PR #1621, opened 2017-09-25
CORDA-205: Supported core types documentation

rick-r3:
This PR introduces the concept of the migration to AMQP (but does not yet detail the format itself) and describes the "core types" that will be supported when we move to AMQP serialization, which are very closely aligned to those supported under Kryo.

### DIFF
--- a/docs/source/serialization.rst
+++ b/docs/source/serialization.rst
@@ -40,3 +40,230 @@ It's reproduced here as an example of both ways you can do this for a couple of 
 
 .. warning:: We will be replacing the use of Kryo in the serialization framework and so additional changes here are
    likely.
+
+AMQP
+====
+
+.. note:: AMQP serialization is not currently live and will be turned on in a future release.
+
+The long term goal is to migrate the current serialization format for everything except checkpoints away from the current
+``Kryo``-based format to a more sustainable and controllable format based on AMQP.  Documentation on that format, and how JVM
+classes are translated to AMQP, will be linked here when it is available.
+
+.. For information on our choice of AMQP 1.0, see :doc:`amqp-choice`.  For detail on how we utilise AMQP 1.0 and represent
+   objects in AMQP types, see :doc:`amqp-format`.
+
+We describe here what is and will be supported in the Corda AMQP format from the perspective
+of CorDapp developers, to allow for CorDapps to take into consideratio the future state.  The AMQP serialization format will of
+course continue to apply the whitelisting functionality that is already in place and described in :doc:`serialization`.
+
+Core Types
+----------
+
+Here we describe the classes and interfaces that the AMQP serialization format will support.
+
+Collection Types
+````````````````
+
+The following collection types are supported.  Any implementation of the following will be mapped to *an* implementation of the interface or class on the other end.
+e.g. If you, for example, use a Guava implementation of a collection it will deserialize as a different implementation,
+but will continue to adhere to the most specific of any of the following interfaces.  You should use only these types
+as the declared types of fields and properties, and not the concrete implementation types.
+
+::
+
+    java.util.Collection
+    java.util.List
+    java.util.Set
+    java.util.SortedSet
+    java.util.NavigableSet
+    java.util.NonEmptySet
+    java.util.Map
+    java.util.SortedMap
+    java.util.NavigableMap
+
+However, we will support the concrete implementation types below explicitly and also as the declared type of a field, as
+a convenience.
+
+::
+
+    java.util.LinkedHashMap
+    java.util.TreeMap
+    java.util.EnumSet
+    java.util.EnumMap (but only if there is at least one entry)
+
+
+JVM primitives
+``````````````
+
+All the primitive types are supported.
+
+::
+
+    boolean
+    byte
+    char
+    double
+    float
+    int
+    long
+    short
+
+Arrays
+``````
+
+We also support arrays of any supported type, primitive or otherwise.
+
+JDK Types
+`````````
+
+The following types are supported from the JDK libraries.
+
+::
+
+    java.io.InputStream
+
+    java.lang.Boolean
+    java.lang.Byte
+    java.lang.Character
+    java.lang.Class
+    java.lang.Double
+    java.lang.Float
+    java.lang.Integer
+    java.lang.Long
+    java.lang.Short
+    java.lang.StackTraceElement
+    java.lang.String
+    java.lang.StringBuffer
+
+    java.math.BigDecimal
+
+    java.security.PublicKey
+
+    java.time.DayOfWeek
+    java.time.Duration
+    java.time.Instant
+    java.time.LocalDate
+    java.time.LocalDateTime
+    java.time.LocalTime
+    java.time.Month
+    java.time.MonthDay
+    java.time.OffsetDateTime
+    java.time.OffsetTime
+    java.time.Period
+    java.time.YearMonth
+    java.time.Year
+    java.time.ZonedDateTime
+    java.time.ZonedId
+    java.time.ZoneOffset
+
+    java.util.BitSet
+    java.util.Currency
+    java.util.UUID
+
+Third Party Types
+`````````````````
+
+The following 3rd party types are supported.
+
+::
+
+    kotlin.Unit
+    kotlin.Pair
+
+    org.apache.activemq.artemis.api.core.SimpleString
+
+    org.bouncycastle.asn1.x500.X500Name
+    org.bouncycastle.cert.X509CertificateHolder
+
+Corda Types
+```````````
+
+Classes and interfaces in the Corda codebase annotated with ``@CordaSerializable`` are of course supported.
+The following un-annotated corda types are supported.
+
+::
+
+    net.corda.core.identity.PartyAndCertificate
+
+All Corda exceptions that are expected to be serialized inherit from ``CordaThrowable`` via either ``CordaException``, for
+checked exceptions, or ``CordaRuntimeException``, for unchecked exceptions.  Any ``Throwable`` that is serialized but does
+not conform to ``CordaThrowable`` will be converted to a ``CordaRuntimeException`` with the original exception type
+and other properties retained within it.
+
+Custom Types
+------------
+
+Here are the rules to adhere to for support of your own types:
+
+Classes
+```````
+
+    #.  A constructor which takes all of the properties that you wish to record in the serialized form.  This is required in
+        order for the serialization framework to reconstruct an instance of your class.
+    #.  If more than one constructor is provided, the serialization framework needs to know which one to use.  The ``@ConstructorForDeserialization``
+        annotation can be used to indicate which one.  For a Kotlin class, without the ``@ConstructorForDeserialization`` annotation, the
+        *primary constructor* will be selected.
+    #.  The class must be compiled with parameter names included in the ``.class`` file.  This is the default in Kotlin
+        but must be turned on in Java (``-parameters`` command line option to ``javac``).
+    #.  A Java Bean getter for each of the properties in the constructor, with the names matching up.  For example, for a constructor
+        parameter ``foo``, there must be a getter called ``getFoo()``.  If the type of ``foo`` is boolean, the getter may
+        optionally be called ``isFoo()``.  This is why the class must be compiled with parameter names turned on.
+    #.  The class is annotated with ``@CordaSerializable``.
+    #.  The declared types of constructor arguments / getters must be supported, and where generics are used the
+        generic parameter must be a supported type, an open wildcard (``*``), or a bounded wildcard which is currently
+        widened to an open wildcard.
+    #.  Any superclass must adhere to the same rules, but can be abstract.
+
+Enums
+`````
+
+    #.  All enums are supported, provided they are annotated with ``@CordaSerializable``.
+
+..  warning::  Use of enums in CorDapps requires potentially deeper consideration than in other application environments
+    due to the challenges of simultaneously upgrading the code on all nodes.  It is therefore important to consider the code
+    evolution perspective, since an older version of the enum code cannot
+    accommodate a newly added element of the enum in a new version of the enum code.  See `Type Evolution`_.  Hence, enums are
+    a good fit for genuinely static data that will *never* change.  e.g. Days of the week is not going to be extended any time
+    soon and is indeed an enum in the Java library.  A Buy or Sell indicator is another.  However, something like
+    Trade Type or Currency Code is likely not, since who's to say a new trade type or currency will not come along soon.  For
+    those it is better to choose another representation: perhaps just a string.
+
+Exceptions
+``````````
+
+The following rules apply to supported ``Throwable`` implementations.
+
+    #.  If you wish for your exception to be serializable and transported type safely it should inherit from either
+        ``CordaException`` or ``CordaRuntimeException``.
+    #.  If not, the ``Throwable`` will deserialize to a ``CordaRuntimeException`` with the details of the original
+        ``Throwable`` contained within it, including the class name of the original ``Throwable``.
+
+Kotlin Objects
+``````````````
+
+    #.  Kotlin ``object`` s are singletons and treated differently.  They are recorded into the stream with no properties
+        and deserialize back to the singleton instance.
+
+Currently, the same is not true of Java singletons, and they will deserialize to new instances of the class.
+
+Future Enhancements
+```````````````````
+
+    #.  Java singleton support.  We will add support for identifying classes which are singletons and identifying the
+        static method responsible for returning the singleton instance.
+    #.  Iterning support.  We will add support for identifying classes that should be resolved against an instances map to avoid
+        creating many duplicate instances that are equal.
+    #.  Enum evolution support.  We *may* introduce an annotation that can be applied to an enum element to indicate that
+        if an unrecognised enum entry is deserialized from a newer version of the code, it should be converted to that
+        element in the older version of the code.  This is dependent on identifying a suitable use case, since it does
+        mutate the data when transported to another node, which could be considered hazardous.
+
+.. Type Evolution:
+
+Type Evolution
+--------------
+
+When we move to AMQP as the serialization format, we will be adding explicit support for interoperability of different versions of the same code.
+We will describe here the rules and features for evolvability as part of a future update to the documentation.
+
--- a/docs/source/serialization.rst
+++ b/docs/source/serialization.rst
@@ -214,6 +214,7 @@ Classes
         generic parameter must be a supported type, an open wildcard (``*``), or a bounded wildcard which is currently
         widened to an open wildcard.
     #.  Any superclass must adhere to the same rules, but can be abstract.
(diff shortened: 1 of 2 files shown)


docs/source/serialization.rst
```diff
@@ -40,3 +40,231 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
 .. warning:: We will be replacing the use of Kryo in the serialization framework and so additional changes here are
    likely.
+
+AMQP
+====
+
+.. note:: AMQP serialization is not currently live and will be turned on in a future release.
+
+The long term goal is to migrate the current serialization format for everything except checkpoints away from the current
+``Kryo``-based format to a more sustainable and controllable format based on AMQP.  Documentation on that format, and how JVM
```
vkolomeyko: Might be worth briefly mentioning what is the problem with Kryo serialisation and why we are steering away from it.

docs/source/serialization.rst
```diff
@@ -40,3 +40,231 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+    java.util.Set
+    java.util.SortedSet
+    java.util.NavigableSet
+    java.util.NonEmptySet
+    java.util.Map
+    java.util.SortedMap
+    java.util.NavigableMap
+
+However, we will support the concrete implementation types below explicitly and also as the declared type of a field, as
+a convenience.
```
vkolomeyko: Might be worth putting a note about generic types within collections. I.e. give some background about fingerprint and compatibility checks we make.

docs/source/serialization.rst
```diff
@@ -40,3 +40,231 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+        and deserialize back to the singleton instance.
+
+Currently, the same is not true of Java singletons, and they will deserialize to new instances of the class.
+
+Future Enhancements
+```````````````````
+
+    #.  Java singleton support.  We will add support for identifying classes which are singletons and identifying the
+        static method responsible for returning the singleton instance.
+    #.  Iterning support.  We will add support for identifying classes that should be resolved against an instances map to avoid
```
vkolomeyko: Interning?

vkolomeyko:
Few minor changes outlined

docs/source/serialization.rst
```diff
@@ -40,3 +40,231 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+Here are the rules to adhere to for support of your own types:
+
+Classes
+```````
+
+    #.  A constructor which takes all of the properties that you wish to record in the serialized form.  This is required in
+        order for the serialization framework to reconstruct an instance of your class.
+    #.  If more than one constructor is provided, the serialization framework needs to know which one to use.  The ``@ConstructorForDeserialization``
+        annotation can be used to indicate which one.  For a Kotlin class, without the ``@ConstructorForDeserialization`` annotation, the
+        *primary constructor* will be selected.
```
fenryka: Might want to mention the DeprecatedConstructorForDeserialization as well here

docs/source/serialization.rst
```diff
@@ -40,3 +40,231 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+    #.  Java singleton support.  We will add support for identifying classes which are singletons and identifying the
+        static method responsible for returning the singleton instance.
+    #.  Iterning support.  We will add support for identifying classes that should be resolved against an instances map to avoid
+        creating many duplicate instances that are equal.
+    #.  Enum evolution support.  We *may* introduce an annotation that can be applied to an enum element to indicate that
+        if an unrecognised enum entry is deserialized from a newer version of the code, it should be converted to that
+        element in the older version of the code.  This is dependent on identifying a suitable use case, since it does
+        mutate the data when transported to another node, which could be considered hazardous.
+
```
fenryka: Doesn't mention the carpenter

docs/source/serialization.rst
```diff
@@ -40,3 +40,231 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+    org.apache.activemq.artemis.api.core.SimpleString
+
+    org.bouncycastle.asn1.x500.X500Name
+    org.bouncycastle.cert.X509CertificateHolder
+
+Corda Types
+```````````
+
+Classes and interfaces in the Corda codebase annotated with ``@CordaSerializable`` are of course supported.
+The following un-annotated corda types are supported.
```
mikehearn: That's weird. Why is it not annotated?

docs/source/serialization.rst
rick-r3: I was saving that for another document, but I'll include an overview here.

docs/source/serialization.rst
rick-r3: I've added some coverage of generics.  After 1.0 it is worth a more detailed section on more complex topics, such as wildcards and type variables.

The fingerprint discussion should come under the type evolution section, which is post 1.0, along with discussion of the carpenter.

docs/source/serialization.rst
rick-r3: As in `String.intern()`.  I'll expand to "internalising", or some such, and maybe add `String.intern()` as a footnote.

docs/source/serialization.rst
rick-r3: Save this for the type evolution section

docs/source/serialization.rst
rick-r3: True.  I think it should get a mention.  A document on it later is warranted, along with the type evolution work.

docs/source/serialization.rst
rick-r3: It has custom serializers in both Kryo and AMQP and thus is implicitly serializable.  We might as well annotate it and get rid of this section.

docs/source/serialization.rst
```diff
@@ -40,3 +40,248 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+The long term goal is to migrate the current serialization format for everything except checkpoints away from the current
+``Kryo``-based format to a more sustainable, self-describing and controllable format based on AMQP 1.0.  The primary drivers for that move are:
+
+    #.  A desire to have a schema describing what has been serialized along-side the actual data:
+        #.  To assist with versioning, both in terms of being able to interpret long ago archived data (e.g. trades from
+            a decade ago, long after the code has changed) and between differing code versions.
+        #.  To make it easier to write user interfaces that can navigate the serialized form of data.
+        #.  To support cross platform (non-JVM) interaction, where the format of a class file is not so easily interpreted.
+    #.  A desire to use a documented and static wire format that is platform independent, and is not subject to change with
```
gendal: i thought a driver was also security?  if so, should we state that up-front and outline the argument?

docs/source/serialization.rst
```diff
@@ -40,3 +40,248 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+    #.  A desire to support open-ended polymorphism, where the number of subclasses of a superclass can expand over time
+        and do not need to be defined in the schema *upfront*, which is key to many Corda concepts, such as contract states.
+
+Documentation on that format, and how JVM classes are translated to AMQP, will be linked here when it is available.
+
+.. For information on our choice of AMQP 1.0, see :doc:`amqp-choice`.  For detail on how we utilise AMQP 1.0 and represent
+   objects in AMQP types, see :doc:`amqp-format`.
+
+We describe here what is and will be supported in the Corda AMQP format from the perspective
+of CorDapp developers, to allow for CorDapps to take into consideratio the future state.  The AMQP serialization format will of
```
gendal: consideration

docs/source/serialization.rst
```diff
@@ -40,3 +40,248 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
 .. warning:: We will be replacing the use of Kryo in the serialization framework and so additional changes here are
    likely.
+
+AMQP
+====
+
+.. note:: AMQP serialization is not currently live and will be turned on in a future release.
+
+The long term goal is to migrate the current serialization format for everything except checkpoints away from the current
+``Kryo``-based format to a more sustainable, self-describing and controllable format based on AMQP 1.0.  The primary drivers for that move are:
```
fenryka: I'm a fan of the Oxford comma... it looks so wrong without it although I know people will disagree

docs/source/serialization.rst
fenryka: Since it's not on in version one do we want a whole document or just a paragraph stating "if the class isn't found on the classpath then one will be carpetned up for you"

If the former, I should probably start writing that

docs/source/serialization.rst
rick-r3: A lot of the benefit was already realised by the whitelisting work, but there is some security benefit to the use of constructors to create objects during deserialization, rather than peeking and poking inside objects as Kryo does.

docs/source/serialization.rst
rick-r3: Done.

docs/source/serialization.rst
rick-r3: I had to google...

docs/source/serialization.rst
rick-r3: I put a paragraph on the carpenter concept above which is good for now.  The next release is all about wire stability and we should write documents on the formats, type evolution, carpenter etc for that.

docs/source/serialization.rst
gendal: ok - thanks.

vkolomeyko:
Looks good to me

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
 
 .. warning:: We will be replacing the use of Kryo in the serialization framework and so additional changes here are
    likely.
+
+AMQP
```
gendal: Immediate question: who is the intended reader of this?  The surrounding context implies the reader should be developers who need to know how to write/annotate their code to play nicely with respect to serialisation, etc... effectively a must-read/must-know piece of info.      So my question would be:  is the section that follows a must-read (as in "you really need to know this because bad things will happen if you don't take it into account") or is it a note-to-one's-future-self?  Or something else?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+====
+
+.. note:: AMQP serialization is not currently live and will be turned on in a future release.
```
gendal: This is the first time AMQP _serialization_ is mentioned.  The reader would think... "oh...  there's a note intended to pre-empt my being confused... but I wasn't confused until I read it".  Change the section header to match? ie call it AMQP Serialisation?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+The long term goal is to migrate the current serialization format for everything except checkpoints away from the current
+``Kryo``-based format to a more sustainable, self-describing and controllable format based on AMQP 1.0.  The primary drivers for that move are:
```
gendal: Do we expect the developer experience to change? (e.g. does what follows invalidate anything that was said above about annotations?)  Either way, should we say whether it does or not? If the act of turning this one will have developer-visible effects, should we tell them so?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+        data poked directy into their fields without an opportunity to validate consistency or intercept attempts to manipulate
+        supposed invariants.
+
+Documentation on that format, and how JVM classes are translated to AMQP, will be linked here when it is available.
+
+.. For information on our choice of AMQP 1.0, see :doc:`amqp-choice`.  For detail on how we utilise AMQP 1.0 and represent
+   objects in AMQP types, see :doc:`amqp-format`.
+
+We describe here what is and will be supported in the Corda AMQP format from the perspective
+of CorDapp developers, to allow for CorDapps to take into consideration the future state.  The AMQP serialization format will of
```
gendal: "take into consideration"... in what way?  Are we about to tell them that things that are currently supported will become unsupported?  The opposite?  That apps will break in exciting ways?  ie if we've done our job well, what's the reason why they should be reading any of this section at all?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+Core Types
+----------
+
+Here we describe the classes and interfaces that the AMQP serialization format will support.
+
+Collection Types
+````````````````
+
+The following collection types are supported.  Any implementation of the following will be mapped to *an* implementation of the interface or class on the other end.
```
gendal: Are there any headline/noteworthy examples of things we won't be supporting?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+Corda Types
+```````````
+
+Classes and interfaces in the Corda codebase annotated with ``@CordaSerializable`` are of course supported.
+
+All Corda exceptions that are expected to be serialized inherit from ``CordaThrowable`` via either ``CordaException``, for
+checked exceptions, or ``CordaRuntimeException``, for unchecked exceptions.  Any ``Throwable`` that is serialized but does
+not conform to ``CordaThrowable`` will be converted to a ``CordaRuntimeException`` with the original exception type
+and other properties retained within it.
```
gendal: as strings?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+Custom Types
+------------
+
+Here are the rules to adhere to for support of your own types:
```
gendal: should we highlight up top that this section is here and that it should be read by CorDapp developers today to maximise chances of things not breaking when we turn this on?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+Classes
+```````
+
+    #.  A constructor which takes all of the properties that you wish to record in the serialized form.  This is required in
+        order for the serialization framework to reconstruct an instance of your class.
+    #.  If more than one constructor is provided, the serialization framework needs to know which one to use.  The ``@ConstructorForDeserialization``
+        annotation can be used to indicate which one.  For a Kotlin class, without the ``@ConstructorForDeserialization`` annotation, the
+        *primary constructor* will be selected.
+    #.  The class must be compiled with parameter names included in the ``.class`` file.  This is the default in Kotlin
```
gendal: feels like this is something most people will forget to do most of the time.  does anything bad happen if so or do they discover they've caused a problem almost immediately?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+        but must be turned on in Java (``-parameters`` command line option to ``javac``).
+    #.  A Java Bean getter for each of the properties in the constructor, with the names matching up.  For example, for a constructor
+        parameter ``foo``, there must be a getter called ``getFoo()``.  If the type of ``foo`` is boolean, the getter may
+        optionally be called ``isFoo()``.  This is why the class must be compiled with parameter names turned on.
+    #.  The class is annotated with ``@CordaSerializable``.
+    #.  The declared types of constructor arguments / getters must be supported, and where generics are used the
+        generic parameter must be a supported type, an open wildcard (``*``), or a bounded wildcard which is currently
+        widened to an open wildcard.
+    #.  Any superclass must adhere to the same rules, but can be abstract.
+    #.  Object graph cycles are not supported, so an object cannot refer to itself, directly or indirectly.
```
gendal: actually - this whole list feels like a breeding ground of opportunities for developers to mess up. if they don't follow this religiously, is it a big problem in the scheme of things (ie it gets detected soon enough) or does it introduce the possibility of horrible problems when we move to version-to-version wire compatibility and/or non-JVM runtimes, etc?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+    #.  All enums are supported, provided they are annotated with ``@CordaSerializable``.
+
+..  warning::  Use of enums in CorDapps requires potentially deeper consideration than in other application environments
+    due to the challenges of simultaneously upgrading the code on all nodes.  It is therefore important to consider the code
+    evolution perspective, since an older version of the enum code cannot
+    accommodate a newly added element of the enum in a new version of the enum code.  See `Type Evolution`_.  Hence, enums are
+    a good fit for genuinely static data that will *never* change.  e.g. Days of the week is not going to be extended any time
+    soon and is indeed an enum in the Java library.  A Buy or Sell indicator is another.  However, something like
+    Trade Type or Currency Code is likely not, since who's to say a new trade type or currency will not come along soon.  For
+    those it is better to choose another representation: perhaps just a string.
```
gendal: ooo... this is interesting.  I would have used a semicolon... but a colon isn't obviously wrong....

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+Exceptions
+``````````
+
+The following rules apply to supported ``Throwable`` implementations.
+
+    #.  If you wish for your exception to be serializable and transported type safely it should inherit from either
```
gendal: we assume a lot of the reader... ie that we ship serialised java objects over the p2p and rpc interfaces, etc.  you could argue.. "well... duhhhh... obviously this is how it works.."...  but I wonder if here or at the top we should include a line or so just to "remind" people why we're talking about all this stuff?  so somebody diving in can catch up

gendal: perhaps more clearly:  "if you wish for your exception to be..."...  how would they know if they wished for this?    should we say something like: "if you throw exceptions that you wish to have received by recipients across the network (rpc/p2p), ..."  (loose wording, i know)

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+
+    #.  Kotlin ``object`` s are singletons and treated differently.  They are recorded into the stream with no properties
+        and deserialize back to the singleton instance.
+
+Currently, the same is not true of Java singletons, and they will deserialize to new instances of the class.
+
+The Carpenter
+`````````````
+
+We will support a class carpenter that can dynamically manufacture classes from the supplied schema when deserializing
```
gendal: is class carpenter a common term of art?  put in quotes if not?

docs/source/serialization.rst
```diff
@@ -40,3 +40,251 @@ It's reproduced here as an example of both ways you can do this for a couple of
...
+in the JVM without the supporting classes on the classpath.  This can be useful where other components might expect to
+be able to use reflection over the deserialized data, and also for ensuring classes not on the classpath can be
+deserialized without loading potentially malicious code dynamically without security review outside of a fully sandboxed
+environment.  A more detailed discussion of the carpenter will be provided in a future update to the documentation.
```
gendal: is @fenryka not planning to write something this week?

PR #1623, opened 2017-09-25
[CORDA-442]  Add filesystem polling to nodeInfos on disk.

al-r3:
Add the logic in node to poll for new serialized nodes to appear on disk.

Newly discovered nodes are automatically added to the PersistentNetworkMapCache


gendal:
not sure if this PR would be the right place but this seems like a big deal... ie the push from the network map service goes away and the (only?) way to tell a running node about a new peer is to copy a file into this directory.  do we / will we document how to do this? e.g. how to generate the NodeInfo files, etc, etc?

al-r3:
The decision to remove the network map service on nodes was taken here:
https://r3-cev.atlassian.net/wiki/spaces/<https://r3-cev.atlassian.net/wiki/spaces/AWG/pages/127710793/Design+doc+network+map+upgrade>AWG<https://r3-cev.atlassian.net/wiki/spaces/AWG/pages/127710793/Design+doc+network+map+upgrade>/pages/127710793/Design+doc+network+map+upgrade<https://r3-cev.atlassian.net/wiki/spaces/AWG/pages/127710793/Design+doc+network+map+upgrade>

Note that no manual action will be required for the demos/demobench/other tools, they'll all be updated to generate the nodeinfos and copy them to the appropriate folders.

In production nodes will talk with doorman to find about other nodes, further details are being finalized to address the finastra use case.


Al


kasiastreich:
@gendal This is a part of work needed for demos (cordformation in this case) to have NodeInfos immediately visible in NetworkMapCache without polling network map service. Definitely not intended to use in normal production environment.

gendal:
@kasiastreich @al-r3  ok... so we don't expect anybody who is "not us" to use this or know about it?

docs/source/changelog.rst
```diff
@@ -18,6 +18,7 @@ UNRELEASED
    ``NetworkMapCache::notaryIdentities`` list to enable to lookup for notary parties known to the network. Related ``CordaRPCOps::notaryIdentities``
    was introduced. Other special nodes parties like Oracles or Regulators need to be specified directly in CorDapp or flow.
   * Moved ``ServiceType`` and ``ServiceInfo`` to ``net.corda.nodeapi`` package as services are only required on node startup.
+  * Nodes now poll the filesystem for new serialized ``NodeInfo`` in the ``additional-node-info`` directory.
```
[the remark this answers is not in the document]
al-r3: This isn't specific for Cordformation, demoBench and other tools will use it.
The polling will also happen regardless of whether a node was deployed with Cordformation or not.

al-r3:
Sorry about this, but the dependent PR has been open for 9 days, what was I supposed to do in the meanwhile?


shamsasari:
Once you've rebased and resolved the conflicts I'll take a look a this.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,59 +3,76 @@ package net.corda.node.services.network
...
+         * Saves the given [NodeInfo] to a path.
+         * The node is 'encoded' as a SignedData<NodeInfo>, signed with the owning key of its first identity.
+         * The name of the written file will be "nodeInfo-" followed by the hash of the content. The hash in the filename
+         * is used so that one can freely copy these files without fearing to overwrite another one.
+         *
+         * @param path the path where to write the file, if non-existent it will be created.
+         * @param nodeInfo the NodeInfo to serialize.
+         * @param keyManager a KeyManagementService used to sign the NodeInfo data.
+         */
+        fun saveToFile(path: Path, nodeInfo: NodeInfo, keyManager: KeyManagementService) {
```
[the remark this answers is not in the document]
al-r3: saveToFile is in the companion object, saveToFile is unrelated with the other code in the class.
saveToFile is meant to be called from the node startup logic, the rest of the class is called by the network map logic.
All the rest of the code is used just from the network map. 
I might even move this out of the companion object and make this a private function in this file (or another one)

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -87,16 +89,20 @@ open class PersistentNetworkMapCache(private val serviceHub: ServiceHubInternal)
...
     }
 
     private fun loadFromFiles() {
         logger.info("Loading network map from files..")
-        for (node in NodeInfoSerializer().loadFromDirectory(serviceHub.configuration.baseDirectory)) {
+        for (node in nodeInfoSerializer.loadFromDirectory()) {
             addNode(node)
         }
+
+        nodeInfoSerializer.directoryObservable().subscribe { node -> addNode(node)}
```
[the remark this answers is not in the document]
al-r3: the observable documentation say one shouldn't use it for IO blocking operations (this doesn't block)
http://reactivex.io/RxJava/javadoc/rx/schedulers/Schedulers.html#computation--
So this should comply with the RxJava requirement.

Also using a full thread for this seems overkill, we'd have all the overhead of a Thread which would just sit there waiting all the time

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -65,21 +82,80 @@ class NodeInfoSerializer {
...
 
-    private fun loadFromFile(file: File): NodeInfo {
-        val signedData = ByteSequence.of(file.readBytes()).deserialize<SignedData<NodeInfo>>()
-        return signedData.verified()
+    /**
+     * @return an [Observable] returning new [NodeInfo]s, there is no guarantee that the same value isn't returned more
+     *      than once.
+     */
+    fun directoryObservable() : Observable<NodeInfo> {
+        return Observable.interval(5, TimeUnit.SECONDS, scheduler)
```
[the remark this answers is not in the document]
al-r3: (see my other reply to the threading issue)

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -65,21 +82,80 @@ class NodeInfoSerializer {
...
     }
 
-    private fun loadFromFile(file: File): NodeInfo {
-        val signedData = ByteSequence.of(file.readBytes()).deserialize<SignedData<NodeInfo>>()
-        return signedData.verified()
+    /**
+     * @return an [Observable] returning new [NodeInfo]s, there is no guarantee that the same value isn't returned more
+     *      than once.
+     */
+    fun directoryObservable() : Observable<NodeInfo> {
```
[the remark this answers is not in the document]
al-r3: created a single function which does reading & polling

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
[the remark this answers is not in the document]
al-r3: discussed off-line, done

node/src/test/kotlin/net/corda/node/services/network/NodeInfoSerializerTest.kt
```diff
@@ -36,11 +40,12 @@ class NodeInfoSerializerTest : NodeBasedTest() {
     fun start() {
         val identityService = InMemoryIdentityService(trustRoot = DEV_TRUST_ROOT)
         keyManagementService = MockKeyManagementService(identityService, ALICE_KEY)
+        nodeInfoSerializer = NodeInfoSerializer(folder.root.toPath(), scheduler)
     }
 
     @Test
     fun `save a NodeInfo`() {
-        nodeInfoSerializer.saveToFile(folder.root.toPath(), nodeInfo, keyManagementService)
+        NodeInfoSerializer.saveToFile(folder.root.toPath(), nodeInfo, keyManagementService)
```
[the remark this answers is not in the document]
al-r3: discussed offline, it's in the companion object

docs/source/changelog.rst
```diff
@@ -11,7 +11,9 @@ UNRELEASED
   * Cordform may not specify a value for ``NetworkMap``, when that happens, during the task execution the following happens:
     1. Each node is started and its signed serialized NodeInfo is written to disk in the node base directory.
     2. Every serialized ``NodeInfo`` above is copied in every other node "additional-node-info" folder under the NodeInfo folder.
-  * Nodes read all the nodes stored in ``additional-node-info`` when the ``NetworkMapService`` starts up.
+
+* Nodes read all the nodes stored in ``additional-node-info`` when the ``NetworkMapService`` starts up.
+* Nodes now poll the filesystem for new serialized ``NodeInfo`` in the ``additional-node-info`` directory.
```
shamsasari: The use of the word "now" implies the additional-node-info feature already exists in the current release. Can you combine all these three bullet points into one so that the entire thing comes across as a new feature.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,83 +3,157 @@ package net.corda.node.services.network
...
+        logger.info("Successfully read ${result.size} NodeInfo files.")
         return result
     }
 
-    private fun loadFromFile(file: File): NodeInfo {
-        val signedData = ByteSequence.of(file.readBytes()).deserialize<SignedData<NodeInfo>>()
-        return signedData.verified()
+    // Polls the watchService for changes to nodeInfoDirectory, return all the newly read NodeInfos.
+    private fun pollWatch() : List<NodeInfo> {
+        val files = ArrayList<Path>()
```
shamsasari: Easy optimisation by moving this down to where it's first needed.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,83 +3,157 @@ package net.corda.node.services.network
...
-        logger.info("Succesfully read ${result.size} NodeInfo files.")
+        logger.info("Successfully read ${result.size} NodeInfo files.")
         return result
     }
 
-    private fun loadFromFile(file: File): NodeInfo {
-        val signedData = ByteSequence.of(file.readBytes()).deserialize<SignedData<NodeInfo>>()
-        return signedData.verified()
+    // Polls the watchService for changes to nodeInfoDirectory, return all the newly read NodeInfos.
+    private fun pollWatch() : List<NodeInfo> {
```
shamsasari: `pollDirectory`?

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,83 +3,157 @@ package net.corda.node.services.network
...
         for (path in Files.list(nodeInfoDirectory)) {
-            val file = path.toFile()
-            if (file.isFile) {
-                try {
-                    logger.info("Reading NodeInfo from file: $file")
-                    val nodeInfo = loadFromFile(file)
-                    result.add(nodeInfo)
-                } catch (e: Exception) {
-                    logger.error("Exception parsing NodeInfo from file. $file" , e)
+            if (path.toFile().isFile) {
```
shamsasari: `path.isRegularFile()`

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,83 +3,157 @@ package net.corda.node.services.network
...
      */
-    fun loadFromDirectory(nodePath: Path): List<NodeInfo> {
+    private fun loadFromDirectory(): List<NodeInfo> {
         val result = mutableListOf<NodeInfo>()
         val nodeInfoDirectory = nodePath / CordformNode.NODE_INFO_DIRECTORY
         if (!nodeInfoDirectory.isDirectory()) {
             logger.info("$nodeInfoDirectory isn't a Directory, not loading NodeInfo from files")
             return result
         }
         for (path in Files.list(nodeInfoDirectory)) {
```
shamsasari: I'm not sure if this for-loop is automatically releasing the resources of `list`. Use the `list` extension method we have which closes the Stream for you.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,83 +3,157 @@ package net.corda.node.services.network
...
-            file.writeBytes(signedData.serialize().bytes)
-        } catch (e : Exception) {
-            logger.warn("Couldn't write node info to file: $e")
-        }
+    fun directoryObservable() : Observable<NodeInfo> {
+        val pollForFiles= Observable.interval(5, TimeUnit.SECONDS, scheduler)
+                .flatMapIterable { pollWatch() }
+
+        val readCurrentFiles = Observable.from(loadFromDirectory())
+
```
shamsasari: Too many blank lines

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
shamsasari: Actually, why not just have this be a Set so you don't have to do the `toSet` at the end.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,83 +3,157 @@ package net.corda.node.services.network
...
-            val file = path.toFile()
-            if (file.isFile) {
-                try {
-                    logger.info("Reading NodeInfo from file: $file")
-                    val nodeInfo = loadFromFile(file)
-                    result.add(nodeInfo)
-                } catch (e: Exception) {
-                    logger.error("Exception parsing NodeInfo from file. $file" , e)
+            if (path.toFile().isFile) {
+                processFile(path)?.let {
```
shamsasari: You're not making use of the `let`'s return value so I'm not sure if it's worth having.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
```diff
@@ -3,83 +3,157 @@ package net.corda.node.services.network
...
+ * Class containing the logic to
+ * - Serialize and de-serialize a [NodeInfo] to disk and reading it back.
+ * - Poll a directory for new serialized [NodeInfo]
+ *
+ * @param path the base path of a node.
+ * @param scheduler a [Scheduler] for the rx [Observable] returned by [directoryObservable], this is mainly useful for
+ *        testing. It defaults to the io scheduler which is the appropriate value for production uses.
  */
-class NodeInfoSerializer {
+class NodeInfoSerializer(private val nodePath: Path,
```
shamsasari: This class isn't a serialiser. A better name would be `NodeInfoWatcher` or similar.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
al-r3: I've done it, but it seems less readable to me.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoSerializer.kt
al-r3: no, but I'm using let to run a block of code only when a variable is not null.
It's a kotlin suggested pattern,
https://kotlinlang.org/docs/reference/idioms.html#execute-if-not-null
how would you better express that?

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -0,0 +1,153 @@
...
+     * Signatures are checked before returning a value.
+     *
+     * @return a list of [NodeInfo]s
+     */
+    private fun loadFromDirectory(): List<NodeInfo> {
+        val result = mutableListOf<NodeInfo>()
+        val nodeInfoDirectory = nodePath / CordformNode.NODE_INFO_DIRECTORY
+        if (!nodeInfoDirectory.isDirectory()) {
+            logger.info("$nodeInfoDirectory isn't a Directory, not loading NodeInfo from files")
+            return result
```
shamsasari: return emptyList

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -0,0 +1,153 @@
...
+        }
+        nodeInfoDirectory.list { paths ->
```
shamsasari: To address both your comments, how about something like:
```
        val result = nodeInfoDirectory.list { files ->
            files.filter { it.isRegularFile() }.distinct().map { processFile(it) }.filter { it != null }.toList()
        }
```

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Read all the files contained in [nodePath] / [CordformNode.NODE_INFO_DIRECTORY] and keep watching
+     * the folder for further updates.
+     *
+     * @return an [Observable] returning [NodeInfo]s, there is no guarantee that the same value isn't returned more
+     *      than once.
+     */
+    fun pollDirectory() : Observable<NodeInfo> {
+        val pollForFiles= Observable.interval(5, TimeUnit.SECONDS, scheduler)
```
shamsasari: whitespace before =

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
shamsasari: The `distinct()` call isn't needed

al-r3: Done, I've moved the filterNotNull after the toList, to have the right type.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -0,0 +1,150 @@
...
+ * Class containing the logic to
+ * - Serialize and de-serialize a [NodeInfo] to disk and reading it back.
+ * - Poll a directory for new serialized [NodeInfo]
+ *
+ * @param path the base path of a node.
+ * @param scheduler a [Scheduler] for the rx [Observable] returned by [pollDirectory], this is mainly useful for
+ *        testing. It defaults to the io scheduler which is the appropriate value for production uses.
+ */
+class NodeInfoWatcher(private val nodePath: Path,
+                      private val scheduler: Scheduler = Schedulers.io()) {
```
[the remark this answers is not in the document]
al-r3: This code will run maybe 10 times in whole life of a node.
But it's quite likely it will run zero times, are we sure that even discussing the performance implications of the various ways of implementing it is worth the time?

PR #1625, opened 2017-09-25
fix several elements associated with the same key bug

patrickkuo:
fix for https://r3-cev.atlassian.net/browse/CORDA-609


josecoll:
LGTM!

PR #1627, opened 2017-09-25
Vault Query API parameter rename to aid clarity

josecoll:
_ContractType_ renamed to _ContractStateType_


mnesbit:
Does HibernateQueryCriteriaParser need any corrections to line up?

PR #1630, opened 2017-09-25
CORDA-599 Fix circular dependency between vault and SH

andr3ej:


testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -42,7 +44,12 @@ import java.util.*
  * A singleton utility that only provides a mock identity, key and storage service. However, this is sufficient for
  * building chains of transactions and verifying them. It isn't sufficient for testing flows however.
  */
-open class MockServices(cordappPackages: List<String> = emptyList(), vararg val keys: KeyPair) : ServiceHub {
+open class MockServices(
+        cordappPackages: List<String> = emptyList(),
+        vararg val keys: KeyPair,
```
mnesbit: Varargs should be last

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -158,8 +164,8 @@ open class MockServices(cordappPackages: List<String> = emptyList(), vararg val
 
     lateinit var hibernatePersister: HibernateObserver
 
-    fun makeVaultService(hibernateConfig: HibernateConfiguration = HibernateConfiguration( { NodeSchemaService() }, makeTestDatabaseProperties(), { identityService })): VaultService {
-        val vaultService = NodeVaultService(this, hibernateConfig)
+    fun makeVaultService(hibernateConfig: HibernateConfiguration = HibernateConfiguration( { NodeSchemaService() }, makeTestDatabaseProperties(), { identityService })): NodeVaultService {
```
mnesbit: I did do the VaultService intentionally. Might need to be a VaultServiceInternal api, but I think the vault will probably have to have pluggable implementations.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
andr3ej: Done, I've had to add a constructor to avoid adding "keys = " in loads of places.

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
```diff
@@ -131,7 +133,7 @@ data class SignedTransaction(val txBits: SerializedBytes<CoreTransaction>,
      */
     @JvmOverloads
     @Throws(SignatureException::class, AttachmentResolutionException::class, TransactionResolutionException::class)
-    fun toLedgerTransaction(services: ServiceHub, checkSufficientSignatures: Boolean = true): LedgerTransaction {
```
Clintonio: I believe this counts as an API change. Will defer to @mikehearn since this is the first one of it's kind.

andr3ej:
If it's source compatibility we want, the changes are OK:
* loadState is still a method on ServicesForResolution (via inheritance)
* Params that used to take ServiceHub are still satisfied by args implementing ServiceHub, and we don't expect users to implement the interfaces having those methods

mikehearn:
Why not just introduce an overload that takes ServiceHub as well, and passes it straight through?

andr3ej:
Anyone can add that without breaking API if it becomes necessary.

mikehearn:
So why not do it? I don't understand why we're considering a ABI break here when it's not necessary.

andr3ej:
Surely as we only enforce source not binary stability, ABI will inevitably break somehow between this release and the next, so cordapps will need to be recompiled anyway.

mikehearn:
No, it's not inevitable. It's always the result of choices we make. We relaxed ABI compatibility in 1.0 only to give ourselves an out if we found we *had* to do it, but in this case we clearly don't, so we may as well get ourselves into the habit of preserving the ABI as well - especially as in this case it's so easy.

andr3ej:
@Clintonio @mikehearn can this be approved now? I've fixed the ABI, and the only thing the tool is still complaining about is loadState which isn't really a break.

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -33,10 +34,24 @@ import org.hibernate.Session
 import rx.Observable
 import rx.subjects.PublishSubject
 import java.security.PublicKey
+import java.time.Clock
 import java.time.Instant
 import java.util.*
 import javax.persistence.Tuple
 
+interface VaultServiceInternal : VaultService {
```
mnesbit: Ideally intra-node interfaces should be in node/services/api. The intent being that we interact between system components using a well defined internal API, rather than concrete types.

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -96,7 +95,7 @@ class NodeSchedulerServiceTest : SingletonSerializeAsToken() {
                     overrideClock = testClock,
                     keyManagement = kms,
                     network = mockMessagingService), TestReference {
-                override val vaultService: VaultService = NodeVaultService(this, database.hibernateConfig)
+                override val vaultService = NodeVaultService(testClock, kms, stateLoader, database.hibernateConfig)
```
mnesbit: Doesn't matter much as this is a test, but I do like overridden properties to have a declared type. Otherwise we tend to leak concrete types.

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
andr3ej: Done

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
andr3ej: Done. I think I removed the type to self-document "I don't care what the declared type is in this test".

PR #1631, opened 2017-09-25
CORDA-540: Ensure that covariance of type is handled correctly when serializing with AMQP

vkolomeyko:

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
@@ -152,7 +152,7 @@ class DeserializationInput(internal val serializerFactory: SerializerFactory) {
     private fun Type.materiallyEquivalentTo(that: Type): Boolean =
         when(that) {
             is ParameterizedType -> asClass() == that.asClass()
-            is TypeVariable<*> -> isSubClassOf(that.bounds.single())
+            is TypeVariable<*> -> isSubClassOf(that.bounds.first())
             else -> false
         }
 }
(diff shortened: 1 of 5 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
```diff
@@ -142,10 +143,16 @@ class DeserializationInput(internal val serializerFactory: SerializerFactory) {
...
+     *  * [java.util.List<Command<?>>] and [java.lang.Class<? extends net.corda.core.contracts.Contract>]
+     *  * [T : Parent] and [Parent]
+     *
      * In the future tighter control might be needed
      */
     private fun Type.materiallyEquivalentTo(that: Type): Boolean =
-            asClass() == that.asClass() && that is ParameterizedType
+        when(that) {
+            is ParameterizedType -> asClass() == that.asClass()
+            is TypeVariable<*> -> isSubClassOf(that.bounds.single())
```
rick-r3: `single()` will error if there are two bounds.  Is that what we want?  Or do we just return false?  e.g.

`isSubClassOf(that.bounds.first()) && that.bound.size == 1`

vkolomeyko: Agree - will change to `first()`

node-api/src/test/java/net/corda/nodeapi/internal/serialization/amqp/ListsSerializationJavaTest.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        public void setValue(int value) {
+            this.value = value;
+        }
+    }
+
+    @CordaSerializable
+    public static class CovariantContainer<T extends Parent> {
+        private List<T> content;
```
rick-r3: Should this not have a constructor which takes content?

vkolomeyko: Yes, I have tried few options in Java including having constructor, doesn't seem to work. Will continue today.

PR #1632, opened 2017-09-25
Clean up of Dokka comments

rnicoll:


core/src/main/kotlin/net/corda/core/contracts/Amount.kt
```diff
@@ -249,8 +255,9 @@ data class Amount<T : Any>(val quantity: Long, val displayTokenSize: BigDecimal,
 
     /**
      * Convert a currency [Amount] to a decimal representation. For example, with an amount with a quantity
-     * of "1234" GBP, returns "12.34". The precise representation is controlled by the displayTokenSize,
-     * which determines the size of a single token and controls the trailing decimal places via it's scale property.
+     * of "1234" GBP, returns "12.34". The precise representation is controlled by the display token size (
```
mnesbit: I think it would be helpful to educate people that currencies DO NOT always have 2 decimal places! See http://www.londonfx.co.uk/ccylist.html which shows that they typically have 0, 2, or even 3 decimal places and that we do actually handle this. If people want to display stuff differently they should have their ViewModel use Amount.toDecimal() and then sort display conversion from the BigDecimal returned.

rnicoll:
@mnesbit ready for re-review

PR #1636, opened 2017-09-25
Extract identity helpers

rick-r3:
Up for debate whether we want to do this, but we need to decide fast if want to get in for RC2.  Move identity helpers off `ServiceHub` into standalone helper.

- Rename `excludeMe` to `excludeHostNode`.
- Downside, they can be harder to find via code complete in Java IDE.
- Upside, we don't add baggage to `ServiceHub` for methods relating to confidential identity which is in a little bit of flux still.


josecoll:
I certainly think this is a good move - as all the above methods have been dropped into _ServiceHub_ in a rather ad hoc fashion (including numerous renaming's).

PR #1639, opened 2017-09-25
Improve NodeTabView in DemoBench

kasiastreich:
Reflect recent changes with advertised services removal.


tools/demobench/src/main/kotlin/net/corda/demobench/model/ServiceController.kt
```diff
@@ -8,31 +8,32 @@ import java.util.logging.Level
 
 class ServiceController(resourceName: String = "/services.conf") : Controller() {
 
-    val services: List<String> = loadConf(resources.url(resourceName))
+    val services: HashMap<String, String> = loadConf(resources.url(resourceName))
```
chrisr3: `Map` rather than `HashMap`. There's no need to expose the implementation, and we don't want this map of serviecs to be mutable anyway.

tools/demobench/src/main/kotlin/net/corda/demobench/model/ServiceController.kt
```diff
@@ -8,31 +8,32 @@ import java.util.logging.Level
...
+    val notaries: Map<String, String> = services.filter { it.value.startsWith("corda.notary.") }
+
+    val issuers: Map<String, String> = services.filter { it.value.startsWith("corda.issuer.") }
 
     /*
      * Load our list of known extra Corda services.
      */
-    private fun loadConf(url: URL?): List<String> {
+    private fun loadConf(url: URL?): HashMap<String, String> {
         return if (url == null) {
```
chrisr3: `Map` rather than `HashMap`. We want this function to return something immutable.

tools/demobench/src/main/kotlin/net/corda/demobench/model/ServiceController.kt
```diff
@@ -8,31 +8,32 @@ import java.util.logging.Level
...
-            emptyList()
+            HashMap()
         } else {
```
chrisr3: `emptyMap()`

tools/demobench/src/main/kotlin/net/corda/demobench/model/ServiceController.kt
```diff
@@ -8,31 +8,32 @@ import java.util.logging.Level
...
             try {
-                val set = sortedSetOf<String>()
+                val map = HashMap<String, String>()
                 InputStreamReader(url.openStream()).useLines { sq ->
```
chrisr3: `linkedMapOf<String, String>()` to ensure consistent ordering of the contents.

tools/demobench/src/main/kotlin/net/corda/demobench/model/ServiceController.kt
```diff
@@ -8,31 +8,32 @@ import java.util.logging.Level
...
+                        map[service[1]] = service[0]
                         log.info("Supports: $service")
                     }
+                    map
                 }
-                set.toList()
             } catch (e: IOException) {
                 log.log(Level.SEVERE, "Failed to load $url: ${e.message}", e)
-                emptyList<String>()
+                HashMap<String, String>()
```
chrisr3: `emptyMap<String, String>()`

tools/demobench/src/main/kotlin/net/corda/demobench/model/ServiceController.kt
```diff
@@ -8,31 +8,32 @@ import java.util.logging.Level
...
             try {
-                val set = sortedSetOf<String>()
+                val map = HashMap<String, String>()
                 InputStreamReader(url.openStream()).useLines { sq ->
                     sq.forEach { line ->
-                        val service = line.trim()
-                        set.add(service)
-
+                        val service = line.split(":").map { it.trim() }
+                        map[service[1]] = service[0]
```
chrisr3: Not to be picky or anything, but what if `service.conf` is corrupt? You have assumed that there will always be 2 tokens per line, but anything more or less would be a problem that we could log.

chrisr3:
BTW, have you just made `demobench-configure-bank.png` and `demobench-initial.png` out-of-date?
Do you have GIMP installed ;-)?

kasiastreich:
Comments addressed. I will update screen shots after I will get final blessing from @mikehearn 

**Bank**
screenshot from 2017-09-26 11-04-41

**Notary**
screenshot from 2017-09-26 11-04-01

mikehearn:
What happens if I tick both validating and non-validating? Is there a check for that?

kasiastreich:
There never was a check for that. Can add.

kasiastreich:
Please re-review.

tools/demobench/src/main/kotlin/net/corda/demobench/model/ServiceController.kt
```diff
@@ -8,31 +8,37 @@ import java.util.logging.Level
...
+                val map = linkedMapOf<String, String>()
                 InputStreamReader(url.openStream()).useLines { sq ->
                     sq.forEach { line ->
-                        val service = line.trim()
-                        set.add(service)
-
-                        log.info("Supports: $service")
+                        val service = line.split(":").map { it.trim() }
+                        if (service.size != 2) {
+                            log.info("Encountered corrupted line while reading the service config from: $url")
```
chrisr3: Surely this is a warning? And perhaps log the contents of the line too? E.g.
```kotlin
"Encountered corrupt line '$line' while reading services..."
```

chrisr3:
It's a _small_ thing, but it would be nice if `$line` were in quotes (or similar) within that log line, so that it was clear from the log message where `$line` started and ended.

PR #1644, opened 2017-09-25
Renamed "plugins" dir to "cordapps"

Clintonio:
Why: because it should have been for a long time now.


chrisr3:
Well, it obviously doesn't build yet... >:)

mnesbit:
Our custom capsule code adds the plugins to the classpath I thought, but I don't see it here?

tools/demobench/build.gradle
```diff
@@ -199,7 +199,7 @@ task javapackage(dependsOn: distZip) {
 
                 fileset(dir: dist_source, type: 'data') {
                     include(name: 'corda/*.jar')
-                    include(name: 'plugins/*.jar')
+                    include(name: 'cordapps/*.jar')
                     include(name: 'explorer/*.jar')
```
chrisr3: You've missed the `distributions` section:
```gradle
distributions {
    main() {
        contents {
            ...
            from(project(':finance').tasks.jar) {
                rename 'corda-finance-(.*)', 'corda-finance.jar'
                into 'plugins'  // <<< HERE
                fileMode = 0444
            }
            from(project(':samples:bank-of-corda-demo').jar) {
                rename 'bank-of-corda-demo-(.*)', 'bank-of-corda.jar'
                into 'plugins'  // <<< AND HERE!
                fileMode = 0444
            }
        }
    }
}
```
You can test the distribution by:
```
$ gradlew tools:demobench:installDist
```

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -49,6 +49,11 @@ class CordappLoader private constructor(private val cordappJarPaths: List<URL>)
...
 
         /**
+         * Cordapp directory name
+         */
+        val CORDAPPS_DIR_NAME = "cordapps"
+
+        /**
          * Creates a default CordappLoader intended to be used in non-dev or non-test environments.
          *
          * @param baseDir The directory that this node is running in. Will use this to resolve the plugins directory
```
chrisr3: Comment out-of-date

tools/demobench/src/main/kotlin/net/corda/demobench/plugin/PluginController.kt
```diff
@@ -16,7 +16,7 @@ import java.util.stream.Stream
 class PluginController : Controller() {
```
chrisr3: You've missed these two extension functions in this file:
```kotlin
fun Path.isPlugin(): Boolean = Files.isReadable(this) && this.fileName.toString().endsWith(".jar")
fun Path.inPluginsDir(): Boolean = (this.parent != null) && this.parent.endsWith("plugins/")
```
_Especially_ the `inPluginsDir()` function.

tools/demobench/src/main/kotlin/net/corda/demobench/plugin/PluginController.kt
```diff
@@ -30,7 +30,7 @@ class PluginController : Controller() {
         }
         if (finance.exists()) {
             finance.copyToDirectory(config.pluginDir, StandardCopyOption.REPLACE_EXISTING)
-            log.info("Installed 'Finance' plugin")
+            log.info("Installed 'Finance' cordapp")
         }
```
chrisr3: Isn't "Bank of Corda" _also_ a cordapp?

tools/demobench/src/main/kotlin/net/corda/demobench/plugin/PluginController.kt
```diff
@@ -30,7 +30,7 @@ class PluginController : Controller() {
...
         // Nodes cannot issue cash unless they contain the "Bank of Corda" plugin.
```
chrisr3: Comment out-of-date...

chrisr3: And check `tools/demobench/README.md` too, because I'm sure it mentions a "plugins" directory.

Clintonio:
Will look at this post V1.

mikehearn:
Can we continue adding stuff in plugins/ even if it's not documented? It seems not hard to do and it avoids a backcompat break.

Clintonio:
Wawrzek and I were talking about using symlinks for that since that way we don't have to pollute the code with the legacy stuff. Thoughts @mikehearn?

chrisr3:
While symlinks _are_ supported on Windows, I believe that they are only enabled for Administrators by default.

mikehearn:
Symlinks don't work well on Windows. We should keep scanning the old path.

Clintonio:
The plugins dir is now still picked up by the starting node as a legacy feature but isn't anywhere else in the codebase. Does this work @mikehearn @chrisr3?

PR #1646, opened 2017-09-26
Removed SSLConfiguration parameter from CordaRPCClient, thus removing…

shamsasari:
… SSL support.

The current use of SSL RPC relies on access to the node's keystore file, and further to that some uses where using the NODE_USER to login on the p2p port.

As far as I'm aware we've not used RPC SSL without replying on the node keystore file so it doesn't make sense to ship 1.0 with RPC in this state.

### DIFF
--- a/client/rpc/src/integration-test/java/net/corda/client/rpc/CordaRPCJavaClientTest.java
+++ b/client/rpc/src/integration-test/java/net/corda/client/rpc/CordaRPCJavaClientTest.java
@@ -57,7 +57,7 @@ public class CordaRPCJavaClientTest extends NodeBasedTest {
         CordaFuture<StartedNode<Node>> nodeFuture = startNode(getALICE().getName(), 1, services, singletonList(rpcUser), emptyMap());
         node = nodeFuture.get();
         node.getInternals().registerCustomSchemas(Collections.singleton(CashSchemaV1.INSTANCE));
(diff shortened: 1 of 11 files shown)


mikehearn:
I think the assumption was that parts of the node would use RPC to communicate between each other, once more components are split out, but I suspect RPC needs a bit more work to support that first. Can you file a JIRA to bring RPC-SSL back? It shouldn't be hard, all we need to do is have the node produce a keystore file that has only its certificate in it and not its private key. The admin can then distribute that as part of apps that use it (and we can provide a way to put a PEM encoded cert in the source code of an app).

mnesbit:
I would to be expect each node to have a local self-signed cert/private key that is used inside the organisation to control RPC access and issue RPC user certificates, but which couldn't be used for spoofing node to node, or legal identity. However, I agree that design work is for another day.

mnesbit:
Can we now push up the SSLConfiguration and NodeSSLConfiguration interfaces into node, so that it is just used for FullNodeConfiguration?

shamsasari:
It's still referenced by the `tcpTransport` in node-api, but that's fine as node-api is currently internal.

mnesbit:
This looks perfectly sensible. Are there any other tests we need to fix up? I guess the templates will be broken where they make an RPCClient? Otherwise, happy to approve if you think this is good to go

PR #1652, opened 2017-09-26
CORDA-599 PersistentNetworkMapCache no longer circularly depends on SH

andr3ej:


core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -47,19 +62,6 @@ interface NetworkMapCache {
     fun track(): DataFeed<List<NodeInfo>, MapChange>
 
     /**
-     * Look up the node info for a specific party. Will attempt to de-anonymise the party if applicable; if the party
```
mnesbit: This is an API change and we are not allowed to make those

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -41,15 +40,32 @@ import java.security.SignatureException
 import java.util.*
 import javax.annotation.concurrent.ThreadSafe
 
+class NodeLookupImpl(private val identityService: IdentityService, private val networkMapCache: NetworkMapCache) : NodeLookup {
```
mnesbit: I like the idea, but perhaps we have to keep it hidden due to api effects.

rick-r3:
This looks like it changes the public API.

rick-r3:
I think we should add @shamsasari since the network map is his area.

andr3ej:
OK, then I think the best I can do here is retain the circularity but make it more explicit, like in the initial impl of #1630

Clintonio:
I'd like this to be updated or closed with a rebase against master first to double check if the new API stability versions allow this or not.

andr3ej:
I have a remix that works, and can be pushed tomorrow after some polishing. Keeping this PR open to highlight the difference between a change without API stability and the same change with.

andr3ej:
Cool, the remaining api changes aren't real, they're 16 methods moved from NMC to NMCLite that NMC extends.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -29,6 +28,22 @@ interface NetworkMapCache {
...
+     *
+     * @param party party to retrieve node information for.
+     * @return the node for the identity, or null if the node could not be found. This does not necessarily mean there is
+     * no node for the party, only that this cache is unaware of it.
+     */
+    fun getNodeByLegalIdentity(party: AbstractParty): NodeInfo?
+}
+
+/** Subset of [NetworkMapCache] that doesn't depend on an [IdentityService]. */
+interface NetworkMapCacheLite {
```
Clintonio: This name is a bit weird. Maybe "NetworkMapCacheBase"?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -459,8 +459,12 @@ abstract class AbstractNode(config: NodeConfiguration,
         attachments = NodeAttachmentService(services.monitoringService.metrics)
         cordappProvider.start(attachments)
         legalIdentity = obtainIdentity(notaryConfig = null)
+        // TODO  We keep only notary identity as additional legalIdentity if we run it on a node . Multiple identities need more design thinking.
```
Clintonio: This function is getting pretty long but I see why you're doing this (so ordering is explicit and DI)

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
andr3ej: i think guava uses similar naming. anyway it's not important, renamed

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
andr3ej: this function will be a victim of CORDA-599

PR #1653, opened 2017-09-26
Problematic CordaRPCClient c'tor made private

shamsasari:
The removal of RPC-SSL is too late to go into v1.0 so I've just deprecated the relevant c'tor. This also tackles the problematic `initialiseSerialization` parameter which I was going to remove but won't now.

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
@@ -35,8 +35,8 @@ data class CordaRPCClientConfiguration(val connectionMaxRetryInterval: Duration)
         /**
          * Returns the default configuration we recommend you use.
          */
-        @JvmStatic
-        val default = CordaRPCClientConfiguration(connectionMaxRetryInterval = RPCClientConfiguration.default.connectionMaxRetryInterval)
+        @JvmField
+        val DEFAULT = CordaRPCClientConfiguration(connectionMaxRetryInterval = RPCClientConfiguration.default.connectionMaxRetryInterval)
     }
 }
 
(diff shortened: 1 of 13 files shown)


mikehearn:
Can't you just make that c'tor private or internal instead of deprecating it? 1.0 isn't out yet.

shamsasari:
Private won't work right? I can make it internal and then have a public factory method in the internal package for those uses which are outside of the module. That should work.

mikehearn:
I guess the risk of such a change is fairly minimal as it's entirely checkable by the compiler.

shamsasari:
Internal is not going to work as an internal c'tor is still public in Java (I guess there's no name to mangle with a c'tor). What'll do instead is make the c'tor private and then have an internal bridge method in CordaRPCClient which can be accessed by the factory method in the internal package. 

Not pretty but hopefully it should work.

shamsasari:
OK that's done. Back to you guys.

samples/irs-demo/src/integration-test/kotlin/net/corda/irs/IRSDemoTest.kt
```diff
@@ -101,7 +97,7 @@ class IRSDemoTest : IntegrationTestCategory {
     }
 
     private fun getFixingDateObservable(config: FullNodeConfiguration): Observable<LocalDate?> {
-        val client = CordaRPCClient(config.rpcAddress!!, initialiseSerialization = false)
+        val client = internalCordaRPCClient(config.rpcAddress!!, initialiseSerialization = false)
```
mikehearn: No, this looks all wrong. Why are samples using internal APIs now? This isn't what I meant at all. 

I guess we just have to suck up the crappy API or push the release out a few more days so we have time to fix the initialiseSerialisation junk a better way.

shamsasari: The samples only use this internal method in the integration tests. Can we get away with that?

Unfortunately leaving the c'tor as is is not an option. It has a direct reference to `SSLConfiguration` which is internal (it's part of node-api).

mikehearn: But the only thing it's doing is changing the initialiseSerialisation parameter, right?

Do we need to take another few days to get rid of this initialiseSerialisation thing properly?

mikehearn: Yes please.

shamsasari:
`initialiseSerialization` removed.

client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -26,27 +26,28 @@ data class CordaRPCClientConfiguration(
...
+private constructor(hostAndPort: NetworkHostAndPort,
+                    sslConfiguration: SSLConfiguration? = null,
+                    configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.DEFAULT
 ) {
+    constructor(hostAndPort: NetworkHostAndPort) : this(hostAndPort, sslConfiguration = null)
+    constructor(hostAndPort: NetworkHostAndPort, configuration: CordaRPCClientConfiguration) :
+            this(hostAndPort, sslConfiguration = null, configuration = configuration)
+
     init {
         // Init serialization.  It's plausible there are multiple clients in a single JVM, so be tolerant of
```
mikehearn: Should this comment now be removed?

PR #1659, opened 2017-09-26
BIGINT fix for H2 coin selection.

rick-r3:
As per fix to M16 - merge it to master


samples/trader-demo/src/main/kotlin/net/corda/traderdemo/TraderDemo.kt
```diff
@@ -44,17 +44,17 @@ private class TraderDemo {
...
 
         // What happens next depends on the role. The buyer sits around waiting for a trade to start. The seller role
-        // will contact the buyer and actually make something happen.
+        // will contact the buyer and actually make something happen.  We intentionally use large amounts here.
         val role = options.valueOf(roleArg)!!
         if (role == Role.BANK) {
             val bankHost = NetworkHostAndPort("localhost", bankRpcPort)
             CordaRPCClient(bankHost).use("demo", "demo") {
-                TraderDemoClientApi(it.proxy).runIssuer(1100.DOLLARS, buyerName, sellerName)
+                TraderDemoClientApi(it.proxy).runIssuer(1_100_000_000.DOLLARS, buyerName, sellerName)
```
mnesbit: Int.DOLLARS is called, which may be confusing. (The internal multiply by 100 will overflow the Int, but even so!) Really we should use the fromDecimal factory on AMOUNT for such things.

PR #1663, opened 2017-09-26
CORDA-641: Enable verifying notary in contract upgrade flow tests

rnicoll:


adagys:
This is now redundant, #1700

PR #1666, opened 2017-09-26
A variety of small fixes

mikehearn:
* RPC docs improvements (mostly moving stuff around)
* Make RPCClient.RPCConnection a top level public interface
* Make package descriptions appear in Kotlin API docs
* Flesh out docs on ServiceHub
* Upgrade the Gradle version check plugin to make it work again
* Add an IntelliJ scope that covers the public API, useful for spotting missing kdocs.
* Suppress Any extension functions that are internal.

### DIFF
--- a/.idea/scopes/Corda_API.xml
+++ b/.idea/scopes/Corda_API.xml
@@ -1,0 +1,3 @@
+<component name="DependencyValidationManager">
+  <scope name="Corda API" pattern="src[core_main]:*..*&amp;&amp;!src[core_main]:net.corda.core.internal..*||src[rpc_main]:net.corda.client.rpc.*||src[jackson_main]:net.corda.client.jackson.*" />
+</component>
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
@@ -10,31 +10,62 @@ import net.corda.nodeapi.ConnectionDirection
 import net.corda.nodeapi.internal.serialization.KRYO_RPC_CLIENT_CONTEXT
 import java.time.Duration
 
-/** @see RPCClient.RPCConnection */
-class CordaRPCConnection internal constructor(
-        connection: RPCClient.RPCConnection<CordaRPCOps>
-) : RPCClient.RPCConnection<CordaRPCOps> by connection
+/**
+ * This class is essentially just a wrapper for an RPCConnection<CordaRPCOps> and can be treated identically.
+ *
+ * @see RPCConnection
+ */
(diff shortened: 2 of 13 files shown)


client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -10,31 +10,62 @@ import net.corda.nodeapi.ConnectionDirection
...
-/** @see RPCClient.RPCConnection */
-class CordaRPCConnection internal constructor(
-        connection: RPCClient.RPCConnection<CordaRPCOps>
-) : RPCClient.RPCConnection<CordaRPCOps> by connection
+/**
+ * This class is essentially just a wrapper for an RPCConnection<CordaRPCOps> and can be treated identically.
+ *
+ * @see RPCConnection
+ */
+class CordaRPCConnection internal constructor(connection: RPCConnection<CordaRPCOps>) : RPCConnection<CordaRPCOps> by connection
```
shamsasari: As I found out internal constructors are completely visible in Java.

core/src/main/kotlin/net/corda/core/node/NodeInfo.kt
```diff
@@ -6,21 +6,34 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.utilities.NetworkHostAndPort
 
 /**
- * Info about a network node that acts on behalf of some form of contract party.
- * @param legalIdentitiesAndCerts is a non-empty list, where the first identity is assumed to be the default identity of the node.
+ * Information about a network node that acts on behalf of some party. NodeInfos can be found via the network map
+ * cache, accessible from a [net.corda.core.node.services.NetworkMapCache].
```
shamsasari: They're also available from RPC `networkMapSnapshot()`

client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
mikehearn: Visible in the IDE? I know they're visible at the bytecode level via reflection.

.idea/scopes/Corda_API.xml
```diff
@@ -0,0 +1,3 @@
+<component name="DependencyValidationManager">
+  <scope name="Corda API" pattern="src[core_main]:*..*&amp;&amp;!src[core_main]:net.corda.core.internal..*||src[rpc_main]:net.corda.client.rpc.*||src[jackson_main]:net.corda.client.jackson.*" />
```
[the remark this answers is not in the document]
mikehearn: Nope. The Scope language is a bit odd. This is correct. Once merged, open up your preferences and look at Appearance & Behaviour > Scopes. You will see that rpc.internal is black (not included). The reason is you need two dots to trigger recursive capture.

PR #1667, opened 2017-09-26
uncheckedCast crusade

andr3ej:
Replace almost all uses of @ Suppress("UNCHECKED_CAST") with a function that is
* more specific as to exactly where you're cheating the type system
* enforces ancestral relationship between the 2 types
* usually less code

### DIFF
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
@@ -17,6 +17,7 @@ import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.AnonymousParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
+import net.corda.core.internal.uncheckedCast
 import net.corda.core.messaging.CordaRPCOps
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.services.IdentityService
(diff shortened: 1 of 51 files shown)


client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
```diff
@@ -17,6 +17,7 @@ import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.AnonymousParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
+import net.corda.core.internal.uncheckedCast
```
mnesbit: I do quite like this code as I agree the suppress is ugly. However, if we are trying to encourage use of this it shouldn't be in an internal utils, rather in the public utils.

mikehearn: The thing is, such utilities might crop up in future releases of the standard library, and it'd be annoying if we were committed to a duplication forever.

client/jfx/src/main/kotlin/net/corda/client/jfx/utils/ObservableUtilities.kt
```diff
@@ -120,13 +119,7 @@ fun <A> ObservableList<out A>.filter(predicate: ObservableValue<(A) -> Boolean>)
...
 fun <A> ObservableList<out A?>.filterNotNull(): ObservableList<A> {
     //TODO This is a tactical work round for an issue with SAM conversion (https://youtrack.jetbrains.com/issue/ALL-1552) so that the M10 explorer works.
-    @Suppress("UNCHECKED_CAST")
-    return (this as ObservableList<A?>).filtered(object : Predicate<A?> {
-        override fun test(t: A?): Boolean {
-            return t != null
-
-        }
-    }) as ObservableList<A>
+    return uncheckedCast(filtered { it != null })
```
mikehearn: If the workaround isn't needed anymore please delete the comment too.

andr3ej: I can't tell whether it's needed, so I've changed this to just add uncheckedCast with no other changes.

PR #1669, opened 2017-09-26
Gradle API Scanner plugin

chrisr3:
Implement a Gradle plugin to summarise the public APIs of published Corda modules in a deterministic way. We will then be able to "diff" these summaries in order to spot API breakages.

See [CORDA-572](https://r3-cev.atlassian.net/browse/CORDA-572).


gradle-plugins/api-scanner/src/main/java/net/corda/plugins/ScanApiTask.java
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        private void writeApis(PrintWriter writer, ScanResult result) {
+            Map<String, ClassInfo> allInfo = result.getClassNameToClassInfo();
+            result.getNamesOfAllClasses().forEach(className -> {
+                if (className.contains(".internal.")) {
+                    return;
+                }
+                ClassInfo classInfo = allInfo.get(className);
+                if (classInfo.getClassLoaders() == null) {
```
Clintonio: Please review this line.

chrisr3: These are classes that belong to the ignored parent `ClassLoader`. They're in the scan because they're superclasses, implemented interfaces, return types etc, but they don't live within the jar that we're scanning.

gradle-plugins/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -0,0 +1,215 @@
...
+                .enableFieldInfo()
+                .verbose(verbose)
+                .scan();
+            writeApis(writer, result);
+        }
+
+        private void writeApis(PrintWriter writer, ScanResult result) {
+            Map<String, ClassInfo> allInfo = result.getClassNameToClassInfo();
+            result.getNamesOfAllClasses().forEach(className -> {
+                if (className.contains(".internal.")) {
```
shamsasari: You need to remove the trailing dot as otherwise packages ending in `.internal` will be included

gradle-plugins/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -0,0 +1,215 @@
...
+            for (FieldInfo field : fields) {
+                if (field.isPublic() || field.isProtected()) {
+                    output.append("  ").println(field);
+                }
+            }
+        }
+    }
+
+    private static String stringOf(Collection<ClassInfo> items) {
+        return items.stream().map(ClassInfo::toString).collect(Collectors.joining(", "));
```
shamsasari: static import

gradle-plugins/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    private static URL toURL(File file) throws MalformedURLException {
+        return file.toURI().toURL();
+    }
+
+    private static URL[] toURLs(Iterable<File> files) throws MalformedURLException {
+        List<URL> urls = new LinkedList<>();
```
shamsasari: You could do this in one line with the Stream API

shamsasari:
This is really good, I'm running it now to see if there's anything we shouldn't be making public. There's still time right??!!

shamsasari:
```
public class net.corda.core.CordaRuntimeException extends java.lang.RuntimeException implements net.corda.core.CordaThrowable
  public void addSuppressed(Throwable[])
  public boolean equals(Object)
  @org.jetbrains.annotations.Nullable public Throwable getCause()
  @org.jetbrains.annotations.Nullable public String getMessage()
  @org.jetbrains.annotations.Nullable public String getOriginalExceptionClassName()
  @org.jetbrains.annotations.Nullable public String getOriginalMessage()
  public int hashCode()
  public void setCause(Throwable)
  public void setMessage(String)
  public void setOriginalExceptionClassName(String)
```

I think we should ignore any public/protected member that's defined in a super class that's part of the JRE. Because we've committed to the JRE they're part of our public API anyway. Otherwise we keep repeating things like `equals`, `hashCode`, `toString`, in the above example `setCause`, etc. We could probably have a message at the top of the file explaining this.

gradle-plugins/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
shamsasari: The class name follows the trailing dot, ignore me.

shamsasari:
It's including stuff like `static final class net.corda.core.Utils$toFuture$1$1 extends kotlin.jvm.internal.Lambda implements kotlin.jvm.functions.Function1`

They're all package protected so you can get away with ignoring package protected stuff, which isn't part of our public API anyway.

gradle-plugins/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
chrisr3: Gradle's `FileCollection` type is an `Iterable`, which means that I'd need to create a stream using its `spliterator` etc. So I fear your "one line" could be quite long... ;-).

gradle-plugins/api-scanner/README.md
```diff
@@ -0,0 +1,45 @@
...
+The plugin writes its output files to the `$buildDir/api` directory.
+
+### Sample Output
+```
+public interface net.corda.core.contracts.Attachment extends net.corda.core.contracts.NamedByHash
+  public abstract void extractFile(String, java.io.OutputStream)
+  @org.jetbrains.annotations.NotNull public abstract List getSigners()
+  @org.jetbrains.annotations.NotNull public abstract java.io.InputStream open()
+  @org.jetbrains.annotations.NotNull public abstract jar.JarInputStream openAsJAR()
+--
```
shamsasari: This has changed

gradle-plugins/api-scanner/src/main/java/net/corda/plugins/GenerateApi.java
```diff
@@ -0,0 +1,61 @@
...
+import java.nio.file.Files;
+
+import static java.util.Comparator.comparing;
+import static java.util.stream.Collectors.toList;
+
+@SuppressWarnings("unused")
+public class GenerateApi extends DefaultTask {
+
+    private final File outputDir;
+    private String baseName;
```
shamsasari: final

gradle-plugins/api-scanner/README.md
```diff
@@ -0,0 +1,63 @@
+# API Scanner
+
+Generates a text summary of Corda's public API that we can check for API-breaking changes.
```
shamsasari: Can you reference api-index.rst here, and also quickly mention what's not public API: e.g. synthentic methods, stuff with $ in it, etc.

gradle-plugins/api-scanner/src/main/java/net/corda/plugins/GenerateApi.java
shamsasari: final

chrisr3: It _cannot_ be `final` because it has a setter:
```java
public void setBaseName(String baseName) {
    this.baseName = baseName;
}
```
This allows us to change `baseName` from `build.gradle`.

shamsasari:
Actually I forgot about cases such as these:
`public static final class net.corda.core.Utils$toFuture$1$subscription$1 extends java.lang.Object implements rx.Observer`
Looking at the generated bytecode, there's a internal Kotlin annotation on this class, `@Metadata`. Would if we checked for class names that had `$` in them *and* had this annotation present. That should be enough to prevent accidentally ignoring things like inner classes, right?

chrisr3:
The specification of "class name contains `$` and class is annotated with `kotlin.Metadata`" also includes all the elements of this `sealed class`, _including_ its nested `enum`:
```kotlin
package net.corda.core.contracts

sealed class TransactionVerificationException(val txId: SecureHash, message: String, cause: Throwable?)
    : FlowException("$message, transaction: $txId", cause) {
    ...

    @CordaSerializable
    enum class Direction {
        INPUT,
        OUTPUT
    }
}
```
[Apparently](https://github.com/JetBrains/kotlin/blob/master/core/runtime.jvm/src/kotlin/Metadata.kt), `kotlin.Metadata` is added to _any_ class that is generated by the Kotlin compiler. But _maybe_ this annotation has fields that we interpret in a more cunning fashion...

PR #1670, opened 2017-09-26
CORDA-645: Fix validating notary flow to handle notary change transactions properly

adagys:


docs/source/example-code/src/main/kotlin/net/corda/docs/CustomNotaryTutorial.kt
```diff
@@ -35,10 +36,17 @@ class MyValidatingNotaryFlow(otherSide: FlowSession, service: MyCustomValidating
...
         try {
             val stx = subFlow(ReceiveTransactionFlow(otherSideSession, checkSufficientSignatures = false))
-            checkNotary(stx.notary)
-            checkSignatures(stx)
-            val wtx = stx.tx
-            return TransactionParts(wtx.id, wtx.inputs, wtx.timeWindow, wtx.notary)
+            val notary = stx.notary
+            checkNotary(notary)
+            var timeWindow: TimeWindow? = null
+            val transactionWithSignatures = if (stx.isNotaryChangeTransaction()) {
```
mnesbit: This if statement is a bad pattern. We should be calling a virtual method that handles the types appropriately.

node/src/main/kotlin/net/corda/node/services/transactions/ValidatingNotaryFlow.kt
```diff
@@ -15,19 +16,24 @@ import java.security.SignatureException
...
     override fun receiveAndVerifyTx(): TransactionParts {
         try {
             val stx = subFlow(ReceiveTransactionFlow(otherSideSession, checkSufficientSignatures = false))
             val notary = stx.notary
             checkNotary(notary)
-            checkSignatures(stx)
-            val wtx = stx.tx
-            return TransactionParts(wtx.id, wtx.inputs, wtx.timeWindow, notary!!)
+            var timeWindow: TimeWindow? = null
+            val transactionWithSignatures = if (stx.isNotaryChangeTransaction()) {
```
mnesbit: As above

docs/source/example-code/src/main/kotlin/net/corda/docs/CustomNotaryTutorial.kt
```diff
@@ -35,10 +36,17 @@ class MyValidatingNotaryFlow(otherSide: FlowSession, service: MyCustomValidating
...
+                stx.resolveNotaryChangeTransaction(serviceHub)
+            } else {
+                timeWindow = stx.tx.timeWindow
```
rnicoll: Confused why we're only using the `timeWindow` on normal transactions, and on notary change we use null always. If it's treated as null, shouldn't it be required to be null on the transaction?

adagys: Notary change transactions don't contain a timestamp in general. This is just to fit the api of the flow for both cases.

docs/source/example-code/src/main/kotlin/net/corda/docs/CustomNotaryTutorial.kt
adagys: As discussed, I agree this is a bad pattern, but the assumption here is that normal user flows will never have to deal with notary change transactions. So this "if" only appears in a few places on our internal flows where we do resolution/validation. Otherwise we'd have to force users to always pass in serviceHub when verifying signatures even if it's not needed.

mnesbit: I think we could have avoided this completely by having a synthetic NotaryChangeCommand on an ordinary WireTransaction. Then a faintly different bit of logic that routes to a NotaryChangeContract when that command is present and the verify of that which enforces all the rules.

Anyway, Rick and I have chatted and there clearly is not time to do anything more right now. So go for it, but there is a silent TODO here to whittle this code down to nothing.

adagys: I considered the command approach initially, but we made a decision to go with separate 
types, since it simplifies the validity checks and is less prone to potential exploits.

rnicoll:
Just realised this is merging into `master`, can you rebase onto `pre-release-M16-2` and then merge into that branch instead please? Alternatively if you merge this I'll then raise a cherry-pick PR for you.

PR #1671, opened 2017-09-27
FlowSnapshot serialization error

mkit:
There is an error when serialising the snapshot object (just before the checkpointing). It is due to the Kryo lack of support for the SubList class (thanks @mnesbit for spotting this).

### DIFF
--- a/testing/node-driver/src/integration-test/kotlin/net/corda/testing/FlowStackSnapshotTest.kt
+++ b/testing/node-driver/src/integration-test/kotlin/net/corda/testing/FlowStackSnapshotTest.kt
@@ -2,23 +2,25 @@ package net.corda.testing
 
 import co.paralleluniverse.fibers.Suspendable
 import net.corda.client.jackson.JacksonSupport
-import net.corda.core.flows.FlowLogic
-import net.corda.core.flows.FlowStackSnapshot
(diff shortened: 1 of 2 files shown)


mnesbit:
Needs a test

testing/node-driver/src/integration-test/kotlin/net/corda/testing/FlowStackSnapshotTest.kt
```diff
@@ -178,6 +180,30 @@ class MultiplePersistingSideEffectFlow(val persistCallCount: Int) : FlowLogic<St
...
+ */
+@StartableByRPC
+@InitiatingFlow
+class FlowStackSnapshotSerializationTestingFlow : FlowLogic<Unit>() {
+
+    @Suspendable
+    override fun call() {
+        val flowStackSnapshot = flowStackSnapshot()
+        val mySession = initiateFlow(ourIdentity)
+        mySession.send("TestMessage")
```
mnesbit: Better to do a ping pong sendAndReceive. The other side can fail without this node knowing.

mkit: Done.

PR #1675, opened 2017-09-27
CORDA-641: Move special case handling of notary transactions into `SignedTransaction`

rnicoll:
Move special case handling of notary transactions into `SignedTransaction` so it's implemented once and reusable, instead of scattered through flows.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
@@ -180,6 +180,30 @@ data class SignedTransaction(val txBits: SerializedBytes<CoreTransaction>,
     fun isNotaryChangeTransaction() = transaction is NotaryChangeWireTransaction
 
     /**
+     * Resolves the underlying base transaction and then returns it, handling any special case transactions such as
+     * [NotaryChangeWireTransaction].
+     */
+    fun resolveBaseTransaction(services: StateLoader): BaseTransaction {
+        return when (transaction) {
+            is NotaryChangeWireTransaction -> resolveNotaryChangeTransaction(services)
+            is TraversableTransaction -> this.tx
+            else -> throw IllegalStateException("Unknown transaction type ${transaction::class.qualifiedName}")
+        }
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


josecoll:
Please can you make this PR clear that it relates to the issue report by https://r3-cev.atlassian.net/browse/CORDA-641:
**Contract upgrade transactions are verified incorrectly**

adagys:
Maybe let's hold off with this until we get the contract upgrade transactions sorted

rnicoll:
This is a worked example of some changes being discussed as part of #1670 and also for the upgrade transactions, to simplify both changes. However yes, we need solutions and then can look at how this fits in (potentially rebasing both off this).

rnicoll:
Waiting on integration tests re-running (attachment loader failed, but known flaky), but this is ready for re-review.

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
```diff
@@ -180,6 +180,30 @@ data class SignedTransaction(val txBits: SerializedBytes<CoreTransaction>,
...
+    }
+
+    /**
+     * Resolves the underlying transaction with signatures and then returns it, handling any special case transactions
+     * such as [NotaryChangeWireTransaction].
+     */
+    fun resolveTransactionWithSignatures(services: ServiceHub): TransactionWithSignatures {
+        return when (transaction) {
+            is NotaryChangeWireTransaction -> resolveNotaryChangeTransaction(services)
+            is TraversableTransaction -> this
```
adagys: Should probably be WireTransaction (doesn't really matter much in this case, but makes it more clear)

rnicoll: Could do, but then means a separate `is FilteredTransaction ->`, is that an improvement overall?

adagys: `FilteredTransaction` should never really be wrapped in a SignedTransaction, we don't want someone accidentally storing it. Hmm, maybe we should enforce that somwhere

PR #1677, opened 2017-09-27
Unification of VaultQuery And VaultService APIs

josecoll:
Purpose is to simplify node bootstrapping (removing circular references) and aiding usability (developers no longer need to work with 2 separate vault services).

**This PR is a proposal to be included for RC3 (thus V1.0) if deemed appropriate.**

Please note that implementation remains identical to pre-merged interfaces.

Public interface is also slightly cleaner:
- removed  _val updatesPublisher: PublishSubject<Vault.Update<ContractState>>_


rick-r3:
Looks ok to me.

PR #1678, opened 2017-09-27
Updating docs w.r.t. new FilteredTransaction and TransactionSignature

kchalkias:


rick-r3 (approved):

josecoll:
There have been a large number of changes since M14 with respect to **security**.

I summarise here _changelog_ entries since M14 which introduce new Classes and Types that are not currently mentioned anywhere in our RST documentation (descriptive usage or samples):

* A new `PrivacySalt` transaction component is introduced, which is now an attribute in ``TraversableTransaction`` and inherently in ``WireTransaction``.
* A new ``nonces: List<SecureHash>`` feature has been added to ``FilteredLeaves``.
*New functions have been added to ``MerkleTransaction``:
  ``fun <T : Any> serializedHash(x: T, privacySalt: PrivacySalt?, index: Int): SecureHash``
  ``fun <T : Any> serializedHash(x: T, nonce: SecureHash): SecureHash``
  ``fun computeNonce(privacySalt: PrivacySalt, index: Int)``.
* A new ``SignatureMetadata`` data class is introduced with two attributes, ``platformVersion: Int`` and ``schemeNumberID: Int`` (the signature scheme used).
* As part of the metadata support in signatures, a new ``data class SignableData(val txId: SecureHash, val signatureMetadata: SignatureMetadata)`` is introduced, which represents the object actually signed.
* To reflect changes in the signing process, the ``Crypto`` object is now equipped with the:
  ``fun doSign(keyPair: KeyPair, signableData: SignableData): TransactionSignature`` and
  ``fun doVerify(txId: SecureHash, transactionSignature: TransactionSignature): Boolean`` functions.

I think transaction tear offs documentation also needs updating as it has inlined code snippet which seem out of date:
eg. * A new helper method ``buildFilteredTransaction(filtering: Predicate<Any>)`` is added to ``SignedTransaction`` to  directly build a ``FilteredTransaction`` using provided filtering functions, without first accessing the ``tx: WireTransaction``.

kchalkias:
I will address them in an upcoming PR, however does the common user needs to know about the underlying logic on how nonces, privacysalt and signatureMetadata are used? All of them are computed in the background, without any real change in the api (apart from the fact that `FilteredLeaves` is now removed which is reflected to this PR). Let me check about `buildFilteredTransaction`.

kchalkias:
It seems that our tutorial examples do not mention how to create a `FilteredTransaction` from a `SignableTransaction`, which explains why `SignableTransaction.buildFilteredTransaction` is not there. However these days, users mostly deal with `SignableTransaction` Vs `WireTransaction` (the latter is rarely used any more and we try to hide it from the users). Actually, it shouldn't be called `WireTransaction` any more :)

PR #1680, opened 2017-09-27
CORDA-540: Fix exception type thrown in AMQP mode

vkolomeyko:


node/src/test/kotlin/net/corda/node/services/transactions/ValidatingNotaryServiceTests.kt
```diff
@@ -69,7 +70,14 @@ class ValidatingNotaryServiceTests {
 
         val ex = assertFailsWith(NotaryException::class) { future.getOrThrow() }
         val notaryError = ex.error as NotaryError.TransactionInvalid
-        assertThat(notaryError.cause).isInstanceOf(SignedTransaction.SignaturesMissingException::class.java)
```
rick-r3: I think we should try and keep the exception type the same.  Given this exception is now on the 1.0 API, we need to be careful.  Can we debate whether re-inheriting from CordaException is a big deal?  Worse case, we have to add some support for CordaThrowable.  Is that right?

vkolomeyko: The trouble is that `SignaturesMissingException` is already extending from *class* `SignatureException` so it cannot re-inherit from `CordaException` without dropping `SignatureException`.
And dropping `SignatureException` is a big deal since there are places in the code where people check for instance of it, e.g. `net.corda.node.services.transactions.ValidatingNotaryFlow.receiveAndVerifyTx`.
So re-inheriting from `CordaException` seems like a "no go" to me.

Another option is to make `SignaturesMissingException` implement `CordaThrowable` - this is possible but we will have to add pretty much all the overrides `CordaException` has. This is when one starts to miss Scala traits :-)

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/PropertySerializer.kt
```diff
@@ -48,11 +49,19 @@ sealed class PropertySerializer(val name: String, val readMethod: Method?, val r
     }
 
     private fun Method.returnsNullable(): Boolean {
-        val returnTypeString = this.declaringClass.kotlin.memberProperties.firstOrNull { it.javaGetter == this }?.returnType?.toString() ?: "?"
-        return returnTypeString.endsWith('?') || returnTypeString.endsWith('!')
+        try {
+            val returnTypeString = this.declaringClass.kotlin.memberProperties.firstOrNull { it.javaGetter == this }?.returnType?.toString() ?: "?"
+            return returnTypeString.endsWith('?') || returnTypeString.endsWith('!')
+        } catch(e: kotlin.reflect.jvm.internal.KotlinReflectionInternalError) {
+            // This might happen for some types, e.g. kotlin.Throwable? - the root cause of the issue is: https://youtrack.jetbrains.com/issue/KT-13077
```
rick-r3: Can we make this a TODO to revisit when bug is fixed

PR #1687, opened 2017-09-27
Fix validating notary flow to handle notary change transactions properly

adagys:
This fix has been approved and merged on master


adagys:
https://github.com/corda/corda/pull/1670

PR #1693, opened 2017-09-27
FlowSession docs

exFalso:
This PR adds more discussion on why we need `FlowSession`s and inlined subflows. Not sure whether this is enough?

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FlowSession.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowSession.kt
@@ -6,8 +6,8 @@ import net.corda.core.utilities.UntrustworthyData
 
 /**
  *
- * A [FlowSession] is a handle on a communication sequence between two flows. It is used to send and receive messages
- *   between flows.
+ * A [FlowSession] is a handle on a communication sequence between two paired flows, possibly running on separate nodes.
+ *   It is used to send and receive messages between the flows as well as to query information about the counter-flow.
  *
  * There are two ways of obtaining such a session:
  *
--- a/core/src/main/kotlin/net/corda/core/flows/FlowSession.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowSession.kt
@@ -5,7 +5,18 @@ import net.corda.core.identity.Party
 import net.corda.core.utilities.UntrustworthyData
 
 /**
- * To port existing flows:
+ *
+ * A [FlowSession] is a handle on a communication sequence between two flows. It is used to send and receive messages
+ *   between flows.
+ *
+ * There are two ways of obtaining such a session:
+ *
+ * 1.  Calling [FlowLogic.initiateFlow]. This will create a [FlowSession] object on which the first send/receive
+ *   operation will attempt to kick off a corresponding [InitiatedBy] flow on the counterparty's node.
+ * 2.  As constructor parameter to [InitiatedBy] flows. This session is the one corresponding to the initiating flow and
+ *   may be used for replies.
+ *
+ * To port flows using the old Party-based API:
  *
  * Look for [Deprecated] usages of send/receive/sendAndReceive/getFlowInfo.
  *
@@ -31,6 +42,10 @@ import net.corda.core.utilities.UntrustworthyData
  *     otherSideSession.send(something)
  */
 abstract class FlowSession {
+    /**
+     * The [Party] on the other side of this session. In the case of a session created by [FlowLogic.initiateFlow]
+     *   [counterparty] is the same Party as the one passed to that function.
+     */
     abstract val counterparty: Party
 
     /**
--- a/docs/source/api-flows.rst
+++ b/docs/source/api-flows.rst
@@ -377,10 +377,11 @@ as it likes, and each party can invoke a different response flow:
         :end-before: DOCEND 6
         :dedent: 12
 
-.. warning:: If you initiate several counter flows from the same ``@InitiatingFlow`` flow then on the receiving side you
-   must be prepared to be initiated by any of the corresponding ``initiateFlow()`` calls! A good way of handling this
-   ambiguity is to send as a first message a "role" message to the initiated flow, indicating which part of the
-   initiating flow the rest of the counter-flow should conform to.
+.. warning:: If you initiate several flows from the same ``@InitiatingFlow`` flow then on the receiving side you must be
+   prepared to be initiated by any of the corresponding ``initiateFlow()`` calls! A good way of handling this ambiguity
+   is to send as a first message a "role" message to the initiated flow, indicating which part of the initiating flow
+   the rest of the counter-flow should conform to. For example send an enum, and on the other side start with a switch
+   statement.
 
 SendAndReceive
 ~~~~~~~~~~~~~~
@@ -445,9 +446,9 @@ Consider the following contrived example using the old ``Party`` based API:
         :start-after: DOCSTART LaunchTheNukesFlow
         :end-before: DOCEND LaunchTheNukesFlow
 
-The intention of the flows is very clear: LaunchTheNukesFlow asks the president whether the nukes should be launched.
-The president in return tells the secretary that they need coffee. Afterwards the president replies that they don't want
-to launch the nukes.
+The intention of the flows is very clear: LaunchTheNukesFlow asks the president whether the nukes should be launched. It
+is expecting a boolean reply. The president in first tells the secretary that they need coffee, which is also
+communicated with a boolean. Afterwards the president replies to the launcher that they don't want to launch the nukes.
 
 However the above can go horribly wrong when the ``launcher`` happens to be the same party ``getSecretary`` returns. In
 this case the boolean meant for the secretary will be received by the launcher!
@@ -471,7 +472,7 @@ Note how the president is now explicit about which session it wants to send to. 
 Porting from the old Party-based API
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-In the old API the first ``send`` or ``receive`` to a ``Party`` was the one kicking off the counterflow. This is now
+In the old API the first ``send`` or ``receive`` to a ``Party`` was the one kicking off the counter-flow. This is now
 explicit in the ``initiateFlow`` function call. To port existing code:
 
 .. container:: codeset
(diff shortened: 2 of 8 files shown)


docs/source/api-flows.rst
```diff
@@ -641,6 +673,20 @@ We can also send and receive a ``StateAndRef`` dependency chain and automaticall
...
+Inlined subflows provide a way to share commonly used flow code `while forcing users to create a parent flow`. Take for
+example ``CollectSignaturesFlow``. Say we made it an initiating flow that automatically kicks off
+``SignTransactionFlow`` that signs the transaction. This would mean malicious nodes can just send any old transaction to
+us using ``CollectSignaturesFlow`` and we would automatically sign it!
+
+By making this pair of flows inlined we provide control to the user over whether to sign the transaction or not by
+forcing them to nest it in their own parent flows.
+
+In general if you're writing a subflow the decision of whether you should make it initiating should depend on whether
+you want your flow to automatically kick off counterflows without the users' intervention.
```
rick-r3: I feel like the sentence is not that clear.  Who is the user?  Should we say it is dependent on whether the counterflow needs a broader context?  e.g. is this deal the deal I was just negotiating, or some random deal the other side has sent to me?

core/src/main/kotlin/net/corda/core/flows/FlowSession.kt
```diff
@@ -5,7 +5,18 @@ import net.corda.core.identity.Party
 import net.corda.core.utilities.UntrustworthyData
 
 /**
- * To port existing flows:
+ *
+ * A [FlowSession] is a handle on a communication sequence between two flows. It is used to send and receive messages
+ *   between flows.
```
gendal: between paired flows on two (or more?) nodes which are communicating with each other?  Perhaps too verbose - but do we want to force the reader's mental model to one where the two flows are two sides of a conversation between two nodes?

docs/source/api-flows.rst
```diff
@@ -351,6 +377,11 @@ as it likes, and each party can invoke a different response flow:
         :end-before: DOCEND 6
         :dedent: 12
 
+.. warning:: If you initiate several counter flows from the same ``@InitiatingFlow`` flow then on the receiving side you
```
gendal: counter flow?

docs/source/api-flows.rst
```diff
@@ -351,6 +377,11 @@ as it likes, and each party can invoke a different response flow:
...
+   must be prepared to be initiated by any of the corresponding ``initiateFlow()`` calls! A good way of handling this
+   ambiguity is to send as a first message a "role" message to the initiated flow, indicating which part of the
+   initiating flow the rest of the counter-flow should conform to.
```
gendal: trying to parse this....  is this equivalent to saying that we commonly want to "invoke" an operation on the other side  but because we don't provide anywhere for an operation name to be specified, we recommend encoding it in the first message and effectively writing the recipient flow as a big switch statement?

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+.. container:: codeset
+
+    .. literalinclude:: ../../docs/source/example-code/src/main/kotlin/net/corda/docs/LaunchTheNukesFlow.kt
+        :language: kotlin
+        :start-after: DOCSTART LaunchTheNukesFlow
+        :end-before: DOCEND LaunchTheNukesFlow
+
+The intention of the flows is very clear: LaunchTheNukesFlow asks the president whether the nukes should be launched.
+The president in return tells the secretary that they need coffee. Afterwards the president replies that they don't want
+to launch the nukes.
```
gendal: for the narrative to work without reference to the code, specify that the president merely replies with a "yes" or "no" to the question of whether to launch the nukes?  (ie it's important that the reply has no context I'm assuming)

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+of flows would look like this:
+
+.. container:: codeset
+
+    .. literalinclude:: ../../docs/source/example-code/src/main/kotlin/net/corda/docs/LaunchTheNukesFlow.kt
+        :language: kotlin
+        :start-after: DOCSTART LaunchTheNukesFlowCorrect
+        :end-before: DOCEND LaunchTheNukesFlowCorrect
+
+Note how the president is now explicit about which session it wants to send to. The people of Earth live another day.
```
gendal: :)

gendal: this section would be redundant after another couple of months I guess.  If the whole thing were scheduled for deletion, would anything be lost that we would want to retain long after anybody even knew the previous design existed?

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+        :language: java
+        :start-after: DOCSTART FlowSession porting
+        :end-before: DOCEND FlowSession porting
+        :dedent: 12
+
+
 Subflows
 --------
+
+Subflows are pieces of reusable flows that may be run by calling ``FlowLogic.subFlow``.
```
gendal: you immediately then go onto several sections which, presumably, describe different types of sub flows.  Perhaps include a line or two to set this up? eg "There are x types of subflows, xxx, used for yyy; zzz, used for aaa; etc"

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+
+Inlined subflows
+^^^^^^^^^^^^^^^^
+
+Inlined subflows inherit their calling flow's type when initiating a new session with a counterparty. For example say
```
gendal: comma after example

docs/source/api-flows.rst
gendal: (just noticed... you use counter-flow above and counterflow below... three different spellings :)

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+we have flow A calling an inlined subflow B, which in turn initiates a session with a party. The FlowLogic type used to
+determine which counterflow should be kicked off will be A, not B. Note that this means that the other side of this
+session must therefore be called explicitly in the kicked off flow as well.
```
gendal: what does this mean? what does it mean for the flow that was kicked off on the other node to "explicitly call the other side of this session"?

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+
+An example of such a flow is ``CollectSignaturesFlow``. It has a counter-flow ``SignTransactionFlow`` that isn't
+annotated with ``InitiatedBy``. This is because both of these flows are inlined, the kick-off relationship will be
```
gendal: semicolon, not comma after inlined

docs/source/api-flows.rst
gendal: ahhh.. from the context below... are you saying that an inlined flow on the initiating side needs to be matched by an inlined flow on the other side? (is that always the case?)... and that the overarching flow on the receiving side therefore needs to kick off this matching inlined flow so it's able to receive the data being sent to it from the initiating side?

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+defined by the parent flows calling ``CollectSignaturesFlow`` and ``SignTransactionFlow``.
+
+In the code inlined subflows appear as regular ``FlowLogic`` instances, `without` either of the ``@InitiatingFlow`` or
+``@InitiatedBy`` annotation.
+
+.. note:: Inlined flows aren't versioned, they inherit their parent flow's version.
```
gendal: semicolon not comma after versioned

gendal: perhaps add some explanation/intuition for why one would choose to mark a flow as inlined?

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
+
+Initiating subflows
+^^^^^^^^^^^^^^^^^^^
+
+Initiating subflows are ones annotated with the ``@InitiatingFlow`` annotation. When such a flow initiates a session its
+type will be used to determine which ``@InitiatedBy`` flow to kick off on the counterparty.
+
+An example is the ``@InitiatingFlow InitiatorFlow``/``@InitiatedBy ResponderFlow`` flow pair in the ``FlowCookbook``.
+
+.. note:: Initiating flows are versioned separately from their parents.
```
gendal: same comment as above.  why would I choose to use one of these?

docs/source/api-flows.rst
```diff
@@ -395,19 +426,123 @@ Our side of the flow must mirror these calls. We could do this as follows:
...
 Corda provides a number of built-in flows that should be used for handling common tasks. The most important are:
 
-* ``CollectSignaturesFlow``, which should be used to collect a transaction's required signatures
-* ``FinalityFlow``, which should be used to notarise and record a transaction
-* ``SendTransactionFlow``, which should be used to send a signed transaction if it needed to be resolved on the other side.
-* ``ReceiveTransactionFlow``, which should be used receive a signed transaction
-* ``ContractUpgradeFlow``, which should be used to change a state's contract
-* ``NotaryChangeFlow``, which should be used to change a state's notary
+* ``CollectSignaturesFlow`` (inlined), which should be used to collect a transaction's required signatures
+* ``FinalityFlow`` (initiating), which should be used to notarise and record a transaction
```
gendal: .. and inform participants that the notarisation has occurred?

docs/source/api-flows.rst
```diff
@@ -538,6 +673,20 @@ We can also send and receive a ``StateAndRef`` dependency chain and automaticall
...
+Inlined subflows provide a way to share commonly used flow code `while forcing users to create a parent flow`. Take for
+example ``CollectSignaturesFlow``. Say we made it an initiating flow that automatically kicks off
+``SignTransactionFlow`` that signs the transaction. This would mean malicious nodes can just send any old transaction to
+us using ``CollectSignaturesFlow`` and we would automatically sign it!
+
+By making this pair of flows inlined we provide control to the user over whether to sign the transaction or not by
+forcing them to nest it in their own parent flows.
+
+In general if you're writing a subflow the decision of whether you should make it initiating should depend on whether
+the counterflow needs broader context to achieve its goal.
```
gendal: a ha! ignore my comment above about reasons you might use them.

docs/source/api-flows.rst
exFalso: Elaborated on this a little. In general an inlined subflow need not have a matching counter-flow, the counter-flow matching the parent initiating flow may decide to implement the other side of the protocol without recourse to an inlined flow.

rick-r3:
Looks good

josecoll:
Sorry to be a pain, but we need to be sensitive to our board members and `LaunchTheNukesFlow` might not go down too well in certain countries.
Please can you rename this flow to something less sinister ;-)

exFalso:
How about this

gendal:
i know this has already been merged but just to say the revisions helped a lot.. thanks!

PR #1695, opened 2017-09-27
Fixed several bugs in the contract constraints work

Clintonio:

PR #1696, opened 2017-09-27
Fixes bugs with contract constraints

Clintonio:
Awaiting review on #1695


Clintonio:
Looks like this needs a rebase.

josecoll:
Make sure you only include your commit (not the entire history from master).

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -340,14 +341,15 @@ open class Node(override val configuration: FullNodeConfiguration,
...
+        val classloader = cordappLoader.appClassLoader
         SerializationDefaults.SERIALIZATION_FACTORY = SerializationFactoryImpl().apply {
             registerScheme(KryoServerSerializationScheme())
             registerScheme(AMQPServerSerializationScheme())
         }
-        SerializationDefaults.P2P_CONTEXT = KRYO_P2P_CONTEXT
-        SerializationDefaults.RPC_SERVER_CONTEXT = KRYO_RPC_SERVER_CONTEXT
-        SerializationDefaults.STORAGE_CONTEXT = KRYO_STORAGE_CONTEXT
-        SerializationDefaults.CHECKPOINT_CONTEXT = KRYO_CHECKPOINT_CONTEXT
+        SerializationDefaults.P2P_CONTEXT = KRYO_P2P_CONTEXT.withClassLoader(classloader)
```
mnesbit: I think this is the wrong way to control classloader choice. When I did my classloader work I found that I needed to set the classloader dynamically in different places. I think you are going to have to make it an extra defaulted param to .deserialize

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -118,12 +136,41 @@ class CordappLoader private constructor(private val cordappJarPaths: List<URL>)
...
+
+        /** A list of the core RPC flows present in Corda */
+        private val coreRPCFlows = listOf(
+                ContractUpgradeFlow.Initiate::class.java,
+                ContractUpgradeFlow.Authorise::class.java,
+                ContractUpgradeFlow.Deauthorise::class.java)
+
+        /** A Cordapp representing the core package which is not scanned automatically. */
+        @VisibleForTesting
+        internal val coreCordapp = CordappImpl(
```
mnesbit: Perhaps used named params as this is meaningless

testing/test-utils/src/main/kotlin/net/corda/testing/node/MockCordappProvider.kt
```diff
@@ -14,7 +14,7 @@ class MockCordappProvider(cordappLoader: CordappLoader) : CordappProviderImpl(co
     val cordappRegistry = mutableListOf<Pair<Cordapp, AttachmentId>>()
 
     fun addMockCordapp(contractClassName: ContractClassName, services: ServiceHub) {
-        val cordapp = CordappImpl(listOf(contractClassName), emptyList(), emptyList(), emptyList(), emptyList(), emptySet(), Paths.get(".").toUri().toURL())
+        val cordapp = CordappImpl(listOf(contractClassName), emptyList(), emptyList(), emptyList(), emptyList(), emptyList(), emptySet(), Paths.get(".").toUri().toURL())
```
mnesbit: Perhaps used named params as this is meaningless

node/src/test/kotlin/net/corda/node/internal/cordapp/CordappLoaderTest.kt
```diff
@@ -0,0 +1,89 @@
...
+    override fun call() { }
+}
+
+class CordappLoaderTest {
+    private companion object {
+        val testScanPackages = listOf("net.corda.node.internal.cordapp")
+        val isolatedContractId = "net.corda.finance.contracts.isolated.AnotherDummyContract"
+        val isolatedFlowName = "net.corda.finance.contracts.isolated.IsolatedDummyFlow\$Initiator"
+    }
+
```
mnesbit: Do we need an assertFails test to check for isolation/not just finding things via the classpath?

mnesbit:
The 'test that attachments retrieved over the network are not used for code' seems to hang  with: java.lang.ClassNotFoundException: net.corda.finance.contracts.isolated.IsolatedDummyFlow$Initiator

node/src/main/kotlin/net/corda/node/internal/Node.kt
Clintonio: This was already merged to master but you are right. This will be changing later since it won't affect public API.

mnesbit:
That failing test did look related I am afraid. Are you sure this won't mess up RC3?

Clintonio:
@mnesbit I'll have another look - time is ticking quickly though

PR #1492, opened 2017-09-12
CORDA-652: Remove uses of `createSomeNodes` for consistency

rnicoll:

PR #1700, opened 2017-09-27
CORDA-641: A temporary fix for contract upgrade transactions

adagys:
During LedgerTransaction verification run the right logic based on whether
it contains the UpgradeCommand.

The current still has some issues:
- Changing contracts break encumbrance links
- We need a separate vault update type for these types of transactions
- Need to make sure expected state equality checks always work properly

Further work will be fixing the above and potentially introducing a new transaction type inline with the current notary change transaction model.


rnicoll:
Can you try including https://github.com/corda/corda/pull/1492/commits/9b41bed4db212ca086b201d7a9273f742f23fe86 as a test of this, please?

rnicoll:
Many thanks for this!

PR #1701, opened 2017-09-27
Address feedback on confidential identities docs

rnicoll:


docs/source/api-identity.rst
```diff
@@ -75,15 +75,16 @@ The swap identities flow goes through the following key steps:
 1. Generate a nonce value.
 2. Send nonce value to all counterparties, and receive their nonce values.
 3. Generate a new confidential identity from our well known identity.
-4. Create a data blob containing the new confidential identity, plus the hash of the nonce values.
+4. Create a data blob containing the new confidential identity, and the hash of the nonce values.
 5. Sign the resulting data blob with the confidential identity's public key.
-6. Send the confidential identity, data blob signature to all counterparties, while receiving theirs.
+6. Send the confidential identity, and data blob signature to all counterparties, while receiving theirs.
```
gendal: remove the first comma - or add one after "signature"

docs/source/api-identity.rst
```diff
@@ -75,15 +75,16 @@ The swap identities flow goes through the following key steps:
 1. Generate a nonce value.
 2. Send nonce value to all counterparties, and receive their nonce values.
```
gendal: explain why a nonce is needed?

docs/source/api-identity.rst
```diff
@@ -101,8 +102,27 @@ entities (counterparties) to require to know which well known identities those c
...
+
+``IdentitySyncFlow`` will serve all confidential identities in the provided transaction, irrespective of well known
+identity. This is important for more complex transaction cases with 3+ parties, for example:
+
+* Alice is building the transaction, and provides some input state *x* owned by a confidential identity of Alice
+* Bob provides some input state *y* owned by a confidential identity of Bob
+* Charlie provides some input state *z* owned by a confidential identity of Charlie
+
+Alice may know all of the confidential identities ahead of time, but Bob not know about Charlie's and vice-versa.
+The assembled transactions therefore has three input states *x*, *y* and *z*, which only Alice knows the certificates for
```
gendal: rogue s on transactions

gendal: , for which only Alice possesses certificates for all confidential identities.

docs/source/api-identity.rst
```diff
@@ -101,8 +102,27 @@ entities (counterparties) to require to know which well known identities those c
...
+all confidential identities. ``IdentitySyncFlow`` must send not just Alice's confidential identity but also any other
+identities in the transaction to the Bob & Charlie.
```
gendal: to Bob and Charlie.

docs/source/api-identity.rst
```diff
@@ -41,7 +41,7 @@ Confidential Identities
 Confidential identities are key pairs where the corresponding X.509 certificate is not made public, so that parties who
```
mnesbit: X509 Certificate Chain is not made public?

docs/source/api-identity.rst
```diff
@@ -72,18 +72,20 @@ You can see an example of it being used in ``TwoPartyDealFlow.kt``:
 
 The swap identities flow goes through the following key steps:
 
-1. Generate a nonce value.
+1. Generate a nonce value to form a challenge to the other nodes, in order to ensure their responses are specific to
```
mnesbit: The whole nonce thing is up for debate to be honest. Should we suppress this level of detail?

rnicoll: I'm warming to the need for it, as while it can always be resolved retroactively, it makes it much harder for a node to cause short-term disruption (as in, there's probably weird DoS attacks that emerge without this). The exact implementation is being hammered out, but this reflects current state.

docs/source/api-identity.rst
rnicoll: I'd argue it's specifically the certificate, because the certificate includes the name, even if you can't validate it. I'll make it "certificate (and path)"

PR #1703, opened 2017-09-27
Contract constraints documentation

Clintonio:
I have a bit of dev-blindness to this feature now so any further questions or areas to expand upon are welcome.


docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+not be tamperable by providing a fake contract.
+
+CorDapps as attachments
+-----------------------
+
+CorDapps JARs (:doc:`cordapp-overview`) that are installed to the node and contain classes implementing the ``Contract``
+interface are automatically loaded into the ``AttachmentStorage`` of a node at startup. When the CorDapp is loaded by
+the node the CorDapp JAR and it's contract classes are associated so that one can be used to lookup the other, this
+is how the automatic resolution of attachments is done by the TransactionBuilder and how, when verifying the constraints
+and contracts, attachment in the attachments are associated with which Contract.
```
gendal: attachment in the attachments?  insane in the membrane!

gendal: this section needs fleshing out.  feels like there needs to be a line or two that explains how when an instance of a State is created, the platform figures out which JAR file the associated .class file lived in, assumes that is the one you intended to be used for verifications in the future and so it's the hash of that JAR which is added in the automatic case? (Assuming that's what's happening)

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+Implementations
+---------------
+
+There are currently three implemented ``AttachmentConstraints`` provided by default in Corda with one more planned.
+
+``AlwaysAcceptAttachmentConstraint``: Any attachment (except a missing one) will satisfy this constraint.
+
+``AutomaticHashConstraint``: This will be resolved in ``TransactionBuilder`` to a ``HashAttachmentConstraint``. The
+result will be the attachment hash of the CorDapp that contains the ``ContractState`` on the
+``TransactionState.contract`` field.
```
gendal: .... at the time the transaction is constructed, right?  ie this depends on the verify() method currently on the classpath (in the relevant class loader?) at the time of transaction creation being the one you intended to use, right?   (wondering idly about whether it's possible for an attacker to force a different implementation onto you through some clever technique (no idea what) so that the automation is used against you when you come to create a transaction?)

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+
+``HashAttachmentConstraint``: Will require that the hash of the attachment containing the contract matches the hash
+stored in the constraint.
```
gendal: so this means I could create a Transaction where the verify code that will be run to check it validity is different to the one currently on the class path/in my class loader?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+
+There will be a future constraint to accept the contract if it is signed by a certain party, and one to require both a
+specific signature and hash.
```
gendal: expand...  "if it is signed by a certain party".  If what is signed?  "There will be a future constraint"...  "We plan to add an additional ContractConstraing implementation that...

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Testing
+-------
+
+Since all tests involving transactions now require attachments it is also required to load the correct attachments
+for tests. Of course this isn't possible or convenient so the test suite has a set of functions to generate CorDapps
```
gendal: why isn't it possible? i get that it isn't convenient but why impossible?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+from package names or to specify JAR URLs.
+
+MockNetwork/MockNode
+********************
+
+The most simple way to ensure that a vanilla instance of a MockNode generates the correct CorDapps is to make a call
```
gendal: what do you mean "generates the correct CorDapps"?  surely we have generated a CorDapp if we're testing it.  Do you mean correct attachments? correct constraints?  it's not clear.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+to ``setCordappPackages`` before the MockNetwork/Node are created and then ``unsetCordappPackages`` after the test
+has finished. These calls will cause the ``AbstractNode`` to use the named packages as sources for CorDapps. All files
+within those packages will be zipped into a JAR and added to the attachment store and loaded as CorDapps by the
+``CordappLoader``.
```
gendal: this isn't very clear to me but possibly because i'm not close enough.  but i do think an example would be really helpful.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+
+MockServices
+************
+
+If your test uses a ``MockServices`` directly you can instantiate it using a constructor that takes a list of packages
+to use as CorDapps using the ``cordappPackages`` parameter.
```
gendal: same comment as above "to use as CorDapps"?  need an example too?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
+Contract Constraints
+====================
+
+A basic understanding of contract key concepts is required reading for this page which can be found in
+:doc:`key-concepts-contracts`.
+
+Transaction states specify a constraint over the contract that will be used to verify it. A typical constraint is the
+hash of the CorDapp JAR that contains the contract and states but will in future releases include constraints that
+require specific signers of the JAR.
```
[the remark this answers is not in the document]
Clintonio: Added more to the description.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+
+By default every ``TransactionState`` has an ``AutomaticHashConstraint``. This will be automatically resolved when
+a ``TransactionBuilder`` is converted to a ``WireTransactions``. This exists to reduce the amount of boilerplate when
```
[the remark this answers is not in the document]
Clintonio: Expanded section.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,80 @@
...
+writing transaction building code. It is possible to specify the constraint manually with any other class that
+implements the ``AttachmentConstraint`` interface. To specify a hash manually the ``HashAttachmentConstraint`` can be
+used and to not provide any constraint the ``AlwaysAcceptAttachmentConstraint`` can be used - though this is intended
+for testing only.
```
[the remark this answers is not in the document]
Clintonio: I've restructured this section leading with the idea that transaction states must specify a constraint - if not it's automagic. thoughts?

docs/source/key-concepts-contract-constraints.rst
Clintonio: It is the contract's JAR that is used to decide which attachments need to be on the transaction. It's currently assumed that the states will be available. More error handling in that area is needed.

docs/source/key-concepts-contract-constraints.rst
Clintonio: When it's converted to a WireTransaction. Made that more specific. 

Right now it'd require you to completely ignore the field to be caught out by it. This field, like the whole transaction, is signed over so if you check it then you're not going to be caught out.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+====================
+
+A basic understanding of contract key concepts, which can be found :doc:`here </key-concepts-contracts>`
+is required reading for this page.
+
+Transaction states specify a constraint over the contract that will be used to verify it.  For a transaction to be
+valid, the verify() function associated with each states must run successfully. However, for this to be secure, it is
+not sufficient to specify the verify() function by name as there may exist multiple different implementations with the
+same method signature and enclosing class. Contract constraints solve this problem by allowing a contract developer to
+constrain which verify() functions out of the universe of implementations which match the signature can be executed.
```
josecoll: What _signature_ are you referring to here? (as signature is an overloaded term we use in many other places).
Perhaps it might be better to use the term _constraint algorithm_ and use that term in the following paragraph aswell.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+
+A typical constraint is the hash of the CorDapp JAR that contains the contract and states but will in future releases
+include constraints that require specific signers of the JAR, or both the signer and the hash. Custom constraints are
+specified when constructing a transaction, if unspecified an automatic constraint is used.
+
+Constraints are specified for each ``Transactionstate``. If unspecified the ``TransactionState`` has a default
+``AutomaticHashConstraint``. This default will be automatically resolved to a specific ``HashAttachmentConstraint`` that
+contains the hash of the attachment which contrains the contract of that ``TransactionState``. This automatic resolution
+occurs when a ``TransactionBuilder`` is converted to a ``WireTransaction``. This reduces boilerplate of finding a
```
josecoll: shouldn't this be ``LedgerTransaction``?

```
    fun verify(services: ServiceHub) {
        toLedgerTransaction(services).verify()
    }
```

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+specific hash constraints when building a transaction.
+
+It is possible to specify the constraint manually with any other class that implements the ``AttachmentConstraint``
```
josecoll: Perhaps give a sample (code snippet) of what you mean by _manually_

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+interface. To specify a hash manually the ``HashAttachmentConstraint`` can be used and to not provide any constraint
+the ``AlwaysAcceptAttachmentConstraint`` can be used - though this is intended for testing only.
+
+This mechanism exists both for integrity and security reasons. It is important not to verify against the wrong contract,
+which could happen if the wrong version of the contract is attached. More importantly when resolving transaction chains
+there will, in a future release, be attachments loaded from the network into the an attachment sandbox that are used
```
josecoll: grammar error: _into the an_

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+to verify the transaction chain. Ensuring the attachment used is the correct one ensures that the verification will
+not be tamperable by providing a fake contract.
+
+CorDapps as attachments
+-----------------------
+
+CorDapps JARs (:doc:`cordapp-overview`) that are installed to the node and contain classes implementing the ``Contract``
+interface are automatically loaded into the ``AttachmentStorage`` of a node at startup.
+
+When CorDapp JARs are loaded any contract classes contained within are associated with it so that it is possible to
```
josecoll: Rephrase as doesn't quite make sense.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+find the attachment for any given contract. This is how the automatic resolution of attachments is done by the
+TransactionBuilder and how, when verifying the constraints and contracts, attachments are associated with their
```
josecoll: ``TransactionBuilder``

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+respective contracts.
+
+Implementations
+---------------
+
+There are currently three implemented ``AttachmentConstraints`` provided by default in Corda with one more planned.
```
josecoll: "three implementations of" ... 
Probably better to just state " with more planned in the future".

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,91 @@
...
+
+``AlwaysAcceptAttachmentConstraint``: Any attachment (except a missing one) will satisfy this constraint.
+
+``AutomaticHashConstraint``: This will be resolved to a ``HashAttachmentConstraint`` when a ``TransactionBuilder`` is
+converted to a ``WireTransaction``. The ``HashAttachmentConstraint`` will include the attachment hash of the CorDapp
+that contains the ``ContractState`` on the ``TransactionState.contract`` field.
```
josecoll: to a ``LedgerTransaction``

josecoll:
Please can you also add a package description for `net.corda.core.cordapp` in `packages.md` (for our Javadoc/Kotlin API docs).

docs/source/key-concepts-contract-constraints.rst
Clintonio: Calling it a constraint algorithm would be a bit of a misnomer too since it's not really an algorithm in the sense that most people think of one. I'm going to go with "satisfy the constraint" which describes it pretty well even if it seems a bit circular.

docs/source/key-concepts-contract-constraints.rst
Clintonio: The resolution refers to the automatic resolution mentioned earlier in the paragraph which occurs when it is converted to a wire transaction specifically. From there it can become either a signed transaction or a ledger transaction. This is important because the first implementation _didn't_ include signed transactions which meant it essentially didn't work.

docs/source/key-concepts-contract-constraints.rst
Clintonio: WireTransaction is correct. It was changed.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Good point, it's just very inconvenient and would make a complete mess of your build systems. Changed to reflect that.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Not exactly. I've updated the above paragraph to explain why we do generation.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Added examples.

josecoll:
Please update packages.md

Clintonio:
Done @josecoll

docs/source/key-concepts-contract-constraints.rst
josecoll: OK - thanks for clarifying, although I don't see a code path thats leads to a `TransactionBuilder` with attachment constraints becoming a `signedTransaction`.
Maybe this is future work.

josecoll:
I think you forgot to push the commit

docs/packages.md
```diff
@@ -24,6 +24,12 @@ RPC client interface to Corda, for use both by user-facing client and integratio
 
 Internal, do not use. These APIs and implementations which are currently being revised and are subject to future change.
 
+# Package net.corda.core.cordapp
+
+This package contains the interface to CorDapps from within a node. A CorDapp can access it's own context by using
```
gendal: its, not it's

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
+Contract Constraints
+====================
+
+A basic understanding of contract key concepts, which can be found :doc:`here </key-concepts-contracts>`
+is required reading for this page.
```
gendal: comma needed after the link to balance the parenthetical comma after "concepts"

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+
+Transaction states specify a constraint over the contract that will be used to verify it.  For a transaction to be
+valid, the verify() function associated with each states must run successfully. However, for this to be secure, it is
```
gendal: each state

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+not sufficient to specify the verify() function by name as there may exist multiple different implementations with the
+same method signature and enclosing class. Contract constraints solve this problem by allowing a contract developer to
+constrain which verify() functions out of the universe of implementations which satisfy the constraint can be executed.
```
gendal: out of the universe of implementations which match the signature  (ie the universe is everything that matches the signature and contract constraints restricts this universe to a subset.)

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+
+A typical constraint is the hash of the CorDapp JAR that contains the contract and states but will in future releases
+include constraints that require specific signers of the JAR, or both the signer and the hash. Custom constraints are
+specified when constructing a transaction, if unspecified an automatic constraint is used.
```
gendal: Constraints can be specified when constructing a transaction; if unspecified, an automatic constraint is used.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+
+Constraints are specified for each ``Transactionstate``. If unspecified the ``TransactionState`` has a default
```
gendal: weaselly, passive voice :)  specified by whom? when?   How about "A party constructing a transaction can specify a constraint for each Transactionstate in the transactions.  A default AutomaticHashConstraint will be applied to any Transactionstates that do not have a constraint specified.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+``AutomaticHashConstraint``. This default will be automatically resolved to a specific ``HashAttachmentConstraint`` that
+contains the hash of the attachment which contrains the contract of that ``TransactionState``. This automatic resolution
+occurs when a ``TransactionBuilder`` is converted to a ``WireTransaction``. This reduces boilerplate of finding a
+specific hash constraints when building a transaction.
```
gendal: "This removes the need for boilerplate code in the most common scenarios."

(Question:  what happens in a complex scenario where multiple parties are using a TransactionBuilder to construct a transaction?   Imagine you're responsible for creating the wire transaction that includes states both of us have contributed. So I add a state and send you the builder but have left the constraint unspecified.... I want the verify method that i know is on my classpath to be used and the documentation here tells me the automatic magic will pick it up.  And so I send the TB to you.  But you're sneaky... you know that the hashes are only resolved when the TB is converted to a WT.  So you arrange for the hash of JAR file containing a verify method more to your liking to be written into the WireTransaction and send it back to me for signing.   Do I need to explicitly check the hash is the one I expected before signing it?  If so, doesn't this defeat the point of letting the developer go with an easy, non-boilerplate default?   If not, should we say that we perform this double-checking?   (Of course, I could just be completely forgetting how Corda works... maybe TBs are conveyed across the wire as WTs and so this does happen before it leaves my node?)

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+
+It is possible to specify the constraint explicitly with any other class that implements the ``AttachmentConstraint``
+interface. To specify a hash manually the ``HashAttachmentConstraint`` can be used and to not provide any constraint
+the ``AlwaysAcceptAttachmentConstraint`` can be used - though this is intended for testing only. An example below
+shows how to construct a ``TransactionState`` with an explicitly specified hash constraint from within a flow;
+
+.. sourcecode:: kotlin
```
gendal: i thought we were committed to Java samples first; Kotlin if we have time?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+
+     // Constructing a transaction state with a custom hash constraint.
+     val notaryParty = ... // a notary party
+     val contractState = DummyState()
+     val myAttachmentsHash = serviceHub.cordappProvider.getContractAttachmentIDDummyContract.PROGRAM_ID)
+     val transactionState = TransactionState(contractState, DummyContract.PROGRAM_ID, notaryParty, AttachmentHashConstraint(myAttachmentsHash))
```
gendal: this snippet is a bit confusing... can we expand it a little to show writing this into the TransactionBuilder.  Feels like this code is just floating around not attached to anything.   what do I do with the transactionState, for example?   

other question:  would a reader know what getContractAttachmentIDDummyContract does?  ie if I'm writing an arbitrary cordapp, what would I write to get that cordapp's jar file's hash?  (ie have the example show how to achieve the same effect as the automatic codepath?)

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+``HashAttachmentConstraint``: Will require that the hash of the attachment containing the contract matches the hash
+stored in the constraint.
+
+We plan to add a future ``AttachmentConstraint`` that will only be satisfied by the presence of signatures on the
+attachment JAR. This allows for trusting of attachments from trusted entities.
+
+Limitations
+-----------
+
+When a constraint is checked the ``isSatisfiedBy`` method is provided only the relevant attachment to the transaction.
```
gendal: you haven't mentioned this method before.  What class is it on?  Provided by whom?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+The constraint can be provided any other kind of data, but the main aim is to verify that attachments match the ones we
```
gendal: can be provided any other kind of data.. How? by whom?

gendal: this sentence would seem to contradict the one before.

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+expect. This feature is not intended to be used to scan the contents of attachments or to perform logic about states
+or other aspects of the transaction.
```
gendal: I wouldn't have expected it to.. but now you mention it, I'm sitting here thinking... oh... there's something I must have misunderstood...!

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+
+Testing
+-------
+
+Since all tests involving transactions now require attachments it is also required to load the correct attachments
+for tests. Unit test environments in JVM ecosystems tend to use class directories over JARs, and so CorDapp JARs aer
```
gendal: rather than JARs?

gendal: typo:  are

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+typically aren't built for testing. Requiring this would add significant complexity to the build systems of Corda and
```
gendal: are typically aren't?    replace aren't with not? or lose the are!

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+MockNetwork/MockNode
+********************
+
+The most simple way to ensure that a vanilla instance of a MockNode generates the correct CorDapps is to make a call
+to ``setCordappPackages`` before the MockNetwork/Node are created and then ``unsetCordappPackages`` after the test
+has finished. These calls will cause the ``AbstractNode`` to use the named packages as sources for CorDapps. All files
+within those packages will be zipped into a JAR and added to the attachment store and loaded as CorDapps by the
+``CordappLoader``. An example of this usage would be:
+
+.. sourcecode:: kotlin
```
gendal: why not java?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+             // are created the CorDapps will not be loaded into the MockNodes correctly.
+             setCordappPackages(listOf("com.domain.cordapp"))
+             network = MockNetwork()
+         }
+
+         @After
+         fun teardown() {
+             // This must be called at the end otherwise the global state set by setCordappPackages may leak into future
+             // tests in the same test runner environment.
+             unsetCordappPackages()
```
gendal: aside:  this seems unsatisfactory.. is there a way post-1.0 to automate this?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,142 @@
...
+
+MockServices
+************
+
+If your test uses a ``MockServices`` directly you can instantiate it using a constructor that takes a list of packages
+to use as CorDapps using the ``cordappPackages`` parameter.
+
+.. sourcecode:: kotlin
+
+    val mockServices = MockServices(listOf("com.domain.cordapp"))
```
gendal: i don't know about testing so can't comment... but this still seems very terse...  if the reader would know all about mockServices, etc., then great.  Otherwise, maybe throw them a bone and expand :)

docs/source/key-concepts-contract-constraints.rst
Clintonio: TransactionBuilder.toWireTransaction is where the attachments are resolved.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Modified to something on the same lines.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Added more context. Are the additions enough to make sense? I was thinking of making this a Unit flow.

That call there is how to get the hash of a contract. I think this is the only reference to it.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Removed that - I was basically re-explaining basic programming concepts and convoluting the point.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Good point - removed.

docs/source/key-concepts-contract-constraints.rst
Clintonio: Unsure about automating - but making it a more sensible part of the mock APIs - definitely.

docs/source/key-concepts-contract-constraints.rst
Clintonio: I'm wondering if we even expose it. I'll discuss tomorrow with Roger.

docs/source/key-concepts-contract-constraints.rst
Clintonio: At this time I believe we blacklist sending over the TransactionBuilder so that can't be propagated. 

When creating a new transaction the hash will need to be in both stores and will be resolved from the contract name. In the case where party B tries to inject a new contract into the JAR - it'll fail for the other side. In the case where party B tries to use a different JAR - this shouldn't be possible since the contract name will also differ. If there are two contracts with the same name one will shadow the other (We need to add a warning for this though, it's completely unsupported to have two contracts at the same package and name on one node since it makes no sense at all).

So both sides should be setting the contract names themselves. If you let the other side tell you which contract to use then you're open to a lot of problems unrelated to this work.

But yeah, this took a lot of thinking to make sure it wasn't exposing issues and I'm always keeping it in the back of my mind trying to figure out any edge cases. Some pentesting on this feature and verifying in general would be welcome.

gendal: OK - I'm still not 100% understanding but maybe my key question is this:

1) Assume A adds a state and doesn't specify the hash
2) There is a back and forth with B, where additional states are added and the Transaction is incrementally built up ready for signing.
3) Do we guarantee that when A comes to sign the resulting transaction, the hash is the same one that would have been added had A immediately converted to a WireTransaction upon adding the state at step 1)  ?
4) Put another way, if B replaces the hash with one of a *different* JAR file it has previously arranged for A to have in its store and which contains a malign implementation of the verify function, will A's node detect this switch before signing?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,149 @@
...
+====================
+
+A basic understanding of contract key concepts, which can be found :doc:`here </key-concepts-contracts>`,
+is required reading for this page.
+
+Transaction states specify a constraint over the contract that will be used to verify it.  For a transaction to be
+valid, the verify() function associated with each state must run successfully. However, for this to be secure, it is
+not sufficient to specify the verify() function by name as there may exist multiple different implementations with the
+same method signature and enclosing class. Contract constraints solve this problem by allowing a contract developer to
+constrain which verify() functions out of the universe of implementations which match the signature
```
gendal: incomplete sentence.  End with "can be used", perhaps?

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,149 @@
...
+(ie the universe is everything that matches the signature and contract constraints restricts this universe to a subset.)
+
+A typical constraint is the hash of the CorDapp JAR that contains the contract and states but will in future releases
+include constraints that require specific signers of the JAR, or both the signer and the hash. Constraints can be
+specified when constructing a transaction; if unspecified, an automatic constraint is used.
+
+``TransactionState``s have a ``constraint`` field that represents that state's attachment constraint. When a party
+constructs a ``TransactionState`` without specifying the constraint parameter a default value
+(``AutomaticHashConstraint``) is used. This default will be automatically resolved to a specific
+``HashAttachmentConstraint`` that contains the hash of the attachment which contrains the contract of that
```
gendal: typo: contains

docs/source/key-concepts-contract-constraints.rst
```diff
@@ -0,0 +1,149 @@
...
+``TransactionState``. This automatic resolution occurs when a ``TransactionBuilder`` is converted to a
+``WireTransaction``. This reduces boilerplate of finding a specific hash constraints when building a transaction.
```
gendal: ... reduces the boilerplate involved in finding a specific hash constraint...

PR #1707, opened 2017-09-28
Hide SerializationContext from public API on TransactionBuilder

rick-r3:

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
@@ -81,9 +81,11 @@ open class TransactionBuilder(
      *
      * @returns A new [WireTransaction] that will be unaffected by further changes to this [TransactionBuilder].
      */
-    @JvmOverloads
     @Throws(MissingContractAttachments::class)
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -81,9 +81,11 @@ open class TransactionBuilder(
      *
      * @returns A new [WireTransaction] that will be unaffected by further changes to this [TransactionBuilder].
      */
-    @JvmOverloads
     @Throws(MissingContractAttachments::class)
-    fun toWireTransaction(services: ServicesForResolution, serializationContext: SerializationContext? = null): WireTransaction {
+    fun toWireTransaction(services: ServicesForResolution): WireTransaction = toWireTransactionWithContext(services)
+
+    @Throws(MissingContractAttachments::class)
```
shamsasari: Not really needed here.

mnesbit:
Clinton is putting some classloader stuff on serialization in https://github.com/corda/corda/pull/1696. Worth you looking at seeing this doesn't hurt anything here.

PR #1708, opened 2017-09-28
A temporary fix for contract upgrade transactions

adagys:
Same as https://github.com/corda/corda/pull/1700 that was merged to the RC branch

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
@@ -3,9 +3,11 @@ package net.corda.core.transactions
 import net.corda.core.contracts.*
 import net.corda.core.crypto.SecureHash
 import net.corda.core.identity.Party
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -63,7 +65,13 @@ data class LedgerTransaction(
     @Throws(TransactionVerificationException::class)
     fun verify() {
         verifyConstraints()
-        verifyContracts()
+        // TODO: make contract upgrade transactions have a separate type
+        if (commands.any { it.value is UpgradeCommand }) {
```
mnesbit: Would it not be cleaner to modify verifyContracts to call a getContractList function and otherwise process the individual verifications as normal?

adagys: As discussed, the contract upgrade verification logics involves invoking a `contract.upgrade`, so it can't be in its own contract

PR #1710, opened 2017-09-28
Consistent use of CordaException and CordaRuntimeException

mnesbit:
Custom exceptions in corda, should either derive from an appropriate closely related java exception, or CordaException, or CordaRuntimeException. They should not inherit just from Exception, or RuntimeException.

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/PermissionException.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/PermissionException.kt
@@ -1,5 +1,6 @@
 package net.corda.client.rpc
 
+import net.corda.core.CordaRuntimeException
 import net.corda.core.serialization.CordaSerializable
 
 /**
@@ -7,4 +8,4 @@ import net.corda.core.serialization.CordaSerializable
  * calling a method).
  */
 @CordaSerializable
-class PermissionException(msg: String) : RuntimeException(msg)
+class PermissionException(msg: String) : CordaRuntimeException(msg)
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -6,6 +6,7 @@ import com.google.common.collect.MutableClassToInstanceMap
 import com.google.common.util.concurrent.MoreExecutors
 import net.corda.confidential.SwapIdentitiesFlow
 import net.corda.confidential.SwapIdentitiesHandler
+import net.corda.core.CordaException
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.cordapp.CordappProvider
 import net.corda.core.flows.*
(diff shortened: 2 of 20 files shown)


mnesbit:
I raised this to make us more consistent. Personally I hate custom exceptions as they add very little, but we should at least be consistent if we have a CordaException class.

node/src/main/kotlin/net/corda/node/services/api/ServiceHubInternal.kt
```diff
@@ -61,7 +62,7 @@ interface NetworkMapCacheInternal : NetworkMapCache {
 }
 
 @CordaSerializable
-sealed class NetworkCacheError : Exception() {
+sealed class NetworkCacheError : CordaException("Network Cache Error", null) {
```
rick-r3: These are not Errors.  How much turmoil is renaming to ...Exception?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
```diff
@@ -360,7 +361,7 @@ class WireNodeRegistration(raw: SerializedBytes<NodeRegistration>, sig: DigitalS
 }
 
 @CordaSerializable
-sealed class NodeMapError : Exception() {
+sealed class NodeMapError : CordaException("Network Map Protocol Error", null) {
```
rick-r3: Same thing.  Shouldn't be Error.  If we rename to Exception, what's the damage?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -465,7 +466,7 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
     }
 
     // Specific class so that MockNode can catch it.
-    class DatabaseConfigurationException(msg: String) : Exception(msg)
+    class DatabaseConfigurationException(msg: String) : CordaException(msg, null)
```
rick-r3: Hadn't spotted this before.  Let's add another constructor to CordaException for the null cause case.  ie.  `constructor(message: String)`

client/rpc/src/main/kotlin/net/corda/client/rpc/PermissionException.kt
```diff
@@ -1,10 +1,11 @@
...
+import net.corda.core.CordaRuntimeException
 import net.corda.core.serialization.CordaSerializable
 
 /**
  * Thrown to indicate that the calling user does not have permission for something they have requested (for example
  * calling a method).
  */
 @CordaSerializable
-class PermissionException(msg: String) : RuntimeException(msg)
+class PermissionException(msg: String) : CordaRuntimeException(msg)
```
vkolomeyko: Minor: `@CordaSerializable` annotation is no longer necessary as `CordaThrowable` already has it.

vkolomeyko:
Is this change really necessary for 1.0 or can it wait till 1.1 ?

mnesbit:
This is public API and these exceptions are for life. People are going to start catching them soon, especially things like notary exceptions, so that they can retry in race conditions.

client/rpc/src/main/kotlin/net/corda/client/rpc/PermissionException.kt
mnesbit: I agree, but the window is closing, so I will leave and hope we marge as soon as this build is completed.

mnesbit:
Missed the cut for 1.0

PR #1711, opened 2017-09-28
CORDA-649: Improve stability of PersistentNetworkMapCacheTest

vkolomeyko:
This is done by:
* Ensuring  that cluster of nodes after restart is stable;
* Connectivity retry is happening faster than using default settings

Test is also moved under "integration-test" pack as it is quite slow and complex to be considered a unit test.

Please see referenced Jira for more information for underlying issue.


node/src/main/kotlin/net/corda/node/services/messaging/LongPropertiesRegistry.kt
```diff
@@ -0,0 +1,28 @@
+package net.corda.node.services.messaging
+
+import net.corda.core.utilities.minutes
+import net.corda.core.utilities.seconds
+
+enum class LongPropertiesRegistry(private val defaultValue: Long) {
```
mnesbit: Can we not do this! We seem to be creating a parallel config system.
i. The equivalent Kotlin construct is an object.
ii. Actually if it needs configuration for a test it probably needs configuration for the real system.
iii. This configuration can easily come from the node config (with existing defaults provided by reference.conf) and thus get into NodeConfiguration which can be passed to service construction.
iv. Perhaps the test startup can be made smarter, so that no delays are required?

vkolomeyko: Apologies, I was not aware that we already have existing config system.
I was just trying to eliminate hardcoded properties values and make them easy to override during testing.

Do we have any good example in Corda when configuration property value is changed within a test and then restored back once test is finished?

vkolomeyko:
As discussed with @mnesbit - I am reworking this PR to align it with existing properties system. Please bear with me.

vkolomeyko:
PR re-worked and TeamCity is happy with my changes.
@mnesbit and @kasiastreich - over to you for PR review.

PR #1713, opened 2017-09-28
CORDA-599 Remove dependency of NetworkMapService impls on ServiceHub

andr3ej:


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -103,6 +103,7 @@ import net.corda.core.crypto.generateKeyPair as cryptoGenerateKeyPair
 abstract class AbstractNode(open val configuration: NodeConfiguration,
                             val advertisedServices: Set<ServiceInfo>,
                             val platformClock: Clock,
+                            private val platformVersion: Int,
```
mnesbit: This is possibly to be regarded as an api change. e.g. Fuzz is running an in-node VertX server.
So I think we should avoid this change.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -572,9 +572,7 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
         return PersistentKeyManagementService(identityService, partyKeys)
     }
 
-    open protected fun makeNetworkMapService(): NetworkMapService {
-        return PersistentNetworkMapService(services, configuration.minimumPlatformVersion)
-    }
+    abstract protected fun makeNetworkMapService(network: MessagingService, networkMapCache: NetworkMapCacheInternal): NetworkMapService
```
mnesbit: This is going away, so it is unclear if this will affect anything. Though I don't have an issue with this in principle.

rick-r3:
I've added @shamsasari as a reviewer, since he runs the network map / doorman area.  Don't know if this conflicts with his teams work etc, so defer to him.  

Nothing otherwise controversial to my eyes.

Suspect we should slim down the reviewer count.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
Clintonio: This is okay since it's in `node`.

al-r3:
can you expand SH to "service hub"?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
```diff
@@ -132,8 +133,10 @@ class InMemoryNetworkMapService(services: ServiceHubInternal, minimumPlatformVer
  * subscriber clean up and is simpler to persist than the previous implementation based on a set of missing messages acks.
  */
 @ThreadSafe
-abstract class AbstractNetworkMapService(services: ServiceHubInternal,
-                                         val minimumPlatformVersion: Int) : NetworkMapService, AbstractNodeService(services) {
+abstract class AbstractNetworkMapService(network: MessagingService,
+                                         ownPlatformVersion: Int,
```
shamsasari: ourPlatformVersion

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
```diff
@@ -161,7 +164,7 @@ abstract class AbstractNetworkMapService(services: ServiceHubInternal,
 
     init {
         require(minimumPlatformVersion >= 1) { "minimumPlatformVersion cannot be less than 1" }
```
shamsasari: Can you move this check into `FullNodeConfiguration`

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -120,7 +121,6 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
 
     protected abstract val log: Logger
     protected abstract val networkMapAddress: SingleMessageRecipient?
-    protected abstract val platformVersion: Int
```
shamsasari: The reason I didn't have a c'tor param was because for `Node` this is taken from the VersionInfo. It felt wasteful to have a field!

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
```diff
@@ -161,7 +164,7 @@ abstract class AbstractNetworkMapService(services: ServiceHubInternal,
...
-        require(minimumPlatformVersion <= services.myInfo.platformVersion) {
+        require(minimumPlatformVersion <= ownPlatformVersion) {
```
shamsasari: Can you move this to AbstractNode

andr3ej:
@al-r3 did you see Clinton's comment? Is your stuff in master?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
andr3ej: it's gone now.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
andr3ej: Done

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
andr3ej: Done

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
andr3ej: I don't remember the initial reason for this, but now VersionInfo is a field on AbstractNode (and platformVersion is gone) to avoid a compiler warning when doing the check above.

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -60,11 +63,11 @@ import kotlin.system.exitProcess
  * @param advertisedServices The services this node advertises. This must be a subset of the services it runs,
  * but nodes are not required to advertise services they run (hence subset).
  */
-open class Node(override val configuration: FullNodeConfiguration,
+open class Node(configuration: FullNodeConfiguration,
```
shamsasari: `override val` and then you won't need the cast below.

andr3ej: Done, this was to avoid an NPE but it seems OK now

PR #1715, opened 2017-09-28
Hide SerializationContext from public API on TransactionBuilder

rick-r3:
Add M16 PR https://github.com/corda/corda/pull/1707 to master

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
@@ -1,7 +1,15 @@
 package net.corda.core.internal
 
+import net.corda.core.contracts.AttachmentResolutionException
+import net.corda.core.contracts.TransactionResolutionException
 import net.corda.core.crypto.SecureHash
 import net.corda.core.crypto.sha256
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -273,4 +281,12 @@ annotation class VisibleForTesting
 @Suppress("UNCHECKED_CAST")
 fun <T, U : T> uncheckedCast(obj: T) = obj as U
 
-fun <K, V> Iterable<Pair<K, V>>.toMultiMap(): Map<K, List<V>> = this.groupBy({ it.first }) { it.second }
\ No newline at end of file
+fun <K, V> Iterable<Pair<K, V>>.toMultiMap(): Map<K, List<V>> = this.groupBy({ it.first }) { it.second }
+
+// Provide access to internal method for AttachmentClassLoaderTests
+fun TransactionBuilder.toWireTransaction(services: ServicesForResolution, serializationContext: SerializationContext): WireTransaction {
```
mnesbit: Should this not be in a separate internal file? Will help if we change things to be able to find this I think.

PR #1718, opened 2017-09-28
Removal of kotlin lambdas from public API due to them generating kotl…

shamsasari:
…in interfaces.

Using Java's functional interfaces instead.

### DIFF
--- a/client/mock/src/main/kotlin/net/corda/client/mock/Generator.kt
+++ b/client/mock/src/main/kotlin/net/corda/client/mock/Generator.kt
@@ -1,8 +1,11 @@
 package net.corda.client.mock
 
 import net.corda.client.mock.Generator.Companion.choice
+import net.corda.core.internal.uncheckedCast
 import net.corda.core.utilities.Try
 import java.util.*
+import java.util.function.BiFunction
+import java.util.function.Function
 
 /**
  * This file defines a basic [Generator] library for composing random generators of objects.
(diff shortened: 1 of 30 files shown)


vkolomeyko:
With 29 files changed so far, how do we know that we caught **all** the possible instances of Kotlin interfaces in the public API?

shamsasari:
I used @chrisr3 api scanner (still in PR) to find them. Hopefully I got them all.

vkolomeyko:
Ah - excellent there is an automated tool that reports violations.
@chrisr3 would it be possible to have a look at PR for that tool? I am really keen to learn about the logic we use to flag public API violations. Is there a Wiki page for this topic?

shamsasari:
It's #1669 but currently it only prints out the public API.

chrisr3:
The tool is a Gradle plugin that writes our public API into a text file. The idea is that we run `diff -u` against our "gold" copy, and then start jumping up and down if there are any lines in the `diff` output starting `- `.

chrisr3:
I think we've missed a few:
```kotlin
public final class net.corda.core.contracts.ContractsDSL extends java.lang.Object
  public static final Object requireThat(kotlin.jvm.functions.Function1);
}

public final class net.corda.core.node.services.vault.QueryCriteriaUtils extends java.lang.Object
  public static final Object builder(kotlin.jvm.functions.Function1);
}

public abstract class net.corda.core.serialization.SerializationFactory extends java.lang.Object
  public final Object asCurrent(kotlin.jvm.functions.Function1);
  public final Object withCurrentContext(net.corda.core.serialization.SerializationContext, kotlin.jvm.functions.Function0);
}

public final class net.corda.core.utilities.KotlinUtilsKt extends java.lang.Object
  public static final void debug(org.slf4j.Logger, kotlin.jvm.functions.Function0);
  public static final void trace(org.slf4j.Logger, kotlin.jvm.functions.Function0);
  @org.jetbrains.annotations.NotNull public static final net.corda.core.utilities.PropertyDelegate transient(kotlin.jvm.functions.Function0);
}

public final class net.corda.core.utilities.UntrustworthyDataKt extends java.lang.Object
  public static final Object unwrap(net.corda.core.utilities.UntrustworthyData, kotlin.jvm.functions.Function1);
}

public final class net.corda.client.rpc.CordaRPCClient extends java.lang.Object
  public final Object use(String, String, kotlin.jvm.functions.Function1);
}
```
But since this doesn't seem to be making v1.0, I guess we're going to have to live with them anyway :-(.

shamsasari:
These ones were intentionally ignored because they're either Kotlin specific (the methods were documented as such) or look like they should actually be internal API (SerializationFactory).

chrisr3:
OK, thanks. That's a relief to know because it means that the API Scanner plugin didn't malfunction.

mikehearn:
Missed the 1.0 cutoff, but in the end Java can call Kotlin SAM interfaces too and this part of the Kotlin API is at least stable and largely invisible to Java users if they use anonymous functions.

PR #1719, opened 2017-09-28
Last minute API changes

mikehearn:
Move/rename CordaPluginRegistry because it's name is no longer reflective of what it does. Simplify the API and serialisation code along the way. Move CompositeSignaturesWithKeys which was the only class in its package.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -13,6 +13,11 @@ UNRELEASED
   either annotated with the @CordaSerializable annotation or explicitly whitelisted then a NotSerializableException is
   thrown.
 
+* If you're using the deprecated web server, it no longer connects to the node using node-internal privileges. Your
+  node configs must specify an RPC user that has sufficient privileges to do what your app needs (usually just starting
+  flows). If using the Cordformation Gradle plugin to build local nodes please see the samples, for instance the IRS
+  sample, for how to specify required permissions.
+
 Release 1.0
 -----------
 
--- a/samples/irs-demo/build.gradle
+++ b/samples/irs-demo/build.gradle
@@ -49,6 +49,11 @@ dependencies {
 }
 
 task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
+    ext.rpcUsers = [['username': "demo", 'password': "demo", 'permissions': [
+            'StartFlow.net.corda.irs.flows.AutoOfferFlow$Requester',
+            'StartFlow.net.corda.irs.flows.UpdateBusinessDayFlow$Broadcast'
+    ]]]
+
     directory "./build/nodes"
     networkMap "O=Notary Service,L=Zurich,C=CH"
     node {
@@ -59,6 +64,7 @@ task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
         webPort 10004
         cordapps = ["net.corda:finance:$corda_release_version"]
         useTestClock true
+        rpcUsers = ext.rpcUsers
     }
     node {
         name "O=Bank A,L=London,C=GB"
@@ -68,6 +74,7 @@ task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
         webPort 10007
         cordapps = ["net.corda:finance:$corda_release_version"]
         useTestClock true
+        rpcUsers = ext.rpcUsers
     }
     node {
         name "O=Bank B,L=New York,C=US"
@@ -77,6 +84,7 @@ task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
         webPort 10010
         cordapps = ["net.corda:finance:$corda_release_version"]
         useTestClock true
+        rpcUsers = ext.rpcUsers
     }
 }
 


testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -233,8 +233,8 @@ class MockNetwork(private val networkSendManuallyPumped: Boolean = false,
 
         // Allow unit tests to modify the plugin list before the node start,
```
rick-r3: Modify the comment: plugin -> whitelist ?

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -233,8 +233,8 @@ class MockNetwork(private val networkSendManuallyPumped: Boolean = false,
...
         // so they don't have to ServiceLoad test plugins into all unit tests.
-        val testPluginRegistries by lazy { super.pluginRegistries.toMutableList() }
-        override val pluginRegistries: List<CordaPluginRegistry>
+        val testPluginRegistries by lazy { super.serializationWhitelists.toMutableList() }
```
rick-r3: rename val: testPluginRegistries -> testWhitelists

docs/source/writing-cordapps.rst
```diff
@@ -44,7 +44,6 @@ The ``src`` directory of the Template CorDapp, where we define our CorDapp's sou
     │   └── resources
     │       ├── META-INF
     │       │   └── services
-    │       │       ├── net.corda.core.node.CordaPluginRegistry
```
rick-r3: Should we have an entry here for `net.corda.core.serialization.SerializationWhitelist` or you want to streamline this section (based on the deleted section below)?  We still mention the WebServerPluginRegistry here though...

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -222,8 +222,8 @@ class CordappLoader private constructor(private val cordappJarPaths: List<URL>)
         return (scanResult.getNamesOfClassesImplementing(Contract::class.java) + scanResult.getNamesOfClassesImplementing(UpgradedContract::class.java)).distinct()
     }
 
-    private fun findPlugins(cordappJarPath: URL): List<CordaPluginRegistry> {
-        return ServiceLoader.load(CordaPluginRegistry::class.java, URLClassLoader(arrayOf(cordappJarPath), appClassLoader)).toList().filter {
+    private fun findPlugins(cordappJarPath: URL): List<SerializationWhitelist> {
```
rick-r3: Rename method?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -160,8 +160,8 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
         CordaX500Name.build(cert.subjectX500Principal).copy(commonName = null)
     }
 
-    open val pluginRegistries: List<CordaPluginRegistry> by lazy {
-        cordappProvider.cordapps.flatMap { it.plugins } + DefaultWhitelist()
+    open val serializationWhitelists: List<SerializationWhitelist> by lazy {
+        cordappProvider.cordapps.flatMap { it.serializationWhitelists } + DefaultWhitelist()
```
rick-r3: If we hard code the `DefaultWhiteList` in here, maybe we should do it everywhere and drop the META-INF entry?  Think it should be visible since in node-api.

node-api/src/main/resources/META-INF/services/net.corda.core.serialization.SerializationWhitelist
```diff
@@ -0,0 +1 @@
+net.corda.nodeapi.internal.serialization.DefaultWhitelist
```
rick-r3: Potentially remove.  See comment in AbstractNode

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
shamsasari: Yes, load it manually in client-rpc and elsewhere. Then you don't even need to use the service loader. The node can look it up with the CorDapp scanning.

rick-r3: I just meant that this has `+ DefaultWhitelist()`.  If we do that in the client too, then we don't need the META-INF for the DefaultWhitelist.

shamsasari: Yeah I forgot the rpc-client would have to scan for whitelist classes as well.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
rick-r3: I decided not to change this as the plugins wording is related to the plugins directory name.  It should change as part of that instead.

shamsasari:
👍

PR #1720, opened 2017-09-28
Fixed AbstractNode to load custom notary services properly

adagys:
Converted the custom notary tutorial into a sample. 

This feature is still experimental and the sample is primarily intended for testing and development purposes, so it's not yet mentioned in the demos documentation.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
@@ -15,12 +15,13 @@ import java.security.PublicKey
 abstract class NotaryService : SingletonSerializeAsToken() {
     companion object {
         const val ID_PREFIX = "corda.notary."
-        fun constructId(validating: Boolean, raft: Boolean = false, bft: Boolean = false): String {
+        fun constructId(validating: Boolean, raft: Boolean = false, bft: Boolean = false, custom: Boolean = false): String {
             require(!raft || !bft)
             return StringBuffer(ID_PREFIX).apply {
                 append(if (validating) "validating" else "simple")
                 if (raft) append(".raft")
                 if (bft) append(".bft")
+                if (custom) append(".custom")
             }.toString()
         }
     }
--- a/core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
@@ -1,5 +1,6 @@
 package net.corda.core.node.services
 
+import com.google.common.primitives.Booleans
 import net.corda.core.contracts.StateRef
 import net.corda.core.contracts.TimeWindow
 import net.corda.core.crypto.*
@@ -16,7 +17,7 @@ abstract class NotaryService : SingletonSerializeAsToken() {
     companion object {
         const val ID_PREFIX = "corda.notary."
         fun constructId(validating: Boolean, raft: Boolean = false, bft: Boolean = false, custom: Boolean = false): String {
-            require(!raft || !bft)
+            require(Booleans.countTrue(raft, bft, custom) <= 1) { "At most one of raft, bft or custom may be true" }
             return StringBuffer(ID_PREFIX).apply {
                 append(if (validating) "validating" else "simple")
                 if (raft) append(".raft")
--- a/docs/source/corda-configuration-file.rst
+++ b/docs/source/corda-configuration-file.rst
@@ -108,11 +108,13 @@ path to the node's base directory.
                     members must be active and be able to communicate with the cluster leader for joining. If empty, a new
                     cluster will be bootstrapped.
 
-        :bftSMaRt: If part of a distributed BFT SMaRt cluster specify this config object, with the following settings:
+        :bftSMaRt: If part of a distributed BFT-SMaRt cluster specify this config object, with the following settings:
 
-                :replicaId:
+                :replicaId: The zero-based index of the current replica. All replicas must specify a unique replica id.
 
-                :clusterAddresses:
+                :clusterAddresses: List of all BFT-SMaRt cluster member addresses.
+
+        :custom: If `true`, will load and install a notary service from a CorDapp. See :doc:`tutorial-custom-notary`.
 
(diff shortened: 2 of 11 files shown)


adagys:
True, it might be best to just specify the class name in config. But that requires removing ServiceType which is a fair chunk of work and depends on the network parameters works.

Maybe it's best to leave this broken and revisit later to fix it properly (or merge this as a temp fix)

shamsasari:
Well ServiceType doesn't need to be removed for this PR. Perhaps just add a TODO somewhere to fix this properly, and a note in the rst doc that the `type` thing will do away.

adagys:
This is ready for another look

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -15,12 +15,13 @@ import java.security.PublicKey
 abstract class NotaryService : SingletonSerializeAsToken() {
     companion object {
         const val ID_PREFIX = "corda.notary."
-        fun constructId(validating: Boolean, raft: Boolean = false, bft: Boolean = false): String {
+        fun constructId(validating: Boolean, raft: Boolean = false, bft: Boolean = false, custom: Boolean = false): String {
             require(!raft || !bft)
```
shamsasari: at max only one of `raft`, bft` and `custom` can be specified, right? So perhaps introduce an enum for them to make things simpler.

docs/source/corda-configuration-file.rst
```diff
@@ -108,11 +108,13 @@ path to the node's base directory.
                     members must be active and be able to communicate with the cluster leader for joining. If empty, a new
                     cluster will be bootstrapped.
 
-        :bftSMaRt: If part of a distributed BFT SMaRt cluster specify this config object, with the following settings:
+        :bftSMaRt: If part of a distributed BFT-SMaRt cluster specify this config object, with the following settings:
 
-                :replicaId:
+                :replicaId: The zero-based index of the current replica. All replicas must specify a unique replica id.
```
shamsasari: Thanks

docs/source/corda-configuration-file.rst
```diff
@@ -108,11 +108,13 @@ path to the node's base directory.
...
 
-                :clusterAddresses:
+                :clusterAddresses: List of all BFT-SMaRt cluster member addresses.
+
+        :custom: If `true`, will load and install a notary service from a CorDapp. See :doc:`tutorial-custom-notary`.
```
shamsasari: Can you add a note that at max only one of these three can be specified.

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
shamsasari: Or you can use `Boolean.countTrue`

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -15,12 +16,13 @@ import java.security.PublicKey
 abstract class NotaryService : SingletonSerializeAsToken() {
     companion object {
         const val ID_PREFIX = "corda.notary."
-        fun constructId(validating: Boolean, raft: Boolean = false, bft: Boolean = false): String {
-            require(!raft || !bft)
+        fun constructId(validating: Boolean, raft: Boolean = false, bft: Boolean = false, custom: Boolean = false): String {
+            require(Booleans.countTrue(raft, bft, custom) == 1) { "Only one of raft, bft or custom may be true" }
```
shamsasari: The requirement has to be <= 1. Otherwise there's no way to support the core single-node notary.

adagys:
Comments addressed

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -257,18 +254,32 @@ abstract class AbstractNode(config: NodeConfiguration,
...
-                installCordaService(it)
-            } catch (e: NoSuchMethodException) {
-                log.error("${it.name}, as a Corda service, must have a constructor with a single parameter of type " +
-                        ServiceHub::class.java.name)
-            } catch (e: ServiceInstantiationException) {
-                log.error("Corda service ${it.name} failed to instantiate", e.cause)
-            } catch (e: Exception) {
-                log.error("Unable to install Corda service ${it.name}", e)
-            }
+        cordappProvider.cordapps
```
shamsasari: Don't you need to prevent more than one custom notary being installed?

adagys: Well, this is a general problem we need to solve with multiple CordaServices registering same flows

shamsasari: CordaServices can't register flows. It's done via annotations.

adagys: But we might have multiple oracle service implementations that define flow handlers that are initiated by the same flow

adagys: (fixed it for the notaries)

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -257,20 +254,43 @@ abstract class AbstractNode(config: NodeConfiguration,
...
+    /**
+     * If the [serviceClass] is a notary service, it will only be enable if the "custom" flag is set in
+     * the notary configuration.
+     */
+    private fun isServiceEnabled(serviceClass: Class<*>): Boolean {
+        return if (isNotaryService(serviceClass)) {
+            if (configuration.notary?.custom == true) {
+                val alreadyInstalledCustomNotary = cordappServices.keys.singleOrNull { isNotaryService(it) }
+                if (alreadyInstalledCustomNotary != null) {
+                    log.warn("A custom notary service has already been installed: ${alreadyInstalledCustomNotary.canonicalName}, ${serviceClass.canonicalName} will be ignored.")
```
shamsasari: I would rather the node failed because now it's a lottery which service gets chosen.

adagys: Hmm, true I'll make it throw an exception

PR #1727, opened 2017-09-29
Improve NetworkMapCache docs.

kasiastreich:
NodeInfo lookup can return more than one node for distribute services.


core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -271,6 +271,8 @@ interface CordaRPCOps : RPCOps {
 
     /**
      * Returns a node's info from the network map cache, where known.
+     * Notice that when there are more than one node for a given name (in case of distributed services) first service node
+     * found will be returned.
```
rnicoll: Would "first registered" be a better description perhaps?

kasiastreich: It's first from database, so the order is undefined.

PR #1740, opened 2017-09-29
CORDA-657: Change IdentitySyncFlow to only offer own identities

rnicoll:
Change IdentitySyncFlow to only offer confidential identities matching the identity the flow is run as. This avoids risks of nodes being convinced to include a state with a confidential identity in it, by a remote node, then feeding the well known identity to the node during identity sync.

### DIFF
--- a/confidential-identities/src/main/kotlin/net/corda/confidential/IdentitySyncFlow.kt
+++ b/confidential-identities/src/main/kotlin/net/corda/confidential/IdentitySyncFlow.kt
@@ -12,14 +12,12 @@ import net.corda.core.utilities.unwrap
 
 object IdentitySyncFlow {
     /**
-     * Flow for ensuring that one or more counterparties to a transaction have the full certificate paths of confidential
-     * identities used in the transaction. This is intended for use as a subflow of another flow, typically between
+     * Flow for ensuring that our counterparties in a transaction have the full certificate paths for *our* confidential
+     * identities used in states present in the transaction. This is intended for use as a subflow of another flow, typically between
      * transaction assembly and signing. An example of where this is useful is where a recipient of a [Cash] state wants
      * to know that it is being paid by the correct party, and the owner of the state is a confidential identity of that
      * party. This flow would send a copy of the confidential identity path to the recipient, enabling them to verify that
      * identity.
-     *
-     * @return a mapping of well known identities to the confidential identities used in the transaction.
      */
     // TODO: Can this be triggered automatically from [SendTransactionFlow]
     class Send(val otherSideSessions: Set<FlowSession>,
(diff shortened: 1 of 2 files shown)


confidential-identities/src/main/kotlin/net/corda/confidential/IdentitySyncFlow.kt
```diff
@@ -12,14 +12,12 @@ import net.corda.core.utilities.unwrap
...
+     * identities used in states present in the transaction. This is intended for use as a subflow of another flow, typically between
      * transaction assembly and signing. An example of where this is useful is where a recipient of a [Cash] state wants
      * to know that it is being paid by the correct party, and the owner of the state is a confidential identity of that
      * party. This flow would send a copy of the confidential identity path to the recipient, enabling them to verify that
      * identity.
-     *
-     * @return a mapping of well known identities to the confidential identities used in the transaction.
      */
     // TODO: Can this be triggered automatically from [SendTransactionFlow]
     class Send(val otherSideSessions: Set<FlowSession>,
```
mnesbit: Surely this now is only valid for 2 parties? For 3 or more parties they don't end up with all identities on all nodes.

rnicoll: It's intended to be called from every party that has identities to share.

gendal:
thanks ross.   just looked at the additional unit test and, assuming i'm understanding it correctly, that would seem to cover the case where Alice+Bob are exchanging conf identities but there's a state in the transaction they're studying that's owned by somebody else (charlie in this case) and ensuring that Alice doesn't provide charlie's cert to Bob.  So fine in principle by me.  But no comment on the actual implementation...  i'll leave that to matthew!

confidential-identities/src/main/kotlin/net/corda/confidential/IdentitySyncFlow.kt
```diff
@@ -61,6 +52,20 @@ object IdentitySyncFlow {
             }
         }
 
+        private fun extractOurConfidentialIdentities(): Map<AbstractParty, PartyAndCertificate?> {
+            val states: List<ContractState> = (tx.inputs.map { serviceHub.loadState(it) }.requireNoNulls().map { it.data } + tx.outputs.map { it.data })
+            val identities: Set<AbstractParty> = states.flatMap(ContractState::participants).toSet()
```
mnesbit: Participants is not the signers. Potentially the participants is a superset of the actual signers.

rnicoll:
@mnesbit This is waiting on review

confidential-identities/src/main/kotlin/net/corda/confidential/IdentitySyncFlow.kt
rnicoll: This is intended to let signing entities know who is involved in the states, so `participants` is correct for extracting identities to distribute. We could use signers for the distribution list (which is currently passed in manually), but that's a different discussion IMHO

PR #1744, opened 2017-09-29
CORDA-654: Migrate test APIs to match identity changes

rnicoll:
Replace special case API usage in unit tests with approaches that more closely model actual production scenarios. Specifically:

* Stop pulling identities from running nodes' own info, and instead extract them via the network map, which is how lookup should behave in production.
* Stop accessing `StartedNode` directly, and instead work on the `ServiceHubInternal`; while in many cases code is limited to `ServiceHub`, it's at least closer to reality.

### DIFF
--- a/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
@@ -55,7 +55,6 @@ class AttachmentTests {
 
     @Test
     fun `download and store`() {
-        val notaryNode = mockNet.createNotaryNode()
         val aliceNode = mockNet.createPartyNode(ALICE.name)
         val bobNode = mockNet.createPartyNode(BOB.name)
 
@@ -94,7 +93,6 @@ class AttachmentTests {
 
     @Test
     fun `missing`() {
-        val notaryServices = mockNet.createNotaryNode().services
         val aliceNode = mockNet.createPartyNode(ALICE.name)
         val bobNode = mockNet.createPartyNode(BOB.name)
 
--- a/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
@@ -15,8 +15,8 @@ import net.corda.node.services.persistence.NodeAttachmentService
 import net.corda.node.utilities.DatabaseTransactionManager
 import net.corda.nodeapi.internal.ServiceInfo
 import net.corda.testing.ALICE
+import net.corda.testing.ALICE_NAME
 import net.corda.testing.BOB
-import net.corda.testing.chooseIdentity
 import net.corda.testing.node.MockNetwork
 import org.junit.After
 import org.junit.Before
@@ -55,13 +55,14 @@ class AttachmentTests {
 
     @Test
     fun `download and store`() {
-        mockNet.createNotaryNode()
+        val notaryNode = mockNet.createNotaryNode()
         val aliceNode = mockNet.createPartyNode(ALICE.name)
         val bobNode = mockNet.createPartyNode(BOB.name)
 
         // Ensure that registration was successful before progressing any further
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
+        val alice = notaryNode.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
 
         aliceNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
         bobNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
@@ -73,7 +74,7 @@ class AttachmentTests {
 
         // Get node one to run a flow to fetch it and insert it.
         mockNet.runNetwork()
-        val bobFlow = bobNode.startAttachmentFlow(setOf(id), aliceNode.info.chooseIdentity())
+        val bobFlow = bobNode.startAttachmentFlow(setOf(id), alice)
         mockNet.runNetwork()
         assertEquals(0, bobFlow.resultFuture.getOrThrow().fromDisk.size)
 
@@ -87,13 +88,13 @@ class AttachmentTests {
         // Shut down node zero and ensure node one can still resolve the attachment.
         aliceNode.dispose()
 
-        val response: FetchDataFlow.Result<Attachment> = bobNode.startAttachmentFlow(setOf(id), aliceNode.info.chooseIdentity()).resultFuture.getOrThrow()
+        val response: FetchDataFlow.Result<Attachment> = bobNode.startAttachmentFlow(setOf(id), alice).resultFuture.getOrThrow()
         assertEquals(attachment, response.fromDisk[0])
     }
 
     @Test
     fun `missing`() {
-        mockNet.createNotaryNode()
+        val notaryServices = mockNet.createNotaryNode().services
         val aliceNode = mockNet.createPartyNode(ALICE.name)
         val bobNode = mockNet.createPartyNode(BOB.name)
 
@@ -107,7 +108,8 @@ class AttachmentTests {
         // Get node one to fetch a non-existent attachment.
         val hash = SecureHash.randomSHA256()
         mockNet.runNetwork()
-        val bobFlow = bobNode.startAttachmentFlow(setOf(hash), aliceNode.info.chooseIdentity())
+        val alice = notaryServices.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
+        val bobFlow = bobNode.startAttachmentFlow(setOf(hash), alice)
         mockNet.runNetwork()
         val e = assertFailsWith<FetchDataFlow.HashNotFound> { bobFlow.resultFuture.getOrThrow() }
         assertEquals(hash, e.requested)
@@ -130,6 +132,7 @@ class AttachmentTests {
         // Ensure that registration was successful before progressing any further
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
+        val alice = aliceNode.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
 
         aliceNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
         bobNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
@@ -151,7 +154,7 @@ class AttachmentTests {
 
         // Get n1 to fetch the attachment. Should receive corrupted bytes.
         mockNet.runNetwork()
-        val bobFlow = bobNode.startAttachmentFlow(setOf(id), aliceNode.info.chooseIdentity())
+        val bobFlow = bobNode.startAttachmentFlow(setOf(id), alice)
         mockNet.runNetwork()
         assertFailsWith<FetchDataFlow.DownloadedVsRequestedDataMismatch> { bobFlow.resultFuture.getOrThrow() }
     }
--- a/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
@@ -62,7 +62,7 @@ class AttachmentTests {
         // Ensure that registration was successful before progressing any further
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
-        val alice = notaryNode.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
+        val alice = aliceNode.services.myInfo.chooseIdentity(ALICE_NAME)
 
         aliceNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
         bobNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
@@ -108,7 +108,7 @@ class AttachmentTests {
         // Get node one to fetch a non-existent attachment.
         val hash = SecureHash.randomSHA256()
         mockNet.runNetwork()
-        val alice = notaryServices.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
+        val alice = aliceNode.services.myInfo.chooseIdentity(ALICE_NAME)
         val bobFlow = bobNode.startAttachmentFlow(setOf(hash), alice)
         mockNet.runNetwork()
         val e = assertFailsWith<FetchDataFlow.HashNotFound> { bobFlow.resultFuture.getOrThrow() }
@@ -132,7 +132,7 @@ class AttachmentTests {
         // Ensure that registration was successful before progressing any further
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
-        val alice = aliceNode.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
+        val alice = aliceNode.services.myInfo.chooseIdentity(ALICE_NAME)
 
         aliceNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
         bobNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
--- a/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
@@ -62,7 +62,7 @@ class AttachmentTests {
         // Ensure that registration was successful before progressing any further
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
-        val alice = aliceNode.services.myInfo.chooseIdentity(ALICE_NAME)
+        val alice = aliceNode.services.myInfo.identityFromX500Name(ALICE_NAME)
 
         aliceNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
         bobNode.internals.registerInitiatedFlow(FetchAttachmentsResponse::class.java)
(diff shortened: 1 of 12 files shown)


rnicoll:
Opening this early for feedback - is this the right approach? There's a lot of tests I'll move over if so, but if it's not I didn't want to go too far.

docs/source/example-code/src/test/kotlin/net/corda/docs/WorkflowTransactionBuildTutorialTest.kt
```diff
@@ -36,13 +39,18 @@ class WorkflowTransactionBuildTutorialTest {
         setCordappPackages("net.corda.docs")
         mockNet = MockNetwork(threadPerNode = true)
         val notaryService = ServiceInfo(ValidatingNotaryService.type)
-        notaryNode = mockNet.createNode(
+        val notaryNode = mockNet.createNode(
```
kasiastreich: Quick question, will we provide notary identity through `overrideServices`? I am thinking about getting rid of that and adding parameter for special notary case in `MockNode`. WDYT?

rnicoll: I should probably actually have a constant that's the notary service name and pull that from the network map too, again trying to match production API as closely as possible

core/src/test/kotlin/net/corda/core/flows/FinalityFlowTests.kt
```diff
@@ -16,21 +16,27 @@ import kotlin.test.assertEquals
...
-        aliceNode = mockNet.createPartyNode(notaryNode.network.myAddress, ALICE.name)
-        bobNode = mockNet.createPartyNode(notaryNode.network.myAddress, BOB.name)
+        val aliceNode = mockNet.createPartyNode(notaryNode.network.myAddress, ALICE_NAME)
+        val bobNode = mockNet.createPartyNode(notaryNode.network.myAddress, BOB_NAME)
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
-        notary = aliceNode.services.getDefaultNotary()
+        aliceServices = aliceNode.services
+        bobServices = bobNode.services
+        alice = notaryNode.services.identityService.wellKnownPartyFromX500Name(ALICE_NAME)!!
```
kasiastreich: I like this approach more. Will we use network map or identity service?

rnicoll:
Reworked to look up notary identity the same way, and to use the network map where available

core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
```diff
@@ -16,9 +16,7 @@ import net.corda.node.services.persistence.NodeAttachmentService
 import net.corda.node.services.transactions.SimpleNotaryService
 import net.corda.node.utilities.DatabaseTransactionManager
 import net.corda.nodeapi.internal.ServiceInfo
-import net.corda.testing.ALICE
-import net.corda.testing.BOB
-import net.corda.testing.chooseIdentity
+import net.corda.testing.*
```
kasiastreich: What is our guideline when it comes to imports? Intellij changes them automatically, but it can be switched off.

core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
```diff
@@ -64,6 +62,7 @@ class AttachmentTests {
         // Ensure that registration was successful before progressing any further
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
+        val alice = notaryNode.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
```
kasiastreich: It is the right direction, but needs to be documented, it will cause some race conditions on normal nodes (or people forgetting to put `mockNet.runNetwork()`).

core/src/test/kotlin/net/corda/core/flows/CollectSignaturesFlowTests.kt
```diff
@@ -38,8 +42,12 @@ class CollectSignaturesFlowTests {
         bobNode = mockNet.createPartyNode(notaryNode.network.myAddress, BOB.name)
         charlieNode = mockNet.createPartyNode(notaryNode.network.myAddress, CHARLIE.name)
         mockNet.runNetwork()
-        notary = notaryNode.services.getDefaultNotary()
+        val notaryServices = notaryNode.services
         aliceNode.internals.ensureRegistered()
+        alice = notaryServices.identityService.certificateFromKey(notaryServices.networkMapCache.getPeerByLegalName(ALICE_NAME)!!.owningKey)!!
+        bob = notaryServices.identityService.certificateFromKey(notaryServices.networkMapCache.getPeerByLegalName(BOB_NAME)!!.owningKey)!!
+        charlie = notaryServices.identityService.certificateFromKey(notaryServices.networkMapCache.getPeerByLegalName(CHARLIE_NAME)!!.owningKey)!!
+        notary = notaryServices.networkMapCache.getPeerByLegalName(DUMMY_NOTARY_SERVICE_NAME)!!
```
kasiastreich: We should use `getNotary(name: CordaX500Name)`. Notaries will be distributed as `NetworkParameters` so you should take them by name from there, not `NetworkMapCache` in general.

core/src/test/kotlin/net/corda/core/flows/CollectSignaturesFlowTests.kt
```diff
@@ -38,8 +42,12 @@ class CollectSignaturesFlowTests {
         bobNode = mockNet.createPartyNode(notaryNode.network.myAddress, BOB.name)
         charlieNode = mockNet.createPartyNode(notaryNode.network.myAddress, CHARLIE.name)
         mockNet.runNetwork()
-        notary = notaryNode.services.getDefaultNotary()
+        val notaryServices = notaryNode.services
         aliceNode.internals.ensureRegistered()
+        alice = notaryServices.identityService.certificateFromKey(notaryServices.networkMapCache.getPeerByLegalName(ALICE_NAME)!!.owningKey)!!
```
kasiastreich: Why it's so complicated? ;)

core/src/test/kotlin/net/corda/core/flows/CollectSignaturesFlowTests.kt
```diff
@@ -38,8 +42,12 @@ class CollectSignaturesFlowTests {
...
+        bob = notaryServices.identityService.certificateFromKey(notaryServices.networkMapCache.getPeerByLegalName(BOB_NAME)!!.owningKey)!!
+        charlie = notaryServices.identityService.certificateFromKey(notaryServices.networkMapCache.getPeerByLegalName(CHARLIE_NAME)!!.owningKey)!!
```
kasiastreich: We don't need charlie's certificate, only party.

core/src/test/kotlin/net/corda/core/flows/FinalityFlowTests.kt
```diff
@@ -16,21 +16,27 @@ import kotlin.test.assertEquals
...
+        val aliceNode = mockNet.createPartyNode(notaryNode.network.myAddress, ALICE_NAME)
+        val bobNode = mockNet.createPartyNode(notaryNode.network.myAddress, BOB_NAME)
         mockNet.runNetwork()
         aliceNode.internals.ensureRegistered()
-        notary = aliceNode.services.getDefaultNotary()
+        aliceServices = aliceNode.services
+        bobServices = bobNode.services
+        alice = notaryNode.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
+        bob = notaryNode.services.networkMapCache.getPeerByLegalName(BOB_NAME)!!
+        notary = notaryNode.services.networkMapCache.getPeerByLegalName(DUMMY_NOTARY_SERVICE_NAME)!!
```
kasiastreich: getNotary

node/src/test/kotlin/net/corda/node/services/NotaryChangeTests.kt
```diff
@@ -219,10 +220,10 @@ fun issueMultiPartyState(nodeA: StartedNode<*>, nodeB: StartedNode<*>, notaryNod
     return stx.tx.outRef(0)
 }
 
-fun issueInvalidState(node: StartedNode<*>, notary: Party): StateAndRef<DummyContract.SingleOwnerState> {
-    val tx = DummyContract.generateInitial(Random().nextInt(), notary, node.info.chooseIdentity().ref(0))
+fun issueInvalidState(services: ServiceHubInternal, identity: Party, notary: Party): StateAndRef<DummyContract.SingleOwnerState> {
```
kasiastreich: `ServiceHub`

node/src/test/kotlin/net/corda/node/services/transactions/ValidatingNotaryServiceTests.kt
```diff
@@ -31,22 +29,26 @@ import kotlin.test.assertFailsWith
...
                 advertisedServices = *arrayOf(ServiceInfo(NetworkMapService.type), ServiceInfo(ValidatingNotaryService.type))
         )
-        clientNode = mockNet.createNode(notaryNode.network.myAddress)
+        val aliceNode = mockNet.createNode(notaryNode.network.myAddress, legalName = ALICE_NAME)
         mockNet.runNetwork() // Clear network map registration messages
         notaryNode.internals.ensureRegistered()
-        notary = clientNode.services.getDefaultNotary()
+        notaryServices = notaryNode.services
+        aliceServices = aliceNode.services
+        notary = notaryServices.networkMapCache.getPeerByLegalName(DUMMY_NOTARY_SERVICE_NAME)!!
```
kasiastreich: getNotary

node/src/test/kotlin/net/corda/node/services/transactions/ValidatingNotaryServiceTests.kt
```diff
@@ -96,16 +98,16 @@ class ValidatingNotaryServiceTests {
...
         mockNet.runNetwork()
         return future
     }
 
-    fun issueState(node: StartedNode<*>): StateAndRef<*> {
-        val tx = DummyContract.generateInitial(Random().nextInt(), notary, node.info.chooseIdentity().ref(0))
-        val signedByNode = node.services.signInitialTransaction(tx)
-        val stx = notaryNode.services.addSignature(signedByNode, notary.owningKey)
-        node.services.recordTransactions(stx)
+    fun issueState(serviceHub: ServiceHubInternal, identity: Party): StateAndRef<*> {
```
kasiastreich: `ServiceHub`

verifier/src/integration-test/kotlin/net/corda/verifier/VerifierTests.kt
```diff
@@ -118,14 +116,15 @@ class VerifierTests {
...
-            val alice = aliceFuture.get()
-            val notary = notaryFuture.get()
-            val notaryIdentity = notary.nodeInfo.legalIdentities[1]
-            startVerifier(notary)
-            alice.rpc.startFlow(::CashIssueFlow, 10.DOLLARS, OpaqueBytes.of(0), notaryIdentity).returnValue.get()
-            notary.waitUntilNumberOfVerifiers(1)
+            val aliceNode = aliceFuture.get()
+            val notaryNode = notaryFuture.get()
+            val alice = notaryNode.rpc.wellKnownPartyFromX500Name(ALICE_NAME)!!
+            val notary = notaryNode.rpc.wellKnownPartyFromX500Name(DUMMY_NOTARY_SERVICE_NAME)!!
```
kasiastreich: We need something similar to `getNotary` on rpc.

docs/source/example-code/src/test/kotlin/net/corda/docs/WorkflowTransactionBuildTutorialTest.kt
kasiastreich: I think I will get rid/change of `overrideServices` on my next PR, because it will be used only for notaries, now is misleading.

core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
rnicoll: I generally optimize imports as we go along, but am happy to put these back and I can do any cleanup in a separate PR

core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
rnicoll: I wonder if we need some sort of `waitForPeerByLegalName` that blocks (with a timeout) until the peer registers...

rnicoll: Also, doesn't the comment "// Ensure that registration was successful before progressing any further" above cover this?

kasiastreich: Ah true, didn't notice that comment. Either way it should be documented/maybe blocking?

node/src/test/kotlin/net/corda/node/services/NotaryChangeTests.kt
rnicoll: Could have sworn that used an internal API... fixing now

core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
rnicoll: Not entirely sure what makes sense to add, beyond that comment above?

testing/test-utils/src/main/kotlin/net/corda/testing/CoreTestUtils.kt
```diff
@@ -169,6 +170,10 @@ fun NodeInfo.chooseIdentity(): Party = chooseIdentityAndCert().party
 /** Returns the identity of the first notary found on the network */
 fun ServiceHub.getDefaultNotary(): Party = networkMapCache.notaryIdentities.first()
 
+fun NetworkMapCache.getPeerCertificateByLegalName(name: CordaX500Name): PartyAndCertificate? {
```
kasiastreich: Will we need that method in `NetworkMapCache` API?

testing/test-utils/src/main/kotlin/net/corda/testing/CoreTestUtils.kt
```diff
@@ -169,6 +170,10 @@ fun NodeInfo.chooseIdentity(): Party = chooseIdentityAndCert().party
...
+    return getNodeByLegalName(name)!!.legalIdentitiesAndCerts.single { it.name == name }
```
kasiastreich: Return type is nullable, but this expression doesn't return null, just throws.

testing/test-utils/src/main/kotlin/net/corda/testing/CoreTestUtils.kt
rnicoll: We shouldn't, as generally only a few node-internal parts ever deal with the party and certificate. It's just for tests we take shortcuts and therefore need extensions.

kasiastreich: On your second PR: https://github.com/corda/corda/pull/1766/files#diff-67b02e51d45cd0c536773abfae390816R81

rnicoll:
I've tweaked this to have a `chooseIdentity()` function on `NodeInfo` that knows it should find an identity (and therefore throws an exception if it's not found). This is somewhat more elegant than having `!!` everywhere after doing identity lookup.

rnicoll:
Rebased, and I think all feedback addressed now.

rnicoll:
Rebased and renamed `chooseIdentity` to `identityFromX500Name` (similar to naming used in the identity service).

core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
```diff
@@ -87,13 +88,13 @@ class AttachmentTests {
...
 
-        val response: FetchDataFlow.Result<Attachment> = bobNode.startAttachmentFlow(setOf(id), aliceNode.info.chooseIdentity()).resultFuture.getOrThrow()
+        val response: FetchDataFlow.Result<Attachment> = bobNode.startAttachmentFlow(setOf(id), alice).resultFuture.getOrThrow()
         assertEquals(attachment, response.fromDisk[0])
     }
 
     @Test
     fun `missing`() {
-        mockNet.createNotaryNode()
+        val notaryServices = mockNet.createNotaryNode().services
```
kasiastreich: Do we need `notaryServices` here?

docs/source/example-code/src/test/kotlin/net/corda/docs/WorkflowTransactionBuildTutorialTest.kt
```diff
@@ -31,10 +33,14 @@ class WorkflowTransactionBuildTutorialTest {
     @Before
     fun setup() {
         mockNet = MockNetwork(threadPerNode = true, cordappPackages = listOf("net.corda.docs"))
-        mockNet.createNotaryNode(legalName = DUMMY_NOTARY.name)
-        nodeA = mockNet.createPartyNode()
-        nodeB = mockNet.createPartyNode()
-        nodeA.internals.registerInitiatedFlow(RecordCompletionFlow::class.java)
+        val notaryNode = mockNet.createNotaryNode(legalName = DUMMY_NOTARY.name)
```
kasiastreich: Don't need `notaryNode`.

kasiastreich: Don't need `notaryNode`.

PR #1754, opened 2017-10-01
Updated documentation indices

josecoll:
I have reverted the proposed **docs specific** release tag as per your explanation that _docs-head_ tracks the latest release documentation and can be moved to accommodate additional/evolving changes post official code release.


josecoll:
I have reverted the proposed docs release tag as per your explanation that *docs-head* tracks the latest release documentation and can be moved to accommodate additional changes post official code release.

PR #1765, opened 2017-10-02
CORDA-540: AMQP specific fixes in "node-api" project

vkolomeyko:
Use special instruction to exclude newly introduced tests from running in AMQP mode.

### DIFF
--- a/node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/SetsSerializationTest.kt
+++ b/node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/SetsSerializationTest.kt
@@ -40,7 +40,7 @@ class SetsSerializationTest : TestDependencyInjectionBase() {
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/SetsSerializationTest.kt
```diff
@@ -39,7 +40,7 @@ class SetsSerializationTest : TestDependencyInjectionBase() {
     }
 
     @Test
-    fun `check empty set serialises as Java emptySet`() {
+    fun `check empty set serialises as Java emptySet`() = kryoSpecific<ListsSerializationTest>("Checks Kryo header properties") {
```
chrisr3: Shouldn't this be `kryoSpecific<SetsSerializationTest>("etc etc")`?
Otherwise it'll use the incorrect `logger`.

vkolomeyko: yes, you are quite right - will change

PR #1766, opened 2017-10-02
Extend NetworkMapCache API

rnicoll:
Add functions to:

* Return PartyAndCertificate rather than just Party
* Return all NodeInfo entries for a name (rather than just by key)
* Return all Party/PartyAndCertificate entries for name, based on the above function

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
@@ -10,7 +10,11 @@ import java.security.cert.*
 /**
  * An identity service maintains a directory of parties by their associated distinguished name/public keys and thus
  * supports lookup of a party given its key, or name. The service also manages the certificates linking confidential
- * identities back to the well known identity (i.e. the identity in the network map) of a party.
+ * identities back to the well known identity.
+ *
+ * Well known identities in Corda are the public identity of a party, registered with the network map directory,
+ * whereas confidential identities are distributed only on a need to know basis (typically between parties in
+ * a transaction being built). See [NetworkMapCache] for retrieving well known identities from the network map.
  */
 interface IdentityService {
     val trustRoot: X509Certificate
@@ -43,8 +47,9 @@ interface IdentityService {
     fun getAllIdentities(): Iterable<PartyAndCertificate>
 
     /**
-     * Get the certificate and path for a known identity's owning key.
+     * Resolves a public key to the well known identity [PartyAndCertificate] instance which is owned by the key.
      *
+     * @param owningKey The [PublicKey] to determine well known identity for.
      * @return the party and certificate, or null if unknown.
      */
     fun certificateFromKey(owningKey: PublicKey): PartyAndCertificate?
@@ -63,16 +68,14 @@ interface IdentityService {
      * lookup from name should be done from the network map (via [NetworkMapCache]) instead, as it is the authoritative
      * source of well known identities.
      *
-     * @param name The [CordaX500Name] to search for.
+     * @param name The [CordaX500Name] to determine well known identity for.
      * @return If known the canonical [Party] with that name, else null.
      */
     fun wellKnownPartyFromX500Name(name: CordaX500Name): Party?
 
     /**
-     * Returns the well known identity from an [AbstractParty]. This is intended to resolve the well known identity,
-     * as visible in the [NetworkMapCache] from a confidential identity.
-     * It transparently handles returning the well known identity back if
-     * a well known identity is passed in.
+     * Resolves a (optionally) confidential identity to the corresponding well known identity [Party].
+     * It transparently handles returning the well known identity back if a well known identity is passed in.
      *
      * @param party identity to determine well known identity for.
      * @return well known identity, if found.
@@ -80,11 +83,12 @@ interface IdentityService {
     fun wellKnownPartyFromAnonymous(party: AbstractParty): Party?
 
     /**
-     * Returns the well known identity from a PartyAndReference. This is intended to resolve the well known identity,
-     * as visible in the [NetworkMapCache] from a confidential identity.
-     * It transparently handles returning the well known identity back if
-     * a well known identity is passed in.
+     * Resolves a (optionally) confidential identity to the corresponding well known identity [Party].
+     * Convenience method which unwraps the [Party] from the [PartyAndReference] and then resolves the
+     * well known identity as normal.
+     * It transparently handles returning the well known identity back if a well known identity is passed in.
      *
+     * @param partyRef identity (and reference, which is unused) to determine well known identity for.
      * @return the well known identity, or null if unknown.
      */
     fun wellKnownPartyFromAnonymous(partyRef: PartyAndReference) = wellKnownPartyFromAnonymous(partyRef.party)
--- a/core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
@@ -59,7 +59,10 @@ interface IdentityService {
     fun partyFromKey(key: PublicKey): Party?
 
     /**
-     * Resolves a party name to the well known identity [Party] instance for this name.
+     * Resolves a party name to the well known identity [Party] instance for this name. Where possible well known identity
+     * lookup from name should be done from the network map (via [NetworkMapCache]) instead, as it is the authoritative
+     * source of well known identities.
+     *
      * @param name The [CordaX500Name] to search for.
      * @return If known the canonical [Party] with that name, else null.
      */
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
@@ -99,7 +99,7 @@ interface NetworkMapCache {
 
     /**
      * Look up the node information entries for a legal name.
-     * Note that normally there will be only one node for a key, but for clusters of nodes or distributed services there
+     * Note that normally there will be only one node for a legal name, but for clusters of nodes or distributed services there
      * can be multiple nodes.
      */
     fun getNodesByLegalName(name: CordaX500Name): List<NodeInfo>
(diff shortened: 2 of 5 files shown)


core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,21 +73,60 @@ interface NetworkMapCache {
     /** Look up the node info for a host and port. */
     fun getNodeByAddress(address: NetworkHostAndPort): NodeInfo?
 
-    fun getPeerByLegalName(name: CordaX500Name): Party? = getNodeByLegalName(name)?.let {
-        it.legalIdentitiesAndCerts.singleOrNull { it.name == name }?.party
+    /**
+     * Look up all well known identities (including certificate path) of a legal name.
```
mnesbit: This is confusing duplication of stuff in identityservice.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,21 +73,68 @@ interface NetworkMapCache {
...
     fun getNodeByAddress(address: NetworkHostAndPort): NodeInfo?
 
-    fun getPeerByLegalName(name: CordaX500Name): Party? = getNodeByLegalName(name)?.let {
-        it.legalIdentitiesAndCerts.singleOrNull { it.name == name }?.party
+    /**
+     * Look up all well known identities (including certificate path) of a legal name. This should be used in preference
+     * to well known identity lookup in the identity service where possible, as the network map is the authoritative
+     * source of well known identities.
+     */
+    fun getPeerCertificatesByLegalName(name: CordaX500Name): Set<PartyAndCertificate> {
```
adagys: when do we need to use peer certificates?

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,21 +73,68 @@ interface NetworkMapCache {
...
+                ?.legalIdentitiesAndCerts
+                ?.singleOrNull { it.name == name }
+    }
+
+    /**
+     * Look up all well known identities for a legal name. This should be used in preference
+     * to well known identity lookup in the identity service where possible, as the network map is the authoritative
+     * source of well known identities.
+     */
+    fun getPeersByLegalName(name: CordaX500Name): Set<Party> {
```
kasiastreich: Will we have peers with the same name but different key in NetworkMap?

mnesbit:
In general can we do API design work ahead of creating a PR?

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,21 +73,68 @@ interface NetworkMapCache {
...
      * In general, nodes can advertise multiple identities: a legal identity, and separate identities for each of
      * the services it provides. In case of a distributed service – run by multiple nodes – each participant advertises
      * the identity of the *whole group*.
      */
     fun getNodesByLegalIdentityKey(identityKey: PublicKey): List<NodeInfo>
 
+    /**
+     * Look up the node information entries for a legal name.
+     * In general, nodes can advertise multiple identities: a legal identity, and separate identities for each of
+     * the services it provides. In case of a distributed service – run by multiple nodes – each participant advertises
```
kasiastreich: Hmmm that mention about services has to be removed, I forgot about that.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -106,7 +106,8 @@ open class PersistentNetworkMapCache(private val serviceHub: ServiceHubInternal)
         return null
     }
 
-    override fun getNodeByLegalName(name: CordaX500Name): NodeInfo? = serviceHub.database.transaction { queryByLegalName(name).firstOrNull() }
+    override fun getNodeByLegalName(name: CordaX500Name): NodeInfo? = queryByLegalName(name).firstOrNull()
```
kasiastreich: Should be in database.transaction too

rnicoll: Meant to actually call `getNodesByLegalName`, fixing that now

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
rnicoll: I've rewritten it to extract what I think is the most important consequences of this, rather than talking as much about what it can do. Hopefully that makes sense now.

mikehearn:
@shamsasari we can't _release_ until that work is finished but we can keep developing - we may just have to go back and change things if we discover we broke the API without realising.

As to the wider issues, I'm thinking we might want to introduce a bit of process around adding API from now on, to make sure we discuss new API before implementing it.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
rnicoll: Good point, the map always has one key at a time, will correct that

core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
```diff
@@ -59,7 +59,10 @@ interface IdentityService {
     fun partyFromKey(key: PublicKey): Party?
 
     /**
-     * Resolves a party name to the well known identity [Party] instance for this name.
+     * Resolves a party name to the well known identity [Party] instance for this name. Where possible well known identity
```
shamsasari: Perhaps you could elaborate here or somewhere else that the definition of a well-known identity is one known by the network map, assuming we don't already.

I only say that because I only realised that a few weeks ago!

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,21 +73,40 @@ interface NetworkMapCache {
...
 
-    fun getPeerByLegalName(name: CordaX500Name): Party? = getNodeByLegalName(name)?.let {
-        it.legalIdentitiesAndCerts.singleOrNull { it.name == name }?.party
+    /**
+     * Look up a well known identity (including certificate path) of a legal name. This should be used in preference
+     * to well known identity lookup in the identity service where possible, as the network map is the authoritative
+     * source of well known identities.
+     */
+    fun getPeerCertificateByLegalName(name: CordaX500Name): PartyAndCertificate? {
+        return getNodeByLegalName(name)
```
shamsasari: This is inefficient because it looks like the SQL query for getNodeByLegalName does exactly what we want but returns the NodeInfo and not the PartyAndCert as we want here. Perhaps add a TODO to address this. May be move this implementation to PersistentNetworkMapCache.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,21 +73,40 @@ interface NetworkMapCache {
...
     /** Return all [NodeInfo]s the node currently is aware of (including ourselves). */
     val allNodes: List<NodeInfo>
 
     /**
-     * Look up the node infos for a specific peer key.
-     * In general, nodes can advertise multiple identities: a legal identity, and separate identities for each of
-     * the services it provides. In case of a distributed service – run by multiple nodes – each participant advertises
-     * the identity of the *whole group*.
+     * Look up the node information entries for a specific identity key.
+     * Note that for distributed services each node advertises the same service identity.
```
shamsasari: Can you elaborate this a bit more by saying normally there will be only one node for that key, but for distributed services there can be multiple.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,21 +73,40 @@ interface NetworkMapCache {
...
      */
     fun getNodesByLegalIdentityKey(identityKey: PublicKey): List<NodeInfo>
 
+    /**
+     * Look up the node information entries for a legal name.
+     * Note that for distributed services each node advertises the same service identity.
```
shamsasari: ditto

core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
rnicoll: I've expanded the note about well known vs confidential at the top of the class, in the next set of commits. It's also now in the node docs for 1.0 (under key concepts), so hopefully we'll have less confusion from here on :)

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
rnicoll: I've added a specialised query in the next commit, and added a unit test for this function to match.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -72,26 +73,42 @@ interface NetworkMapCache {
...
-     * the identity of the *whole group*.
+     * Look up the node information entries for a specific identity key.
+     * Note that normally there will be only one node for a key, but for clusters of nodes or distributed services there
+     * can be multiple nodes.
      */
     fun getNodesByLegalIdentityKey(identityKey: PublicKey): List<NodeInfo>
 
+    /**
+     * Look up the node information entries for a legal name.
+     * Note that normally there will be only one node for a key, but for clusters of nodes or distributed services there
```
shamsasari: key -> legal name

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
```diff
@@ -5,6 +5,7 @@ import net.corda.core.crypto.DigitalSignature
 import net.corda.core.crypto.SignedData
 import net.corda.core.crypto.isFulfilledBy
 import net.corda.core.crypto.random63BitValue
+import net.corda.core.identity.CordaX500Name
```
kasiastreich: Unused import

PR #1770, opened 2017-10-02
Instantiate Contract classes outside of LedgerTransaction.verify()

chrisr3:
The `LedgerTransaction.verify()` function does not have access to Java Reflection, so instantiate the contract classes when creating the `LedgerTransaction` instance instead.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
@@ -48,6 +48,22 @@ data class LedgerTransaction(
         checkEncumbrancesValid()
     }
 
+    private companion object {
+        @JvmStatic
+        fun createContractFor(className: ContractClassName): ContractResult {
+            return try {
+                ContractResult(this::class.java.classLoader.loadClass(className).asSubclass(Contract::class.java).getConstructor().newInstance())
+            } catch (e: Exception) {
+                ContractResult(null, e)
+            }
+        }
+    }
+
+    private data class ContractResult(val contract: Contract?, val error: Throwable? = null)
+
+    private val contracts: Map<ContractClassName, ContractResult> = (inputs.map { it.state.contract } + outputs.map { it.contract })
+            .toSet().map { it to createContractFor(it) }.toMap()
+
     val inputStates: List<ContractState> get() = inputs.map { it.state.data }
 
     /**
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -48,6 +48,23 @@ data class LedgerTransaction(
...
+            return try {
+                ContractResult(contract = this::class.java.classLoader.loadClass(className).asSubclass(Contract::class.java).getConstructor().newInstance())
+            } catch (e: Exception) {
+                ContractResult(error = e)
+            }
+        }
+    }
+
+    @CordaSerializable
+    private data class ContractResult(val contract: Contract? = null, val error: Throwable? = null)
```
exFalso: This should be a `sealed class` instead, i.e.
```kotlin
sealed class ContractResult {
  data class Success(val contract: Contract) : ContractResult()
  data class Failure(val failure: Throwable) : ContractResult()
}
```
This way there is no `!!` coercion needed when we're consuming it later.

Actually just realised this is precisely `Try`, so let's use that instead

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -48,6 +48,23 @@ data class LedgerTransaction(
         checkEncumbrancesValid()
     }
 
+    private companion object {
+        @JvmStatic
+        private fun createContractFor(className: ContractClassName): ContractResult {
+            return try {
```
exFalso: can use `Try.on` here

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -98,15 +109,12 @@ data class LedgerTransaction(
...
     private fun verifyContracts() {
-        val contracts = (inputs.map { it.state.contract } + outputs.map { it.contract }).toSet()
-        for (contractClassName in contracts) {
+        for (contractEntry in contracts.entries) {
             val contract = try {
-                assert(javaClass.classLoader == ClassLoader.getSystemClassLoader())
-                javaClass.classLoader.loadClass(contractClassName).asSubclass(Contract::class.java).getConstructor().newInstance()
-            } catch (e: ClassNotFoundException) {
-                throw TransactionVerificationException.ContractCreationError(id, contractClassName, e)
+                contractEntry.value.getOrThrow()
```
exFalso: This re-throws the exception, instead this could simply be a `when(contractEntry)`

chrisr3: How does the `when(contractEntry.entry)` statement help? I need to wrap the exception as an instance of `TransactionVerificationException.ContractCreationError`, and using `getOrThrow()` is the only way to extract it from the `Failure` object.

PR #1773, opened 2017-10-02
Various fixes cherry picked from V1

Clintonio:
Why: I made these changes on V1 and they should be in master. Also fixed the flaky intergration test.


Clintonio:
And yes this does change the API - that was the last API change that went into V1.

vkolomeyko:
Seems like still unstable, see:
https://ci-master.corda.r3cev.com/viewLog.html?buildId=25869&buildTypeId=Corda_Build_PullRequests_IntegrationTests

Clintonio:
Indeed. That change should have fixed it, will go back to testing it.

PR #1776, opened 2017-10-02
BREAK RPC API!!!! Muhahaha!!!!

chrisr3:
This BAD CHANGE should fail the "API Change Gate" on TeamCity.


Clintonio:
It failed to fail.

Clintonio:
Ah yes - please rebase. I need to add a fail on error condition.

mnesbit:
Will it catch parameter signature changes?

chrisr3:
Yes, it does.

PR #1778, opened 2017-10-02
Added new API check script and API version

Clintonio:
Why: I want to move the script into TC and the API descriptor is out of date with some V1 merges.


chrisr3:
I think that "Gradlifying" the `diff` script is a Good Idea, _provided_ it can be done in a platform-independent way. At the moment, we are assuming that the API build will only be run on a Linux box.

chrisr3:
What we _don't_ want is for developers to break the API build, and then try to "fix" it by updating the `api-current.txt` file! What I'm thinking is:
- Create a separate repository with restricted commit writes, containing `api-current.txt` and this script.
- Import this repository into `corda-public` as a subtree.

mikehearn:
We can go there if it's a problem in practice. Perhaps a comment at the top of the file is sufficient to tell developers what to do, or a sufficiently clear warning printed by the script. Let's not fuss with ACLs until we find they're really needed.

PR #1783, opened 2017-10-03
Removed the network map service type as it's no longer needed.

shamsasari:
The absence or presence of the networkMapService config is what determines if a node is the network map or not.

This is part of my work to remove the adverisedServices config.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -484,7 +484,7 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
 
     private fun makeNetworkServices(tokenizableServices: MutableList<Any>) {
         val serviceTypes = advertisedServices.map { it.type }
-        inNodeNetworkMapService = if (NetworkMapService.type in serviceTypes) makeNetworkMapService() else NullNetworkMapService
+        inNodeNetworkMapService = if (configuration.networkMapService == null) makeNetworkMapService() else NullNetworkMapService
         val notaryServiceType = serviceTypes.singleOrNull { it.isNotary() }
         if (notaryServiceType != null) {
             val service = makeCoreNotaryService(notaryServiceType)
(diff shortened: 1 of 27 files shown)


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -484,7 +484,7 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
 
     private fun makeNetworkServices(tokenizableServices: MutableList<Any>) {
         val serviceTypes = advertisedServices.map { it.type }
-        inNodeNetworkMapService = if (NetworkMapService.type in serviceTypes) makeNetworkMapService() else NullNetworkMapService
+        inNodeNetworkMapService = if (configuration.networkMapService == null) makeNetworkMapService() else NullNetworkMapService
```
kasiastreich: Isn't @al-r3 PR with pregeneration of `NodeInfo`s for cordformation based on an assumption that `networkMapService` in config is null?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -260,10 +258,10 @@ open class NodeStartup(val args: Array<String>) {
     }
 
     private fun printPluginsAndServices(node: Node) {
-        node.configuration.extraAdvertisedServiceIds.filter { it.startsWith("corda.notary.") || it.startsWith("corda.network_map") }.let {
+        node.configuration.extraAdvertisedServiceIds.filter { it.startsWith("corda.notary.") }.let {
```
kasiastreich: We could print out if node runs network map based on `Node.inNodeNetworkMapService`? It's minor, because network map will be gone soon.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
shamsasari: This nullability assumptions on `networkMapService` has always been there, I haven't changed those, so there shouldn't be anything different for @al-r3

al-r3: I haven't changed any assumptions in AbstractNode regarding the networkMapService,

shamsasari: @al-r3 can you rebase your PR onto master and double-check that it still works.

al-r3: I'll wait for the green light from Clinton, then rebase and redo a full round of checks

PR #1785, opened 2017-10-03
Fixed a few instances of waitUntilNetworkReady() to fix flaky tests.

Clintonio:
Why: Flaky test fixes.


vkolomeyko:
I have started a pack of Integration tests runs:
https://ci-master.corda.r3cev.com/viewType.html?buildTypeId=Corda_Build_PullRequests_IntegrationTests&branch_Corda_Build_PullRequests=1785&tab=buildTypeStatusDiv

Let's see if the test is completely stable.

My guess is that it is unlikely to be. The reason for that is:
`net.corda.core.messaging.CordaRPCOps#waitUntilNetworkReady` may ensure that a given node is registered with NetrworkMap, but it doesn't provide a guarantee that all the sibling nodes are registered with this node's NetworkMap and can be spoken to.

vkolomeyko:
10 of 10 Integration build completed successfully - we clearly have a stability improvement.

PR #1786, opened 2017-10-03
CORDA-654: Add CordaRPCOps.notaryPartyFromX500Name

rnicoll:
Add new function for https://github.com/corda/corda/pull/1744#discussion_r142072757

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
@@ -176,11 +176,7 @@ class CordaRPCOpsImpl(
 
     override fun currentNodeTime(): Instant = Instant.now(services.clock)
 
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
```diff
@@ -200,6 +200,12 @@ class CordaRPCOpsImpl(
         }
     }
 
+    override fun notaryPartyFromX500Name(x500Name: CordaX500Name): Party? {
+        return database.transaction {
+            services.networkMapCache.getNotary(x500Name)
```
kasiastreich: Doesn't have to be in `database.transaction`

rnicoll: Not sure what happened there, second commit must have involved an old version of the file and reverted the code from the first.

PR #1793, opened 2017-10-03
Add README for our automated API check.

chrisr3:
Add README file to warn users not to modify `api-current.txt` because we have committed to API stability.


.ci/README.md
```diff
@@ -0,0 +1,10 @@
+# !! DO NOT MODIFY THE API FILE IN THIS DIRECTORY !!
```
Clintonio: Can you add a caveat that the file will be changed intentionally at releases or under special circumstances that require extra approval? Right now it looks like an immutable file.

PR #1794, opened 2017-10-03
CORDA-530 Don't soft-lock non-fungible states

andr3ej:
* I'm probably missing something here, as effectively I've just added an instanceof
* No strong feelings about this particular design, but we should support all the test cases?


core/src/main/kotlin/net/corda/core/contracts/FungibleAsset.kt
```diff
@@ -39,3 +40,9 @@ interface FungibleAsset<T : Any> : OwnableState {
      */
     fun withNewOwnerAndAmount(newAmount: Amount<Issued<T>>, newOwner: AbstractParty): FungibleAsset<T>
 }
+
+/** May be implemented by states e.g. [FungibleAsset] to influence soft-locking behaviour. */
+interface SoftLockableState {
```
josecoll: Move into separate interface as not _FungibleAsset_ specific.

node/src/test/kotlin/net/corda/node/services/vault/VaultSoftLockManagerTest.kt
```diff
@@ -0,0 +1,123 @@
...
+        override fun verify(tx: LedgerTransaction) {}
+    }
+
+    private fun run(expectSoftLock: Boolean, state: ContractState) {
+        val f = node.services.startFlow(FlowLogicImpl(expectSoftLock, state)).resultFuture
+        mockNet.runNetwork()
+        f.getOrThrow()
+    }
+
+    @Test
```
josecoll: Shouldn't these tests have some assertions? (eg. checking for specific failure behaviour or actual states)

core/src/main/kotlin/net/corda/core/contracts/FungibleAsset.kt
andr3ej: Done

node/src/test/kotlin/net/corda/node/services/vault/VaultSoftLockManagerTest.kt
andr3ej: Moved assert to the foreground, was inside flow

andr3ej:
Plan now that I have more context:
* drop general mechanism in favour of simple instanceof on FungibleAsset, if a user doesn't want their FA soft-locked they can make one from scratch
* the root cause of slowness is H2's handling of multiple states in the query, but we usually soft lock one at a time so it may be worth putting effort into a fast path for that case
  * i reckon that's enforceable via carefully crafted unit test, which should be attempted first

josecoll:
Not sure I agree with your statement, re: "we usually soft lock one at a time so it may be worth putting effort into a fast path for that case"
Both call paths to `softLock` (see `unconsumedCashStatesForSpending` and `tryLockFungibleStatesForSpending`) specify a list of states which are combined in a single SQL Update statement.

josecoll:
@andr3ej what is the status of this request?
Did you check with Matthew and Andras with regards to benchmarking this optimisation?
eg. what load tests scenarios to re-run to validate improvement.

andr3ej:
I checked with Matthew. According to him H2 executes the SQL update inefficiently which is a performance bottleneck in practice. That's where the above plan comes from. On hold at the moment as #1839 is higher priority.

andr3ej:
The latest commit adds a unit test enforcing that the unlock query doesn't run if no states were locked. (That behaviour already existed, the only functionality added by this PR is that non-FungibleAsset states aren't auto-soft-locked.) So now a flow with no FungibleAssets won't touch the database at all via the vault soft-lock api.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -274,8 +274,7 @@ abstract class AbstractNode(config: NodeConfiguration,
                 require(customNotaryServiceList.size == 1) {
                     "Attempting to install more than one notary service: ${customNotaryServiceList.joinToString()}"
                 }
-            }
-            else return loadedServices - customNotaryServiceList
+            } else return loadedServices - customNotaryServiceList
```
andr3ej: formatter

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -490,9 +489,9 @@ abstract class AbstractNode(config: NodeConfiguration,
     protected open fun makeTransactionStorage(): WritableTransactionStorage = DBTransactionStorage()
 
     private fun makeVaultObservers() {
-        VaultSoftLockManager(services.vaultService, smm)
-        ScheduledActivityObserver(services)
-        HibernateObserver(services.vaultService.rawUpdates, services.database.hibernateConfig)
+        VaultSoftLockManager.install(services.vaultService, smm)
+        ScheduledActivityObserver.install(services.vaultService, services.schedulerService)
+        HibernateObserver.install(services.vaultService.rawUpdates, database.hibernateConfig)
```
andr3ej: constructors should not have side effects

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -68,12 +68,10 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
...
     override val logger: Logger = LoggerFactory.getLogger("net.corda.flow.$id")
-
-    @Transient private var _resultFuture: OpenFuture<R>? = openFuture()
+    @Transient private var resultFutureTransient: OpenFuture<R>? = openFuture()
+    private val _resultFuture get() = resultFutureTransient ?: openFuture<R>().also { resultFutureTransient = it }
     /** This future will complete when the call method returns. */
-    override val resultFuture: CordaFuture<R>
-        get() = _resultFuture ?: openFuture<R>().also { _resultFuture = it }
-
+    override val resultFuture: CordaFuture<R> get() = _resultFuture
```
andr3ej: these changes are so that VaultSoftLockManagerTest doesn't get a resultFuture that never completes

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -269,7 +269,7 @@ class NodeVaultService(private val clock: Clock, private val keyManagementServic
                 update.where(stateStatusPredication, lockIdPredicate, *commonPredicates)
             }
             if (updatedRows > 0 && updatedRows == stateRefs.size) {
-                log.trace("Reserving soft lock states for $lockId: $stateRefs")
+                log.trace { "Reserving soft lock states for $lockId: $stateRefs" }
```
andr3ej: creating an object should be much faster than string formatting

node/src/test/kotlin/net/corda/node/services/vault/VaultSoftLockManagerTest.kt
```diff
@@ -0,0 +1,161 @@
...
+        server.internals.internalRegisterFlowFactory(AbstractClientLogic::class.java, InitiatedFlowFactory.Core { ServerLogic(it, serverRunning) }, ServerLogic::class.java, false)
+        client.services.startFlow(clientLogic)
+        while (!serverRunning.get()) mockNet.runNetwork(1)
+        if (rebootClient) {
+            client.dispose()
+            client = mockNet.createNode(client.internals.id)
+        }
+        return uncheckedCast(client.smm.allStateMachines.single().stateMachine)
+    }
+}
```
andr3ej: i may port AttachmentSerializationTest to use this in a separate PR

node/src/test/kotlin/net/corda/node/services/vault/VaultSoftLockManagerTest.kt
```diff
@@ -0,0 +1,161 @@
...
+    }
+        private set
+
+    fun <T> communicate(clientLogic: AbstractClientLogic<T>, rebootClient: Boolean): FlowStateMachine<T> {
+        server.internals.internalRegisterFlowFactory(AbstractClientLogic::class.java, InitiatedFlowFactory.Core { ServerLogic(it, serverRunning) }, ServerLogic::class.java, false)
+        client.services.startFlow(clientLogic)
+        while (!serverRunning.get()) mockNet.runNetwork(1)
+        if (rebootClient) {
+            client.dispose()
+            client = mockNet.createNode(client.internals.id)
```
andr3ej: reboot needed so the flow is actually loaded from the db

PR #1795, opened 2017-10-03
[CORDA-519] Hash agility - step1

kchalkias:
This is the first part for hash agility enabling `.hash` that uses the default SHA256 algorithm.
Application specific calls of `SecureHash` were remained intact and they still use `.sha256`.
Leaf computation is using the default algorithm.
This is step 1: in an upcoming PR, the Merkle tree computation will take the hash algorithm as input and will validate data, also the tx should know which algorithm is used, probably this information will be a metadata leaf.
See [here ](https://r3-cev.atlassian.net/wiki/spaces/AWG/pages/132068319/Hash+Agility) for the related design doc. We should agree on the HEX prefix method + store in DBs.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
@@ -71,23 +71,11 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
 
         /**
          * Method that receives a HEX [String] which represents a hash output and returns its corresponding [SecureHash]
-         * object. Selection of the Hash algorithm is based on the input String size and if there are more than one hash
-         * functions offering the same hash output size, then the default per size is selected.
+         * object. The hash secureHashAlgorithm is defined by the input [secureHashAlgorithm]. If the algorithm is
+         * not provided, then the [DEFAULT_ALGORITHM] is used.
          */
         @JvmStatic
-        fun parse(str: String) = str.toUpperCase().parseAsHex().let {
-            when (it.size) {
-                32 -> SHA256(it)
-                64 -> SHA512(it) // TODO: define default 64-bit SecureHash function when SHA3-512 is introduced.
-                else -> throw IllegalArgumentException("Provided string is ${it.size} bytes not 32 or 64 bytes in hex: $str")
-            }
-        }
-
-        /**
-         * Method that receives a HEX [String] which represents a hash output and returns its corresponding [SecureHash]
-         * object. The hash secureHashAlgorithm is defined by the input [secureHashAlgorithm].
-         */
-        @JvmStatic
+        @JvmOverloads
         fun parse(str: String, secureHashAlgorithm: SecureHashAlgorithm = DEFAULT_ALGORITHM) = str.toUpperCase().parseAsHex().let {
             when(secureHashAlgorithm) {
                 SHA256_ALGORITHM -> SHA256(it)
@@ -130,7 +118,11 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
          */
         @JvmStatic
         fun randomHash(secureHashAlgorithm: SecureHashAlgorithm = DEFAULT_ALGORITHM): SecureHash =
-                hash(newSecureRandom().generateSeed(secureHashAlgorithm.outputSize), secureHashAlgorithm)
+            when (secureHashAlgorithm) {
+                SHA256_ALGORITHM -> SHA256(secureRandomBytes(secureHashAlgorithm.outputSize))
+                SHA512_ALGORITHM -> SHA512(secureRandomBytes(secureHashAlgorithm.outputSize))
+                else -> throw IllegalArgumentException("Hash secureHashAlgorithm $secureHashAlgorithm is not supported")
+            }
 
         /** Computes the hash output of the input [bytes] and returns the corresponding [SHA256] object. */
         @JvmStatic fun sha256(bytes: ByteArray) = hash(bytes, SHA256_ALGORITHM) as SHA256
@@ -164,6 +156,8 @@ fun ByteArray.hash(): SecureHash = SecureHash.hash(this, SecureHash.DEFAULT_ALGO
 fun OpaqueBytes.hash(): SecureHash = this.bytes.hash()
 
 /** Compute the hash of this [ByteArray] using the [SecureHash.SHA256_ALGORITHM] and return its corresponding [SecureHash.SHA256] object. */
+@Deprecated("Use hash(SecureHash.SHA256_ALGORITHM)", level = DeprecationLevel.WARNING)
 fun ByteArray.sha256(): SecureHash.SHA256 = SecureHash.sha256(this)
 /** Compute the hash of this [OpaqueBytes] object using the [SecureHash.SHA256_ALGORITHM] and return its corresponding [SecureHash.SHA256] object. */
+@Deprecated("Use hash(SecureHash.SHA256_ALGORITHM)", level = DeprecationLevel.WARNING)
 fun OpaqueBytes.sha256(): SecureHash.SHA256 = this.bytes.sha256()
--- a/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
@@ -5,13 +5,21 @@ import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.utilities.parseAsHex
 import net.corda.core.utilities.toHexString
 import java.security.MessageDigest
+import kotlin.reflect.KClass
 
 /**
  * Container for a cryptographically secure hash value.
- * Provides utilities for generating a cryptographic hash using different algorithms (currently only SHA-256 supported).
+ * Provides utilities for generating a cryptographic hash using different algorithms.
+ * Currently, the following cryptographic hash algorithms are supported:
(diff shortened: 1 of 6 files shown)


kchalkias:
@mikehearn api stability complains due to generalisation from `SecureHash.SHA256` to `SecureHash`, what's our strategy on that?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -5,46 +5,165 @@ import net.corda.core.utilities.OpaqueBytes
...
+        @JvmStatic
+        fun hashTwice(bytes: ByteArray, secureHashAlgorithm: SecureHashAlgorithm = DEFAULT_ALGORITHM): SecureHash =
+                hash(hash(bytes, secureHashAlgorithm).bytes, secureHashAlgorithm)
+
+        /**
+         * Returns a random [SecureHash] object using the provided [SecureHashAlgorithm]. If the algorithm
+         * is not provided, then the [DEFAULT_ALGORITHM] is used.
+         */
+        @JvmStatic
+        fun randomHash(secureHashAlgorithm: SecureHashAlgorithm = DEFAULT_ALGORITHM): SecureHash =
```
mnesbit: This seems pretty pointless, since SecurerRandom will typically involve a hashing step?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -5,46 +5,165 @@ import net.corda.core.utilities.OpaqueBytes
...
+        /** SHA256 is currently the default [SecureHash] algorithm. */
+        val DEFAULT_ALGORITHM = SHA256_ALGORITHM
+
+        /**
+         * Method that receives a HEX [String] which represents a hash output and returns its corresponding [SecureHash]
+         * object. Selection of the Hash algorithm is based on the input String size and if there are more than one hash
+         * functions offering the same hash output size, then the default per size is selected.
+         */
         @JvmStatic
         fun parse(str: String) = str.toUpperCase().parseAsHex().let {
```
mnesbit: Can we perhaps put a prefix on the SecureHash.toString, thus allowing a more type safe parse?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -5,46 +5,165 @@ import net.corda.core.utilities.OpaqueBytes
...
+/** Compute the hash of this [ByteArray] using the [SecureHash.DEFAULT_ALGORITHM] and return its corresponding [SecureHash] object. */
+fun ByteArray.hash(): SecureHash = SecureHash.hash(this, SecureHash.DEFAULT_ALGORITHM)
+/** Compute the hash of this [OpaqueBytes] object using the [SecureHash.DEFAULT_ALGORITHM] and return its corresponding [SecureHash] object. */
+fun OpaqueBytes.hash(): SecureHash = this.bytes.hash()
+
+/** Compute the hash of this [ByteArray] using the [SecureHash.SHA256_ALGORITHM] and return its corresponding [SecureHash.SHA256] object. */
 fun ByteArray.sha256(): SecureHash.SHA256 = SecureHash.sha256(this)
-fun OpaqueBytes.sha256(): SecureHash.SHA256 = SecureHash.sha256(this.bytes)
+/** Compute the hash of this [OpaqueBytes] object using the [SecureHash.SHA256_ALGORITHM] and return its corresponding [SecureHash.SHA256] object. */
+fun OpaqueBytes.sha256(): SecureHash.SHA256 = this.bytes.sha256()
```
mnesbit: Can we deprecate?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -5,46 +5,165 @@ import net.corda.core.utilities.OpaqueBytes
...
+ * @property allOnesHash a flag-object used as a root of an empty Merkle tree (without leaves). By convention, this is a [SecureHash] object with all bits set to one.
+ */
+data class SecureHashAlgorithm(val algorithmName: String, val outputSize: Int, val zeroHash: SecureHash, val allOnesHash: SecureHash)
+
+/** Compute the hash of this [ByteArray] using the [SecureHash.DEFAULT_ALGORITHM] and return its corresponding [SecureHash] object. */
+fun ByteArray.hash(): SecureHash = SecureHash.hash(this, SecureHash.DEFAULT_ALGORITHM)
+/** Compute the hash of this [OpaqueBytes] object using the [SecureHash.DEFAULT_ALGORITHM] and return its corresponding [SecureHash] object. */
+fun OpaqueBytes.hash(): SecureHash = this.bytes.hash()
+
+/** Compute the hash of this [ByteArray] using the [SecureHash.SHA256_ALGORITHM] and return its corresponding [SecureHash.SHA256] object. */
```
mnesbit: Can we deprecate?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -5,46 +5,165 @@ import net.corda.core.utilities.OpaqueBytes
...
+     */
+    fun hashConcat(other: SecureHash): SecureHash {
+        return if (this::class == other::class) {
+            SecureHash.hash(this.bytes + other.bytes, this::class)
+        } else throw IllegalArgumentException("You can only hashConcat SecureHash objects of the same type.")
+    }
+
     companion object {
+        /** Properties of the [SHA256] algorithm. */
+        val SHA256_ALGORITHM = SecureHashAlgorithm(
```
mnesbit: Should we have a SHA2-256 implementation?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
kchalkias: Hm, `generateSeed` returns a random seed to the hash function, but construction of this seed might not include hashing necessarily in the future, although it does now as we only use nonBlocking seed generation. But, my sense is we are actually agnostic on how this random seed is produced, we just need some random value with enough entropy to avoid reuse of the same seed. We also seed it with enough random bytes `generateSeed(secureHashAlgorithm.outputSize)`.

On the other hand, it's true we currently use it for testing purposes only, with the exception of `TransactionViewer` (again for a non critical operation). But it's use is usually for `StateRef(SecureHash.randomSHA256(), Random().nextInt(32))` which means, that if we change to SHA512 or any other, we should support  a generic `randomHash`.

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
kchalkias: SHA256 is SHA2-256, there is no SHA1-256.
ported from wikipedia: _SHA-2 includes significant changes from its predecessor, SHA-1. The SHA-2 family consists of six hash functions with digests (hash values) that are 224, 256, 384 or 512 bits: SHA-224, SHA-256, SHA-384, SHA-512, SHA-512/224, SHA-512/256_

kchalkias:
@mnesbit, I removed `parse(str)` and I now use the overloaded
`parse(str: String, secureHashAlgorithm: SecureHashAlgorithm = DEFAULT_ALGORITHM)` to avoid size checks for the time being.
I had a discussion with @rick-r3 and need to evaluate various suggested alternatives, such as
1. prefix Vs suffix
2. 1 or 2 bytes for identification
3. Do not allow HexCharacters [0-9a-fA-F] on the identification bytes.
4. No {pre,suf}-fix and always provide the algorithm; thus, 2 cols in the DB

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
mnesbit: sorry I mean SHA3-256

mikehearn:
Is there a design doc for this feature? It seems to me that making hash functions pluggable has quite a large impact on many things and there should be a design doc that takes into account our compatibility constraints.

The inferred types for some of those functions being SHA256 specifically is unfortunate, but this is telling us something important - these functions *have* in fact always returned 256 bit wide hashes, and it's entirely reasonable that people may be already relying on that. E.g. for database column widths.

I don't think hash agility is a feature we can just start writing PRs for, even though this one may be fine. We shouldn't just start dropping 64 byte hashes on people unexpectedly, especially as there's no justification for such a change at this time!

kchalkias:
Thanks @mikehearn, I will park this and re-open after we agree on the design doc. I've some design notes already (was about to send on an upcoming MTree + Vault update PR to support hashes up to 64 bytes).

mnesbit:
My concern is that without hash agility we can't have algorithmic agility. Our Tx ID's are basically hashes, so we are limited in how we can change the merkle trees. Also, it is ambiguous what a 256-bit hash is. How would we get onto SHA3-256? Also, people are hard coding the Any.serialize.sha256 everywhere which is really dangerous and inflexible.
I totally agree it needs design and I had wanted this work to have been done before 1.0.

mikehearn:
@mnesbit fully agree with the concerns, I'd just like to see this thought out in a bit more depth first.

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
kchalkias: @mnesbit SHA3 is not supported by the default "SUN" provider of Java 8, it is now in Java 9 though. Of course, I could use BouncyCastle, are you ok with it?

PR #1796, opened 2017-10-03
[CORDA-442] DemoBench copies around the NodeInfos for running nodes.

al-r3:
NodeBench now copies the nodeInfo files of running nodes to each other node.

In detail, each node writes its own nodeInfo-<HASH> to its base folder and polls the filesystem for new ones in <nodepath>/additional-node-infos

Demobench now monitors the nodes path, waiting for them to write their own nodeInfo and then copies each nodeInfo to all other running nodes' additional-node-info directories.

(nodes polling the filesystem is happening here https://github.com/corda/corda/pull/1623, however the logic in this PR is independent of that)

Note that currently DemoBench makes the first node to start the 'networkmapnode' this PR is in preparation for when the 'networkmapnode' will be dismantled

### DIFF
--- a/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
+++ b/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
@@ -1,0 +1,126 @@
+package net.corda.demobench.model
+
+import net.corda.cordform.CordformNode
+import net.corda.core.internal.createDirectories
+import net.corda.core.internal.div
+import net.corda.core.internal.isRegularFile
+import net.corda.core.internal.list
+import rx.Observable
+import rx.Scheduler
+import rx.schedulers.Schedulers
+import tornadofx.*
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.StandardCopyOption
+import java.nio.file.attribute.BasicFileAttributes
+import java.nio.file.attribute.FileTime
+import java.util.concurrent.TimeUnit
+import java.util.logging.Level
+
+/**
+ * Utility class which copies nodeInfo files across a set of running nodes.
+ *
+ * This class will create paths that it needs to poll and to where it needs to copy files in case those
+ * don't exist yet.
+ */
+class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() {
+
+    private val nodeDataMap = mutableMapOf<Path, NodeData>()
+
+    init {
+        Observable.interval(5, TimeUnit.SECONDS, scheduler)
+                .subscribe { poll() }
+    }
+
+    /**
+     * @param nodeConfig the configuration to be added.
+     * Add a [NodeConfig] for a node which is about to be started.
+     * Its nodeInfo file will be copied to other nodes' additional-node-infos directory, and conversely,
+     * other nodes' nodeInfo files will be copied to this node additional-node-infos directory.
+     */
+    @Synchronized
+    fun addConfig(nodeConfig: NodeConfigWrapper) {
+        val newNodeFile = NodeData(nodeConfig.nodeDir)
+        nodeDataMap[nodeConfig.nodeDir] = newNodeFile
+
+        for (previouslySeenFile in allPreviouslySeenFiles()) {
+            copy(previouslySeenFile, newNodeFile.destination.resolve(previouslySeenFile.fileName))
+        }
+        log.info("Now watching: ${nodeConfig.nodeDir}")
+    }
+
+    /**
+     * @param nodeConfig the configuration to be removed.
+     * Remove the configuration of a node which is about to be stopped or already stopped.
+     * No files written by that node will be copied to other nodes, nor files from other nodes will be copied to this
+     * one.
+     */
+    @Synchronized
+    fun removeConfig(nodeConfig: NodeConfigWrapper) {
+        nodeDataMap.remove(nodeConfig.nodeDir) ?: return
+        log.info("Stopped watching: ${nodeConfig.nodeDir}")
+    }
+
+    @Synchronized
+    fun reset() {
+        nodeDataMap.clear()
+    }
+
+    private fun allPreviouslySeenFiles() = nodeDataMap.values.map { it.previouslySeenFiles.keys }.flatten()
+
+    @Synchronized
+    private fun poll() {
+        for (nodeData in nodeDataMap.values) {
+            nodeData.nodeDir.list { paths ->
+                paths.filter { it.isRegularFile() }
+                        .filter { it.fileName.toString().startsWith("nodeInfo-") }
+                        .forEach { path -> processPath(nodeData, path) }
+            }
+        }
+    }
+
+    // Takes a path under nodeData config dir and decides whether the file represented by that path needs to
+    // be copied.
+    private fun processPath(nodeData: NodeData, path: Path) {
+        val newTimestamp = Files.readAttributes(path, BasicFileAttributes::class.java).lastModifiedTime()
+        val previousTimestamp = nodeData.previouslySeenFiles.put(path, newTimestamp) ?: FileTime.fromMillis(-1)
+        if (newTimestamp > previousTimestamp) {
+            for (destination in nodeDataMap.values.filter { it.nodeDir != nodeData.nodeDir }.map { it.destination }) {
+                val fullDestinationPath = destination.resolve(path.fileName)
+                copy(path, fullDestinationPath)
+            }
+        }
+    }
+
+    private fun copy(source: Path, destination: Path) {
+        val tempDestination = destination.parent / ("." + destination.fileName.toString())
+        try {
+            // First copy the file to a temporary file within the appropriate directory.
+            Files.copy(source, tempDestination, StandardCopyOption.COPY_ATTRIBUTES)
+        } catch (exception: Exception) {
+            log.log(Level.WARNING, "Couldn't copy $source to $tempDestination.", exception)
+            throw exception
+        }
+        try {
+            // Then rename it to the desired name. This way the file 'appears' on the filesystem as an atomic operation.
+            Files.move(tempDestination, destination, StandardCopyOption.COPY_ATTRIBUTES, StandardCopyOption.REPLACE_EXISTING)
+        } catch (exception: Exception) {
+            log.log(Level.WARNING, "Couldn't move $tempDestination to $destination.", exception)
+            Files.delete(tempDestination)
+            throw exception
+        }
+    }
+
+    /**
+     * Convenience holder for all the paths and files relative to a single node.
+     */
+    private class NodeData(val nodeDir: Path) {
+        val destination: Path = nodeDir.resolve(CordformNode.NODE_INFO_DIRECTORY)
+        // Map from Path to its lastModifiedTime.
+        val previouslySeenFiles = mutableMapOf<Path, FileTime>()
+
+        init {
+            destination.createDirectories()
+        }
+    }
+}
--- a/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
+++ b/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
@@ -105,6 +105,7 @@ class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() 
             Files.copy(source, tempDestination, COPY_ATTRIBUTES, REPLACE_EXISTING)
         } catch (exception: IOException) {
             log.log(Level.WARNING, "Couldn't copy $source to $tempDestination.", exception)
+            Files.delete(tempDestination)
             throw exception
         }
         try {
--- a/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
+++ b/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
@@ -2,7 +2,6 @@ package net.corda.demobench.model
 
 import net.corda.cordform.CordformNode
 import net.corda.core.internal.createDirectories
-import net.corda.core.internal.div
 import net.corda.core.internal.isRegularFile
 import net.corda.core.internal.list
 import rx.Observable
@@ -95,7 +94,12 @@ class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() 
     }
 
     private fun copy(source: Path, destination: Path) {
-        val tempDestination = destination.parent / ("." + destination.fileName.toString())
+        val tempDestination = try {
+            Files.createTempFile(destination.parent, ".", null)
+        } catch (exception: IOException) {
+            log.log(Level.WARNING, "Couldn't create a temporary file to copy $source", exception)
+            throw exception
+        }
         try {
             // First copy the file to a temporary file within the appropriate directory.
             Files.copy(source, tempDestination, COPY_ATTRIBUTES, REPLACE_EXISTING)
--- a/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
+++ b/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
@@ -9,9 +9,11 @@ import rx.Observable
 import rx.Scheduler
 import rx.schedulers.Schedulers
 import tornadofx.*
+import java.io.IOException
 import java.nio.file.Files
 import java.nio.file.Path
-import java.nio.file.StandardCopyOption
+import java.nio.file.StandardCopyOption.COPY_ATTRIBUTES
+import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.attribute.FileTime
 import java.util.concurrent.TimeUnit
@@ -96,15 +98,15 @@ class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() 
         val tempDestination = destination.parent / ("." + destination.fileName.toString())
         try {
             // First copy the file to a temporary file within the appropriate directory.
-            Files.copy(source, tempDestination, StandardCopyOption.COPY_ATTRIBUTES)
-        } catch (exception: Exception) {
+            Files.copy(source, tempDestination, COPY_ATTRIBUTES, REPLACE_EXISTING)
+        } catch (exception: IOException) {
             log.log(Level.WARNING, "Couldn't copy $source to $tempDestination.", exception)
             throw exception
         }
         try {
             // Then rename it to the desired name. This way the file 'appears' on the filesystem as an atomic operation.
-            Files.move(tempDestination, destination, StandardCopyOption.REPLACE_EXISTING)
-        } catch (exception: Exception) {
+            Files.move(tempDestination, destination, REPLACE_EXISTING)
+        } catch (exception: IOException) {
             log.log(Level.WARNING, "Couldn't move $tempDestination to $destination.", exception)
             Files.delete(tempDestination)
             throw exception
--- a/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
+++ b/tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
@@ -103,7 +103,7 @@ class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() 
         }
         try {
             // Then rename it to the desired name. This way the file 'appears' on the filesystem as an atomic operation.
-            Files.move(tempDestination, destination, StandardCopyOption.COPY_ATTRIBUTES, StandardCopyOption.REPLACE_EXISTING)
+            Files.move(tempDestination, destination, StandardCopyOption.REPLACE_EXISTING)
         } catch (exception: Exception) {
             log.log(Level.WARNING, "Couldn't move $tempDestination to $destination.", exception)
             Files.delete(tempDestination)
--- a/tools/demobench/src/test/kotlin/net/corda/demobench/model/NodeInfoFilesCopierTest.kt
+++ b/tools/demobench/src/test/kotlin/net/corda/demobench/model/NodeInfoFilesCopierTest.kt
@@ -1,0 +1,156 @@
+package net.corda.demobench.model
+
+import net.corda.cordform.CordformNode
+import net.corda.core.identity.CordaX500Name
+import net.corda.core.utilities.NetworkHostAndPort
+import net.corda.testing.eventually
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import rx.schedulers.TestScheduler
+import java.nio.file.Files
+import java.nio.file.Path
+import java.time.Duration
+import java.util.concurrent.TimeUnit
+import kotlin.streams.toList
+import kotlin.test.assertEquals
+
+/**
+ * tests for [NodeInfoFilesCopier]
+ */
+class NodeInfoFilesCopierTest {
+
+    @Rule @JvmField var folder = TemporaryFolder()
+    private val rootPath get() = folder.root.toPath()
+    private val scheduler = TestScheduler()
+    companion object {
+        private const val ORGANIZATION = "Organization"
+        private const val NODE_1_PATH = "node1"
+        private const val NODE_2_PATH = "node2"
+
+        private val content = "blah".toByteArray(Charsets.UTF_8)
+        private val GOOD_NODE_INFO_NAME = "nodeInfo-test"
+        private val GOOD_NODE_INFO_NAME_2 = "nodeInfo-anotherNode"
+        private val BAD_NODE_INFO_NAME = "something"
+        private val legalName = CordaX500Name(organisation = ORGANIZATION, locality = "Nowhere", country = "GB")
+        private val hostAndPort = NetworkHostAndPort("localhost", 1)
+    }
+
+    private fun nodeDir(nodeBaseDir : String) = rootPath.resolve(nodeBaseDir).resolve(ORGANIZATION.toLowerCase())
+
+    private val node1Config by lazy { createConfig(NODE_1_PATH) }
+    private val node2Config by lazy { createConfig(NODE_2_PATH) }
+    private val node1RootPath by lazy { nodeDir(NODE_1_PATH) }
+    private val node2RootPath by lazy { nodeDir(NODE_2_PATH) }
+    private val node1AdditionalNodeInfoPath by lazy { node1RootPath.resolve(CordformNode.NODE_INFO_DIRECTORY) }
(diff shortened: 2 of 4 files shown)


tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,116 @@
...
+import java.util.concurrent.TimeUnit
+
+
+/**
+ * Utility class which copies nodeInfo files across a set of running nodes.
+ *
+ * This class will create paths that it needs to poll and to where it needs to copy files in case those
+ * don't exist yet.
+ */
+class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()) {
```
chrisr3: Make into a `Controller` bean:
```kotlin
class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()) : Controller() {
   ...
}
```

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,116 @@
...
+                    }
+                }
+            }
+            if (!watchKey.reset()) {
+                logger.warn("Couldn't reset watchKey for path ${watchTarget.path}, it was probably deleted.")
+            }
+        }
+    }
+
+    private fun copy(source : Path, destination: Path) {
```
chrisr3: Remove space between `source` and `:` for consistency.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            // REPLACE_EXISTING is needed in case we copy a file being written and we need to overwrite it with the
+            // "full" file.
+            Files.copy(source, destination, StandardCopyOption.REPLACE_EXISTING)
+        } catch (exception : Exception) {
+            logger.warn("Couldn't copy $source to $destination. Exception: $exception")
+        }
+    }
+
+    private fun addDestination(destination : Path) {
```
chrisr3: Remove space between `destination` and `:` for consistency.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,116 @@
+package net.corda.demobench.model
+
+import net.corda.cordform.CordformNode
+import net.corda.core.internal.createDirectories
+import net.corda.core.internal.isDirectory
+import net.corda.core.internal.isRegularFile
+import net.corda.core.utilities.loggerFor
+import rx.Observable
+import rx.Scheduler
+import rx.Subscription
```
chrisr3: Remove unused import `rx.Subscription`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,116 @@
...
+    companion object {
+        private val logger = loggerFor<NodeInfoFilesCopier>()
+    }
+
+    // Paths we are copying into.
+    private val destinations = mutableListOf<Path>()
+    // WatchServices watching paths we are watching.
+    private val watchTargets = mutableListOf<WatchTarget>()
+    // Set of paths of files which have been copied from already..
+    private val previouslySeenFiles = mutableSetOf<Path>()
```
chrisr3: You add to these lists but never empty them. What happens if you create some nodes, save them as a profile and then reload the profile? Would DemoBench continue polling the paths belonging to the old (dead) nodes?

al-r3:
All done, PTAL

tools/demobench/src/test/kotlin/net/corda/demobench/model/NodeInfoFilesCopierTest.kt
```diff
@@ -0,0 +1,154 @@
...
+        // Create 2 files, a nodeInfo and another file in node1 folder.
+        Files.write(node1RootPath.resolve(GOOD_NODE_INFO_NAME), content)
+        Files.write(node1RootPath.resolve(BAD_NODE_INFO_NAME), content)
+
+        // Configure the second node.
+        nodeInfoFilesCopier.addConfig(node2Config)
+        advanceTime()
+
+        // Give some time to the filesystem to report the change.
+        Thread.sleep(100)
```
chrisr3: Do you need this sleep? Aren't you polling below?

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Utility class which copies nodeInfo files across a set of running nodes.
+ *
+ * This class will create paths that it needs to poll and to where it needs to copy files in case those
+ * don't exist yet.
+ */
+class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() {
+
+    class NodeWasNeverRegisteredException: IllegalStateException()
```
chrisr3: Unused exception, please remove.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * @param nodeConfig the configuration to be removed.
+     * Remove the configuration of a node which is about to be stopped or already stopped.
+     * No files written by that node will be copied to other nodes, nor files from other nodes will be copied to this
+     * one.
+     */
+    @Synchronized
+    fun removeConfig(nodeConfig: NodeConfig) {
+        val nodeData = nodeDataMap.get(nodeConfig.nodeDir) ?: return
```
chrisr3: I think this can be better written as:
```kotlin
val nodeData = nodeDataMap.remove(nodeConfig.nodeDir) ?: return
nodeData.watchService.close()
logger.info("Stopped watching: ${nodeConfig.nodeDir}")
```
You should also note that `watchService.close()` can potentially throw `IOException`. However, not seeing this log message is probably harmless in that case. It's almost certainly more harmless than leaving the data inside `nodeDataMap`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+            val path = nodeFile.nodeDir
+            val watchService = nodeFile.watchService
+            val watchKey: WatchKey = watchService.poll() ?: continue
+
+            for (event in watchKey.pollEvents()) {
+                val kind = event.kind()
+                if (kind == StandardWatchEventKinds.OVERFLOW) continue
+
+                @Suppress("UNCHECKED_CAST")
+                val fileName: Path = (event as WatchEvent<Path>).context()
```
chrisr3: Replace `@Suppress("UNCHECKED_CAST")` with
```kotlin
val fileName: Path = uncheckedCast<WatchEvent<*>, WatchEvent<Path>>(event).context()
```

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        val watchService: WatchService
+        val destination: Path
+        val previouslySeenFiles = mutableSetOf<Path>()
+
+        init {
+            this.watchService = initWatch(nodeDir)
+            this.destination = nodeDir.resolve(CordformNode.NODE_INFO_DIRECTORY)
+            destination.createDirectories()
```
chrisr3: Are you checking whether these directories have been created successfully?

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Convenience holder for all the paths and files relative to a single node.
+     */
+    private class NodeData(val nodeDir: Path) {
+        companion object {
+            private fun initWatch(path: Path): WatchService {
+                if (!path.isDirectory()) {
+                    logger.info("Creating $path which doesn't exist.")
+                    path.createDirectories()
```
chrisr3: What happens if the directories could not be created for any reason?

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * @param nodeConfig the configuration to be added.
+     * Add a [NodeConfig] for a node which is about to be started.
+     * Its nodeInfo file will be copied to other nodes' additional-node-infos directory, and conversely,
+     * other nodes' nodeInfo files will be copied to this node additional-node-infos directory.
+     */
+    @Synchronized
+    fun addConfig(nodeConfig: NodeConfig) {
+        val newNodeFile = NodeData(nodeConfig.nodeDir)
+        nodeDataMap.put(nodeConfig.nodeDir, newNodeFile)
```
chrisr3: Alternatively, a more Kotliny syntax is:
```kotlin
nodeDataMap[nodeConfig.nodeDir] = newNodeFile
```

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Convenience holder for all the paths and files relative to a single node.
+     */
+    private class NodeData(val nodeDir: Path) {
+        companion object {
+            private fun initWatch(path: Path): WatchService {
+                if (!path.isDirectory()) {
+                    logger.info("Creating $path which doesn't exist.")
```
chrisr3: No need to say "which doesn't exist" here - that's implicit in the fact there we've just said we're creating it.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+                .subscribe { poll() }
+    }
+
+    /**
+     * @param nodeConfig the configuration to be added.
+     * Add a [NodeConfig] for a node which is about to be started.
+     * Its nodeInfo file will be copied to other nodes' additional-node-infos directory, and conversely,
+     * other nodes' nodeInfo files will be copied to this node additional-node-infos directory.
+     */
+    @Synchronized
```
chrisr3: This is a "big" function to synchronize, and it's full of I/O. What are you protecting with this? Just the map? Do we care if `nodeConfig.nodeDir` already exists inside the map?

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,145 @@
...
+            }
+            if (!watchKey.reset()) {
+                logger.warn("Couldn't reset watchKey for path $path, it was probably deleted.")
+            }
+        }
+    }
+
+    private fun copy(source: Path, destination: Path) {
+        try {
+            // REPLACE_EXISTING is needed in case we copy a file being written and we need to overwrite it with the
```
chrisr3: Yikes! Can this scenario actually happen? How would we end up with multiple copies of the same file simultaneously being written to the same location? Because that sounds like a nasty bug to me!

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
al-r3: This code runs on in a thread pool from observables (see Schedulers.io()) that is meant for blocking IO so this usage should be fine.
The Synchronized are mainly to protect the map, yes

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
al-r3: oh good to know :)

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
al-r3: It could.
1 Process A starts to write file X
2 Process B sees a file X and copies what is sees (which is just the first part of X)
3 Process A finishes writing file X
4 Process B now sees the whole file and should do a copy overwriting the target file

The fundamental issue is that I can't know whether a file has been completely written.
Note that these files are minuscule (~4Kb) so I think this event is next to impossible, though I think it's better to code assuming it might happen.

tools/demobench/src/test/kotlin/net/corda/demobench/model/NodeInfoFilesCopierTest.kt
al-r3: seems these tests are hard (if not impossible) to make work on a mac, we're having a separate discussion on how to handle them, I'll wait to see how we address the ones that I've already written and apply that solution in this case.

I might end up dropping the watch service entirely and simply reread the directory contents every now and then

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
al-r3: this is the node root directory if it's not created here I expect node bench will fail to create them as well and the node won't even be deployed.
There isn't much meaningful handling of this exception I can do here, I think

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
al-r3: No, but (see above) there isn't much I can think I can do if it fails.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
chrisr3: But isn't DemoBench single-threaded? Where are these other threads coming from?

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
chrisr3: You can always log a message.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
chrisr3: Then maybe synchronise only the map access?

tools/demobench/src/test/kotlin/net/corda/demobench/model/NodeInfoFilesCopierTest.kt
```diff
@@ -0,0 +1,154 @@
...
+        nodeInfoFilesCopier.addConfig(node2Config)
+        advanceTime()
+
+        nodeInfoFilesCopier.reset()
+
+        advanceTime()
+        Files.write(node2RootPath.resolve(GOOD_NODE_INFO_NAME_2), content)
+
+        // Give some time to the filesystem to report the change.
+        Thread.sleep(100)
```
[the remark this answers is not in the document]
al-r3: I've removed all the sleep but one.
`eventually` works when I need to check something will change, it doesn't let me write negative tests where I check for something not to happen.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
al-r3: I'm iterating over the map, so that doesn't work out of the box.
Keep in mind that there might be threads waiting only at node start/stop time which are fairly scattered.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
al-r3: That is correct and I can change the writing logic to do that.
However I wouldn't go and change this code, if I made a single copy I'd be assuming that the writes are atomic, which might be true or not.
Making a single copy is more fragile overall, it seems.

al-r3:
I've replaced all the polling logic with just re-reading the folder every time, seems polling is somewhat unreliable on MacOS (it's probably just slow, but hard to test)

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
chrisr3: I actually just meant
```kotlin
fun addConfig(nodeConfig: NodeConfig) {
    val newNodeFile = NodeData(nodeConfig.nodeDir)
    synchronized(this) {
        nodeDataMap[nodeConfig.nodeDir] = newNodeFile
    }

    // ... etc
}
```
However, I now see that you have `private fun allPreviouslySeenFiles()`, which also accesses the map.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
chrisr3: I'm not sure I'm understanding you. My issue here is that having Thread-2 start copying a file from B -> C before Thread-1 has finished coping it from A -> B should not be allowed to happen in the first place.
And I don't see anything wrong with single copies either, if that single copy is reliable.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    private fun copy(source: Path, destination: Path) {
+        try {
+            // First copy the file to a temporary file within the appropriate directory.
+            val tempDestination = destination.parent / ("." + destination.fileName.toString())
+            Files.copy(source, tempDestination)
+            // Then rename it to the desired name. This way the file 'appears' on the filesystem as an atomic operation.
+            Files.move(tempDestination, destination)
```
chrisr3: You're going to need to using `StandardCopyOptions.REPLACE_EXISTING` here, or the `move` will throw a `FileAlreadyExistsException`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    // Takes a path under nodeData config dir and decides whether the file represented by that paths needs to
+    // be copied.
+    private fun processPath(nodeData: NodeData, path: Path) {
+        val newTimestamp = Files.readAttributes(path, BasicFileAttributes::class.java).lastModifiedTime()
+        val previousTimestamp = nodeData.previouslySeenFiles.put(path, newTimestamp)
+        if (previousTimestamp == null || newTimestamp > previousTimestamp) {
+            for (destination in nodeDataMap.values.map { it.destination }) {
```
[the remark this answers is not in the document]
al-r3: I'm filtering on the nodeDir being the different.
I don't really like the idea of NodeData being a data class, since it contains a mutable map.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,127 @@
...
+                val fullDestinationPath = destination.resolve(path.fileName)
+                copy(path, fullDestinationPath)
+            }
+        }
+    }
+
+    private fun copy(source: Path, destination: Path) {
+        try {
+            // First copy the file to a temporary file within the appropriate directory.
+            val tempDestination = destination.parent / ("." + destination.fileName.toString())
```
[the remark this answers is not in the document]
al-r3: There is a single temporary file, I'm deleting that if the move fails, if the initial copy fails there is hardly anything can do.
I'm re-throwing the exceptions as if any of these operation fails it means the nodes won't work as expected, and there is no handling which can really fix that.

I can't use createTempFile as it returns a java.io.File (old io) and I've been told not to use that in code.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,127 @@
...
+            Files.copy(source, tempDestination)
+            // Then rename it to the desired name. This way the file 'appears' on the filesystem as an atomic operation.
+            Files.move(tempDestination, destination)
+        } catch (exception: Exception) {
```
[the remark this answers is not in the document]
al-r3: (see reply above)
I'm logging something got wrong (you told me to do that in a previous round) and rethrowing the exception as it is not really recoverable.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,127 @@
...
+            logger.warn("Couldn't copy $source to $destination.", exception)
+        }
+    }
+
+    /**
+     * Convenience holder for all the paths and files relative to a single node.
+     */
+    private class NodeData(val nodeDir: Path) {
```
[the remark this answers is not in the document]
al-r3: It's not needed, I never have two of these I'll need to compare for equality (or hash), also it contains a mutableMap which still makes it a "suspicious" data class imho

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,126 @@
...
+            // First copy the file to a temporary file within the appropriate directory.
+            Files.copy(source, tempDestination, StandardCopyOption.COPY_ATTRIBUTES)
+        } catch (exception: Exception) {
+            log.log(Level.WARNING, "Couldn't copy $source to $tempDestination.", exception)
+            throw exception
+        }
+        try {
+            // Then rename it to the desired name. This way the file 'appears' on the filesystem as an atomic operation.
+            Files.move(tempDestination, destination, StandardCopyOption.REPLACE_EXISTING)
+        } catch (exception: Exception) {
```
chrisr3: `IOException` rather than `Exception`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+    }
+
+    private fun copy(source: Path, destination: Path) {
+        val tempDestination = destination.parent / ("." + destination.fileName.toString())
+        try {
+            // First copy the file to a temporary file within the appropriate directory.
+            Files.copy(source, tempDestination, StandardCopyOption.COPY_ATTRIBUTES)
+        } catch (exception: Exception) {
```
chrisr3: `IOException` rather than `Exception`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,126 @@
...
+            log.log(Level.WARNING, "Couldn't copy $source to $tempDestination.", exception)
+            throw exception
+        }
+        try {
+            // Then rename it to the desired name. This way the file 'appears' on the filesystem as an atomic operation.
+            Files.move(tempDestination, destination, StandardCopyOption.REPLACE_EXISTING)
```
chrisr3: Use `import java.nio.file.StandardCopyOption.*` so that you can just write `REPLACE_EXISTING`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,126 @@
...
+        if (newTimestamp > previousTimestamp) {
+            for (destination in nodeDataMap.values.filter { it.nodeDir != nodeData.nodeDir }.map { it.destination }) {
+                val fullDestinationPath = destination.resolve(path.fileName)
+                copy(path, fullDestinationPath)
+            }
+        }
+    }
+
+    private fun copy(source: Path, destination: Path) {
+        val tempDestination = destination.parent / ("." + destination.fileName.toString())
```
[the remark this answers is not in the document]
al-r3: but this is unique, I know the names of this files. They're all NodeInfo-<HASH> so if I add a prefix I know I'm not writing over another file.
Also the directory in which I'm writing these files is there exactly and just for these files, so there shouldn't be anything else in there anyway.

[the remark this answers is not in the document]
al-r3: By the way java.nio has
http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory%28java.nio.file.Path,%20java.lang.String,%20java.nio.file.attribute.FileAttribute...%29

which creates a whole directory for me, but still it doesn't seem I need that when I can safely generate one.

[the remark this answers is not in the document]
chrisr3: Actually, I see now that you could use:
```kotlin
val tempDestination = Files.createTempFile(destination.parent, ".nodeInfo", ".tmp")
```
and be done instead. I didn't know this function was there, and it's cleaner than the Kotlin one.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,132 @@
...
+        } catch (exception: IOException) {
+            log.log(Level.WARNING, "Couldn't create a temporary file to copy $source", exception)
+            throw exception
+        }
+        try {
+            // First copy the file to a temporary file within the appropriate directory.
+            Files.copy(source, tempDestination, COPY_ATTRIBUTES, REPLACE_EXISTING)
+        } catch (exception: IOException) {
+            log.log(Level.WARNING, "Couldn't copy $source to $tempDestination.", exception)
+            throw exception
```
chrisr3: You'll also need to delete the empty file created by `Files.createTempFile()` if this copy fails.

al-r3: done

PR #1800, opened 2017-10-04
CORDA-540: Add verification to ensure that private keys can only be serialized with specific contexts

vkolomeyko:
At the moment, even in Kryo mode, we allow unconstrained serialization of Private Keys.
This could be dangerous as the the key may leak to the external party.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/AllButBlacklisted.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/AllButBlacklisted.kt
@@ -30,6 +30,9 @@ import kotlin.collections.LinkedHashSet
  * we can often end up pulling in a lot of objects that do not make sense to put in a checkpoint.
  * Thus, by blacklisting classes/interfaces we don't expect to be serialised, we can better handle/monitor the aforementioned behaviour.
  * Inheritance works for blacklisted items, but one can specifically exclude classes from blacklisting as well.
+ * Note: Custom serializer registration trumps white/black lists. So if a given type has a custom serializer and has its name
+ * in the blacklist - it will still be serialized as specified by custom serializer.
+ * For more details, see [net.corda.nodeapi.internal.serialization.CordaClassResolver.getRegistration]
  */
 object AllButBlacklisted : ClassWhitelist {
 
--- a/node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/PrivateKeySerializationTest.kt
+++ b/node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/PrivateKeySerializationTest.kt
@@ -1,30 +1,21 @@
 package net.corda.nodeapi.internal.serialization
 
+import net.corda.core.crypto.Crypto
 import net.corda.core.serialization.SerializationContext.UseCase.*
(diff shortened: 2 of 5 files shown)


node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/PrivateKeySerializationTest.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.security.PrivateKey
+import java.security.SecureRandom
+import kotlin.test.assertTrue
+
+class PrivateKeySerializationTest : TestDependencyInjectionBase() {
+
+    private val privateKeys: List<PrivateKey>
+
+    init {
+        val generator = KeyPairGenerator()
```
kchalkias: better create keys  using `Crypto.generateKeyPair(Crypto.EDDSA_ED25519_SHA512)` as EdDSA is currently our default algorithm.
Also, let's test for the rest of key types (ECDSA K1 and R1, RSA and Sphincs) take a look at `CryptoUtilsTest` where all 5 algorithms are tested.

vkolomeyko: Sure - will do

kchalkias:
@rick-r3 We could also add `PrivateKey` to the blacklist of specific CONTEXTs, right?

vkolomeyko:
I have discussed with @rick-r3 and we agreed that I will try to achieve the same by using white/black lists.
Let me re-work the implementation in a separate branch and then we will be able to decide which one we like more.

vkolomeyko:
It doesn't seem to be easy to achieve the same effect with whitelist.
This is due to `net.corda.nodeapi.internal.serialization.CordaClassResolver#getRegistration` is defined as:
```
override fun getRegistration(type: Class<*>): Registration? {
        val targetType = typeForSerializationOf(type)
        return super.getRegistration(targetType) ?: checkClass(targetType)
    }
```
We do check in whitelist in `checkClass` method, but we check for custom serializer first. So if custom serializer is defined (like in case of `PrivateKey`) it is non-material if the class is in whitelist or not.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/AllButBlacklisted.kt
```diff
@@ -30,6 +30,9 @@ import kotlin.collections.LinkedHashSet
  * we can often end up pulling in a lot of objects that do not make sense to put in a checkpoint.
  * Thus, by blacklisting classes/interfaces we don't expect to be serialised, we can better handle/monitor the aforementioned behaviour.
  * Inheritance works for blacklisted items, but one can specifically exclude classes from blacklisting as well.
+ * Note: Custom serializer registration trumps white/black lists. So if a given type has a custom serializer and has its name
+ * in the blacklist - it will still be serialized as specified by custom serializer.
+ * For more details, see [net.corda.nodeapi.internal.serialization.CordaClassResolver.getRegistration]
```
kchalkias: could we have an {@link .... } here?

rick-r3: that's what this (the square brackets) does, no?

PR #1801, opened 2017-10-04
CORDA-644: Only serialise Kotlin lambdas when checkpointing.

chrisr3:
This local function is indistinguishable (at a JVM level) from the potentially problematic lambda parameter on the example flow in [CORDA-644](https://r3-cev.atlassian.net/browse/CORDA-644). I am removing it now to prevent this code from breaking when we switch from Kryo to AMQP.

### DIFF
--- a/client/rpc/src/integration-test/kotlin/net/corda/client/rpc/BlacklistKotlinClosureTest.kt
+++ b/client/rpc/src/integration-test/kotlin/net/corda/client/rpc/BlacklistKotlinClosureTest.kt
@@ -1,0 +1,92 @@
+package net.corda.client.rpc
+
+import co.paralleluniverse.fibers.Suspendable
+import com.esotericsoftware.kryo.KryoException
+import net.corda.core.flows.*
+import net.corda.core.identity.Party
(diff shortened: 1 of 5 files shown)


chrisr3:
Tricky. As I mentioned, Kotlin implemented the local function as a lambda class. The only thing I can think of would be to log a warning in `CordaClassResolver.checkClass()` whenever it whitelists a lambda type.

mikehearn:
You can annotate lambdas:

`foo.bar @Suspendable { ... }`

is that sufficient?

chrisr3:
As I understand it, we have _already_ denied that possibility by declaring `@CordaSerializable` as:
```kotlin
@Target(AnnotationTarget.CLASS)
@Retention(AnnotationRetention.RUNTIME)
@Inherited
annotation class CordaSerializable
```
Its `@Target` would need to include `AnnotationTarget.EXPRESSION` for it to be applicable to lambdas.

mikehearn:
Hmm - sending lambdas over P2P is weird, but if the lambda is annotated, I'd guess it's not a security hole. Anyway, no big deal either way.

Issue #1479, opened 2017-09-11
Provide a means for a @CordaService to access a CordaRPCOps to the local node


dazraf:
# Context

I'm writing  `@CordaService` that exposes a more advanced webservice. I'm doing this because I want to take advantage of [vertx's](http://vertx.io) performance and websockets; and further, I wish this webservice to be co-located with the node to simplify the deployment and lifecycle for the node. 

# Issue

I cannot create a RPC client easily because the RPC host and port is not exposed on `PluginServiceHub`. The best I can do is to either:

1. re-read the node's config - this solution doesn't work with the test driver.
2. provide a second, private, webserver config containing the RPC host and port  - which also doesn't solve for the test driver case.

# Suggestion

Provide some means that a `@CordaService` can acquire a `CordaRPCOps`.
This can be done in a number of ways. Some suggestions:

* Provide a builder on `PluginServiceHub` or `ServiceHub` to acquire a `CordaRPCOps`
* Provide a `rpcHostAndPort: NetworkHostAndPort` onf `PluginServiceHub`
* Provide access to the Corda node Config on `ServiceHub` or derived.

thanks and regards
Fuzz

mikehearn:
CCing Matthew.

Which RPCs do you want, exactly? As far as I know, the RPCs we have are all essentially just fronts to ServiceHub functionality.

mnesbit:
I also think everything in CordaRPCOps is just thin wrappers over ServiceHub calls at the minute (once the transport layer code and security checks are passed).

shamsasari:
I believe @dazraf whats the ability to start flows from Corda services. The `ServiceHub` that's provided doesn't expose the `startFlow` method, it's only available in `ServiceHubInternal`.

mikehearn:
Oh, odd. Why is startFlow not in ServiceHub?

mnesbit:
Because when we last had this discussion we regarded starting flows a key permission checking and audit capturing life cycle point. Therefore flows should only be able to start subFlows, or registered remote flows. That was precisely why startFlow was on PluginServiceHub as a privileged method. However, we now seem to have thinned that down to an empty extension of ServcieHub and then it got removed.

mikehearn:
Ah, I see, this is "if we put it on ServiceHub then it's available to services _and_ flows".

[...]

PR #1803, opened 2017-10-05
Start Flows from services

mnesbit:
This adds the AppServiceHub which allows CordaService classes to start flows marked with StartableByService in a many analogous to RPC flow support.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/StartableByService.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/StartableByService.kt
@@ -1,0 +1,12 @@
+package net.corda.core.flows
+
+import kotlin.annotation.AnnotationTarget.CLASS
+
+/**
+ * Any [FlowLogic] which is to be started by the [AppServiceHub] interface from
+ * within a [CordaService] must have this annotation. If it's missing the
+ * flow will not be allowed to start and an exception will be thrown.
+ */
+@Target(CLASS)
+@MustBeDocumented
+annotation class StartableByService
--- a/core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
@@ -1,0 +1,30 @@
+package net.corda.core.node
+
+import net.corda.core.flows.FlowLogic
+import net.corda.core.messaging.FlowHandle
+import net.corda.core.messaging.FlowProgressHandle
+import rx.Observable
+
+/**
+ * A [CordaService] annotated class requires a constructor taking a
+ * single parameter of type [AppServiceHub].
+ * With the [AppServiceHub] parameter a [CordaService] is able to access to privileged operations.
+ * In particular such a [CordaService] can initiate and track flows marked with [net.corda.core.flows.StartableByService].
+ */
+interface AppServiceHub : ServiceHub {
+
+    /**
+     * Start the given flow with the given arguments. [flow] must be annotated
+     * with [net.corda.core.flows.StartableByService].
+     * TODO it is assumed here that the flow object has an appropriate classloader.
+     */
+    fun <T> startFlow(flow: FlowLogic<T>): FlowHandle<T>
+
+    /**
+     * Start the given flow with the given arguments, returning an [Observable] with a single observation of the
+     * result of running the flow. [flow] must be annotated with [net.corda.core.flows.StartableByService].
+     * TODO it is assumed here that the flow object has an appropriate classloader.
+     */
+    fun <T> startTrackedFlow(flow: FlowLogic<T>): FlowProgressHandle<T>
+
+}
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -20,9 +20,8 @@ import net.corda.core.internal.concurrent.flatMap
 import net.corda.core.internal.concurrent.openFuture
 import net.corda.core.internal.toX509CertHolder
 import net.corda.core.internal.uncheckedCast
-import net.corda.core.messaging.CordaRPCOps
-import net.corda.core.messaging.RPCOps
-import net.corda.core.messaging.SingleMessageRecipient
+import net.corda.core.messaging.*
+import net.corda.core.node.AppServiceHub
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.ServiceHub
 import net.corda.core.node.services.*
@@ -265,6 +264,49 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
     }
 
     /**
+     * This customizes the ServiceHub for each CordaService that is initiating flows
+     */
+    private class AppServiceHubImpl<T : SerializeAsToken>(val serviceHub: ServiceHubInternal): AppServiceHub, ServiceHub by serviceHub {
+        lateinit var serviceInstance: T
+        override fun <T> startTrackedFlow(flow: FlowLogic<T>): FlowProgressHandle<T> {
+            val stateMachine = startFlowChecked(flow)
+            return FlowProgressHandleImpl(
+                    id = stateMachine.id,
+                    returnValue = stateMachine.resultFuture,
+                    progress = stateMachine.logic.track()?.updates ?: Observable.empty()
+            )
+        }
+
+        override fun <T> startFlow(flow: FlowLogic<T>): FlowHandle<T> {
+            val stateMachine = startFlowChecked(flow)
+            return FlowHandleImpl(id = stateMachine.id, returnValue = stateMachine.resultFuture)
+        }
+
+        private fun <T> startFlowChecked(flow: FlowLogic<T>): FlowStateMachineImpl<T> {
+            val logicType = flow.javaClass
+            require(logicType.isAnnotationPresent(StartableByService::class.java)) { "${logicType.name} was not designed for starting by a CordaService" }
+            val currentUser = FlowInitiator.Service(serviceInstance.javaClass.name)
+            return serviceHub.startFlow(flow, currentUser)
+        }
+
+        override fun equals(other: Any?): Boolean {
+            if (this === other) return true
+            if (other !is AppServiceHubImpl<*>) return false
+
+            if (serviceHub != other.serviceHub) return false
+            if (serviceInstance != other.serviceInstance) return false
+
+            return true
+        }
+
+        override fun hashCode(): Int {
+            var result = serviceHub.hashCode()
+            result = 31 * result + serviceInstance.hashCode()
+            return result
+        }
+    }
+
+    /**
      * Use this method to install your Corda services in your tests. This is automatically done by the node when it
      * starts up for all classes it finds which are annotated with [CordaService].
      */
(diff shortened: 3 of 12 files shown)


mnesbit:
This tackles https://github.com/corda/corda/issues/1479

core/src/main/kotlin/net/corda/core/flows/StartableByService.kt
```diff
@@ -0,0 +1,12 @@
+package net.corda.core.flows
+
+import kotlin.annotation.AnnotationTarget.CLASS
+
+/**
+ * Any [FlowLogic] which is to be started by the AppServiceHub interface from
```
shamsasari: [AppServiceHub]

core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
```diff
@@ -0,0 +1,25 @@
+package net.corda.core.node
+
+import net.corda.core.flows.FlowLogic
+import net.corda.core.messaging.FlowHandle
+import net.corda.core.messaging.FlowProgressHandle
+import rx.Observable
+
+/**
+ * A [CordaService] annotated class requires
```
shamsasari: Looks incomplete

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -265,6 +264,48 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
...
+            )
+        }
+
+        override fun <T> startFlowDynamic(logicType: Class<out FlowLogic<T>>, vararg args: Any?): FlowHandle<T> {
+            val stateMachine = startFlow(logicType, args)
+            return FlowHandleImpl(id = stateMachine.id, returnValue = stateMachine.resultFuture)
+        }
+
+        private fun <T> startFlow(logicType: Class<out FlowLogic<T>>, args: Array<out Any?>): FlowStateMachineImpl<T> {
+            require(logicType.isAnnotationPresent(StartableByService::class.java)) { "${logicType.name} was not designed for RPC" }
```
shamsasari: Not RPC

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -265,6 +264,48 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
...
+            val currentUser = FlowInitiator.Service(serviceInstance.javaClass.name)
+            return serviceHub.invokeFlowAsync(logicType, currentUser, *args)
+        }
+
+        override fun equals(other: Any?): Boolean {
```
shamsasari: Do we need equals and hashCode overridden? The default will suffice as there's only one instance of this class per `serviceInstance`.

node/src/test/kotlin/net/corda/node/internal/cordapp/CordaServiceTest.kt
```diff
@@ -0,0 +1,143 @@
...
+        return stateMachine.flowInitiator
+    }
+}
+
+@CordaService
+class TestCordaService(val appServiceHub: AppServiceHub): SingletonSerializeAsToken() {
+    fun startServiceFlow() {
+        val handle = appServiceHub.startFlowDynamic(DummyServiceFlow::class.java)
+        val initiator = handle.returnValue.get()
+        assert(initiator is FlowInitiator.Service)
```
shamsasari: The cast below will throw ClassCastException if not the correct type so this isn't needed. Plus this isn't the testing assert.

node/src/test/kotlin/net/corda/node/internal/cordapp/CordaServiceTest.kt
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    fun startServiceFlowAndTrack() {
+        val handle = appServiceHub.startTrackedFlowDynamic(DummyServiceFlow::class.java)
+        val count = AtomicInteger(0)
+        val subscriber = handle.progress.subscribe { count.incrementAndGet() }
+        handle.returnValue.get()
+        // Simply prove some progress was made.
+        // The actual number is currently 11, but don't want to hard code an implementation detail.
+        assert(count.get() > 1)
```
shamsasari: Use Junit asserts

node/src/test/kotlin/net/corda/node/internal/cordapp/CordaServiceTest.kt
```diff
@@ -0,0 +1,143 @@
...
+    @Test
+    fun `Can start StartableByService flows`() {
+        val service = nodeA.services.cordaService(TestCordaService::class.java)
+        service.startServiceFlow()
+    }
+
+    @Test
+    fun `Can't start StartableByRPC flows`() {
+        val service = nodeA.services.cordaService(TestCordaService2::class.java)
+        assertFails { service.startInvalidRPCFlow() }
```
shamsasari: This will be prone to false-positives if the exception thrown is not related to what we're testing. We need assertions on the expected exception.

node/src/test/kotlin/net/corda/node/internal/cordapp/CordaServiceTest.kt
```diff
@@ -0,0 +1,143 @@
+package net.corda.node.internal.cordapp
```
shamsasari: Perhaps this should be in `net.corda.node.internal`

core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+interface AppServiceHub : ServiceHub {
+
+    /**
+     * Start the given flow with the given arguments. [logicType] must be annotated
+     * with [net.corda.core.flows.StartableByService].
+     */
+    fun <T> startFlowDynamic(logicType: Class<out FlowLogic<T>>, vararg args: Any?): FlowHandle<T>
```
shamsasari: We don't need to serialise this over the wire like in RPC so passing in a `FlowLogic` instance is probably what we want

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
mnesbit: I ended up not making this SerializeAsToken as it just contains two such objects already. In my unit tests I got surprising results without a valid equals method.

shamsasari: Hang on, may be it needs to be SerializeAsToken anyway. @rick-r3 ?

core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
mnesbit: Just been playing with this. I actually prefer the Class, param1, param2 style to use. Having to construct my flow and then run it just seems clunky.

mnesbit: It has been pointed out none of these signatures lets us get control of the classloaders used to instantiate flows.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
rick-r3: Not really needed, since the two members already are.

core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
shamsasari: Yes I was thinking about that. Does that mean we use Strings?!

mnesbit: Perhaps as a trusted component the classloader of the service will already be appropriate. Really a question of whether I go with a version like the RPC functions, or one like the raw ServiceHubInternal startFlow?

core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
```diff
@@ -0,0 +1,28 @@
+package net.corda.core.node
+
+import net.corda.core.flows.FlowLogic
+import net.corda.core.messaging.FlowHandle
+import net.corda.core.messaging.FlowProgressHandle
+import rx.Observable
+
+/**
+ * A [CordaService] annotated class requires a constructor taking a
+ * single parameter of type [AppServiceHub], or for legacy support of type [ServiceHub].
```
shamsasari: Perhaps not document the legacy support.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -276,8 +318,15 @@ abstract class AbstractNode(open val configuration: NodeConfiguration,
...
             } else {
-                val constructor = serviceClass.getDeclaredConstructor(ServiceHub::class.java).apply { isAccessible = true }
-                constructor.newInstance(services)
+                try {
+                    val extendedServiceConstructor = serviceClass.getDeclaredConstructor(AppServiceHub::class.java).apply { isAccessible = true }
+                    val serviceContext = AppServiceHubImpl<T>(services)
+                    serviceContext.serviceInstance = extendedServiceConstructor.newInstance(serviceContext)
+                    serviceContext.serviceInstance
+                } catch (ex: NoSuchMethodException) {
+                    val constructor = serviceClass.getDeclaredConstructor(ServiceHub::class.java).apply { isAccessible = true }
```
shamsasari: Can we add a info or warn level log here so that at least in tests the devs know to use the new API.

core/src/main/kotlin/net/corda/core/node/AppServiceHub.kt
```diff
@@ -0,0 +1,28 @@
...
+ * With the [AppServiceHub] parameter a [CordaService] is able to access to privileged operations.
+ * In particular such a [CordaService] can initiate and track flows marked with [net.corda.core.flows.StartableByService].
+ */
+interface AppServiceHub : ServiceHub {
+
+    /**
+     * Start the given flow with the given arguments. [flow] must be annotated
+     * with [net.corda.core.flows.StartableByService].
+     */
+    fun <T> startFlow(flow: FlowLogic<T>): FlowHandle<T>
```
shamsasari: Can you add a TODO here regarding classloaders and the impact it may have on the API

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -207,14 +209,18 @@ class CordappLoader private constructor(private val cordappJarPaths: List<URL>)
...
+    private fun Class<out FlowLogic<*>>.isUserInvokable(): Boolean {
+        return Modifier.isPublic(modifiers) && !isLocalClass && !isAnonymousClass && (!isMemberClass || Modifier.isStatic(modifiers))
+    }
 
+    private fun findRPCFlows(scanResult: ScanResult): List<Class<out FlowLogic<*>>> {
         return scanResult.getClassesWithAnnotation(FlowLogic::class, StartableByRPC::class).filter { it.isUserInvokable() }
     }
 
+    private fun findServiceFlows(scanResult: ScanResult): List<Class<out FlowLogic<*>>> {
+        return scanResult.getClassesWithAnnotation(FlowLogic::class, StartableByService::class).filter { it.isUserInvokable() }
```
shamsasari: Use of `isUserInvokable` will prevent users from starting non-public flows (perhaps they want their flow to be `private`). In fact the filter is no longer needed now that the start method takes in a `FlowLogic` instance.

PR #1812, opened 2017-10-05
Change IdentitySyncFlow to only offer own identities

rnicoll:
Change IdentitySyncFlow to only offer confidential identities matching the identity the flow is run as. This avoids risks of nodes being convinced to include a state with a confidential identity in it, by a remote node, then feeding the well known identity to the node during identity sync.


rnicoll:
1.1 equivalent to #1740

PR #1814, opened 2017-10-05
CORDA-540: Change how standard mutable whitelists structured

vkolomeyko:


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
```diff
@@ -170,8 +170,8 @@ class GlobalTransientClassWhiteList(val delegate: ClassWhitelist) : MutableClass
 /**
  * A whitelist that can be customised via the [net.corda.core.node.SerializationWhitelist], since implements [MutableClassWhitelist].
  */
-class TransientClassWhiteList(val delegate: ClassWhitelist) : MutableClassWhitelist, ClassWhitelist by delegate {
-    val whitelist: MutableSet<String> = Collections.synchronizedSet(mutableSetOf())
+class TransientClassWhiteList(private val delegate: ClassWhitelist) : MutableClassWhitelist {
+    private val whitelist: MutableSet<String> = Collections.synchronizedSet(mutableSetOf())
 
     override fun hasListed(type: Class<*>): Boolean {
         return (type.name in whitelist) || delegate.hasListed(type)
```
rick-r3: I was thinking yesterday.... should we re-order this OR ?

Rationale: If we wanted to prevent the whitelisting by apps of certain types, with a blacklist, we need to check our delegate `ClassWhitelist` implementation first before the one they are allowed to add to.  Our blacklist implementation throws an exception rather than returning, so this would then work.

What do you think?  If we change it, can we write a test to make sure we don't re-order again later, and add a comment here that the order matters?

kchalkias: Very interesting trick, also agree.

vkolomeyko: Sure will make a change.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
```diff
@@ -152,36 +152,33 @@ object AllWhitelist : ClassWhitelist {
...
-        val whitelist: MutableSet<String> = Collections.synchronizedSet(mutableSetOf())
-    }
+sealed class AbstractMutableClassWhitelist(private val whitelist: MutableSet<String>,  private val delegate: ClassWhitelist) : MutableClassWhitelist {
 
     override fun hasListed(type: Class<*>): Boolean {
-        return (type.name in whitelist) || delegate.hasListed(type)
+        /**
+         * There are certain delegates like [net.corda.nodeapi.internal.serialization.AllButBlacklisted]
+         * which may throw when asked whether the type is listed.
+         * In such situations - it may be a good idea to ask [delegate] first before making a check against own [whitelist]
```
kchalkias: full stop

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
```diff
@@ -152,36 +152,33 @@ object AllWhitelist : ClassWhitelist {
...
-        whitelist += entry.name
+// TODO: Need some concept of from which class loader
+class GlobalTransientClassWhiteList(delegate: ClassWhitelist) : AbstractMutableClassWhitelist(GlobalTransientClassWhiteList.whitelist, delegate) {
+    companion object {
+        private val whitelist: MutableSet<String> = Collections.synchronizedSet(mutableSetOf())
     }
 }
 
+/**
+ * A whitelist that can be customised via the [net.corda.core.node.SerializationWhitelist], since implements [MutableClassWhitelist].
```
kchalkias: since _it_ implements

vkolomeyko:
@kchalkias and @rick-r3 - Requested changes been made - over to you

PR #1815, opened 2017-10-05
Break API!

chrisr3:
Break the RPC API to test the API Scanner gate in V1.


shamsasari:
Adding is not an breaking API change. You'll need to remove or change.

chrisr3:
@shamsasari Yes, that's right. Which was why my first test was to change the `notUsed()` RPC function to `neverUsed()`. When this broke (correctly) the API gate, I reverted that change and then added this `world()` function instead to ensure that the gate now passed.

Trust us, @Clintonio and I had this covered.

shamsasari:
Yes, I figured later that's probably what you had done! 👍

PR #1818, opened 2017-10-05
CORDA-654: Handle non-standard file systems in NodeInfoWatcher

rnicoll:
Handle non-standard file systems such as JimFs, in NodeInfoWatcher. Instead of using toFile() to convert a Path to a File, open the Path for writing to directly.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
+++ b/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
@@ -49,9 +49,7 @@ class NodeInfoWatcher(private val nodePath: Path,
                 val serializedBytes = nodeInfo.serialize()
                 val regSig = keyManager.sign(serializedBytes.bytes, nodeInfo.legalIdentities.first().owningKey)
                 val signedData = SignedData(serializedBytes, regSig)
-                Files.newOutputStream(path / ("nodeInfo-" + serializedBytes.hash.toString())).use { out ->
-                    out.write(signedData.serialize().bytes)
-                }
+                signedData.serialize().open().copyTo(path / "nodeInfo-${serializedBytes.hash}")
             } catch (e: Exception) {
                 logger.warn("Couldn't write node info to file", e)
             }
--- a/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
+++ b/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
@@ -1,14 +1,8 @@
 package net.corda.node.services.network
 
 import net.corda.cordform.CordformNode
-import net.corda.core.crypto.SecureHash
 import net.corda.core.crypto.SignedData
-import net.corda.core.internal.createDirectories
-import net.corda.core.internal.div
-import net.corda.core.internal.isDirectory
-import net.corda.core.internal.isRegularFile
-import net.corda.core.internal.list
-import net.corda.core.internal.readAll
+import net.corda.core.internal.*
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.services.KeyManagementService
 import net.corda.core.serialization.deserialize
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -52,11 +48,11 @@ class NodeInfoWatcher(private val nodePath: Path,
             try {
                 path.createDirectories()
                 val serializedBytes = nodeInfo.serialize()
-                val regSig = keyManager.sign(serializedBytes.bytes,
-                        nodeInfo.legalIdentities.first().owningKey)
+                val regSig = keyManager.sign(serializedBytes.bytes, nodeInfo.legalIdentities.first().owningKey)
                 val signedData = SignedData(serializedBytes, regSig)
-                val file = (path / ("nodeInfo-" + SecureHash.sha256(serializedBytes.bytes).toString())).toFile()
-                file.writeBytes(signedData.serialize().bytes)
+                Files.newOutputStream(path / ("nodeInfo-" + serializedBytes.hash.toString())).use { out ->
```
shamsasari: I was going to fix this in my PR but didn't bother. You can improve further:
```
val file = path / ("nodeInfo-" + SecureHash.sha256(serializedBytes.bytes).toString()))  // You could even convert this to a string template to remove the toString() call
signedData.serialize().open().copyTo(file)
```

mikehearn:
I'm curious what happens if you try to convert a JimFS Path to a File()? File() is backed by the nio implementation but I can imagine that a JimFS path can't be expressed in the File() structures.

shamsasari:
That's correct. It throws an exception, can't remember which one.

shamsasari:
For some reason my PR in question passes on TC but not on my laptop, so that's something I may have to look into.

rnicoll:
Looks like the `polling folder` test has a race condition, investigating now

rnicoll:
It turns out that WatchService on OS X is incredibly slow because instead of using native functionality of the OS, it manually polls every now and then (see https://stackoverflow.com/questions/9588737/is-java-7-watchservice-slow-for-anyone-else ). As a result the test is almost universally going to fail on OS X, and isn't terribly reliable on other operating systems. @al-r3 is re-writing the implementation at the moment, and we'll then get back to these more minor changes

mikehearn:
There's a way to speed up polling on OS X such that it's fairly fast. See here:

https://github.com/vinumeris/lighthouse/blob/ba19625cfe88a1a76095807ee64b91fff094b809/common/src/main/java/lighthouse/files/DirectoryWatcher.kt

look at SensitivityWatchModifier.HIGH

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -12,11 +12,7 @@ import net.corda.core.utilities.loggerFor
 import rx.Observable
 import rx.Scheduler
 import rx.schedulers.Schedulers
-import java.nio.file.Path
-import java.nio.file.StandardWatchEventKinds
-import java.nio.file.WatchEvent
-import java.nio.file.WatchKey
-import java.nio.file.WatchService
+import java.nio.file.*
```
al-r3: Do we have any conventions here on import *?
I was told not to use them in a previous PR of mine.

al-r3:
I think it's easier to just list the folder content instead of polling at this time, it's still a single implementation and should just work everywhere.
It will be less performant but we'll have a single implementation for all filesystems

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
shamsasari: The only time I've ever commented on imports if they're the only change in a file and reverting can reduce the number of files changes, useful in a large PR. Otherwise the import statements are noise and shouldn't be a point of issue.

al-r3:
fyi PR to remove WatchService entirely: https://github.com/corda/corda/pull/1836

PR #1819, opened 2017-10-05
CORDA-654: Continuation of moving to chooseIdentity() with name

rnicoll:


node/src/test/kotlin/net/corda/node/services/statemachine/FlowFrameworkTests.kt
```diff
@@ -67,39 +67,46 @@ class FlowFrameworkTests {
...
 
         // We don't create a network map, so manually handle registrations
         mockNet.registerIdentities()
-        notary1Identity = notary1.services.myInfo.legalIdentities[1]
-        notary2Identity = notary2.services.myInfo.legalIdentities[1]
+
+        // Extract identities
+        alice = notary1.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
+        bob = notary1.services.networkMapCache.getPeerByLegalName(BOB_NAME)!!
+        notary1Identity = notary1.services.myInfo.legalIdentities.single { it.name == notaryName }
```
kasiastreich: You should use `NetworkMapCache` lookup on `notaryIdentities`.

node/src/test/kotlin/net/corda/node/services/statemachine/FlowFrameworkTests.kt
```diff
@@ -111,24 +118,24 @@ class FlowFrameworkTests {
 
     @Test
     fun `newly added flow is preserved on restart`() {
-        node1.services.startFlow(NoOpFlow(nonTerminating = true))
-        node1.internals.acceptableLiveFiberCountOnStop = 1
-        val restoredFlow = node1.restartAndGetRestoredFlow<NoOpFlow>()
+        aliceNode.services.startFlow(NoOpFlow(nonTerminating = true))
```
kasiastreich: Like in base PR, extract `aliceServices` to get rid of `StartedNode`?

kasiastreich: Ah my bad, in these tests it doesn't make sense.

node/src/test/kotlin/net/corda/node/services/transactions/NotaryServiceTests.kt
```diff
@@ -151,16 +156,16 @@ class NotaryServiceTests {
...
         mockNet.runNetwork()
         return future
     }
 
-    fun issueState(node: StartedNode<*>): StateAndRef<*> {
-        val tx = DummyContract.generateInitial(Random().nextInt(), notary, node.info.chooseIdentity().ref(0))
-        val signedByNode = node.services.signInitialTransaction(tx)
-        val stx = notaryNode.services.addSignature(signedByNode, notary.owningKey)
-        node.services.recordTransactions(stx)
+    fun issueState(services: ServiceHub, identity: Party): StateAndRef<*> {
```
kasiastreich: Are `services` form different node than `identity`? Because it would imply something like that, but you generate contract with you as an owner and sign it with your keys.

rnicoll: They're the same, it's just there so you it knows which identity of the node to use.

rnicoll:
Unfortunately passing in an identity involves a bunch of changes to how node identity is loaded, although I suppose in theory we could save then load them in some manner. It would also worsen problems we've seen already where the fixed keys used in `ALICE` & co. are sometimes incorrect (created from the same entropy values in multiple places), which we're trying to move away from.

I've added a `chooseIdentity()` function that takes in an identity name and throws an exception if not matched, which should be a neater solution. That's in #1744 the precursor PR to this one.

confidential-identities/src/test/kotlin/net/corda/confidential/IdentitySyncFlowTests.kt
```diff
@@ -40,12 +40,12 @@ class IdentitySyncFlowTests {
...
-        mockNet.createNotaryNode()
-        val aliceNode = mockNet.createPartyNode(ALICE.name)
-        val bobNode = mockNet.createPartyNode(BOB.name)
-        val alice: Party = aliceNode.services.myInfo.chooseIdentity()
-        val bob: Party = bobNode.services.myInfo.chooseIdentity()
-        val notary = aliceNode.services.getDefaultNotary()
+        val notaryNode = mockNet.createNotaryNode()
+        val aliceNode = mockNet.createPartyNode(ALICE_NAME)
+        val bobNode = mockNet.createPartyNode(BOB_NAME)
+        val alice: Party = notaryNode.services.networkMapCache.getPeerByLegalName(ALICE_NAME)!!
```
shamsasari: Why not have a `singleIdentity()` method on `MockNode` which throws an exception if there's more than one identity. `getPeerByLegalName` hasn't improved things over `chooseIdentity` - it's more boilerplate.

shamsasari:
You say in your JIRA ticket:

"Change mock node startup to return the initial identity's `Party` instance. This then allows elimination of `chooseIdentity()` and the treatment of the first identity as special, as part of progress towards full multiple identity support in both production and mock nodes."

Is this what you're suggesting:
```
val (node, identity) = mockNet.startNode()
```

That would seem better than what we have here. I don't think this PR is substantive enough.

mikehearn:
At this point if we want to make major changes to the unit testing API, please split out the parts that you want to change into a separate module, clone it, and then make the changes in the new module. Whilst the test APIs weren't frozen in 1.0 that doesn't mean we should just gratuitously break them all either because we found a slightly better way to do things. At this point I'd like to see only small tweaks to the test APIs - a total restructure would deserve a new module that we maintain alongside the old one (but probably not worth it).

shamsasari:
@mikehearn I don't think the test APIs are in a state where they just deserve small tweaks. We'll be releasing a poor API otherwise IMO. For example, `test-utils` depends on `node` and so all the node code becomes public. `AbstractNode` has not be written with the mind that it's accessible from testing. There's stuff visible in there that shouldn't be. 

Further `test-utils` has been serving two purposes thus far: as a testing module for CorDapp developers and as a testing framework for us to test Corda itself. These two concerns have to be separated out. The former shouldn't be depending on `node` and later should be, but as an internal API.

mikehearn:
Sure, but we can tolerate a split/merge of modules that we haven't stabilised yet. My point is that changes which require people to do major rewrites of their unit tests need to be considered carefully. Just hiding stuff that isn't useful for app developer tests or clarifying module boundaries is unlikely to break things in any major way.

shamsasari:
I'd like someone to go through the test modules and make sure what we have makes sense and is thought out, because the level of care and concern that we've given to `core` as a public API dwarfs what we've done here.

As an example, there are three ways to start a node, NodeBasedTest, the Driver (which has in-process and out-process modes) and the MockNetwork. NodeBasedTest is easy - we should get rid of it since in-process driver already does the same thing, or perhaps it's the other way around and we get rid of in-process driver? 

As long as we have someone do that due diligence I'm OK.

kasiastreich:
I agree with @shamsasari, test framework becomes overcomplicated and not well thought. For NetworkParameters work I realised that we have 3 ways of doing the same thing (for example million ways of starting normal nodes, starting notary clusters), which is unnecessary code replication and should be unified to the same behaviour (bft is only implemented for MockNet, raft for driver). Probably NodeBasedTest should go too.

mikehearn:
Yep, I know. We can always fork test-utils and produce a corda-app-test-framework that's cleaner, better designed etc and let people migrate at their own pace however. Perhaps one day the old module can be deprecated and eventually removed if nobody uses it anymore. 

Realistically though, we need to stabilise the test framework soon, and without more major dislocations, even if they are justifiable from a design perspective. We don't have to keep extending the bad APIs but we do need to avoid just ripping them out entirely. I'll send a note about this soon.

confidential-identities/src/test/kotlin/net/corda/confidential/IdentitySyncFlowTests.kt
rnicoll: So the idea is to give an example of how it's used in production, rather than using test case special code, in the theory that developers are likely to try use our tests as a very large library of worked samples.

rnicoll:
Keep in mind this doesn't even remove any functions yet, this is just moving from a test special API, to the standard production API.

PR #1820, opened 2017-10-06
Removes useless @JVMOverloads.

joeldudleyr3:
Removes a useless @JVMOverloads annotation on a method with no default arguments.


chrisr3:
This will be unblocked by #1829.

PR #1821, opened 2017-10-06
KDoc Tidyup -> @param -> @parameter in AMQP lib

fenryka:
Just updating some class documentation


andr3ej:
i don't get it, it's documented as @ param on https://kotlinlang.org/docs/reference/kotlin-doc.html

also in TC api check,
> Task 'generateApi' not found in root project 'corda-project'.

are you up to date with latest master?

PR #1822, opened 2017-10-06
Better error for driver instantiated outside package.

joeldudleyr3:
Throws helpful error message when driver is instantiated by a function outside a package.


testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -990,7 +990,8 @@ class DriverDSL(
             return Exception()
                     .stackTrace
                     .first { it.fileName != "Driver.kt" }
-                    .let { Class.forName(it.className).`package`.name }
+                    .let { Class.forName(it.className).`package`?.name }
+                    ?: throw IllegalArgumentException("Function instantiating driver must be defined in a package.")
```
exFalso: Perhaps IllegalState is more appropriate

PR #1824, opened 2017-10-06
[CORDA-682]: Claim Jigsaw module names

sollecitom:
It is now possible to require Corda artifacts from Java 9 Jigsaw modules.

An example: https://github.com/sollecitom/corda-jigsaw-example


docs/source/corda-modules.rst
```diff
@@ -0,0 +1,39 @@
+Corda modules
+=============
+
+Corda artifacts can be required from Java 9 Jigsaw modules.
+From within a ``module-info.java``, you can reference one of the modules e.g., ``requires net.corda.core;``.
+
+.. warning:: while Corda artifacts can be required from ``module-info.java`` files, they are still not proper Jigsaw modules.
```
mikehearn: still not proper Jigsaw modules .... because they rely on the automatic module mechanism and declare no module descriptors themselves. We plan to integrate Jigsaw more thoroughly in future.

docs/source/corda-modules.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Corda comprises the following modules:
```
mikehearn: Hmm. It's a great list, but we should be super clear which are part of the public API and which are internal. There's such a list already in the api-index.rst section, perhaps the discussion of versions and stability there can be moved to the new page, and we can split this list into public vs private so developers know what they can't depend on.

docs/source/corda-modules.rst
```diff
@@ -4,19 +4,38 @@ Corda modules
 Corda artifacts can be required from Java 9 Jigsaw modules.
 From within a ``module-info.java``, you can reference one of the modules e.g., ``requires net.corda.core;``.
 
-.. warning:: while Corda artifacts can be required from ``module-info.java`` files, they are still not proper Jigsaw modules.
+.. warning:: while Corda artifacts can be required from ``module-info.java`` files, they are still not proper Jigsaw modules,
+because they rely on the automatic module mechanism and declare no module descriptors themselves. We plan to integrate Jigsaw more thoroughly in the future.
```
mikehearn: Does this format properly? Have you tried a docsite build and checked? I thought the second line had to be indented too.

docs/source/corda-modules.rst
```diff
@@ -4,19 +4,38 @@ Corda modules
...
+
+* **net.corda.node.driver**: test utilities to run nodes programmatically
+* **net.corda.confidential.identities**: experimental support for confidential identities on the ledger
+* **net.corda.node.test.utils**: generic test utilities
 * **net.corda.client.jfx**: support for Java FX UI
 * **net.corda.client.mock**: client mock utilities
-* **net.corda.client.rpc**: client RPC
-* **net.corda.confidential.identities**: experimental support for confidential identities on the ledger
-* **net.corda.core**: core Corda libraries such as crypto functions, types for Corda's building blocks: states, contracts, transactions, attachments, etc. and some interfaces for nodes and protocols
 * **net.corda.sandbox-creator**: sandbox utilities
```
mikehearn: I'm not sure about these latter too, I don't know what they actually even do. Let's move them to the third list.

docs/source/corda-modules.rst
```diff
@@ -4,19 +4,38 @@ Corda modules
...
 * **net.corda.quasar.hook**: agent to hook into Quasar and provide types exclusion lists
+
+Corda unstable modules
+----------------------
+
+The following modules are available but we do not commit to their stability or continuation in any sense:
+
+* **net.corda.buildSrc**: necessary gradle plugins to build Corda
 * **net.corda.finance**: a range of elementary contracts (and associated schemas) and protocols, such as abstract fungible assets, cash, obligation and commercial paper
```
mikehearn: Let's move this (aspirationally) to the second list, as that aligns with our 1.0 messaging.

docs/source/corda-modules.rst
sollecitom: Fixed, my bad.

docs/source/corda-modules.rst
sollecitom: Fxied

docs/source/corda-modules.rst
sollecitom: Moved.

PR #1826, opened 2017-10-06
Removes broken link.

joeldudleyr3:
Removes a broken API link on the flows testing page.


shamsasari:
Why not just point to the correct link?

joeldudleyr3:
There's no correct link until we put the testing API docs back in after stabilisation (discussed with Mike).

PR #1827, opened 2017-10-06
[CORDA-442] Removed the NetworkMap option from Cordform, changed all the examples…

al-r3:
Removed the option for Cordformation to specify a networkMap.

All the samples have been migrated to not specify a networkMap.

PRs to update the demo apps will follow shortly.

### DIFF
--- a/docs/source/deploying-a-node.rst
+++ b/docs/source/deploying-a-node.rst
@@ -11,13 +11,12 @@ Cordform is the local node deployment system for CorDapps. The nodes generated a
 debugging, and testing node configurations, but not for production or testnet deployment.
 
 Here is an example Gradle task called ``deployNodes`` that uses the Cordform plugin to deploy three nodes, plus a
-notary/network map node:
+notary node:
 
 .. sourcecode:: groovy
 
     task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
         directory "./build/nodes"
-        networkMap "O=Controller,OU=corda,L=London,C=UK"
         node {
             name "O=Controller,OU=corda,L=London,C=UK"
             notary = [validating : true]
@@ -56,9 +55,7 @@ notary/network map node:
         }
     }
 
-You can extend ``deployNodes`` to generate any number of nodes you like. The only requirement is that you must specify
-one node as running the network map service, by putting their name in the ``networkMap`` field. In our example, the
-``Controller`` is set as the network map service.
+You can extend ``deployNodes`` to generate any number of nodes you like.
 
 .. warning:: When adding nodes, make sure that there are no port clashes!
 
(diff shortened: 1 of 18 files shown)


docs/source/deploying-a-node.rst
```diff
@@ -55,9 +54,7 @@ notary/network map node:
         }
     }
 
-You can extend ``deployNodes`` to generate any number of nodes you like. The only requirement is that you must specify
-one node as running the network map service, by putting their name in the ``networkMap`` field. In our example, the
-``Controller`` is set as the network map service.
+You can extend ``deployNodes`` to generate any number of nodes you like.
```
shamsasari: I doesn't make sense that the changelog has more detail on the gradle plugin with regards to node info files than this doc does. Can you update this to say that the plugin copies the node information files to each node so that the deployed nodes form their own private network, etc.

al-r3: I've added it, tough I'm not sure I fully get the purpose of this file at this point.

Is this meant to be a complete guide to how Cordformation works internally? Or just a tutorial on how to use it?

I'm saying this because all those information on nodeInfos seem implementation details and they're not much actionable by a user.

al-r3:
I went through all the mentions of a NetworkMap node I could find and I killed them.

docs/source/corda-configuration-file.rst
```diff
@@ -106,13 +106,6 @@ path to the node's base directory.
     members must be active and be able to communicate with the cluster leader for joining. If empty, a new cluster will be
     bootstrapped. Required only when running a distributed notary service.
 
-:networkMapService: If `null`, or missing the node is declaring itself as the NetworkMapService host. Otherwise this is
```
shamsasari: This is a bit premature. We haven't removed network map node concept yet.

docs/source/deploying-a-node.rst
shamsasari: I don't think the files are implementation detail.

docs/source/corda-configuration-file.rst
al-r3: reverted

docs/source/setting-up-a-corda-network.rst
```diff
@@ -48,13 +48,12 @@ The most important fields regarding network configuration are:
   resolvable name of a machine in a VPN.
 * ``rpcAddress``: The address to which Artemis will bind for RPC calls.
 * ``webAddress``: The address the webserver should bind. Note that the port must be distinct from that of ``p2pAddress`` and ``rpcAddress`` if they are on the same machine.
-* ``networkMapService``: Details of the node running the network map service. If it's this node that's running the service
-  then this field must not be specified.
+
```
shamsasari: d

shamsasari: There's more needing reverting

al-r3:
@shamsasari yes I've manually tried out each demo

al-r3:
I've bumped the version, I think it was un-bumped by a rebase

shamsasari:
You will need to update the gradle plugin version again, as my PR has already updated master to the same one you're using.

PR #1830, opened 2017-10-06
Clean up NodeInfoWatcher tests

rnicoll:


node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -36,6 +36,8 @@ class NodeInfoWatcher(private val nodePath: Path,
     private val watchService : WatchService? by lazy { initWatch() }
 
     companion object {
+        const val POLL_INTERVAL = 5L
```
al-r3: I'd just have a single parameter named
POLL_INTERVAL_MSEC 
so that it's meaning and content are immediately clear

rnicoll:
Ready for re-review

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -21,20 +21,25 @@ import kotlin.streams.toList
...
 class NodeInfoWatcher(private val nodePath: Path,
                       pollFrequencyMsec: Long = 5.seconds.toMillis(),
                       private val scheduler: Scheduler = Schedulers.io()) {
 
     private val nodeInfoDirectory = nodePath / CordformNode.NODE_INFO_DIRECTORY
-    private val pollFrequencyMsec: Long
+    /**
+     * How often to poll the filesystem, in milliseconds.
+     */
+    val pollFrequencyMsec: Long
```
al-r3: You don't really need to make this visible.
You can simply set the value to that you want in the test (the default just works)

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -61,7 +66,7 @@ class NodeInfoWatcher(private val nodePath: Path,
     }
 
     init {
-        this.pollFrequencyMsec = maxOf(pollFrequencyMsec, 5.seconds.toMillis())
+        this.pollFrequencyMsec = maxOf(pollFrequencyMsec, POLL_INTERVAL_UNIT.toMillis(MIN_POLL_INTERVAL))
```
al-r3: you don't really need those constants.
They're not used externally, so they should be private.
Also their meaning is clear from the context they're used in.

And if you really want them, have a single value with a clear meaning, either a Duration or something with Msec or seconds in their name MIN_POLL_INTERVAL is quite obscure in meaning to me

rnicoll:
At this point as I think the core of this has been merged in other PRs, it's easier to close than to tackle.

PR #1832, opened 2017-10-06
Initial stub for IntelliJ Code Style settings

vkolomeyko:


mkit:
Don't we need .gitignore update as well for this?

vkolomeyko:
I do not think we need to fiddle with `.gitignore` as it only prevents *adding* files to Git.
Once the file is added (I had to run: `git add -f`) it can be version controlled just like any other file.

mikehearn:
If you do a reformatting run with these settings, does it cause a lot of churn?

vkolomeyko:
Without imports optimisation enabled - 463 files in the whole Corda codebase.
Changes are mainly around managing rogue spaces to make code look consistent.

tlil:
@mikehearn see [here](https://github.com/corda/corda/compare/tlil/reformat-2017-oct?expand=1) for a snapshot of the output using the current settings.

PR #1835, opened 2017-10-06
[CORDA-694] Commands visibility for Oracles (without sacrificing privacy)

kchalkias:
A new functionality is implemented to enable command visibility for Oracles without sacrificing privacy.
Previously, one could achieve the above functionality by requiring full access to every command in the original transaction.
In contrast, this approach suggests we only send the list of `signers` per `command`, but not `CommandData` for unrelated `Command`s.
oraclevisibility_commands
The algorithm works as follows:
a) a new `SIGNERS_GROUP` with ordinal 6 is added.

b) Unlike other components, signers is NOT a property in `TraversableTransaction`, but of course it is part of the `ComponentGroup` `OpaqueBytes`.

c) Due to the above, we ensure that filtering over `signers` is not possible (eg. someone by mistake sends some commands but filters out the signers)

d) If even one `Command` is included in the `FilteredTransaction`, the list of `signers` for all commands should be sent.

e) The ftx receiver (eg an Oracle) does the following check:
 - `checkAllComponentsVisible(SIGNERS_GROUP)`
 - Match the number of received commands (for its owning key) with the expected commands to sign based on the list of signers received.

Note that although a new component_type is added, this model remains backwards compatible with Corda 1.0 (both as an api and as a process).

Based on various conversations with @rick-r3 
command_visibility_versiontable

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
@@ -25,7 +25,7 @@ abstract class TraversableTransaction(open val componentGroups: List<ComponentGr
     override val outputs: List<TransactionState<ContractState>> = deserialiseComponentGroup(ComponentGroupEnum.OUTPUTS_GROUP, { SerializedBytes<TransactionState<ContractState>>(it).deserialize(context = SerializationFactory.defaultFactory.defaultContext.withAttachmentsClassLoader(attachments)) })
 
     /** Ordered list of ([CommandData], [PublicKey]) pairs that instruct the contracts what to do. */
-    val commands: List<Command<*>> = deserialiseComponentGroup(ComponentGroupEnum.COMMANDS_GROUP, { SerializedBytes<Command<*>>(it).deserialize(context = SerializationFactory.defaultFactory.defaultContext.withAttachmentsClassLoader(attachments)) })
+    val commands: List<Command<*>> = deserialiseCommands()
 
     override val notary: Party? = let {
         val notaries: List<Party> = deserialiseComponentGroup(ComponentGroupEnum.NOTARY_GROUP, { SerializedBytes<Party>(it).deserialize() })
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
```diff
@@ -229,18 +250,57 @@ class FilteredTransaction private constructor(
...
+     * @param publicKey signer's [PublicKey]
+     * @throws ComponentVisibilityException if not all of the related commands are visible.
+     */
+    @Throws(ComponentVisibilityException::class)
+    fun checkCommandVisibility(publicKey: PublicKey) {
+        val commandSigners = componentGroups.firstOrNull { it.groupIndex == ComponentGroupEnum.SIGNERS_GROUP.ordinal }
+        if (commandSigners == null) {
+            // Because a list of signing keys is not provided, we should at least ensure all commands are visible.
+            // TODO: Consider removing this requirement, as old clients might not follow this rule.
+            checkAllComponentsVisible(ComponentGroupEnum.COMMANDS_GROUP)
```
kchalkias: Let's decide on this requirement. Briefly: until Corda 1.0, an Oracle could use  `checkAllComponentsVisible(ComponentGroupEnum.COMMANDS_GROUP)` to verify it sees all commands, BUT we don't enforce this rule.
However, IMO it should be the recommended approach; new clients shouldn't sign if they can't ensure they see all related commands. There is indeed a trade-off between privacy and security and IMO security in this particular case is more important. Of course this is debatable.

rick-r3:
Does anything stop me from massaging a `WireTransaction` such that the signers does not accurately reflect those on the commands?

kchalkias:
@rick-r3 not with this version. We could construct commands however using the signers fields, ignoring `signers` field inside each `Command`. And eventually the command component will only include `CommandData`. I didn't remove it now, as it wouldn't be backwards compatible anymore.

kchalkias:
@rick-r3 Practically, the suggested approach is (for old clients) similar to the [ComponentMapper](https://www.lucidchart.com/invitations/accept/11477866-4cd0-4472-8a01-b2fadce1fa47) solution. As you mentioned, WireTX massaging is possible in both cases.
This is better than nothing however, mainly because previously it was much easier (programmatically or accidentally) to deceive the Oracle by filtering out a related command, even if the original WireTx was as expected.
Now, you need to hack the WireTx first, but if it is formed correctly and people sign it, then none of the users cannot play games with the FilteredTransaction.

kchalkias:
@mikehearn, @adagys and @rick-r3 we should decide if this approach should be applied. It's more secure than our current v1.0 model, but it will be 100% secure against WireTx massaging attacks if everyone uses the new version.

mikehearn:
If the implementation changed such that Command objects are always built by combining the different parts of the tree, would this be API compatible? (obviously not wire compatible)

What specific API change would be needed to ensure you can't mis-align Command.signers vs the signers field on the tx? Note that we used to have a signers field on wtx and took it out for this specific reason!

kchalkias:
Thank you @mikehearn. To resolve the aforementioned privacy/security issues, `Command`s should always be constructed from both subTrees [our slack conversation here](https://cordaledger.slack.com/messages/C3J04VC3V/). I think the sub-tree structure can guarantee each commandData is securely linked to its corresponding signers list.

kchalkias:
back to you @rick-r3, see my previous comment (we've agreed with Mike that we can tolerate wire compatibility for this one).

kchalkias:
@rick-r3 changelog update and rebase

kchalkias:
back to you @rick-r3

kchalkias:
@mikehearn I am going to merge this tomorrow.

mikehearn:
Unfortunately I wasn't able to catch up with this yesterday as I was travelling.

Did you update the docsite to discuss the new API? Did we update the sample code?

PR #1836, opened 2017-10-06
Deflake node infoWatcherTest

al-r3:
Change the implementation not to rely on the Java WatchService and instead just list the whole directory contents every time.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
+++ b/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
@@ -3,7 +3,12 @@ package net.corda.node.services.network
 import net.corda.cordform.CordformNode
 import net.corda.core.crypto.SecureHash
 import net.corda.core.crypto.SignedData
-import net.corda.core.internal.*
+import net.corda.core.internal.createDirectories
+import net.corda.core.internal.div
+import net.corda.core.internal.isDirectory
+import net.corda.core.internal.isRegularFile
+import net.corda.core.internal.list
+import net.corda.core.internal.readAll
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.services.KeyManagementService
 import net.corda.core.serialization.deserialize
@@ -13,10 +18,6 @@ import rx.Observable
 import rx.Scheduler
 import rx.schedulers.Schedulers
 import java.nio.file.Path
-import java.nio.file.StandardWatchEventKinds
-import java.nio.file.WatchEvent
-import java.nio.file.WatchKey
-import java.nio.file.WatchService
 import java.util.concurrent.TimeUnit
 import kotlin.streams.toList
 
@@ -33,7 +34,6 @@ class NodeInfoWatcher(private val nodePath: Path,
                       private val scheduler: Scheduler = Schedulers.io()) {
 
     private val nodeInfoDirectory = nodePath / CordformNode.NODE_INFO_DIRECTORY
-    private val watchService : WatchService? by lazy { initWatch() }
 
     companion object {
         private val logger = loggerFor<NodeInfoWatcher>()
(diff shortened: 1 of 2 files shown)


shamsasari:
Does Mike's suggestion not work?

rnicoll:
Certainly even when sensitivity is set to it took over 15 seconds (didn't measure how long exactly) to spot a directory change. Given it's a low-frequency check it seems neater to ditch the whole thing until the `WatchService` works correctly on all major platforms.

rnicoll:
Works on OS X with this applied

shamsasari:
We're going about this the wrong way. The reason the test took long is because the poll time is set to 5 secs, which eons for a unit test. We should be updating the code to have the poll time passed in and use the `HIGH` setting.

shamsasari:
My bad, the test is advancing the time manually.

al-r3:
The fundamental issue is that a WatchService makes no promise on the time it takes to realize something has changed really.
After some discussion with Victor I've relocated the test under integration tests as this is relying on an actual file system implementation.

Another option would be for the code to take a FileSystem implementation as an argument, so that in a test I can pass a mock and make it do whatever I think a real FileSystem would do.

That would make the test, smaller, faster but it wouldn't exercise an actual filesystem polling

shamsasari:
Making it an integration test makes sense.

al-r3:
With the help of Ross we tried Mike's suggestion but it didn't work.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -67,12 +67,15 @@ class NodeInfoWatcher(private val nodePath: Path,
      * Read all the files contained in [nodePath] / [CordformNode.NODE_INFO_DIRECTORY] and keep watching
      * the folder for further updates.
      *
+     * We simply list the direcotry content every 5 seconds, the Java implementation of WatchService has been proven to
```
vkolomeyko: Spelling mistake in word `directory`
Also why do we have 5 seconds hardcoded? Would it not make sense to add a configurable parameter into `reference.conf` ?

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -67,12 +67,15 @@ class NodeInfoWatcher(private val nodePath: Path,
...
+     * be unreliable on MacOs and given the fairly simple use case we have, this simple implementation should do.
+     *
      * @return an [Observable] returning [NodeInfo]s, there is no guarantee that the same value isn't returned more
      *      than once.
      */
     fun nodeInfoUpdates(): Observable<NodeInfo> {
         val pollForFiles = Observable.interval(5, TimeUnit.SECONDS, scheduler)
-                .flatMapIterable { pollWatch() }
+                .flatMapIterable { loadFromDirectory() }
         val readCurrentFiles = Observable.from(loadFromDirectory())
```
vkolomeyko: Why is it necessary to use call to `loadFromDirectory()` twice when composing resulting `Observable` ?

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
al-r3: fixed the typo.

It seems overkill to configure this really. At most someone might not enable the feature entirely, but you get that for free if you don't create additional-node-infos

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
al-r3: Listing the directory makes this actually unneeded.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
vkolomeyko: I am still a bit unsold on what sort of harm will it cause to make it configurable rather than hardcoded.

If anything, it should be possible to reduce this directory scan interval in the integration test to make it pass faster.

shamsasari: That's a good point.

al-r3: I'll do that in a separate PR, so that we can deflake the tests under macOS

vkolomeyko:
As discussed with @al-r3, introducing interval as parameter will be done as a separate PR.

PR #1847, opened 2017-10-09
Fix coin selection with Flow-friendly sleep

rick-r3:
Adds to the `FlowLogic` API with a new sleep function that correctly checkpoints the flow, which is needed in the coin selection to avoid leaking the database connection.

Ideally we should backport this to 1.0.  There's a nasty workaround to avoid this API change in 1.0 by setting retries to 1, but then coin selection will fail in some scenarios that should succeed.

Coin selection refactored:
1. Avoid having JDBC statement on the stack, via use of private function, when `sleep()` checkpoints since statements are blacklisted in checkpoints.
2. Added a small amount of exponential back-off since this actually improved things when testing with a modified trader demo to reproduce the problem.
3. Fixed up slightly inconsistent clear of states set.

Added a test that fails on the old implementation.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
@@ -50,6 +50,7 @@ abstract class FlowLogic<out T> {
         /**
          * Return the outermost [FlowLogic] instance, or null if not in a flow.
          */
+        @JvmStatic
         val currentTopLevel: FlowLogic<*>? get() = (Strand.currentStrand() as? FlowStateMachine<*>)?.logic
 
         /**
@@ -60,6 +61,7 @@ abstract class FlowLogic<out T> {
          * support for flow migration!
          */
         @Suspendable
+        @JvmStatic
         fun sleep(duration: Duration) {
             (Strand.currentStrand() as? FlowStateMachine<*>)?.sleepUntil(Instant.now() + duration) ?: Strand.sleep(duration.toMillis())
         }
--- a/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
@@ -55,14 +55,21 @@ abstract class FlowLogic<out T> {
 
         /**
          * If on a flow, suspends the flow and only wakes it up after at least [duration] time has passed.  Otherwise,
-         * just sleep for [duration].
+         * just sleep for [duration].  This sleep function is not designed to aid scheduling, for which you should
+         * consider using [SchedulableState].  It is designed to aid with managing contention for which you have not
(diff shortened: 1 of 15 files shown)


core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -43,6 +46,25 @@ abstract class FlowLogic<out T> {
     /** This is where you should log things to. */
     val logger: Logger get() = stateMachine.logger
 
+    companion object {
+        /**
+         * Return the outermost [FlowLogic] instance, or null if not in a flow.
+         */
+        val currentTopLevel: FlowLogic<*>? get() = (Strand.currentStrand() as? FlowStateMachine<*>)?.logic
```
mikehearn: Remember that these need `@JvmStatic` for Java users!

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -43,6 +46,25 @@ abstract class FlowLogic<out T> {
...
+
+        /**
+         * If on a flow, suspends the flow and only wakes it up after at least [duration] time has passed.  Otherwise,
+         * just sleep for [duration].
+         *
+         * Note that long sleeps and in general long running flows are highly discouraged, as there is currently no
```
mikehearn: Should we assert that the duration is, say, <5 minutes?

rick-r3: Yeah, I think I will update the comments to suggest this is really designed for resolving an issue with contention (that hasn't been solved through some other means), rather than for scheduling, for which `SchedulableState` should be used.

For resolving contention, something like 5 minutes sounds like plenty.  If they need more than 5 minutes, then `sleep` isn't what they need.  I'll add the check.

rick-r3:
Added changelog.  So now have to rebase.

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -494,6 +486,10 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
             }
         }
 
+        if (ioRequest is Sleep) {
```
exFalso: Should this be here or in `actionOnSuspend` with the other IOs?

rick-r3: Needs to be outside the `parkAndSerialize` so that it executes on the Fiber.

rick-r3: But I do think I should check there wasn't an exception during suspend.   Will modify.

PR #1849, opened 2017-10-09
Add confidential identity support to cash exit logic

rnicoll:
generateExit() previously required all states to have a single owner, which was used as the change output's owner. This deprecates that function but enables it to work as expected using `firstOrNull()` instead of `singleOrNull()`, and adds new functions which take in the change output's owner.


mnesbit:
I think people will want to cherry pick this onto 1.1

mikehearn:
Possibly - it's probably better to leave it until 2.0 unless it's requested though.

rnicoll:
It's been requested for Ubin, I'll cherry-pick it once merged on master.

rnicoll:
@roger3cev This needs a review, if you could!

rnicoll:
Going to merge this as-is and Roger can tackle anything minor on the 1.1 version

Issue #1730, opened 2017-09-29
Problems loading Corda Node on an Android App


yshprasanth:
I am working on a POC to deploy a Corda Node on an Android App.

First step for is to ensure the Corda libraries are included in our APK and deployed to Android Emulator.

In this process, I have add below dependency in my app's gradle.build file:

	implementation("net.corda:corda-node:0.14.0", {
        exclude group: 'com.github.bft-smart', module: 'library'
        exclude group: 'com.github.corda.crash', module: 'crash.shell'
        exclude group: 'org.jetbrains.exposed', module: 'exposed'
        //exclude group: 'io.netty'
        exclude group: 'io.netty', module: 'netty-transport'
        exclude group: 'io.netty', module: 'netty-handler'
        transitive = true
    })
	
	packagingOptions {
        // Excluding the below files to ensure the build of APK with Corda Node library works ok
        exclude 'META-INF/INDEX.LIST'
        exclude 'META-INF/DEPENDENCIES'
        exclude 'META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat'
        exclude 'META-INF/io.netty.versions.properties'
    }
	
The APK build was fine after the above configuration with multiple dex files. However, when I am starting the APK, I see the below error:

	_java.io.IOException: Failed to open dex files from /data/app/com.example4.corda.cordamobileapp4-Ma2gE0QLACHd94ICti9wGg==/split_lib_dependencies_apk.apk because: Failure to verify dex file '/data/app/com.example4.corda.cordamobileapp4-Ma2gE0QLACHd94ICti9wGg==/split_lib_dependencies_apk.apk': _Invalid method name: 'owned by'_
                 
This is because of having the method names as `issued by` an `owned by` with back ticks and spaces in below files:

		○ corda\finance\src\main\kotlin\net\corda\finance\contracts\asset\Cash.kt
		○ corda\finance\src\main\kotlin\net\corda\finance\contracts\asset\Obligation.kt
		○ corda\finance\src\main\kotlin\net\corda\finance\contracts\CommercialPaper.kt
		○ corda\finance\src\main\kotlin\net\corda\finance\Currencies.kt
		○ corda\samples\irs-demo\src\test\kotlin\net\corda\irs\api\NodeInterestRatesTest.kt

		
I have made the changes in my local copy, to remove space, re-compile to JAR. 

I was then able to launch the APP with the modified corda-finances.jar.

[...]

PR #1850, opened 2017-10-09
[GH 1730] Remove functions with spaces in their name

al-r3:
See https://github.com/corda/corda/issues/1730

Remove all the Kotlin functions with spaces in them since the Android doesn't support them.

### DIFF
--- a/finance/src/main/kotlin/net/corda/finance/contracts/asset/Cash.kt
+++ b/finance/src/main/kotlin/net/corda/finance/contracts/asset/Cash.kt
@@ -129,10 +129,10 @@ class Cash : OnLedgerAsset<Currency, Cash.Commands, Cash.State>() {
         override fun toString() = "${Emoji.bagOfCash}Cash($amount at ${amount.token.issuer} owned by $owner)"
 
         override fun withNewOwner(newOwner: AbstractParty) = CommandAndState(Commands.Move(), copy(owner = newOwner))
-        fun ownedBy(owner: AbstractParty) = copy(owner = owner)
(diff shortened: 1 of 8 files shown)


finance/src/main/kotlin/net/corda/finance/contracts/asset/Cash.kt
```diff
@@ -401,10 +401,17 @@ class Cash : OnLedgerAsset<Currency, Cash.Commands, Cash.State>() {
 
 // Small DSL extensions.
 
-/** @suppress */ infix fun Cash.State.`owned by`(owner: AbstractParty) = ownedBy(owner)
-/** @suppress */ infix fun Cash.State.`issued by`(party: AbstractParty) = issuedBy(party)
-/** @suppress */ infix fun Cash.State.`issued by`(deposit: PartyAndReference) = issuedBy(deposit)
-/** @suppress */ infix fun Cash.State.`with deposit`(deposit: PartyAndReference): Cash.State = withDeposit(deposit)
+/** @suppress */
+infix fun Cash.State.ownedBy(owner: AbstractParty) = ownedBy(owner)
```
rnicoll: I believe these were just convenience wrappers around functions already present on `Cash.State`, so probably need to deleted rather than rewritten.

al-r3: They're also infix, the wrapped functions aren't infix.
Other than that I agree these functions here don't really add anything new, what do you think?

rnicoll: Can we change the underlying functions to infix and remove  these? Hopefully that will also fix the build errors.

al-r3:
All done, I didn't immediately realize you can call infix functions as regular functions.

PR #1851, opened 2017-10-09
Add a configuration option to set how often should a node check for new nodeinfos on disk

al-r3:
This was discussed in the PR https://github.com/corda/corda/pull/1836


node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -86,7 +88,8 @@ data class FullNodeConfiguration(
         override val devMode: Boolean = false,
         val useTestClock: Boolean = false,
         val detectPublicIp: Boolean = true,
-        override val activeMQServer: ActiveMqServerConfiguration
+        override val activeMQServer: ActiveMqServerConfiguration,
+        override val additionalNodeInfoPollingFrequencyMsec: Long = 5.seconds.toMillis()
```
vkolomeyko: I think it is more traditional to put default values in `reference.conf` rather than here.
@mnesbit - please feel free to correct me if I am wrong.

al-r3:
NodeInfoWatcherTest never had a real time. It used Observer's way of advancing time, this has no effect on the time tests take. Nor the polling interval ever affected the test runtime anyway.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
al-r3: There are a few lines above with defaults.

mnesbit: My concern is this looks likely to cause multiple parallel scans (especially if someone sets it to 1 ms). It is always good practice to protect this sort of code against overlapping timer events. I have seen too many production issues when that wasn't followed.

al-r3: I hardly see people having the need to change this.
Someone might want to disable it, said that, I can change the code so that it always takes the largest of (5 minutes, user specified value), what do you think?

al-r3: I've made the code not poll faster than once every 5 seconds

PR #1852, opened 2017-10-09
Non-material public API change due to interfaces being re-arranged

vkolomeyko:


vkolomeyko:
Actual change was made by #1630

PR #1853, opened 2017-10-09
Protect the serialization/deserialization code of SessionMessage data

mnesbit:
The current SessionMessage deserialization is not within an exception catcher and also the payload is directly deserialized, which prevents any opportunity to apply the checking logic if the class is not available to the node. This puts some basic checks in, although the full fix requires the safe AMQP serializers, rather than Kryo.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/SessionMessage.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/SessionMessage.kt
@@ -5,7 +5,10 @@ import net.corda.core.flows.UnexpectedFlowEndException
 import net.corda.core.identity.Party
 import net.corda.core.internal.castIfPossible
 import net.corda.core.serialization.CordaSerializable
+import net.corda.core.serialization.SerializationDefaults
+import net.corda.core.serialization.SerializedBytes
 import net.corda.core.utilities.UntrustworthyData
+import java.io.IOException
 
 @CordaSerializable
 interface SessionMessage
@@ -25,7 +28,7 @@ data class SessionInit(val initiatorSessionId: Long,
                        val initiatingFlowClass: String,
                        val flowVersion: Int,
                        val appName: String,
-                       val firstPayload: Any?) : SessionMessage
+                       val firstPayload: SerializedBytes<Any>?) : SessionMessage
 
 data class SessionConfirm(override val initiatorSessionId: Long,
                           val initiatedSessionId: Long,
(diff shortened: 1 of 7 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/SessionMessage.kt
```diff
@@ -25,7 +27,7 @@ data class SessionInit(val initiatorSessionId: Long,
                        val initiatingFlowClass: String,
                        val flowVersion: Int,
                        val appName: String,
-                       val firstPayload: Any?) : SessionMessage
+                       val firstPayload: SerializedBytes<Any>?) : SessionMessage
```
shamsasari: Perhaps have this as `ByteArray` to avoid the extra object indirection.

node/src/main/kotlin/net/corda/node/services/statemachine/SessionMessage.kt
```diff
@@ -34,16 +36,23 @@ data class SessionConfirm(override val initiatorSessionId: Long,
...
 data class ReceivedSessionMessage<out M : ExistingSessionMessage>(val sender: Party, val message: M)
 
-fun <T> ReceivedSessionMessage<SessionData>.checkPayloadIs(type: Class<T>): UntrustworthyData<T> {
-    return type.castIfPossible(message.payload)?.let { UntrustworthyData(it) } ?:
+fun <T : Any> ReceivedSessionMessage<SessionData>.checkPayloadIs(type: Class<T>): UntrustworthyData<T> {
+    val payloadData: T = try {
+        val serializer = SerializationDefaults.SERIALIZATION_FACTORY
+        serializer.deserialize<T>(message.payload, type, SerializationDefaults.P2P_CONTEXT)
+    } catch (ex: Exception) {
+        throw UnexpectedFlowEndException("Payload invalid")
```
shamsasari: I don't think this is needed. `UnexpectedFlowEndException` is automatically thrown when the other side flow has finished unexpectedly. If we're trying to finish ourselves then just leaving the exception to propagate is sufficient. The other side will automatically get a `UnexpectedFlowEndException`. Basically this is not needed.

node/src/main/kotlin/net/corda/node/services/statemachine/StateMachineManager.kt
```diff
@@ -290,7 +290,12 @@ class StateMachineManager(val serviceHub: ServiceHubInternal,
     }
 
     private fun onSessionMessage(message: ReceivedMessage) {
-        val sessionMessage = message.data.deserialize<SessionMessage>()
+        val sessionMessage = try {
+            message.data.deserialize<SessionMessage>()
+        } catch (ex: Exception) {
+            logger.error("Received corrupt message data from ${message.peer}")
```
shamsasari: "Receive corrupt session message from ..."

node/src/main/kotlin/net/corda/node/services/statemachine/SessionMessage.kt
mnesbit: I will use ByteSequence as ByteArray doesn't have the correct equality logic

shamsasari: ByteSequence has the same issue - it wrap ByteArray. If you're after correct equality then what you had originally is fine. My suggestion was purely an optimisation.

PR #1855, opened 2017-10-09
Make integration tests pass in AMQP mode, part 1

vkolomeyko:


docs/source/serialization.rst
```diff
@@ -259,9 +259,10 @@ Kotlin Objects
 ``````````````
 
     #.  Kotlin ``object`` s are singletons and treated differently.  They are recorded into the stream with no properties
-        and deserialize back to the singleton instance.
-
-Currently, the same is not true of Java singletons, and they will deserialize to new instances of the class.
+        and deserialize back to the singleton instance. Currently, the same is not true of Java singletons,
+        and they will deserialize to new instances of the class.
+    #.  Kotlin's anonymous ``object`` s are not currrently supported. I.e. constructs like:
+        ``object : Contract {...}`` will not serialize corrrectly and need to be re-written as an explicit class declarations.
```
rick-r3: declarations -> declaration

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationOutputTests.kt
```diff
@@ -534,8 +534,22 @@ class SerializationOutputTests {
...
-    val FOO_PROGRAM_ID = "net.corda.nodeapi.internal.serialization.amqp.SerializationOutputTests.FooContract"
+    @Test
+    fun `test custom object`() {
+        serdes(FooContract)
+    }
+
+    @Test
+    @Ignore("Cannot serialize due to known Kotlin/serialization limitation")
+    fun `test custom anonymous object`() {
+        val anonymous: Contract = object : Contract {
```
rick-r3: Yeah, think I agree that this is undesirable.  The types should be nice, public, well named types since they are supposed to be describing a schema.

docs/source/serialization.rst
vkolomeyko: Fixed + few other typos

PR #1856, opened 2017-10-09
Removed extraAdvertisedServiceIds config

shamsasari:
The remaining use for it was the finance CorDapp for permissioning CCY issuers. Instead this is now taken from a custom config in node.conf.

The DemoBench also needed updating. I used the `toConfig()` and `parseAs()` extension methods to make it easier to deal with the changes in the node configuration. As a result DemoBench's config handling code is much simpler.

In my previous PR the notary configuration was moved from `extraAdvertisedServiceIds` to the new `notary` config. I hadn't updated DemoBench then since it would be easier to update it in one go when `extraAdvertisedServiceIds` is removed.


finance/src/main/kotlin/net/corda/finance/flows/CashConfigDataFlow.kt
```diff
@@ -1,38 +1,59 @@
...
+@CordaService
+class ConfigHolder(services: AppServiceHub) : SingletonSerializeAsToken() {
+    companion object {
+        val supportedCurrencies = listOf(USD, GBP, CHF, EUR)
+    }
+
+    val issuableCurrencies: List<Currency>
+
+    init {
+        // Warning!! You are about to see a major hack!
```
kasiastreich: It made my day instantly better.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeConfig.kt
```diff
@@ -1,87 +1,79 @@
...
+            cordapp.copyToDirectory(pluginDir, StandardCopyOption.REPLACE_EXISTING)
         }
     }
 
-}
-
-private fun <T> valueFor(any: T): ConfigValue? = ConfigValueFactory.fromAnyRef(any)
 
-private fun addressValueFor(port: Int) = valueFor("localhost:$port")
+    }
```
chrisr3: Whitespace...!

PR #1857, opened 2017-10-10
Create nodeInfoDirectory on NodeInfoWatcher init and don't log when it does not exist

al-r3:
Let NodeInfoWatcher create the directory it wants to poll at startup.
Also log failure in creating the directory, but don't log anything if it can be found at poll time.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
+++ b/node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
@@ -12,6 +12,7 @@ import net.corda.core.utilities.seconds
 import rx.Observable
 import rx.Scheduler
 import rx.schedulers.Schedulers
+import java.io.IOException
 import java.nio.file.Path
 import java.util.concurrent.TimeUnit
 import kotlin.streams.toList
@@ -32,7 +33,8 @@ class NodeInfoWatcher(private val nodePath: Path,
                       private val scheduler: Scheduler = Schedulers.io()) {
 
     private val nodeInfoDirectory = nodePath / CordformNode.NODE_INFO_DIRECTORY
-    private val pollFrequencyMsec: Long
+    private val pollFrequencyMsec: Long = maxOf(pollFrequencyMsec, 5.seconds.toMillis())
+    private val successfullyProcessedFiles = mutableSetOf<Path>()
 
     companion object {
         private val logger = loggerFor<NodeInfoWatcher>()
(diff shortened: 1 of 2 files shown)


al-r3:
It isn't going to be (much) better if it logs when it reads something, since it will keep on logging it at every scan. It seems more sane to just not log anything. (logging new previously unseen files would require keeping track of returned files which seems overkill to do just for logging)

chrisr3:
Interesting. Why don't you move each file into a different location after you've seen it once? Is there a reason why you need to keep reprocessing them each time?

al-r3:
Well I can't move nor rename the files because I'd need to move them back when the node quits (to be able to read them at the next start-up).
And the reason I reprocess them each time is that it doesn't take much to do so (the initial approach which didn't do that has been scrapped due to the Java polling issues on MacOS).

I've now changed the code to keep track of files which where successfully processed and log something only when a new file is seen.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -87,16 +94,20 @@ class NodeInfoWatcher(private val nodePath: Path,
...
+                    .map { path ->
+                        val nodeInfo = processFile(path)
+                        if (nodeInfo != null) successfullyProcessedFiles.add(path)
+                        return@map nodeInfo
+                    }
                     .toList()
                     .filterNotNull()
         }
-        logger.info("Successfully read ${result.size} NodeInfo files.")
+        if (result.size != 0) logger.info("Successfully read ${result.size} NodeInfos from disk.")
```
chrisr3: Kotlin has `isNotEmpty()` as shorthand for `size != 0`. Also prefer `if` with curly brackets, even for a single line:
```kotlin
if (result.isNotEmpty()) {
    logger.info("...")
}
```
And "NodeInfo files" reads better than "NodeInfos".

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -87,16 +94,20 @@ class NodeInfoWatcher(private val nodePath: Path,
...
-            logger.info("$nodeInfoDirectory isn't a Directory, not loading NodeInfo from files")
             return emptyList()
         }
         val result = nodeInfoDirectory.list { paths ->
-            paths.filter { it.isRegularFile() }
-                    .map { processFile(it) }
+            paths.filter { it !in successfullyProcessedFiles }
+                    .filter { it.isRegularFile() }
+                    .map { path ->
+                        val nodeInfo = processFile(path)
```
chrisr3: I think this can be simplified to:
```kotlin
.map { path -> processFile(path)?.apply { successfullyProcessedFiles.add(path) } }
```

al-r3: While writing that code I suspected there was a more kotlin-esque way of doing that.

al-r3:
All done

PR #1860, opened 2017-10-10
CORDA-680 Retire setCordappPackages

andr3ej:


experimental/src/test/kotlin/net/corda/finance/contracts/universal/Caplet.kt
```diff
@@ -53,17 +48,6 @@ class Caplet {
...
-    @Before
-    fun setup() {
-        setCordappPackages("net.corda.finance.contracts.universal")
-    }
-
-    @After
-    fun tearDown() {
-        unsetCordappPackages()
-    }
-
```
andr3ej: trick: these tests will now use the transaction function defined in Cap.kt

rick-r3:
This makes sense to me.  I can see us wanting to control the packages per node.  e.g. the notary shouldn't really run with the same "cordapp" as the other nodes.  The network-wide setting makes sense as a default (it's just easiest).  A follow up PR to override / supplement per node, rather than making this change bigger?

andr3ej:
Definitely a separate PR, as the focus of this one is to make the mechanism non-static (and this one doesn't make the granularity any worse than it currently is). That work can go in the same jira ticket though.

PR #1861, opened 2017-10-10
 Add confidential identity support to cash exit logic

rnicoll:
release-V1 equivalent of #1849


rnicoll:
Having some sort of weird issue with `PersistentNetworkMapCacheTest` and this, but I think that's in `release-V1` too

Clintonio:
There is no 1.1 that I am aware of (despite it existing in GH). This will need to be pushed into a later release. Is it in master already?

rnicoll:
This is in master, so seems a bit pointless merging into the 1.1 branch now

PR #1862, opened 2017-10-10
Fixed the dozens of artifactory warnings at gradle initialisation.

Clintonio:
Why: Because they are annoying.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -163,25 +163,6 @@ allprojects {
     }
 }
 
-
-artifactory {
-    publish {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -293,12 +293,16 @@ artifactory {
     publish {
         contextUrl = 'https://ci-artifactory.corda.r3cev.com/artifactory'
         repository {
-            repoKey = 'corda-releases'
+            repoKey = 'corda-dev'
```
Clintonio: Made this corda dev since nobody should ever push to releases from local (and this block is mostly for TC which overrides this particular field)

PR #1863, opened 2017-10-11
Aslemmer kryo hook

exFalso:
This is just a small javaagent tool that hooks into kryo `write() ` calls to gather the offsets in the output in order to get a feel for how much space certain objects take.

Perhaps we can extend it to hook into amqp or other things later

We could also attach stacktraces to map our serialize() calls around the codebase


experimental/kryo-hook/src/main/kotlin/net/corda/kryohook/KryoHook.kt
```diff
@@ -0,0 +1,168 @@
...
+ * TODO add Field constructor.
+ */
+sealed class StatsTree {
+    data class Object(
+            val className: String,
+            val size: Long,
+            val children: List<StatsTree>
+    ) : StatsTree()
+}
+
```
rick-r3: Remove blank line

exFalso: removed

PR #1866, opened 2017-10-11
Clean up Crypto 

kchalkias:
Also, `PublicKey.toSHA256Bytes()` now uses `.encoded`

### DIFF
--- a/core/src/main/kotlin/net/corda/core/utilities/EncodingUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/utilities/EncodingUtils.kt
@@ -3,9 +3,8 @@
 package net.corda.core.utilities
 
 import net.corda.core.crypto.Base58
+import net.corda.core.crypto.Crypto
 import net.corda.core.crypto.sha256
-import net.corda.core.serialization.deserialize
-import net.corda.core.serialization.serialize
 import java.nio.charset.Charset
 import java.security.PublicKey
 import java.util.*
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/net/corda/core/utilities/EncodingUtils.kt
```diff
@@ -65,7 +67,15 @@ fun String.hexToBase64(): String = hexToByteArray().toBase64()
...
 //       structure, e.g. mapping a PublicKey to a condition with the specific feature (ED25519).
+/**
+ * Method to return the [PublicKey] object given its Base58-[String] representation.
+ * @param base58String the Base58 encoded format of the serialised [PublicKey].
+ * @return the resulted [PublicKey] after decoding the [base58String] input and then deserialising to a [PublicKey] object.
+ */
 fun parsePublicKeyBase58(base58String: String): PublicKey = base58String.base58ToByteArray().deserialize<PublicKey>()
 
+/** Return the Base58 representation of the serialised public key. */
 fun PublicKey.toBase58String(): String = this.serialize().bytes.toBase58()
```
patrickkuo: do we need to change this as well? Maybe we need to blacklist public key in kyro, or make kyro serialize keys by using `encoded`

kchalkias: I decided not to do, because we need to deserialise to read back. But, internally kryo uses encoded anyway.

kchalkias: Convinced, we should use use `.encoded`. Back to you @patrickkuo

PR #1867, opened 2017-10-11
Adds a length method to calculate a time-window's length.

joeldudleyr3:


core/src/main/kotlin/net/corda/core/contracts/TimeWindow.kt
```diff
@@ -64,6 +64,17 @@ abstract class TimeWindow {
      */
     abstract val midpoint: Instant?
 
+    /**
+     * Returns the duration between [fromTime] and [untilTime] if both are non-null. Otherwise returns null.
+     */
```
adagys: can use Duration.INDEFINITE instead of null

joeldudleyr3: I can't find it - where is it? https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html

adagys: ah, yes it's not in the java.time one

core/src/test/kotlin/net/corda/core/contracts/TimeWindowTest.kt
```diff
@@ -1,9 +1,12 @@
 package net.corda.core.contracts
 
+import net.corda.core.internal.div
+import net.corda.core.internal.times
```
kchalkias: do we use those two?

joeldudleyr3: Yes, for dividing/multiplying doubles in the new tests.

kchalkias: just noticed

kchalkias:
Could you also add changelog notes?

joeldudleyr3:
Added a change-log entry.

PR #1870, opened 2017-10-11
CORDA-540: Make Verifier work in AMQP mode

vkolomeyko:
Also ensure Attachments serialize correctly in AMQP mode


node-api/src/main/kotlin/net/corda/nodeapi/VerifierApi.kt
```diff
@@ -21,12 +29,22 @@ object VerifierApi {
...
+                val context = establishSuitableContext(bytesSequence.obtainHeaderSignature())
                 return VerificationRequest(
                         message.getLongProperty(VERIFICATION_ID_FIELD_NAME),
-                        ByteArray(message.bodySize).apply { message.bodyBuffer.readBytes(this) }.deserialize(),
+                        bytesSequence.deserialize(context = context),
                         MessageUtil.getJMSReplyTo(message)
                 )
             }
+
+            private fun establishSuitableContext(headerSignature: ByteSequence): SerializationContext =
```
rick-r3: I don't think this is needed.  The `SerializationFactory` will select the correct scheme based on the header for you.

vkolomeyko: Well... I have a different experience.
`VerifierAPI` runs as a separate process and separate VM. With `establishSuitableContext` method removed `.deserialize` will be called with `serializationFactory.defaultContext` which in the absence of `currentContext` defaults to `SerializationDefaults.P2P_CONTEXT` - which also may not be set.
In this case the following exception will be thrown:
```
[ERROR] 15:15:57,622 [Thread-0 (ActiveMQ-client-global-threads)] (ClientConsumerImpl.java:1122) core.client.run - AMQ214000: Failed to call onMessage
java.lang.IllegalStateException: Write-once property var net.corda.core.serialization.SerializationDefaults.P2P_CONTEXT: net.corda.core.serialization.SerializationContext not set.
	at net.corda.core.internal.WriteOnceProperty.getValue(WriteOnceProperty.kt:12) ~[classes/:?]
	at net.corda.core.serialization.SerializationDefaults.getP2P_CONTEXT(SerializationAPI.kt) ~[classes/:?]
	at net.corda.core.serialization.SerializationFactory.getDefaultContext(SerializationAPI.kt:42) ~[classes/:?]
	at net.corda.nodeapi.VerifierApi$VerificationRequest$Companion.fromClientMessage(VerifierApi.kt:79) ~[classes/:?]
	at net.corda.verifier.Verifier$Companion$main$3.onMessage(Verifier.kt:68) ~[classes/:?]
	at org.apache.activemq.artemis.core.client.impl.ClientConsumerImpl.callOnMessage(ClientConsumerImpl.java:997) ~[artemis-core-client-2.1.0.jar:2.1.0]
	at org.apache.activemq.artemis.core.client.impl.ClientConsumerImpl.access$400(ClientConsumerImpl.java:49) ~[artemis-core-client-2.1.0.jar:2.1.0]
	at org.apache.activemq.artemis.core.client.impl.ClientConsumerImpl$Runner.run(ClientConsumerImpl.java:1120) [artemis-core-client-2.1.0.jar:2.1.0]
	at org.apache.activemq.artemis.utils.OrderedExecutorFactory$OrderedExecutor$ExecutorTask.run(OrderedExecutorFactory.java:101) [artemis-commons-2.1.0.jar:2.1.0]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_144]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_144]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_144]
```

rick-r3: Yeah, the defaults needs setting, or we specify one of the P2Ps directly, but once it's using P2P the AMQP/Kryo bit should switch automatically (now you have registered the 2 implementations).

vkolomeyko: We have discussed this with @exFalso and agreed that it would be a more flexible solution to be able to handle messages in **both** encodings: Kryo and AMQP - especially given that it is not difficult to do at all.
As far as production roll-out is concerned, this will also make sense as not everyone might be able to upgrade at the same time, so if we will be able to handle Kryo and AMQP it might be a good thing.
Also current AMQP enable property named `net.corda.testing.amqp.enable` - it would be incorrect to use this testing property in production code of `net.corda.verifier.Verifier`

vkolomeyko: When we are saying "set the default", I am assuming we are talking about assigning value to `SerializationDefaults.P2P_CONTEXT`.
But this is `WriteOnceProperty` property at the moment, so first mechanism always wins and given that `Verifier` process can be running for some time, this may meant that second assignment with result `IllegalStateException`.

exFalso: No, the default should be set only once (in fact it's already set).
The decision on which scheme to use is in `SerializationScheme.kt:SerializationFacoryImpl.schemeFor`, which already checks this header and chooses the right scheme.

So in theory all we need to do is register AMQP.

vkolomeyko:
Discussed with @rick-r3 - will make some changes to capture the encoding/context of the inbound message to use the same encoding when responding back.
More changes on the way.

vkolomeyko:
More changes pushed.
@rick-r3 and @exFalso - over to you for review.

.ci/api-current.txt
```diff
@@ -2435,6 +2435,17 @@ public final class net.corda.core.serialization.MissingAttachmentsException exte
   public <init>(List)
   @org.jetbrains.annotations.NotNull public final List getIds()
 ##
+public final class net.corda.core.serialization.ObjectWithVersionHeader extends java.lang.Object
```
exFalso: @chrisr3 Should we be updating the api file like this?

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolverTests.kt
```diff
@@ -116,6 +116,10 @@ class CordaClassResolverTests {
         override fun <T : Any> serialize(obj: T, context: SerializationContext): SerializedBytes<T> {
             TODO("not implemented")
         }
+
+        override fun <T : Any> deserializeWithVersionHeader(byteSequence: ByteSequence, clazz: Class<T>, context: SerializationContext): ObjectWithVersionHeader<T> {
```
exFalso: perhaps we can use `val factory: SerializationFactory = mock()` from mockito instead of this

vkolomeyko: The method is not called in the test atm, so I think it should be fine to have a `TODO` implementation.

exFalso: I know, I meant replacing this whole object with a `mock()` call (which also throws exceptions by default), then we don't need to worry about extending it with TODOs

vkolomeyko: It is also possible to delete `factory` from `CordaClassResolverTests` as it is no longer used.

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -7,6 +7,8 @@ import net.corda.core.utilities.ByteSequence
 import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.utilities.sequence
 
+data class ObjectWithVersionHeader<out T : Any>(val obj: T, val versionHeader: VersionHeader)
```
rick-r3: I was thinking this should be:

`data class ObjectWithResponseContext<out T : Any>(val obj: T, val responseContext: SerializationContext)`

and the `responseContext` is populated with `context.withPreferredSerializationVersion(versionHeader)` for them, rather than exposing the header directly.

`Response` might not be the right word.  Maybe `Implicit` ?  `ObjectWithImplicitContext`.... not sure.

vkolomeyko: Changes are done and pushed. As discussed, the name for the container will be `ObjectWithCompatibleContext`.

PR #1871, opened 2017-10-11
Ensure that the contents of OpaqueBytes cannot be modified.

chrisr3:
Force `OpaqueBytes.bytes` to return a clone of its contents, and then make this field `final`. This effectively makes the contents immutable.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -2923,7 +2923,7 @@ public static final class net.corda.core.utilities.NonEmptySet$iterator$1 extend
 ##
 public class net.corda.core.utilities.OpaqueBytes extends net.corda.core.utilities.ByteSequence
   public <init>(byte[])
-  @org.jetbrains.annotations.NotNull public byte[] getBytes()
+  @org.jetbrains.annotations.NotNull public final byte[] getBytes()
   public int getOffset()
   public int getSize()
   public static final net.corda.core.utilities.OpaqueBytes$Companion Companion
--- a/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
@@ -19,13 +19,29 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
         }
     }
 
+    /**
+     * Convert the hash value to a hexadecimal [String].
+     */
     override fun toString(): String = bytes.toHexString()
 
+    /**
+     * Returns the first [prefixLen] hexadecimal digits of the [SecureHash] value.
+     * @param prefixLen
+     */
     fun prefixChars(prefixLen: Int = 6) = toString().substring(0, prefixLen)
+
+    /**
+     * Append a second hash value to this hash value, and then compute the SHA-256 hash of the result.
+     * @param other
+     */
     fun hashConcat(other: SecureHash) = (this.bytes + other.bytes).sha256()
 
     // Like static methods in Java, except the 'companion' is a singleton that can have state.
     companion object {
+        /**
+         * Converts a SHA-256 hash value represented as a hexadecimal [String] into a [SecureHash].
+         * @param str A sequence of 64 hexadecimal digits that represents a SHA-256 hash value.
+         */
         @JvmStatic
         fun parse(str: String) = str.toUpperCase().parseAsHex().let {
             when (it.size) {
@@ -34,24 +50,55 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
             }
         }
 
+        /**
+         * Computes the SHA-256 hash value of the [ByteArray].
+         * @param bytes
+         */
         @JvmStatic
         fun sha256(bytes: ByteArray) = SHA256(MessageDigest.getInstance("SHA-256").digest(bytes))
 
+        /**
+         * Computes the SHA-256 hash of the [ByteArray], and then computes the SHA-256 hash of the hash.
+         * @param bytes
+         */
         @JvmStatic
         fun sha256Twice(bytes: ByteArray) = sha256(sha256(bytes).bytes)
 
+        /**
+         * Computes the SHA-256 hash of the [String]'s UTF-8 byte contents.
+         * @param str
+         */
         @JvmStatic
         fun sha256(str: String) = sha256(str.toByteArray())
 
+        /**
+         * Generates a random SHA-256 value.
+         */
         @JvmStatic
         fun randomSHA256() = sha256(newSecureRandom().generateSeed(32))
 
+        /**
+         * A SHA-256 hash value consisting of 32 0x00 bytes.
+         */
+        @JvmField
         val zeroHash = SecureHash.SHA256(ByteArray(32, { 0.toByte() }))
+
+        /**
+         * A SHA-256 hash value consisting of 32 0xFF bytes.
+         */
+        @JvmField
         val allOnesHash = SecureHash.SHA256(ByteArray(32, { 255.toByte() }))
     }
 
     // In future, maybe SHA3, truncated hashes etc.
 }
 
+/**
+ * Compute the SHA-256 hash for the contents of the [ByteArray].
+ */
 fun ByteArray.sha256(): SecureHash.SHA256 = SecureHash.sha256(this)
+
+/**
+ * Compute the SHA-256 hash for the contents of the [OpaqueBytes].
+ */
 fun OpaqueBytes.sha256(): SecureHash.SHA256 = SecureHash.sha256(this.bytes)
--- a/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
@@ -32,7 +32,7 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
 
     /**
      * Append a second hash value to this hash value, and then compute the SHA-256 hash of the result.
-     * @param other
+     * @param other The hash to append to this one.
      */
     fun hashConcat(other: SecureHash) = (this.bytes + other.bytes).sha256()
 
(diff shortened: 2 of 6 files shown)


.ci/api-current.txt
```diff
@@ -771,10 +771,10 @@ public abstract class net.corda.core.crypto.SecureHash extends net.corda.core.ut
   @kotlin.jvm.JvmStatic @org.jetbrains.annotations.NotNull public static final net.corda.core.crypto.SecureHash$SHA256 sha256Twice(byte[])
   @org.jetbrains.annotations.NotNull public String toString()
   public static final net.corda.core.crypto.SecureHash$Companion Companion
+  @kotlin.jvm.JvmField @org.jetbrains.annotations.NotNull public static final net.corda.core.crypto.SecureHash$SHA256 allOnesHash
+  @kotlin.jvm.JvmField @org.jetbrains.annotations.NotNull public static final net.corda.core.crypto.SecureHash$SHA256 zeroHash
 ##
 public static final class net.corda.core.crypto.SecureHash$Companion extends java.lang.Object
-  @org.jetbrains.annotations.NotNull public final net.corda.core.crypto.SecureHash$SHA256 getAllOnesHash()
```
shamsasari: I guess we should have caught this before 1.0, but now we're stuck with these. The solution is to just deprecate these getters

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -19,13 +19,29 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
...
     }
 
+    /**
+     * Convert the hash value to a hexadecimal [String].
+     */
     override fun toString(): String = bytes.toHexString()
 
+    /**
+     * Returns the first [prefixLen] hexadecimal digits of the [SecureHash] value.
+     * @param prefixLen
```
shamsasari: Remove or document

.ci/api-current.txt
```diff
@@ -771,10 +771,10 @@ public abstract class net.corda.core.crypto.SecureHash extends net.corda.core.ut
   @kotlin.jvm.JvmStatic @org.jetbrains.annotations.NotNull public static final net.corda.core.crypto.SecureHash$SHA256 sha256Twice(byte[])
   @org.jetbrains.annotations.NotNull public String toString()
   public static final net.corda.core.crypto.SecureHash$Companion Companion
+  @kotlin.jvm.JvmField @org.jetbrains.annotations.NotNull public static final net.corda.core.crypto.SecureHash$SHA256 allOnesHash
```
shamsasari: Convention is all uppercase with underscores for static constants.

core/src/test/kotlin/net/corda/core/serialization/TransactionSerializationTests.kt
```diff
@@ -63,12 +64,6 @@ class TransactionSerializationTests : TestDependencyInjectionBase() {
 
         // Now check that the signature we just made verifies.
         stx.verifyRequiredSignatures()
-
-        // Corrupt the data and ensure the signature catches the problem.
-        stx.id.bytes[5] = stx.id.bytes[5].inc()
```
shamsasari: LOL

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -19,13 +19,29 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
         }
     }
 
+    /**
+     * Convert the hash value to a hexadecimal [String].
```
shamsasari: Is it uppercase?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -19,13 +19,29 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+     * Append a second hash value to this hash value, and then compute the SHA-256 hash of the result.
+     * @param other
+     */
     fun hashConcat(other: SecureHash) = (this.bytes + other.bytes).sha256()
 
     // Like static methods in Java, except the 'companion' is a singleton that can have state.
     companion object {
+        /**
+         * Converts a SHA-256 hash value represented as a hexadecimal [String] into a [SecureHash].
+         * @param str A sequence of 64 hexadecimal digits that represents a SHA-256 hash value.
```
shamsasari: Add `@throws IllegalArgumentException`

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -34,24 +51,63 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
...
 
-        val zeroHash = SecureHash.SHA256(ByteArray(32, { 0.toByte() }))
-        val allOnesHash = SecureHash.SHA256(ByteArray(32, { 255.toByte() }))
+        /**
+         * A SHA-256 hash value consisting of 32 0x00 bytes.
+         */
+        @JvmField
+        val ZERO_HASH = SecureHash.SHA256(ByteArray(32, { 0.toByte() }))
+        /** Compatibility field for Corda v1.0 */
+        @Deprecated("Replaced by ZERO_HASH in Corda v1.1")
```
mikehearn: See my email on the topic: when there's no particularly good reason to change to a new API we don't want to mark it as @Deprecated, as that will generated warnings. And if we do have a good reason, we should provide replaceBy so IntelliJ can do the fix automatically.

In this case, it doesn't seem worth renaming. Just leave it be.

core/src/test/kotlin/net/corda/core/serialization/TransactionSerializationTests.kt
mikehearn: You can't just remove this test because you can't be bothered fixing it :) Sorry. Go back and make sure that modified data in the wire transaction is caught by the signature check.

chrisr3:
@mikehearn I removed the test because it's no longer possible to alter the contents of the `OpaqueBytes`!

mikehearn:
I know why you removed the test. Tough! The test was checking something important - that the signature checking codepath actually works in this context. I don't care that this API change makes it slightly more awkward to test that ... it's an important property to check!

Use reflection or something to override the finality, or serialise, corrupt the serialised bytes, unserialise and check.

chrisr3:
It turns out that the `OpaqueBytes.bytes` backing field _was_ accessible via reflection after alll.

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -19,13 +19,30 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
...
 
+    /**
+     * Returns the first [prefixLen] hexadecimal digits of the [SecureHash] value.
+     * @param prefixLen The number of characters in the prefix.
+     */
     fun prefixChars(prefixLen: Int = 6) = toString().substring(0, prefixLen)
+
+    /**
+     * Append a second hash value to this hash value, and then compute the SHA-256 hash of the result.
+     * @param other
```
shamsasari: likewise

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -19,13 +19,30 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+     */
     fun hashConcat(other: SecureHash) = (this.bytes + other.bytes).sha256()
 
     // Like static methods in Java, except the 'companion' is a singleton that can have state.
     companion object {
+        /**
+         * Converts a SHA-256 hash value represented as a hexadecimal [String] into a [SecureHash].
+         * @param str A sequence of 64 hexadecimal digits that represents a SHA-256 hash value.
+         * @throws IllegalArgumentException The input string does not contain 64 hexadecimal digits.
```
shamsasari: It's not just that. `parseAsHex` also throws IllegalArgumentException

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -19,13 +19,30 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+         */
         @JvmStatic
         fun parse(str: String) = str.toUpperCase().parseAsHex().let {
```
shamsasari: I forgot to mention this before, can you convert this to a code block. The `it.size` is very confusing. At first glance it looks like it's comparing on `str`.

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -34,24 +51,53 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
             }
         }
 
+        /**
+         * Computes the SHA-256 hash value of the [ByteArray].
+         * @param bytes
```
shamsasari: ditto, and for the other methods

core/src/main/kotlin/net/corda/core/utilities/ByteArrays.kt
```diff
@@ -128,13 +131,35 @@ open class OpaqueBytes(override val bytes: ByteArray) : ByteSequence() {
         require(bytes.isNotEmpty())
     }
 
-    override val size: Int get() = bytes.size
-    override val offset: Int get() = 0
+    /* The bytes are always cloned so that this object becomes immutable. This has been done
```
shamsasari: /**

core/src/test/kotlin/net/corda/core/serialization/TransactionSerializationTests.kt
```diff
@@ -65,7 +66,10 @@ class TransactionSerializationTests : TestDependencyInjectionBase() {
         stx.verifyRequiredSignatures()
 
         // Corrupt the data and ensure the signature catches the problem.
-        stx.id.bytes[5] = stx.id.bytes[5].inc()
+        val bytesField = stx.id::bytes.javaField?.apply { setAccessible(true) }
```
shamsasari: Use `declaredField`

chrisr3: I don't understand. This is the Kotlin reflection API for accessing the backing field I need, and it works.

shamsasari: Fair enough. `declaredField` is very useful for removing the boilerplate when doing reflection. It may not be that useful here.

core/src/main/kotlin/net/corda/core/utilities/ByteArrays.kt
```diff
@@ -128,13 +131,37 @@ open class OpaqueBytes(override val bytes: ByteArray) : ByteSequence() {
...
+     */
+    override final val bytes: ByteArray = bytes
+        get() = field.clone()
+    override val size: Int = bytes.size
+    override val offset: Int = 0
 }
 
+/**
+ * Copy [size] bytes from this [ByteArray] starting from [offset] into a new [ByteArray].
+ * @param offset
```
shamsasari: I'm actually OK with no `@param` entries if the KDoc already explains what they do. That's better than having no description like this. So you can remove these empty descriptions.

core/src/test/kotlin/net/corda/core/serialization/TransactionSerializationTests.kt
chrisr3: The `declaredField` equivalent of `std.id::bytes.javaField` would be something like:
```kotlin
val bytes = declaredField<ByteArray>(stx.id::class.superclasses[0].superclasses[0], "bytes").value
```
I definitely prefer the Kotlin reflection version.

docs/source/changelog.rst
```diff
@@ -6,6 +6,8 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
+* ``OpaqueBytes.bytes`` now returns a clone of its underlying ``ByteArray``, and has been redeclared as ``final``.
```
shamsasari: Add that this is a slight API change, but required to make sure classes like SecureHash are immutable.

PR #1876, opened 2017-10-12
Revert javadocs cleanup as API change

cburlinchon:


chrisr3:
Only two of them need to be reverted.

PR #1878, opened 2017-10-12
Add AbstractPartyDescriptor to stop Hibernate warning

rnicoll:


node/src/main/kotlin/net/corda/node/services/persistence/HibernateConfiguration.kt
```diff
@@ -36,6 +37,8 @@ class HibernateConfiguration(val schemaService: SchemaService, private val datab
 
     init {
         logger.info("Init HibernateConfiguration for schemas: ${schemaService.schemaOptions.keys}")
+        // Register the AbstractPartyDescriptor so Hibernate doesn't warn
+        JavaTypeDescriptorRegistry.INSTANCE.addDescriptor(AbstractPartyDescriptor(createIdentityService))
```
josecoll: I don't quite understand why we are creating a 2nd duplicate custom converter implementation which does exactly the same as `AbstractPartyToX500NameAsStringConverter` which gets registered using:

```
            // register custom converters
            applyAttributeConverter(AbstractPartyToX500NameAsStringConverter(createIdentityService))registered
```

rnicoll: I've added a much longer in-line comment to explain the background on this, but basically Hibernate warns on the class not having a descriptor, but as far as I can tell it won't use the descriptor for conversion without providing a type for it, and that requires annotating every `AbstractParty` with `@Type`. It's all a bit absurd, unfortunately.

josecoll: Yes - but why do we need a second converter when the original one already performs the conversion? Does this new one get invoked again per request?

rnicoll:
Rebased to match latest master. So the rationale behind this is that Hibernate warns if it can't find a descriptor, even if there's a converter available. Unfortunately it won't use a descriptor by default, despite looking for one by default, without type annotations on everything, so we keep the converter in as well.

If there's a way I've missed, to convince Hibernate to use up the descriptor by default, do let me know :)

rnicoll:
@josecoll this needs a re-review, please

PR #1883, opened 2017-10-13
Attempting to make VerifierTests more stable

vkolomeyko:
The test `net.corda.verifier.VerifierTests#single verifier works with a node` is quite flaky in `master` at the moment.

Main point for improvement: Alice node will always be faster to locate Alice by name than NotaryNode. Especially given that it may take some time for Alice node to get reflected on the Notary's NetworkMap.

Also additional logging which might be helpful during debug and IntelliJ changed project structure reflection.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
@@ -883,14 +883,16 @@ class DriverDSL(
                 }
                 establishRpc(nodeConfiguration, processDeathFuture).flatMap { rpc ->
                     // Call waitUntilNetworkReady in background in case RPC is failing over:
-                    val networkMapFuture = executorService.fork {
+                    val forked = executorService.fork {
                         rpc.waitUntilNetworkReady()
-                    }.flatMap { it }
+                    }
+                    val networkMapFuture = forked.flatMap { it }
                     firstOf(processDeathFuture, networkMapFuture) {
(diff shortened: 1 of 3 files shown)


vkolomeyko:
Yes, I like idea for this utility method and I think it will make sense to create it and add it in places where `net.corda.testing.EventuallyKt#eventually` is used and nodes need to be able to locate/talk to each other.

But I would argue that change on this PR makes the test more stable and it would be worth merging it anyway to unblock progress of other people.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -706,7 +718,7 @@ class DriverDSL(
                         "webAddress" to webAddress.toString(),
                         "networkMapService" to networkMapServiceConfigLookup(name),
                         "useTestClock" to useTestClock,
-                        "rpcUsers" to if (rpcUsers.isEmpty()) defaultRpcUserList else rpcUsers.map { it.toMap() },
+                        "rpcUsers" to if (rpcUsers.isEmpty()) defaultRpcUserList else rpcUsers.map { it.toConfig().root().unwrapped() },
```
exFalso: Why this change?

vkolomeyko: Apparently `toMap` is a deprecated method - so I replaced the use of it as per recommendation.

PR #1886, opened 2017-10-16
Isolation level per transaction

mkit:
Allowing to override isolation level on per transaction basis.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/utilities/CordaPersistence.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/CordaPersistence.kt
@@ -40,10 +40,21 @@ class CordaPersistence(var dataSource: HikariDataSource, private val schemaServi
         }
     }
 
-    fun createTransaction(): DatabaseTransaction {
+    /**
+     * Creates an instance of [DatabaseTransaction], with the given isolation level.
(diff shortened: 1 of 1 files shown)


shamsasari:
I'll leave this to @szymonsztuka.

node/src/main/kotlin/net/corda/node/utilities/CordaPersistence.kt
```diff
@@ -40,10 +40,21 @@ class CordaPersistence(var dataSource: HikariDataSource, private val schemaServi
...
-        return DatabaseTransactionManager.currentOrNew(transactionIsolationLevel)
+        return DatabaseTransactionManager.currentOrNew(isolationLevel)
+    }
+
+    /**
+     * Creates an instance of [DatabaseTransaction], with the transaction isolation level specified at the creation time.
+     */
+    fun createTransaction(): DatabaseTransaction {
+        return createTransaction(transactionIsolationLevel)
     }
```
szymonsztuka: I think better to have one `createTransaction` method with a default argument ( value `transactionIsolationLevel`)

szymonsztuka: I forgot about backward compatibility issue, so this code is ok.

PR #1888, opened 2017-10-16
NotaryFlow get notary type

kasiastreich:
Get validating/non-validating type at the beginning of communication with the notary service

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -63,7 +63,7 @@ class NotaryFlow {
             val response = try {
                 progressTracker.currentStep = DISCOVERING
                 val session = initiateFlow(notaryParty)
-                val isValidating: Boolean = session.receive<Boolean>().unwrap { it }
+                val isValidating = session.receive<Boolean>().unwrap { it }
                 progressTracker.currentStep = REQUESTING
                 if (isValidating) {
                     subFlow(SendTransactionWithRetry(session, stx))
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -37,17 +37,18 @@ class NotaryFlow {
         constructor(stx: SignedTransaction) : this(stx, tracker())
 
         companion object {
+            object CHECKING_TX : ProgressTracker.Step("Checking validity of the transaction")
+            object DISCOVERING : ProgressTracker.Step("Discovering the type of Notary service: validating/non-validating")
             object REQUESTING : ProgressTracker.Step("Requesting signature by Notary service")
             object VALIDATING : ProgressTracker.Step("Validating response from Notary service")
 
-            fun tracker() = ProgressTracker(REQUESTING, VALIDATING)
+            fun tracker() = ProgressTracker(CHECKING_TX, DISCOVERING, REQUESTING, VALIDATING)
         }
 
         @Suspendable
         @Throws(NotaryException::class)
         override fun call(): List<TransactionSignature> {
-            progressTracker.currentStep = REQUESTING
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -60,8 +61,11 @@ class NotaryFlow {
             }
 
             val response = try {
+                progressTracker.currentStep = DISCOVERING
                 val session = initiateFlow(notaryParty)
-                if (serviceHub.networkMapCache.isValidatingNotary(notaryParty)) {
+                val isValidating: Boolean = session.receive<Boolean>().unwrap { it }
```
shamsasari: Don't need the type declaration on `isValidating`

node/src/main/kotlin/net/corda/node/services/transactions/BFTNonValidatingNotaryService.kt
```diff
@@ -42,6 +39,7 @@ class BFTNonValidatingNotaryService(override val services: ServiceHubInternal,
         private val log = loggerFor<BFTNonValidatingNotaryService>()
     }
 
+    override val isValidating: Boolean = false
```
shamsasari: `get()  = false`, and ditto for the others

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -116,7 +116,9 @@ interface NetworkMapCache {
     fun isNotary(party: Party): Boolean = party in notaryIdentities
 
     /** Checks whether a given party is an validating notary identity. */
-    // TODO This implementation will change after introducing of NetworkParameters.
+    // TODO We probably should remove that function, because in NetworkParameters we specify notaries without the type.
+    //  It could be problematic if we had more types of notaries.
+    @Deprecated("Notary type is checked at the beginning of a communication with the notary service, not based on NetworkMap data")
```
shamsasari: We're deprecating this, not because it's no longer used by the notary flow, but because we can no longer determine if an identity is a validating notary just based on the Party.

node/src/main/kotlin/net/corda/node/services/transactions/BFTNonValidatingNotaryService.kt
```diff
@@ -74,6 +72,7 @@ class BFTNonValidatingNotaryService(override val services: ServiceHubInternal,
     private class ServiceFlow(val otherSideSession: FlowSession, val service: BFTNonValidatingNotaryService) : FlowLogic<Void?>() {
         @Suspendable
         override fun call(): Void? {
+            otherSideSession.send(service.isValidating)
```
shamsasari: Shouldn't this class extend `NotaryFlow.Service`? Can you try and fix that in this PR

chrisr3:
This is breaking the API Stability build partly because you've annotated one method as `@Deprecated`. I think Mike is against us deprecating methods like this, c.f. his "Backwards Compatibility" email from 10th October: 
> That means that often we should not be marking things deprecated via the annotations even if they're old. It's sufficient to note in the docs that there's a better way to do things.

shamsasari:
@chrisr3 Well it's an interesting question. If we don't deprecate then we must lock down the network map design that a validating notary *must* have the string "validating" in its identity CN. That would be ironic since this was only suggested as a temporary hack to gets things moving as part of the network map redesign!

@kasiastreich This does need @mikehearn's input. For now assuming validating notaries must have "validating" in their CN.

kasiastreich:
@shamsasari other solutions are: have validating and non-validating notaries in NetworkParameters, or have some metadata for notaries. Having "validating" in name isn't the requirement. I think that the implementation of that method is irrelevant.

shamsasari:
Fair enough, I think then splitting the notary list into two makes most sense. Perhaps add a TODO there and link to this PR.

PR #1873, opened 2017-10-11
Cordformation in Kotlin

Clintonio:
Why: Brings Cordformation back in line with the rest of the Corda repo (ie; Kotlin or Java) and fixes some bugs on Windows relating to deployNodes. 

The changes may seem trivial but because gradle uses _dynamically typed_ groovy by default it ends up being much easier to maintain the Kotlin version than the groovy (I would know, I've spent more time than anyone with this part of the codebase).

PR #1890, opened 2017-10-16
[CORDA-442] let Driver run without network map

al-r3:
Let driver started nodes run without a network map.

- Nodes started by driver run without a networkMapNode.

- Driver does not take a networkMapStartStrategy anymore

- a new parameter in the configuration "noNetworkMapServiceMode" allows for a node not to be a networkMapNode nor to connect to one.

- Driver now waits for each node to write its own NodeInfo file to disk and then copies it into each other node.

- When driver starts a node N, it waits for *every* node to be have N nodes in their network map.

Note: the code to copy around the NodeInfo files was already in DemoBench, the NodeInfoFilesCopier class was just moved from DemoBench into core (I'm very open to core not being the best place, please advise)


core/src/main/kotlin/net/corda/core/utilities/NodeInfoFilesCopier.kt
```diff
@@ -57,9 +59,9 @@ class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller()
...
     @Synchronized
-    fun removeConfig(nodeConfig: NodeConfigWrapper) {
-        nodeDataMap.remove(nodeConfig.nodeDir) ?: return
-        log.info("Stopped watching: ${nodeConfig.nodeDir}")
+    fun removeConfig(nodeDir: Path) {
+        nodeDataMap.remove(nodeDir) ?: return
+        log.info("Stopped watching: $nodeDir")
     }
 
     @Synchronized
```
shamsasari: `allPreviouslySeenFiles()` below can use `flatMap` instead

core/src/main/kotlin/net/corda/core/utilities/NodeInfoFilesCopier.kt
```diff
@@ -34,20 +36,20 @@ class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller()
...
-    fun addConfig(nodeConfig: NodeConfigWrapper) {
-        val newNodeFile = NodeData(nodeConfig.nodeDir)
-        nodeDataMap[nodeConfig.nodeDir] = newNodeFile
+    fun addConfig(nodeDir: Path) {
+        val newNodeFile = NodeData(nodeDir)
+        nodeDataMap[nodeDir] = newNodeFile
 
         for (previouslySeenFile in allPreviouslySeenFiles()) {
             copy(previouslySeenFile, newNodeFile.destination.resolve(previouslySeenFile.fileName))
```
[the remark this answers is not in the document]
al-r3: I don't understand this comment, copy is a function I've defined below and does a copy with some additional twists.

Does this mean I should rename that to copyTo and make it an extension function of path?

al-r3:
I've added a NODE_INFO_FILE_NAME_PREFIX in NodeInfoFilesCopier to for the "nodeInfo-" constant.
I'll move it to CordformNode (or somethere in the same package) once I can also apply the change to the gradle plug-in which is now being rewritten in Kotlin: https://github.com/corda/corda/pull/1873

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -846,28 +816,65 @@ class DriverDSL(
...
+    /**
+     * @param initial number of nodes currently in the network map of a running node.
+     * @param observable an observable returning the updates to the node network map.
+     * @return a [ConnectableObservable] which emits a new [Int] every time the number of registered nodes changes
+     *   the initial value emitted is always [initial]
+     */
+    private fun countNodesObservable(initial: Int, observable: Observable<NetworkMapCache.MapChange>):
+            ConnectableObservable<Int> {
+        var count = initial
+        return observable.map { it ->
```
shamsasari: It looks like what you need here is `Observable.collect`

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -846,28 +816,65 @@ class DriverDSL(
...
-                        "useTestClock" to useTestClock)
-        )
-        return startNodeInternal(config, webAddress, startInProcess, maximumHeapSize)
+    /**
+     * @param initial number of nodes currently in the network map of a running node.
+     * @param observable an observable returning the updates to the node network map.
+     * @return a [ConnectableObservable] which emits a new [Int] every time the number of registered nodes changes
+     *   the initial value emitted is always [initial]
+     */
+    private fun countNodesObservable(initial: Int, observable: Observable<NetworkMapCache.MapChange>):
```
shamsasari: The param name `observable` is too vague in this context.

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
```diff
@@ -16,15 +16,20 @@ import java.nio.file.StandardCopyOption.REPLACE_EXISTING
...
  * don't exist yet.
  */
-class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() {
+class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()) {
+
+    companion object {
+        private val log = loggerFor<NodeInfoFilesCopier>()
+    }
 
     private val nodeDataMap = mutableMapOf<Path, NodeData>()
```
shamsasari: Can you use a `ThreadBox` around this, instead of synchronisation

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
```diff
@@ -16,15 +16,20 @@ import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.attribute.FileTime
 import java.util.concurrent.TimeUnit
-import java.util.logging.Level
+
+const val NODE_INFO_FILE_NAME_PREFIX = "nodeInfo-"
```
shamsasari: Better to put it inside the companion object below

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -211,13 +207,15 @@ sealed class NodeHandle {
             override val configuration: FullNodeConfiguration,
             override val webAddress: NetworkHostAndPort,
             val debugPort: Int?,
-            val process: Process
+            val process: Process,
+            val action: () -> Unit
```
shamsasari: make this private and rename

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -228,14 +226,16 @@ sealed class NodeHandle {
             override val configuration: FullNodeConfiguration,
             override val webAddress: NetworkHostAndPort,
             val node: StartedNode<Node>,
-            val nodeThread: Thread
+            val nodeThread: Thread,
+            val action: () -> Unit
```
shamsasari: ditto

tools/demobench/src/main/kotlin/net/corda/demobench/model/DemoBenchNodeInfoFilesCopier.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    private val nodeInfoFilesCopier = NodeInfoFilesCopier(scheduler)
+
+    /**
+     * @param nodeConfig the configuration to be added.
+     * Add a [NodeConfig] for a node which is about to be started.
+     * Its nodeInfo file will be copied to other nodes' additional-node-infos directory, and conversely,
+     * other nodes' nodeInfo files will be copied to this node additional-node-infos directory.
+     */
+    fun addConfig(nodeConfig: NodeConfigWrapper) = nodeInfoFilesCopier.addConfig(nodeConfig.nodeDir)
```
shamsasari: Specify return types on public methods

core/src/main/kotlin/net/corda/core/utilities/NodeInfoFilesCopier.kt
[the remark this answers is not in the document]
shamsasari: I see. It looked you were using `Files.copy`. Can you rename your method to `atomicCopy`. Why doesn't it use the `ATOMIC_MOVE` copy option?

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
al-r3: Why? It doesn't make any difference as far as I get it. 
Also it's not really a part of the class NodeInfoFilesCopier per se, it's a shared concept between a few classes.

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
al-r3: That wouldn't be enough I have methods iterating over it while other might be adding/removing elements, a ThreadBox is just a lock around the map object.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -833,9 +805,7 @@ class DriverDSL(
     override fun start() {
         _executorService = Executors.newScheduledThreadPool(2, ThreadFactoryBuilder().setNameFormat("driver-pool-thread-%d").build())
         _shutdownManager = ShutdownManager(executorService)
-        if (networkMapStartStrategy.startDedicated) {
-            startDedicatedNetworkMapService().andForget(log) // Allow it to start concurrently with other nodes.
-        }
+        // TODO
```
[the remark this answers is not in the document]
al-r3: I forgot it there, sorry.

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
shamsasari: Kotlin generates a `NodeInfoFilesCopierKt` class to put it in, so it makes sense to put it in the companion object and keep things "clean".

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
shamsasari: Use the `locked` method

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: collect seem to have been renamed reduce: http://reactivex.io/documentation/operators/reduce.html
That does give a single value which is a combination of all the values from all observables, and you're not able to tell which observable produced which value.

I can't use that because:
- I can't wait for netMap observables to be completed, they never are
- I need to know which observable does each value comes from

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -846,28 +816,65 @@ class DriverDSL(
...
+        val requiredNodes = countObservables.size
+        val future = openFuture<Unit>()
+
+        // This is an observable which yield the minimum number of nodes in each node network map.
+        val latest = Observable.combineLatest(countObservables.values.toList(), { args ->
+            val ints = args.map { it as Int }
+            return@combineLatest ints.min() ?: 0
+        })
+
+        latest.subscribe(object : Subscriber<Int>() {
```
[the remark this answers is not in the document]
al-r3: It should, though .toFuture unsubscribes from the observable if the future is cancelled (which is fine for me)
Though it does not unsubscribe after the first emission, which seems the correct behavior in general, as far as I get it.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: I'm referring to collect, not reduce: http://reactivex.io/RxJava/1.x/javadoc/rx/Observable.html#collect-rx.functions.Func0-rx.functions.Action2-

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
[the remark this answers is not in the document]
shamsasari: It uses `first()` which does unsubscribe after the first emission.

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
al-r3: moved

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
al-r3: If I do that, I'll need to wrap all the methods in a lambda and the only difference will be that the lock I'm using is the one in the threadbox VS the one object monitor in the class instance, I'm not sure I see the benefit

core/src/main/kotlin/net/corda/core/utilities/NodeInfoFilesCopier.kt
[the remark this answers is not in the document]
al-r3: My understanding of ATOMIC_MOVE is that it just makes the move fail if the file system does not support an atomic move.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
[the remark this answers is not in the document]
al-r3: ok - then that works, I needed to add a 'funny' .map{ } at the end to make that a future<Unit> otherwise it's a future<Int> and there's not much point in returning a value

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: ok googling for observables reduce returns' obaservables collect'

Still I'm not sure how are you suggesting to use collect, that just returns lists with previous values of an observable, here I need to combine the values returned by a set of observables

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
shamsasari: The benefit is that it makes it harder to forget (for the next developer) to use locks/synchronisation. It's an aid to prevent bugs, at the small expense of a few more characters. ThreadBox is used throughout the codebase.

core/src/main/kotlin/net/corda/core/utilities/NodeInfoFilesCopier.kt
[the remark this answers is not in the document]
shamsasari: If you're not going to use it then what's the purpose of the method? What are you trying to avoid with all this copying? Seeing half-copied files? Is that something we've come across or even likely to? And is it even something we care about considering the context we're dealing with?

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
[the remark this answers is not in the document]
shamsasari: Hmm, perhaps have the method return `CordaFuture<*>`. Either way is fine.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: This is what I came up with. I haven't tested it though.
```
       return networkMapCacheChangeObservable.collect({ AtomicInteger(initial) }) { count, change ->
            when (change) {
                is NetworkMapCache.MapChange.Added -> count.incrementAndGet()
                is NetworkMapCache.MapChange.Removed -> count.decrementAndGet()
            }
        }.map(AtomicInteger::get).replay()
```
You might need to add `startsWith` as well.

al-r3: Oh that.
Out of the box it doesn't (even with the .startWith).
Though what's wrong in the current implementation? Seems it's doing something relatively straightforward.

shamsasari: If `collect` works then it's much easier to reason with. Also, can multiple threads emit values on `networkMapCacheChangeObservable`? If so then your `count` variable needs additional synchronisation. If there's a primitive that's designed for what you're trying to do, why not use it?

core/src/main/kotlin/net/corda/core/utilities/NodeInfoFilesCopier.kt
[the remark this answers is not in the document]
al-r3: The purpose it to try to avoid seeing half-copied files.
I've not seen that, but I can't be sure it won't happen. The current shape of the code came to be in the review of this code for DemoBench, where it was originally, with Cris R.

I can do a move with ATOMIC_MOVE, catch AtomicMoveNotSupportedException, and then try to an non-necessarily-atomic move.
I'd expect code written like that to be equivalent to the current implementation.

As whether we care about in this context, I'd say no. But for code reuse sake I've taken the code which did that for demobench, since it's exactly the same behavior we want.

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
al-r3: Switched to ThreadBox,

core/src/main/kotlin/net/corda/core/utilities/NodeInfoFilesCopier.kt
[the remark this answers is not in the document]
shamsasari: It's just that I don't think this is necessary, especially if the node is already dealing with half-copied files. Anyway, I won't belabour the point.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: I'm not sure if different threads can call this code, I don't know the code and observables well enough to tell. (I've added some logging and it doesn't seem it happens, but that doesn't mean much)
I've changed the code to use an AtomicInteger. 
As whether collect is much easier to reason with, I don't see it. collect is a more complex operation than map, it doesn't save you from concurrency issues. Overall what this transformation does is fairly simple, however you write it.

core/build.gradle
```diff
@@ -21,6 +21,8 @@ dependencies {
     // Guava: Google test library (collections test suite)
     testCompile "com.google.guava:guava-testlib:$guava_version"
 
+    compile "net.corda.plugins:cordform-common:$gradle_plugins_version"
```
shamsasari: Is this needed?

node-api/src/main/kotlin/net/corda/nodeapi/NodeInfoFilesCopier.kt
```diff
@@ -16,15 +16,19 @@ import java.nio.file.StandardCopyOption.REPLACE_EXISTING
...
-import java.util.logging.Level
 
 /**
  * Utility class which copies nodeInfo files across a set of running nodes.
  *
  * This class will create paths that it needs to poll and to where it needs to copy files in case those
  * don't exist yet.
  */
-class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()): Controller() {
+class NodeInfoFilesCopier(scheduler: Scheduler = Schedulers.io()) {
```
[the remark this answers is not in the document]
al-r3: I've added a comment on the instantiation of NodeInfoFilesCopier in DriverDSL. These delays are fine for the DemoBench use case.

core/build.gradle
al-r3: Almost. I've moved it into node-api.
It might "leak" to node one day, this is here because it contains the definition of NODE_INFO_DIRECTORY which is shared with the gradle plug-in

shamsasari: I don't understand, is the gradle plugins dependency going to be moved to node-api?

al-r3: yes (it's not the whole plug-in, it's just cordform-common which is a subset of it).
This is needed since they both need NODE_INFO_DIRECTORY which is in CordformNode

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -846,28 +820,50 @@ class DriverDSL(
...
+    /**
+     * @param rpc the [CordaRPCOps] of a newly started node.
+     * @return a [CordaFuture] which resolves when every node started by driver has in its network map a number of nodes
+     *   equal to the number of running nodes. The future will yield the number of connected nodes.
+     */
+    private fun waitForNodes(rpc: CordaRPCOps): CordaFuture<Int> {
+        val (snapshot, updates) = rpc.networkMapFeed()
+        val counterObservable = nodeCountObservable(snapshot.size, updates)
+        countObservables.put(rpc.nodeInfo().legalIdentities.first().name, counterObservable)
+        val requiredNodes = countObservables.size
```
shamsasari: Please add TODO as discussed.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -846,28 +820,50 @@ class DriverDSL(
...
+    private fun nodeCountObservable(initial: Int, networkMapCacheChangeObservable: Observable<NetworkMapCache.MapChange>):
+            ConnectableObservable<Int> {
+        val count = AtomicInteger(initial)
+        return networkMapCacheChangeObservable.map { it ->
+            when (it) {
+                is NetworkMapCache.MapChange.Added -> count.incrementAndGet()
+                is NetworkMapCache.MapChange.Removed -> count.decrementAndGet()
+                is NetworkMapCache.MapChange.Modified -> count.get()
+            }
+        }.startWith(initial).replay()
```
shamsasari: You're not connecting to the ConnectableObservable. Or is the `replay()` not needed?

al-r3: I missed the connect in a review pass

PR #1895, opened 2017-10-17
Attempt to make NodeStatePersistenceTests more stable

vkolomeyko:
By ensuring that the nodes are properly started and aware of each other before firing any flows through them.
Also minor refactoring.


vkolomeyko:
I have executed CI Integration tests 11 times and they all passed.
@sollecitom @Clintonio and @kasiastreich - over to you for review and approval.

PR #1897, opened 2017-10-17
Plumbing code for system tests

mat-rizzo-r3:
Please review this code for launching nodes in a similar way to NodeRunner. This plumbing code will be reusable for all the system tests.


samples/trader-demo/build.gradle
```diff
@@ -39,6 +48,9 @@ dependencies {
     testCompile project(':test-utils')
     testCompile "junit:junit:$junit_version"
     testCompile "org.assertj:assertj-core:${assertj_version}"
+
+    // https://mvnrepository.com/artifact/net.sf.expectit/expectit-core
+    systemTestCompile group: 'net.sf.expectit', name: 'expectit-core', version: "${expectit_version}"
```
andr3ej: can you use the colon-separated notation for consistency

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+//
+//        var result = expect
+//            .withTimeout(60, TimeUnit.SECONDS)
+//            .expect(Matchers.regexp("(?m)^Node for \"Bank A\" started up and registered in .*"))
+//
+//        println(result!!.input
+//
+//        println(result.isSuccessful)
+////        Node for "Bank A" started up and registered in 5.59 sec
+////        Node for "Bank A" started up and registered in 5.59 sec
```
andr3ej: what's going on here?

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+import java.io.FileNotFoundException
+import java.lang.UnsupportedOperationException
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.util.Locale
+import java.util.concurrent.TimeUnit
+import javax.naming.ConfigurationException
+
+class TraderDemoSystemTest {
+
```
andr3ej: r3 coding style says to minimise blank lines, such as this one (unless the formatter puts them back)

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+    private var _bankA: NodeExpectingProcess? = null
+    private var _bankB: NodeExpectingProcess? = null
+    private var _bankOfCorda: NodeExpectingProcess? = null
+    private var _notaryService: NodeExpectingProcess? = null
```
andr3ej: these can be lateinit, and we only use the underline prefix in rare cases e.g.
```
private var _foo: SpecificTypeImpl
val foo: MoreGeneralType get() = _foo
```
please fix all such names

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+        return NodeCommandBuilder()
+                .withNodeFileInfo(nodeFileInfo)
+                .build()
+    }
+
+    private val _nodesDirectory: File by lazy {
+        Paths
+                .get(System.getProperty("user.dir"), "build", "nodes")
+                .toFile()
+    }
```
andr3ej: why is this lazy?

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+
+private class NodeExpectingProcess(private val _expectingProcess: ExpectingProcess) {
+
+    fun expectCordaLogo(): Result {
+        val pattern = "?(m)" +
+                """^\Q   ______               __ \E.*$\r\n?|\n""" +
+                """^\Q  / ____/     _________/ /___ _\E.*$\r\n?|\n""" +
+                """^\Q / /     __  / ___/ __  / __ `/\E.*$\r\n?|\n""" +
+                """^\Q/ /___  /_/ / /  / /_/ / /_/ /\E.*$\r\n?|\n""" +
+                """^\Q\____/     /_/   \__,_/\__,_/\E.*$\r\n?|\n"""
```
andr3ej: presumably this is to synchronize with the output, rather than to enforce particular output? in which case this will break if/when the logo changes. can this (and the writer) refer to a constant rather hardcoding the logo. similarly for the other expect methods

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+        _process
+                .destroyForcibly()
+                .waitFor(5, TimeUnit.SECONDS)
+    }
+}
+
+private class ExpectingProcessBuilder {
+
+    private var _command: List<String>? = null
+    private var _workingDirectory: File? = null
```
andr3ej: these should be non-nullable constructor params, assuming start will inevitably be called

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+
+    private fun validateConfigOrThrow(configFile: File): Unit {
+        val isConfigValid = Files
+                .lines(configFile.toPath())
+                .anyMatch { "webAddress" in it }
+        if (!isConfigValid) {
+            throw ConfigurationException("Node config does not contain webAddress.")
+        }
+    }
+}
```
andr3ej: this stuff looks familiar. can it be unduplicated w.r.t. NodeRunner?

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+                add("--no-local-shell")
+            }
+            addAll(arguments())
+        }
+    }
+
+    private fun directory(): File = _nodeFileInfo!!.directory()
+    private fun directoryName(): String = _nodeFileInfo!!.directoryName()
+    private fun jarFile(): File = _nodeFileInfo!!.jarFile()
+    private fun jarName(): String = _nodeFileInfo!!.jarName()
```
andr3ej: these bangs tell me nodeFileInfo probably shouldn't be nullable

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+            add("-jar")
+            add(jarName())
+            addAll(jarArguments())
+        }
+    }
+
+    private val _javaPathString: String by lazy {
+        Paths
+                .get(System.getProperty("java.home"), "bin", "java")
+                .toString()
```
andr3ej: there's some code somewhere to do this that works in both jdk and jre cases

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+    else
+        throw UnsupportedOperationException("OS not supported.")
+}
+
+private object debugPortNumberGenerator {
+
+    private var _portNumber: Int = 5005
+
+    fun next(): Int = _portNumber++
+}
```
andr3ej: this class also looks very familiar

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,457 @@
...
+
+private fun File.existsOrThrow(): Unit {
+    if (!this.exists()) {
+        throw FileNotFoundException("${this.path} does not exist.")
+    }
+}
+
+private fun Any?.expectNullOrThrow(name: String = "Nullable"): Unit {
+    this?.let { throw IllegalArgumentException("$name has already been set to a value.") }
```
andr3ej: use run rather than let if you're not going to use "it"

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
mat-rizzo-r3: line 18 or line 20 ?

samples/trader-demo/build.gradle
mat-rizzo-r3: yes

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
andr3ej: 20 and lots more. easiest to remove all blanks and run the formatter

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -101,6 +101,11 @@ fun <T> List<T>.indexOfOrThrow(item: T): Int {
 fun Path.createDirectory(vararg attrs: FileAttribute<*>): Path = Files.createDirectory(this, *attrs)
 fun Path.createDirectories(vararg attrs: FileAttribute<*>): Path = Files.createDirectories(this, *attrs)
 fun Path.exists(vararg options: LinkOption): Boolean = Files.exists(this, *options)
+fun Path.existsOrThrow(vararg options: LinkOption) {
```
shamsasari: `ensureExists`?

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -303,3 +308,27 @@ fun TransactionBuilder.toLedgerTransaction(services: ServiceHub, serializationCo
 
 /** Convenience method to get the package name of a class literal. */
 val KClass<*>.packageName get() = java.`package`.name
+
+/**
+ * Throws an [IllegalArgumentException] if the [value] is not null.
+ * Provides the converse of the requireNotNull kotlin builtin. See kotlin-stdlib/kotlin/util/Preconditions.kt .
+ */
+inline fun <T:Any> requireNull(value: T?) = requireNull(value) { "Required value was not null." }
```
shamsasari: single space around the colon: `T : Any`, but in fact you don't even need the type parameter as it's not used. `value` can simply be `Any?`

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -303,3 +308,27 @@ fun TransactionBuilder.toLedgerTransaction(services: ServiceHub, serializationCo
...
+
+/**
+ * Throws an [IllegalArgumentException] with the result of calling [lazyMessage] if the [value] is null.
+ * Provides the converse of the requireNotNull kotlin builtin. See kotlin-stdlib/kotlin/util/Preconditions.kt .
+ */
+inline fun <T:Any> requireNull(value: T?, lazyMessage: () -> Any) = require(value == null) { lazyMessage() }
```
shamsasari: You can pass in `lazyMessage` directly: `require(value == null, lazyMessage)`

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,391 @@
...
+    @After
+    fun stopNodes(): Unit {
+        _bankA?.close()
+        _bankB?.close()
+        _bankOfCorda?.close()
+        _notaryService?.close()
+    }
+
+    @Test
+    fun `runs trader demo system test`(): Unit {
```
shamsasari: This test is not actually doing anything.

shamsasari: No need to specify `Unit` for the method return type

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
shamsasari: Yeah I don't think the logo is that important. The other expectations you have are sufficient I think.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
mat-rizzo-r3: I don't think that's a good idea, as `ensureExists` generally tends to mean ensure the path is created, and if it isn't created then create it. `existsOrThrow` is a bit more explicit in terms of what the function is doing.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
mat-rizzo-r3: requireNotNull is defined in the kotlin stdlib in this way (definition and file path included below), is there any particular reason we would do it differently?

`kotlin-stdlib/1.1.50/6649061a364e109349b4d1fb644b4ae14ee21905/kotlin-stdlib-1.1.50-sources.jar!/kotlin/util/Preconditions.kt`
```
/**
 * Throws an [IllegalArgumentException] if the [value] is null. Otherwise returns the not null value.
 */
@kotlin.internal.InlineOnly
public inline fun <T:Any> requireNotNull(value: T?): T = requireNotNull(value) { "Required value was null." }
```

shamsasari: Because `requireNotNull` is returning`T`

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
shamsasari: Fair enough

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
mat-rizzo-r3: Commented as discussed.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
mat-rizzo-r3: fair point.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -101,6 +101,11 @@ fun <T> List<T>.indexOfOrThrow(item: T): Int {
 fun Path.createDirectory(vararg attrs: FileAttribute<*>): Path = Files.createDirectory(this, *attrs)
 fun Path.createDirectories(vararg attrs: FileAttribute<*>): Path = Files.createDirectories(this, *attrs)
 fun Path.exists(vararg options: LinkOption): Boolean = Files.exists(this, *options)
+fun Path.existsOrThrow(vararg options: LinkOption) {
+    if (!this.exists(*options)) {
```
andr3ej: redundant this

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -101,6 +101,11 @@ fun <T> List<T>.indexOfOrThrow(item: T): Int {
...
+        throw IOException("$this does not exist.")
```
andr3ej: FileNotFoundException is more idiomatic, the jdk is very fond of it

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -101,6 +101,11 @@ fun <T> List<T>.indexOfOrThrow(item: T): Int {
...
+    }
+}
```
andr3ej: the formatter says insert blank line after this one

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
andr3ej: further to what i was saying earlier, going forwards please push code like this along with something/anything that uses it. otherwise someone may feel like deleting this method.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -303,3 +308,27 @@ fun TransactionBuilder.toLedgerTransaction(services: ServiceHub, serializationCo
...
+ * Throws an [IllegalArgumentException] with the result of calling [lazyMessage] if the [value] is null.
+ * Provides the converse of the requireNotNull kotlin builtin. See kotlin-stdlib/kotlin/util/Preconditions.kt .
+ */
+inline fun requireNull(value: Any?, lazyMessage: () -> Any) = require(value == null, lazyMessage)
+
+/**
+ * Throws an [IllegalStateException] if the [value] is false.
+ * Provides the converse of the checkNotNull kotlin builtin. See kotlin-stdlib/kotlin/util/Preconditions.kt .
+ */
+inline fun <T:Any> checkNull(value: T?) = checkNull(value) { "Required value was not null." }
```
andr3ej: please run formatter on new code

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -303,3 +308,27 @@ fun TransactionBuilder.toLedgerTransaction(services: ServiceHub, serializationCo
 
 /** Convenience method to get the package name of a class literal. */
 val KClass<*>.packageName get() = java.`package`.name
+
+/**
+ * Throws an [IllegalArgumentException] if the [value] is not null.
+ * Provides the converse of the requireNotNull kotlin builtin. See kotlin-stdlib/kotlin/util/Preconditions.kt .
+ */
+inline fun requireNull(value: Any?) = requireNull(value) { "Required value was not null." }
```
andr3ej: message confusing - did we require the value to be null, or did we require the value?

samples/trader-demo/build.gradle
```diff
@@ -39,6 +48,10 @@ dependencies {
     testCompile project(':test-utils')
     testCompile "junit:junit:$junit_version"
     testCompile "org.assertj:assertj-core:${assertj_version}"
+
+
```
andr3ej: why so many blanks

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+
+    }
+
+    @Test
+    fun `runs trader demo system test`() {
+
+    }
+}
+
+private object cordaMatchers {
```
andr3ej: objects are UpperCase

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
+package net.corda.traderdemo
```
andr3ej: @file:Suppress("unused") to see the real warnings

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+    private fun workingDirectoryAsFile(): File = workingDirectory.toFile()
+}
+
+private abstract class JarType(private val nodeDirectory: Path, private val jarName: String) {
+    fun directory(): Path = nodeDirectory
+    fun directoryName(): String = nodeDirectory.fileName.toString()
+    fun jarName(): String = jarName
+    fun jarFile(): Path = nodeDirectory.resolve(jarName())
+    open fun validateFilesOrThrow() {
+        throw NotImplementedError()
```
andr3ej: probably more idiomatic to throw this via the TODO() function

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+    protected fun validateFilesExistOrThrow() {
+        nodeDirectory.existsOrThrow()
+        jarFile().existsOrThrow()
+        configFile().existsOrThrow()
+    }
+
+    protected fun configFile(): Path = nodeDirectory.resolve("node.conf")
+}
+
+private class CordaJarType(private val nodeDirectory: Path) : JarType(nodeDirectory, "corda.jar") {
```
andr3ej: shouldn't be a val, superclass already puts it in a field

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+    override fun validateFilesOrThrow() = validateFilesExistOrThrow()
+}
+
+private class WebserverJarType(private val nodeDirectory: Path) : JarType(nodeDirectory, "corda-webserver.jar") {
```
andr3ej: same

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+        validateFilesExistOrThrow()
+        validateConfigOrThrow()
+    }
+
+    private fun validateConfigOrThrow() {
+        val isConfigValid = Files
+                .lines(configFile())
+                .anyMatch { "webAddress" in it }
+        if (!isConfigValid) {
+            throw ConfigurationException("Node config does not contain webAddress.")
```
andr3ej: i think you wanted to import a different ConfigurationException

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+        return when (os) {
+            OS.WINDOWS -> buildWindows()
+            OS.OSX -> buildOsx()
+            OS.LINUX -> buildLinux()
+        }
+    }
+
+    private fun buildWindows(): List<String> {
+        return mutableListOf<String>().apply {
+            TODO() // TODO test this and also apply working directory
```
andr3ej: this blows up with NotImplementedError without a message, is that intentional?

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+        return mutableListOf<String>().apply {
+            add("/bin/bash")
+            add("-c")
+            add("cd $workingDirectory ; ${commandString()} && exit")
+        }
+    }
+
+    private fun buildLinux(): List<String> {
+        return mutableListOf<String>().apply {
+            TODO() // TODO Implement this
```
andr3ej: similarly, surely the message should be an arg to the TODO function

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
```diff
@@ -0,0 +1,427 @@
...
+        throw UnsupportedOperationException("OS not supported.")
+}
+
+private object debugPortNumberGenerator {
+    private var portNumber: Int = 5005
+    fun next(): Int = portNumber++
+}
+
+private fun FluentRe.compile(flags: Int): Pattern {
+    val patternString = this
```
andr3ej: redundant this, and maybe a comment outlining that we're converting to a jdk regex

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
mat-rizzo-r3: Suggestions?
1) Required value was non-null.
2) Required value was expected to be null.
3) Required value was not null as expected.
4) Required value was expected as null.
5) Required null was value.
6) Required null, got a value.

open to other suggestions.

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
mat-rizzo-r3: That isn't consistent with `debugPortNumberGenerator` as used in NodeRunner which uses lowerCamelCase. So does that mean that the convention used in NodeRunner needs correcting to CamelCase as well ?

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
mat-rizzo-r3: intentional for now, can apply the message into the todo function if you'd prefer.

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
mat-rizzo-r3: hmm. Do you have ideas for which one you'd prefer? we have one in `net.corda.node.internal.ConfigurationException` but it doesn't really have a description of what it does. Also, we're supposed to be using system exceptions, so I'd rather not use this which derives from a custom exception type.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
andr3ej: i like 6

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
andr3ej: yes, NodeRunner is wrong but you don't have to fix it now

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
andr3ej: javax.naming.ConfigurationException suggests that javax.naming is complaining, so definitely not that one. you'd expect the node one to have been thrown by the node, so not that one either. it's hard to know what the right answer is without knowing what's going to call this method. it's fine to make a completely custom exception.

samples/trader-demo/src/system-test/kotlin/net/corda/traderdemo/TraderDemoSystemTest.kt
andr3ej: yes, i would want to see the message without having to dig for it

PR #1899, opened 2017-10-18
Disable unstable test on Windows

vkolomeyko:
The test still proven to be unstable on Windows.
(of course it passes locally when executed in isolation)
More work is needed to establish the root cause.


andr3ej:
is there a jira ticket (or equivalent) with a list of all the unreliable tests that have been disabled? i'm concerned that many of them were enforcing important behaviour

vkolomeyko:
In the past I was raising Jiras for flaky tests, e.g.: https://r3-cev.atlassian.net/browse/CORDA-649
For that flaky test no Jira been raised.

Flaky tests can be seen at any time in TC using: https://ci-master.corda.r3cev.com/project.html?projectId=Corda_Build&tab=flakyTests&branch_Corda_Build=__all_branches__
Recently changes been made to battle flakiness in all the tests mentioned there.

If we are to raise Jiras for flaky test, I think we need to have 1 Jira per 1 flaky test/class. As different tests are largely unrelated and could be flaky due to various reasons.

PR #1903, opened 2017-10-18
CORDA-704: Implement `@DoNotImplement` annotation

chrisr3:
Some public APIs are not intended to be extended. Annotate these as `@DoNotImplement` and enhance the API Stability check to ignore such classes when warning about new abstract API methods.

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/RPCConnection.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/RPCConnection.kt
@@ -1,5 +1,6 @@
 package net.corda.client.rpc
 
+import net.corda.core.DoNotImplement
 import net.corda.core.messaging.RPCOps
 import java.io.Closeable
 
@@ -10,6 +11,7 @@ import java.io.Closeable
  * [Closeable.close] may be used to shut down the connection and release associated resources. It is an
  * alias for [notifyServerAndClose].
  */
+@DoNotImplement
 interface RPCConnection<out I : RPCOps> : Closeable {
     /**
      * Holds a synthetic class that automatically forwards method calls to the server, and returns the response.
--- a/core/src/main/kotlin/net/corda/core/messaging/RPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/RPCOps.kt
@@ -1,9 +1,12 @@
 package net.corda.core.messaging
 
+import net.corda.core.DoNotImplement
+
 /**
  * Base interface that all RPC servers must implement. Note: in Corda there's only one RPC interface. This base
  * interface is here in case we split the RPC system out into a separate library one day.
  */
+@DoNotImplement
 interface RPCOps {
     /** Returns the RPC protocol version. Exists since version 0 so guaranteed to be present. */
     val protocolVersion: Int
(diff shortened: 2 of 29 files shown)


core/src/main/kotlin/net/corda/core/concurrent/CordaFuture.kt
```diff
@@ -1,12 +1,14 @@
...
 
+import net.corda.core.DoNotImplement
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.Future
 
 /**
  * Same as [Future] with additional methods to provide some of the features of [java.util.concurrent.CompletableFuture] while minimising the API surface area.
  * In Kotlin, to avoid compile errors, whenever CordaFuture is used in a parameter or extension method receiver type, its type parameter should be specified with out variance.
  */
+@DoNotImplement
```
mnesbit: They might legitimately need to implement this I am afraid e.g. to do Future combining

core/src/main/kotlin/net/corda/core/messaging/RPCOps.kt
```diff
@@ -1,9 +1,12 @@
 package net.corda.core.messaging
 
+import net.corda.core.DoNotImplement
+
 /**
  * Base interface that all RPC servers must implement. Note: in Corda there's only one RPC interface. This base
  * interface is here in case we split the RPC system out into a separate library one day.
  */
+@DoNotImplement
```
mnesbit: I am still a believer that we may end up with RPC tear-offs for pluggable RPC, so not sure this should be annotated, although it is true at teh minute.

core/src/main/kotlin/net/corda/core/node/services/UniquenessProvider.kt
```diff
@@ -12,6 +13,7 @@ import net.corda.core.serialization.CordaSerializable
  *
  * A uniqueness provider is expected to be used from within the context of a flow.
  */
+@DoNotImplement
```
mnesbit: Would they need to implement a custom UniquenessProvider as part of the pluggable notary framework?

client/rpc/src/main/kotlin/net/corda/client/rpc/RPCConnection.kt
```diff
@@ -10,6 +11,7 @@ import java.io.Closeable
  * [Closeable.close] may be used to shut down the connection and release associated resources. It is an
  * alias for [notifyServerAndClose].
  */
+@DoNotImplement
```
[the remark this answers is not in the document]
chrisr3: We wouldn't generate any KDoc for anything in `.internal.` package.

chrisr3:
This PR is motivated by "API Stability". TeamCity needs to be able to identify the interfaces and abstract classes that we can legitimately extend ourselves without risk of breaking other people's CorDapps, and then not block the PR if we add a new methods. We obviously still cannot modify or remove any existing methods from such interfaces.

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -13,6 +14,7 @@ import net.corda.core.utilities.loggerFor
 import org.slf4j.Logger
 import java.security.PublicKey
 
+@DoNotImplement
```
shamsasari: This can be implemented by CorDapps if they're writing their own notary service (an experimental feature).

PR #1904, opened 2017-10-18
CORDA-553 - First steps towards evolvability

fenryka:
Define the two transforms that will be useful for enum evolvability (see
the design document for more details).

Furthermore, define the generic mechanism by which transform annotations
on classes are encoded into the AMQP envelope

With nothing to check for these annotations at either end, this is
mostly a no-op, but an important step toward getting evolvability in
place


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
```diff
@@ -0,0 +1,24 @@
+package net.corda.nodeapi.internal.serialization.amqp
+
+import org.apache.qpid.proton.amqp.UnsignedLong
+
+// TODO: get an assigned number as per AMQP spec
+const val DESCRIPTOR_TOP_32BITS: Long = 0xc0da0000
+
+enum class AMQPDescriptorRegistry(val id: Long) {
+
```
rick-r3: Remove blank line

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
```diff
@@ -0,0 +1,24 @@
+package net.corda.nodeapi.internal.serialization.amqp
+
+import org.apache.qpid.proton.amqp.UnsignedLong
+
+// TODO: get an assigned number as per AMQP spec
+const val DESCRIPTOR_TOP_32BITS: Long = 0xc0da0000
```
rick-r3: We now have one of these allocated.  50530

See https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers

core/src/main/kotlin/net/corda/core/serialization/CordaSerializationTransformEnumDefault.kt
```diff
@@ -0,0 +1,18 @@
+package net.corda.core.serialization
+
+/**
+ *
```
rick-r3: Want to add at least introductory kdocs ?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SupportedTransforms.kt
```diff
@@ -0,0 +1,66 @@
...
+import net.corda.core.serialization.CordaSerializationTransformRenames
+import net.corda.core.serialization.CordaSerializationTransformRename
+
+/**
+ * Utility class that defines an instance of a transform we support
+ *
+ * @property type The transform annotation
+ * @property enum Maps the annotaiton onto a transform type, we expect there are multiple annotations that
+ * would map to a single transform type
+ * @property f Anonymous function that should return a list of Annotations encapsualted by the parent annotation
```
rick-r3: encapsulated

rick-r3: Full stops are not endangered.  You can use them!

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
```diff
@@ -0,0 +1,268 @@
...
+        val o = other as? RenameSchemaTransform ?: return super.equals(other)
+        return o.from == from && o.to == to
+    }
+
+    override val name: String get() = typeName
+}
+
+
+/**
+ * @property types is a list of serialised types that have transforms, each list element is a
```
rick-r3: This comment seems to end abruptly.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
fenryka: Should probably go in as a different PR

fenryka: but I"ll change it

core/src/main/kotlin/net/corda/core/serialization/CordaSerializationTransformEnumDefault.kt
fenryka: good point :)

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
```diff
@@ -0,0 +1,268 @@
...
+         * one for any transform annotations. If there are any build up a set that can be
+         * encoded into the AMQP [Envelope]
+         *
+         * @param schema should be a [Schema] generated for a serialised data structure
+         * @param sf should be provided by the same serialization context that generated the schema
+         */
+        fun build(schema: Schema, sf: SerializerFactory): TransformsSchema {
+            val rtn = mutableMapOf<String, EnumMap<TransformTypes, MutableList<Transform>>>()
+
+            schema.types.forEach { type ->
```
rick-r3: Thinking....

Each time we send some one a schema, are we recomputing all of this?

Can it be statically computed?  e.g. We contribute entries to the schema as we encounter a serializer for the first time, so if we made the transforms static properties of serializers, we could add them at the same time as adding type to the schema.  Does that work, or have I misunderstood when this is called and/or what it does?

fenryka: That's a good point, we shuld probably cache some of this

fenryka:
@rick-r3 yes, you're correct, at the moment this doesn't do anything that impacts serialisation and deserialisation aside from making the envelope a little bigger if there are transforms applied to a class

This is a fairly contained bt of work that the rest of the evolvability stuff can build on and thus made sense to get it reviewed and in to avoid making later PR's overly large

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
fenryka: raised this as https://github.com/corda/corda/pull/1905

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
fenryka: xorrected

fenryka:
KDocs updated, at least a start has been made
Transform schemas are now cached
Fixed the Java unit tests

fenryka:
@mnesbit 

ByteCode switch on an enum is reduced to a tableswith on the ordinal vale

    INVOKEVIRTUAL net/corda/nodeapi/internal/serialization/amqp/custom/test$E.ordinal ()I
    IALOAD
    TABLESWITCH
      1: L2
      2: L3
      3: L4

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/Envelope.kt
```diff
@@ -0,0 +1,69 @@
...
+ */
+// TODO: make the schema parsing lazy since mostly schemas will have been seen before and we only need it if we
+// TODO: don't recognise a type descriptor.
+data class Envelope(val obj: Any?, val schema: Schema, val transformsSchema: TransformsSchema) : DescribedType {
+    companion object : DescribedTypeConstructor<Envelope> {
+        val DESCRIPTOR = AMQPDescriptorRegistry.ENVELOPE.amqpDescriptor
+        val DESCRIPTOR_OBJECT = Descriptor(null, DESCRIPTOR)
+
+        // described list should either be two or three elements long
+        private const val envelopeWithTransforms = 3
```
mnesbit: In Kotlin consts are expected to be CAPS_FORM

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
```diff
@@ -0,0 +1,276 @@
...
+import java.util.*
+
+// NOTE: We are effectively going to replicate the annotations, we need to do this because
+// we can't instantiate instances of those annotation classes and this code needs to
+// work at the de-serialising end
+/**
+ * Base class for representations of specific types of transforms as applied to a type within the
+ * Corda serialisation framework
+ */
+sealed class Transform : DescribedType {
```
mnesbit: Are we sure about the sealed? This does restrict all future transforms to be in this file.

core/src/main/kotlin/net/corda/core/serialization/CordaSerializationTransformEnumDefault.kt
```diff
@@ -0,0 +1,93 @@
+package net.corda.core.serialization
+
+/**
+ * This annotation is used to mark an enumerated type as having had a multiple members added, It acts
```
rick-r3: `had a multiple` -> `had multiple`

core/src/main/kotlin/net/corda/core/serialization/CordaSerializationTransformEnumDefault.kt
```diff
@@ -0,0 +1,93 @@
...
+ * IMPORTANT - Once added (and in production) do NOT remove old annotations. See documentation for
+ * more discussion on this point!.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class CordaSerializationTransformEnumDefaults(vararg val value: CordaSerializationTransformEnumDefault)
+
+/**
+ * This annotation is used to mark an enumerated type as having had a new constant appended to it. For
+ * each additional constant added a new annotation should be appended to the class. Once more than one
```
rick-r3: `Once` -> `If` ?

core/src/main/kotlin/net/corda/core/serialization/CordaSerializationTransformRename.kt
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * IMPORTANT - Once added (and in production) do NOT remove old annotations. See documentation for
+ * more discussion on this point!.
+ */
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class CordaSerializationTransformRenames(vararg val value: CordaSerializationTransformRename)
+
+/**
+ * This annotation is used to mark a class has having had an element renamed. It is used by the
```
rick-r3: What do we mean by element.  e.g. do we mean a property/constructor arg *and* the class name (including package)?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * R3 AMQP assigned enterprise number
+ *
+ * see [here](https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers)
+ *
+ * Repeated here for brevity:
+ *   50530 - R3 - Mike Hearn - mike&r3.com
+ */
+const val DESCRIPTOR_TOP_32BITS: Long = 0xc5620000
```
rick-r3: Is this correct?  Shouldn't there be 8 zeros?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
```diff
@@ -0,0 +1,31 @@
...
+
+enum class AMQPDescriptorRegistry(val id: Long) {
```
rick-r3: Worth a comment on here saying never delete or change the id number of an entry?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
```diff
@@ -0,0 +1,276 @@
...
+    abstract val name: String
+}
+
+/**
+ * Transform to be used on an Enumerated Type whenever a new element is added
+ *
+ * @property old The value the [new] instance should default to when not available
+ * @property new the value (as a String) that has been added
+ */
+class EnumDefaultSchemeTransform(val old: String, val new: String) : Transform() {
```
rick-r3: `Scheme` -> `Schema` ?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
```diff
@@ -0,0 +1,276 @@
...
+
+
+/**
+ * Represents the set of all transforms that can be a applied to all classes represented as part of
+ * an AMQP schema. It forms a part of the AMQP envelope alongside the [Schema] and the serialized bytes
+ *
+ * @property types maps class names to a map of transformation types. In turn those transformation types
+ * are each a list of instances o that transform.
+ */
+data class TransformsSchema(val types: Map<String, EnumMap<TransformTypes, MutableList<Transform>>>) : DescribedType {
```
rick-r3: Thinking..... if we use an enum inside the transforms, and we add a new transform.... what does the old code deserializing the transforms do?

node-api/src/test/java/net/corda/nodeapi/internal/serialization/amqp/JavaSerializationOutputTests.java
```diff
@@ -186,6 +186,9 @@ private Object serdes(Object obj) throws NotSerializableException {
         decoder.register(CompositeType.Companion.getDESCRIPTOR(), CompositeType.Companion);
         decoder.register(Choice.Companion.getDESCRIPTOR(), Choice.Companion);
         decoder.register(RestrictedType.Companion.getDESCRIPTOR(), RestrictedType.Companion);
+        decoder.register(Transform.Companion.getDESCRIPTOR(), Transform.Companion);
+        decoder.register(TransformsSchema.Companion.getDESCRIPTOR(), TransformsSchema.Companion);
+
```
rick-r3: Remove blank line

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
```diff
@@ -0,0 +1,276 @@
...
+
+            val describedType = described as? DescribedType ?: return TransformsSchema(rtn)
+
+            if (describedType.descriptor != DESCRIPTOR) {
+                throw NotSerializableException("Unexpected descriptor ${describedType.descriptor}.")
+            }
+
+            val map = describedType.described as? Map<*, *> ?:
+                    throw NotSerializableException("Transform schema must be encoded as a map")
+
```
rick-r3: Remove blank line

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
```diff
@@ -0,0 +1,276 @@
...
+        override fun getTypeClass(): Class<*> = TransformsSchema::class.java
+
+        /**
+         * Constructs an instance of the object from the serialised form of an instance
+         * of this object
+         */
+        override fun newInstance(described: Any?): TransformsSchema {
+            val rtn = mutableMapOf<String, EnumMap<TransformTypes, MutableList<Transform>>>()
+
```
rick-r3: Remove blank line

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
fenryka: Fair point, there aren't any compelling reasons whilst it's sealed beyond the fact "this is all the types there will be at the moment" but as it will extend sometime into the future then simply mkaing it abstract is as good and as you say will allow us to factor things int oseperate files

core/src/main/kotlin/net/corda/core/serialization/CordaSerializationTransformRename.kt
fenryka: I was going to have class rename be a separate annotation as that triggers very different behaviour from the evolver, including the at startup scan for renamed classes.

I could go both ways on this

From a user perspective I can see how a rename could be applied in multipel contexts and they'd want us to magically work out what we're talking about, on the other that complicates the code and seperate class and property rename annotations are clearer in what they express

fenryka: For now I've added a comment that this isn't for class renaming

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPDescriptorRegistry.kt
fenryka: yeah... getting confused with the 32 bits vs the top 32 bit sof a 64 bit number... replced with a specific bit shift

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
fenryka: *thinks aloud*

On the serialiser side, that doesn't matter clearly because it will always have the right version of the code, so potential problems exist on the derserialised end

The reality being we attempt to deserialise something with a transform we don't recognise (assuming for a second we are going to say we're never going to remove a transform type (as we would end up in a recursive problem of having to define evolution for the evolution framework ad infinitum (which we could do but damn that would overcomplicate things to the point of insanity))) 

Now, it may be that we never change the defined transforms, and whilst it's extremely unlikely we ever would the chance is always there we would have some reason... However... I think the best thing to do is just have a default option that is a no op for whatever transform we recieve we don't understand

node-api/src/test/java/net/corda/nodeapi/internal/serialization/amqp/JavaSerializationOutputTests.java
fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
fenryka: done

fenryka:
I'll add a default option to the ENUM as per Ricks sugestion then push that

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/TransformsSchema.kt
rick-r3: Yeah, we could have a default feature of enums which looks for entry called "Unrecognised" or something.  Just defaulting to ordinal 0 might be dangerous.  And we make sure we add it to `TransformTypes`.  Or we have a special case in the `EnumSerializer` that recognises it's `TransformTypes` and does custom defaulting (again to special Enum entry).  Which is a bit of a cheat but stops us making the enum evolvability feature even more complex for app developers to understand since it's secret to us.

fenryka: Added an unknown transform type that we set anything we don't recognise to

fenryka:
Back to @rick-r3 / @mnesbit

PR #1907, opened 2017-10-19
Network map service REST API wrapper

patrickkuo:
A network map rest api wrapper, to be used by the network map cache to update the local network map.


core/src/main/kotlin/net/corda/core/node/services/NetworkMapClient.kt
```diff
@@ -0,0 +1,71 @@
...
+interface NetworkMapClient {
+    /**
+     *  Publish node info to network map service.
+     */
+    fun publish(signedNodeInfo: SignedData<NodeInfo>)
+
+    /**
+     *  Retrieve [NetworkMap] from the network map service containing list of node info hashes and network parameter hash.
+     */
+    // TODO: Use NetworkMap object when available.
```
kasiastreich: I think it would be better if I raised a PR after this is in, and wire everything up. Extending old artemis code is pointless.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapClient.kt
```diff
@@ -0,0 +1,71 @@
...
+    fun getNetworkMap(): List<SecureHash>
+
+    /**
+     *  Retrieve [NodeInfo] from network map service using the node info hash.
+     */
+    fun getNodeInfo(nodeInfoHash: SecureHash): NodeInfo?
+
+    // TODO: Implement getNetworkParameter when its available.
+    //fun getNetworkParameter(networkParameterHash: SecureHash): NetworkParameter
```
kasiastreich: So we finally settled on this method? @shamsasari

docs/source/network-map.rst
```diff
@@ -0,0 +1,39 @@
+Network Map
+===========
+
+Protocol Design
+---------------
+.. note:: This section is intended for developers who want to implement their own network map service.
```
kasiastreich: Do we need that sentence? It's almost like doorman docs. Maybe we need some write up on network map service instead.

docs/source/network-map.rst
```diff
@@ -0,0 +1,39 @@
...
+* Serialise the signed data and POST the data to the network map.
+
+* The network map server validate the signature and acknowledge the registration with a http 200 response, it will return http 400 "Bad Request" if the data failed validation or if the public key wasn't registered with the network.
+
+* The network map server will distribute the new network map containing the new node info according to the network parameter ``eventHorizon`` attribute.
+
+
+Node side network map update protocol:
+
+* The Corda node will query the network map service periodically according to the network parameter.
```
kasiastreich: network parameter**s**

docs/source/network-map.rst
```diff
@@ -0,0 +1,39 @@
...
+
+* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameter hashes.
```
kasiastreich: ditto everywhere NetworkParameters

docs/source/network-map.rst
```diff
@@ -0,0 +1,39 @@
...
+
+* The node updates its local copy of NodeInfos and NetworkParameter if it is different from the newly downloaded NetworkMap.
```
kasiastreich: This is not true. NetworkParameters are updated only if you agree on them.

docs/source/network-map.rst
patrickkuo: This is just for the REST api, I am expecting we add in more network map docs.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapClient.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper
+import net.corda.core.crypto.SecureHash
+import net.corda.core.crypto.SignedData
+import net.corda.core.node.NodeInfo
+import net.corda.core.serialization.deserialize
+import net.corda.core.serialization.serialize
+import java.net.HttpURLConnection
+import java.net.URL
+
+interface NetworkMapClient {
```
shamsasari: This all belongs in node

core/build.gradle
```diff
@@ -67,6 +67,16 @@ dependencies {
 
     // JPA 2.1 annotations.
     compile "org.hibernate.javax.persistence:hibernate-jpa-2.1-api:1.0.0.Final"
+
+    // Web stuff: for HTTP[S] servlets
```
shamsasari: These dependencies move as well?

docs/source/network-map.rst
```diff
@@ -0,0 +1,37 @@
+Network Map
+===========
+
+Protocol Design
+---------------
+The node info publishing protocol:
+
+* Create a node info object, and sign it with node's private key to create a ``SignedData<NodeInfo>`` object.
```
shamsasari: Can you elaborate on "node's private key", or is that not needed as it's obvious?

I'm assuming this is the (public) key that was signed by the doorman, CORDA_CLIENT_CA? If that is correct then the doorman already has said key so we can optimise a little by using our own SignedNodeInfo:
```
data class SignedNodeInfo(
  val raw: SerializedBytes<NodeInfo>,
  val signature: DigitalSignature
)
```

`SignedData` uses `DigitalSignature.WithKey`. The extra key may not be needed.

docs/source/network-map.rst
```diff
@@ -0,0 +1,37 @@
...
+
+* Serialise the signed data and POST the data to the network map.
+
+* The network map server validate the signature and acknowledge the registration with a http 200 response, it will return http 400 "Bad Request" if the data failed validation or if the public key wasn't registered with the network.
```
shamsasari: typos: validates, acknowledges

docs/source/network-map.rst
shamsasari: node info -> `NodeInfo`

docs/source/network-map.rst
shamsasari: HTTP

shamsasari: "or if the public key wasn't registered with the network" - see comment above

docs/source/network-map.rst
```diff
@@ -0,0 +1,37 @@
...
+
+* The network map server will distribute the new network map containing the new node info according to the network parameter ``eventHorizon`` attribute.
```
shamsasari: The event horizon is not about the polling interval of the network map, but the time which a node is allowed to be offline before the network map removes it. It means other nodes can safely terminate their flows with this node and flush any pending messages with them, etc.

docs/source/network-map.rst
```diff
@@ -0,0 +1,37 @@
...
+
+
+Node side network map update protocol:
+
+* The Corda node will query the network map service periodically according to the network parameter.
```
shamsasari: Mention the HTTP cache header to look for

docs/source/network-map.rst
```diff
@@ -0,0 +1,37 @@
...
+
+* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameter hashes.
+
+* The node updates its local copy of NodeInfos and NetworkParameters if it is different from the newly downloaded NetworkMap.
```
shamsasari: Elaborate to say the node will download any node info hashes it doesn't know about and remove from its cache any node info hashes not in the network map.

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import java.security.cert.Certificate
+import java.security.cert.CertificateFactory
+import java.security.cert.X509Certificate
+import javax.ws.rs.*
+import javax.ws.rs.core.MediaType
+import javax.ws.rs.core.Response
+import javax.ws.rs.core.Response.ok
+import kotlin.test.assertEquals
+
+internal class HTTPNetworkMapClientTest {
```
shamsasari: Doesn't need to be internal

docs/source/network-map.rst
```diff
@@ -0,0 +1,37 @@
...
+| GET            | /api/network-map                 | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and NetworkParameters hash.                        |
++----------------+----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
+| GET            | /api/network-map/node-info/{hash}| Retrieve ``NodeInfo`` object with the node info hash.                                                                                                  |
++----------------+----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
+| GET            | /api/network-map/parameter/{hash}| Retrieve ``NetworkParameters`` object with the network parameter hash.                                                                                  |
```
gendal: Are these requests authenticated?  ie do we allow anybody who knows the doorman URI to learn who is in the CZ?  Or do you have to have a valid cert from us before you're allowed to find out who else is on?

patrickkuo: Currently there are no authentication in this implementation. 

The network map files will be served using CDN according to the design doc, and depends on which CDN we use there will have different access control, for instance Azura CDN uses token, which can be obtain by the doorman and gives to the node at the time of initial registration (https://docs.microsoft.com/en-us/azure/cdn/cdn-token-auth).

I will add a TODO here to add more auth info once we have decided which CDN we are using.

shamsasari: This means our doorman-node REST API will be determined by the CDN we use.  We should look to come up with something that this is agnostic, e.g. sending a nonce signed by the dooman key, etc.

docs/source/network-map.rst
```diff
@@ -29,9 +28,11 @@ Network Map service REST API:
...
 +----------------+----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
 | GET            | /api/network-map/node-info/{hash}| Retrieve ``NodeInfo`` object with the node info hash.                                                                                                  |
 +----------------+----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /api/network-map/parameter/{hash}| Retrieve ``NetworkParameters`` object with the network parameter hash.                                                                                  |
+| GET            | /api/network-map/parameter/{hash}| Retrieve ``NetworkParameters`` object with the network parameter hash.                                                                                 |
```
shamsasari: /api/network-map/parameters/{hash}

shamsasari: Retrieve ``NetworkParameters`` object with the same hash

docs/source/network-map.rst
```diff
@@ -0,0 +1,38 @@
...
+* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameters hashes.
+
+* The node updates its local copy of ``NodeInfos`` if it is different from the newly downloaded ``NetworkMap``.
+
+Network Map service REST API:
+
++----------------+----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
+| Request method | Path                             | Description                                                                                                                                            |
++================+==================================+========================================================================================================================================================+
+| POST           | /api/network-map/publish         | Publish new node info to the network map service                                                                                                       |
```
shamsasari: Mention that this must match with a previous registration with the doorman

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -0,0 +1,173 @@
...
+    private lateinit var networkMapClient: NetworkMapClient
+    private val rootCAKey = Crypto.generateKeyPair(X509Utilities.DEFAULT_TLS_SIGNATURE_SCHEME)
+    private val rootCACert = X509Utilities.createSelfSignedCACertificate(CordaX500Name(commonName = "Corda Node Root CA", organisation = "R3 LTD", locality = "London", country = "GB"), rootCAKey)
+    private val intermediateCAKey = Crypto.generateKeyPair(X509Utilities.DEFAULT_TLS_SIGNATURE_SCHEME)
+    private val intermediateCACert = X509Utilities.createCertificate(CertificateType.INTERMEDIATE_CA, rootCACert, rootCAKey, X500Name("CN=Corda Node Intermediate CA,L=London"), intermediateCAKey.public)
+
+    companion object {
+        @BeforeClass
+        @JvmStatic
+        fun initSerialization() {
```
shamsasari: use `initialiseTestSerialization` and `resetTestSerialization`, or even `withTestSerialization`

patrickkuo: this cause unit test to fail in team city, i will make the test extend TestDependencyInjectionBase instead.

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -0,0 +1,151 @@
...
+        val signedNodeInfo2 = createStubedNodeInfo("Test2")
+        val nodeInfo2 = signedNodeInfo2.verified()
+        networkMapClient.publish(signedNodeInfo2)
+
+        val nodeInfoHash2 = nodeInfo2.serialize().sha256()
+        assertEquals(listOf(nodeInfoHash, nodeInfoHash2).sorted(), networkMapClient.getNetworkMap().sorted())
+        assertEquals(nodeInfo2, networkMapClient.getNodeInfo(nodeInfoHash2))
+    }
+
+    private fun createStubedNodeInfo(name: String): SignedData<NodeInfo> {
```
shamsasari: rename param to organisation

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+        server.start()
+
+        while (!server.isStarted) {
+            Thread.sleep(100)
+        }
+
+        val hostAndPort = server.connectors.mapNotNull { it as? ServerConnector }
+                .first()
```
shamsasari: I think you can move this up to keep things a bit cleaner

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -0,0 +1,151 @@
...
+
+        assertEquals(listOf(nodeInfoHash), networkMapClient.getNetworkMap())
+        assertEquals(nodeInfo, networkMapClient.getNodeInfo(nodeInfoHash))
+
+        val signedNodeInfo2 = createStubedNodeInfo("Test2")
+        val nodeInfo2 = signedNodeInfo2.verified()
+        networkMapClient.publish(signedNodeInfo2)
+
+        val nodeInfoHash2 = nodeInfo2.serialize().sha256()
+        assertEquals(listOf(nodeInfoHash, nodeInfoHash2).sorted(), networkMapClient.getNetworkMap().sorted())
```
shamsasari: If you used assertJ then you could write something like
`assertThat(networkMapClient.getNetworkMap()).containsExactly(nodeInfoHash, nodeInfoHash2)`

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -0,0 +1,151 @@
...
+        networkMapClient = HTTPNetworkMapClient("http://${hostAndPort.host}:${hostAndPort.localPort}/api/network-map")
+    }
+
+    @After
+    fun tearDown() {
+        server.stop()
+    }
+
+    @Test
+    fun register() {
```
shamsasari: `registered node is added to the network map`

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
shamsasari: This is not really a stub right?

PR #1911, opened 2017-10-20
CORDA-351: added dependency check plugin to gradle build script

bpaunescu:
Added the OWASP dependency check plugin to the gradle build script. It comes with several configurable tasks, 3 of which are the most important:

1. `gradle dependencyCheckAnalyze` will run for all projects and create a report containing detailed descriptions of known library vulnerabilities. The report can be found in several formats(html, xml, json, csv) in /corda/build/reports/

2. `gradle dependencyCheckUpdate` will update the local h2 database with the latest information in the National Vulnerability Database

3. `gradle :dependencyCheckPurge` will delete the h2 database. This task needs to be run only for the root project or it will fail when it's run for the subprojects(database already deleted)

The h2 database is create by default in the .gradle directory and takes up about 300MB. A different location can be configured.

### DIFF
--- a/.ci/dependency-checker/suppressedLibraries.xml
+++ b/.ci/dependency-checker/suppressedLibraries.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+    <!--  Example of a suppressed library -->
+    <!-- The suppress node can be generated from the HTML report by using the 'suppress' option for each vulnerability found
+    <suppress>
+      <notes><![CDATA[
+      file name: some.jar
+      ]]></notes>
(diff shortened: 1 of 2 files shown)


mnesbit:
I am fine with this as it won't affect the normal builds. I would like to know how we might suppress warnings once we have analysed them? Do we need a text file to control that.

bpaunescu:
Warnings and false positives can be suppressed by adding a property to the plugin configuration. It's called `suppressionFile` and it's normally used for false positives but I guess we could use it to exclude vulnerabilities that we are accepting. The file is in XML format and the contents can be generated by using the suppress functions in the HTML report.

mnesbit:
We wouldn't accept any vulnerabilities, but we probably will analyse the problem and decide that they aren't exposed. So I do think this PR needs to include a stub for this file.

mikehearn:
Yes, exactly. Some of the vulns this plugin highlights are blocked already by our global disable on Java object serialisation.

PR #1916, opened 2017-10-20
Adds receiver side of sync flow to ID docs. Cleanup.

joeldudleyr3:
- Adds information about other side of `IdentitySyncFlow`
- Style changes

### DIFF
--- a/docs/source/api-identity.rst
+++ b/docs/source/api-identity.rst
@@ -28,11 +28,11 @@ identity information. For example, it is possible to construct a transaction usi
 learn of your involvement by inspection of the transaction), yet prove to specific counterparts that this
 ``AnonymousParty`` actually corresponds to your well-known identity. This is achieved using the
 ``PartyAndCertificate`` data class, which contains the X.509 certificate path proving that a given ``AnonymousParty``
-corresponds to a given ``Party``. These ``PartyAndCertificate`` can be propagated to counterparties on a need-to-know
+corresponds to a given ``Party``. Each ``PartyAndCertificate`` can be propagated to counterparties on a need-to-know
 basis.
 
 The ``PartyAndCertificate`` class is also used by the network map service to represent well-known identities, with the
-certificate path proving the certificate was issuance by the doorman service.
+certificate path proving the certificate was issued by the doorman service.
 
 Confidential identities
 -----------------------
@@ -44,15 +44,12 @@ public keys of these confidential identities are then used when generating outpu
 transaction.
 
 Where using outputs from a previous transaction in a new transaction, counterparties may need to know who the involved
-parties are. One example is in ``TwoPartyTradeFlow`` which delegates to ``CollectSignaturesFlow`` to gather
-certificates from both parties. ``CollectSignaturesFlow`` requires that a confidential identity of the initiating node
-has signed the transaction, and verifying this requires the receiving node has a copy of the confidential identity for
-the input state. ``IdentitySyncFlow`` can be used to synchronize the confidential identities we have the certificate
-paths for, in a single transaction, to another node.
-
-.. note:: ``CollectSignaturesFlow`` requires that the initiating node has signed the transaction, and as such all nodes
-   providing signatures must recognise the signing key used by the initiating node as being either its well known
-   identity or a confidential identity they have the certificate for.
+parties are. One example is the ``TwoPartyTradeFlow``, where an existing asset is exchanged for cash. If confidential
+identities are being used, the buyer will want to ensure that the asset being transferred is owned by the seller, and
+the seller will likewise want to ensure that the cash being transferred is owned by the buyer. Verifying this requires
+both nodes to have a copy of the confidential identities for the asset and cash input states. ``IdentitySyncFlow`` can
+be used to synchronize the confidential identities we have the certificate paths for, in a single transaction, to
+another node.
 
 SwapIdentitiesFlow
 ~~~~~~~~~~~~~~~~~~
@@ -75,7 +72,7 @@ You can see an example of using ``SwapIdentitiesFlow`` in ``TwoPartyDealFlow.kt`
 
 1. Generate a nonce value to form a challenge to the other nodes
 2. Send nonce value to all counterparties, and receive their nonce values
-3. Generate a new confidential identity from our well known identity
+3. Generate a new confidential identity from our well-known identity
 4. Create a data blob containing the new confidential identity (public key, name and X.509 certificate path),
    and the hash of the nonce values
 5. Sign the resulting data blob with the confidential identity's private key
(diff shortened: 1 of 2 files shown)


docs/source/api-identity.rst
```diff
@@ -7,61 +7,61 @@ API: Identity
...
-involvement by inspection of the transaction, yet prove to specific counterparts that this AnonymousParty actually is
-owned by your well known identity. This disclosure is achieved through the use of the PartyAndCertificate data class
-which can be propagated to those who need to know, and contains the Party's X.509 certificate path to provide proof of
-ownership by a well known identity.
+Support for both ``Party`` and ``AnonymousParty`` classes in Corda enables sophisticated selective disclosure of
+identity information. For example, it is possible to construct a transaction using an ``AnonymousParty`` (so nobody can
+learn of your involvement by inspection of the transaction), yet prove to specific counterparts that this
+``AnonymousParty`` actually corresponds to your well-known identity. This is achieved using the
+``PartyAndCertificate`` data class, which contains the X.509 certificate path proving that a given ``AnonymousParty``
+corresponds to a given ``Party``. These ``PartyAndCertificate`` can be propagated to counterparties on a need-to-know
```
gendal: missing 's' after the closing backticks on PartyAndCertificate

docs/source/api-identity.rst
```diff
@@ -7,61 +7,61 @@ API: Identity
...
+basis.
 
-The PartyAndCertificate class is also used in the network map service to represent well known identities, in which
-scenario the certificate path proves its issuance by the Doorman service.
+The ``PartyAndCertificate`` class is also used by the network map service to represent well-known identities, with the
+certificate path proving the certificate was issuance by the doorman service.
```
gendal: issued

docs/source/api-identity.rst
```diff
@@ -7,61 +7,61 @@ API: Identity
...
+transaction.
 
 Where using outputs from a previous transaction in a new transaction, counterparties may need to know who the involved
-parties are. One example is in ``TwoPartyTradeFlow`` which delegates to ``CollectSignaturesFlow`` to gather certificates
-from both parties. ``CollectSignaturesFlow`` requires that a confidential identity of the initiating node has signed
-the transaction, and verifying this requires the receiving node has a copy of the confidential identity for the input
-state. ``IdentitySyncFlow`` can be used to synchronize the confidential identities we have the certificate paths for, in
-a single transaction, to another node.
+parties are. One example is in ``TwoPartyTradeFlow`` which delegates to ``CollectSignaturesFlow`` to gather
+certificates from both parties. ``CollectSignaturesFlow`` requires that a confidential identity of the initiating node
```
gendal: is it that it needs a _confidential_ identity to signed or just that _any_ signature(s) required by the initiating node have been applied before embarking on the process to collect everybody else's?

docs/source/api-identity.rst
```diff
@@ -7,61 +7,61 @@ API: Identity
...
+has signed the transaction, and verifying this requires the receiving node has a copy of the confidential identity for
+the input state. ``IdentitySyncFlow`` can be used to synchronize the confidential identities we have the certificate
```
gendal: Actually - can you expand on this a little?  ie help the reader understand why it's so important that the receiving node be sure that the initiating node has indeed signed the tx?  (eg is it because our design requires that the initiator can not play games by collecting signatures from everybody else but then not signing themselves? ie we require, by convention, that the initiator always (fully?) signed before asking everybody else to?  If so, do we also need to make it clear that there could be situations where MULTIPLE signatures are required by the initiating node and that we must be sure ALL have been applied?  And if that's not an issue/requirement, say so?)

docs/source/api-identity.rst
```diff
@@ -71,29 +71,35 @@ You can see an example of it being used in ``TwoPartyDealFlow.kt``:
...
+   and the hash of the nonce values
+5. Sign the resulting data blob with the confidential identity's private key
+6. Send the confidential identity and data blob signature to all counterparties, while receiving theirs
+7. Verify the signatures to ensure that identities were generated by the involved set of parties
+8. Verify the confidential identities are owned by the expected well known identities
+9. Store the confidential identities and return them to the calling flow
+
+This ensures not only that the confidential identity X.509 certificates are signed by the correct well-known
+identities, but also that the confidential identity private key is held by the counterparty, and that a party cannot
+claim ownership of another party's confidential identities.
```
gendal: do we need to include a sentence explaining the importance of the nonce?  ie what problem arises if we don't use one?

docs/source/api-identity.rst
```diff
@@ -7,61 +7,61 @@ API: Identity
...
+public keys of these confidential identities are then used when generating output states and commands for the
+transaction.
 
 Where using outputs from a previous transaction in a new transaction, counterparties may need to know who the involved
-parties are. One example is in ``TwoPartyTradeFlow`` which delegates to ``CollectSignaturesFlow`` to gather certificates
-from both parties. ``CollectSignaturesFlow`` requires that a confidential identity of the initiating node has signed
-the transaction, and verifying this requires the receiving node has a copy of the confidential identity for the input
-state. ``IdentitySyncFlow`` can be used to synchronize the confidential identities we have the certificate paths for, in
-a single transaction, to another node.
+parties are. One example is in ``TwoPartyTradeFlow`` which delegates to ``CollectSignaturesFlow`` to gather
```
roger3cev: Swap identities flow not collect signatures flow

roger3cev: .. or to gather signatures?!

docs/source/api-identity.rst
roger3cev: maybe we can re-word this. The identity doesn't have to be confidential.

docs/source/api-identity.rst
```diff
@@ -7,61 +7,61 @@ API: Identity
...
+certificates from both parties. ``CollectSignaturesFlow`` requires that a confidential identity of the initiating node
+has signed the transaction, and verifying this requires the receiving node has a copy of the confidential identity for
+the input state. ``IdentitySyncFlow`` can be used to synchronize the confidential identities we have the certificate
+paths for, in a single transaction, to another node.
 
 .. note:: ``CollectSignaturesFlow`` requires that the initiating node has signed the transaction, and as such all nodes
-   providing signatures must recognise the signing key used by the initiating node as being either its well known identity
-   or a confidential identity they have the certificate for.
+   providing signatures must recognise the signing key used by the initiating node as being either its well known
+   identity or a confidential identity they have the certificate for.
```
roger3cev: This is referring to the check in `SignTransactionFlow` that fails if any one of the counter-parties doesn't have a certificate for any of the other signers.

joeldudleyr3:
Feedback addressed, except for the use of a nonce. I don't understand why it's used, so I'll discuss with Ross and include the explanation in a future update.

Rewrote the example for IdentitySyncFlow to address the concerns raised.

PR #1918, opened 2017-10-20
Add observed states to vault

rnicoll:
Add the concept of an observed state to the vault, for use-cases where
the vault explicitely should store a state which it is not a participant
of.

Add a demo of these observed states, derived from the old trade finance
code.


rnicoll:
Raising early while I finish docs on this. I'm aware we may have to extend the vault API to allow recording of all states in a transaction, but want to raise that as its own JIRA & PR later.

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -118,6 +118,18 @@ interface LinearState : ContractState {
 }
 // DOCEND 2
 
+/**
+ * A state which is observed by parties who are not participants in the state. This does not enforce that the observers
+ * are sent a copy of the state, if enforcement of observation is required it must be provided by the contract itself.
```
gendal: I don't understand this. How does a contract enforce the sending?  (Also...  full stop before "if")

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -118,6 +118,18 @@ interface LinearState : ContractState {
...
+ */
+interface ObservedState : ContractState {
+    /**
+     * Parties who observe this state being created; they are not conventional owners or participants, and cannot
+     * affect the state, but retain a copy of the state for their own purposes.
+     */
```
gendal: why would you add somebody to this list?  Is there a mechanism that means it will be sent to them (by whom?).

core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
```diff
@@ -39,11 +39,16 @@ class Vault<out T : ContractState>(val states: Iterable<StateAndRef<T>>) {
      *
      * If the vault observes multiple transactions simultaneously, where some transactions consume the outputs of some of the
      * other transactions observed, then the changes are observed "net" of those.
+     *
+     * @property consumed the set of states which have been consumed (spent).
+     * @property produced the set of states which have been produced (issued).
+     * @property observed the set of states which have been observed (produced but are not owned by us).
```
gendal: produced by us?

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
mnesbit: I like the vault changes, but the more I look at this the more I think we need to have go with an observerTransaction(tx: FilteredTransaction) rather than defining a ContractState interface. The problem is that states are data not verbs.
If we enforce things properly, we should also delete/retire observed states when the hosts do, which may not be desired. The alternative is that we start getting a lack of synchrony on the ledger, which doesn't sound wise.

core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
```diff
@@ -98,13 +108,13 @@ class Vault<out T : ContractState>(val states: Iterable<StateAndRef<T>>) {
...
-        val NoUpdate = Update(emptySet(), emptySet(), type = Vault.UpdateType.GENERAL)
-        val NoNotaryUpdate = Vault.Update(emptySet(), emptySet(), type = Vault.UpdateType.NOTARY_CHANGE)
+        val NoUpdate = Update(emptySet(), emptySet(), emptySet(), type = Vault.UpdateType.GENERAL)
+        val NoNotaryUpdate = Vault.Update(emptySet(), emptySet(), emptySet(), null, type = Vault.UpdateType.NOTARY_CHANGE)
     }
 
     @CordaSerializable
     enum class StateStatus {
-        UNCONSUMED, CONSUMED, ALL
+        UNCONSUMED, CONSUMED, OBSERVED, ALL
```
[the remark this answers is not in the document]
mnesbit: Also, what are we doing to prevent OBSERVED states being used as inputs into transactions?

mikehearn:
I really wish we could get (back) into the habit of writing the docs first, before the code. I'm not at all sure we need/should have this ObservedState interface. Simply sending a transaction to a node in a flow should be sufficient for it to be inserted into the vault as an observation. Why should the state itself be involved at all?

rnicoll:
I've rebased, expanded the documentation to try talking more about the rationale, and corrected API backwards compatibility issues.

Something I'm not sure that's too obvious about why I've done it this way, and doesn't fit well into the docs - the updates from the vault contain produced and consumed states, and part of this is to expand the updates to have a separate list for observed states, so CorDapps looking for produced states don't find states they can't spend. Hope that makes sense.

mikehearn:
I put together an alternative approach yesterday, see what you think:

https://github.com/corda/corda/pull/1946

rnicoll:
Closing as there's an alternative solution

PR #1926, opened 2017-10-23
CORDA-736 Add some new features to corda.jar via node.conf for testing

rick-r3:
1. Ability to configure system properties for switching A / B testing and other runtime configurable features.
2. Include additional JARs for e.g. JDBC drivers.
3. JVM arguments for tweaking heap size, profiling, debug arguments etc.

It also requires us to respect command line args understood by the node for specifying the configuration file for completeness.  This requires the caplet to be able to parse the config file and see reference.conf.

### DIFF
--- a/node/capsule/build.gradle
+++ b/node/capsule/build.gradle
@@ -31,10 +31,9 @@ task buildCordaJAR(type: FatCapsule, dependsOn: project(':node').compileJava) {
             "$rootDir/config/dev/log4j2.xml"
     )
     from 'NOTICE' // Copy CDDL notice
-    from { project(':node').configurations.runtime.filter { // Include config library JAR.
-        (((it.toString().contains("/config-")) ||
-                (it.toString().contains("\\com\\typesafe\\config\\") && it.toString().contains("\\config-"))) && it.toString().endsWith(".jar"))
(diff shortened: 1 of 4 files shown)


node/capsule/build.gradle
```diff
@@ -31,7 +31,10 @@ task buildCordaJAR(type: FatCapsule, dependsOn: project(':node').compileJava) {
             "$rootDir/config/dev/log4j2.xml"
     )
     from 'NOTICE' // Copy CDDL notice
-
+    from { project(':node').configurations.runtime.filter { // Include config library JAR.
+        (it.toString().contains("/config-") || it.toString().contains("\\config-")) && it.toString().endsWith(".jar")
```
josecoll: Where does this magical jar get built and what does it contain?

rick-r3: It's one of our 3rd party dependencies.  File name is `config-<version>.jar`.  Seemed high maintenance to have to update the version number here.  I'm open to better ideas for locating it.  Maybe @Clintonio might have some ideas?

rick-r3: Actually, it should also contain `/com/typesafe/config/` (and perhaps with Windows style slashes too).  Let me add that.  More definitive.

PR #1931, opened 2017-10-23
Increase logging level of FlowStateMachineImpl.suspend() exception (#…

IgorNitto:

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -480,7 +480,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
                 // Quasar does not terminate the fiber properly if an exception occurs during a suspend. We have to
                 // resume the fiber just so that we can throw it when it's running.
                 exceptionDuringSuspend = t
-                logger.trace("Resuming so fiber can it terminate with the exception thrown during suspend process", t)
+                logger.error("Resuming so fiber can terminate with the exception thrown during suspend process", t)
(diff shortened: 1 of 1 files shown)


shamsasari:
@roger3cev can you elaborate on why this change is needed?

roger3cev:
@shamsasari apologies just seen this. It was in relation to an exception that was thrown but it was only logged at info level despite it being more critical. If i recall correctly, it was due to a flow checkpoint that was not restorable. The checkpoint never gets restored and you don't know this is the case (unless you read the logs).

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -480,7 +480,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
                 // Quasar does not terminate the fiber properly if an exception occurs during a suspend. We have to
                 // resume the fiber just so that we can throw it when it's running.
                 exceptionDuringSuspend = t
-                logger.trace("Resuming so fiber can it terminate with the exception thrown during suspend process", t)
+                logger.error("Resuming so fiber can terminate with the exception thrown during suspend process", t)
```
shamsasari: As the log message implies, `t` should be thrown anyway and terminate the flow. If that's not happening then we need to look into that and fix. This was trace level precisely because we're expecting the exception to be thrown.

shamsasari:
@roger3cev I think you're referring to the checkpoint checker that runs in the background in devMode. This checker does it checks in a background thread so it isn't going to stop the flow. If you're seeing this when using the MockNetwork, then make sure to call `stopNodes` in an `@After` method as that will fail the test.

PR #1932, opened 2017-10-23
Minor fix for api checker script to support macOS

adagys:


chrisr3:
Can you also try adding `set +o posix` to the script like this please?
```bash
#!/bin/bash
set +o posix
```
I am thinking that your `bash` shell has been compiled to enable POSIX mode by default, which means that process substitution is disabled. The above line should _disable_ POSIX mode.

PR #1933, opened 2017-10-24
CORDA-351: update dependencies to resolve found vulnerabilities

bpaunescu:
Updated some dependencies to newer versions which have solved the vulnerability issues found by the dependency-check analyzer.


mikehearn:
Which ones affect us? We should have notes so we can do a 1.1 release.

bpaunescu:
@mikehearn I will prepare notes today after I take a closer look at how we're affected by the remaining vulnerabilities.

PR #1938, opened 2017-10-24
DO NOT REVIEW OR COMMIT

fenryka:
This is a test, please ignore,

when AMQP enabled this fails, without it passes

Seriously!!!! DON'T COMMIT THIS OR REVIEW IT


rnicoll:
YOU'RE NOT THE BOSS OF ME!!!

(but okay, won't)

fenryka:
@rnicoll you're supposed to be on holiday

mikehearn:
Is this just a way to make TeamCity run? I think you can request runs on the TC UI. At any rate, it's not great to have PRs that aren't meant to be reviewed, a PR is literally a "pull request". I'll close this - please do trigger runs against your branches via TC directly. If you don't have the right access level to do that can you work with Ben to get it please? Thanks!

PR #1941, opened 2017-10-24
Perform sensible currency conversion

vkolomeyko:
It is kind of sad to see 1 GBP to be equal to 1 USD and also it looks misleading in the blotter.

### DIFF
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/ExchangeRateModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/ExchangeRateModel.kt
@@ -3,8 +3,12 @@ package net.corda.client.jfx.model
 import javafx.beans.property.SimpleObjectProperty
 import javafx.beans.value.ObservableValue
 import net.corda.core.contracts.Amount
+import net.corda.finance.CHF
+import net.corda.finance.EUR
+import net.corda.finance.GBP
+import net.corda.finance.USD
 import java.math.BigDecimal
-import java.math.RoundingMode
+import java.math.MathContext
 import java.util.*
 
 /**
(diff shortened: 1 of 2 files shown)


client/jfx/src/main/kotlin/net/corda/client/jfx/model/ExchangeRateModel.kt
```diff
@@ -24,9 +24,25 @@ abstract class ExchangeRate {
 /**
  * Default implementation of an exchange rate model, which uses a fixed exchange rate.
  */
+private val usdExchangeRates: Map<Currency, BigDecimal> = mapOf(
+        Currency.getInstance("GBP") to BigDecimal(1.31),
```
shamsasari: You can use the currency constants instead

vkolomeyko: Sure - will do

vkolomeyko:
All the comments addressed and TC is happy with my changes.
@shamsasari - could you please re-review my work?

client/jfx/src/main/kotlin/net/corda/client/jfx/model/ExchangeRateModel.kt
```diff
@@ -24,9 +28,24 @@ abstract class ExchangeRate {
 /**
  * Default implementation of an exchange rate model, which uses a fixed exchange rate.
  */
+private val usdExchangeRates: Map<Currency, BigDecimal> = mapOf(
+        GBP to BigDecimal(1.31),
```
shamsasari: The recommended way to create BigDecimals from doubles is to use `BigDecimal.valueOf`.

PR #1942, opened 2017-10-24
Remove for good network map service node

al-r3:
Remove NetworkMapService from AbstractNode
Also make NodeBasedTest not depend on the NetworkMapService


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -619,14 +574,10 @@ abstract class AbstractNode(config: NodeConfiguration,
...
-    protected open fun noNetworkMapConfigured(): CordaFuture<Unit> {
-        if (services.networkMapCache.loadDBSuccess || configuration.noNetworkMapServiceMode) {
-            return doneFuture(Unit)
-        } else {
+    open protected fun checkNetworkMapIsInitialized() {
+        if (!services.networkMapCache.loadDBSuccess) {
             // TODO: There should be a consistent approach to configuration error exceptions.
-            throw IllegalStateException("Configuration error: this node isn't being asked to act as the network map, nor " +
-                    "has any other map node been configured.")
+            throw NetworkMapCacheEmptyException()
```
shamsasari: I don't see the benefit of a custom exception here, especially as it's not being caught anywhere. IllegalStateException will suffice.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -232,10 +222,13 @@ abstract class AbstractNode(config: NodeConfiguration,
...
 
             runOnStop += network::stop
-            StartedNodeImpl(this, _services, info, checkpointStorage, smm, attachments, inNodeNetworkMapService, network, database, rpcOps, flowStarter, schedulerService)
+            StartedNodeImpl(this, _services, info, checkpointStorage, smm, attachments, network, database, rpcOps, flowStarter, schedulerService)
         }
         // If we successfully  loaded network data from database, we set this future to Unit.
-        _nodeReadyFuture.captureLater(registerWithNetworkMapIfConfigured())
+        services.networkMapCache.addNode(info)
+        checkNetworkMapIsInitialized()
+        _nodeReadyFuture.set(Unit)
```
shamsasari: I guess @patrickkuo you will be fixing this so that it's set on the first poll from the network map end-point.

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -197,16 +192,14 @@ open class Node(configuration: FullNodeConfiguration,
 
     /**
      * Checks whether the specified [host] is a public IP address or hostname. If not, tries to discover the current
-     * machine's public IP address to be used instead. It first looks through the network interfaces, and if no public IP
-     * is found, asks the network map service to provide it.
+     * machine's public IP address to be used instead by looking through the network interfaces.
+     * TODO this code used to rely on the networkmap node, we might want to look at a different solution.
```
shamsasari: The doorman already exposes a REST end-point for this. Liaise with @patrickkuo on how this works.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -15,12 +14,6 @@ interface NodeConfiguration : NodeSSLConfiguration {
     // myLegalName should be only used in the initial network registration, we should use the name from the certificate instead of this.
     // TODO: Remove this so we don't accidentally use this identity in the code?
     val myLegalName: CordaX500Name
-    /**
-     * If null then configure the node to run as the netwok map service, otherwise use this to connect to the network map
-     * service.
-     */
-    val networkMapService: NetworkMapInfo?
```
shamsasari: Delete NetworkMapInfo

node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -413,17 +397,10 @@ class ArtemisMessagingServer(override val config: NodeConfiguration,
     // This is called on one of Artemis' background threads
     internal fun hostVerificationFail(expectedLegalNames: Set<CordaX500Name>, errorMsg: String?) {
```
shamsasari: All this method does now is log. You can inline it as it's only used once.

node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -413,17 +397,10 @@ class ArtemisMessagingServer(override val config: NodeConfiguration,
...
         log.error(errorMsg)
-        if (config.networkMapService?.legalName in expectedLegalNames) {
-            // If the peer that failed host verification was the network map node then we're in big trouble and need to bail!
-            _networkMapConnectionFuture!!.setException(IOException("${config.networkMapService} failed host verification check"))
-        }
     }
 
     // This is called on one of Artemis' background threads
     internal fun onTcpConnection(peerLegalName: CordaX500Name) {
```
shamsasari: delete

node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -413,17 +397,10 @@ class ArtemisMessagingServer(override val config: NodeConfiguration,
...
-        if (peerLegalName == config.networkMapService?.legalName) {
-            _networkMapConnectionFuture!!.set(Unit)
-        }
     }
 
     private fun handleIpDetectionRequest(queueName: String) {
```
shamsasari: This is no longer used.

node/src/main/kotlin/net/corda/node/services/messaging/NodeMessagingClient.kt
```diff
@@ -634,9 +629,7 @@ class NodeMessagingClient(override val config: NodeConfiguration,
     // TODO Rethink PartyInfo idea and merging PeerAddress/ServiceAddress (the only difference is that Service address doesn't hold host and port)
     override fun getAddressOfParty(partyInfo: PartyInfo): MessageRecipients {
         return when (partyInfo) {
-            is PartyInfo.SingleNode -> {
-                getArtemisPeerAddress(partyInfo.party, partyInfo.addresses.first(), config.networkMapService?.legalName)
-            }
+            is PartyInfo.SingleNode -> getArtemisPeerAddress(partyInfo.party, partyInfo.addresses.first())
```
shamsasari: `partyInfo.addresses[0]`

node/src/main/kotlin/net/corda/node/services/messaging/NodeMessagingClient.kt
```diff
@@ -175,11 +174,7 @@ class NodeMessagingClient(override val config: NodeConfiguration,
     /**
      * Apart from the NetworkMapService this is the only other address accessible to the node outside of lookups against the NetworkMapCache.
```
shamsasari: Is this comment still relevant?

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -152,40 +138,10 @@ open class PersistentNetworkMapCache(private val database: CordaPersistence, con
...
-            Unit
-        }
-        _registrationFuture.captureLater(future.map { null })
-
-        return future
-    }
-
     override fun addNode(node: NodeInfo) {
         logger.info("Adding node with info: $node")
+        _loadDBSuccess = true // This is used in AbstractNode to indicate that node is ready.
```
shamsasari: Can you at least comment that this is temporary until the network map cache starts polling the network map HTTP end-point.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetworkMapCache.kt
```diff
@@ -27,32 +27,12 @@ class MockNetworkMapCache(database: CordaPersistence, configuration: NodeConfigu
     }
 
     override val changed: Observable<NetworkMapCache.MapChange> = PublishSubject.create<NetworkMapCache.MapChange>()
+    override val nodeReady: CordaFuture<Void?> get() = doneFuture(null)
```
shamsasari: Can you make `PersistentNetworkMapCache._registrationFuture` protected and then complete it in the init of this class. It's ugly, but better than return a new done future each time.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -239,7 +237,7 @@ class MockNetwork(defaultParameters: MockNetworkParameters = MockNetworkParamete
         }
 
         // It's OK to not have a network map service in the mock network.
```
shamsasari: Comment is out of date.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -554,11 +547,7 @@ abstract class AbstractNode(config: NodeConfiguration,
     }
 
     private fun setupInNodeNetworkMapService(networkMapCache: NetworkMapCacheInternal) {
-        inNodeNetworkMapService =
-                if (configuration.networkMapService == null && !configuration.noNetworkMapServiceMode)
-                    makeNetworkMapService(network, networkMapCache)
-                else
-                    NullNetworkMapService
+        inNodeNetworkMapService = NullNetworkMapService
```
shamsasari: In fact it looks like `inNodeNetworkMapService` isn't used

node-api/src/main/kotlin/net/corda/nodeapi/ArtemisMessagingComponent.kt
```diff
@@ -84,11 +83,8 @@ abstract class ArtemisMessagingComponent : SingletonSerializeAsToken() {
...
 
     // Used for bridges creation.
-    fun getArtemisPeerAddress(party: Party, address: NetworkHostAndPort, netMapName: CordaX500Name? = null): ArtemisPeerAddress {
-        return if (party.name == netMapName) {
-            NetworkMapAddress(address)
-        } else {
-            NodeAddress.asSingleNode(party.owningKey, address) // It also takes care of services nodes treated as peer nodes
-        }
+    fun getArtemisPeerAddress(party: Party, address: NetworkHostAndPort): ArtemisPeerAddress {
+        return NodeAddress.asSingleNode(party.owningKey, address) // It also takes care of services nodes treated as peer nodes
```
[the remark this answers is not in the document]
al-r3: done, it's a secondary constructor though as data classes primary constructor can't take something which isn't a val.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: I had in mind to come back and revisit all the dbLoaded/nodeReady future and related.
Making this succeed after the first net poll seems wrong we should be fine with just the DB and/or files on disk.
It was also a requirement to have nodes starting even if they can't poll the network (of course you could mark the node ready after a failed poll, though....)

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: I've literally followed this: https://docs.corda.net/codestyle.html#assertions-and-errors

node/src/main/kotlin/net/corda/node/internal/Node.kt
al-r3: I imagine in a future PR, I don't this PR needs to be any longer than it is already

node/src/main/kotlin/net/corda/node/services/messaging/NodeMessagingClient.kt
al-r3: I remember seeing that comment and not fully understanding what it meant.
The node address is put in the networkmap cache. So even now it doesn't mean much, I believe, I'm removing it.

node/src/main/kotlin/net/corda/node/services/messaging/NodeMessagingClient.kt
al-r3: Why? .first does the same thing, and we use it all over the place.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
al-r3: I've added a comment saying this needs rethinking, polling the HTTP endpoint is imho not the 100% right solution.
For the record if we did that we'd need another way of making driver, mocknetwork and friends work

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetworkMapCache.kt
al-r3: I can just directly return _registrationFuture and achieve the same

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: yup that feels right!

al-r3:
@shamsasari I thought you were on holiday!

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
shamsasari: Interesting, I never knew. I have to say I disagree with that.

node/src/main/kotlin/net/corda/node/internal/Node.kt
shamsasari: Fair enough, we already have a JIRA ticket for making sure this works, so I guess it can be done for that.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
shamsasari: I don't think it will be a problem since for the driver, et al the doorman URL config won't be specified and so the future won't have to wait on that.

shamsasari:
I'll leave approval to @patrickkuo.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: fwiw I agree with that, it's easier to catch specifically that (whether you're doing it already or not), it's easier to find where it has been thrown if you see a stacktrace with one.

shamsasari:
You also need to update corda-configuration-file.rst

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
patrickkuo: so if we only have 1 node this will always be false? maybe better to put this at the end of `loadFromDB` method?
unless this is use to indicate "someone joined the network"?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -818,3 +763,8 @@ internal class FlowStarterImpl(private val serverThread: AffinityExecutor, priva
         return serverThread.fetchFrom { smm.startFlow(logic, flowInitiator, ourIdentity) }
     }
 }
+
+/**
+ * Thrown when a node is about to start and its network map cache doesn't contain any node.
+ */
+internal class NetworkMapCacheEmptyException: Exception()
```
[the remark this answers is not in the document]
al-r3: no, it's in the class name.

shamsasari:
And `net.corda.demobench.model.NodeConfig`.

al-r3:
Tests are good, both net.corda.demobench.model.NodeConfig and gradlew smokeTest

PR #1944, opened 2017-10-25
CORDA-351: force update dependencies and suppress vulnerabilities not…

bpaunescu:
Updated some dependencies to newer versions which have solved the vulnerability issues found by the dependency-check analyzer.

### DIFF
--- a/.ci/dependency-checker/suppressedLibraries.xml
+++ b/.ci/dependency-checker/suppressedLibraries.xml
@@ -10,5 +10,22 @@
       <cpe>cpe:/a:apache:struts:2.0.0</cpe>
    </suppress>
     -->
-
+    <suppress>
+        <!-- Vulnerability when using SSLv2 Hello messages. Corda uses TLS1.2-->
+        <notes><![CDATA[file name: catalyst-netty-1.1.2.jar]]></notes>
+        <gav regex="true">^io\.atomix\.catalyst:catalyst-netty:.*$</gav>
+        <cve>CVE-2014-3488</cve>
+    </suppress>
+    <suppress>
+        <!-- Vulnerability to LDAP poisoning attacks. Corda doesn't use LDAP-->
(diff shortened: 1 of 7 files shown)


fenryka:
adding @Clintonio to the review for the Gradle stuff, LGTM but I'll give him the final OK

.ci/dependency-checker/suppressedLibraries.xml
```diff
@@ -10,5 +10,22 @@
       <cpe>cpe:/a:apache:struts:2.0.0</cpe>
    </suppress>
     -->
-
+    <suppress>
```
Clintonio: Can we add a JIRA to re-visit these later? We'll want to enable them if they get fixed.

client/jfx/build.gradle
```diff
@@ -37,6 +37,9 @@ dependencies {
     compile 'org.fxmisc.easybind:easybind:1.0.3'
 
     // Artemis Client: ability to connect to an Artemis broker and control it.
+    // TODO: remove the forced update of commons-collections and beanutils when artemis updates them
+    compile group: 'org.apache.commons', name: 'commons-collections4', version: "$commons_collections_version"
```
Clintonio: Please change these to the stringly form. While this form is more type safe it isn't standard across our gradle files.

.ci/dependency-checker/suppressedLibraries.xml
bpaunescu: Added https://r3-cev.atlassian.net/browse/CORDA-745

mikehearn:
I have filed https://issues.apache.org/jira/browse/ARTEMIS-1483 to get Artemis to upgrade Apache CC. They already did upgrade CC once before to fix this issue, however the fix didn't take because they are importing CC via BeanUtils as well. 

Thanks!

PR #1946, opened 2017-10-25
Add support for observer nodes (alternative approach)

mikehearn:
This PR adds support for observer nodes in a different way to PR #1918 - the primary differences are:

* No tagging of states
* No new demo, just a small extension to the IRS demo
* There is no distinction in vault queries or observables between observed or not-observed states.
* Whether states should be "observed" or not is passed through the API from the flows, rather than inferred by the vault based on the state contents.

The latter approach avoids verb/noun type conceptual mixup and solves the probably common case where two sides to a transaction wish to report the same transaction to different regulators (which would otherwise require both to be listed in the observers and involve both receiving duplicated reports). In the case where the regulatory reporting process is more involved than just dispatching a transaction e.g. more country-specific data is required, it makes sense to me to let the developer control this process.

It's possible that some apps could benefit from separation between relevant/irrelevant states at the level of the observables. This would seem to be an issue when a regulator is e.g. also making payments of their own, on the ledger, and an app is written to assume all states entering the vault are relevant. However, it's easy to solve this later by simply extending the vault API to let you request filtered streams of updates, and coin selection should at any rate ignore irrelevant transactions. For now we can simply punt on the issue of a regulator that also interacts on the ledger themselves using apps not written with that scenario in mind.

The change is a fair bit smaller as a consequence of these differences.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/ReceiveTransactionFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/ReceiveTransactionFlow.kt
@@ -3,6 +3,7 @@ package net.corda.core.flows
 import co.paralleluniverse.fibers.Suspendable
 import net.corda.core.contracts.*
 import net.corda.core.internal.ResolveTransactionsFlow
+import net.corda.core.node.StatesToRecord
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.utilities.unwrap
 import java.security.SignatureException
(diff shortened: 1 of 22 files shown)


core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
```diff
@@ -144,6 +160,16 @@ interface ServiceHub : ServicesForResolution {
 
     /**
      * Stores the given [SignedTransaction]s in the local transaction storage and then sends them to the vault for
+     * further processing if [statesToRecord] is not [StatesToRecord.NONE].
+     * This is expected to be run within a database transaction.
+     *
+     * @param txs The transactions to record.
+     * @param statesToRecord how the vault should treat the output states of the transaction.
+     */
+    fun recordTransactions(statesToRecord: StatesToRecord, txs: Iterable<SignedTransaction>)
```
mnesbit: I was thinking instead of an observeTransaction function that takes a FilteredTransaction and extracts outputs into the vault as OBSERVED, because people seem to be asking to hide some states from each other. This will work, but I feel that we have to be careful that people don't try to spend states they don't own. Also it breaks the globally consistent view of the ledger, because there is no guarantee that an vault state will be signaled as consumed to the original observers.

mikehearn: This whole set of APIs will need an overhaul anyway once SGX is available. Then the data will be popping out of the enclave in filtered form, so it'll have to take an alternative route into the vault (e.g. a private API). This API can stick around for the case where SGX is not in use or where it's OK for the transaction to be entirely visible.

Coin selection should eliminate any irrelevant states.

mikehearn: (specifically, coin selection *should* eliminate any irrelevant states, but probably doesn't for now. in the interests of time I think it's OK to ship the feature like this with caveats in the docs, and then update the interactions with other features later).

mikehearn:
Discussion with the customers indicates they're happy with this approach. Will rebase. Let's get it merged and a release with it included out the door.

mikehearn:
Rebased. @mnesbit can you review the rest of the patch please?

core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
```diff
@@ -51,6 +51,22 @@ interface ServicesForResolution : StateLoader {
...
+ */
+enum class StatesToRecord {
+    /** The received transaction is not sent to the vault at all. This is used within transaction resolution. */
+    NONE,
+    /**
+     * All states that can be seen in the transaction will be recorded by the vault, even if none of the identities
+     * on this node are a participant or owner.
+     */
+    ALL_VISIBLE,
+    /** Only states that involve one of our public keys will be stored in the vault. This is the default. */
```
mnesbit: Is this sufficiently accurate? We call the isRelevant method which uses owner key for OwnableSTate, else the participants for other state types.

mikehearn: I think it is? The word "involve" here is being used as shorthand for "one of our public keys is the owner, or in the participants list" - do you think that's excessively short?

core/src/main/kotlin/net/corda/core/flows/ReceiveTransactionFlow.kt
```diff
@@ -14,25 +15,29 @@ import java.security.SignatureException
...
             TransactionResolutionException::class,
             TransactionVerificationException::class)
     override fun call(): SignedTransaction {
-        return otherSideSession.receive<SignedTransaction>().unwrap {
+        val stx = otherSideSession.receive<SignedTransaction>().unwrap {
             subFlow(ResolveTransactionsFlow(it, otherSideSession))
             it.verify(serviceHub, checkSufficientSignatures)
             it
         }
+        if (checkSufficientSignatures)
```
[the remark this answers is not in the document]
mikehearn: Done, and added a comment.

node/src/main/kotlin/net/corda/node/services/CoreFlowHandlers.kt
```diff
@@ -16,8 +17,7 @@ import net.corda.core.transactions.SignedTransaction
 class FinalityHandler(private val sender: FlowSession) : FlowLogic<Unit>() {
     @Suspendable
     override fun call() {
-        val stx = subFlow(ReceiveTransactionFlow(sender))
-        serviceHub.recordTransactions(stx)
+        val stx = subFlow(ReceiveTransactionFlow(sender, true, StatesToRecord.ONLY_RELEVANT))
```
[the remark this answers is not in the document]
mikehearn: Done and added some logging.

mikehearn:
Actually, I'll add another comment summarising our discussion about coin selection.

core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
mnesbit: I think in general people struggle to understand the subtleties of participants, owners, signers and we have not made it very clear where we use things. I agree it doesn't need to be war and peace, but maybe a little bit more colour.

PR #1950, opened 2017-10-26
checkpoint checker thread

blsemo:
- Give  the checkpoint checker thread a meaningful name so it can be identified while debugging or profiling
- Add devModeOptions as a bag of options that can be set for dev mode and add an options to turn the checkpoint checker off in dev mode (e.g. for profiling)

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
+++ b/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
@@ -36,6 +36,10 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val notary: NotaryConfig?
     val activeMQServer: ActiveMqServerConfiguration
     val additionalNodeInfoPollingFrequencyMsec: Long
+
+    companion object {
+        val disableCheckpointCheckerFlag = "disableCheckpointChecker"
+    }
 }
 
 fun NodeConfiguration.isDevModeOptionsFlagSet(flag: String):Boolean{
--- a/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
+++ b/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
@@ -28,6 +28,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val database: Properties?
     val rpcUsers: List<User>
     val devMode: Boolean
+    val debugOptions: Properties?
     val certificateSigningService: URL
     val certificateChainCheckPolicies: List<CertChainPolicyConfig>
     val verifierType: VerifierType
@@ -93,6 +94,7 @@ data class FullNodeConfiguration(
         override val notary: NotaryConfig?,
         override val certificateChainCheckPolicies: List<CertChainPolicyConfig>,
         override val devMode: Boolean = false,
+        override val debugOptions: Properties? = null,
         val useTestClock: Boolean = false,
         val detectPublicIp: Boolean = true,
         override val activeMQServer: ActiveMqServerConfiguration,
@@ -103,6 +105,7 @@ data class FullNodeConfiguration(
     init {
         // This is a sanity feature do not remove.
         require(!useTestClock || devMode) { "Cannot use test clock outside of dev mode" }
+        require(debugOptions == null || devMode){"Cannot use debugOptions outside of dev mode"}
         // TODO Move this to ArtemisMessagingServer
         rpcUsers.forEach {
             require(it.username.matches("\\w+".toRegex())) { "Username ${it.username} contains invalid characters" }
--- a/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
+++ b/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
@@ -42,7 +42,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     }
 }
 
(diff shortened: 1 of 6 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/StateMachineManagerImpl.kt
```diff
@@ -92,7 +89,7 @@ class StateMachineManagerImpl(
     private val scheduler = FiberScheduler()
     private val mutex = ThreadBox(InnerState())
     // This thread (only enabled in dev mode) deserialises checkpoints in the background to shake out bugs in checkpoint restore.
-    private val checkpointCheckerThread = if (serviceHub.configuration.devMode) Executors.newSingleThreadExecutor() else null
+    private val checkpointCheckerThread = if (serviceHub.configuration.devMode && !serviceHub.configuration.isDevModeOptionsFlagSet("disableCheckpointChecking")) newNamedSingleThreadExecutor("CheckpointChecker") else null
```
al-r3: please use a constant for "disableCheckpointChecking"

node/src/main/kotlin/net/corda/node/utilities/NamedThreadFactory.kt
```diff
@@ -0,0 +1,19 @@
+package net.corda.node.utilities
+
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.ThreadFactory
+import java.util.concurrent.atomic.AtomicInteger
+
+class NamedThreadFactory(private val name:String, private val underlyingFactory: ThreadFactory) : ThreadFactory{
```
al-r3: why do you need all this? seems you're creating a single thread anyway, so the counter should always be 1, or am I missing something?
Also add a comment to the class.

node/src/main/kotlin/net/corda/node/utilities/NamedThreadFactory.kt
```diff
@@ -0,0 +1,19 @@
...
+    val threadNumber = AtomicInteger(1)
+    override fun newThread(r: Runnable?): Thread {
```
al-r3: rename r to runnable and t to thread

node/src/main/kotlin/net/corda/node/utilities/NamedThreadFactory.kt
blsemo: Agreed that adding a comment is a good idea. 
I just put the numbering in there as it doesn't really cost much and is in line with the default thread factory. After all, you could use this thread factory with any other executor that might create more than on thread.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -28,13 +28,22 @@ interface NodeConfiguration : NodeSSLConfiguration {
...
     val certificateSigningService: URL
     val certificateChainCheckPolicies: List<CertChainPolicyConfig>
     val verifierType: VerifierType
     val messageRedeliveryDelaySeconds: Int
     val notary: NotaryConfig?
     val activeMQServer: ActiveMqServerConfiguration
     val additionalNodeInfoPollingFrequencyMsec: Long
+
+    companion object {
+        val disableCheckpointCheckerFlag = "disableCheckpointChecker"
```
al-r3: make it const (I think the only difference is that you can use it in annotation params)
and make the name ALL_CAPS

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -103,6 +113,7 @@ data class FullNodeConfiguration(
     init {
         // This is a sanity feature do not remove.
         require(!useTestClock || devMode) { "Cannot use test clock outside of dev mode" }
+        require(devModeOptions == null || devMode){ "Cannot use devModeOptions outside of dev mode" }
```
rick-r3: Whitespace between `)` and `{`

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -28,13 +28,22 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val database: Properties?
     val rpcUsers: List<User>
     val devMode: Boolean
+    val devModeOptions: Properties?
```
shamsasari: Why not simply have `devMode` be an optional data class like `val devMode: DevModeOptions?`
```
data class DevModeOptions(val disableCheckPointChecker: Boolean)
```

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -103,6 +113,7 @@ data class FullNodeConfiguration(
     init {
         // This is a sanity feature do not remove.
         require(!useTestClock || devMode) { "Cannot use test clock outside of dev mode" }
+        require(devModeOptions == null || devMode) { "Cannot use devModeOptions outside of dev mode" }
```
shamsasari: See previous comment

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
rick-r3: I forgot it will recurse and parse into any `data class`.  We should make the flag `Boolean?`, I think, so that we don't need to specify all dev flags in the config file every time we want to set one of them (should we add more flags in the future).

PR #1951, opened 2017-10-26
CORDA-742 Mrizzo fix capsule error in deploy nodes

mat-rizzo-r3:
This issue ONLY exhibits itself on the TC build agents.

### DIFF
--- a/gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
+++ b/gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
@@ -148,42 +148,48 @@ open class Cordform : DefaultTask() {
         installNodeInfos()
     }
 
-
     private fun generateNodeInfos() {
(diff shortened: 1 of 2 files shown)


gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -192,13 +215,20 @@ open class Cordform : DefaultTask() {
...
-                            into(fullNodePath(destination).resolve(CordformNode.NODE_INFO_DIRECTORY).toString())
+                            into(destination.fullPath().resolve(CordformNode.NODE_INFO_DIRECTORY).toString())
                         }
                     }
                 }
             }
         }
     }
+
+    private const val capsuleCacheDir: String = "./cache"
```
Clintonio: Could you move all of these Node extensions to the Node object itself, as well as the capsule cache dir value which is a property of the nodes.

PR #1954, opened 2017-10-27
Retire FullNodeConfiguration

al-r3:
Remove all references to FullNodeConfiguration and just use NodeConfiguration everywhere.


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -99,7 +98,7 @@ open class Node(configuration: FullNodeConfiguration,
     }
 
     override val log: Logger get() = logger
-    override val configuration get() = super.configuration as FullNodeConfiguration // Necessary to avoid init order NPE.
+    override val configuration get() = super.configuration  // Necessary to avoid init order NPE.
```
shamsasari: Is this overridden getter still needed?

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -168,13 +167,13 @@ open class Node(configuration: FullNodeConfiguration,
         return NodeMessagingClient(
                 configuration,
                 versionInfo,
-                serverAddress,
+                serverAddress!!,
```
shamsasari: It looks like the `if (messagingServerAddress != null)` check above is not smart casting away the nullability. If you fix that then these two double-bangs can go away.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -68,7 +76,16 @@ data class BridgeConfiguration(val retryIntervalMs: Long,
 
 data class ActiveMqServerConfiguration(val bridge: BridgeConfiguration)
 
-data class FullNodeConfiguration(
+fun Config.parseAsNodeConfiguration(): NodeConfiguration {
```
shamsasari: You can make this one line

node/src/test/kotlin/net/corda/node/services/config/NodeConfigurationTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import net.corda.testing.ALICE
+import net.corda.testing.node.MockServices.Companion.makeTestDataSourceProperties
+import net.corda.testing.node.MockServices.Companion.makeTestDatabaseProperties
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.junit.Test
+
+class NodeConfigurationTest {
+    @Test
+    fun `Artemis special characters not permitted in RPC usernames`() {
+        val map = mapOf<String, Any?>(
```
shamsasari: This test can be done simply by creating a NodeConfigurationImpl - no need to use the typesafe config API. In fact there's a TODO in NodeConfigurationImpl.init  to move the username checks to ArtemisMessagingServer. There you could probably just need to deal with `User` objects for test.

node/src/main/kotlin/net/corda/node/internal/Node.kt
al-r3: fixed, and removed that funny construct with 'with' & friends.

node/src/test/kotlin/net/corda/node/services/config/NodeConfigurationTest.kt
al-r3: I'd like NodeConfigurationImpl to be private, hence I don't want to use it in tests either.
Said that, I can't find any use of rpcUsers, unless I'm missing some reflection magic, there's an RPCUserService which gets somehow a list of Users, I'm not sure if that would be the place I should put that check in

shamsasari: If it's too much then just create a NodeConfigurationImpl directly. You can always make it internal and annotate it visiblefortesting.

al-r3: found where it was used, (which was actually RPCUserServiceImpl). Moved all the logic and test logic there

node/src/main/kotlin/net/corda/node/internal/Node.kt
al-r3: removed (I was scared by the comment)

PR #1955, opened 2017-10-27
Refactor Kryo into sub module just like AMQP

fenryka:
Prior to enabling AMQP RPC move Kryo code to it's own submodule just like AMQP

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
@@ -7,6 +7,7 @@ import net.corda.core.messaging.CordaRPCOps
 import net.corda.core.utilities.NetworkHostAndPort
 import net.corda.nodeapi.ArtemisTcpTransport.Companion.tcpTransport
 import net.corda.nodeapi.ConnectionDirection
+import net.corda.nodeapi.internal.serialization.AMQP_RPC_CLIENT_CONTEXT
 import net.corda.nodeapi.internal.serialization.KRYO_RPC_CLIENT_CONTEXT
 import java.time.Duration
 
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
@@ -12,6 +12,7 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.serialization.SerializationContext
 import net.corda.core.utilities.loggerFor
 import net.corda.nodeapi.internal.serialization.amqp.hasAnnotationInHierarchy
+import net.corda.nodeapi.internal.serialization.kryo.ThrowableSerializer
 import java.io.PrintWriter
 import java.lang.reflect.Modifier.isAbstract
 import java.nio.charset.StandardCharsets
--- a/node/src/main/kotlin/net/corda/node/internal/Node.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/Node.kt
@@ -39,6 +39,7 @@ import net.corda.nodeapi.ConnectionDirection
 import net.corda.nodeapi.internal.ShutdownHook
 import net.corda.nodeapi.internal.addShutdownHook
 import net.corda.nodeapi.internal.serialization.*
+import net.corda.nodeapi.internal.serialization.amqp.AMQPServerSerializationScheme
 import org.apache.activemq.artemis.api.core.ActiveMQNotConnectedException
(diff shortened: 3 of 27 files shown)


client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -77,6 +78,7 @@ class CordaRPCClient @JvmOverloads constructor(
     private val rpcClient = RPCClient<CordaRPCOps>(
             tcpTransport(ConnectionDirection.Outbound(), hostAndPort, config = null),
             configuration.toRpcClientConfiguration(),
+            //AMQP_RPC_CLIENT_CONTEXT
```
andr3ej: commented code shouldn't really go into master

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerialisationScheme.kt
```diff
@@ -0,0 +1,132 @@
+package net.corda.nodeapi.internal.serialization.kryo
```
andr3ej: british english is fine in comments and docs but this filename should be US english for consistency with others

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -370,6 +371,11 @@ open class Node(configuration: FullNodeConfiguration,
             registerScheme(KryoServerSerializationScheme())
             registerScheme(AMQPServerSerializationScheme())
         }
+        /*
+        SerializationDefaults.P2P_CONTEXT = AMQP_P2P_CONTEXT.withClassLoader(classloader)
+        SerializationDefaults.RPC_SERVER_CONTEXT = AMQP_RPC_SERVER_CONTEXT.withClassLoader(classloader)
+        SerializationDefaults.STORAGE_CONTEXT = AMQP_STORAGE_CONTEXT.withClassLoader(classloader)
+        */
```
andr3ej: if this is effectively a todo, i don't think it's appropriate to use master for that

fenryka: You haven't seem master recently, it's full of TODO's,

mnesbit: I agree I don't like commented out code either. If we need a config switch lets make it one.

andr3ej: @fenryka master is probably not the best place to learn best practices

fenryka:
So the commented code is really only there because it's all going to change in the next commit given this is just a prelim to actually enabling AMQP, so meh, but people seem super keen on them going away now and not soon so I'll remove it

node/src/main/kotlin/net/corda/node/internal/Node.kt
fenryka: I learned those 15 years ago ;)

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerialisationScheme.kt
fenryka: that is a good point!

client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
fenryka: done

fenryka:
@mnesbit It's kinda the way it was done "at the time", ultimately given we're going to rip out Kryo except for storage the lack of matrix aplicability won't be too noticable

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
```diff
@@ -12,6 +12,7 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.serialization.SerializationContext
 import net.corda.core.utilities.loggerFor
 import net.corda.nodeapi.internal.serialization.amqp.hasAnnotationInHierarchy
+import net.corda.nodeapi.internal.serialization.kryo.ThrowableSerializer
```
rick-r3: Is this spurious?

fenryka: it is not :)

When ThrowableSEializer was in the higher level serialisation package we could just resolve it, now it's moved to a sub module we need to explicitly import it

rick-r3: Makes sense

PR #1956, opened 2017-10-27
Clean up of devModeOptions

shamsasari:


shamsasari:
Scaling back my changes. Making `testConfiguration` return NodeConfigurationImpl is not as simple as initially thought.

.idea/runConfigurations/BankOfCordaDriverKt___Issue_Web.xml
```diff
@@ -1,6 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="BankOfCordaDriverKt - Issue Web" type="JetRunConfigurationType" factoryName="Kotlin">
     <extension name="coverage" enabled="false" merge="false" sample_coverage="true" runner="idea" />
+    <extension name="krasa.grepconsole.plugin.runConfiguration.GrepRunConfigurationExtension" selectedProfileId="0" />
```
blsemo: Is adding this extension part of the clean-up? Is that used in IntelliJ or in some output?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -101,6 +97,8 @@ data class NodeConfigurationImpl(
         override val minimumPlatformVersion: Int = 1,
         override val rpcUsers: List<User>,
         override val verifierType: VerifierType,
+        // TODO typesafe config supports the notion of durations. Make use of that by mapping it to java.time.Duration.
```
blsemo: Shouldn't these TODO really be on the interface? Would be nice to have all proposed changes in one place, but not really important...

.idea/runConfigurations/BankOfCordaDriverKt___Issue_Web.xml
shamsasari: No, its just that I have this plugin installed and Intellij insists on updating this file. This is going to be an ongoing issue I think.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
shamsasari: Probably, but I'm not convinced yeah need the separate interface...

PR #1958, opened 2017-10-30
Add helper for checking whether to check checkpoints and unit test it.

blsemo:
Put check for checkpoint checking in a small helper function and test that

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
+++ b/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
@@ -49,7 +49,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val detectPublicIp: Boolean get() = true
 }
 
-fun NodeConfiguration.shouldCheckCheckpoints(): Boolean{
+fun NodeConfiguration.shouldCheckCheckpoints(): Boolean {
     return this.devMode && this.devModeOptions?.disableCheckpointChecker != true
 }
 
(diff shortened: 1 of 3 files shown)


node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -49,6 +49,10 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val detectPublicIp: Boolean get() = true
 }
 
+fun NodeConfiguration.shouldCheckCheckpoints(): Boolean{
```
shamsasari: space before {

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -49,6 +49,10 @@ interface NodeConfiguration : NodeSSLConfiguration {
...
+    return this.devMode && this.devModeOptions?.disableCheckpointChecker != true
```
shamsasari: I think the problem is with the check in NodeConfigurationImpl.init. Shouldn't it be `require((devModeOptions != null) == devMode)`?

blsemo: Not really - the whole point of having devModeOptions as an extra parameter rather than replacing the devMode flag was to keep existing configurations valid - but if you do the check like you proposed, a configuration with devMode == true but without specifying devModeOptions would be invalid. After all, we can't default the devMode options as they are supposed to be null in non-devMode.
We could drop the requirement to not break old configurations and replace devMode with devModeOptions.

shamsasari: If we did replace devMode with devModeOptions then who would we be breaking? Nothing deployed on testnet (I hope).

The problem I have is that `devMode &&` should be a redundant check. If we add anything to DevModeOptions then wherever it's used will also need to check for `devMode`, so it's a design smell. `devModeOptions == null` should imply `devMode == true`.

PR #1960, opened 2017-10-30
Upgraded to gradle 4.3

Clintonio:
https://github.com/gradle/gradle/releases/tag/v4.3.0


gradle-plugins/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -61,7 +61,8 @@ class CordappPlugin : Plugin<Project> {
         }
         filteredDeps.forEach {
             // net.corda or com.r3.corda.enterprise may be a core dependency which shouldn't be included in this cordapp so give a warning
-            if ((it.group.startsWith("net.corda.") || it.group.startsWith("com.r3.corda.enterprise."))) {
+            val group = it.group?.toString()
```
chrisr3: Maybe avoid the null-check?
```kotlin
val group = it.group?.toString() ?: ""
```

PR #1969, opened 2017-10-31
[CORDA-446] Remove mentions of networkmap node in DemoBench

al-r3:
Remove mentions of networkmap node from Demobench.

Note that Demobench used to check for a network map to decide whether one could launch a Notary or a non-notary node.
Now it checks for there being a running notary.


al-r3:
Yes, we don't support nodes removal right now and I don't know when we will.
Also only because a node is down it doesn't mean you can send a message to it, which should be sent to it when/if it comes up. To what extent this makes sense within Demobench I'm unclear.

PR #1970, opened 2017-10-31
Replace artemis network map with http network map

patrickkuo:


patrickkuo:
this is not ready for review yet, the PR is open to make sure it passes integration test.

patrickkuo:
This PR is ready for review.
Integration test will require modification in driver, I will raise another PR for that.

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -72,7 +73,7 @@ class HTTPNetworkMapClientTest {
         }
 
         val hostAndPort = server.connectors.mapNotNull { it as? ServerConnector }.first()
-        networkMapClient = HTTPNetworkMapClient("http://${hostAndPort.host}:${hostAndPort.localPort}/api/network-map")
+        networkMapClient = HTTPNetworkMapClient(URL("http://${hostAndPort.host}:${hostAndPort.localPort}"))
```
al-r3: Use the 3 param constructor of URL:
https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#URL(java.lang.String,%20java.lang.String,%20int,%20java.lang.String)

that should escape things as needed

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -232,6 +239,68 @@ abstract class AbstractNode(config: NodeConfiguration,
         }
     }
 
+    private fun startNetworkMapUpdate() {
```
al-r3: This code shouldn't be in AbstractNode, I'd rather have it as a free function close to where the rest of the network map code lives.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -232,6 +239,68 @@ abstract class AbstractNode(config: NodeConfiguration,
...
+                // Register node info if its changed.
+                val serialisedNodeInfo = info.serialize()
+                val signature = services.keyManagementService.sign(serialisedNodeInfo.bytes, info.legalIdentities.first().owningKey)
+                val signedNodeInfo = SignedData(serialisedNodeInfo, signature)
+                networkMapClient?.let {
+                    // Register with network map if its configured.
+                    while (true) {
+                        // Retry until successfully registered with the network map.
+                        it.publish(signedNodeInfo)
+                        try {
```
al-r3: I think this try should be 1 line up

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -232,6 +239,68 @@ abstract class AbstractNode(config: NodeConfiguration,
...
+            // Compare node info without timestamp.
+            // Only publish to network map server and write to disk if there are changes to the node info.
+            if (info.copy(serial = 0L) != oldInfo?.copy(serial = 0L)) {
+                // Register node info if its changed.
+                val serialisedNodeInfo = info.serialize()
+                val signature = services.keyManagementService.sign(serialisedNodeInfo.bytes, info.legalIdentities.first().owningKey)
+                val signedNodeInfo = SignedData(serialisedNodeInfo, signature)
+                networkMapClient?.let {
+                    // Register with network map if its configured.
+                    while (true) {
```
al-r3: This could loop forever, I'd have a maximum number of tries and then give up, and log an error message.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -232,6 +239,68 @@ abstract class AbstractNode(config: NodeConfiguration,
...
+                        toBeRemoved.forEach {
+                            services.networkMapCache.removeNode(it)
+                        }
+                        // Add new node info to the network map cache.
+                        toBeAdded.forEach {
+                            services.networkMapCache.addNode(it)
+                        }
+                        Thread.sleep(cacheTimeout)
+                    } catch (e: IllegalArgumentException) {
+                        log.warn("Error encountered while fetching network map update, will retry in $retryInterval", e)
```
al-r3: specify the unit of measure of $retryInterval in the error message

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -94,6 +95,10 @@ interface NetworkMapCacheBase {
     /** Return all [NodeInfo]s the node currently is aware of (including ourselves). */
     val allNodes: List<NodeInfo>
 
+    val allNodeHashes: List<SecureHash>
```
al-r3: can't this be easily computed by just getting all the nodes and then hashing them? (that way you need one less API)

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -94,6 +95,10 @@ interface NetworkMapCacheBase {
...
+
+    fun getNodeByHash(nodeHash: SecureHash): NodeInfo?
```
al-r3: Add a comment on what does this do

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -232,6 +239,68 @@ abstract class AbstractNode(config: NodeConfiguration,
...
+                    try {
+                        val (networkMap, cacheTimeout) = networkMapClient.getNetworkMap()
+                        val currentNodeHashes = services.networkMapCache.allNodeHashes
+                        val toBeAdded = networkMap.subtract(currentNodeHashes).mapNotNull {
+                            // Download new node info from network map
+                            networkMapClient.getNodeInfo(it)
+                        }
+                        val toBeRemoved = currentNodeHashes.subtract(networkMap).mapNotNull { services.networkMapCache.getNodeByHash(it) }
+                        // Remove node info from network map.
+                        toBeRemoved.forEach {
```
al-r3: There might be a corner case here.
If a node changes, say it's port name, we'll see a different hash and first call remove on the "oldNode" and then add the "newNode"
This will make the netmapcache trigger first a MapChange.Removed and then immediately a MapChange.Added, whereas it should trigger single MapCange.Modified event.


Also if we load NodeInfos from disk this will wipe them entirely at the first run.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: All of this code has no tests, please add some

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
patrickkuo: no, that will require loading everything from database and serialising them to get the hash, which is expensive.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
patrickkuo: testing this will require removing hard wired nodeInfo copying from the driver, which will be in a separate PR, as mentioned in PR comment

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
patrickkuo: we want it to loop forever incase of network map registration outage, however the registration shouldn't block getting network map update, will add some code to handle that.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
patrickkuo: retryInterval is `Duration`, the to string method handles that

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
patrickkuo: starting network map registration and update is part of the node start up sequences

al-r3: What you're saying is that It should be called by AbstractNode, which it is.
That is not a good enough reason to keep it within the AbstractNode class, arguing like that you'd copy half of the codebase there.

Also if this is a separate function it's easier to test.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -23,19 +22,20 @@ import kotlin.streams.toList
...
+ * @param nodePath the base path of a node.
+ * @param pollFrequency how often to poll the filesystem in milliseconds. Any value smaller than 5 seconds will
  *        be treated as 5 seconds.
  * @param scheduler a [Scheduler] for the rx [Observable] returned by [nodeInfoUpdates], this is mainly useful for
  *        testing. It defaults to the io scheduler which is the appropriate value for production uses.
  */
+// TODO: Use NIO watch service instead?
 class NodeInfoWatcher(private val nodePath: Path,
-                      pollFrequencyMsec: Long = 5.seconds.toMillis(),
+                      pollFrequency: Long = 5.seconds.toMillis(),
```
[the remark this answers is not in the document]
shamsasari: Use `Duration`

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -67,4 +61,90 @@ class HTTPNetworkMapClient(private val networkMapUrl: String) : NetworkMapClient
...
+}
+
+data class NetworkMapResponse(val networkMap: List<SecureHash>, val cacheMaxAge: Long)
+
+class NetworkMapUpdater(private val serviceHub: ServiceHubInternal, private val networkMapClient: NetworkMapClient?) {
+    companion object {
+        val logger = loggerFor<NetworkMapUpdater>()
+    }
+
+    fun startUpdate() {
```
shamsasari: This is the wrong approach. Instead there should be an `updateNodeInfo` method somewhere (in AbstractNode or ServiceHubInternal?) which other areas of the node can call if they need to update the node info. The method does some checks like the `serial` is later than the current one, etc and then tries to publish it to the network map. If it succeeds then it updates its database and overwrites the node-info on disk.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -67,4 +61,90 @@ class HTTPNetworkMapClient(private val networkMapUrl: String) : NetworkMapClient
...
+        // TODO: Make this configurable?
+        val defaultRetryInterval = 1.minutes
+        // Compare node info without timestamp.
+        if (newInfo.copy(serial = 0L) != oldInfo?.copy(serial = 0L)) {
+            // Only publish and write to disk if there are changes to the node info.
+            // This method is non-blocking.
+            tryPublish(defaultRetryInterval, newInfo)
+        }
+        val fileWatcher = NodeInfoWatcher(serviceHub.configuration.baseDirectory, serviceHub.configuration.additionalNodeInfoPollingFrequencyMsec)
+        fileWatcher.nodeInfoUpdates().subscribe { node -> serviceHub.networkMapCache.addNode(node) }
```
shamsasari: It doesn't make sense to use both the additional-node-infos *and* the network map. The node should only poll the directory if it hasn't been setup with the czURL. Better to keep the two sources separate.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -67,4 +61,90 @@ class HTTPNetworkMapClient(private val networkMapUrl: String) : NetworkMapClient
...
+            thread(name = "Network Map Update Thread") {
+                while (true) {
+                    try {
+                        val (networkMap, cacheTimeout) = networkMapClient.getNetworkMap()
+                        val currentNodeHashes = serviceHub.networkMapCache.allNodeHashes
+                        val toBeAdded = networkMap.subtract(currentNodeHashes).mapNotNull {
+                            // Download new node info from network map
+                            networkMapClient.getNodeInfo(it)
+                        }
+                        val toBeRemoved = currentNodeHashes.subtract(networkMap).mapNotNull { serviceHub.networkMapCache.getNodeByHash(it) }
```
shamsasari: This won't work if node infos can come from the node infos directory as well. We'd need to add some sort of concept of origin. But why bother with all that, better to keeps things simple as previously stated.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -67,4 +61,90 @@ class HTTPNetworkMapClient(private val networkMapUrl: String) : NetworkMapClient
...
+                        // Remove node info from network map.
+                        toBeRemoved.forEach {
+                            serviceHub.networkMapCache.removeNode(it)
+                        }
+                        // Add new node info to the network map cache.
+                        toBeAdded.forEach {
+                            serviceHub.networkMapCache.addNode(it)
+                        }
+                        Thread.sleep(cacheTimeout)
```
shamsasari: Use a ScheduledExecutorService

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -67,4 +61,90 @@ class HTTPNetworkMapClient(private val networkMapUrl: String) : NetworkMapClient
...
+        }
+    }
+
+    private fun maybeSubscriptToNetworkMap(retryInterval: Duration) {
+        if (networkMapClient != null) {
+            thread(name = "Network Map Update Thread") {
+                while (true) {
+                    try {
+                        val (networkMap, cacheTimeout) = networkMapClient.getNetworkMap()
+                        val currentNodeHashes = serviceHub.networkMapCache.allNodeHashes
```
shamsasari: `allNodeHashes` is not a public API. You might need a NetworkMapCacheInternal interface

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -67,4 +61,90 @@ class HTTPNetworkMapClient(private val networkMapUrl: String) : NetworkMapClient
...
+        thread(name = "Node Info Publish Thread") {
+            val serialisedNodeInfo = nodeInfo.serialize()
+            val signature = serviceHub.keyManagementService.sign(serialisedNodeInfo.bytes, nodeInfo.legalIdentities.first().owningKey)
+            val signedNodeInfo = SignedData(serialisedNodeInfo, signature)
+            if (networkMapClient != null) {
+                while (true) {
+                    // Retry until successfully registered with the network map.
+                    try {
+                        networkMapClient.publish(signedNodeInfo)
+                    } catch (e: IllegalArgumentException) {
```
shamsasari: Why this exception?

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -147,20 +165,22 @@ open class PersistentNetworkMapCache(private val database: CordaPersistence, con
                 }
             }
             val previousNode = registeredNodes.put(node.legalIdentities.first().owningKey, node) // TODO hack... we left the first one as special one
-            if (previousNode == null) {
-                logger.info("No previous node found")
-                database.transaction {
-                    updateInfoDB(node)
-                    changePublisher.onNext(MapChange.Added(node))
+            when {
```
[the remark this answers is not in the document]
shamsasari: I have to say I don't agree with the suggestion that Intellij gives for this. The previous way was fine.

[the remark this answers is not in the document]
patrickkuo: just bugged by the intellij highlight complaining "use when instead", can change it back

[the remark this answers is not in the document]
al-r3: I wouldn't take each and every suggestion from Intellij in general.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
patrickkuo: Can I assume cacheTimeout will never change after the first request?

shamsasari: No, check the cache timeout on every HTTP header received. This way the network map can change the timeout with minimum fuss.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -67,4 +61,90 @@ class HTTPNetworkMapClient(private val networkMapUrl: String) : NetworkMapClient
             else -> throw IllegalArgumentException("Unexpected response code ${conn.responseCode}, response error message: '${conn.errorStream.bufferedReader().readLines()}'")
         }
     }
+
+    fun myPublicHostname(): String {
+        val nodeInfoURL = URL("$networkMapUrl/my-hostname")
+        val conn = nodeInfoURL.openConnection() as HttpURLConnection
+
+        return when (conn.responseCode) {
+            HttpURLConnection.HTTP_OK -> conn.inputStream.readBytes().deserialize()
```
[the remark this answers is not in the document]
patrickkuo: tried the suggested doHttp method but didn't reduce much code, I have removed the throws Exception in the code and let conn.inputStream throws IOException instead, if the response code not = 200.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -93,6 +94,10 @@ interface NetworkMapCacheBase {
     /** Return all [NodeInfo]s the node currently is aware of (including ourselves). */
     val allNodes: List<NodeInfo>
 
+    val allNodeHashes: List<SecureHash>
```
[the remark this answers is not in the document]
patrickkuo: i have moved these to internal

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -242,16 +245,12 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
         }
-
         info = NodeInfo(
                 myAddresses(),
                 setOf(identity, myNotaryIdentity).filterNotNull(),
                 versionInfo.platformVersion,
                 platformClock.instant().toEpochMilli()
         )
-
-        NodeInfoWatcher.saveToFile(configuration.baseDirectory, info, identityKeyPair)
```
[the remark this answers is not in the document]
patrickkuo: it will now write to file regardless of the registration, in `NetworkMapUpdater.subscriptToNetworkMap`

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+
+        // This will throw IOException if the response code is not HTTP 200.
+        conn.inputStream
+    }
+
+    fun getNetworkMap(): NetworkMapResponse {
+        val conn = networkMapUrl.openHTTPConnection()
+        val response = conn.inputStream.bufferedReader().use { it.readLine() }
+        val networkMap = ObjectMapper().readValue(response, List::class.java).map { SecureHash.parse(it.toString()) }
+        val timeout = conn.headerFields["Cache-Control"]?.single { it.startsWith("max-age=") }?.removePrefix("max-age=")?.toLong() ?: throw IllegalArgumentException("Network map cache timeout is missing from the respond header.")
```
al-r3: We should use some standard libraries to parse http headers.
https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/CacheControl.html is a way, there might be better ones.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+                tryPublishNodeInfoAsync(signedNodeInfo, networkMapClient)
             }
-            else -> throw IllegalArgumentException("Unexpected response code ${conn.responseCode}, response error message: '${conn.errorStream.bufferedReader().readLines()}'")
         }
     }
 
-    override fun getNodeInfo(nodeInfoHash: SecureHash): NodeInfo? {
-        val nodeInfoURL = URL("$networkMapUrl/$nodeInfoHash")
-        val conn = nodeInfoURL.openConnection() as HttpURLConnection
+    fun subscriptToNetworkMap(serviceHub: ServiceHubInternal, networkMapClient: NetworkMapClient?) {
```
al-r3: subscribeToNetworkMap

al-r3: I think it's better to just take the networkMapCache and the nodeConf instead of a whole ServiceHub.
This makes it clear what parts of the code you're interacting with.

node/src/main/kotlin/net/corda/node/utilities/registration/HTTPNetworkRegistrationService.kt
```diff
@@ -21,7 +23,7 @@ class HTTPNetworkRegistrationService(val server: URL) : NetworkRegistrationServi
     @Throws(CertificateRequestException::class)
     override fun retrieveCertificates(requestId: String): Array<Certificate>? {
         // Poll server to download the signed certificate once request has been approved.
-        val url = URL("$server/api/certificate/$requestId")
+        val url = URL("$registrationURL/$requestId")
```
al-r3: Use https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriBuilder.html
manually assembling URLs is error-prone and doesn't do proper escaping

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -72,7 +73,7 @@ class HTTPNetworkMapClientTest {
         }
 
         val hostAndPort = server.connectors.mapNotNull { it as? ServerConnector }.first()
-        networkMapClient = HTTPNetworkMapClient("http://${hostAndPort.host}:${hostAndPort.localPort}/api/network-map")
+        networkMapClient = NetworkMapClient(URL("http://${hostAndPort.host}:${hostAndPort.localPort}"))
```
al-r3: use https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriBuilder.html

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -24,19 +23,20 @@ import kotlin.streams.toList
...
 class NodeInfoWatcher(private val nodePath: Path,
-                      pollFrequencyMsec: Long = 5.seconds.toMillis(),
+                      pollFrequency: Duration = 5.seconds,
                       private val scheduler: Scheduler = Schedulers.io()) {
 
     private val nodeInfoDirectory = nodePath / CordformNode.NODE_INFO_DIRECTORY
-    private val pollFrequencyMsec: Long = maxOf(pollFrequencyMsec, 5.seconds.toMillis())
-    private val successfullyProcessedFiles = mutableSetOf<Path>()
+    private val pollFrequency: Duration = if (pollFrequency < 5.seconds) 5.seconds else pollFrequency
+    private val processedNodeInfo = mutableSetOf<Path>()
```
al-r3: Why this variable name change? Also this is a collection of paths, so I think it should have path or file in its name

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+object NetworkMapUpdater {
+    private val logger = loggerFor<NetworkMapUpdater>()
+    private val retryInterval = 1.minutes
 
-        return when (conn.responseCode) {
-            HttpURLConnection.HTTP_OK -> {
-                val response = conn.inputStream.bufferedReader().use { it.readLine() }
-                ObjectMapper().readValue(response, List::class.java).map { SecureHash.parse(it.toString()) }
+    // This method store
+    fun updateNodeInfo(serviceHub: ServiceHubInternal, networkMapClient: NetworkMapClient?) {
```
al-r3: Just take a NodeInfo and a keyManagementService.
This will make this code easier to use (say if you will ever need it before there is a whole ServiceHub to pass in)

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
patrickkuo: currently corda node doesn't have dependancy on java-rs, it might be overkilled to added the dependancy just for parsing the header? @shamsasari what's your thought?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+                            // Add new node info to the network map cache, these could be new node info or modification of node info for existing nodes.
+                            serviceHub.networkMapCache.addNode(it)
+                        }
 
-        return when (conn.responseCode) {
-            HttpURLConnection.HTTP_OK -> conn.inputStream.readBytes().deserialize()
-            HttpURLConnection.HTTP_NOT_FOUND -> null
-            else -> throw IllegalArgumentException("Unexpected response code ${conn.responseCode}, response error message: '${conn.errorStream.bufferedReader().readLines()}'")
+                        // Remove node info from network map.
+                        serviceHub.networkMapCache.allNodeHashes.subtract(networkMap)
```
shamsasari: Extract `serviceHub.networkMapCache.allNodeHashes` into a val

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+        val fileWatcher = NodeInfoWatcher(serviceHub.configuration.baseDirectory, Duration.ofMillis(serviceHub.configuration.additionalNodeInfoPollingFrequencyMsec))
+        fileWatcher.nodeInfoUpdates().subscribe { node -> serviceHub.networkMapCache.addNode(node) }
+        // Subscribe to remote network map if configured.
+        if (networkMapClient != null) {
+            val executor = Executors.newSingleThreadScheduledExecutor(NamedThreadFactory("Network Map Update Thread", Executors.defaultThreadFactory()))
+            val task = object : Runnable {
+                override fun run() {
+                    try {
+                        val (networkMap, cacheTimeout) = networkMapClient.getNetworkMap()
+                        networkMap.subtract(serviceHub.networkMapCache.allNodeHashes).mapNotNull {
```
shamsasari: The `-` operator is nice and clear

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+                        // Remove node info from network map.
+                        serviceHub.networkMapCache.allNodeHashes.subtract(networkMap)
+                                .mapNotNull { serviceHub.networkMapCache.getNodeByHash(it) }
+                                .forEach { serviceHub.networkMapCache.removeNode(it) }
+
+                        // Schedule the next update.
+                        executor.schedule(this, cacheTimeout, TimeUnit.MILLISECONDS)
+                    } catch (t: Throwable) {
+                        logger.warn("Error encountered while updating network map, will retry in $retryInterval", t)
+                        executor.schedule(this, retryInterval.toMillis(), TimeUnit.MILLISECONDS)
```
shamsasari: Extract out the schedule from the try-catch.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+                    }
+                }
+            }
+            executor.submit(task) // The check may be expensive, so always run it in the background even the first time.
```
shamsasari: We need to capture the latest Future so that we can cancel it during node shutdown.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+        }
+    }
+
+    private fun tryPublishNodeInfoAsync(signedNodeInfo: SignedData<NodeInfo>, networkMapClient: NetworkMapClient) {
+        val executor = Executors.newSingleThreadScheduledExecutor(NamedThreadFactory("Node Info Publish Thread", Executors.defaultThreadFactory()))
```
shamsasari: Have one executor for both actions. Also this executor will need to be shutdown on node shutdown.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,134 @@ package net.corda.node.services.network
...
+        val task = object : Runnable {
+            override fun run() {
+                try {
+                    networkMapClient.publish(signedNodeInfo)
+                } catch (t: Throwable) {
+                    logger.warn("Error encountered while publishing node info, will retry in $retryInterval.", t)
+                    executor.schedule(this, retryInterval.toMillis(), TimeUnit.MILLISECONDS)
```
shamsasari: We probably want some sort of exponential backoff. Add a TODO.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
shamsasari: Perhaps `processedNodeInfoFiles` if you want to make it clear it's node info files we're dealing with.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
patrickkuo: do we need to keep track of individual futures or will `executor.shutdown`be enough? the only way to keep track is to use a var field variable but it looks odd to me.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
patrickkuo: the CacheControl we have is from okhttp3, which require some manipulation to the header map to make it parse without error. but i guess is better then fiddling with String.  `CacheControl.parse(Headers.of(conn.headerFields.filterKeys { it != null }.mapValues { (_, v) -> v.first() })).maxAgeSeconds()`

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
shamsasari: Yes you're right, executor shutdown is sufficient.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
shamsasari: I see three CacheControl classes in Intellij. Use the javax.ws one if you can.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,153 @@ package net.corda.node.services.network
...
+                            .forEach { networkMapCache.removeNode(it) }
 
-        return when (conn.responseCode) {
-            HttpURLConnection.HTTP_OK -> conn.inputStream.readBytes().deserialize()
-            HttpURLConnection.HTTP_NOT_FOUND -> null
-            else -> throw IllegalArgumentException("Unexpected response code ${conn.responseCode}, response error message: '${conn.errorStream.bufferedReader().readLines()}'")
+                    cacheTimeout
+                } catch (t: Throwable) {
+                    logger.warn("Error encountered while updating network map, will retry in $retryInterval", t)
+                    // TODO: Exponential backoff?
```
shamsasari: The TODO for exponential backoff is for tryPublishNodeInfoAsync. I don't think it makes sense to worry about it here since this should be the CDN.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -24,19 +23,20 @@ import kotlin.streams.toList
...
+// TODO: Use NIO watch service instead?
 class NodeInfoWatcher(private val nodePath: Path,
-                      pollFrequencyMsec: Long = 5.seconds.toMillis(),
+                      pollFrequency: Duration = 5.seconds,
                       private val scheduler: Scheduler = Schedulers.io()) {
 
     private val nodeInfoDirectory = nodePath / CordformNode.NODE_INFO_DIRECTORY
-    private val pollFrequencyMsec: Long = maxOf(pollFrequencyMsec, 5.seconds.toMillis())
-    private val successfullyProcessedFiles = mutableSetOf<Path>()
+    private val pollFrequency: Duration = if (pollFrequency < 5.seconds) 5.seconds else pollFrequency
```
shamsasari: I didn't notice this. Better to throw an exception instead.

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -90,18 +92,24 @@ class HTTPNetworkMapClientTest {
...
         val signedNodeInfo2 = createNodeInfo("Test2")
         val nodeInfo2 = signedNodeInfo2.verified()
         networkMapClient.publish(signedNodeInfo2)
 
         val nodeInfoHash2 = nodeInfo2.serialize().sha256()
-        assertThat(networkMapClient.getNetworkMap()).containsExactly(nodeInfoHash, nodeInfoHash2)
+        assertThat(networkMapClient.getNetworkMap().networkMap).containsExactly(nodeInfoHash, nodeInfoHash2)
+        assertEquals(100000.seconds, networkMapClient.getNetworkMap().cacheMaxAge)
         assertEquals(nodeInfo2, networkMapClient.getNodeInfo(nodeInfoHash2))
     }
```
shamsasari: You also need to test the case that the network map no longer includes a node info. And also test cases which involve node infos taken from the directory.

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
```diff
@@ -28,6 +29,7 @@ import org.junit.Test
 import java.io.ByteArrayInputStream
 import java.io.InputStream
 import java.net.InetSocketAddress
+import java.net.URL
 import java.security.cert.CertPath
 import java.security.cert.Certificate
 import java.security.cert.CertificateFactory
```
shamsasari: This class should be called `NetworkMapClientTest` to match the class under test

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,153 @@ package net.corda.node.services.network
...
         }
     }
 
-    override fun getNodeInfo(nodeInfoHash: SecureHash): NodeInfo? {
-        val nodeInfoURL = URL("$networkMapUrl/$nodeInfoHash")
-        val conn = nodeInfoURL.openConnection() as HttpURLConnection
+    fun subscribeToNetworkMap(config: NodeConfiguration, networkMapCache: NetworkMapCacheInternal, networkMapClient: NetworkMapClient?) {
+        // Subscribe to file based networkMap
+        val fileWatcher = NodeInfoWatcher(config.baseDirectory, Duration.ofMillis(config.additionalNodeInfoPollingFrequencyMsec))
+        fileWatcher.nodeInfoUpdates().subscribe { node -> networkMapCache.addNode(node) }
```
shamsasari: `networkMapCache::addNode`

shamsasari: You need to keep hold of the Subscription so that it can be unsubscribe from in `close`

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -141,14 +141,15 @@ open class NodeStartup(val args: Array<String>) {
     }
 
     open protected fun maybeRegisterWithNetworkAndExit(cmdlineOptions: CmdLineOptions, conf: NodeConfiguration) {
-        if (!cmdlineOptions.isRegistration) return
+        val compatibilityZoneURL = conf.compatibilityZoneURL
+        if (!cmdlineOptions.isRegistration || compatibilityZoneURL == null) return
```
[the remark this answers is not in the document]
patrickkuo: the new check only ensure when isRegistration is true, compatibilityZoneURL cannot be null, we still need to skip this method when isRegistration = false.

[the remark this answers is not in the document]
shamsasari: Yeah I had a feeling that's what it was.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -3,68 +3,153 @@ package net.corda.node.services.network
...
         }
     }
 
-    override fun getNetworkMap(): List<SecureHash> {
-        val conn = URL(networkMapUrl).openConnection() as HttpURLConnection
+    fun getNetworkMap(): NetworkMapResponse {
+        val conn = networkMapUrl.openHttpConnection()
+        val response = conn.inputStream.bufferedReader().use(BufferedReader::readLine)
+        val networkMap = ObjectMapper().readValue(response, List::class.java).map { SecureHash.parse(it.toString()) }
+        val timeout = CacheControl.parse(Headers.of(conn.headerFields.filterKeys { it != null }.mapValues { (_, v) -> v.first() })).maxAgeSeconds().seconds
```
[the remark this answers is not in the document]
patrickkuo: the response code is in the header with a null key, and okhttp3 doesn't like null in the map so have to filter it out, its not ideal....

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
patrickkuo: this class only test the rest client, its not testing the network map cache and the file watcher.
This should cover by NetworkMapUpdaterTest, I will see if I can create a unit test for this.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -184,7 +186,18 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
         val startedImpl = initialiseDatabasePersistence(schemaService) {
             val transactionStorage = makeTransactionStorage()
             val stateLoader = StateLoaderImpl(transactionStorage)
-            val services = makeServices(keyPairs, schemaService, transactionStorage, stateLoader)
+            val nodeServices = makeServices(keyPairs, schemaService, transactionStorage, stateLoader)
+
+            val networkMapUpdater = NetworkMapUpdater(configuration, services.networkMapCache, networkMapClient)
+                    .apply { runOnStop += this::close }
```
shamsasari: You already have the updater handle. I'm not sure the `apply` is giving you much.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -1,70 +1,155 @@
...
+        NodeInfoWatcher.saveToFile(config.baseDirectory, signedNodeInfo)
+        if (networkMapClient != null) {
+            tryPublishNodeInfoAsync(signedNodeInfo, networkMapClient)
+        }
+    }
+
+    fun subscribeToNetworkMap() {
+        // Subscribe to file based networkMap
+        val fileWatcher = NodeInfoWatcher(config.baseDirectory, Duration.ofMillis(config.additionalNodeInfoPollingFrequencyMsec))
+        fileWatcherSubscription?.unsubscribe()
```
shamsasari: Doesn't make sense to call this method twice. Simply check that this is null

node/src/main/resources/reference.conf
```diff
@@ -14,7 +14,6 @@ database = {
     initDatabase = true
 }
 devMode = true
-certificateSigningService = "https://cordaci-netperm.corda.r3cev.com"
```
shamsasari: Update the changelog

node/src/test/kotlin/net/corda/node/services/network/HTTPNetworkMapClientTest.kt
shamsasari: The tests?

patrickkuo: I have got the test in new branch together with the Integration test, i will raise another PR on top of this.

shamsasari: Would make more sense to include in this PR which introduced it?

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+import net.corda.testing.SerializationEnvironmentRule
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import kotlin.test.assertEquals
+
+class NetworkMapUpdaterTest {
+    @Rule
+    @JvmField
+    val tempFolder = TemporaryFolder()
```
shamsasari: Jimfs

shamsasari: Jimfs

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+            on { getNodeByLegalIdentity(any()) }.then { data[it.arguments.first()] }
+            on { allNodeHashes }.then { data.values.map { it.serialize().hash } }
+            on { getNodeByHash(any()) }.then { mock -> data.values.single { it.serialize().hash == mock.arguments.first() } }
+        }
+
+        val nodeInfoMap = mutableMapOf<SecureHash, SignedData<NodeInfo>>()
+        val networkMapClient = mock<NetworkMapClient> {
+            on { publish(any()) }.then {
+                val signedNodeInfo: SignedData<NodeInfo> = uncheckedCast(it.arguments.first())
+                nodeInfoMap.put(signedNodeInfo.verified().serialize().hash, signedNodeInfo)
```
shamsasari: Why not just create a `TestNetworkMapClient` that has all this implementation. The benefits of mocking seems little here.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        val nodeInfo1 = TestNodeInfoFactory.createNodeInfo("Info 1")
+        val nodeInfo2 = TestNodeInfoFactory.createNodeInfo("Info 2")
+        val nodeInfo3 = TestNodeInfoFactory.createNodeInfo("Info 3")
+        val nodeInfo4 = TestNodeInfoFactory.createNodeInfo("Info 4")
+        val fileNodeInfo = TestNodeInfoFactory.createNodeInfo("Info from file")
+
+        val networkMapCache = mock<NetworkMapCacheInternal> {
+            val data = mutableMapOf<Party, NodeInfo>()
```
shamsasari: See comment below.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        val signedSameNodeInfoDifferentTime = TestNodeInfoFactory.sign(keyPair, sameNodeInfoDifferentTime)
+
+        val differentNodeInfo = nodeInfo1.copy(addresses = listOf(NetworkHostAndPort("my.new.host.com", 1000)))
+        val signedDifferentNodeInfo = TestNodeInfoFactory.sign(keyPair, differentNodeInfo)
+
+        val networkMapCache = mock<NetworkMapCacheInternal> {
+            val data = mutableMapOf<Party, NodeInfo>()
+            on { addNode(any()) }.then {
+                val nodeInfo = it.arguments.first() as NodeInfo
+                data.put(nodeInfo.legalIdentities.first(), nodeInfo)
```
shamsasari: See comment below.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+            }
+            on { getNodeByLegalIdentity(any()) }.then { data[it.arguments.first()] }
+        }
+
+        val networkMapClient = mock<NetworkMapClient>()
+        val updater = NetworkMapUpdater(nodeConfig, networkMapCache, networkMapClient)
```
shamsasari: Shouldn't you also unit test the updater when it doesn't have a NetworkMapClient?

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,165 @@
...
+
+        updater.close()
+    }
+
+    @Test
+    fun `subscribe to network map`() {
+        val baseDir = tempFolder.root.toPath() / "networkMapUpdaterTest"
+        val nodeConfig = mock<NodeConfiguration> {
+            on { baseDirectory }.thenReturn(baseDir)
+            on { additionalNodeInfoPollingFrequencyMsec }.thenReturn(5.seconds.toMillis())
```
shamsasari: Would removing the 5 second minimum be good enough?

patrickkuo: mock will default it to 0 if its not provided

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -100,10 +99,13 @@ class NodeInfoWatcher(private val nodePath: Path,
             return emptyList()
         }
         val result = nodeInfoDirectory.list { paths ->
-            paths.filter { it !in successfullyProcessedFiles }
+            paths.filter { it !in processedNodeInfoFiles }
                     .filter { it.isRegularFile() }
                     .map { path ->
```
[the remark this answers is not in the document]
patrickkuo: paths is a Stream which doesn't have mapNotNull..... hence the toList as well

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
shamsasari: I meant, why not just reduce this frequency to some milliseconds

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
patrickkuo: NetworkMapClient is not a interface, i can make it a interface and implement the Test class but I prefer using mock to have different behaviour in different test case.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
patrickkuo: that will require removing the 5 second rule @al-r3 added, is that ok?

shamsasari: Yep. It can default to 5 sec

al-r3: @mnesbit asked for that in the PR in which I wrote this. Arguing that someone might incidentally configure that to a too low value and just spend all of the cpu rereading a folder which never changes.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,204 @@
...
+        // Publish different node info.
+        updater.updateNodeInfo(differentNodeInfo) { signedDifferentNodeInfo }
+        Thread.sleep(200)
+        verify(networkMapClient, times(1)).publish(signedDifferentNodeInfo)
+
+        updater.close()
+    }
+
+    @Test
+    fun `subscribe to network map`() {
```
shamsasari: subscribe to network map, with additional node infos from dir

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,204 @@
...
+
+        // Node info from file should not be deleted
+        assertEquals(1, networkMapCache.allNodeHashes.size)
+        assertEquals(fileNodeInfo.verified().serialize().hash, networkMapCache.allNodeHashes.first())
+
+        updater.close()
+    }
+
+    @Test
+    fun `subscribe to network map without network map client`() {
```
shamsasari: receive node infos from directory, without a network map

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,204 @@
...
+        val networkMapClient = mock<NetworkMapClient>()
+
+        val scheduler = TestScheduler()
+        val fileWatcher = NodeInfoWatcher(baseDir, scheduler = scheduler)
+        val updater = NetworkMapUpdater(networkMapCache, fileWatcher, networkMapClient)
+
+        // Publish node info for the first time.
+        updater.updateNodeInfo(nodeInfo1) { signedNodeInfo }
+        // Sleep as publish is asynchronous.
+        Thread.sleep(200)
```
shamsasari: I'm still not happy with the sleeps, can you add a TODO

mnesbit:
How does the timing of this interact with the bridge refresh at node restart?

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,204 @@
...
+        scheduler.advanceTimeBy(10, TimeUnit.SECONDS)
+        Thread.sleep(200)
+
+        // 4 node info from network map, and 1 from file.
+        verify(networkMapCache, times(5)).addNode(any())
+        verify(networkMapCache, times(1)).addNode(nodeInfo3.verified())
+        verify(networkMapCache, times(1)).addNode(nodeInfo4.verified())
+        verify(networkMapCache, times(1)).addNode(fileNodeInfo.verified())
+
+        // Test remove node.
```
shamsasari: I still think this should be a separate test. Also there is a lot of code duplication in this class, which could cut-down on the size of the test methods, which are big for unit tests.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
patrickkuo: same here, I will see if i can use the scheduler instead.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -1,70 +1,153 @@
...
+
+    private val executor = Executors.newSingleThreadScheduledExecutor(NamedThreadFactory("Network Map Updater Thread", Executors.defaultThreadFactory()))
+    private var fileWatcherSubscription: Subscription? = null
+
+    override fun close() {
+        fileWatcherSubscription?.unsubscribe()
+        MoreExecutors.shutdownAndAwaitTermination(executor, 50, TimeUnit.SECONDS)
+    }
+
+    fun updateNodeInfo(newInfo: NodeInfo, signNodeInfo: (NodeInfo) -> SignedData<NodeInfo>) {
```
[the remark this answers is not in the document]
patrickkuo: myNodeInfo is now added to the cache `networkMapCache.addNode(newInfo)`

patrickkuo:
@mnesbit Network map updates runs periodically after network map cache has been restored from the DB and also after `makeService` which creates the message broker, so any network map changes should be captured by the messaging service via network map cache.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -0,0 +1,242 @@
...
+import kotlin.test.assertEquals
+
+class NetworkMapUpdaterTest {
+    @Rule
+    @JvmField
+    val testSerialization = SerializationEnvironmentRule(true)
+
+    @Test
+    fun `publish node info`() {
+        val jimFs = Jimfs.newFileSystem(Configuration.unix())
```
shamsasari: You can move this and baseDir out, it's used in all tests

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
patrickkuo: I have added TODO for now as the changes is quite involved, will raise another PR for that.

PR #1972, opened 2017-10-31
Move StateLoader logic into ServicesForResolution

adagys:
As a consequence of https://github.com/corda/corda/pull/1927, loading output states for some transactions now requires resolving a full transaction – e.g. for contract upgrade transactions  outputs get generated by running contract upgrade code on the inputs. So StateLoader requires access to cordapp provider, tx storage and attachments.

I ended up moving StateLoader logic into ServicesForResolution, but that re-introduces the circular dependency between ServiceHub and NodeVaultService.. Not sure what's the best solution here.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
@@ -1,6 +1,5 @@
 package net.corda.core.node
 
-import net.corda.core.DoNotImplement
 import net.corda.core.contracts.*
 import net.corda.core.cordapp.CordappProvider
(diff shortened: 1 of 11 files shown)


andr3ej:
it's hard to review this without seeing the code that requires these changes. could the things that StateLoader(Impl) now needs be simply passed into it? the target design i have in mind (for CORDA-599) is that ServiceHub (and anything like it) is a facade for the benefit of users on top of objects with more specific purposes, such as StateLoader, and those objects form a graph rather than the flattened setup we have now. this will mean that some lower-level objects e.g. AttachmentStorage are assigned to fields in multiple places, but that's ok. @mnesbit may have opinions.

PR #1976, opened 2017-11-01
Move node info schema to internal package

patrickkuo:
NodeInfoSchema shouldn't be part of the public API, moving it to internal package and modified api-current.txt.

This changes is  needed to allow new ``node_info_hash`` column to be added for the network map redesign work.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -72,6 +72,9 @@ UNRELEASED
 * The ``ReceiveTransactionFlow`` can now be told to record the transaction at the same time as receiving it. Using this
   feature, better support for observer/regulator nodes has been added. See :doc:`tutorial-observer-nodes`.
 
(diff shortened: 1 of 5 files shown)


shamsasari:
Update the change log which explains this change.

docs/source/changelog.rst
```diff
@@ -71,6 +71,8 @@ UNRELEASED
 * The ``ReceiveTransactionFlow`` can now be told to record the transaction at the same time as receiving it. Using this
   feature, better support for observer/regulator nodes has been added. See :doc:`tutorial-observer-nodes`.
 
+* Moved ``NodeInfoSchema`` to internal package as it is not part of the public API.
```
shamsasari: Elaborate a bit more. The node info db schema is not part of the public API.

patrickkuo:
@rick-r3 can I have you blessing as well for this changes?

PR #1978, opened 2017-11-01
Introduction of network parameters.

shamsasari:
Note: This work was mostly done by @kasiastreich. I took over as she's on holiday. I've made changes of my own off of her work.

Introducing the concept of network parameters, which is a list of parameters which all nodes on the same network must abide by. For now this is a file in the node's directory which is read on startup. 

Most of the parameters are not used yet. The one that is is the list of notaries. The network map cache uses the network parameters to get the list of notaries (including whether they're validating or not). This means the list of notaries is fix at the network level, and so that is mirrored in the driver and MockNetwork as well. For these you have to specify a list of notary specs and accordingly notary nodes will be started up automatically. The various `startNotaryNode` methods have been removed as it would cause confusion since these nodes won't be visible as notaries to the other members since they won't be part of the network parameters (the idea of updating network parameters is not handled in this PR). The bulk of the changes are related to this.

This PR also removes the second notary identity for nodes running as single notaries. For now only nodes which are part of a distributed notary will have a second identity, which is the composite identity of the notary cluster.

There's a whole bunch still not implemented, such as network parameter updates, downloading it from the doorman on first startup, actually making use of the other parameters, etc. These are separate pieces of work.

There are things missing in this PR which should have been addressed but will be in a separate PR so to unblock several people. These are:
* docs
* gradle/demos

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
@@ -16,6 +16,7 @@ import net.corda.core.identity.Party
 import net.corda.core.internal.*
 import net.corda.core.internal.concurrent.*
 import net.corda.core.messaging.CordaRPCOps
+import net.corda.core.node.NetworkParameters
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.NotaryInfo
 import net.corda.core.node.services.NetworkMapCache
@@ -33,13 +34,12 @@ import net.corda.nodeapi.NodeInfoFilesCopier
 import net.corda.nodeapi.User
 import net.corda.nodeapi.config.toConfig
 import net.corda.nodeapi.internal.addShutdownHook
-import net.corda.testing.ALICE
-import net.corda.testing.BOB
-import net.corda.testing.DUMMY_BANK_A
+import net.corda.testing.*
 import net.corda.testing.common.internal.NetworkParametersCopier
 import net.corda.testing.common.internal.testNetworkParameters
-import net.corda.testing.initialiseTestSerialization
+import net.corda.testing.node.ClusterSpec
 import net.corda.testing.node.MockServices.Companion.MOCK_VERSION_INFO
+import net.corda.testing.node.NotarySpec
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import org.slf4j.Logger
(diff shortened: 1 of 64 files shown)


shamsasari:
I'll introduce a `defaultNotary` method.

samples/notary-demo/src/main/kotlin/net/corda/notarydemo/RaftNotaryCordform.kt
```diff
@@ -58,6 +57,6 @@ class RaftNotaryCordform : CordformDefinition("build" / "notary-demo-nodes") {
     }
 
     override fun setup(context: CordformContext) {
-        ServiceIdentityGenerator.generateToDisk(notaryNames.map { context.baseDirectory(it.toString()) }, clusterName)
+//        ServiceIdentityGenerator.generateToDisk(notaryNames.map { context.baseDirectory(it.toString()) }, clusterName)
```
mkit: Can we remove it?

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -67,19 +73,26 @@ import kotlin.concurrent.thread
...
+/**
  * This is the interface that's exposed to DSL users.
  */
 interface DriverDSLExposedInterface : CordformContext {
+    /** Returns a list of [NotaryHandle]s matching the list of [NotarySpec]s passed into [driver]. */
+    val notaries: List<NotaryHandle>
+
     /**
-     * Starts a [net.corda.node.internal.Node] in a separate process.
+     * Start a node.
```
mkit: Does it mean it is no longer started in a separate process?

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -893,7 +918,7 @@ class DriverDSL(
     private fun allNodesConnected(rpc: CordaRPCOps): CordaFuture<Int> {
         val (snapshot, updates) = rpc.networkMapFeed()
         val counterObservable = nodeCountObservable(snapshot.size, updates)
-        countObservables.put(rpc.nodeInfo().legalIdentities.first().name, counterObservable)
+        countObservables[rpc.nodeInfo().legalIdentities[0].name] = counterObservable
```
mkit: Is there a reason why we replace "first()" with "[0]"?

tools/explorer/src/main/kotlin/net/corda/explorer/ExplorerSimulation.kt
```diff
@@ -67,30 +65,23 @@ class ExplorerSimulation(val options: OptionSet) {
...
+            val notaryNode = notaries[0].nodeHandles.get()[0]
             aliceNode = alice.get()
             bobNode = bob.get()
             issuerNodeGBP = issuerGBP.get()
             issuerNodeUSD = issuerUSD.get()
 
-            arrayOf(notaryNode, aliceNode, bobNode, issuerNodeGBP, issuerNodeUSD).forEach {
-                println("${it.nodeInfo.legalIdentities.first()} started on ${it.configuration.rpcAddress}")
+            listOf(notaryNode, aliceNode, bobNode, issuerNodeGBP, issuerNodeUSD).forEach {
+                println("${it.nodeInfo.legalIdentities[0]} started on ${it.configuration.rpcAddress}")
```
mkit: Here again [0] instead of first()?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -641,22 +656,16 @@ abstract class AbstractNode(config: NodeConfiguration,
     private fun obtainIdentity(notaryConfig: NotaryConfig?): PartyAndCertificate {
         val keyStore = KeyStoreWrapper(configuration.nodeKeystore, configuration.keyStorePassword)
 
-        val (id, singleName) = if (notaryConfig == null) {
-            // Node's main identity
+        val (id, singleName) = if (notaryConfig == null || (notaryConfig.bftSMaRt == null && notaryConfig.raft == null)) {
+            // Node's main identity or if it's single notary
```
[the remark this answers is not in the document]
shamsasari: Fixed

samples/notary-demo/src/main/kotlin/net/corda/notarydemo/RaftNotaryCordform.kt
shamsasari: Fixed

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: Whether the node is in-process or out depends on the parameter.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: Uses less characters and is more idiomatic Kotlin IMO

shamsasari:
I've introduced helper methods to the driver and MockNetwork for dealing with the notary in the default case. Back to you @mnesbit.

shamsasari:
To get around the failing tests, #1995 was created which introduced the concept of network parameters. This PR now just deals with automatically starting nodes for the driver and MockNetwork.

PR #1980, opened 2017-11-01
[CORDA-446] Remove entities from netmap cache & friends

al-r3:
Remove NetworkMapService and all implementations.
They are not used in the new NetworkMap implementation


core/src/test/kotlin/net/corda/core/flows/ContractUpgradeFlowTest.kt
```diff
@@ -127,7 +128,8 @@ class ContractUpgradeFlowTest {
 
     @Test
     fun `2 parties contract upgrade using RPC`() {
-        rpcDriver(initialiseSerialization = false) {
+        rpcDriver(initialiseSerialization = false,
+                portAllocation = PortAllocation.RandomFree) {
```
shamsasari: You need to undo this?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
```diff
@@ -1,16 +1,5 @@
 package net.corda.node.services.network
```
shamsasari: This file doesn't have anything in it.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -492,10 +491,10 @@ fun addressMustBeBoundFuture(executorService: ScheduledExecutorService, hostAndP
 
 /*
  * The default timeout value of 40 seconds have been chosen based on previous node shutdown time estimate.
- * It's been observed that nodes can take up to 30 seconds to shut down, so just to stay on the safe side the 40 seconds
+ * It's been observed that nodes can take up to 30 seconds to shut down, so just to stay on the safe side the 60 seconds
  * timeout has been chosen.
  */
-fun addressMustNotBeBound(executorService: ScheduledExecutorService, hostAndPort: NetworkHostAndPort, timeout: Duration = 40.seconds) {
+fun addressMustNotBeBound(executorService: ScheduledExecutorService, hostAndPort: NetworkHostAndPort, timeout: Duration = 60.seconds) {
```
shamsasari: revert

node/src/main/kotlin/net/corda/node/services/network/NetworkMapService.kt
al-r3: fair point :)

I've just deleted all the comments, they seems significantly outdated.

shamsasari:
Also remove the minimumPlatformVersion config. It's only used when the node was a network map

al-r3:
The minimumPlatformVersion is used at the startup in AbstractNode to compare the version in the config with the one in the manifest.
I think that sill makes sense (and is otherwise unrelated with the network map node)

shamsasari:
But that serves no purpose on its own. That config was the first network parameter and is now no longer relevant. I'll remove it my PR.

PR #1981, opened 2017-11-01
Re-introduce demo docs

rnicoll:
Create a new "Advanced" section to the documentation, and re-introduce demo docs split into the
simpler demos and advanced implementations. Simple demos are now under tutorials rather than quick start, and advanced demos are in the new "Advanced" section, so readers should encounter the demos at a more natural part of their learning experience.


rnicoll:
Reworked this heavily in a second commit, to remove content that's now in the readme files (sorry Joel, missed that). I'd rather we do keep documentation in the main pages, so users know why we have these samples and have an idea what to expect, and I'd also rather keep a page at the existing address for the `running-the-demos` page so that external links won't break.

@gendal I'm going to add you as a reviewer as Joel and I disagree on whether this is a good idea, and we'd appreciate your thoughts on whether this helps users understand why there's a samples directory and what's in it (as well as where to start), or overly complicates things.

gendal:
I guess there are a few questions we have to be clear on:

1) Is the distinction between "simpler" and "advanced" based on the complexity/sophistication of the code and/or concepts being demonstrated or based on how simple/hard they are to run/understand?
2) Is the trader demo really "advanced" by either measure in any case?
3) Do the samples (or the tutorials for that matter) showcase known good practice? ie if somebody took any of the "advanced" samples and used them as a guide to writing their own CorDapp, would we be happy with the result?

@joeldudleyr3 It seems from @rnicoll that there is a disagreement here. What is it?

joeldudleyr3:
@gendal The devrel team's view is that the samples are not useful as examples of how to develop on Corda:

- They often include esoteric finance terms (e.g. the SIMM Valuation demo had a wiki page explaining what a SIMM is)
- They have complex front-ends and other cruft that isn't core to understanding how a CorDapp works, and confuses the user

Putting this in an `Advanced` section is better than in the `Quickstart` section they were in, but we think we can do better in terms of creating examples such as https://www.corda.net/samples/. Each of the examples in the Samples usually showcases a single thing, with the minimum amount of cruft.

I'd rather put the observer demo there, along with perhaps the notary demo.

gendal:
ok @rnicoll...  seems that the older samples (with the exception, perhaps, of trader-demo) are not considered examples of good practice and are confusing.  So making it clear that people should not use them as guides feels important?   We obviously wouldn't want to lose them as _integration tests_ of course.  

If I'm being honest, I still don't understand the disagreement.

Is it: "Joel thinks they should be killed and if we can't do that, hide them in the attic somewhere.  Ross thinks they should be displayed just as prominently as all the other samples and hence actively promoted to newcomers as things to pay attention to / learn from".

Is that a fair characterisation?

rnicoll:
Sorry about the delay updating this - so... I think for the samples still in this PR (it's been heavily cut back), they're reasonably simple examples which are useful for developer to learn from.

I also feel we need more continuity in the documentation, simply ripping out the whole page means anyone who has used it before, has no context on where to look for the replacements. We may end up ripping the whole page out later, but I'd rather keep it in for a version or three, directing to the new documentation, then remove it.

gendal:
ok - so would we have agreement if we did the following:

1) Caveat these demos to say they're intended to show how to use certain features but not intended to be examples of best practice?
2) Signpost/link people to the broader set of samples that devrel maintain?

joeldudleyr3:
@gendal @rnicoll Yes, let's keep the demos around then. How about updating this page (https://docs.corda.net/building-a-cordapp-index.html) so that it has an `Examples` entry in the ToC that links to both the devrel samples and the demos. Just keen to keep them out of the `Quickstart` section to reduce initial mental load for new devs.

@rnicoll If you agree with those changes, are you happy to make them as part of this PR?

gendal:
sounds reasonable.  to summarise, I think we want:

1) The default path taken by new developers is the easiest/simplest one and we signpost them to something that gets them productive quickly
2) we distinguish between samples that showcase a specific feature (but which are not intended to be canonical from a good practices perspective) and those that people should use as examples of good practice for developing an app.

joeldudleyr3:
@gendal Re: your second point - it's not that either type of CorDapp is not displaying best practices, but rather that the demos are sometimes needlessly complex (e.g. a complex scenario such as SIMM, a complex front-end). I'd prefer to rework each demo into a minimal sample that only has the code required to demonstrate a given feature, and no more.

Agreed on the first point.

rnicoll:
I think this update now matches the agreed on solution, please do let me know if you need any more changes!

rnicoll:
How's this @joeldudleyr3 ?

joeldudleyr3:
@rnicoll Sorry to be a pain, but could you also link to corda.net/samples?

Maybe "...to a number of common designs could be implemented in a CorDapp.  The former can be found at https://www.corda.net/samples/. In particular, new developers should start with the :doc:`example CorDapps <tutorial-cordapp>`."

rnicoll:
How's that?

PR #1982, opened 2017-11-01
Checking that the smoke tests don't have the node on their classpath.

shamsasari:


core/src/smoke-test/kotlin/net/corda/core/cordapp/CordappSmokeTest.kt
```diff
@@ -40,7 +39,7 @@ class CordappSmokeTest {
 
     @Test
     fun `FlowContent appName returns the filename of the CorDapp jar`() {
-        val cordappsDir = (factory.baseDirectory(aliceConfig) / CordappLoader.CORDAPPS_DIR_NAME).createDirectories()
+        val cordappsDir = (factory.baseDirectory(aliceConfig) / "cordapps").createDirectories()
```
Clintonio: Can you collect these both into a single variable in this test class? Just in case we end up using this string more over time.

shamsasari:
Done.

PR #1983, opened 2017-11-01
[CORDA-757] Deflake  contractUpgradeFlowTests. Use random ports and allow a longer timeout for ports

al-r3:


al-r3:
Still working on it, it seems this causes some (relatively unrelated) tests to fail

PR #1984, opened 2017-11-01
Add demo for observer node

rnicoll:
Add a demo with an observer node which is required to observe all transactions containing a contract's states, to demonstrate how that can be enforced (i.e. for regulatory requirements).


rnicoll:
This resurrects the demo from #1918

rnicoll:
@joeldudleyr3 Conclusion on this was that it should be in a different repo, wasn't it?

mikehearn:
Is this PR still relevant? If so, should it be moved out into a separate repo (the IRS demo shows how to use the current support for observers)?

rnicoll:
Closing this, to move it to its own repo instead, as an independent sample

PR #1985, opened 2017-11-02
[CORDA-758]: Permissions are now checked for each RPC method.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-758

### DIFF
--- a/client/rpc/src/test/kotlin/net/corda/client/rpc/RPCPermissionsTests.kt
+++ b/client/rpc/src/test/kotlin/net/corda/client/rpc/RPCPermissionsTests.kt
@@ -1,7 +1,9 @@
 package net.corda.client.rpc
 
+import net.corda.core.messaging.CordaRPCOps
 import net.corda.core.messaging.RPCOps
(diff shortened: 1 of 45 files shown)


client/rpc/src/test/kotlin/net/corda/client/rpc/RPCPermissionsTests.kt
```diff
@@ -28,7 +32,7 @@ class RPCPermissionsTests : AbstractRPCTest() {
 
     class TestOpsImpl : TestOps {
         override val protocolVersion = 1
-        override fun validatePermission(str: String) = getRpcContext().requirePermission(str)
+        override fun validatePermission(str: String) { rpcContext().requirePermission(str) }
```
szymonsztuka: Maybe leave the assignment instead of {} ?

sollecitom: Doesn't compile, for now `requirePermission()` is fluent and returns the `RpcContext`.

PR #1988, opened 2017-11-02
CORDA-716 Rename one TestClock to DemoClock, and unduplicate code

andr3ej:


andr3ej:
@mnesbit i think this is as good as it gets. in particular, NodeSchedulerService likes things the way they are.

Issue #1935, opened 2017-10-24
Serialisation of large contract attachments causes OOM exception.


roger3cev:
If one has a`LedgerTransaction` on the stack when a flow is check-pointed, then the CorDapp JAR is serialised and persisted to disk. This is problematic because the JAR is copied during the serialisation process and if one has a large JAR file, then it's highly likely that the node will throw a OOM Exception. To date, CorDapp developers understand that they only need to produce one JAR and that JAR constitutes their CorDapp. However, with the above process, this cannot be the case going forward. 

Question: Is the intention that only the contract and state definition is ever included in the JAR? If so, then we very clearly need to make developers aware of this. I.e. putting your front-end in the same JAR is not recommended..!

mikehearn:
Do you  have a copy of the serialisation trace? E.g. the flow snapshot API will give one.

joeldudleyr3:
Fixed.

PR #1991, opened 2017-11-02
Serialisation of large contract attachments causes OOM exception

cburlinchon:
Fix for: https://github.com/corda/corda/issues/1935
Instead of serialising the whole contract attachment(CorDapp JAR), only serialise the attachment id and contract class name when we checkpoint

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
@@ -6,6 +6,7 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.serialization.SerializationToken
 import net.corda.core.serialization.SerializeAsToken
 import net.corda.core.serialization.SerializeAsTokenContext
+import java.util.*
 
 /**
  * Wrap an attachment in this if it is to be used as an executable contract attachment
@@ -27,11 +28,5 @@ class ContractAttachment(override val id: SecureHash, dataLoader: () -> ByteArra
 
     override fun equals(other: Any?) = other === this || other is ContractAttachment && other.id == this.id && other.contract == this.contract
 
-    override fun hashCode(): Int {
-        var result = super.hashCode()
-        result = 31 * result + id.hashCode()
-        result = 31 * result + contract.hashCode()
-        return result
-    }
-
+    override fun hashCode() = super.hashCode() + 31 * Objects.hash(id, contract)
 }
--- a/core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
@@ -1,32 +1,12 @@
 package net.corda.core.contracts
 
-import net.corda.core.crypto.SecureHash
-import net.corda.core.internal.AbstractAttachment
 import net.corda.core.serialization.CordaSerializable
-import net.corda.core.serialization.SerializationToken
-import net.corda.core.serialization.SerializeAsToken
-import net.corda.core.serialization.SerializeAsTokenContext
-import java.util.*
 
 /**
  * Wrap an attachment in this if it is to be used as an executable contract attachment
  *
- * @property id Attachment id
- * @param dataLoader Function to load attachment
+ * @property attachment The attachment representing the contract JAR
  * @property contract The contract name contained within the JAR
  */
 @CordaSerializable
-class ContractAttachment(override val id: SecureHash, dataLoader: () -> ByteArray, val contract: ContractClassName) : AbstractAttachment(dataLoader), SerializeAsToken {
-
-    private class Token(private val id: SecureHash, val contract: ContractClassName) : SerializationToken {
-        override fun fromToken(context: SerializeAsTokenContext) = ContractAttachment(id, context.attachmentDataLoader(id), contract)
-    }
-
-    override fun toToken(context: SerializeAsTokenContext): SerializationToken {
-        return Token(id, contract)
-    }
-
-    override fun equals(other: Any?) = other === this || other is ContractAttachment && other.id == this.id && other.contract == this.contract
-
-    override fun hashCode() = super.hashCode() + 31 * Objects.hash(id, contract)
-}
+class ContractAttachment(val attachment: Attachment, val contract: ContractClassName) : Attachment by attachment
--- a/core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
@@ -1,12 +1,37 @@
 package net.corda.core.contracts
 
+import net.corda.core.crypto.SecureHash
+import net.corda.core.internal.AbstractAttachment
 import net.corda.core.serialization.CordaSerializable
+import net.corda.core.serialization.SerializationToken
+import net.corda.core.serialization.SerializeAsToken
+import net.corda.core.serialization.SerializeAsTokenContext
 
 /**
  * Wrap an attachment in this if it is to be used as an executable contract attachment
  *
- * @property attachment The attachment representing the contract JAR
+ * @property id Attachment id
+ * @param dataLoader Function to load attachment
  * @property contract The contract name contained within the JAR
  */
 @CordaSerializable
-class ContractAttachment(val attachment: Attachment, val contract: ContractClassName) : Attachment by attachment
+class ContractAttachment(override val id: SecureHash, dataLoader: () -> ByteArray, val contract: ContractClassName) : AbstractAttachment(dataLoader), SerializeAsToken {
+
+    private class Token(private val id: SecureHash, val contract: ContractClassName) : SerializationToken {
+        override fun fromToken(context: SerializeAsTokenContext) = ContractAttachment(id, context.attachmentDataLoader(id), contract)
+    }
+
+    override fun toToken(context: SerializeAsTokenContext): SerializationToken {
+        return Token(id, contract)
+    }
+
+    override fun equals(other: Any?) = other === this || other is ContractAttachment && other.id == this.id && other.contract == this.contract
+
+    override fun hashCode(): Int {
+        var result = super.hashCode()
+        result = 31 * result + id.hashCode()
+        result = 31 * result + contract.hashCode()
+        return result
+    }
+
+}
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -243,7 +243,12 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
         val contractAttachments = (outputs + resolvedInputs.map { it.state }).map { Pair(it, resolveContractAttachment(it)) }
         val missingAttachments = contractAttachments.filter { it.second == null }
         return if (missingAttachments.isEmpty()) {
-            contractAttachments.map { ContractAttachment(resolveAttachment(it.second!!) ?: throw AttachmentResolutionException(it.second!!), it.first.contract) }
+            contractAttachments.map {
+                ContractAttachment(it.second as SecureHash, {
+                    val attachment = resolveAttachment(it.second!!) ?: throw AttachmentResolutionException(it.second!!)
+                    attachment.open().readBytes()
+                }, it.first.contract)
```
andr3ej: can you decompose "it"? e.g. contractAttachments.map { (goodNameForFirst, goodNameForSecond) ->

core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
```diff
@@ -1,12 +1,37 @@
...
+
+    override fun equals(other: Any?) = other === this || other is ContractAttachment && other.id == this.id && other.contract == this.contract
+
+    override fun hashCode(): Int {
+        var result = super.hashCode()
+        result = 31 * result + id.hashCode()
+        result = 31 * result + contract.hashCode()
+        return result
+    }
+
```
andr3ej: remove blank line, and i think the one before Token can also be removed (i.e. formatter won't put it back)

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
```diff
@@ -196,15 +199,34 @@ object DefaultKryoCustomizer {
...
+                    val attachmentStorage = kryo.serializationContext()!!.serviceHub.attachments
+                    val attachment = attachmentStorage.openAttachment(attachmentHash) ?: throw MissingAttachmentsException(listOf(attachmentHash))
+                    attachment.open().readBytes()
+                }) {
+                    override val id = attachmentHash
+                }
+
+                return ContractAttachment(lazyAttachment, input.readString())
+            } else {
+                val attachment = GeneratedAttachment(input.readBytesWithLength())
```
shamsasari: The context should always be there, right? It should be an error if it's not.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
```diff
@@ -196,15 +197,22 @@ object DefaultKryoCustomizer {
 
     private object ContractAttachmentSerializer : Serializer<ContractAttachment>() {
         override fun write(kryo: Kryo, output: Output, obj: ContractAttachment) {
-            val buffer = ByteArrayOutputStream()
-            obj.attachment.open().use { it.copyTo(buffer) }
-            output.writeBytesWithLength(buffer.toByteArray())
+            output.writeBytesWithLength(obj.attachment.id.bytes)
```
shamsasari: Minor optimisation: SHA-256 has to be 32 bytes so need no need to write out the length

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
```diff
@@ -196,15 +197,22 @@ object DefaultKryoCustomizer {
...
             output.writeString(obj.contract)
         }
 
         override fun read(kryo: Kryo, input: Input, type: Class<ContractAttachment>): ContractAttachment {
-            val attachment = GeneratedAttachment(input.readBytesWithLength())
-            return ContractAttachment(attachment, input.readString())
+            val attachmentHash = SecureHash.SHA256(input.readBytesWithLength())
+
+            val lazyAttachment = object : AbstractAttachment({
+                val attachmentStorage = kryo.serializationContext()!!.serviceHub.attachments
```
shamsasari: Move `attachmentStorage` and `attachment` outside of the lambda so that only `attachment` is captured and not the Kryo instance.

shamsasari: Instead of the `!!` do something like `val context = kryo.serializationContext() ?: throw IllegalStateException(...)`  so that we have a more meaningful error message.

shamsasari:
Do we have tests for this?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
cburlinchon: In regards to your comment of moving attachmentStorage out of lambda, I get:
```com.esotericsoftware.kryo.KryoException: java.lang.IllegalStateException: Was expecting to find CordaPersistence set on current thread: Thread[CheckpointChecker-1,5,main]```

So this happens in NodeAttachmentService, val attachment = currentDBSession().get(NodeAttachmentService.DBAttachment::class.java, id.toString())

Guessing Kryo is doing stuff on a separate thread and we don't have CordaPersistence on that

File: node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/ContractAttachmentSerializerTest.kt [hunk 6]
```diff
@@ -0,0 +1,104 @@
...
+
+
+    @Test
+    fun `throws when no serializationContext`() {
+        val contractAttachment = ContractAttachment(GeneratedAttachment(ByteArray(0)), DummyContract.PROGRAM_ID)
+        // don't pass context in serialize
+        val serialized = contractAttachment.serialize()
+
+        Assertions.assertThatThrownBy { serialized.deserialize() }
+                .isInstanceOf(KryoException::class.java).hasMessageContaining(
```
fenryka: needs to work for AMQP as well as Kryo as attachments will pass over the p2p context (we know this as we've talked on slack)

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
```diff
@@ -196,15 +198,23 @@ object DefaultKryoCustomizer {
 
     private object ContractAttachmentSerializer : Serializer<ContractAttachment>() {
```
fenryka: Write an AMQP version :)

shamsasari: Perhaps that should be a separate PR since this is resolving a specific issue.

verifier/src/main/kotlin/net/corda/verifier/Verifier.kt
```diff
@@ -98,6 +113,50 @@ class Verifier {
              * request received, see use of [context] in [main] method.
              */
             SerializationDefaults.P2P_CONTEXT = KRYO_P2P_CONTEXT
+
+            // Need a way to load ContractAttachments
+            val mockServiceHub = object : ServiceHub {
```
shamsasari: On second thought, if this is what's required to avoid a null check in the Serializer, let's stick with the null check!

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
```diff
@@ -196,15 +198,23 @@ object DefaultKryoCustomizer {
...
+            val attachmentStorage = context.serviceHub.attachments
+
+            val lazyAttachment = object : AbstractAttachment({
+                val attachment = attachmentStorage.openAttachment(attachmentHash) ?: throw MissingAttachmentsException(listOf(attachmentHash))
+                attachment.open().readBytes()
+            }) {
+                override val id = attachmentHash
+            }
+
+            return ContractAttachment(lazyAttachment, input.readString())
```
shamsasari: I would avoid inlining the readString like this. Best to put it after the attachmentHash. It's very easy to change the ordering of reads if they're inlined.

verifier/src/main/kotlin/net/corda/verifier/Verifier.kt
```diff
@@ -98,6 +113,50 @@ class Verifier {
...
+                override val attachments: AttachmentStorage = object : AttachmentStorage {
+                    override fun openAttachment(id: AttachmentId): Attachment? {
+                        return GeneratedAttachment(kotlin.ByteArray(0))
```
shamsasari: I don't understand, how does this work when serialising attachments to the verifier?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
fenryka: I'm fine with that, although it's just going to break when I merge my changes and KRYO goes away for storage and P2P but splitting that out into a second PR is fine

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/ContractAttachmentSerializerTest.kt
fenryka: you can see this failing when AMQP is enabled in the CI gate

fenryka: Since we want to do the AMQP work as a separate PR Mark that test as Kryo only (there's an annotation) and in the other PR bring it back in

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/DefaultKryoCustomizer.kt
```diff
@@ -196,15 +199,38 @@ object DefaultKryoCustomizer {
...
+                obj.attachment.open().use { it.copyTo(buffer) }
+                output.writeBytesWithLength(buffer.toByteArray())
+            }
             output.writeString(obj.contract)
         }
 
         override fun read(kryo: Kryo, input: Input, type: Class<ContractAttachment>): ContractAttachment {
-            val attachment = GeneratedAttachment(input.readBytesWithLength())
-            return ContractAttachment(attachment, input.readString())
+            if (kryo.serializationContext() != null) {
```
shamsasari: Does using `UseCase` not work here? And would be because the use case is set in the context and the verifier doesn't have one? Perhaps we need to add a TODO to introduce a use case for the verifier.

For the meantime can you comment serializationContext == null case.

PR #1992, opened 2017-11-02
 Claim agent's port before attempting to open it

andr3ej:


exFalso:
If this is to avoid port clashes between tests then I think we could use the check done by `addressMustNotBeBoundFuture` instead of poking the filesystem. That will also work for inter-test-run lingering bound ports

andr3ej:
is it safe to assume the gradle processes never overlap?

exFalso:
they do overlap (in terms of port pools), from what i understand gradle spawns a new process for each test task

andr3ej:
the advantage of this solution is it works even if Gradle Test Executor processes run concurrently. but i've now done a test and it looks like they run one at a time, so addressMustNotBeBoundFuture is sufficient. do you want to implement that in your pr or shall i change this one?

exFalso:
Perhaps change this one, this includes the parent dir traversal fix

exFalso:
@andr3ej is this still relevant?

andr3ej:
now that things are sort of working again, i feel like this change may hide new problems, so closing

PR #1994, opened 2017-11-02
Simplify inheritance trees of NetworkMap

al-r3:
Bestow some order upon the various *NetworkMap* classes:
 NetworkMapCache, NetworkMapCacheBase, NetworkMapCacheInternal, NetworkMapCacheBaseInternal, NetworkMapCacheImpl and PersistentNetWorkMapCache

The only meaningful piece of logic was in NetworkMapCacheImpl, NetworkMapCacheImpl contains just 2 pieces of logic:
1. getNodeByLegalIdentity, 
It was used just by CordaRPCOpsImpl.kt, that logic has been moved there.
2. the constructor:
It took a NetworkMapCache and an IdentityService and subscribed to the updades from the cache and  updated the IdentityService.
That logic now lives in registerNetworkMapUpdatesInIdentityService which is called by AbstractNode when it creates the NetmapCache

Now the inheritance tree is:
NetworkMapCache   (has all the methods used by the RPC)
       |
NetworkMapCacheInternal   (extends NetworkMapCache has add/remove node)
       |
PersistentNetworkMapCache  (concrete implementation used by Node)
There's also a MockNetworkMap extending PersistentNetworkMapCache
 
NetworkMapCache has been kept back not the make the API checker freak out.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
@@ -121,7 +121,7 @@ interface ServiceHub : ServicesForResolution {
      * from an authoritative service, and adds easy lookup of the data stored within it. Generally it would be initialised
      * with a specified network map service, which it fetches data from and then subscribes to updates of.
      */
-    val networkMapCache: NetworkMapCache
+    val networkMapCacheBase: NetworkMapCacheBase
 
     /**
      * INTERNAL. DO NOT USE.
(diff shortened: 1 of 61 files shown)


core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
```diff
@@ -121,7 +121,7 @@ interface ServiceHub : ServicesForResolution {
      * from an authoritative service, and adds easy lookup of the data stored within it. Generally it would be initialised
      * with a specified network map service, which it fetches data from and then subscribes to updates of.
      */
-    val networkMapCache: NetworkMapCache
+    val networkMapCacheBase: NetworkMapCacheBase
```
shamsasari: This cannot be renamed as it's part of public API now, and you've restricted the type which is also a incompatible change.

andr3ej:
why wasn't i added to this pr? i touched this area most recently

al-r3:
We've usually done reviews among ourselves in the network map team.
Anyway this will be postponed to until some other work affecting those classes has been completed.

mikehearn:
Hmm, this PR seems to be inactive now. It will need a rebase at any rate, so I'm closing this temporarily to free up CI resources. No implication as to the content is intended!

PR #1995, opened 2017-11-02
Introducing network parameters.

shamsasari:
network-parameters file read in by the node at startup, of which only the list of notaries is used. For now, the driver and MockNetwork have been updated to require notaries to be started first. This is so that the same set of network parameters can be defined for all the nodes.

CN in the legal name is not longer disallowed since it's no longer reserved for distributed notary names.

Single-node notaries now only have one identity, their main identity. Nodes part of a cluster continue to have two.

(Based off Kasia's work)

This PR is cutdown version of #1978, which has already been approved. That PR has issues in its tests, and rather spend time hunting them down this smaller PR should have no issues.

This PR doesn't have the driver and MockNetwork automatically starting notary nodes, they still have to be created, but must be created first. The plan is to rebase #1978 once this is merged.


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -138,19 +132,19 @@ open class Node(override val configuration: NodeConfiguration,
...
+        val advertisedAddress = info.addresses.single()
 
         printBasicNodeInfo("Incoming connection address", advertisedAddress.toString())
 
         return NodeMessagingClient(
                 configuration,
                 versionInfo,
                 serverAddress,
-                legalIdentity.owningKey,
+                info.legalIdentities[0].owningKey,
```
mkit: does `info.legalIdentities[0]` refers to a notary? If so, shouldn't we replace it with some helper function?

node/src/main/kotlin/net/corda/node/utilities/ServiceIdentityGenerator.kt
```diff
@@ -39,9 +38,8 @@ object ServiceIdentityGenerator {
         keyPairs.zip(dirs) { keyPair, dir ->
             val serviceKeyCert = X509Utilities.createCertificate(CertificateType.CLIENT_CA, issuer.certificate, issuer.keyPair, serviceName, keyPair.public)
             val compositeKeyCert = X509Utilities.createCertificate(CertificateType.CLIENT_CA, issuer.certificate, issuer.keyPair, serviceName, notaryKey)
-            val certPath = Files.createDirectories(dir / "certificates") / "distributedService.jks"
+            val certPath = (dir / "certificates").createDirectories() / "distributedService.jks"
```
mkit: distributedService.jks is used also in another place. Do you think we could have this name (or entire path) defined in a single place - e.g. a companion object of some sort?

node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
```diff
@@ -615,18 +615,25 @@ class TwoPartyTradeFlowTests(val anonymous: Boolean) {
             notaryNode: StartedNode<*>,
             vararg extraSigningNodes: StartedNode<*>): Map<SecureHash, SignedTransaction> {
 
+        val notaryParty = notaryNode.info.legalIdentities[0]
```
mkit: Again, shouldn't we have a helper function here? It is test though, so not sure about this though.

samples/network-visualiser/src/main/kotlin/net/corda/netmap/simulation/IRSSimulation.kt
```diff
@@ -140,7 +140,7 @@ class IRSSimulation(networkSendManuallyPumped: Boolean, runAsync: Boolean, laten
         node1.internals.registerInitiatedFlow(FixingFlow.Fixer::class.java)
         node2.internals.registerInitiatedFlow(FixingFlow.Fixer::class.java)
 
-        val notaryId = notary.info.legalIdentities[1]
+        val notaryId = notary.info.legalIdentities[0]
```
mkit: Handle function?

shamsasari:
Also removed the minimumPlatformVersion config, using the one retrieved from the network parameters instead.

core/src/main/kotlin/net/corda/core/node/services/NetworkMapCache.kt
```diff
@@ -116,12 +115,15 @@ interface NetworkMapCacheBase {
...
-    /** Checks whether a given party is an advertised notary identity. */
+    /** Returns true if and only if the given [Party] is a notary, which is defined by the network parameters. */
     fun isNotary(party: Party): Boolean = party in notaryIdentities
 
-    /** Checks whether a given party is an validating notary identity. */
-    // TODO This implementation will change after introducing of NetworkParameters.
-    fun isValidatingNotary(party: Party): Boolean = isNotary(party) && "validating" in party.name.commonName!!
+    /**
+     * Returns true if and only if the given [Party] is validating notary. For every party that is a validating notary,
+     * [isNotary] is only true.
```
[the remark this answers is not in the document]
shamsasari: It should be "also:

node/src/main/kotlin/net/corda/node/internal/Node.kt
shamsasari: Refers to the node's main identity, which is what NodeMessagingClient is asking for.

node/src/main/kotlin/net/corda/node/utilities/ServiceIdentityGenerator.kt
shamsasari: Maybe, but this identity generator is a crutch until we have a proper way to generate composite identities

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -135,7 +139,7 @@ class NotaryFlow {
         protected fun checkNotary(notary: Party?) {
             // TODO This check implies that it's OK to use the node's main identity. Shouldn't it be just limited to the
             // notary identities?
```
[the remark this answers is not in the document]
shamsasari: I'm not sure so I left it as is.

node/src/main/kotlin/net/corda/node/internal/Node.kt
mkit: Since we have this concept of a "main node identity", do you think it is worth of having a helper function that would be called something like "getNodeMainIdentity" that would encapsulate the whole logic of referring to the first item in the list?

shamsasari: This is the problem; will this concept of a "main node identity" stay when multiple identities is implemented? Or is that each identity is a valid first class identity and there is no main identity.

PR #1997, opened 2017-11-03
CORDA-716 Retire marker interface

andr3ej:
useless kdoc, nothing refers to it, can't tell what it does

### DIFF
--- a/samples/irs-demo/src/integration-test/kotlin/net/corda/irs/IRSDemoTest.kt
+++ b/samples/irs-demo/src/integration-test/kotlin/net/corda/irs/IRSDemoTest.kt
@@ -31,8 +31,7 @@ import rx.Observable
 import java.time.Duration
 import java.time.LocalDate
 
-class IRSDemoTest : IntegrationTestCategory {
-
+class IRSDemoTest {
     companion object {
         val log = loggerFor<IRSDemoTest>()
     }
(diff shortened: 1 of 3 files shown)


vkolomeyko:
Hm... this fails to compile for me in IntelliJ with:
```
Z:\corda\samples\simm-valuation-demo\src\integration-test\kotlin\net\corda\vega\SimmValuationTest.kt
Error:(8, 26) Kotlin: Unresolved reference: IntegrationTestCategory
```
Also on this PR I do not see TC checks passed...

andr3ej:
something wrong your end, you can expand the section that says "5 checks passed"

vkolomeyko:
Yes, Windows is "special" in many ways, but if we have a look at this file:
https://github.com/corda/corda/blob/master/samples/simm-valuation-demo/src/integration-test/kotlin/net/corda/vega/SimmValuationTest.kt
It still has (un-necessary) import of "IntegrationTestCategory", do you agree?

andr3ej:
i see what's going on here, it's nothing to do with windows. we've setup TC so it doesn't rebuild in response to upstream changes, so shams commit yesterday broke it https://github.com/corda/corda/commit/3bb018a5cef14ec8da1026b80f1a788aaaa749c1#diff-45a17da83ea5217a71a1bff415470041

andr3ej:
it's just an import so the problem will fix itself

mikehearn:
It was originally intended to allow integration tests to be excluded from unit test runs, but we ended up doing it a different way.

PR #2001, opened 2017-11-06
Moved ShutdownManager into an internal package.

shamsasari:
This meant having to also move the shutdownManager property into the interal driver DSL interface, which further meant also moving RPCDriver into an internal package (which makes sense as it's not a public testing API).


shamsasari:
Invalid assumption that ShutdownManager is internal

PR #2003, opened 2017-11-06
On exit, the driver will automaticallly shutdown any nodes which were…

shamsasari:
…n't waited for.

The motivation for this came with the recent change that a default notary is started by the driver, which if ignored will leak the notary process.

This change removes the need for waitForAllNodesToFinish().

### DIFF
--- a/samples/irs-demo/cordapp/src/test/kotlin/net/corda/irs/Main.kt
+++ b/samples/irs-demo/cordapp/src/test/kotlin/net/corda/irs/Main.kt
@@ -10,7 +10,7 @@ import net.corda.testing.driver.driver
  * Do not use in a production environment.
  */
 fun main(args: Array<String>) {
-    driver(useTestClock = true, isDebug = true) {
+    driver(useTestClock = true, isDebug = true, waitForAllNodesToFinish = true) {
         val (nodeA, nodeB) = listOf(
                 startNode(providedName = DUMMY_BANK_A.name),
                 startNode(providedName = DUMMY_BANK_B.name)
@@ -20,7 +20,5 @@ fun main(args: Array<String>) {
         startWebserver(controller)
         startWebserver(nodeA)
         startWebserver(nodeB)
-
-        waitForAllNodesToFinish()
     }
 }
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
@@ -37,7 +37,6 @@ import net.corda.nodeapi.internal.addShutdownHook
 import net.corda.testing.*
 import net.corda.testing.common.internal.NetworkParametersCopier
 import net.corda.testing.common.internal.testNetworkParameters
-import net.corda.testing.setGlobalSerialization
 import net.corda.testing.internal.ProcessUtilities
 import net.corda.testing.node.ClusterSpec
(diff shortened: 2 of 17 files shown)


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -724,12 +724,15 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
-            NetworkMapCacheImpl(
+            val cache = NetworkMapCacheImpl(
                     PersistentNetworkMapCache(
                             this@AbstractNode.database,
                             this@AbstractNode.configuration,
                             networkParameters.notaries),
-                    identityService)
+                    identityService
+            )
+            runOnStop += cache::close
```
exFalso: perhaps put this into a `makeNetworkMapCache()` mirroring the rest of these initialisations

samples/irs-demo/cordapp/src/test/kotlin/net/corda/irs/Main.kt
```diff
@@ -20,7 +20,5 @@ fun main(args: Array<String>) {
         startWebserver(controller)
         startWebserver(nodeA)
         startWebserver(nodeB)
-
-        waitForAllNodesToFinish()
```
exFalso: If we remove this won't the driver immediately shut down?

shamsasari: Doh, yes of course! I'll add it back, but then the method is at odds with this new feature. And also it's only meant to be called at the end of the dsl block? I wonder if it even makes sense to use the driver in these scenarios and instead use the new cordformation thing.

exFalso: Agreed. Hmm then these Main.kts should be removed. Perhaps ask devrel first whether these are used at all?

shamsasari: I added a `waitForNodesToFinish` parameter instead.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -396,6 +395,7 @@ data class DriverParameters(
     fun setUseTestClock(useTestClock: Boolean) = copy(useTestClock = useTestClock)
     fun setInitialiseSerialization(initialiseSerialization: Boolean) = copy(initialiseSerialization = initialiseSerialization)
     fun setStartNodesInProcess(startNodesInProcess: Boolean) = copy(startNodesInProcess = startNodesInProcess)
+    fun setTerminateNodesOnShutdown(terminateNodesOnShutdown: Boolean) = copy(waitForNodesToFinish = terminateNodesOnShutdown)
```
exFalso: Perhaps name this the same as the parameter for consistency

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -585,8 +588,9 @@ class DriverDSL(
     override val notaryHandles: List<NotaryHandle> get() = _notaries
     private lateinit var networkParameters: NetworkParametersCopier
 
+
```
exFalso: whitespace 🚓

shamsasari:
Comments addressed.

PR #2004, opened 2017-11-06
CORDA-296: added rpc that returns an observable for node state

bpaunescu:
A first step towards making the node shutdown and overall RPC client-to-node connection better.
RPC clients can subscribe to an observable and be notified when the node will shutdown intentionally.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -62,6 +62,7 @@ import net.corda.node.utilities.*
 import org.apache.activemq.artemis.utils.ReusableLatch
 import org.slf4j.Logger
 import rx.Observable
+import rx.subjects.PublishSubject
 import java.io.IOException
 import java.lang.reflect.InvocationTargetException
 import java.security.KeyPair
@@ -124,6 +125,7 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
     protected val services: ServiceHubInternal get() = _services
     private lateinit var _services: ServiceHubInternalImpl
     protected lateinit var info: NodeInfo
+    protected val nodeStateObservable: PublishSubject<String> = PublishSubject.create<String>()
     protected var myNotaryIdentity: PartyAndCertificate? = null
     protected lateinit var checkpointStorage: CheckpointStorage
     protected lateinit var smm: StateMachineManager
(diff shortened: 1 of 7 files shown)


bpaunescu:
I'm not overly happy with adding an RPC that simply returns an observable. Also, at the moment the node sends one observation as a string message when the shutdown starts. Any suggestions on a prettier way of doing it? I considered an enum but I can't think of any other observations that the node could send regarding to its state.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -636,6 +638,9 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
         // Meanwhile, we let the remote service send us updates until the acknowledgment buffer overflows and it
         // unsubscribes us forcibly, rather than blocking the shutdown process.
 
+        // Notify observers that the node is shutting down
+        nodeStateObservable.onNext("Node shutting down.")
```
rnicoll: I was thinking an enum of the possible states; at the least can put put this string into a `const val` in a companion object, so it's defined in one place that CorDapps/clients can reference, rather than others having to have their own copy of this string (and potentially therefore not match).

But ideally can we replace the string with an enum please?

bpaunescu: That's what I initially wanted as having strings like this one is a poor choice. I just couldn't decide what other states to put in the enum.

rnicoll:
@exFalso https://r3-cev.atlassian.net/browse/CORDA-296 specifically refers to a clean shutdown, so while I see your point, I'm inclined to go with this and then have a combined network/node events observable in a later PR.

exFalso:
@rnicoll if we do the proper shutdown handling later then at that time the merged code from this PR will need to be removed, a generic connection-state-changed observable is strictly better than an event about "clean shutdown".
Furthermore the proper shutdown observable won't live on the RPC interface (as it won't be a proxied event) so that'll be removed, and we'll be listening on network events on the client side so the `close()` hook as well as the PublishSubject on the node object will *also* be removed. So practically we'll have to remove all of this code.
Doing the proper disconnect detection should be quite straightforward, so why not do it properly from the get go?

mikehearn:
Can we re-review this please, possibly rolling it back? This PR adds a new API that we already plan to remove. Unless this new API is removed *very fast* there's a risk that we'll ship it *and then be committed to supporting it forever*.

Please do not add new APIs that you don't actually think are right or which are intended to be temporary! Can we undo this and fix it properly please?

PR #2008, opened 2017-11-06
Various structural cleanups of node-driver:

shamsasari:
* Extracted out ShutdownManager into its own file
* Moved RPCDriver and ProcessUtilities into internal package
* Made n.c.testing.performance package internal


testing/node-driver/src/main/kotlin/net/corda/testing/NodeTestUtils.kt
```diff
@@ -45,7 +45,6 @@ fun ledger(
  */
 @JvmOverloads
 fun transaction(
-        transactionLabel: String? = null,
```
andr3ej: yay!

testing/node-driver/src/main/kotlin/net/corda/testing/driver/ShutdownManager.kt
```diff
@@ -0,0 +1,113 @@
...
+import net.corda.core.utilities.getOrThrow
+import net.corda.core.utilities.loggerFor
+import net.corda.core.utilities.seconds
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicInteger
+
+class ShutdownManager(private val executorService: ExecutorService) {
```
andr3ej: can this be internal? i don't think users should be calling it directly

testing/node-driver/src/main/kotlin/net/corda/testing/driver/ShutdownManager.kt
```diff
@@ -0,0 +1,113 @@
...
+    private class State {
+        val registeredShutdowns = ArrayList<CordaFuture<() -> Unit>>()
+        var isShutdown = false
+    }
+
+    private val state = ThreadBox(State())
+
+    companion object {
+        val log = loggerFor<ShutdownManager>()
```
andr3ej: private

testing/node-driver/src/main/kotlin/net/corda/testing/driver/ShutdownManager.kt
shamsasari: I had created a PR to do exactly this, which also removed the `shutdownManager` property of the driver. But I subsequently realised it probably does make sense to keep it public, if you're starting 3rd party entities when using the driver which should be shutdown automatically. So I closed it. In any case it may need more discussion.

shamsasari:
Back to you.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/ShutdownManager.kt
andr3ej: i don't think it's driver's responsibility to (offer to) stop things it didn't start. anyway, another pr

PR #2009, opened 2017-11-06
Retired getDefaultNotary test extension method.

shamsasari:
Most uses where with MockNetwork which recently got a defaultNotaryIdentity property for dealing with the default single notary case. The remaining uses where in flows.

### DIFF
--- a/core/src/test/kotlin/net/corda/core/flows/CollectSignaturesFlowTests.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/CollectSignaturesFlowTests.kt
@@ -10,7 +10,6 @@ import net.corda.core.identity.groupAbstractPartyByWellKnownParty
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.transactions.TransactionBuilder
 import net.corda.core.utilities.getOrThrow
-import net.corda.core.utilities.unwrap
 import net.corda.node.internal.StartedNode
 import net.corda.testing.*
 import net.corda.testing.contracts.DummyContract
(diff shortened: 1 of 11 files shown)


core/src/test/kotlin/net/corda/core/flows/CollectSignaturesFlowTests.kt
```diff
@@ -87,7 +87,7 @@ class CollectSignaturesFlowTests {
             @Suspendable
             override fun call(): SignedTransaction {
                 val state = initiatingSession.receive<DummyContract.MultiOwnerState>().unwrap { it }
-                val notary = serviceHub.getDefaultNotary()
+                val notary = serviceHub.networkMapCache.notaryIdentities[0]
```
rnicoll: I get why it's being replaced where `MockNet.defaultNotaryIdentity` can be used, but is this an improvement?

shamsasari:
Resolved, back to you.

core/src/test/kotlin/net/corda/core/flows/CollectSignaturesFlowTests.kt
```diff
@@ -55,60 +54,14 @@ class CollectSignaturesFlowTests {
...
-                val stx = ptx + signature
-                return subFlow(FinalityFlow(stx))
-            }
-        }
-    }
-
     // With this flow, the initiator starts the "CollectTransactionFlow". It is then the responders responsibility to
     // override "checkTransaction" and add whatever logic their require to verify the SignedTransaction they are
     // receiving off the wire.
     object TestFlowTwo {
```
rnicoll: Can this be renamed to `TestFlow` as there's no other flow now, please.

shamsasari:
Done.

PR #2012, opened 2017-11-07
CORDA-770: Sync-up setting in "build.gradle"

vkolomeyko:


samples/bank-of-corda-demo/src/main/kotlin/net/corda/bank/BankOfCordaDriver.kt
```diff
@@ -56,35 +55,30 @@ private class BankOfCordaDriver {
...
-                                        startFlowPermission<CashExitFlow>(),
                                         startFlowPermission<CashIssueAndPaymentFlow>(),
-                                        startFlowPermission<CashConfigDataFlow>()
-                                ))
+                                        startFlowPermission<CashExitFlow>()))
+                        val bigCorpUser = User(BIGCORP_USERNAME, "test",
+                                permissions = setOf(
+                                        startFlowPermission<CashPaymentFlow>()))
+                        startNode(providedName = DUMMY_NOTARY.name,
+                                advertisedServices = setOf(ServiceInfo(SimpleNotaryService.type)))
```
mkit: Here we add SimpleNotaryService, however in line 58 we use validating version. I might be confusing things but shouldn't they be the same?

fenryka: I agree,

would make more sense to just rename the node in build.gradle to match what's expected by the dirver

vkolomeyko: Sure - will do that

PR #2016, opened 2017-11-08
[CORDA-760]: Propagate invocation context across the codebase.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-760

**This does not cover actions initiated by services.**


client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -444,12 +449,19 @@ object RpcClientObservableSerializer : Serializer<Observable<*>>() {
         }.dematerialize()
     }
 
+    private fun Input.readInvocationId() : InvocationId? {
+
+        val value = readString() ?: return null
```
szymonsztuka: The only invocation of the method shows (line 434) that returning null results in `IllegalStateException`. However the next line may throw different exception. Maybe instead of returning `null` throw this exception here. Also maybe catch exception from `readLong` and re-throw `IllegalStateException` to be consistent with exception case for readString()?

node/src/main/kotlin/net/corda/node/services/logging/ContextualLogger.kt
```diff
@@ -0,0 +1,229 @@
+package net.corda.node.services.logging
+
+import net.corda.core.context.InvocationContext
+import org.slf4j.Logger
+import org.slf4j.Marker
+
+// Wish we didn't expose org.slf4j.Logger from public API.
```
szymonsztuka: I understand your pain ;) but maybe a different description - something like /** Add Invocation Context log information to org.slf4j.Logger */

samples/simm-valuation-demo/src/main/kotlin/net/corda/vega/flows/SimmFlow.kt
```diff
@@ -310,7 +310,7 @@ object SimmFlow {
             val seller = TwoPartyDealFlow.Instigator(
                     replyToSession,
                     TwoPartyDealFlow.AutoOffer(offer.notary, offer.dealBeingOffered))
-            logger.info("Starting two party deal initiator with: ${replyToSession.counterparty.name}")
+            logger.info("Starting two party deal context with: ${replyToSession.counterparty.name}")
```
szymonsztuka: I think I would leave `initiator`.

core/src/main/kotlin/net/corda/core/context/InvocationContext.kt
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Models an initiator in Corda, can be a user, a service, etc.
+ */
+@CordaSerializable
+data class Actor(val id: Id, val serviceId: AuthServiceId, val owningLegalIdentity: CordaX500Name) {
+
+    companion object {
+        fun service(serviceClassName: String, owningLegalIdentity: CordaX500Name): Actor = Actor(Id(serviceClassName), AuthServiceId("SERVICE"), owningLegalIdentity)
```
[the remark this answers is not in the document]
sollecitom: Technically it does create a service actor. Why would a constructor be preferable to a factory method?

core/src/main/kotlin/net/corda/core/utilities/UuidGenerator.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.core.utilities
+
+import java.util.*
+
+class UuidGenerator {
+
```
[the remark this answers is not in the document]
sollecitom: Yes, it is indeed. I'm thinking about switching it to time based before it even gets merged.

samples/simm-valuation-demo/src/main/kotlin/net/corda/vega/flows/SimmFlow.kt
sollecitom: My bad, yeah, I'll revert it.

node/src/main/kotlin/net/corda/node/services/logging/ContextualLogger.kt
sollecitom: Not done with it, tomorrow I'll use the thread local :) I'll make sure the comment won't be there haha

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
sollecitom: I'd rather keep this function pure and handle the problem directly. About `readLong()`, that returns a `long` primitive, so it would throw `NullPointerException` and there is no boxed equivalent, but id and timestamp are set together, so if the id is there there's no chance the timestamp will be missing.

node/src/main/kotlin/net/corda/node/services/statemachine/SessionMessage.kt
```diff
@@ -11,7 +12,9 @@ import net.corda.core.utilities.UntrustworthyData
 import java.io.IOException
 
 @CordaSerializable
-interface SessionMessage
+interface SessionMessage {
+    val context: InvocationContext
```
exFalso: Why is this on every sessionmessage? Shouldn't we only propagate external IDs on the initiating message?

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -25,13 +29,53 @@ import java.io.InputStream
...
+            }
+            is FlowInitiator.Scheduled -> {
+                actor = Actor(Actor.Id("SCHEDULED"), AuthServiceId("SCHEDULED"), unknownName)
+                origin = Origin.Scheduled(initiator.scheduledState)
+            }
+        }
+        return InvocationContext(actor, origin)
+    }
+
+    fun copy(id: StateMachineRunId = this.id,
```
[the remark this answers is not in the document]
sollecitom: To avoid breaking our public API. Unfortunately adding a field to an interface is not a breaking change (provided we require them not to implement it), but adding it to a constructor is.

node/src/main/kotlin/net/corda/node/services/statemachine/SessionMessage.kt
sollecitom: As discussed, I'll change the implementation to send a new single-use invocation context along with a session message, making sure to log the "parent" context and the "child" context together on the sender's side.

node-api/src/main/kotlin/net/corda/nodeapi/RPCApi.kt
```diff
@@ -101,28 +104,32 @@ object RPCApi {
          */
         data class RpcRequest(
                 val clientAddress: SimpleString,
-                val id: RpcRequestId,
                 val methodName: String,
-                val serialisedArguments: ByteArray
+                val serialisedArguments: ByteArray,
+                val trace: Trace,
+                val externalTrace: Trace? = null,
```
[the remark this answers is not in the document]
sollecitom: This actually breaks a few tests and would force me to pass `null` explicitly. No harm in leaving it optional I guess.

.ci/api-current.txt
```diff
@@ -53,6 +53,144 @@ public interface net.corda.core.concurrent.CordaFuture extends java.util.concurr
...
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Actor copy(net.corda.core.context.Actor$Id, net.corda.core.context.AuthServiceId, net.corda.core.identity.CordaX500Name)
+  public boolean equals(Object)
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Actor$Id getId()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.identity.CordaX500Name getOwningLegalIdentity()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.AuthServiceId getServiceId()
+  public int hashCode()
+  public String toString()
+  public static final net.corda.core.context.Actor$Companion Companion
+##
+public static final class net.corda.core.context.Actor$Companion extends java.lang.Object
```
exFalso: This should be @JvmStatic instead

.ci/api-current.txt
```diff
@@ -53,6 +53,144 @@ public interface net.corda.core.concurrent.CordaFuture extends java.util.concurr
...
+  public boolean equals(Object)
+  @org.jetbrains.annotations.Nullable public final net.corda.core.context.Actor getActor()
+  @org.jetbrains.annotations.Nullable public final net.corda.core.context.Trace getExternalTrace()
+  @org.jetbrains.annotations.Nullable public final net.corda.core.context.Actor getImpersonatedActor()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Origin getOrigin()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Trace getTrace()
+  public int hashCode()
+  @org.jetbrains.annotations.NotNull public final java.security.Principal principal()
+  public String toString()
+  public static final net.corda.core.context.InvocationContext$Companion Companion
```
exFalso: @JvmStatic

.ci/api-current.txt
```diff
@@ -53,6 +53,144 @@ public interface net.corda.core.concurrent.CordaFuture extends java.util.concurr
...
+  public <init>(net.corda.core.context.Trace$InvocationId, net.corda.core.context.Trace$SessionId)
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Trace$InvocationId component1()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Trace$SessionId component2()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Trace copy(net.corda.core.context.Trace$InvocationId, net.corda.core.context.Trace$SessionId)
+  public boolean equals(Object)
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Trace$InvocationId getInvocationId()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.context.Trace$SessionId getSessionId()
+  public int hashCode()
+  public String toString()
+  public static final net.corda.core.context.Trace$Companion Companion
```
exFalso: @JvmStatic, also for the rest of the companions

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/CordaRPCClientTest.kt
```diff
@@ -131,31 +132,45 @@ class CordaRPCClientTest : NodeBasedTest(listOf("net.corda.finance.contracts", C
...
+        node.services.startFlow(CashIssueFlow(2000.DOLLARS, OpaqueBytes.of(0), nodeIdentity), InvocationContext.shell()).flatMap { it.resultFuture }.getOrThrow()
+        proxy.startFlow(::CashIssueFlow, 123.DOLLARS, OpaqueBytes.of(0), nodeIdentity).returnValue.getOrThrow()
+        proxy.startFlowDynamic(CashIssueFlow::class.java, 1000.DOLLARS, OpaqueBytes.of(0), nodeIdentity).returnValue.getOrThrow()
+
+        updates.expectEvents(isStrict = false) {
+            sequence(
+                    expect { update: StateMachineUpdate.Added ->
+                        checkShellNotification(update.stateMachineInfo)
+                    },
+                    expect { update: StateMachineUpdate.Added ->
```
exFalso: I think there's a `repeat` or `replicate` function or something for replicated events, but it's not that important

core/src/main/kotlin/net/corda/core/flows/FlowInitiator.kt
```diff
@@ -10,29 +10,35 @@ import java.security.Principal
  * communication started by peer node [FlowInitiator.Peer], scheduled flows [FlowInitiator.Scheduled]
  * or via the Corda Shell [FlowInitiator.Shell].
  */
+@Deprecated("Do not use these types. Future releases might remove them.")
 @CordaSerializable
 sealed class FlowInitiator : Principal {
     /** Started using [net.corda.core.messaging.CordaRPCOps.startFlowDynamic]. */
+    @Deprecated("Do not use these type. Future releases might remove it.")
```
exFalso: typo `these` -> `this`

finance/src/main/kotlin/net/corda/finance/flows/CashConfigDataFlow.kt
```diff
@@ -33,7 +33,8 @@ class ConfigHolder(services: AppServiceHub) : SingletonSerializeAsToken() {
         // Warning!! You are about to see a major hack!
         val baseDirectory = services.declaredField<Any>("serviceHub").value
```
exFalso: what on earth is this code

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -345,8 +346,9 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
         private fun <T> startFlowChecked(flow: FlowLogic<T>): FlowStateMachine<T> {
             val logicType = flow.javaClass
             require(logicType.isAnnotationPresent(StartableByService::class.java)) { "${logicType.name} was not designed for starting by a CordaService" }
-            val currentUser = FlowInitiator.Service(serviceInstance.javaClass.name)
-            return flowStarter.startFlow(flow, currentUser).getOrThrow()
+            // TODO check service permissions
+            val context = InvocationContext.service(serviceInstance.javaClass.name, myInfo.legalIdentities[0].name)
```
exFalso: Perhaps add TODO about identity selection

node/src/main/kotlin/net/corda/node/services/statemachine/StateMachineManagerImpl.kt
```diff
@@ -82,7 +86,7 @@ class StateMachineManagerImpl(
         val fibersWaitingForLedgerCommit = HashMultimap.create<SecureHash, FlowStateMachineImpl<*>>()!!
 
         fun notifyChangeObservers(change: StateMachineManager.Change) {
-            changesPublisher.bufferUntilDatabaseCommit().onNext(change)
+             changesPublisher.bufferUntilDatabaseCommit().onNext(change)
```
exFalso: whitespace 🚓

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/CordaRPCClientTest.kt
sollecitom: Uhm, I'll actually add more checks that are specific to each message.

finance/src/main/kotlin/net/corda/finance/flows/CashConfigDataFlow.kt
sollecitom: This mitigates the current hacky workaround by allowing it to work against mocks. I believe somebody else is working on a proper fix but it was breaking some tests.

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/CordaRPCClientTest.kt
exFalso: Good call

PR #2018, opened 2017-11-09
CORDA-768: Raft Notary: bootstrap command line flag

thschroeter:
When the bootstrap-raft-cluster flag is set, the node forms a single node cluster (ignoring configured peer addresses), acting as a seed for other nodes to join the cluster.

### DIFF
--- a/docs/source/running-a-notary.rst
+++ b/docs/source/running-a-notary.rst
@@ -28,3 +28,6 @@ For ``ValidatingNotaryService``, it is:
 
 Setting up a Raft notary is currently slightly more involved and is not recommended for prototyping purposes. There is
 work in progress to simplify it. To see it in action, however, you can try out the :ref:`notary-demo`.
+
+Use the `--bootstrap-raft-cluster` command line argument when starting the first node of a notary cluster for the first
+time. When the flag is set, the node will act as a seed for the cluster that other members can join.
--- a/node/src/main/kotlin/net/corda/node/ArgsParser.kt
+++ b/node/src/main/kotlin/net/corda/node/ArgsParser.kt
@@ -36,6 +36,7 @@ class ArgsParser {
     private val isVersionArg = optionParser.accepts("version", "Print the version and exit")
     private val justGenerateNodeInfoArg = optionParser.accepts("just-generate-node-info",
             "Perform the node start-up task necessary to generate its nodeInfo, save it to disk, then quit")
+    private val bootstrapRaftClusterArg = optionParser.accepts("bootstrap-raft-cluster", "Bootstraps Raft cluster. The node forms a single node cluster (ignoring otherwise configured peer addresses), acting as a seed for other nodes to join the cluster.")
     private val helpArg = optionParser.accepts("help").forHelp()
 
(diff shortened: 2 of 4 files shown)


thschroeter:
@rick-r3 please take a look when you have time, thanks.

node/src/main/kotlin/net/corda/node/ArgsParser.kt
```diff
@@ -52,8 +53,9 @@ class ArgsParser {
         val noLocalShell = optionSet.has(noLocalShellArg)
         val sshdServer = optionSet.has(sshdServerArg)
         val justGenerateNodeInfo = optionSet.has(justGenerateNodeInfoArg)
+        val bootstrapRaftCluster = optionSet.has(bootstrapRaftClusterArg)
```
rick-r3: We should perhaps consider having the notary be distinct from regular nodes, since this option is specific to the raft notary.  How does a node decide it wants to be a notary currently?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -62,7 +63,16 @@ open class NodeStartup(val args: Array<String>) {
 
         drawBanner(versionInfo)
         Node.printBasicNodeInfo(LOGS_CAN_BE_FOUND_IN_STRING, System.getProperty("log-path"))
-        val conf = loadConfigFile(cmdlineOptions)
+        val conf0 = loadConfigFile(cmdlineOptions)
+
+        val conf = if (cmdlineOptions.bootstrapRaftCluster && conf0 is NodeConfigurationImpl) {
```
rick-r3: Should we raise an error if it's not `NodeConfigurationImpl` but they requested bootstrap ?

thschroeter: Yes, that's a good idea.

docs/source/running-a-notary.rst
```diff
@@ -28,3 +28,6 @@ For ``ValidatingNotaryService``, it is:
 
 Setting up a Raft notary is currently slightly more involved and is not recommended for prototyping purposes. There is
 work in progress to simplify it. To see it in action, however, you can try out the :ref:`notary-demo`.
+
+Use the `--bootstrap-raft-cluster` command line argument when starting the first node of a notary cluster. When the flag
```
rick-r3: For extra clarity, should this be "starting the first node for the first time" ?  We only need it for the first run, yes?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
rick-r3: I think this would be clearer (excuse the lack of indents!):

`if(cmdlineOptions.bootstrapRaftCluster) {`
`   if(conf0 is NodeConfigurationImpl) {`
`   } else { `
`     // ERROR`
`  }`
`} else {`
`}`

docs/source/running-a-notary.rst
thschroeter: Yes, that's right.

PR #2022, opened 2017-11-09
AttachmentCriteriaQuery class and infrastructure

m4ksio:
for attachment metadata query.
Decision was made to add this changes to OS version - this is https://github.com/corda/enterprise/pull/55 ported for OS.
@mikehearn please review api-current.txt changes


mikehearn:
This was already reviewed in the Enterprise repo and then moved here, so I only checked the API diff. Approving. As far as I can tell the changes would be ABI compatible. We need to push more of the checking into the tool though.

PR #2023, opened 2017-11-09
CORDA-311 - SSH server instead of local shell

m4ksio:
Fully permissioned and trackable shell


node/src/integration-test/kotlin/net/corda/node/SSHServerTest.kt
```diff
@@ -0,0 +1,166 @@
...
+import java.net.ConnectException
+import kotlin.test.assertTrue
+import kotlin.test.fail
+import org.assertj.core.api.Assertions.assertThat
+import java.util.regex.Pattern
+
+class SSHServerTest {
+
+    @Test()
+    fun `ssh server does not start be default`() {
```
sollecitom: On port 2222 :) not sure how easy is to test for absence on initialisation.

node/src/integration-test/kotlin/net/corda/node/SSHServerTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+
+    @Test
+    fun `ssh server verify credentials`() {
+        val user = User("u", "p", setOf())
+        // The driver will automatically pick up the annotated flows below
+        driver {
+            val node = startNode(providedName = ALICE.name, rpcUsers = listOf(user),
+                    customOverrides = mapOf("sshd" to mapOf("port" to 2222)))
```
sollecitom: It'd maybe be good to have a dedicated configuration for SSH? Is there anything like that for web?

node/src/integration-test/kotlin/net/corda/node/SSHServerTest.kt
```diff
@@ -0,0 +1,166 @@
...
+            node.getOrThrow()
+
+            val session = JSch().getSession("u", "localhost", 2222)
```
sollecitom: Are these sessions closeable? We could fix this with `.use()`.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -213,6 +217,9 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
             FlowLogicRefFactoryImpl.classloader = cordappLoader.appClassLoader
 
             runOnStop += network::stop
+
+            startShell(rpcOps)
```
sollecitom: Do we check the enabling property later on? Perhaps doing here instead would be clearer.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -289,11 +279,6 @@ open class NodeStartup(val args: Array<String>) {
                     "Computers are useless. They can only\ngive you answers.  -- Picasso"
             )
 
-            // TODO: Delete this after CordaCon.
```
sollecitom: Good riddance :)

sollecitom: Is the new Cordacon registration open? In case let's update the date and the link instead.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -105,7 +106,9 @@ data class NodeConfigurationImpl(
         override val detectPublicIp: Boolean = true,
         override val activeMQServer: ActiveMqServerConfiguration,
         // TODO See TODO above. Rename this to nodeInfoPollingFrequency and make it of type Duration
-        override val additionalNodeInfoPollingFrequencyMsec: Long = 5.seconds.toMillis()
+        override val additionalNodeInfoPollingFrequencyMsec: Long = 5.seconds.toMillis(),
+        override val sshd: SSHDConfiguration? = null
```
sollecitom: Perhaps default to a `SSHDConfiguration.DISABLED`?

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -76,71 +66,47 @@ import kotlin.concurrent.thread
...
-            // TODO: Finish and enable SSH access.
-            // This means bringing the CRaSH SSH plugin into the Corda tree and applying Marek's patches
-            // found in https://github.com/marekdapps/crash/commit/8a37ce1c7ef4d32ca18f6396a1a9d9841f7ff643
-            // to that local copy, as CRaSH is no longer well maintained by the upstream and the SSH plugin
-            // that it comes with is based on a very old version of Apache SSHD which can't handle connections
-            // from newer SSH clients. It also means hooking things up to the authentication system.
-            Node.printBasicNodeInfo("SSH server access is not fully implemented, sorry.")
-            runSSH = false
-        }
+            val sshKeysDir = dir / "sshkey"
```
sollecitom: Can this come from the sshd configuration, perhaps with the default there?

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -76,71 +66,47 @@ import kotlin.concurrent.thread
...
+            sshKeysDir.toFile().mkdirs()
 
-        if (runSSH) {
             // Enable SSH access. Note: these have to be strings, even though raw object assignments also work.
-            config["crash.ssh.keypath"] = (dir / "sshkey").toString()
+            config["crash.ssh.keypath"] = (sshKeysDir / "hostkey.pem").toString()
```
sollecitom: Same as above, perhaps from the configuration?

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -76,71 +66,47 @@ import kotlin.concurrent.thread
...
             config["crash.ssh.keygen"] = "true"
-            // config["crash.ssh.port"] = node.configuration.sshdAddress.port.toString()
-            config["crash.auth"] = "simple"
-            config["crash.auth.simple.username"] = "admin"
-            config["crash.auth.simple.password"] = "admin"
+            config["crash.ssh.port"] = configuration.sshd?.port.toString()
+            config["crash.auth"] = "corda"
```
sollecitom: Same as above?

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -168,27 +134,25 @@ object InteractiveShell {
                     // Don't use the Java language plugin (we may not have tools.jar available at runtime), this
                     // will cause any commands using JIT Java compilation to be suppressed. In CRaSH upstream that
                     // is only the 'jmx' command.
-                    return super.getPlugins().filterNot { it is JavaLanguage }
+                    return super.getPlugins().filterNot { it is JavaLanguage } + CordaAuthenticationPlugin(rpcOps, userService)
```
sollecitom: No way to add this only in case sshd is enabled?

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -217,41 +181,47 @@ object InteractiveShell {
...
+                    throw e.rootCause
+                } catch (e: InvocationTargetException) {
+                    throw e.rootCause
+                }
             }
         } catch (e: NoApplicableConstructor) {
             output.println("No matching constructor found:", Color.red)
             e.errors.forEach { output.println("- $it", Color.red) }
+        } catch (e:PermissionException) {
+            output.println(e.message ?: "Access denied")
```
sollecitom: Colour red?

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -278,6 +287,10 @@ class MockNetwork(defaultParameters: MockNetworkParameters = MockNetworkParamete
             return E2ETestKeyManagementService(identityService, keyPairs)
         }
 
+        override fun startShell(rpcOps: CordaRPCOps) {
```
sollecitom: This might not be needed if we move the if.

node/src/integration-test/kotlin/net/corda/node/SSHServerTest.kt
m4ksio: Not sure either. But if contracts with other tests, so I think it is useful somehow.

node/src/integration-test/kotlin/net/corda/node/SSHServerTest.kt
m4ksio: You mean configuration object? We have one, but here (in overrides) don't support custom ones

node/src/integration-test/kotlin/net/corda/node/SSHServerTest.kt
m4ksio: They can be "disconnected()" but they don't use Closable interface...

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
m4ksio: I guess not. We can add this if/when we have new one..

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
m4ksio: The idea is, that this is abstract class, so implementations can choose what and how to enable shell

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
m4ksio: But then DISABLED is an object, and this empty object has to be represented/seralizable in config file. Null / absence of value in config file seems easier and I even clearer IMO

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
m4ksio: This is a bigger problem of "what shall be the hostkey", so I would leave it this for now

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
m4ksio: Same response

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
m4ksio: Port comes from config, "crash.auth" options is used internally to choose CordaAuthenticator so it should be constant

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
m4ksio: As above, I think it's not up to abstract node to decide

sollecitom:
Not sure why it doesn't show me the approve button. Anyway, approved, you can dismiss the review and get it merged.

mikehearn:
The local shell shouldn't have been removed! That's used in DemoBench and in various other demo-related places, it's also needed/very useful for local testing purposes. I don't recall removal of that feature being a part of the request?

docs/source/shell.rst
```diff
@@ -18,11 +18,44 @@ Some of its features include:
...
+
+Connecting
+----------
+
+Linux and MacOS computers usually come with SSH client preinstalled. On Windows it usually require extra download.
+Usual connection syntax is ``ssh user@host -p 2222`` - where ``user`` is a RPC username, and ``-p`` specifies a port parameters -
+it's the same as setup in ``node.conf`` file. ``host`` should point to a node hostname, usually ``localhost`` if connecting and
+running node on the same computer. Password will be asked after estabilishing connection.
+
+:note: While developing, checking multiple samples or simply restarting a node frequently host key may be regenerated. SSH usually
```
mikehearn: It'd be nice if we could generate the SSH host key as a constant when run in --dev-mode

node/src/main/kotlin/net/corda/node/shell/RPCOpsWithContext.kt
```diff
@@ -0,0 +1,201 @@
...
+import net.corda.node.services.messaging.RpcContext
+import net.corda.nodeapi.User
+import rx.Observable
+import java.io.InputStream
+import java.security.PublicKey
+import java.time.Instant
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Future
+
+class RPCOpsWithContext(val cordaRPCOps: CordaRPCOps, val user:User) : CordaRPCOps {
```
mikehearn: Do we really need this? If so could the javadoc explain why? I'd have thought 1 ssh connection would = 1 thread, and so the RPC user TLS could be set up once at connect time. This seems like unfortunate duplication: we'll have to remember to add it here every time we add a new RPC.

sollecitom:
We were told to remove unauthorised access to the shell.
For local test purposes it's like 1 SSH login away.

node/src/main/kotlin/net/corda/node/utilities/ANSIProgressRenderer.kt
```diff
@@ -1,197 +0,0 @@
-package net.corda.node.utilities
```
mikehearn: Why is this removed? What's that got to do with ssh access?

mikehearn:
Local shell access isn't unauthorised. It has implicit administrator access. If you have access to the local shell you have access to the node anyway, implicitly (modulo convoluted setups involving the screen utility).

m4ksio:
Introduction of permissioning in the shell has a consequence of restricting access without authentication. Also, the decision to channel all shell operations via RPC had some minor changes in functionality itself. That was the common consensus we had in a team. Otherwise - we had to either provide a local login-in / keep userless/uberuser.
However, shell still can be accessed by simply SSHing into the node.

mikehearn:
Anyone who can access the local tty via the OS provided sshd has sufficient kernel rights to debug the host VM (which is running as the same UNIX user) and do whatever they like with it. Restricting OS-level access to a node isn't a part of this task, although this task is a part of it, and it's a lot more involved than just disabling the local shell.

Please put the local shell back. It makes no security difference. For one, this breaks the DemoBench workflow, which is intended to support interactive interaction via the shell.

docs/source/shell.rst
m4ksio: Thats good idea!

node/src/main/kotlin/net/corda/node/shell/RPCOpsWithContext.kt
m4ksio: Unfortunately, yes. Apache SSHD does not create thread per connection but rather uses dynamic pool, then CRaSH uses another thread pool to execute commands, and they both use different sessions objects etc. 
I think there is some work to introduce proper "user context" propagation being done, so once we move away from LocalThread to store user info, we won't need this proxy.

node/src/main/kotlin/net/corda/node/utilities/ANSIProgressRenderer.kt
m4ksio: CRaSH exposes only it's own PrintWriter to commands (not even OutputStream), so with SSH only there was no use for this renderer at all.

m4ksio:
Just a general comment - we didn't try to harden security in this task, rather bring some coherence and permissions usage inside the node - so we don't have flows started by almighty entity. 
Maybe the local shell shall be enabled in --dev-mode only?

mikehearn:
Yes, we could do that. Alternatively, maybe we should impose a new rule - all nodes require at least one administrator user? Then the local shell runs as that? Do we want a node that can't be admin'd at all except by poking at files and JVMs using dev tools - probably not. I guess there must always be at least one admin.

Note that the local shell disables itself if there's no tty anyway.

node/src/main/kotlin/net/corda/node/shell/RPCOpsWithContext.kt
mikehearn: OK, makes sense. Can you stick that explanation in a Javadoc for this class, so we can see why?

Also given the boilerplate nature of the class I wonder if we can use a java.lang.reflect.Proxy to do it instead of implementing the whole thing manually.

node/src/main/kotlin/net/corda/node/utilities/ANSIProgressRenderer.kt
mikehearn: Hm, but CRaSH does have commands that use smart ANSI rendering, like for drawing tables and doing the command line editing. So there must be a way to emit ANSI escapes.

m4ksio: ANSIProgressRenderer was writing directly to stdout so there was no quick way to adapt it for real CRaSH provided input/output. However since we are brining local shell back I will restore this.

m4ksio:
The one-admin rule is a brilliant idea. We just need to define what an admin means, as now we merely have bag of permissions.

mikehearn:
Indeed, the definition of admin is quite subtle. For now let's use a simple definition: a node admin (vs an operating system admin) is anyone who has the set of all possible permissions.

PR #2030, opened 2017-11-10
Vault query - don't return repeated column which groups multiple aggregation functions.

szymonsztuka:
When a vault query has multiple aggregate function (e.g sum and avg), the column used to group by was repeated accordingly to number of aggregate functions.

Example aggregate sum and max of `pennies` group by `currency`:

    val sumCriteria = VaultCustomQueryCriteria(builder { CashSchemaV1.PersistentCashState::pennies.sum(groupByColumns = listOf(CashSchemaV1.PersistentCashState::currency)) })
    val maxCriteria = VaultCustomQueryCriteria(builder { CashSchemaV1.PersistentCashState::pennies.max(groupByColumns = listOf(CashSchemaV1.PersistentCashState::currency)) })
    vaultService.queryBy<FungibleAsset<*>>(sumCriteria.and(maxCriteria))

Before the change the query would return the following columns:
`sum, currency, max, currency`,
after the change (the column with currency is not duplicated):
`sum, max, currency`.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
+++ b/node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
@@ -196,6 +196,7 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
                             AggregateFunctionType.MAX -> criteriaBuilder.max(column)
                             AggregateFunctionType.MIN -> criteriaBuilder.min(column)
                         }
+                //TODO investigate possibility to avoid producing redundant joins in SQL for multiple aggregate functions against the same table
                 aggregateExpressions.add(aggregateExpression)
                 // optionally order by this aggregate function
                 expression.orderBy?.let {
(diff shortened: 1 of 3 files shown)


josecoll:
Great! This solves the general query failure observed.

It removes the redundant selects for currency_code (and thus returns an optimal result set now)

The generated query could still be improved by removing the redundant joins to the same table:

```
    select
        sum(cashschema1_.pennies) as col_0_0_,
        max(cashschema2_.pennies) as col_1_0_,
        min(cashschema3_.pennies) as col_2_0_,
        avg(cashschema4_.pennies) as col_3_0_,
        cashschema4_.ccy_code as col_4_0_ 
    from
        vault_states vaultschem0_ cross 
    join
        contract_cash_states cashschema1_ cross 
    join
        contract_cash_states cashschema2_ cross 
    join
        contract_cash_states cashschema3_ cross 
    join
        contract_cash_states cashschema4_ 
    where
        vaultschem0_.output_index=cashschema1_.output_index 
        and vaultschem0_.transaction_id=cashschema1_.transaction_id 
        and vaultschem0_.output_index=cashschema2_.output_index 
        and vaultschem0_.transaction_id=cashschema2_.transaction_id 
        and vaultschem0_.output_index=cashschema3_.output_index 
        and vaultschem0_.transaction_id=cashschema3_.transaction_id 
        and vaultschem0_.output_index=cashschema4_.output_index 
        and vaultschem0_.transaction_id=cashschema4_.transaction_id 
        and 1=1 
        and vaultschem0_.state_status=? 
        and (
            vaultschem0_.contract_state_class_name in (
                ?
            )
        ) 
    group by
        cashschema4_.ccy_code fetch first ? rows only
```

josecoll:
Perhaps we should add a TODO in the relevant source file where this join inefficiency needs addressing.

szymonsztuka:
Yes, I'll add TODO a this may be nontrivial change.

PR #2031, opened 2017-11-10
Refactor RPC permissions to use off-the-shelf solution [ENT-1068]

IgorNitto:
A refactoring of our RPC authentication and authorization logic based for integrating the Apache Shiro framework (see tickets [ENT-1068], [ENT-1000]).

This change include:
 - Reuse the Shiro Permission interface and related classes to model our RPC permission scheme and syntax. 
- Replacing the RPCUserService interface and implementation with a run-time configurable Authenticator/Authorizer object taken from Shiro. This is where external security stores (DB, LDAP...) can be plugged in (see RPCRealmFactory)
- Allow pointing Shiro framework to an external config file (more on configuration in general comments)
 
The following is an example configuration for linking to an external H2 database storing credentials:
[shiro.ini.txt](https://github.com/corda/corda/files/1474563/shiro.ini.txt). A sample could be built that automatically set up a credentials database and runs a node against it, this is not implemented here.  

Note: there are a lots of tiny changes to many unit tests. This is unfortunately a consequence of a dependency of many tests on the Permissions class, which has been replaced.


client/rpc/src/test/kotlin/net/corda/client/rpc/RPCPermissionsTests.kt
```diff
@@ -32,7 +38,13 @@ class RPCPermissionsTests : AbstractRPCTest() {
 
     class TestOpsImpl : TestOps {
         override val protocolVersion = 1
-        override fun validatePermission(str: String) { rpcContext().requirePermission(str) }
+        override fun validatePermission(str: String) {
+            val permission = RPCPermission(str)
+            val userPermissions = rpcContext().currentUser.permissions.map {RPCPermission(it)}
+            if (!userPermissions.any {it.implies(permission)}) {
```
m4ksio: Can we use .none() instead of negated any() ?

core/src/smoke-test/kotlin/net/corda/core/cordapp/CordappSmokeTest.kt
```diff
@@ -20,6 +20,7 @@ import java.nio.file.Paths
 import java.util.concurrent.atomic.AtomicInteger
 import kotlin.streams.toList
 
+
```
m4ksio: Unnecessary newline

node/src/main/kotlin/net/corda/node/internal/RpcAuthorisationProxy.kt
```diff
@@ -1,158 +1,221 @@
...
-
-    // TODO change to KFunction reference after Kotlin fixes https://youtrack.jetbrains.com/issue/KT-12140
-    private inline fun <RESULT> guard(methodName: String, action: () -> RESULT) = guard(methodName, emptyList(), action)
-
-    // TODO change to KFunction reference after Kotlin fixes https://youtrack.jetbrains.com/issue/KT-12140
-    private inline fun <RESULT> guard(methodName: String, args: List<Any?>, action: () -> RESULT): RESULT {
-
-        context.invoke().requireEitherPermission(permissionsAllowing.invoke(methodName, args))
-        return action.invoke()
+import net.corda.node.services.security.RPCPermission
```
m4ksio: Do we need this class? Just a speculation, but shouldn't we simply put check in real implementation, possibly with anotations? Or this shall wait for Michele work with propagating context?

client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
```diff
@@ -52,17 +51,17 @@ class NodeMonitorModelTest {
...
-                    startFlow<CashPaymentFlow>(),
-                    startFlow<CashExitFlow>(),
-                    invokeRpc(CordaRPCOps::notaryIdentities),
-                    invokeRpc("vaultTrackBy"),
-                    invokeRpc("vaultQueryBy"),
-                    invokeRpc(CordaRPCOps::internalVerifiedTransactionsFeed),
-                    invokeRpc(CordaRPCOps::stateMachineRecordedTransactionMappingFeed),
-                    invokeRpc(CordaRPCOps::stateMachinesFeed),
-                    invokeRpc(CordaRPCOps::networkMapFeed))
+            val cashUser = User("user1", "test", permissions = setOfPermissionStrings(
```
m4ksio: Why are we still using string permissions rather then permissions classes you prepared?

IgorNitto: In this instance, the reason is that I was not keen on modifying the definition of User for various reasons. This would impact lots of code which currently rely on it and would require to have a dependency from node-api to node.  I preferred to keep User as  a "lower level" data class which merely encode a Config entry. It might actually become an unnecessary class once the authentication/authorization is done behind the scene by Shiro (or whatever).

client/rpc/src/test/kotlin/net/corda/client/rpc/RPCPermissionsTests.kt
IgorNitto: Yes, good point.

client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
m4ksio: I just think, that now we had to change all tests, but if we move to "proper" permissions objects we would have to change again.. but I guess it's fine for now though

josecoll:
General comments.

I think we are jumping the gun a little with this work.

1) The intent was to **evaluate** an off-the-shelf entitlements solution (Apache Shiro is just one of three I proposed) to determine whether it is a good solution to provide abstraction (and generalised implementation of boilerplate code), pluggability (to different underlying sources such as LDAP, JDBC, AD), and configurability (eg. ease setup/removal of users, permissions)

2) Although this PR touches only on the permissioning side of things, my immediate feedback is

- represents a large amount of boilerplate coding (which I was hoping we could avoid).
- does not demonstrate externalized configuration of any kind.
- duplicates a lot of the code that @sollecitom originally wrote (twice already)

I was hoping for something as simple as:
```
user = context.getPrincipal()
isUserInRole(user, role)
```
applied in isolated proxied locations (eg. RCPProxy, FlowLogicImpl)

node/src/main/kotlin/net/corda/node/internal/RpcAuthorisationProxy.kt
IgorNitto: Indeed annotations can be used to define permission checks in a nice and compact way and security frameworks such as Shiro enable such usage.
However in our case there is the complication that a permission might depend on the value of an argument (such as in the case of flow invocation) and I could not find a way to represent this with Shiro annotations

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -137,6 +138,9 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
      * or has loaded network map data from local database */
     val nodeReadyFuture: CordaFuture<Unit>
         get() = _nodeReadyFuture
+
+    protected abstract val rpcSecurityManager : SecurityManager
```
sollecitom: As discussed, this needs to be only inside the implementation of `RPCUserService` and nowhere else.

node/src/main/kotlin/net/corda/node/services/security/RPCPermission.kt
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Provide a representation of RPC permissions based on Apache Shiro permissions framework.
+ * A permission represents a set of actions: for example, the set of all RPC invocations, or the set
+ * of RPC invocations acting on a given class of Flows in input. A permission `implies` another one if
+ * its set of actions contains the set of actions in the other one. In Apache Shiro, permissions are
+ * represented by instances of the [Permission] interface which offers a single method: [implies], to
+ * test if the 'x implies y' binary predicate is satisfied.
+ */
+class RPCPermission : DomainPermission {
```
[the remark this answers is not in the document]
IgorNitto: Agreed. This class can actually be "demoted" to an implementation detail necessary to send our string representation in the world of Shiro.

IgorNitto:
This PR was indeed a PoC "previewing" the more serious work on CORDA-827. It should not be merged and so I am closing it now.

PR #2032, opened 2017-11-10
Cross provider Issuer Reference database storage

josecoll:
Enforce size checking of _Issuer Reference_ type (max 512), and ensure consistently mapped to an equivalently sized binary database type.


node/src/main/kotlin/net/corda/node/services/persistence/HibernateConfiguration.kt
```diff
@@ -133,4 +140,51 @@ class HibernateConfiguration(val schemaService: SchemaService, private val datab
...
+    }
+
+    private object OpaqueBytesDescriptor : AbstractTypeDescriptor<OpaqueBytes>(OpaqueBytes::class.java) {
+
+        override fun fromString(text: String): OpaqueBytes {
+            return OpaqueBytes.of(text.toByte())
+        }
+
+        override fun toString(value: OpaqueBytes): String {
+            return value.toString().trim('[',']')
```
rick-r3: I think it is worth adding `toHexString()` to `ByteSequence`, rather than trimming `toString()`

node/src/main/kotlin/net/corda/node/services/persistence/HibernateConfiguration.kt
```diff
@@ -133,4 +140,51 @@ class HibernateConfiguration(val schemaService: SchemaService, private val datab
...
+        }
+
+        override fun getName(): String {
+            return "opaquebytes"
+        }
+    }
+
+    private object OpaqueBytesDescriptor : AbstractTypeDescriptor<OpaqueBytes>(OpaqueBytes::class.java) {
+
+        override fun fromString(text: String): OpaqueBytes {
```
rick-r3: What does this `toByte()` do?  We have `String.parseAsHex()` extension function, which seems more explicit / readable, even if `toByte()` does the right thing.

finance/src/test/kotlin/net/corda/finance/schemas/SampleCommercialPaperSchemaV2.kt
```diff
@@ -30,8 +32,9 @@ object SampleCommercialPaperSchemaV2 : MappedSchema(schemaFamily = CommercialPap
             @Column(name = "face_value_issuer_key_hash", length = MAX_HASH_HEX_SIZE)
             var faceValueIssuerPartyHash: String,
 
-            @Column(name = "face_value_issuer_ref")
-            var faceValueIssuerRef: ByteArray,
+            @Column(name = "face_value_issuer_ref", columnDefinition = "varchar(16)")
+            @Type(type = "opaquebytes")
```
rick-r3: If we have to specify the `@Type` here, is it better to mention we are storing as hex ?  e.g. `bytes-as-hex` or something.

finance/src/test/kotlin/net/corda/finance/schemas/SampleCommercialPaperSchemaV2.kt
```diff
@@ -44,6 +47,6 @@ object SampleCommercialPaperSchemaV2 : MappedSchema(schemaFamily = CommercialPap
             @Transient
             val _issuerParty: AbstractParty,
             @Transient
-            val _issuerRef: ByteArray
+            val _issuerRef: OpaqueBytes
```
rick-r3: I think we should broaden this to be `ByteSequence` (the parent interface / sealed class).  It's not clear to me whether we want the immutable / copying semantics of `OpaqueBytes` or the non-copying / mutable semantics of other implementations.

node/src/main/kotlin/net/corda/node/services/persistence/HibernateConfiguration.kt
```diff
@@ -133,4 +140,51 @@ class HibernateConfiguration(val schemaService: SchemaService, private val datab
             }
         }
     }
+    // Type helper to enable storage and query of Corda OpaqueBytes type
+    private object OpaqueBytesType : AbstractSingleColumnStandardBasicType<OpaqueBytes>(VarcharTypeDescriptor.INSTANCE, OpaqueBytesDescriptor), DiscriminatorType<OpaqueBytes> {
```
rick-r3: Let's broaden to parent of `OpaqueBytes` (`ByteSequence`).  Also, I think we should involve "hex" in the naming here (and the name of the type in `getName()`.  `ByteSequenceAsHexType` and `hexbytes` (as the name)?

node/src/main/kotlin/net/corda/node/services/persistence/HibernateConfiguration.kt
```diff
@@ -133,4 +140,51 @@ class HibernateConfiguration(val schemaService: SchemaService, private val datab
...
+        override fun objectToSQLString(value: OpaqueBytes, dialect: Dialect?): String {
+            return value.toString().trim('[',']')
+        }
+
+        override fun getName(): String {
+            return "opaquebytes"
+        }
+    }
+
+    private object OpaqueBytesDescriptor : AbstractTypeDescriptor<OpaqueBytes>(OpaqueBytes::class.java) {
```
rick-r3: Likewise broaden to parent `ByteSequence` type.

rick-r3:
I've added some general code review comments, but as discussed, we should start defining max lengths for variable length fields, validating that in code and documenting that on the doc site, besides the inferred limit from database schema.

Seems like the reference but of `PartyAndReference` will be the first....

josecoll:
Captured Slack discussion on how to proceed with this PR:

colljos [9:24 AM] 
@mike discussed setting max lengths for variable length fields with @rick.parker and enforcing that in the code (and subsequently setting database attribute types to correct fixed lengths). This discussion came out of a review some Database work to standardise how we store/retrieve `IssuerReference` which is of  type `OpaqueBytes`. What would you specify as the max size of an `IssuerReference`? (I had used 16 in my PR but I couldn’t find any specification of what it should be)


mike [9:38 AM] 
hmmm - it wasn't meant to have a limit, that's why


[9:39] 
can we use a blob for it? or do some db engines, not work with lots of small blobs (perhaps a hybrid approach)


colljos [10:14 AM] 
surely an IssuerReference has a limit?


[10:14] 
(eg. akin to a serial number on a bank note)


[10:14] 
The issue was storing IssuerRef as a Blob is inefficient and queries perform poorly in Oracle.


mike [10:32 AM] 
the idea was that "IssueReference" could contain e.g. encrypted local state.


[10:32] 
like a cookie. it can either be an opaque id that's looked up in a local db, or it can just be the directly stored data itself encrypted under a key owned by the institution. note that encryption can increase the size of a piece of data.


[10:33] 
perhaps we can store the contents as a hash, with the actual contents held elsewhere. query = hash(input)


[10:33] 
are we disk space constrained? if we just make the limits quite large (like 512 bytes or 1kb) does that trash performance too badly?


colljos [10:39 AM] 
I guess we could …


colljos [10:40 AM] 
added and commented on this Plain Text snippet
      @Column(name = "issuer_key_hash", length = MAX_HASH_HEX_SIZE)
      var issuerPartyHash: String,
      @Column(name = "issuer_ref", columnDefinition = "varchar(16)")
      @Type(type = "bytes-as-hex")
      var issuerRef: ByteSequence
1 Comment Collapse
current proposal from PR review feedback



colljos [10:41 AM] 
added this Plain Text snippet: becomes
      @Column(name = "issuer_key_hash", length = MAX_HASH_HEX_SIZE)
      var issuerPartyHash: String,
      @Column(name = "issuer_ref_hash", length = MAX_HASH_HEX_SIZE)
      var issuerRefHash: String
Add Comment Click to expand inline 6 lines



mike [10:52 AM] 
one thing i went back and forth on in the early days was whether to automatically encrypt the issuerRef fields.

[10:52] 
e.g. the obvious thing for an app dev to do, is to put customer IDs into them

[10:52] 
but then if your customer IDs are incrementing integers you leak the size of your customer base and possibly its growth rate. partial visibility helps a lot here.

roger3cev [10:52 AM] 
Also has data protection implications, apparently

mike [10:52 AM] 
transparent encryption/decryption is nice but introduces a new key management problem (currently only encryption is via TLS).


[10:53] 
i ended up concluding that maybe SGX was sufficient, and anyway, we never implemented it.

roger3cev [10:53 AM] 
(Even if the data is encrypted) (edited)

mike [10:53 AM] 
yes i know stupid gdpr


roger3cev [10:53 AM] 
aye :confused:


mike [10:54 AM] 
it would be nice to think this through a bit with a few use cases, perhaps we can optimise it such that the issuerRef field could be *either* a small piece of data (unencrypted in the db) or a hash pointer to something else. so if we introduce transparent encrypt/decrypt, it's only visible on the wire, and in the db that field can just be used as a join key with other tables that may predate corda


[10:55] 
that was the intention really ... that this field can be used as a join key. so for that the db would need to see the contents unencrypted.


[10:55] 
hmmm.


[10:57] 
maybe we can punt it for now and impose a restriction on the size


[10:57] 
and then release that restriction later.


[10:57] 
OTOH we probably won't revisit it for years if we do that.


rick.parker [11:00 AM] 
I feel like if we said 512 bytes or 1KB, then that doesn’t sound *too* constrained does it?


mike [11:00 AM] 
that would probably be fine for all use cases i can imagine especially if that's 1kb *after* decryption


[11:00] 
and assuming a tight binary format is used


[11:00] 
however - do db's handle 1kb columns well?


[11:00] 
at what point to blobs start getting blobby?


rick.parker [11:02 AM] 
It varies.  Little exercise for @colljos to figure out what type we’d use for Oracle/AzureSQL/SQLServer/Postgres if we said that was the limit

rick-r3:
I thought the idea was to go with a small-ish (but bigger than 16 bytes) binary field.  e.g. VARBINARY(512) or something like that?

josecoll:
That was an alternative suggestion.

core/src/main/kotlin/net/corda/core/schemas/CommonSchema.kt
```diff
@@ -74,7 +77,8 @@ object CommonSchemaV1 : MappedSchema(schemaFamily = CommonSchema.javaClass, vers
             @Column(name = "issuer_name")
             var issuer: AbstractParty,
 
-            @Column(name = "issuer_reference")
-            var issuerRef: ByteArray
+            // TODO: store the raw issuer reference contents in a separate join table
```
rick-r3: I don't see the point of another table.  Might as well remove the TODO.

finance/src/main/kotlin/net/corda/finance/schemas/CashSchemaV1.kt
```diff
@@ -36,7 +39,8 @@ object CashSchemaV1 : MappedSchema(schemaFamily = CashSchema.javaClass, version
             @Column(name = "issuer_key_hash", length = MAX_HASH_HEX_SIZE)
             var issuerPartyHash: String,
 
-            @Column(name = "issuer_ref")
-            var issuerRef: ByteArray
+            // TODO: store the raw issuer reference contents in a separate join table
```
rick-r3: I don't see the point of another table. Might as well remove the TODO.

finance/src/main/kotlin/net/corda/finance/schemas/CommercialPaperSchemaV1.kt
```diff
@@ -48,7 +51,8 @@ object CommercialPaperSchemaV1 : MappedSchema(schemaFamily = CommercialPaperSche
             @Column(name = "face_value_issuer_key_hash", length = MAX_HASH_HEX_SIZE)
             var faceValueIssuerPartyHash: String,
 
-            @Column(name = "face_value_issuer_ref")
-            var faceValueIssuerRef: ByteArray
+            // TODO: store the raw issuer reference contents in a separate join table
```
rick-r3: I don't see the point of another table. Might as well remove the TODO.

finance/src/test/kotlin/net/corda/finance/schemas/SampleCashSchemaV1.kt
```diff
@@ -35,7 +38,8 @@ object SampleCashSchemaV1 : MappedSchema(schemaFamily = CashSchema.javaClass, ve
             @Column(name = "issuer_key_hash", length = MAX_HASH_HEX_SIZE)
             var issuerPartyHash: String,
 
-            @Column(name = "issuer_ref")
-            var issuerRef: ByteArray
+            // TODO: store the raw issuer reference contents in a separate join table
```
rick-r3: I don't see the point of another table. Might as well remove the TODO.

finance/src/test/kotlin/net/corda/finance/schemas/SampleCashSchemaV3.kt
```diff
@@ -37,7 +41,8 @@ object SampleCashSchemaV3 : MappedSchema(schemaFamily = CashSchema.javaClass, ve
             @Column(name = "issuer_name")
             var issuer: AbstractParty,
 
-            @Column(name = "issuer_ref")
-            var issuerRef: ByteArray
+            // TODO: store the raw issuer reference contents in a separate join table
+            @Column(name = "issuer_ref_hash", length = MAX_HASH_HEX_SIZE)
```
rick-r3: I don't see the point of another table. Might as well remove the TODO.

finance/src/test/kotlin/net/corda/finance/schemas/SampleCommercialPaperSchemaV1.kt
```diff
@@ -46,7 +49,8 @@ object SampleCommercialPaperSchemaV1 : MappedSchema(schemaFamily = CommercialPap
             @Column(name = "face_value_issuer_key_hash", length = MAX_HASH_HEX_SIZE)
             var faceValueIssuerPartyHash: String,
 
-            @Column(name = "face_value_issuer_ref")
-            var faceValueIssuerRef: ByteArray
+            // TODO: store the raw issuer reference contents in a separate join table
```
rick-r3: I don't see the point of another table. Might as well remove the TODO.

finance/src/test/kotlin/net/corda/finance/schemas/SampleCommercialPaperSchemaV2.kt
```diff
@@ -30,8 +32,9 @@ object SampleCommercialPaperSchemaV2 : MappedSchema(schemaFamily = CommercialPap
             @Column(name = "face_value_issuer_key_hash", length = MAX_HASH_HEX_SIZE)
             var faceValueIssuerPartyHash: String,
 
-            @Column(name = "face_value_issuer_ref")
-            var faceValueIssuerRef: ByteArray,
+            // TODO: store the raw issuer reference contents in a separate join table
```
rick-r3: I don't see the point of another table. Might as well remove the TODO.

node/src/main/kotlin/net/corda/node/services/vault/VaultSchema.kt
```diff
@@ -125,14 +129,15 @@ object VaultSchemaV1 : MappedSchema(schemaFamily = VaultSchema.javaClass, versio
             @Column(name = "issuer_name")
             var issuer: AbstractParty,
 
-            @Column(name = "issuer_reference")
-            var issuerRef: ByteArray
+            // TODO: store the raw issuer reference contents in a separate join table
```
rick-r3: I don't see the point of another table. Might as well remove the TODO.

node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
```diff
@@ -349,8 +350,8 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
 
         // issuer reference
         criteria.issuerRef?.let {
-            val issuerRefs = (criteria.issuerRef as List<OpaqueBytes>).map { it.bytes }
-            predicateSet.add(criteriaBuilder.and(vaultFungibleStates.get<ByteArray>("issuerRef").`in`(issuerRefs)))
+            val issuerRefs = (criteria.issuerRef as List<OpaqueBytes>).map { it.toHashedString() }
```
rick-r3: So we only convert the hash for `VaultSchemaV1.VaultFungibleStates` ?

node/src/main/kotlin/net/corda/node/services/vault/VaultSchema.kt
rick-r3: Not related to this PR, but it highlighted it.  How does `VaultSchemaV1. VaultFungibleStates` relate to `CommonSchemaV1. FungibleState`?  Should the former not extend the latter (and basically add no fields).

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -103,3 +104,10 @@ fun ByteArray.sha256(): SecureHash.SHA256 = SecureHash.sha256(this)
  * Compute the SHA-256 hash for the contents of the [OpaqueBytes].
  */
 fun OpaqueBytes.sha256(): SecureHash.SHA256 = SecureHash.sha256(this.bytes)
```
rick-r3: Can we make these (this and below) methods on `ByteSequence` rather than extension methods?

Here's my logic.  Written as is, this will copy the underlying ByteArray unnecessarily, then hash that copy.  There's no need to generate the new array.  Adding them directly to `ByteSequence` would allow us to fix this later, but with an extension method, we can't.

rick-r3:
OK, I get the idea with the hash, but it's only supported on the FungibleAsset query and none of the others, right?  So all the other schema changes are not magically handled?

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
josecoll: Removed.

PR #2034, opened 2017-11-10
Removal of transaction contract state as BLOB in VaultStates table.

josecoll:
Certain database providers have limitations on join table queries where LOB attributes are included in the select criteria. This PR removes the duplicated copy of the Transaction (Contract State) from the `VaultStates` table to prevent Vault Query failure.

Transaction contract state now resolved using `StateLoader` (from `DBTransactionStorage`).
An additional `loadStates(Set<StateRef>)` has also been added to the StateLoader interface.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/StartedNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/StartedNode.kt
@@ -37,4 +37,9 @@ class StateLoaderImpl(private val validatedTransactions: TransactionStorage) : S
         val stx = validatedTransactions.getTransaction(stateRef.txhash) ?: throw TransactionResolutionException(stateRef.txhash)
         return stx.resolveBaseTransaction(this).outputs[stateRef.index]
     }
+
+    @Throws(TransactionResolutionException::class)
+    override fun loadStates(stateRefs: Set<StateRef>): Set<TransactionState<*>> {
+        return (stateRefs.map { loadState(it) }).toSet()
+    }
 }
--- a/node/src/main/kotlin/net/corda/node/internal/StartedNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/StartedNode.kt
@@ -1,8 +1,6 @@
 package net.corda.node.internal
 
-import net.corda.core.contracts.StateRef
-import net.corda.core.contracts.TransactionResolutionException
-import net.corda.core.contracts.TransactionState
+import net.corda.core.contracts.*
 import net.corda.core.flows.FlowLogic
 import net.corda.core.messaging.CordaRPCOps
 import net.corda.core.node.NodeInfo
@@ -39,7 +37,8 @@ class StateLoaderImpl(private val validatedTransactions: TransactionStorage) : S
     }
 
     @Throws(TransactionResolutionException::class)
-    override fun loadStates(stateRefs: Set<StateRef>): Set<TransactionState<*>> {
-        return (stateRefs.map { loadState(it) }).toSet()
+    // TODO: future implementation to retrieve contract states from a Vault BLOB store
+    override fun loadStates(stateRefs: Set<StateRef>): Set<StateAndRef<ContractState>> {
+        return (stateRefs.map { StateAndRef(loadState(it), it) }).toSet()
     }
 }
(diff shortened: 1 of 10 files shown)


node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -473,7 +466,7 @@ class NodeVaultService(
                                     return@forEachIndexed
                                 val vaultState = result[0] as VaultSchemaV1.VaultStates
                                 val stateRef = StateRef(SecureHash.parse(vaultState.stateRef!!.txId!!), vaultState.stateRef!!.index!!)
-                                val state = vaultState.contractState.deserialize<TransactionState<T>>(context = STORAGE_CONTEXT)
+                                val state = stateLoader.loadState(stateRef)
```
rick-r3: Another place we should be using `loadStates` (plural) outside the loop.

node/src/main/kotlin/net/corda/node/services/vault/VaultSchema.kt
```diff
@@ -29,6 +29,8 @@ object VaultSchemaV1 : MappedSchema(schemaFamily = VaultSchema.javaClass, versio
     @Table(name = "vault_states",
             indexes = arrayOf(Index(name = "state_status_idx", columnList = "state_status")))
     class VaultStates(
+            /** NOTE: serialized transaction state (including contract state) is now resolved from transaction store **/
```
rick-r3: We should mention that we think this is temporary, until we have created a distinct table to hold serialized state data for our vault states.  We need to make that change before anyone deploys for real, otherwise the upgrade is going to get fun.

node/src/main/kotlin/net/corda/node/internal/StartedNode.kt
```diff
@@ -35,4 +35,9 @@ class StateLoaderImpl(private val validatedTransactions: TransactionStorage) : S
         val stx = validatedTransactions.getTransaction(stateRef.txhash) ?: throw TransactionResolutionException(stateRef.txhash)
         return stx.resolveBaseTransaction(this).outputs[stateRef.index]
     }
+
+    @Throws(TransactionResolutionException::class)
+    override fun loadStates(stateRefs: Set<StateRef>): Set<TransactionState<*>> {
+        return (stateRefs.map { loadState(it) }).toSet()
```
rick-r3: Add a TODO to make this do it in bulk, and to use a "per state" table, since the TX store will at some point be encrypted.

josecoll:
Added JIRA for follow-up activity to introduce a new Vault States table.
https://r3-cev.atlassian.net/browse/CORDA-791

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -227,7 +220,7 @@ class NodeVaultService(
             results.asSequence().forEach {
                 val txHash = SecureHash.parse(it.stateRef?.txId!!)
                 val index = it.stateRef?.index!!
-                val state = it.contractState.deserialize<TransactionState<ContractState>>(context = STORAGE_CONTEXT)
+                val state = stateLoader.loadState(StateRef(txHash,index))
```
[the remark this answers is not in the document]
josecoll: Refactored to use VaultQuery query function.

josecoll:
@rick-r3 there was a subtle but significant change (in terms of transaction visibility correctness) which impacted several VaultQuery tests which were not correctly setting transaction boundaries.
Specifically the private `loadStates` method in `NodeVaultService` now uses the VaultQuery API which creates a new session boundary (whereas previous code was re-using existing NodeVaultService session).

josecoll:
Interestingly the changes in `NodeVaultService` to re-use the same transaction (session) do not ripple down to the `AbstractCashSelection` code, which creates a new `Connection` from the `ServiceHub jdbcSession()` for execution of its Prepared Statement native query.

The implication of this is that Cash Spending does not have visibility of prior Cash Issuance unless it was in a separately demarcated transaction (aprox. 6 unit tests surfaced this behaviour in `VaultQueryTests`).

PR #2035, opened 2017-11-10
Added some more details regarding setting log levels for individual modules

ragmondo:
As titled - more details regarding setting log levels for modules.

### DIFF
--- a/docs/source/node-administration.rst
+++ b/docs/source/node-administration.rst
@@ -10,9 +10,43 @@ Logging
 
 By default the node log files are stored to the ``logs`` subdirectory of the working directory and are rotated from time
 to time. You can have logging printed to the console as well by passing the ``--log-to-console`` command line flag.
-The default logging level is ``INFO`` which can be adjusted by the ``--logging-level`` command line argument. For more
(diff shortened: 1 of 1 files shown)


docs/source/node-administration.rst
```diff
@@ -10,9 +10,43 @@ Logging
...
+
+    <?xml version="1.0" encoding="UTF-8"?>
+        <Configuration status="WARN">
+            <Appenders>
+                <Console name="Console" target="SYSTEM_OUT">
+                    <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+                </Console>
+            </Appenders>
+            <Loggers>
+                <Logger name="org.hibernate" level="debug" additivity="false">
```
roger3cev: How do we know what the logger is called? Is there an easy way to find out? Cheers

docs/source/node-administration.rst
```diff
@@ -10,9 +10,43 @@ Logging
 
 By default the node log files are stored to the ``logs`` subdirectory of the working directory and are rotated from time
```
m4ksio: This may not be true with your custom config - should we describe that custom config changes everything?

PR #2042, opened 2017-11-13
Fix the BFT startup consensus workaround

andr3ej:
i broke it in #2036, must have been lucky with the tests

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -108,7 +108,7 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
             override val database: CordaPersistence,
             override val rpcOps: CordaRPCOps,
             flowStarter: FlowStarter,
-            internal val schedulerService: NodeSchedulerService) : StartedNode<N> {
+            override val notaryService: NotaryService?) : StartedNode<N> {
         override val services: StartedNodeServices = object : StartedNodeServices, ServiceHubInternal by services, FlowStarter by flowStarter {}
     }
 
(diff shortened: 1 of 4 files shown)


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -108,7 +108,7 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
             override val database: CordaPersistence,
             override val rpcOps: CordaRPCOps,
             flowStarter: FlowStarter,
-            internal val schedulerService: NodeSchedulerService) : StartedNode<N> {
```
andr3ej: schedulerService was only using StartedNode as a ferry

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -323,13 +323,13 @@ class MockNetwork(defaultParameters: MockNetworkParameters = MockNetworkParamete
...
-                override fun waitUntilAllReplicasHaveInitialized(notaryService: BFTNonValidatingNotaryService) {
-                    val clusterNodes = mockNet.nodes.filter { notaryKey in it.started!!.info.legalIdentities.map { it.owningKey } }
+                override fun waitUntilAllReplicasHaveInitialized() {
+                    val clusterNodes = mockNet.nodes.map { it.started!! }.filter { notaryKey in it.info.legalIdentities.map { it.owningKey } }
                     if (clusterNodes.size != bftSMaRtConfig.clusterAddresses.size) {
                         throw IllegalStateException("Unable to enumerate all nodes in BFT cluster.")
                     }
                     clusterNodes.forEach {
-                        notaryService.waitUntilReplicaHasInitialized()
+                        (it.notaryService as BFTNonValidatingNotaryService).waitUntilReplicaHasInitialized()
```
andr3ej: it needs to query the notaryService of **other** nodes (which is why only mock framework supports this workaround)

PR #2043, opened 2017-11-13
Consistent database columns naming convention.

szymonsztuka:
* Consistent database column naming convention.
* Remove some column length constraints.

### DIFF
--- a/node/src/test/kotlin/net/corda/node/services/schema/NodeSchemaServiceTest.kt
+++ b/node/src/test/kotlin/net/corda/node/services/schema/NodeSchemaServiceTest.kt
@@ -87,7 +87,7 @@ object TestSchema : MappedSchema(SchemaFamily::class.java, 1, setOf(Parent::clas
     @Table(name = "Children")
     class Child {
         @Id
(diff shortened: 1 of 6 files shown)


node/src/test/kotlin/net/corda/node/services/schema/NodeSchemaServiceTest.kt
```diff
@@ -87,7 +87,7 @@ object TestSchema : MappedSchema(SchemaFamily::class.java, 1, setOf(Parent::clas
     @Table(name = "Children")
     class Child {
         @Id
-        @GeneratedValue(strategy = GenerationType.IDENTITY)
+        @GeneratedValue(strategy = GenerationType.AUTO)
```
josecoll: GenerationType.AUTO is the default so no need to specify.
https://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#identifiers-generators

szymonsztuka: Fixed

PR #2045, opened 2017-11-14
Removed ValidateClasspathTest smoke test, as NodeProcess.Factory does…

shamsasari:
… the same thing and has wider coverage


.idea/compiler.xml
```diff
@@ -54,8 +59,13 @@
       <module name="finance_test" target="1.8" />
       <module name="graphs_main" target="1.8" />
       <module name="graphs_test" target="1.8" />
+      <module name="irs-demo-cordapp_integrationTest" target="1.8" />
       <module name="irs-demo-cordapp_main" target="1.8" />
+      <module name="irs-demo-cordapp_main~1" target="1.8" />
```
andr3ej: this doesn't look right at all. can you revert this file to master and refresh gradle in intellij and see what happens?

shamsasari: Urgh. Yeah sure

shamsasari: Same thing, it goes back to this.

andr3ej: i've seen this once before, trying to remember what i was doing - maybe switching between 2 corda repos? or maybe you've upgraded intellij or some plugin recently? because these changes don't make sense - if i remove all module elements and refresh gradle, it brings back fewer.

shamsasari: I've just reverted the file. I'm not sure what's going on either, but I don't want to spend time right now looking into it.

PR #2046, opened 2017-11-14
CORDA-767: Raft Notary: remove snapshotting

thschroeter:
We remove snapshotting since it is not effective in our use case. We keep the Raft command log "forever" to allow new nodes to join the cluster and replay the commands.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/transactions/DistributedImmutableMap.kt
+++ b/node/src/main/kotlin/net/corda/node/services/transactions/DistributedImmutableMap.kt
@@ -27,9 +27,9 @@ class DistributedImmutableMap<K : Any, V : Any, E, EK>(val db: CordaPersistence,
     object Commands {
         class PutAll<K, V>(val entries: Map<K, V>) : Command<Map<K, V>> {
             override fun compaction(): Command.CompactionMode {
-                // The SNAPSHOT compaction mode indicates that a command can be removed from the Raft log once
-                // a snapshot of the state machine has been written to disk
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/services/transactions/DistributedImmutableMap.kt
```diff
@@ -58,7 +65,10 @@ class DistributedImmutableMap<K : Any, V : Any, E, EK>(val db: CordaPersistence,
             db.transaction {
                 val entries = commit.operation().entries
                 log.debug("State machine commit: storing entries with keys (${entries.keys.joinToString()})")
-                for (key in entries.keys) map[key]?.let { conflicts[key] = it }
+                for (key in entries.keys) map[key]?.let {
+                    if (entries[key] != it)
```
rick-r3: This changes the behaviour I think.  Is it correct?  Previously, if the we submitted the same transaction twice, we'd register a conflict it seems here (i.e. value was the same), but also report a conflict to the client.

Now we won't register a conflict if the value is the same, which seems correct, but we will put them into the map again (I think this generates a warning).  

There seems to be a test for this at the `DistributedImmutableMap` level, but I think there should be `UniquenessProvider` tests against the `RaftUniquenessProvider` in some way, because I'm not convinced it was correct before (or now).

rick-r3: So it looks like the client filters for duplicate source transactions if the states were previously committed.  Therefore we should report a conflict even if the value is the same.

PR #2047, opened 2017-11-14
Add config library and default config to webserver JAR

rick-r3:
Changes to make node.conf more powerful resulted in webserver capsule JAR throwing class not found exception due to missing config parsing library.  Also include default config to avoid error about missing default config.

### DIFF
--- a/webserver/webcapsule/build.gradle
+++ b/webserver/webcapsule/build.gradle
@@ -10,6 +10,12 @@ description 'Corda node web server capsule'
 
 configurations {
     runtimeArtifacts
+    capsuleRuntime
+}
+
+dependencies {
+    // TypeSafe Config: for simple and human friendly config files.
+    capsuleRuntime "com.typesafe:config:$typesafe_config_version"
 }
 
 // Force the Caplet to target Java 6. This ensures that running 'java -jar corda.jar' on any Java 6 VM upwards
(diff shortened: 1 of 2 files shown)


rick-r3:
Makes my webserver testing locally pass.

webserver/webcapsule/build.gradle
```diff
@@ -31,6 +31,10 @@ task buildWebserverJar(type: FatCapsule, dependsOn: project(':node').compileJava
             "$rootDir/config/dev/log4j2.xml"
     )
     from 'NOTICE' // Copy CDDL notice
+    from { project(':node').configurations.runtime.allDependencies.matching { // Include config library JAR.
+        it.group.equals("com.typesafe") && it.name.equals("config")
+    }.collect { zipTree(project(':node').configurations.runtime.files(it).first()) } }
+    from { "$rootDir/node/build/resources/main/reference.conf" }
```
rick-r3: These lines are a direct lift from node/capsule/build.gradle

They include the config library and the default config file, both of which are now used by the capsule wrapper.

Clintonio: The more "gradle" way of doing this would be to specify a new configuration set called say, capsuleRuntime, and then add these libs to those and collect the dependencies specifically from that configuration rather than filtering. The only reason we do the filtering in the plugins is because we can't expect users to know about even more specific configuration types.

Could you either change both to do this or add a TODO?

rick-r3:
Applied @Clintonio suggestion to webserver and node capsule building.

PR #2051, opened 2017-11-14
Add generation of network parameters to deployNodes task

kasiastreich:
Add utility NetworkParametersGenerator run during deployNodes task as a
separate process after generation of node infos.
Move NetworkParametersCopier to node-api/internal.


kasiastreich:
Closing after discussion that class loading is better solution.

PR #2052, opened 2017-11-14
Clean up identity tests

rnicoll:
* Move test identities into MockServices
* Step through cert path when registering


rnicoll:
Closing as now part of #1993

PR #2053, opened 2017-11-14
Added exception if same attachment uploaded. Added test

ragmondo:
As found in https://stackoverflow.com/questions/47284059/can-not-get-file-exist-exception-after-corda-1-0

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
@@ -46,6 +46,13 @@ interface AttachmentStorage {
     fun importAttachment(jar: InputStream, uploader: String, filename: String): AttachmentId
 
     /**
+     * Inserts or returns Attachment Id of attachment. Does not throw an exception if already uploaded.
+     * @param jar [InputStream] of Jar file
+     * @return [AttachmentId] of uploaded attachment
+     */
+    fun importOrGetAttachment(jar: InputStream): AttachmentId
+
+    /**
      * Searches attachment using given criteria and optional sort rules
      * @param criteria Query criteria to use as a filter
      * @param sorting Sorting definition, if not given, order is undefined
@@ -53,5 +60,12 @@ interface AttachmentStorage {
      * @return List of AttachmentId of attachment matching criteria, sorted according to given sorting parameter
      */
     fun queryAttachments(criteria: AttachmentQueryCriteria, sorting: AttachmentSort? = null): List<AttachmentId>
+
+    /**
+     * Searches for an attachment already in the store
+     * @param attachmentId The attachment Id
+     * @return true if it's in there
+     */
+    fun hasAttachment(attachmentId: AttachmentId): Boolean
 }
 
--- a/node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
@@ -2,18 +2,24 @@ package net.corda.node.internal.cordapp
 
 import com.google.common.collect.HashBiMap
 import net.corda.core.contracts.ContractClassName
-import net.corda.core.crypto.SecureHash
-import net.corda.core.node.services.AttachmentStorage
 import net.corda.core.cordapp.Cordapp
 import net.corda.core.cordapp.CordappContext
+import net.corda.core.crypto.SecureHash
 import net.corda.core.node.services.AttachmentId
+import net.corda.core.node.services.AttachmentStorage
 import net.corda.core.serialization.SingletonSerializeAsToken
+import net.corda.core.utilities.loggerFor
 import java.net.URL
 
 /**
  * Cordapp provider and store. For querying CorDapps for their attachment and vice versa.
  */
 open class CordappProviderImpl(private val cordappLoader: CordappLoader, attachmentStorage: AttachmentStorage) : SingletonSerializeAsToken(), CordappProviderInternal {
+
+    companion object {
+        private val log = loggerFor<CordappProviderImpl>()
+    }
+
     override fun getAppContext(): CordappContext {
         // TODO: Use better supported APIs in Java 9
         Exception().stackTrace.forEach { stackFrame ->
@@ -45,7 +51,7 @@ open class CordappProviderImpl(private val cordappLoader: CordappLoader, attachm
 
     private fun loadContractsIntoAttachmentStore(attachmentStorage: AttachmentStorage): Map<SecureHash, URL> {
         val cordappsWithAttachments = cordapps.filter { !it.contractClassNames.isEmpty() }.map { it.jarPath }
-        val attachmentIds = cordappsWithAttachments.map { it.openStream().use { attachmentStorage.importAttachment(it) } }
+        val attachmentIds = cordappsWithAttachments.map { it.openStream().use { attachmentStorage.importOrGetAttachment(it) }}
         return attachmentIds.zip(cordappsWithAttachments).toMap()
     }
 
--- a/node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
+++ b/node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
@@ -1,18 +1,19 @@
 package net.corda.node.services.persistence
 
 import com.codahale.metrics.MetricRegistry
-import net.corda.core.internal.VisibleForTesting
 import com.google.common.hash.HashCode
 import com.google.common.hash.Hashing
 import com.google.common.hash.HashingInputStream
 import com.google.common.io.CountingInputStream
 import net.corda.core.CordaRuntimeException
-import net.corda.core.contracts.*
-import net.corda.core.internal.AbstractAttachment
+import net.corda.core.contracts.Attachment
 import net.corda.core.crypto.SecureHash
+import net.corda.core.internal.AbstractAttachment
+import net.corda.core.internal.VisibleForTesting
 import net.corda.core.node.services.AttachmentId
 import net.corda.core.node.services.AttachmentStorage
-import net.corda.core.node.services.vault.*
+import net.corda.core.node.services.vault.AttachmentQueryCriteria
+import net.corda.core.node.services.vault.AttachmentSort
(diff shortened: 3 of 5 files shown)


ragmondo:
Hmmm. Premature commit.. Started running all the tests on my machine - they were going ok but it seems as if I have unearthed a number of failing samples now we have the exception back.. Watch this space.

core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
```diff
@@ -16,6 +16,7 @@ typealias AttachmentId = SecureHash
  */
 @DoNotImplement
 interface AttachmentStorage {
+
```
josecoll: remove blank line

core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
```diff
@@ -53,5 +54,19 @@ interface AttachmentStorage {
...
+     * @return true if it's in there
+     */
+    fun hasAttachment(attachmentId: AttachmentId): Boolean
+
+    /**
+     * Returns what would be the AttachmentId and also the bytes of the jar if this InputStream was saved into the Vault.
+     * @param jar The InputStream of the jar file to be potentially saved into the Vault
+     * @return Pair<AttachmentId, ByteArray> as constructed from the inbound stream
+     */
+    fun getAttachmentIdAndBytes(jar: InputStream): Pair<AttachmentId, ByteArray>
```
josecoll: Seems a rather strange public API function.
Why would you want a to retrieve an attachment as a `ByteArray` ?

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
```diff
@@ -45,7 +46,15 @@ open class CordappProviderImpl(private val cordappLoader: CordappLoader, attachm
 
     private fun loadContractsIntoAttachmentStore(attachmentStorage: AttachmentStorage): Map<SecureHash, URL> {
         val cordappsWithAttachments = cordapps.filter { !it.contractClassNames.isEmpty() }.map { it.jarPath }
-        val attachmentIds = cordappsWithAttachments.map { it.openStream().use { attachmentStorage.importAttachment(it) } }
+        val attachmentIds = cordappsWithAttachments.map {
+            val attachmentId = try {
+                attachmentStorage.importAttachment(it.openStream())
+            } catch (exception: FileAlreadyExistsException) {
+                // We don't care if a cordapp has already been installed into the vault.
+                AttachmentId.parse(exception.message!!)
```
josecoll: Shouldn't you just throw the exception after logging it ? (or just log a WARNING and do nothing more)
(eg. this code will most likely through a misleading exception: `IllegalArgumentException("Provided string is ${it.size} bytes not 32 bytes in hex: $str")`)

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -171,36 +166,69 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
         return import(jar, uploader, filename)
     }
 
-    // TODO: PLT-147: The attachment should be randomised to prevent brute force guessing and thus privacy leaks.
-    private fun import(jar: InputStream, uploader: String?, filename: String?): AttachmentId {
-        require(jar !is JarInputStream)
+    companion object {
```
josecoll: Move back to top of file (general convention for logger).

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -171,36 +166,69 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
...
+        private val log = loggerFor<NodeAttachmentService>()
 
-        // Read the file into RAM, hashing it to find the ID as we go. The attachment must fit into memory.
-        // TODO: Switch to a two-phase insert so we can handle attachments larger than RAM.
-        // To do this we must pipe stream into the database without knowing its hash, which we will learn only once
-        // the insert/upload is complete. We can then query to see if it's a duplicate and if so, erase, and if not
-        // set the hash field of the new attachment record.
+        private fun checkIsAValidJAR(stream: InputStream) {
+            // Just iterate over the entries with verification enabled: should be good enough to catch mistakes.
```
josecoll: Move comment back out of function.

core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
josecoll: Seems like this function is only used in a `private` method of the NodeAttachmentService (implementation), so does it need to be in public API?

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -171,36 +166,69 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
...
+                require(!entryPath.isAbsolute) { "Path $entryPath is absolute" }
+                require(entryPath.normalize() == entryPath) { "Path $entryPath is not normalised" }
+                require(!('\\' in cursor.name || cursor.name == "." || cursor.name == "..")) { "Bad character in $entryPath" }
+                count++
+            }
+            require(count > 0) { "Stream is either empty or not a JAR/ZIP" }
+        }
+    }
+
+    override fun getAttachmentIdAndBytes(jar: InputStream): Pair<AttachmentId, ByteArray> {
```
josecoll: This function is only used privately in this class.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
josecoll: Could probably do with a new JUnit in `CordappProviderImplTests` to validate condition.

core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
ragmondo: Sooo... the issue is the following: One needs to know the hash of an attachment to determine if the attachment is already in the store. However, in calculating the hash, you need to consume the input stream, so if you don't have the attachment stored, you no longer have the stream in order to store the attachment. Hence, the function needs to return both the hash (so you can check to see if you have it stored), plus the original data so you can store if you need to. I can't think of a way of getting round this *apart* from storing the attachment as you stream it and then discarding it if it's a duplicate - but then this also has it's issues. 

Nonetheless, I will move the function back out of the public API. I was actually intending to write the exception handling part a bit better .. but then that would probably require re-doing the whole class design.

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
ragmondo: moved

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
ragmondo: Ok.

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
ragmondo: moved

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
ragmondo: The whole point is that this exception is thrown only in the case of a duplicate attachment being uploaded. However, the CordApp code doesn't care about uploading the attachment N times :-/ 

Then I started trying to work around some kind of hash calculation (hence the move of `getAttachementIdandBytes` to public), but then didn't want to pollute this part of the code.

core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
josecoll: Right - so just to keep the Public API clean and meaningful, I would just add another function:
`fun importOrGetAttachment(jar: InputStream): Attachment`
which always succeeds (and just re-uses the original import and open attachment logic)

core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
```diff
@@ -45,6 +45,13 @@ interface AttachmentStorage {
     @Throws(FileAlreadyExistsException::class, IOException::class)
     fun importAttachment(jar: InputStream, uploader: String, filename: String): AttachmentId
 
+    /**
+     * Inserts or returns Attachment Id of attachment. Does not throw an exception if already uploaded.
+     * @param uploader Uploader name
+     * @param filename Name of the file
+     */
+    fun importOrGetAttachment(jar: InputStream): AttachmentId
```
josecoll: Minor change here. 
Return the `Attachment` itself (which contains a reference to its `AttachmentId`) so end user doesn't need to make a subsequent call to `openAttachment`.

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -231,6 +231,15 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
         }
     }
 
+    override fun importOrGetAttachment(jar: InputStream): AttachmentId {
```
josecoll: Add a new JUnit test to cover this new call (validating both possible execution paths).

core/src/main/kotlin/net/corda/core/node/services/AttachmentStorage.kt
ragmondo: I guess we could do this - but then this would not be the same as the `importAttachment` function which might cause confusion later, and we can't change the `importAttachment` function as that's been published now.

josecoll: Understood. In that case, its good to go!

PR #2054, opened 2017-11-14
Use NetworkMap and SignedNetworkMap in NetworkMapClient, and enable signature verification.

patrickkuo:
New NetworkMap and SignedNetworkMap object and verify signature and root certificate in Signed network map and node info

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/NetworkMap.kt
+++ b/core/src/main/kotlin/net/corda/core/node/NetworkMap.kt
@@ -1,9 +1,20 @@
 package net.corda.core.node
 
+import net.corda.core.crypto.DigitalSignature
+import net.corda.core.crypto.SecureHash
+import net.corda.core.crypto.SignedData
 import net.corda.core.identity.Party
 import net.corda.core.serialization.CordaSerializable
+import net.corda.core.serialization.serialize
+import java.security.cert.CertPath
 import java.time.Duration
 import java.time.Instant
+
+/**
+ * Data class containing hash of [NetworkParameters] and network participant's [NodeInfo] hashes.
+ */
+@CordaSerializable
+data class NetworkMap(val nodeInfoHashes: List<SecureHash>, val networkParameterHash: SecureHash)
 
 /**
  * @property minimumPlatformVersion
@@ -38,3 +49,9 @@ data class NetworkParameters(
  */
 @CordaSerializable
 data class NotaryInfo(val identity: Party, val validating: Boolean)
+
+class SignedNetworkMap(networkMap: NetworkMap, val signatureAndCert: DigitalSignatureWithCertPath)
+    : SignedData<NetworkMap>(networkMap.serialize(), signatureAndCert)
+
+class DigitalSignatureWithCertPath(val certPath: CertPath, signatureBytes: ByteArray)
+    : DigitalSignature.WithKey(certPath.certificates.first().publicKey, signatureBytes)
--- a/node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
+++ b/node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
@@ -1,11 +1,13 @@
 package net.corda.node.services.network
 
-import com.fasterxml.jackson.databind.ObjectMapper
 import com.google.common.util.concurrent.MoreExecutors
 import net.corda.core.crypto.SecureHash
 import net.corda.core.crypto.SignedData
 import net.corda.core.internal.openHttpConnection
+import net.corda.core.node.NetworkMap
+import net.corda.core.node.NetworkParameters
 import net.corda.core.node.NodeInfo
+import net.corda.core.node.SignedNetworkMap
 import net.corda.core.serialization.deserialize
 import net.corda.core.serialization.serialize
 import net.corda.core.utilities.loggerFor
@@ -13,6 +15,7 @@ import net.corda.core.utilities.minutes
 import net.corda.core.utilities.seconds
 import net.corda.node.services.api.NetworkMapCacheInternal
 import net.corda.node.utilities.NamedThreadFactory
+import net.corda.node.utilities.X509Utilities
 import okhttp3.CacheControl
 import okhttp3.Headers
 import rx.Subscription
@@ -20,11 +23,12 @@ import java.io.BufferedReader
 import java.io.Closeable
 import java.net.HttpURLConnection
 import java.net.URL
+import java.security.cert.X509Certificate
 import java.time.Duration
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 
-class NetworkMapClient(compatibilityZoneURL: URL) {
+class NetworkMapClient(compatibilityZoneURL: URL, private val trustedRoot: X509Certificate) {
     companion object {
         val logger = loggerFor<NetworkMapClient>()
     }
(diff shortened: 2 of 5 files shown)


node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -46,14 +51,29 @@ class NetworkMapClient(compatibilityZoneURL: URL) {
...
 
     fun getNodeInfo(nodeInfoHash: SecureHash): NodeInfo? {
-        val conn = URL("$networkMapUrl/$nodeInfoHash").openHttpConnection()
+        val conn = URL("$networkMapUrl/node-info/$nodeInfoHash").openHttpConnection()
+        return if (conn.responseCode == HttpURLConnection.HTTP_NOT_FOUND) {
+            null
+        } else {
+            val signedNodeInfo = conn.inputStream.use { it.readBytes() }.deserialize<SignedData<NodeInfo>>()
+            val nodeInfo = signedNodeInfo.verified()
+            // Verify node info is signed by node identity
```
shamsasari: Can you add a TODO here that the NodeInfo will be signed by its multiple identities

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -111,21 +131,27 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
                     val (networkMap, cacheTimeout) = networkMapClient.getNetworkMap()
                     val currentNodeHashes = networkMapCache.allNodeHashes
-                    (networkMap - currentNodeHashes).mapNotNull {
+                    (networkMap.nodeInfoHashes - currentNodeHashes).mapNotNull {
                         // Download new node info from network map
-                        networkMapClient.getNodeInfo(it)
+                        try {
+                            networkMapClient.getNodeInfo(it)
+                        } catch (t: Throwable) {
+                            // Fail to retrieve one node info shouldn't stop the whole update, log and return null instead.
```
shamsasari: Typo: Failure

core/src/main/kotlin/net/corda/core/node/NetworkMap.kt
```diff
@@ -1,11 +1,24 @@
 package net.corda.core.node
```
shamsasari: Move this to node-api

core/src/main/kotlin/net/corda/core/node/NetworkMap.kt
```diff
@@ -38,3 +51,9 @@ data class NetworkParameters(
  */
 @CordaSerializable
 data class NotaryInfo(val identity: Party, val validating: Boolean)
+
+class SignedNetworkMap(val networkMap: NetworkMap, val signatureAndCert: DigitalSignatureWithCertificate)
+    : SignedData<NetworkMap>(networkMap.serialize(), signatureAndCert)
+
+class DigitalSignatureWithCertificate(val certificate: X509CertificateHolder, val signatureBytes: ByteArray)
```
shamsasari: Move this inside `DigitalSignature`, like `WithKey` and call it `WithCert`

shamsasari: This also needs docs

core/src/main/kotlin/net/corda/core/node/NetworkMap.kt
```diff
@@ -38,3 +51,9 @@ data class NetworkParameters(
  */
 @CordaSerializable
 data class NotaryInfo(val identity: Party, val validating: Boolean)
+
+class SignedNetworkMap(val networkMap: NetworkMap, val signatureAndCert: DigitalSignatureWithCertificate)
```
shamsasari: Follow the pattern in `SignedData`, i.e. use of SerializedBytes

shamsasari: docs

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -111,21 +131,27 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+                            logger.warn("Error encountered when downloading node info '$it', skipping...", t)
+                            null
+                        }
                     }.forEach {
                         // Add new node info to the network map cache, these could be new node info or modification of node info for existing nodes.
                         networkMapCache.addNode(it)
                     }
                     // Remove node info from network map.
-                    (currentNodeHashes - networkMap - fileWatcher.processedNodeInfoHashes)
+                    (currentNodeHashes - networkMap.nodeInfoHashes - fileWatcher.processedNodeInfoHashes)
```
shamsasari: You can extract out `networkMap.nodeInfoHashes` to a `val`

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -111,21 +131,27 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
                             .mapNotNull(networkMapCache::getNodeByHash)
                             .forEach(networkMapCache::removeNode)
-
+                    // TODO: Update NetworkParameter.
```
shamsasari: The updating of network parameters is a separate process, discussed in my design doc. Here we just check the network parameters is the one we expect. If it's not we shutdown and log an error.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -46,14 +51,29 @@ class NetworkMapClient(compatibilityZoneURL: URL) {
...
+        val networkMap = signedNetworkMap.verified()
+        // Verify certificate path is valid.
+        X509Utilities.validateCertificateChain(trustedRoot, signedNetworkMap.signatureAndCert.certificate.cert, trustedRoot)
         val timeout = CacheControl.parse(Headers.of(conn.headerFields.filterKeys { it != null }.mapValues { it.value.first() })).maxAgeSeconds().seconds
         return NetworkMapResponse(networkMap, timeout)
     }
 
     fun getNodeInfo(nodeInfoHash: SecureHash): NodeInfo? {
-        val conn = URL("$networkMapUrl/$nodeInfoHash").openHttpConnection()
+        val conn = URL("$networkMapUrl/node-info/$nodeInfoHash").openHttpConnection()
```
mkit: This changes the URL for the network map. Shall we add a TODO here?

core/src/main/kotlin/net/corda/core/crypto/DigitalSignature.kt
```diff
@@ -46,6 +48,10 @@ open class DigitalSignature(bytes: ByteArray) : OpaqueBytes(bytes) {
          */
         @Throws(InvalidKeyException::class, SignatureException::class)
         fun isValid(content: ByteArray) = by.isValid(content, this)
-        fun withoutKey() : DigitalSignature = DigitalSignature(this.bytes)
+
+        fun withoutKey(): DigitalSignature = DigitalSignature(this.bytes)
     }
+
+    /** A digital signature that identifies who the public key is owned by, and the certificate which provides prove of the identity */
+    class WithCert(val certificate: X509CertificateHolder, val signatureBytes: ByteArray) : WithKey(certificate.cert.publicKey, signatureBytes)
```
shamsasari: Use `X509Certificate`. I've update the serializers to not use X509CertificateHolder. That was merged into master and now our branch.

shamsasari: This shouldn't extend WithKey

node-api/src/main/kotlin/net/corda/nodeapi/NetworkMap.kt
```diff
@@ -38,3 +48,11 @@ data class NetworkParameters(
...
 @CordaSerializable
 data class NotaryInfo(val identity: Party, val validating: Boolean)
+
+
+/**
+ * A serialized [NetworkMap] and its signature and certificate. Enforces signature validity in order to deserialize the data
+ * contained within.
+ */
+class SignedNetworkMap(val rawNetworkMap: SerializedBytes<NetworkMap>, val signatureWithCert: DigitalSignature.WithCert)
+    : SignedData<NetworkMap>(rawNetworkMap, signatureWithCert)
```
shamsasari: Let's not extend SignedData

patrickkuo: hmm why? Will have to copy the verify method if not extend the class, same for the Signature.WithKey class.

shamsasari: I don't see how a `Signature.WithCert` instance is also a `Signature.WithKey`. This subclass relationship doesn't seem right.

patrickkuo: A certificate contains the key, x500name and the signature of the issuer; `Signature.WithCert` is really just `Signature.WithKey` + those extra data... and they share the same signature verification logic.

shamsasari: Well if Signature.WithCert is a subclass of Signature.WithKey then technically you could use SignedData as is and pass in a Signature.WithCert to it. But then you'd need to do casting. Neither this or extending SignedData here is clean.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
shamsasari: Make sure to update the docs

shamsasari:
What was the issue with the API test? Was it failing because we added new API?

patrickkuo:
moving NetworkParameter breaks the API test, its in the api-current list.

shamsasari:
In what way did it break it? We didn't change any of the existing APIs. Was it because we added new APIs?

patrickkuo:
looks like you have added NetworkParameter to the api-current.txt last week. I want to move the class  to node-api package, and the API test fail because of that. I think better to do the move on a separate PR?

shamsasari:
Yes, add a TODO

patrickkuo:
Ah... thats only on our feature branch, maybe I should just change the api-current file?

shamsasari:
That would be the simpler option

PR #2049, opened 2017-11-14
CORDA-787 Remove unused class

andr3ej:
and minor refactorings

PR #2063, opened 2017-11-15
CORDA-787 Split NodeMessagingClient into 3

andr3ej:
includes commits from #2049


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -199,9 +206,10 @@ open class Node(configuration: NodeConfiguration,
             runOnStop += this::stop
             start()
         }
-
-        // Start up the MQ client.
-        (network as NodeMessagingClient).start(rpcOps, userService)
```
mnesbit: Need to check shutdown

andr3ej: done. network itself is stopped in AbstractNode, i've made it a bit more sensible in #2077

PR #2064, opened 2017-11-15
Consistent database tables naming convention.

szymonsztuka:
* Added explicit table names (mostly for join tables).
* Shorten or alter 2 tables backing notaries.
* Change a compound index declaration to one column index.


josecoll:
Changes look good, and now apply & work consistently across all 3 database providers tested.

PR #2065, opened 2017-11-16
ENT-1069 Allow for nodes to be started in registration mode

al-r3:
Let Driver start nodes in registration mode.

Added a simple test to verify that the started node hits a URL

### DIFF
--- a/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
+++ b/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
@@ -1,5 +1,8 @@
 package net.corda.testing.driver
 
+import com.sun.net.httpserver.HttpExchange
+import com.sun.net.httpserver.HttpHandler
+import com.sun.net.httpserver.HttpServer
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.internal.div
 import net.corda.core.internal.list
@@ -13,6 +16,7 @@ import net.corda.testing.ProjectStructure.projectRootDir
 import net.corda.testing.node.NotarySpec
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
+import java.net.InetSocketAddress
 import java.util.concurrent.Executors
 import java.util.concurrent.ScheduledExecutorService
 
@@ -49,6 +53,37 @@ class DriverTests {
             nodeMustBeUp(nodeInfo)
         }
         nodeMustBeDown(nodeHandle)
+    }
+
+    @Test
+    fun `node registration`() {
+        // Very simple Http handler which counts the requests it has received and always returns the same payload.
+        val handler = object : HttpHandler {
+            var requests = 0
+                private set
+
+            override fun handle(exchange: HttpExchange) {
+                val response = "reply"
+                exchange.responseHeaders.set("Content-Type", "text/html; charset=" + Charsets.UTF_8)
+                exchange.sendResponseHeaders(200, response.length.toLong())
+                exchange.responseBody.use { it.write(response.toByteArray()) }
+                requests++
+            }
+        }
+
+        val inetSocketAddress = InetSocketAddress(0)
+        val server = HttpServer.create(inetSocketAddress,  /* backlog */0)
+        val port = server.address.port
+        server.createContext("/", handler)
+        server.executor = null // creates a default executor
+        server.start()
+
+        driver(portAllocation = PortAllocation.RandomFree) {
+            registerNode(providedName = DUMMY_BANK_A.name, compatibilityZoneURL = "http://localhost:${port}")
+        }
+
+        // We're getting a request to sign the certificate and one poll request to see if the request has been approved.
+        assertThat(handler.requests).isEqualTo(2)
     }
 
     @Test
--- a/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
+++ b/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
@@ -81,7 +81,8 @@ class DriverTests {
         server.start()
 
         driver(compatibilityZoneURL = URL("http://localhost:$port")) {
-            startNode(providedName = DUMMY_BANK_A.name).get()
+            // Wait for the notary to have started.
+            notaryHandles.first().nodeHandles.get()
         }
 
         // We're getting:
--- a/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
+++ b/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
@@ -17,6 +17,7 @@ import net.corda.testing.node.NotarySpec
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
 import java.net.InetSocketAddress
+import java.net.URL
 import java.util.concurrent.Executors
 import java.util.concurrent.ScheduledExecutorService
 
@@ -72,14 +73,14 @@ class DriverTests {
         }
 
         val inetSocketAddress = InetSocketAddress(0)
-        val server = HttpServer.create(inetSocketAddress,  /* backlog */0)
+        val server = HttpServer.create(inetSocketAddress, 0)
         val port = server.address.port
         server.createContext("/", handler)
         server.executor = null // creates a default executor
         server.start()
 
         driver(portAllocation = PortAllocation.RandomFree) {
-            registerNode(providedName = DUMMY_BANK_A.name, compatibilityZoneURL = "http://localhost:${port}")
+            startNode(providedName = DUMMY_BANK_A.name, compatibilityZoneURL = URL("http://localhost:${port}"))
         }
 
         // We're getting a request to sign the certificate and one poll request to see if the request has been approved.
--- a/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
+++ b/testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
@@ -80,7 +80,7 @@ class DriverTests {
         server.executor = null // creates a default executor
         server.start()
 
-        driver(portAllocation = PortAllocation.RandomFree, startNodesInProcess = false) {
+        driver {
             startNode(providedName = DUMMY_BANK_A.name, compatibilityZoneURL = URL("http://localhost:$port"))
                     .get()
         }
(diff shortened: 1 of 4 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -146,6 +145,22 @@ interface DriverDSLExposedInterface : CordformContext {
...
+     *
+     * @param defaultParameters The default parameters for the node. Allows the node to be configured in builder style
+     *   when called from Java code.
+     * @param providedName Name of the node, which will be its legal name in [Party]. The expected use is that one first
+     *   starts a node in registration mode, the node registers and quits, and then a node with the same name is
+     *   started again.
+     * @param compatibilityZoneURL The URL for the compatibility zone, the node will look for a doorman registration
+     *   service on that URL.
+     */
+    fun registerNode(defaultParameters: NodeParameters = NodeParameters(),
```
shamsasari: I'm not sure we want to have an extra method just for this, considering it's not something I think app developers will use. How about the existing startNode method have an extra optional `compatibilityZoneURL`, which if specified will initiate the registration process *and* then once that's approved (and the node shuts down), it will automatically start up the node again but without the `--initial-registration` flag. That way `startNode` returns a Future handle as normal but which only completes once the node is fully registered.

testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
```diff
@@ -52,6 +56,37 @@ class DriverTests {
     }
 
     @Test
+    fun `node registration`() {
+        // Very simple Http handler which counts the requests it has received and always returns the same payload.
+        val handler = object : HttpHandler {
+            var requests = 0
+                private set
```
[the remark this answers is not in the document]
al-r3: Well it changes the behavior of the field, still.
The rest of the unittest code (outside this object) can't write this value with "private set" if I drop this becomes writable.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: Done, 
though we might want one day to have tests for when the request is rejected or similar so we might need to bring this back.

in case you missed that: the node registration is always out of process (regardless of settings) that logic sits in NodeStartup.kt which isn't not even on the codepath if you start the node in process.

shamsasari: Check that out-of-process flag is on

al-r3: Why?
Now it is possible to have the reg be out of process and the rest to be in process which is probably useful.

If I'm checking for out of process then the API of driver is just confusing I can set the out-of-process flag when I create driver but then there is an option that works only if I set out-of-process.

For the record, this was one of the reasons I went for a separate API, so that you knew that startNode respected in-process-ness while nodeRegistration behaved its own way.

shamsasari: I'm not convinced the out-of-process flag override on startNode actually makes any sense.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -856,6 +875,17 @@ class DriverDSL(
...
 
+    private fun startNodeForRegistration(config: Config) {
+        val maximumHeapSize = "200m"
+        val configuration = config.parseAsNodeConfiguration()
+
+        val debugPort = if (isDebug) debugPortAllocation.nextPort() else null
+        val processFuture = startOutOfProcessNode(configuration, config, quasarJarPath, debugPort,
+                systemProperties, cordappPackages, maximumHeapSize, initialRegistration = true)
+        // Wait for the process to complete.
+        processFuture.waitFor(1, MINUTES)
```
shamsasari: We shouldn't be waiting. Return the future and make it part of the startNode pipline.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: Actually, why not just make the registration process exit normally via `main`, rather than with a `System.exit`. That way it can honour the out-of-process flag.

al-r3: That is a good idea, but:
1. shouldn't be part of this PR, as this is about driver (happy to write, raise, merge that one before this one merges)
2. It's still not enough as the registration happens in NodeStartup.kt which is not even touched if you run the process in-node. We might want to refactor that as well, though as above doesn't really belong to a PR about driver.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: Why not? This seems to contraddict your previous advice to just start the node twice when doing registration I need to wait for the node to have quit to restart in non registration mode, I might make this function return a future though the caller will need to wait ifself anyway.

shamsasari: Returning a future is the better thing to do anyway, and in terms of your registerNode API we would want to move the waiting to the caller, just like for the startNode methods. 

However, as I said in my previous comment, I don't want a separate registerNode API. There is no contradiction, this register process returns a Future which completes when the node exits, which is then mapped onto the normal node start process.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: OK

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: ok, got it done, this just makes the start up slightly faster, that's fine.

testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
```diff
@@ -51,6 +56,42 @@ class DriverTests {
...
+        }
+
+        val inetSocketAddress = InetSocketAddress(0)
+        val server = HttpServer.create(inetSocketAddress, 0)
+        val port = server.address.port
+        server.createContext("/", handler)
+        server.executor = null // creates a default executor
+        server.start()
+
+        driver(portAllocation = PortAllocation.RandomFree, startNodesInProcess = false) {
```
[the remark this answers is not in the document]
al-r3: I don't really recall how it ended it's still used in place.
Either way I reverted to the default - which is incremental.

al-r3:
ok - so what should I do here?
Move mine implementation to pass the URL to driver instead?

shamsasari:
I think so, yes. Also can you change this PR to be merged into `feature-network-parameters`.

al-r3:
Done, I've changed the base

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -644,25 +654,50 @@ class DriverDSL(
...
-                        "rpcUsers" to if (users.isEmpty()) defaultRpcUserList else users.map { it.toConfig().root().unwrapped() },
-                        "verifierType" to verifierType.name
-                ) + customOverrides
+                        "p2pAddress" to "localhost:1222",  // required argument, not really used
+                        "compatibilityZoneURL" to compatibilityZoneURL.toString(),
+                        "myLegalName" to providedName.toString())
         )
-        return startNodeInternal(config, webAddress, startInSameProcess, maximumHeapSize)
+        if (startNodesInProcess) {
+            // This is a bit cheating, we're not stating a full node, we're just calling the code nodes call
```
shamsasari: typo: starting

shamsasari: typo: node's

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -991,13 +1042,18 @@ class DriverDSL(
                     "-javaagent:$quasarJarPath=$excludePattern"
             val loggingLevel = if (debugPort == null) "INFO" else "DEBUG"
 
-            return ProcessUtilities.startCordaProcess(
+            val arguments = mutableListOf<String>(
+                    "--base-directory=${nodeConf.baseDirectory}",
+                    "--logging-level=$loggingLevel",
+                    "--no-local-shell").also {
```
shamsasari: This doesn't help with clarity. You're already declaring `arguments` as a mutable list. Use that without `also`. Also the `toList()` doesn't seem to be needed.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -1042,6 +1098,7 @@ class DriverDSL(
 
 fun writeConfig(path: Path, filename: String, config: Config) {
     val configString = config.root().render(ConfigRenderOptions.defaults())
+    path.createDirectories()
```
shamsasari: Can you instead track down where the dir needs creating, rather than here.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: no, it's 'nodes' plural of 'node'.
I can make it singular and have 'calling the code a node calls' but it feels better (to me) as is.

shamsasari: Yes, you're right, but the sentence is not so easy to read.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: I don't follow, the "toList" makes it immutable.
This is a fairly standard pattern to me, I'm creating a mutable list, I add some stuff to it and when I'm done I create an immutable one.

I can keep all in a mutable list, but no modifications of it are expected outside of this statement.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: Moved that in startNodeForRegistration, though that line is where they are first used, by the statement below.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: That's fair enough, but realise that kotlin's `List` and `MutableList` interfaces don't necessarily map to immutable implementations on the JVM. It's just a nice compile-time check. For example, have a look at the implementation of `toList`. In the general case it calls `toMutableList`!

al-r3: I know! That's for the programmer. JVM has little concept of immutability for anything beyond base types.

PR #2066, opened 2017-11-16
Another approach to fixing deployNodes task and network parameters generation

kasiastreich:
As discussed, I use class loading approach in `deployNodes` gradle task just after the `NodeInfos` generation. Unfortunately, I had some crypto packages loading problems that need to be solved. As a quick fix I disabled checking of the signature for cordformation in devMode.


gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -28,8 +27,10 @@ open class Cordform : DefaultTask() {
      */
     @Suppress("MemberVisibilityCanPrivate")
     var definitionClass: String? = null
+    var parametersClass: String? = "net.corda.node.internal.networkParametersGenerator.TestNetworkParametersGenerator"
```
Clintonio: This should be a private val, not public var.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -142,14 +157,39 @@ open class Cordform : DefaultTask() {
         }
     }
 
+    private fun generateAndInstallNetworkParameters() {
+        project.logger.info("Generating network parameters")
+        gatherNotaries()
```
Clintonio: This should return the value and use it locally instead of storing in a class variable, because it isn't used outside of this function.

Clintonio: (Function and callees)

node-api/src/main/kotlin/net/corda/nodeapi/internal/ParametersUtilities.kt
```diff
@@ -1,4 +1,4 @@
-package net.corda.testing.common.internal
+package net.corda.nodeapi.internal
```
Clintonio: Do these need to be in nodeapi?

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -91,20 +94,55 @@ class NodeInfoWatcher(private val nodePath: Path,
...
 
     /**
+     * Loads all NodeInfo files stored in node's base directory and returns the [Party]s.
+     * Scans main directory and [CordformNode.NODE_INFO_DIRECTORY].
+     * Signatures are not checked before returning a value. The latest value stored for a given name is returned.
+     *
+     * @return a list of [Party]s
+     */
+    fun loadAndGatherNotaryIdentities(notaries: Map<CordaX500Name, Boolean>): List<NotaryInfo> {
+        val nodeInfos = loadFromDirectory(false)
```
Clintonio: 104-115 could be a separate, private, function called something like "getAllNodeInfos".

node-api/src/main/kotlin/net/corda/nodeapi/internal/ParametersUtilities.kt
kasiastreich: I needed to have an access to it from node, any better place to have it?

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -91,20 +94,60 @@ class NodeInfoWatcher(private val nodePath: Path,
...
+                notaryInfos.add(NotaryInfo(notaryId, notaries[info.legalIdentities[0].name]!!))
+            }
+        }
+        return notaryInfos.toList()
+    }
+
+    private fun getAllNodeInfos(): Map<CordaX500Name, NodeInfo> {
+        val nodeInfos = loadFromDirectory(false)
+        // NodeInfos are currently stored in 2 places: in [CordformNode.NODE_INFO_DIRECTORY] and in baseDirectory of the node.
+        val myFiles = nodePath.list {it.filter { it.toString().contains("nodeInfo-") }.toList() }
```
shamsasari: `in` instead of `contains`

shamsasari: Space after `{`

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -91,20 +94,60 @@ class NodeInfoWatcher(private val nodePath: Path,
...
+        val myNodeInfos = myFiles.mapNotNull { processFile(it, false) }
+        val infosMap = mutableMapOf<CordaX500Name, NodeInfo>()
+        // Running deployNodes more than once produces new NodeInfos. We need to load the latest NodeInfos based on serial field.
```
shamsasari: This is not a feature but a bug. I thought this was fixed? If it's not can you add a TODO in the relevant bit of the code. Assume there's only one nodeInfo file in the baseDirectory.

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -91,20 +94,60 @@ class NodeInfoWatcher(private val nodePath: Path,
...
+    fun loadAndGatherNotaryIdentities(notaries: Map<CordaX500Name, Boolean>): List<NotaryInfo> {
+        val infosMap = getAllNodeInfos()
+        // Now get the notary identities based on names passed from Cordform configs. There is one problem, for distributed notaries
+        // Cordfom specifies in config only node's main name, the notary identity isn't passed there. It's read from keystore on
+        // node startup, so we have to look it up from node info as a second identity, which is ugly.
+        val notaryInfos = mutableSetOf<NotaryInfo>()
+        for (info in infosMap.values) {
+            // Here the ugliness happens.
+            // TODO Change Cordform definition so it specifies distributed notary identity.
+            if (info.legalIdentities[0].name in notaries.keys) {
```
shamsasari: You can use a series of `map`, `filter`, `duplicate`, etc on `infosMap.value` to do the same thing. It will probably look cleaner

samples/simm-valuation-demo/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
     cordaCompile project(path: ":node:capsule", configuration: 'runtimeArtifacts')
     cordaCompile project(path: ":webserver:webcapsule", configuration: 'runtimeArtifacts')
     cordaCompile project(':core')
+    cordaCompile project(':node')
```
shamsasari: The demos are not supposed to have a dependency to the node. Why is this needed?

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
kasiastreich: When i was working on that it wasn't fixed. Need to check if there is a commit that addresses this issue.

gradle-plugins/cordform-common/src/main/java/net/corda/cordform/NetworkParametersGenerator.java
```diff
@@ -0,0 +1,9 @@
+package net.corda.cordform;
+
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+
+public interface NetworkParametersGenerator {
+    void run(Map<String, Boolean> notaryMap, List<Path> nodesDirs);
```
shamsasari: Document the parameters

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -124,14 +123,11 @@ class Node(private val project: Project) : CordformNode() {
...
-            if (o.size > 0) {
-                o.first().first.value.toString()
-            } else {
-                name
-            }
-        } catch(_ : IllegalArgumentException) {
-            // Can't parse as an X500 name, use the full string
+            // Using sun.security.x509.X500Name because importing BouncyCastle provider in Cordformation causes problems
+            // with loading our custom X509EdDSAEngine.
+            X500Name(name).organization
```
shamsasari: Why not just write a regex to extract out the `O=` bit of the string? Seems better than to import BC or use a private internal API.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -31,5 +31,11 @@ class Cordformation : Plugin<Project> {
 
     override fun apply(project: Project) {
         Utils.createCompileConfiguration("cordapp", project)
+        // Add runtime dependency to node except for the corda project itself to avoid a circular dependency of Corda with itself.
```
shamsasari: Perhaps add a comment on why this needs to be done

node/src/main/kotlin/net/corda/node/internal/networkParametersGenerator/TestNetworkParametersGenerator.kt
```diff
@@ -0,0 +1,62 @@
...
+        return infos.filter { it.legalIdentities[0].name in notaries.keys }.map {
+            NotaryInfo(it.notaryIdentity(), notaries[it.legalIdentities[0].name]!!)
+        }.distinct()
+    }
+
+    private fun NodeInfo.notaryIdentity() = if (legalIdentities.size == 2) legalIdentities[1] else legalIdentities[0]
+
+    private fun initialiseSerialization() {
+        val context = if (java.lang.Boolean.getBoolean("net.corda.testing.amqp.enable"))
+            AMQP_P2P_CONTEXT
```
shamsasari: I think this if-else is small enough to all go on the line above

node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -90,6 +93,29 @@ class NodeInfoWatcher(private val nodePath: Path,
...
     fun saveToFile(signedNodeInfo: SignedData<NodeInfo>) = Companion.saveToFile(nodePath, signedNodeInfo)
 
+    /**
+     * Loads latest NodeInfo files stored in node's base directory.
+     * Scans main directory and [CordformNode.NODE_INFO_DIRECTORY].
+     * Signatures are checked before returning a value. The latest value stored for a given name is returned.
+     *
+     * @return list of latest [NodeInfo]s
+     */
+    fun getAllNodeInfos(): List<NodeInfo> {
```
shamsasari: I still think this should be in `TestNetworkParametersGenerator`. Just make `loadFromDirectory` public.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -121,12 +119,11 @@ class Node(private val project: Project) : CordformNode() {
...
         }
-
         val dirName = try {
-            // Using sun.security.x509.X500Name because importing BouncyCastle provider in Cordformation causes problems
+            // Parsing O= part directly because importing BouncyCastle provider in Cordformation causes problems
             // with loading our custom X509EdDSAEngine.
-            X500Name(name).organization
-        } catch(_ : IOException) {
+            name.trim().split(",").filter { it.startsWith("O=") }.first().substringAfter("=")
+        } catch(_ : NoSuchElementException) {
```
shamsasari: use `firstOrNull {  }`to avoid catching this exception

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -119,14 +119,10 @@ class Node(private val project: Project) : CordformNode() {
...
-            // Parsing O= part directly because importing BouncyCastle provider in Cordformation causes problems
-            // with loading our custom X509EdDSAEngine.
-            name.trim().split(",").filter { it.startsWith("O=") }.first().substringAfter("=")
-        } catch(_ : NoSuchElementException) {
-            // Can't parse as an X500 name or no organisation part, use the full string
-            name
-        }
+        // Parsing O= part directly because importing BouncyCastle provider in Cordformation causes problems
+        // with loading our custom X509EdDSAEngine.
+        val organizationName = name.trim().split(",").filter { it.startsWith("O=") }.firstOrNull()?.substringAfter("=")
```
shamsasari: Can you merge the filter and firstOrNull with just firstOrNull {}

constants.properties
```diff
@@ -1,4 +1,4 @@
-gradlePluginsVersion=2.0.8
+gradlePluginsVersion=2.0.9
```
shamsasari: Append -NETWORKMAP to avoid any conflict with master

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -28,6 +27,7 @@ open class Cordform : DefaultTask() {
      */
     @Suppress("MemberVisibilityCanPrivate")
     var definitionClass: String? = null
+    private val parametersClass: String = "net.corda.node.internal.networkParametersGenerator.TestNetworkParametersGenerator"
```
shamsasari: networkParametersGenClass

node/src/main/kotlin/net/corda/node/internal/networkParametersGenerator/TestNetworkParametersGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+import net.corda.nodeapi.internal.serialization.*
+import net.corda.nodeapi.internal.serialization.amqp.AMQPServerSerializationScheme
+import java.nio.file.Path
+import java.time.Instant
+import kotlin.streams.toList
+
+// This class is used by deployNodes task to generate NetworkParameters in [Cordformation].
+class TestNetworkParametersGenerator : NetworkParametersGenerator {
+    override fun run(notaryMap: Map<String, Boolean>, nodesDirs: List<Path>) {
+        initialiseSerialization()
```
shamsasari: Shouldn't we also be turning off the serialisation context at the end of this method?

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -121,7 +121,7 @@ class Node(private val project: Project) : CordformNode() {
         }
         // Parsing O= part directly because importing BouncyCastle provider in Cordformation causes problems
         // with loading our custom X509EdDSAEngine.
-        val organizationName = name.trim().split(",").filter { it.startsWith("O=") }.firstOrNull()?.substringAfter("=")
+        val organizationName = name.trim().split(",").firstOrNull { it.startsWith("O=") } ?.substringAfter("=")
```
shamsasari: remove space before ?.

node/src/main/kotlin/net/corda/node/internal/networkParametersGenerator/TestNetworkParametersGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+     * Loads latest NodeInfo files stored in node's base directory.
+     * Scans main directory and [CordformNode.NODE_INFO_DIRECTORY].
+     * Signatures are checked before returning a value. The latest value stored for a given name is returned.
+     *
+     * @return list of latest [NodeInfo]s
+     */
+    private fun getAllNodeInfos(baseDirectory: Path): List<NodeInfo> {
+        val infoWatcher = NodeInfoWatcher(baseDirectory)
+        val nodeInfos = infoWatcher.loadFromDirectory()
+        // NodeInfos are currently stored in 2 places: in [CordformNode.NODE_INFO_DIRECTORY] and in baseDirectory of the node.
```
shamsasari: Why do you even need to scan NODE_INFO_DIRECTORY when you can look at the baseDirectory? And if we just need to look at baseDirectory do we even need to use the NodeInfoWatcher? `processFile` is straightforward enough to re-implement.

node/src/main/kotlin/net/corda/node/internal/networkParametersGenerator/TestNetworkParametersGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+        val myFiles = baseDirectory.list { it.filter { "nodeInfo-" in it.toString() }.toList() }
+        val myNodeInfos = myFiles.mapNotNull { infoWatcher.processFile(it) }
+        val infosMap = mutableMapOf<CordaX500Name, NodeInfo>()
+        // Running deployNodes more than once produces new NodeInfos. We need to load the latest NodeInfos based on serial field.
```
shamsasari: Let's not work around this bug as if it's a feature. Just take the first node info file.

node/src/main/kotlin/net/corda/node/internal/networkParametersGenerator/TestNetworkParametersGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+import net.corda.nodeapi.internal.NetworkParametersCopier
+import net.corda.nodeapi.internal.serialization.*
+import net.corda.nodeapi.internal.serialization.amqp.AMQPServerSerializationScheme
+import java.nio.file.Path
+import java.time.Instant
+import kotlin.streams.toList
+
+// This class is used by deployNodes task to generate NetworkParameters in [Cordformation].
+class TestNetworkParametersGenerator : NetworkParametersGenerator {
+    override fun run(notaryMap: Map<String, Boolean>, nodesDirs: List<Path>) {
```
shamsasari: Since we have all the node directories, can we move more of the processing into this method, namely removing the need for `notaryMap` param. Parse each node.conf file and extract out the notary config.

node/src/main/kotlin/net/corda/node/internal/networkParametersGenerator/TestNetworkParametersGenerator.kt
kasiastreich: Because I was using only one node to generate the parameters based on all information it had. I can change it to scan all nodes, so we need only one nodeInfo from each `baseDirectory`.

shamsasari: That makes more sense

node/src/main/kotlin/net/corda/node/internal/networkParametersGenerator/TestNetworkParametersGenerator.kt
kasiastreich: I am not sure how?

shamsasari: Have a look at the companion object of `NodeProcess.Factory`

node-api/src/main/kotlin/net/corda/nodeapi/internal/TestNetworkParametersGenerator.kt
```diff
@@ -0,0 +1,114 @@
...
+        val context = if (java.lang.Boolean.getBoolean("net.corda.testing.amqp.enable")) AMQP_P2P_CONTEXT else KRYO_P2P_CONTEXT
+        _contextSerializationEnv.set(SerializationEnvironmentImpl(
+                SerializationFactoryImpl().apply {
+                    registerScheme(KryoParametersSerializationScheme)
+                    registerScheme(AMQPServerSerializationScheme())
+                },
+                context))
+    }
+
+    private object KryoParametersSerializationScheme : AbstractKryoSerializationScheme() {
```
shamsasari: A comment perhaps why we need this. You were using the stock scheme before.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -113,6 +113,19 @@ open class Cordform : DefaultTask() {
...
+    /**
+     * The parametersGenerator needn't be compiled until just before our build method, so we load it manually via sourceSets.main.runtimeClasspath.
+     */
+    private fun loadParametersGenerator(): NetworkParametersGenerator {
+        val plugin = project.convention.getPlugin(JavaPluginConvention::class.java)
+        val classpath = plugin.sourceSets.getByName(MAIN_SOURCE_SET_NAME).runtimeClasspath
+        val urls = classpath.files.map { it.toURI().toURL() }.toTypedArray()
+        return URLClassLoader(urls, NetworkParametersGenerator::class.java.classLoader)
+                .loadClass(networkParametersGenClass)
+                .asSubclass(NetworkParametersGenerator::class.java)
```
shamsasari: Because we already know the class to instantiate, do you think we could get away with removing the `NetworkParametersGenerator` interface entirely and just call the `run` method using reflection? `NetworkParametersGenerator` is giving us compile-time safety but nothing more than that. WDYT?

node-api/src/main/kotlin/net/corda/nodeapi/internal/TestNetworkParametersGenerator.kt
kasiastreich: It's because the previous one isn't accessible from `node-api`.

shamsasari: I mean in a comment in the code

constants.properties
```diff
@@ -1,4 +1,4 @@
-gradlePluginsVersion=2.0.9
+gradlePluginsVersion=2.0.9-NETWORKMAP
```
Clintonio: Please change to 3.0.0. Also why the -NETWORKMAP?

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
Clintonio: I'm fine keeping it in because in the case this goes wrong the exceptions produced will be much more helpful.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -250,34 +239,17 @@ class Node(private val project: Project) : CordformNode() {
...
-    /**
-     * Find the corda JAR amongst the dependencies
+     * Find the given JAR amongst the dependencies
+     * @param jarName JAR name without the version part, for example for corda-2.0-SNAPSHOT.jar provide only "corda" as jarName
      *
-     * @return A file representing the Corda webserver JAR
+     * @return A file representing found JAR
      */
-    private fun verifyAndGetWebserverJar(): File {
+    private fun verifyAndGetJar(jarName: String): File {
```
Clintonio: Please rename to `verifyAndGetRuntimeJar`

constants.properties
shamsasari: Oops!

kasiastreich: NETWORKMAP was requested by Shams.

shamsasari: (Btw, ignore my "Oops" I thought I was looking at my own PR!)

W.r.t. to the -NETWORKMAP here, we need it to so that we don't clash with any gradle plugins on master. This will be removed once we merge this feature branch into master.

constants.properties
```diff
@@ -1,4 +1,4 @@
-gradlePluginsVersion=2.0.9-NETWORKMAP
+gradlePluginsVersion=3.0.0-NETWORKMAP
```
Clintonio: What is the purpose of -NETWORKMAP here?

shamsasari: https://github.com/corda/corda/pull/2066#discussion_r153797980

PR #2067, opened 2017-11-16
Move class to the only place that uses it

andr3ej:
and clean up after network map team. No jira, just annoying.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/Node.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/Node.kt
@@ -1,7 +1,6 @@
 package net.corda.node.internal
 
 import com.codahale.metrics.JmxReporter
-import net.corda.core.CordaException
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.identity.CordaX500Name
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -317,7 +317,3 @@ open class Node(configuration: NodeConfiguration,
         log.info("Shutdown complete")
     }
 }
-
-class ConfigurationException(message: String) : CordaException(message)
```
andr3ej: it being here causes unnecessary errors during merges

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -317,7 +317,3 @@ open class Node(configuration: NodeConfiguration,
...
-
-data class NetworkMapInfo(val address: NetworkHostAndPort, val legalName: CordaX500Name)
```
andr3ej: unused

fenryka: I'd do this as a separate PR

PR #2070, opened 2017-11-16
Revert "Introducing network parameters."

shamsasari:
This reverts commit d04e487

The above commit broke `deployNodes` because the network-parameters file that it introduced wasn't generated for `deployNodes`, and it was a non-trivial task since the set of valid notaries is defined by the network parameters.

This commit reverts that back, namely notaries are now defined by the contents of the node's commonName.


shamsasari:
I've tested the notary demo and it all works.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -1,40 +0,0 @@
-package net.corda.core.node
```
al-r3: I think it's better to leave this file in the tree it's referenced from enterprise now, removing it would break the dependency doorman has on corda

shamsasari: I doesn't matter because further network map work is moving to a separate feature branch. We will need to get TC building that branch for doorman.

PR #2072, opened 2017-11-17
Bug fix for cash selection on H2 where the pennies amount is larger t…

shamsasari:
…han max int

### DIFF
--- a/finance/src/test/kotlin/net/corda/finance/contracts/asset/cash/selection/CashSelectionH2ImplTest.kt
+++ b/finance/src/test/kotlin/net/corda/finance/contracts/asset/cash/selection/CashSelectionH2ImplTest.kt
@@ -1,8 +1,12 @@
-package net.corda.finance.contracts.asset
+package net.corda.finance.contracts.asset.cash.selection
 
+import net.corda.core.internal.concurrent.transpose
+import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.utilities.getOrThrow
 import net.corda.finance.DOLLARS
+import net.corda.finance.POUNDS
 import net.corda.finance.flows.CashException
+import net.corda.finance.flows.CashIssueFlow
 import net.corda.finance.flows.CashPaymentFlow
 import net.corda.testing.node.MockNetwork
 import net.corda.testing.node.MockNodeParameters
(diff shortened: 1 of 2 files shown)


finance/src/test/kotlin/net/corda/finance/contracts/asset/cash/selection/CashSelectionH2ImplTest.kt
```diff
@@ -11,14 +15,26 @@ import org.assertj.core.api.Assertions.assertThatThrownBy
...
+        val node = mockNet.createNode()
+        val overflowAmount = (Int.MAX_VALUE / 100) + 1  // Smallest £ amount which maps to pennies larger than max int
+        // Split the amount across two states
+        listOf(overflowAmount, 1).map { issueAmount ->
+            node.services.startFlow(CashIssueFlow(issueAmount.POUNDS, OpaqueBytes.of(1), mockNet.defaultNotaryIdentity)).resultFuture
+        }.transpose().getOrThrow()
+        // Attempt to gather the full amount
+        node.services.startFlow(CashPaymentFlow((overflowAmount + 1).POUNDS, node.info.legalIdentities[0])).resultFuture.getOrThrow()
+    }
+
```
josecoll: Is this quite correct?
You issue 2 states:
- `overflowAmount`
- 1
I assume the `overflowAmount` throws an exception (which is ignored) so only one state gets stored in the vault.
Now you request a Cash transfer of `overflowAmount + 1` (why plus 1 ?) which should also throw an exception as there is only one state of 1 in the vault, but I guess this exception is also silently ignored.
So what is this test actually asserting (validating)?

shamsasari: No the issuance of `overflowAmount` doesn't throw an exception, and that's expected because `PersistentCashState.pennies` is a long. The exception does occur on the payment step.

Why spend `overflowAmount + 1`? I think it's because the issue doesn't occur until the SQL logic moves onto the second state to increment to accumulator. It's probably the seem reason why this issue doesn't occur if I create a single cash state.

I'll add more comments to explain.

shamsasari: Oops. I just accidently deleted my comment!

shamsasari: I'll add more comments.

shamsasari:
Added more comments, and used max int values for the pennies to make things clearer.

finance/src/test/kotlin/net/corda/finance/contracts/asset/cash/selection/CashSelectionH2ImplTest.kt
josecoll: OK - got it! Looks good.

PR #2074, opened 2017-11-17
Adding comment to test

mkit:
Adding comment explaining test design.

### DIFF
--- a/testing/node-driver/src/integration-test/kotlin/net/corda/testing/node/FlowStackSnapshotTest.kt
+++ b/testing/node-driver/src/integration-test/kotlin/net/corda/testing/node/FlowStackSnapshotTest.kt
@@ -293,6 +293,8 @@ class FlowStackSnapshotTest {
         node.registerInitiatedFlow(DummyFlow::class.java)
         node.services.startFlow(FlowStackSnapshotSerializationTestingFlow()).resultFuture.get()
         val thrown = try {
+            // Due to the [MockNetwork] implementation, the easiest way to trigger object serialization process is at
+            // the network stopping stage.
             mockNet.stopNodes()
             null
         } catch (exception: Exception) {


testing/node-driver/src/integration-test/kotlin/net/corda/testing/FlowStackSnapshotTest.kt
```diff
@@ -294,6 +294,8 @@ class FlowStackSnapshotTest {
         node.internals.registerInitiatedFlow(DummyFlow::class.java)
         node.services.startFlow(FlowStackSnapshotSerializationTestingFlow()).resultFuture.get()
         val thrown = try {
```
shamsasari: Why do you need to try-catch the exception and then assert nothing was thrown? Isn't that just the same as having no try-catch. If an exception is thrown then the test fails

mkit: So it would be something like this:
`
    @Test
    fun "flowStackSnapshot object is serializable"() {
        val mockNet = MockNetwork(threadPerNode = true)
        val node = mockNet.createPartyNode()
        node.internals.registerInitiatedFlow(DummyFlow::class.java)
        node.services.startFlow(FlowStackSnapshotSerializationTestingFlow()).resultFuture.get()
        mockNet.stopNodes()
    }
`

For me this code as a test is confusing as it doesn't have any assertion/verification in it so it gives a vague impression what the code/test is supposed to do.

By putting it explicitly in the code that we are asserting that no exception is thrown serves also the purpose of explaining (documenting) to the user what is expected from this test.
I agree that from the functional point of view both approaches would end up with a similar result. 
Although, one would be an exception thrown from the code, the other is assertion violation.

mkit: Preferably there should be a method like `assertNoExceptionIsThrown(mockNet.stopNodes())`. 
However, I am not aware of any such a function, so this approach is just a work around to achieve a similar result.

PR #2077, opened 2017-11-17
CORDA-716 Call stop on InMemoryMessagingNetwork

andr3ej:
and related refactorings.

### DIFF
--- a/core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
@@ -6,45 +6,44 @@ import net.corda.core.utilities.UntrustworthyData
 import net.corda.core.utilities.getOrThrow
 import net.corda.core.utilities.unwrap
 import net.corda.testing.chooseIdentity
-import net.corda.testing.node.network
+import net.corda.testing.node.MockNetwork
 import net.corda.testing.startFlow
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.After
 import org.junit.Test
 
 class ReceiveMultipleFlowTests {
+    private val mockNet = MockNetwork()
+    private val nodes = (1..3).map { mockNet.createPartyNode() }
+    @After
+    fun stopNodes() {
+        mockNet.stopNodes()
+    }
+
     @Test
     fun `receive all messages in parallel using map style`() {
-        network(3) { nodes ->
-            val doubleValue = 5.0
-            nodes[1].registerAnswer(AlgorithmDefinition::class, doubleValue)
-            val stringValue = "Thriller"
-            nodes[2].registerAnswer(AlgorithmDefinition::class, stringValue)
-
-            val flow = nodes[0].services.startFlow(ParallelAlgorithmMap(nodes[1].info.chooseIdentity(), nodes[2].info.chooseIdentity()))
-            runNetwork()
-
-            val result = flow.resultFuture.getOrThrow()
-
-            assertThat(result).isEqualTo(doubleValue * stringValue.length)
-        }
+        val doubleValue = 5.0
+        nodes[1].registerAnswer(AlgorithmDefinition::class, doubleValue)
+        val stringValue = "Thriller"
+        nodes[2].registerAnswer(AlgorithmDefinition::class, stringValue)
+        val flow = nodes[0].services.startFlow(ParallelAlgorithmMap(nodes[1].info.chooseIdentity(), nodes[2].info.chooseIdentity()))
+        mockNet.runNetwork()
+        val result = flow.resultFuture.getOrThrow()
+        assertThat(result).isEqualTo(doubleValue * stringValue.length)
     }
 
     @Test
     fun `receive all messages in parallel using list style`() {
-        network(3) { nodes ->
-            val value1 = 5.0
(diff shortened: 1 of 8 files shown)


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -212,8 +212,6 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
             registerCordappFlows()
             _services.rpcFlows += cordappLoader.cordapps.flatMap { it.rpcFlows }
             FlowLogicRefFactoryImpl.classloader = cordappLoader.appClassLoader
-
-            runOnStop += network::stop
```
andr3ej: moved to where NodeMessagingClient/InMemoryMessaging are started

testing/node-driver/src/main/kotlin/net/corda/testing/node/InMemoryMessagingNetwork.kt
```diff
@@ -103,38 +88,15 @@ class InMemoryMessagingNetwork(
         get() = _receivedMessages
 
     val endpoints: List<InMemoryMessaging> @Synchronized get() = handleEndpointMap.values.toList()
-    fun endpoint(peer: PeerHandle): InMemoryMessaging? = handleEndpointMap.get(peer)
```
andr3ej: unused

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -119,7 +117,7 @@ class MockNetwork(defaultParameters: MockNetworkParameters = MockNetworkParamete
                   private val defaultFactory: (MockNodeArgs) -> MockNode = defaultParameters.defaultFactory,
                   initialiseSerialization: Boolean = defaultParameters.initialiseSerialization,
                   private val notarySpecs: List<NotarySpec> = listOf(NotarySpec(DUMMY_NOTARY.name)),
-                  private val cordappPackages: List<String> = defaultParameters.cordappPackages) : Closeable {
```
andr3ej: stopNodes should be called in an @\After method, so Closeable adds negligible value

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -399,6 +396,7 @@ class MockNetwork(defaultParameters: MockNetworkParameters = MockNetworkParamete
     fun stopNodes() {
         nodes.forEach { it.started?.dispose() }
         serializationEnv.unset()
+        messagingNetwork.stop()
```
andr3ej: in practice redundant as we've stopped the individual InMemoryMessaging objs via AbstractNode, but MockNetwork shouldn't assume that it is

core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
```diff
@@ -5,46 +5,45 @@ import net.corda.core.identity.Party
...
-import net.corda.testing.node.network
+import net.corda.testing.node.MockNetwork
 import net.corda.testing.singleIdentity
 import net.corda.testing.startFlow
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.After
 import org.junit.Test
 
 class ReceiveMultipleFlowTests {
+    private val mockNet = MockNetwork()
```
rnicoll: Shouldn't this be in a `@Before` function? My understanding is `@After` is run on every test, so we're creating a network once, stopping all nodes after the first test but not creating a new network for the next test.

I mean, clearly it passes CI, so maybe I'm wrong, but...

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -214,7 +214,10 @@ open class Node(configuration: NodeConfiguration,
             start()
             runOnStop += this::stop
         }
-        (network as P2PMessagingClient).start()
+        (network as P2PMessagingClient).apply {
+            start()
+            runOnStop += this::stop
```
rnicoll: I can see this is following the same pattern as above, but unless there's a good reason not to, I think it would be better to set the `runOnStop` then call `start()`. Theoretically here there's a race condition where the client could stop before `runOnStop` is set.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
rnicoll: Why is an `@After` better than other approaches?

core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
andr3ej: junit creates a new object for each @\Test method. an actual difference is that @\Rule rules aren't in effect when MockNetwork is created, but it doesn't need any.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
andr3ej: not better, just idiomatic junit

node/src/main/kotlin/net/corda/node/internal/Node.kt
andr3ej: done

PR #2079, opened 2017-11-17
Stabilise NodeStatePersistenceTests

rnicoll:


node/src/integration-test/kotlin/net/corda/test/node/NodeStatePersistenceTests.kt
```diff
@@ -40,26 +41,29 @@ class NodeStatePersistenceTests {
 
         val user = User("mark", "dadada", setOf(startFlow<SendMessageFlow>(), invokeRpc("vaultQuery")))
         val message = Message("Hello world!")
+        var stateAndRef: StateAndRef<MessageState>? = null
         driver(isDebug = true, startNodesInProcess = isQuasarAgentSpecified()) {
             val nodeName = {
                 val nodeHandle = startNode(rpcUsers = listOf(user)).getOrThrow()
                 val nodeName = nodeHandle.nodeInfo.chooseIdentity().name
                 nodeHandle.rpcClientToNode().start(user.username, user.password).use {
+                    // TODO: Wait until notaries have registered
```
shamsasari: Correct thing to is `defaultNotaryNode.getOrThrow()` as the first thing in the driver

node/src/integration-test/kotlin/net/corda/test/node/NodeStatePersistenceTests.kt
```diff
@@ -40,26 +41,29 @@ class NodeStatePersistenceTests {
...
-            nodeHandle.rpcClientToNode().start(user.username, user.password).use {
+            stateAndRef = nodeHandle.rpcClientToNode().start(user.username, user.password).use {
                 val page = it.proxy.vaultQuery(MessageState::class.java)
-                val stateAndRef = page.states.singleOrNull()
-                assertNotNull(stateAndRef)
-                val retrievedMessage = stateAndRef!!.state.data.message
-                assertEquals(message, retrievedMessage)
+                page.states.singleOrNull()
             }
+            nodeHandle.stop()
```
shamsasari: return `stateAndRef` at the end assign it outside to a val. That way you don't need to declare it as nullable

node/src/integration-test/kotlin/net/corda/test/node/NodeStatePersistenceTests.kt
```diff
@@ -142,7 +146,8 @@ class SendMessageFlow(private val message: Message) : FlowLogic<SignedTransactio
 
     @Suspendable
     override fun call(): SignedTransaction {
-        val notary = serviceHub.networkMapCache.notaryIdentities.first()
+        // We fall back to the expected notary identity, but shouldn't start the flow until notaries are registered
+        val notary = serviceHub.networkMapCache.notaryIdentities.firstOrNull() ?: DUMMY_NOTARY
```
shamsasari: I was thinking of changing this test. Best thing to do is pass in the notary `Party`. But my suggested fix above means this change isn't needed anyway.

rnicoll: I think that's a lot neater; I've changed it to that approach now

node/src/integration-test/kotlin/net/corda/test/node/NodeStatePersistenceTests.kt
```diff
@@ -40,26 +40,30 @@ class NodeStatePersistenceTests {
...
-            nodeHandle.rpcClientToNode().start(user.username, user.password).use {
+            val result = nodeHandle.rpcClientToNode().start(user.username, user.password).use {
                 val page = it.proxy.vaultQuery(MessageState::class.java)
-                val stateAndRef = page.states.singleOrNull()
-                assertNotNull(stateAndRef)
-                val retrievedMessage = stateAndRef!!.state.data.message
-                assertEquals(message, retrievedMessage)
+                page.states.singleOrNull()
             }
+            nodeHandle.stop()
```
shamsasari: This is probably not needed since the driver stops all started processes. And I recently made a change to make sure of that.

node/src/integration-test/kotlin/net/corda/test/node/NodeStatePersistenceTests.kt
```diff
@@ -40,26 +40,30 @@ class NodeStatePersistenceTests {
 
         val user = User("mark", "dadada", setOf(startFlow<SendMessageFlow>(), invokeRpc("vaultQuery")))
         val message = Message("Hello world!")
-        driver(isDebug = true, startNodesInProcess = isQuasarAgentSpecified()) {
+        val stateAndRef: StateAndRef<MessageState>? = driver(isDebug = true, startNodesInProcess = isQuasarAgentSpecified()) {
             val nodeName = {
                 val nodeHandle = startNode(rpcUsers = listOf(user)).getOrThrow()
                 val nodeName = nodeHandle.nodeInfo.chooseIdentity().name
+                // Ensure the notary node has finished starting up, before starting a flow that needs a notary
```
shamsasari: Whatever issue this test had, I'm not sure this was actually the cause. Remember Corda is designed to be able to send messages to other nodes if they're offline, and so even if the notary is not fully up it shouldn't matter. The notarisation request should be handled when the notary is ready. So I don't think we need to be waiting on the line below.

If waiting for the notary is required then it affects all the tests using the driver! But like I say I don't think that's the issue.

PR #2080, opened 2017-11-17
remove calls to exitnodes from NodeStartup

al-r3:
remove all the calls to exitProcess in NodeStartup.kt  and move them up to Corda.kt

### DIFF
--- a/node/src/main/kotlin/net/corda/node/Corda.kt
+++ b/node/src/main/kotlin/net/corda/node/Corda.kt
@@ -4,10 +4,11 @@
 package net.corda.node
 
 import net.corda.node.internal.NodeStartup
+import kotlin.system.exitProcess
 
 fun main(args: Array<String>) {
     // Pass the arguments to the Node factory. In the Enterprise edition, this line is modified to point to a subclass.
     // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
     // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
-    NodeStartup(args).run()
+    exitProcess(NodeStartup(args).run())
 }
--- a/node/src/main/kotlin/net/corda/node/Corda.kt
+++ b/node/src/main/kotlin/net/corda/node/Corda.kt
@@ -10,5 +10,5 @@ fun main(args: Array<String>) {
     // Pass the arguments to the Node factory. In the Enterprise edition, this line is modified to point to a subclass.
     // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
     // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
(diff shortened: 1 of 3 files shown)


shamsasari:
Is this now event needed since we're calling the registration helper class directly in in-process mode?

al-r3:
Seems a good thing to do in general, there were a few TODOs around about doing this. It also makes it possible to test some failures related to registration

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -62,12 +61,12 @@ class NetworkRegistrationHelper(private val config: NodeConfiguration, private v
 
             val certificates = try {
                 pollServerForCertificates(requestId)
-            } catch (e: CertificateRequestException) {
-                System.err.println(e.message)
+            } catch (certificateRequestException: CertificateRequestException) {
+                System.err.println(certificateRequestException.message)
```
rnicoll: I think we should be outputting consistently here; can you make the lines below output to `System.err` as well please?

al-r3: done for the error messages.
It is actually quite fun that 2 consecutive lines did something different

Clintonio:
The only comment I have is that if we're not planning to use any codes other than success and failure we could make it a boolean until we need it to do more.

al-r3:
All done

node/src/main/kotlin/net/corda/node/Corda.kt
```diff
@@ -4,10 +4,11 @@
...
 
 import net.corda.node.internal.NodeStartup
+import kotlin.system.exitProcess
 
 fun main(args: Array<String>) {
     // Pass the arguments to the Node factory. In the Enterprise edition, this line is modified to point to a subclass.
     // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
     // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
-    NodeStartup(args).run()
+    exitProcess(if (NodeStartup(args).run()) 0 else 1)
```
Clintonio: Something that I would find greatly useful would be to have it print out "Corda has exited successfully" on one branch and "Corda has exited with an error" on the other. While it contains no details it will be helpful in system tests.

al-r3: Why? The exit code gives you the exact same information.
The specific error should be logged and detailed by whatever raised and detected the error

Clintonio: Ah, I just realised exitProcess does a force kill on itself. My issue is that I've had hanging threads in CordaZone where we simply return from main.

PR #2081, opened 2017-11-17
Upgrade Kotlin and add a comment about RPC role security

mikehearn:


node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -235,8 +235,10 @@ class ArtemisMessagingServer(private val config: NodeConfiguration,
         securityRoles["$INTERNAL_PREFIX#"] = setOf(nodeInternalRole)  // Do not add any other roles here as it's only for the node
         securityRoles[P2P_QUEUE] = setOf(nodeInternalRole, restrictedRole(PEER_ROLE, send = true))
         securityRoles[RPCApi.RPC_SERVER_QUEUE_NAME] = setOf(nodeInternalRole, restrictedRole(RPC_ROLE, send = true))
-        // TODO remove the NODE_USER role once the webserver doesn't need it
+        // TODO: remove the NODE_USER role below once the webserver doesn't need it anymore.
         securityRoles["${RPCApi.RPC_CLIENT_QUEUE_NAME_PREFIX}.$NODE_USER.#"] = setOf(nodeInternalRole)
+        // Each RPC user must have its own role and its own queue. This prevents users accessing each other's queues
+        // and stealing RPC responses.
         for ((username) in userService.users) {
```
IgorNitto: Very reasonable to document this fact.

node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -235,8 +235,10 @@ class ArtemisMessagingServer(private val config: NodeConfiguration,
         securityRoles["$INTERNAL_PREFIX#"] = setOf(nodeInternalRole)  // Do not add any other roles here as it's only for the node
         securityRoles[P2P_QUEUE] = setOf(nodeInternalRole, restrictedRole(PEER_ROLE, send = true))
         securityRoles[RPCApi.RPC_SERVER_QUEUE_NAME] = setOf(nodeInternalRole, restrictedRole(RPC_ROLE, send = true))
-        // TODO remove the NODE_USER role once the webserver doesn't need it
+        // TODO: remove the NODE_USER role below once the webserver doesn't need it anymore.
```
shamsasari: This is interesting. I think we can now do this TODO because the webserver no longer uses the NODE_USER. It connects using RPC.

mikehearn: @IgorNitto want to take that as a small clean up task?

IgorNitto: Yes, of course.

PR #2082, opened 2017-11-17
remove cordacon code

al-r3:


ragmondo:
Another contribution bites the dust !!

LGTM :-(

PR #2086, opened 2017-11-20
Change default driver permissions such that NodeExplorer works.

vkolomeyko:
Also fix-up Kotlin doc.


vkolomeyko:
Sure - I will add the comment.

This change is needed due to RPC permissions enabled recently.
Of course it is possible to achieve the same effect by changing NodeExplorer node start-up code, but since this is all our testing/PoC stuff, I thought it would not be a good idea to duplicate the effort like:
`NodeMonitorModelTest#setup` does, but instead change default permissions such that they are adequate and allow Driver code to work the way it was working before RPC permissions were introduced.

PR #2087, opened 2017-11-20
CORDA-311 Shell via SSH server 

m4ksio:
SSH server can be enabled, so users can connect and manage the node the same way as via local shell (but with proper users/permissions support)

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
@@ -348,6 +348,20 @@ abstract class FlowLogic<out T> {
         }
     }
 
+    fun trackStepsTreeIndex(): DataFeed<Int, Int>? {
+        // TODO this is not threadsafe, needs an atomic get-step-and-subscribe
+        return progressTracker?.let {
+            DataFeed(it.stepsTreeIndex, it.stepsTreeIndexChanges)
+        }
+    }
+
+    fun trackStepsTree(): DataFeed<List<Pair<Int,String>>, List<Pair<Int,String>>>? {
+        // TODO this is not threadsafe, needs an atomic get-step-and-subscribe
+        return progressTracker?.let {
+            DataFeed(it.allStepsLabels, it.stepsTreeChanges)
+        }
+    }
+
     /**
      * Suspends the flow until the transaction with the specified ID is received, successfully verified and
      * sent to the vault for processing. Note that this call suspends until the transaction is considered
--- a/core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
@@ -55,7 +55,7 @@ data class FlowHandleImpl<A>(
 }
 
 @CordaSerializable
-data class FlowProgressHandleImpl<A>(
+data class FlowProgressHandleImpl<A> @JvmOverloads constructor(
         override val id: StateMachineRunId,
         override val returnValue: CordaFuture<A>,
         override val progress: Observable<String>,
--- a/core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
@@ -62,6 +62,11 @@ data class FlowProgressHandleImpl<A> @JvmOverloads constructor(
         override val stepsTreeIndexFeed: DataFeed<Int, Int>? = null,
         override val stepsTreeFeed: DataFeed<List<Pair<Int, String>>, List<Pair<Int, String>>>? = null) : FlowProgressHandle<A> {
 
+    // For API compatibility
+    fun copy(id: StateMachineRunId, returnValue: CordaFuture<A>, progress: Observable<String>): FlowProgressHandleImpl<A> {
+        return copy(id = id, returnValue = returnValue, progress = progress, stepsTreeFeed = null, stepsTreeIndexFeed = null)
+    }
+
     // Remember to add @Throws to FlowProgressHandle.close() if this throws an exception.
     override fun close() {
         try {
--- a/core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
@@ -31,6 +31,9 @@ interface FlowHandle<A> : AutoCloseable {
 interface FlowProgressHandle<A> : FlowHandle<A> {
     val progress: Observable<String>
 
+    val stepsTreeIndexFeed: DataFeed<Int, Int>?
+
+    val stepsTreeFeed: DataFeed<List<Pair<Int, String>>, List<Pair<Int, String>>>?
     /**
      * Use this function for flows whose returnValue and progress are not going to be used or tracked, so as to free up
      * server resources.
@@ -55,7 +58,9 @@ data class FlowHandleImpl<A>(
 data class FlowProgressHandleImpl<A>(
         override val id: StateMachineRunId,
         override val returnValue: CordaFuture<A>,
-        override val progress: Observable<String>) : FlowProgressHandle<A> {
+        override val progress: Observable<String>,
+        override val stepsTreeIndexFeed: DataFeed<Int, Int>? = null,
+        override val stepsTreeFeed: DataFeed<List<Pair<Int, String>>, List<Pair<Int, String>>>? = null) : FlowProgressHandle<A> {
 
     // Remember to add @Throws to FlowProgressHandle.close() if this throws an exception.
     override fun close() {
--- a/core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
+++ b/core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
@@ -6,9 +6,6 @@ import rx.Subscription
 import rx.subjects.PublishSubject
 import java.util.*
 
-// TODO: Expose the concept of errors.
-// TODO: It'd be helpful if this class was at least partly thread safe.
-
 /**
  * A progress tracker helps surface information about the progress of an operation to a user interface or API of some
  * kind. It lets you define a set of _steps_ that represent an operation. A step is represented by an object (typically
@@ -34,16 +31,16 @@ import java.util.*
 @CordaSerializable
 class ProgressTracker(vararg steps: Step) {
     @CordaSerializable
-    sealed class Change {
-        data class Position(val tracker: ProgressTracker, val newStep: Step) : Change() {
+    sealed class Change(val progressTracker: ProgressTracker) {
+        data class Position(val tracker: ProgressTracker,  val newStep: Step) : Change(tracker) {
             override fun toString() = newStep.label
         }
 
-        data class Rendering(val tracker: ProgressTracker, val ofStep: Step) : Change() {
+        data class Rendering(val tracker: ProgressTracker, val ofStep: Step) : Change(tracker) {
             override fun toString() = ofStep.label
         }
 
-        data class Structural(val tracker: ProgressTracker, val parent: Step) : Change() {
+        data class Structural(val tracker: ProgressTracker, val parent: Step) : Change(tracker) {
             override fun toString() = "Structural step change in child of ${parent.label}"
         }
     }
@@ -70,16 +67,22 @@ class ProgressTracker(vararg steps: Step) {
         override fun equals(other: Any?) = other is DONE
     }
 
-    /** The steps in this tracker, same as the steps passed to the constructor but with UNSTARTED and DONE inserted. */
-    val steps = arrayOf(UNSTARTED, *steps, DONE)
-
-    // This field won't be serialized.
-    private val _changes by transient { PublishSubject.create<Change>() }
-
     @CordaSerializable
     private data class Child(val tracker: ProgressTracker, @Transient val subscription: Subscription?)
 
     private val childProgressTrackers = mutableMapOf<Step, Child>()
+
+    /** The steps in this tracker, same as the steps passed to the constructor but with UNSTARTED and DONE inserted. */
+    val steps = arrayOf(UNSTARTED, *steps, DONE)
+
+    private var _allStepsCache: List<Pair<Int, Step>> = _allSteps()
+
+    // This field won't be serialized.
+    private val _changes by transient { PublishSubject.create<Change>() }
+    private val _stepsTreeChanges by transient { PublishSubject.create<List<Pair<Int, String>>>() }
+    private val _stepsTreeIndexChanges by transient { PublishSubject.create<Int>() }
+
+
 
     init {
         steps.forEach {
@@ -92,7 +95,15 @@ class ProgressTracker(vararg steps: Step) {
 
     /** The zero-based index of the current step in the [steps] array (i.e. with UNSTARTED and DONE) */
     var stepIndex: Int = 0
-        private set
+        private set(value) {
+            field = value
+        }
+
+    var stepsTreeIndex: Int = -1
+        private set(value) {
+            field = value
+            _stepsTreeIndexChanges.onNext(value)
+        }
 
     /**
      * Reading returns the value of steps[stepIndex], writing moves the position of the current tracker. Once moved to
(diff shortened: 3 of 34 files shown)


core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -348,6 +348,20 @@ abstract class FlowLogic<out T> {
         }
     }
 
+    fun trackStepsTreeIndex(): DataFeed<Int, Int>? {
+        // TODO this is not threadsafe, needs an atomic get-step-and-subscribe
```
sollecitom: How about these 2 TODOs?

core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
```diff
@@ -52,10 +55,17 @@ data class FlowHandleImpl<A>(
 }
 
 @CordaSerializable
-data class FlowProgressHandleImpl<A>(
+data class FlowProgressHandleImpl<A> @JvmOverloads constructor(
         override val id: StateMachineRunId,
         override val returnValue: CordaFuture<A>,
-        override val progress: Observable<String>) : FlowProgressHandle<A> {
+        override val progress: Observable<String>,
+        override val stepsTreeIndexFeed: DataFeed<Int, Int>? = null,
```
sollecitom: Do we need default values here? Seems a bit down the stack to expose default values.

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -34,16 +31,16 @@ import java.util.*
 @CordaSerializable
 class ProgressTracker(vararg steps: Step) {
     @CordaSerializable
-    sealed class Change {
-        data class Position(val tracker: ProgressTracker, val newStep: Step) : Change() {
+    sealed class Change(val progressTracker: ProgressTracker) {
+        data class Position(val tracker: ProgressTracker,  val newStep: Step) : Change(tracker) {
```
sollecitom: Change `val tracker: ProgressTracker` to `tracker: ProgressTracker` for all sub-types. No need to have a duplicated variable given that the parent now has it.

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -118,22 +129,39 @@ class ProgressTracker(vararg steps: Step) {
...
+
+            if (currentStep == DONE) {
+                _changes.onCompleted()
+                _stepsTreeIndexChanges.onCompleted()
+                _stepsTreeChanges.onCompleted()
+            }
         }
 
     /** Returns the current step, descending into children to find the deepest step we are up to. */
     val currentStepRecursive: Step
```
sollecitom: Any chance we flag this recursive lookup with `tailrec`? Not a massive problem, but it might help performance.

Not entirely sure it's possible from GitHub.

docs/source/changelog.rst
```diff
@@ -6,9 +6,13 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
+
 * ``AttachmentStorage`` now allows providing metadata on attachments upload - username and filename, currently as plain
   strings. Those can be then used for querying, utilizing ``queryAttachments`` method of the same interface.
 
+* ``SSH Server`` - Instead the node can now expose SSH server with proper authorization
```
sollecitom: I'm not sure "Instead the node can now expose" works in English. How about something along the lines of:

* ``SSH Server`` - Nodes can now choose to expose a shell that requires users to login through SSH and enforces permissions checking as part of RPC method invocation.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -96,15 +96,15 @@ class Node(private val project: Project) : CordformNode() {
...
+     * Enables SSH access on given port
      *
-     * @param sshdPort The SSHD port.
+     * @param sshdPort The port for SSH server to listen on
      */
-    fun sshdPort(sshdPort: Int) {
-        config = config.withValue("sshdAddress",
-                ConfigValueFactory.fromAnyRef("$DEFAULT_HOST:$sshdPort"))
+    fun sshdPort(sshdPort: Int?) {
+        config = config.withValue("sshd.port", ConfigValueFactory.fromAnyRef(sshdPort))
```
sollecitom: What happens if `sshdPort` is `null`? Does it get unset?

node/build.gradle
```diff
@@ -157,6 +163,9 @@ dependencies {
     integrationTestCompile "junit:junit:$junit_version"
     integrationTestCompile "org.assertj:assertj-core:${assertj_version}"
 
+    // Jsh: Testing SSH server
+    integrationTestCompile group: 'com.jcraft', name: 'jsch', version: '0.1.54'
```
sollecitom: Shall we put the version in the `buildscript` section?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -39,6 +39,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     // TODO Move into DevModeOptions
     val useTestClock: Boolean get() = false
     val detectPublicIp: Boolean get() = true
+    val sshd: SSHDConfiguration?
```
sollecitom: Sounds like a good time to move these into `DevModeOptions`. At least can you please put `sshd` in there?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -109,7 +110,9 @@ data class NodeConfigurationImpl(
         override val detectPublicIp: Boolean = true,
         override val activeMQServer: ActiveMqServerConfiguration,
         // TODO See TODO above. Rename this to nodeInfoPollingFrequency and make it of type Duration
-        override val additionalNodeInfoPollingFrequencyMsec: Long = 5.seconds.toMillis()
+        override val additionalNodeInfoPollingFrequencyMsec: Long = 5.seconds.toMillis(),
+        override val sshd: SSHDConfiguration? = null
```
sollecitom: I'd personally have `SSHDConfiguration` as not nullable, and have a `port` field inside it with type `Int?`. Not sure if it changes that much though.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -144,3 +147,5 @@ data class CertChainPolicyConfig(val role: String, private val policy: CertChain
             }
         }
 }
+
+data class SSHDConfiguration(val port: Int)
```
sollecitom: See above.

node/src/main/kotlin/net/corda/node/services/messaging/RpcAuthContext.kt
```diff
@@ -23,6 +23,7 @@ data class RpcPermissions(private val values: Set<String> = emptySet()) {
 
     companion object {
         val NONE = RpcPermissions()
+        val ALL = RpcPermissions(setOf("ALL"))
```
sollecitom: Nice

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -78,58 +85,57 @@ object InteractiveShell {
     @VisibleForTesting
     internal lateinit var database: CordaPersistence
 
+    private lateinit var rpcOps:CordaRPCOps
+    private lateinit var userService:RPCUserService
+    private lateinit var identityService:IdentityService
+
+    private var shell:Shell? = null
```
sollecitom: Whitespace. Same for variables above.

node/src/main/kotlin/net/corda/node/shell/RPCOpsWithContext.kt
```diff
@@ -0,0 +1,210 @@
...
+import net.corda.node.services.messaging.RpcAuthContext
+import net.corda.node.services.messaging.RpcPermissions
+import rx.Observable
+import java.io.InputStream
+import java.security.PublicKey
+import java.time.Instant
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Future
+
+class RPCOpsWithContext(val cordaRPCOps: CordaRPCOps, val invocationContext:InvocationContext, val rpcPermissions: RpcPermissions) : CordaRPCOps {
```
sollecitom: Why do we need this type? If the Shell goes through RPC, setting the `CURRENT_RPC_CONTEXT` when the user logs in is enough. What am I missing?

core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
m4ksio: API compatibility

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
m4ksio: SSHD is not a dev mode option

node/src/main/kotlin/net/corda/node/shell/RPCOpsWithContext.kt
m4ksio: Shell runs in a different threads. AFAIK SSHD does not allocate one thread per client, moreover CRaSH executes commands in yet another thread-pool.

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
m4ksio: They are copied from original function - I assume it's not trivial to do this, so TODO is also here to remind to change those too, once somebody does the original

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
m4ksio: Data classes cannot have non-property parameters.

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
m4ksio: That would require either changing this to method (API compatibility problem) or using delegate - don't think it's worth it

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
sollecitom: OK then, but what's the point of moving that field to the super-type then?

m4ksio: So it can be accessed in a places we use supertype

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
mikehearn: This is new public API so it needs KDocs. Also, for fixing the thread safety issue, there's no time like the present!

core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
```diff
@@ -31,6 +31,9 @@ interface FlowHandle<A> : AutoCloseable {
 interface FlowProgressHandle<A> : FlowHandle<A> {
     val progress: Observable<String>
 
+    val stepsTreeIndexFeed: DataFeed<Int, Int>?
```
mikehearn: Likewise

core/src/main/kotlin/net/corda/core/messaging/FlowHandle.kt
```diff
@@ -31,6 +31,9 @@ interface FlowHandle<A> : AutoCloseable {
...
+
+    val stepsTreeFeed: DataFeed<List<Pair<Int, String>>, List<Pair<Int, String>>>?
```
mikehearn: Likewise. Please enable the missing KDoc inspection in your IntelliJ to catch these.

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -92,7 +95,15 @@ class ProgressTracker(vararg steps: Step) {
 
     /** The zero-based index of the current step in the [steps] array (i.e. with UNSTARTED and DONE) */
     var stepIndex: Int = 0
-        private set
+        private set(value) {
+            field = value
+        }
+
+    var stepsTreeIndex: Int = -1
```
mikehearn: needs kdoc

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
mikehearn: It's not tail recursive by the definition needed by the compiler.

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -177,11 +218,15 @@ class ProgressTracker(vararg steps: Step) {
...
+    private fun _allStepsLabels(level: Int = 0): List<Pair<Int, String>> = _allSteps(level).map { Pair(it.first, it.second.label) }
+
     /**
      * A list of all steps in this ProgressTracker and the children, with the indent level provided starting at zero.
      * Note that UNSTARTED is never counted, and DONE is only counted at the calling level.
      */
-    val allSteps: List<Pair<Int, Step>> get() = _allSteps()
+    val allSteps: List<Pair<Int, Step>> get() = _allStepsCache
+
+    val allStepsLabels: List<Pair<Int, String>> get() = _allStepsLabels()
```
mikehearn: kdoc

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -200,8 +245,15 @@ class ProgressTracker(vararg steps: Step) {
      */
     val changes: Observable<Change> get() = _changes
 
+    val stepsTreeChanges: Observable<List<Pair<Int,String>>> get() = _stepsTreeChanges
```
mikehearn: kdoc

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -200,8 +245,15 @@ class ProgressTracker(vararg steps: Step) {
...
+
+    val stepsTreeIndexChanges: Observable<Int> get() = _stepsTreeIndexChanges
```
mikehearn: kdoc

core/src/test/kotlin/net/corda/core/utilities/ProgressTrackerTest.kt
```diff
@@ -81,6 +92,118 @@ class ProgressTrackerTest {
...
+        //assert no structure changes and proper steps propagation
+        assertThat(stepsIndexNotifications).containsExactlyElementsOf(listOf(0, 1, 3, 5))
+        assertThat(stepsTreeNotification).isEmpty()
+    }
+
+    @Test
+    fun `structure changes are pushed down when progress trackers are added`() {
+        pt.setChildProgressTracker(SimpleSteps.TWO, pt2)
+
+        //capture notifications
```
mikehearn: Code style guideline requires a space after // and to use full sentences with capitals and full stops:

https://docs.corda.net/head/codestyle.html

docs/source/shell.rst
```diff
@@ -18,11 +18,47 @@ Some of its features include:
...
+
+.. code:: bash
+
+    sshd {
+        port = 2222
+    }
+
+Authentication and authorization
+--------------------------------
+SSH require user to login first - using the same users as RPC system. In fact, shell serves as a proxy to RPC and communicates
```
mikehearn: Typo: SSH access *requires* *users* (plural) to log in first. In fact *the* shell communicates with *the* node.

docs/source/shell.rst
```diff
@@ -18,11 +18,47 @@ Some of its features include:
...
+with node using RPC calls. This also means that RPC permissions are enforced. No permissions are required to allow the connection
+and login in.
```
mikehearn: typo: login in

docs/source/shell.rst
```diff
@@ -18,11 +18,47 @@ Some of its features include:
...
+Watching flows (``flow watch``) requires ``InvokeRpc.stateMachinesFeed`` while starting flows requires
```
mikehearn: I mentioned this to @sollecitom but this seems rather redundant. Being granted a permission to start a flow should grant you access to the RPC to start flows as well, implicitly. This is just awkward.

docs/source/shell.rst
```diff
@@ -18,11 +18,47 @@ Some of its features include:
...
+--------
+
+The host key is loaded from ``sshkey/hostkey.pem`` file. If the file does not exist, it will be generated randomly, however
+in the development mode seed may be tuned to give the same results on the same computer - in order to avoid host checking
+errors.
+
+Connecting
+----------
+
+Linux and MacOS computers usually come with SSH client preinstalled. On Windows it usually require extra download.
```
mikehearn: usually *requires an* extra download

node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -77,59 +84,55 @@ object InteractiveShell {
...
 
         ExternalResolver.INSTANCE.addCommand("run", "Runs a method from the CordaRPCOps interface on the node.", RunShellCommand::class.java)
         ExternalResolver.INSTANCE.addCommand("flow", "Commands to work with flows. Flows are how you can change the ledger.", FlowShellCommand::class.java)
         ExternalResolver.INSTANCE.addCommand("start", "An alias for 'flow start'", StartShellCommand::class.java)
-        val shell = ShellLifecycle(dir).start(config)
+        shell = ShellLifecycle(dir).start(config)
 
-        if (runSSH) {
-            // printBasicNodeInfo("SSH server listening on address", node.configuration.sshdAddress.toString())
+        if (runSshDeamon) {
```
mikehearn: typo: daemon (it's a painful word i know, i always hated it)

node/src/main/kotlin/net/corda/node/shell/RPCOpsWithContext.kt
mikehearn: Still, the boilerplate is kind of unfortunate. A java.lang.reflect.Proxy would work and avoid the need to always update this file.

node/src/main/kotlin/net/corda/node/utilities/ANSIProgressRenderer.kt
```diff
@@ -152,46 +127,137 @@ object ANSIProgressRenderer {
...
-                    lines += child.renderLevel(ansi, indent + 1, error)
             }
             return lines
         }
     }
+
+
+}
+
+class CRaSHNSIProgressRenderer(val renderPrintWriter:RenderPrintWriter) : ANSIProgressRenderer() {
```
mikehearn: typo: NSI instead of ANSI

node/src/main/kotlin/net/corda/node/utilities/ANSIProgressRenderer.kt
```diff
@@ -152,46 +127,137 @@ object ANSIProgressRenderer {
...
+        renderPrintWriter.println(line)
+    }
+
+    override fun printAnsi(ansi: Ansi) {
+        renderPrintWriter.print(ansi)
+        renderPrintWriter.flush()
+    }
+
+    override fun setup() {
+        //we assume SSH always use ansi
```
mikehearn: comment style

PR #2094, opened 2017-11-20
Corda runs powered by Spring Boot.

sollecitom:
**DO NOT MERGE!**

Refactored `Corda.kt` to leverage Spring Boot as a Dependency Injection container.

How to run it:
- Run/debug `Corda.kt` from Intellij with program arguments `--config-file=node/src/main/resources/testnode.conf` and Quasar agent.

- Run/debug `TestCorda.kt` to see how Spring Boot lets you override implementations in tests (from Intellij with program arguments `--config-file=node/src/main/resources/testnode.conf` and Quasar agent).

Notes:

- The logging is slightly different than the original format. I believe it can be fixed quite easily but it was beyond the scope of this spike.
- It took me ~15m to make it work.
-  Here's a PR showcasing Guice instead (https://github.com/corda/corda/pull/2131).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ buildscript {
     ext.jolokia_version = '2.0.0-M3'
     ext.assertj_version = '3.8.0'
     ext.slf4j_version = '1.7.25'
-    ext.log4j_version = '2.9.1'
+    ext.log4j_version = '2.7'
     ext.bouncycastle_version = constants.getProperty("bouncycastleVersion")
     ext.guava_version = constants.getProperty("guavaVersion")
     ext.okhttp_version = '3.5.0'
@@ -48,6 +48,11 @@ buildscript {
     ext.dependency_checker_version = '3.0.1'
     ext.commons_collections_version = '4.1'
     ext.beanutils_version = '1.9.3'
+
+    ext.spring_boot_version = '1.5.8.RELEASE'
+    ext.spring_version = '5.0.0.RELEASE'
+    ext.jsr_250_version = '1.0'
+    ext.javax_inject_version = '1'
 
     // Update 121 is required for ObjectInputFilter and at time of writing 131 was latest:
     ext.java8_minUpdateVersion = '131'
--- a/node/build.gradle
+++ b/node/build.gradle
@@ -109,6 +109,7 @@ dependencies {
     compile "org.springframework.boot:spring-boot:$spring_boot_version"
     compile "org.springframework.boot:spring-boot-autoconfigure:$spring_boot_version"
     runtime "org.springframework.boot:spring-boot-starter-log4j2:$spring_boot_version"
+    testCompile "org.hibernate:hibernate-validator:5.2.4.Final"
 
     // Unit testing helpers.
     testCompile "junit:junit:$junit_version"
--- a/node/build.gradle
+++ b/node/build.gradle
@@ -103,6 +103,13 @@ dependencies {
     // TypeSafe Config: for simple and human friendly config files.
     compile "com.typesafe:config:$typesafe_config_version"
 
+    compile "javax.inject:javax.inject:$javax_inject_version"
+    compile "javax.annotation:jsr250-api:$jsr_250_version"
+
+    compile "org.springframework.boot:spring-boot:$spring_boot_version"
+    compile "org.springframework.boot:spring-boot-autoconfigure:$spring_boot_version"
+    runtime "org.springframework.boot:spring-boot-starter-log4j2:$spring_boot_version"
+
     // Unit testing helpers.
     testCompile "junit:junit:$junit_version"
     testCompile "org.assertj:assertj-core:${assertj_version}"
--- a/node/src/main/kotlin/net/corda/node/Corda.kt
+++ b/node/src/main/kotlin/net/corda/node/Corda.kt
@@ -3,11 +3,21 @@
 
 package net.corda.node
 
-import net.corda.node.internal.NodeStartup
+import org.springframework.boot.Banner
+import org.springframework.boot.SpringApplication
+import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.context.annotation.ComponentScan
+
+@SpringBootApplication
+@ComponentScan("net.corda")
+open class NodeStartupApp
 
 fun main(args: Array<String>) {
-    // Pass the arguments to the Node factory. In the Enterprise edition, this line is modified to point to a subclass.
-    // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
-    // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
-    NodeStartup(args).run()
+
+    val application = SpringApplication(NodeStartupApp::class.java)
+
+    application.setBannerMode(Banner.Mode.OFF)
+    application.isWebEnvironment = false
+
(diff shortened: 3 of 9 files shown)


sollecitom:
Guys please have a look so we can decide whether we want to go forward. This is just an idea, this PR will never be merged.

exFalso:
inb4 storm of comments. I'd like to quickly mention another use case for DI: enterprise feature injection. Our current "top-down" enterprise feature specialisation will not and does not scale to the whole codebase, especially if the to-be-injected feature lives in some submodule of the system.
I think DI *is* the right solution for open-source vs enterprise feature switching, this will allow simple interface-based specification of features, an introduction of an enterprise feature will not result in a log(n) change in the code tree and in general will minimise long term hair loss.
I view cordapp injection and e.g. node initialisation dependency graph resolution as an additional potential use case, to me enabling easy enterprise dev is a good enough reason to introduce DI in and of itself.

sollecitom:
@exFalso plugin-based resolution at runtime is exactly the use case we have this Sprint. It would allow a clean switch at runtime without polluting the classpath with multiple implementations that are never required at the same time.

andr3ej:
further to @exFalso's comment, currently it's difficult to replace a service for testing and it always has to be done in a bespoke way, see how VaultSoftLockManagerTest hacks the vault service. a DI solution should provide a general mechanism for swapping out services in tests.

sollecitom:
@andr3ej yeah, I agree. Fortunately Spring Boot in particular makes substituting a service in a test quite a simple task. All that's needed is to create a Kotlin Spring `@Configuration` annotated class that imports the production one, excludes the production scoped service, and provide a test scoped one.

With a bit of wrapping, this can be done with just a few lines of code.

sollecitom:
@andr3ej Run/debug `TestCorda.kt` to see how Spring Boot lets you override implementations in tests (from Intellij with program arguments `--config-file=node/src/main/resources/testnode.conf` and Quasar agent).

andr3ej:
since apparently this isn't going away, some things i hate about spring:
* does too much, none of its parts as good as dedicated libs, like buying only microsoft
* doesn't play well with other libs in the di space as hackability so poor
* spring way or highway, vendor lock-in
* can't integrate with code without polluting it with its annotations
* astonishingly crude config system with call site defaulting and key composition, you never know when a key is safe to delete
* terrible error messages that point you to where your autowiring problem isn't
* autowiring arrays does not work
* ignores visibility modifiers
* not practical to use context in an ad-hoc way as it's so slow
* i have to be a spring expert now which i find quite distasteful
* spring absolutely loves aop
* huge list of bean lifecycle steps to remember
* "bean"
* same team that brought us 50k lines of undebuggable xml
* too magic e.g. how the devil does that test override example work
* facilitates poor software engineering e.g. circular deps, setters

chrisr3:
So speaking as a Spring Boot N00b, I am assuming here that this example works because there's only one `ExampleService` bean to be injected anywhere. How would it handle ambiguous bean resolution? Also, my experience with CDI/WELD was that beans could only be injected into other beans, which meant that suddenly _everything_ had to become a bean. (WELD never solved the problem of which order to destroy beans in either.)
For reference, having using CDI/WELD considerably in the past, I would _never_ have recommended using it for Corda...

sollecitom:
@andr3ej 

I do not have strong feelings about Spring Boot, I'm more than happy to use another DI container, provided someone creates a similar branch with the same features. I have a feeling some of your concerns might be related to previous versions of the Spring / Spring Boot framework, but here are my thoughts about your questions:

- does too much, none of its parts as good as dedicated libs, like buying only microsoft : it's an extremely modular project actually.

- doesn't play well with other libs in the di space as hackability so poor : not sure what you mean, would you use multiple di libs together? An example?

- spring way or highway, vendor lock-in : they actually made quite a bit of effort for this not to be the case, look at allowing usage of non-Spring annotations.

- can't integrate with code without polluting it with its annotations : as above, I'm not using any Spring annotation in this example. Also, other configuration means do exist, based on manual binding or file configuration.

- astonishingly crude config system with call site defaulting and key composition, you never know when a key is safe to delete : what config system? Do you refer to injecting values? Not a big fan either, but we don't have to use it.

- terrible error messages that point you to where your autowiring problem isn't : if you try to remove the annotation from `LoggingExampleService` and run the application it will tell exactly where the problem is.

- autowiring arrays does not work : not sure about arrays, but `LIst` and `Set` do work.

- ignores visibility modifiers : not true anymore, works with internal but not private.

- not practical to use context in an ad-hoc way as it's so slow : not sure what you mean.

- i have to be a spring expert now which i find quite distasteful : I disagree, production usage I saw wasn't much different from showcased simple wiring.

- spring absolutely loves aop : not sure and don't see what the problem would be.

- huge list of bean lifecycle steps to remember : not sure what you mean.

- "bean" : :)

- same team that brought us 50k lines of undebuggable xml : frameworks do evolve, as do the teams behind them.

- too magic e.g. how the devil does that test override example work : it works by Spring knowing what test scope is, but I don't get what's the problem with it.

- facilitates poor software engineering e.g. circular deps, setters : true, like most DI containers, although we can enforce constructor dependencies only, like the example.

sollecitom:
@chrisr3 about your comments:

- Naming disambiguation: in case more than one candidate is found when the dependencies tree is constructed, Spring terminates the application with an error, complaining about an ambiguous setup. Solving it requires adding a qualifier at the injection point (can be based on the name or other factors). From previous experiences this pretty much never happens.

- About everything needing to be managed by the DI container, well, yes, but it is up to us to decide what gets injected and what doesn't. Usually, implementations that are defined at runtime get managed by the DI container at high level, and injected into types at high level, meaning that the number of them tends to stay contained.

chrisr3:
@sollecitom Looking closer, it seems that `@Named` is in fact a simple CDI qualifier! However, I'm suspecting that Spring Boot does not also support creating custom additional Qualifier annotations with optional parameters for passing metadata - correct? (With CDI, your Qualifier's parameters could be either "binding" or "non-binding", where "non-binding" values were ignored by the container's bean resolution logic. Conversely, `@Named`'s `name` parameter here is apparently a "binding" one.)

Does Spring Boot allow for the injection of things which _aren't_ beans, i.e. POJOs, generated values etc?

al-r3:
@chrisr3 @sollecitom 
In my experience using other DI in case of conflicts (which were quite common in my workplace)
One creates an annotation annotated with: @Qualifier and that becomes "part of the type" you're injecting.

chrisr3:
@al-r3 For my previous employer, I wrote a CDI "configuration library" JAR that allowed system properties and other configuration values to be injected into beans as simple POJOs. This library used the Full Monty of CDI, including custom `@Qualifier`s, beans with `@Produces` / `@Disposes` methods, "eager" beans created by observing events and `@Alternative` producers ranked by `@Priority`.

And this experience has taught me _not_ to recommend using CDI/WELD in Corda ;-).

andr3ej:
the little bit of spring in corda is already causing problems. can someone explain to me how IRSDemoTest is supposed to find InterestRatesSwapDemoAPI when running from intellij? an example of spring magic wasting development time.

sollecitom:
@andr3ej let's not confuse a specific problem with our own usage of a technology in a particular demo with the merits of the technology themselves.

andr3ej:
it's a big minus if it makes it this easy to screw up

PR #2095, opened 2017-11-20
Mutual TLS auth - mixed keys

kchalkias:
Various unit tests for EC NIST P-256 - RSA mixed key mutual TLS authentication.
RSA functionality will be required for nodes running on cloud (e.g. Azure), where HSMs provide support for RSA keys only. I am also checking for mixed CA keys required during cert path validation.
Also changed the `Crypto.kt` RSA OIDs due to PKCS1 cert validation compatibility purposes.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
@@ -63,21 +63,21 @@ import javax.crypto.spec.SecretKeySpec
  */
 object Crypto {
     /**
-     * RSA_SHA256 signature scheme using SHA256 as hash algorithm and MGF1 (with SHA256) as mask generation function.
+     * RSA_SHA256 signature scheme using SHA256 as hash algorithm.
      * Note: Recommended key size >= 3072 bits.
      */
     @JvmField
     val RSA_SHA256 = SignatureScheme(
             1,
             "RSA_SHA256",
-            AlgorithmIdentifier(PKCSObjectIdentifiers.rsaEncryption, null),
-            emptyList(),
+            AlgorithmIdentifier(PKCSObjectIdentifiers.sha256WithRSAEncryption, null),
+            listOf(AlgorithmIdentifier(PKCSObjectIdentifiers.rsaEncryption, null)),
             BouncyCastleProvider.PROVIDER_NAME,
             "RSA",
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
```diff
@@ -63,21 +63,21 @@ import javax.crypto.spec.SecretKeySpec
...
             1,
             "RSA_SHA256",
-            AlgorithmIdentifier(PKCSObjectIdentifiers.rsaEncryption, null),
-            emptyList(),
+            AlgorithmIdentifier(PKCSObjectIdentifiers.sha256WithRSAEncryption, null),
+            listOf(AlgorithmIdentifier(PKCSObjectIdentifiers.rsaEncryption, null)),
             BouncyCastleProvider.PROVIDER_NAME,
             "RSA",
-            "SHA256WITHRSAANDMGF1",
+            "SHA256WITHRSAEncryption",
```
mikehearn: Is this backwards compatible?

kchalkias: Obviously it is api-compatible, I tried various combinations to make it work both with TLS and Corda, to reply on your comment - backwards compatibility:
key generation YES
verifying old signatures NO
TLS YES (as it is always a new session key agreement)

So, we are only interested in Corda DLT keys (not TLS). However, because each key is created via `KeyManagementService`, there is (was) no support to generate keys of specific types (other than EdDSA).

mikehearn: OK, that's fine, I was worried that changing the algorithm here would just break checking of pre-existing digital signatures for example.

kchalkias: @mikehearn Now that I see it again, what about running the unit tests without using Corda's default `Crypto.kt` functionality for RSA key gen and signing, so I won't change `SHA256WITHRSAANDMGF1` and be 100% backwards wire-compatible as well.
I have to re-implement some methods for the unit tests, but then TLS and Corda stay completely independent. What are your thoughts?

kchalkias:
It turns out that we need cross TLS and CORDA_ID support, mainly due to the shared cert-path validation logic (both keys have the same intermediate and root CAs). Thus, I am merging this version.

PR #2099, opened 2017-11-21
Fixed some permission issues with demos.

sollecitom:


samples/bank-of-corda-demo/build.gradle
```diff
@@ -71,7 +71,8 @@ task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
                                  "StartFlow.net.corda.finance.flows.CashConfigDataFlow",
                                  "StartFlow.net.corda.finance.flows.CashExitFlow",
                                  "StartFlow.net.corda.finance.flows.CashIssueAndPaymentFlow",
-                                 "StartFlow.net.corda.finance.flows.CashConfigDataFlow"]]
+                                 "StartFlow.net.corda.finance.flows.CashConfigDataFlow",
+                                 "ALL"]]
```
mikehearn: If you add ALL then the others become redundant, right? You could just strip it down to ALL on its own, in this case.

That said, it would be more secure to specify the permissions individually, and better for teaching too. How many permissions are missing? Perhaps we can have a permission mode ALL-LOG where permission checks are disabled for that user but we get a log.warning() that includes the permission that was needed. So we can run the demos, look at the logs, and see all the permissions that we need to add (assuming our exploration of the app was sufficiently good).

samples/attachment-demo/build.gradle
```diff
@@ -35,7 +35,10 @@ dependencies {
 }
 
 task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
-    ext.rpcUsers = [['username': "demo", 'password': "demo", 'permissions': ["StartFlow.net.corda.attachmentdemo.AttachmentDemoFlow"]]]
+    ext.rpcUsers = [['username': "demo", 'password': "demo", 'permissions': ["StartFlow.net.corda.attachmentdemo.AttachmentDemoFlow",
```
sollecitom: Checked this locally, it looks well formatted and there's no whitespace.

samples/bank-of-corda-demo/build.gradle
sollecitom: I added the minimum required set, rather than switching to "ALL", good point.

sollecitom: We could provide an "ALL-LOG" permission, as a separate story.

sollecitom:
Hey @mikehearn , the permissions we introduced for MVP1 are fine-grained, because any assumptions we make on what clients might want to decide for their users will limit their options forever.

I see the point of having coarser-grained permissions, and we have roles flagged for MVP2, so I'd say we are definitely going to do something about this, but I would personally not remove required permissions from any RPC method.

Typical usage sees a Corda node called through RPC from a customers' back-end application, meaning "ALL" is pretty safe anyway, in case they don't want to bother adding one permission for each RPC method they call, because what the application does is fixed at code level.

mikehearn:
OK, fair enough. If a notion of roles is coming (perhaps with a default role that grants some safe subset of permissions) then I'm a happy camper.

PR #2107, opened 2017-11-22
docs for additional-node-infos

al-r3:
Add documentation about the additional-node-infos directory

### DIFF
--- a/docs/source/network-map.rst
+++ b/docs/source/network-map.rst
@@ -1,8 +1,8 @@
 Network Map
 ===========
 
-The network map stores a collection of ``NodeInfo`` objects, each representing another node with with one can interact.
-There 2 sources from which a Corda node can retrieve ``NodeInfo`` objects:
+The network map stores a collection of ``NodeInfo`` objects, each representing another node with which the node can interact.
+There two sources from which a Corda node can retrieve ``NodeInfo`` objects:
 
 1. the REST protocol with the network map service, which also provides a publishing API,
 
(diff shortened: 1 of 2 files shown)


docs/source/network-map.rst
```diff
@@ -35,4 +43,15 @@ Network Map service REST API:
...
+
+The ``additional-node-infos`` directory
+---------------------------------------
+Each Corda node reads, and continuously polls, the files contained in a directory named ``additional-node-infos`` inside the node base directory.
+
+Nodes expect to find a serialized ``SignedData<NodeInfo>`` object, the same object which is sent to network map server.
+
+Whenever a node starts it writes on disk a file containing its own ``NodeInfo``, this file is called ``nodeInfo-XXX`` where ``XXX`` is a long string.
+
+Hence if an operators wants node A to see node B they can pick A's ``NodeInfo`` file from A base directory and drop it into B's ``additional-node-infos`` directory.
```
gendal: surely the other way round?

al-r3: yup, nice catch.
It is somewhat likely they'll have to do that both ways.

gendal: oh - also just spotted: operator, not operators.

docs/source/network-map.rst
```diff
@@ -1,6 +1,14 @@
 Network Map
 ===========
 
+The network map stores a collection of ``NodeInfo`` objects, each representing another node with with one can interact.
```
shamsasari: typo: double with

shamsasari: "the node" instead of "one"

docs/source/network-map.rst
```diff
@@ -1,6 +1,14 @@
...
+There 2 sources from which a Corda node can retrieve ``NodeInfo`` objects:
```
shamsasari: "two"

PR #2108, opened 2017-11-22
Add X500 name constraints for non-organisation attributes

rnicoll:


tools/demobench/src/main/kotlin/net/corda/demobench/views/NodeTabView.kt
```diff
@@ -211,7 +211,7 @@ class NodeTabView : Fragment() {
                 error("Node with this name already exists")
             } else {
                 try {
-                    LegalNameValidator.validateLegalName(LegalNameValidator.normaliseLegalName(it))
+                    LegalNameValidator.validateOrganisation(LegalNameValidator.normaliseLegalName(it))
```
patrickkuo: let's change "normaliseLegalName" as well? Maybe just `normalise`

rnicoll:
Ready for re-review @patrickkuo

core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
```diff
@@ -45,7 +45,7 @@ data class CordaX500Name(val commonName: String?,
 
     init {
         // Legal name checks.
-        LegalNameValidator.validateLegalName(organisation)
+        LegalNameValidator.validateOrganization(organisation)
```
patrickkuo: LegalNameValidator.validateOrganization(**_organisation_**)

rnicoll: Yeah... that's my screwup... is it worth adding a deprecated `organisation` value and changing to US spelling (which is how it's spelt in the standard)?

rnicoll: Fixing the name in `CordaX500Name` is... lets say rather invasive. Going to skip for now and tackle later on.

PR #2111, opened 2017-11-22
Removes outdated devmode cert info, links to existing docs to avoid duplication.

joeldudleyr3:


docs/source/setting-up-a-corda-network.rst
```diff
@@ -6,29 +6,26 @@ Creating a Corda network
 A Corda network consists of a number of machines running nodes, including a single node operating as the network map
```
rnicoll: This changes to "A Corda network consists of a number of machines running nodes. These nodes communicate using persistent protocols in order to create and validate transactions." in the original PR, but not here? https://github.com/corda/corda/pull/2105/files

PR #2078, opened 2017-11-17
Integration test for http network map service

patrickkuo:
* make node info file copying optional by setting "compatabilityZoneURL" in driver
* integration test for node using http network map with driver
* some bug fixes

PR #2117, opened 2017-11-23
Don't make nodes crash if doorman returns a 500 on /my-hostname

al-r3:
Don't make nodes crash if doorman returns an error on /network-map URL


patrickkuo:
duplicate https://github.com/corda/corda/pull/2078/

PR #2118, opened 2017-11-23
CORDA-654 Pass key constants into VaultFiller

andr3ej:
first commit reviewed in #2141


testing/test-utils/src/main/kotlin/net/corda/testing/contracts/VaultFiller.kt
```diff
@@ -8,28 +8,38 @@ import net.corda.core.crypto.SignatureMetadata
...
-class VaultFiller(private val services: ServiceHub) {
+/**
+ * The service hub should provide at least a key management service and a storage service.
+ * @param defaultNotary used in [fillWithSomeTestDeals] and [fillWithSomeTestLinearStates].
+ * @param altNotary used in [fillWithSomeTestCash], [fillWithSomeTestCommodity] and consume/evolve methods. If not specified, same as [defaultNotary].
+ * @param rngFactory used by [fillWithSomeTestCash] if no custom [Random] provided.
+ */
+class VaultFiller @JvmOverloads constructor(
+        private val services: ServiceHub,
+        defaultNotary: Pair<Party, KeyPair>,
```
rnicoll: I'm not sure about providing these as a pair. In Java where otherwise the user could try providing one half but not another I'd understand, but as Kotlin enforces non-null, I'm not sure it's worth wrapping the types just to immediately unwrap them again in the class. I'd rather just have `defaultNotary` and `defaultNotaryKeyPair` as separate value parameters.

andr3ej: done, with a check that the public keys match as VaultFiller probably assumes they do

PR #2119, opened 2017-11-23
Overhaul of the Bank of Corda demo to fix two problems it had:

shamsasari:
1. The runRPCCashIssue and runWebCashIssue gradle tasks didn't work because they were using the wrong ports
2. Notary lookup was failing because the lookup name didn't include the correct CN for the notary name (this slipped through when reverting the network parameters)

The ports change occurred in #1922 which was attempting the fix the runIssuer gradle task. This is actually a misleading and redundant task as all it does is start up the nodes, which is what the documented deployNodes already does. The ports runIssuer allocated to the nodes were different to the ones specified in deployNodes.

To make sure we have integration tests which closely match deployNodes, the BoC demo has been updated to make use of CordformDefinition. This keeps the node definitions in one place, removing the need to have disparate files in sync. runIssuer has been removed.


shamsasari:
With regards to your first point, I've not included any unrelated cleanups and refactorings. The changes here were what was needed to first isolate the issues and get the demo to be failing correctly (yes that's right, when they failed it still said "BUILD SUCCESSFUL"!) in a TDD manner.

With regard to your second, I think `CordformDefinition` is the key to avoid future mess ups with the demos because it allows us to write integration tests which match the real usage of the demos much more closely. In fact I'd like the other demos to be follow this pattern and have smoke tests for all of them. But I'm happy for @mikehearn to chime in on this.

shamsasari:
@Clintonio I reverted that change. Now the list of CorDapp packages specified in CordformationDefinition gets converted to a list of CorDapp jars that contain them.

mikehearn:
Hmm ... have I suggested that? I think I'm getting senile. It sounds possible but I don't quite recall when or where I suggested we retire Cordformation. Perhaps what I was getting at is that having both driver and cordform duplicates configuration and it'd be nice to find a fix for that.

mikehearn:
I scan-read the PR: seems ideal! Correct me if I'm wrong but it looks like after this change:

* `gradle deployNodes` calls into Java/Kotlin code which uses a DSL that looks very similar to before and creates a corda formation ("cordform" is perhaps a rather ugly word...) based on the configured objects, so config files are generated and runnodes still exists.
* However the DSL also becomes executable, implemented using the driver. Running the gradle integration test tasks thus calls into Kotlin from Groovy and from Kotlin up *back* into the Gradle plugin, which is implemented in Java, to create the configuration objects, which is then fed to the driver and used to spin up the nodes for integration testing. Phew!

The end result of this gradle judo is that we have configuration for both command line/manual testing, and integration testing, in one place and one place only, and codepaths for them are heavily shared.

Is that right? If so, bravo, it would seem to resolve the issues we have.

josecoll:
@shamsasari what I meant on the 1st point, is that all the code changes in this PR under the 
 gradle-plugins/cordform-common
 gradle-plugins/cordformation
are not related to Bank of Corda fixes, and should be in a separate PR.
(eg. I can review the the BoC changes but know nothing about Cordformation)

shamsasari:
@josecoll The changes to cordformation were needed to make BoC demo work with it. I added @Clintonio for those changes and you for the demo itself :).

josecoll:
Do the instructions in the Bank of Corda README.cmd (used primarily by QA) still apply after these refactorings?

samples/irs-demo/src/integration-test/kotlin/net/corda/irs/IRSDemoTest.kt
```diff
@@ -117,7 +117,7 @@ class IRSDemoTest {
         log.info("Running trade against ${nodeApi.root}")
         val fileContents = loadResourceFile("net/corda/irs/web/simulation/example-irs-trade.json")
         val tradeFile = fileContents.replace("tradeXXX", "trade1").replace("oracleXXX", oracle.name.toString())
-        assertThat(nodeApi.postJson("deals", tradeFile)).isTrue()
+        nodeApi.postJson("deals", tradeFile)
```
josecoll: loss of test assertion. How do we know this test passed successfully now?

samples/irs-demo/web/src/test/kotlin/net/corda/irs/web/demo/IrsDemoClientApi.kt
```diff
@@ -11,10 +11,10 @@ import org.apache.commons.io.IOUtils
 class IRSDemoClientApi(private val hostAndPort: NetworkHostAndPort) {
     private val api = HttpApi.fromHostAndPort(hostAndPort, apiRoot)
 
-    fun runTrade(tradeId: String, oracleName: CordaX500Name): Boolean {
+    fun runTrade(tradeId: String, oracleName: CordaX500Name) {
         val fileContents = IOUtils.toString(javaClass.classLoader.getResourceAsStream("net/corda/irs/simulation/example-irs-trade.json"), Charsets.UTF_8.name())
         val tradeFile = fileContents.replace("tradeXXX", tradeId).replace("oracleXXX", oracleName.toString())
-        return api.postJson("deals", tradeFile)
+        api.postJson("deals", tradeFile)
```
josecoll: wasn't this status flag used to provide assertion checking?
(or is it a meaningless check)

shamsasari:
W.r.t the loss of assertions for one of the HTTP helpers, that was changed to throw an IOException instead of returning a boolean, so the tests fail on the exception. The reason for this was simply because the demo wasn't checking the false case and so would fail with an error message but gradle would say "BUILD SUCCESSFUL"! We shouldn't be using a boolean success indicator for these reasons. The other HTTP helpers need to be changed as well, so I added to TODO.

And with regards to the README.md, it's instructions are correct. I haven't modified the documented process of running the demo.

josecoll:
I give my approval from the Bank of Corda side.
Over to @Clintonio for the CordFormation magic.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -129,10 +132,16 @@ open class Cordform : DefaultTask() {
     private fun initializeConfiguration() {
         if (definitionClass != null) {
             val cd = loadCordformDefinition()
+            require(directory === defaultDirectory) {
```
Clintonio: Can you add a comment to explain this?

samples/bank-of-corda-demo/build.gradle
```diff
@@ -48,51 +48,7 @@ dependencies {
...
-                ['username'   : "bigCorpUser",
-                 'password'   : "test",
-                 'permissions': ["StartFlow.net.corda.finance.flows.CashPaymentFlow",
-                                 "StartFlow.net.corda.finance.flows.CashConfigDataFlow",
-                                 "InvokeRpc.waitUntilNetworkReady",
-                                 "InvokeRpc.wellKnownPartyFromX500Name",
-                                 "InvokeRpc.notaryIdentities"]]
-        ]
-    }
+    definitionClass = 'net.corda.bank.BankOfCordaCordform'
```
Clintonio: Please put a comment here explaining this is still experimental.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
shamsasari: Done

samples/bank-of-corda-demo/build.gradle
shamsasari: Done

PR #2122, opened 2017-11-24
Added tests to make sure the platform version is correctly available

shamsasari:

### DIFF
--- a/constants.properties
+++ b/constants.properties
@@ -1,5 +1,6 @@
 gradlePluginsVersion=2.0.9
 kotlinVersion=1.1.60
+platformVersion=2
 guavaVersion=21.0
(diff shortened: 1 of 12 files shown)


constants.properties
```diff
@@ -1,5 +1,6 @@
 gradlePluginsVersion=2.0.9
 kotlinVersion=1.1.60
+platformVersion=2
```
sollecitom: These are just the tests, so maybe not so relevant, but are we sure about having the version as an integer? Shall we not use a `String` instead?

shamsasari: Using an integer makes comparing straightforward in app code. It's similar to Andriod's API Level, which is used to allow apps to know what features are available to them.

PR #2124, opened 2017-11-24
Corda now works with H2 without the need to allow Hibernate to create the database automatically.

sollecitom:
Also renamed property `database.initDatabase` to `database.createSchemaAutomatically`.

[CORDA-815]: https://r3-cev.atlassian.net/browse/CORDA-815


sollecitom:
Hey Jose, Mike stated clearly the need to rename it. I'm happy to leave it
as it is or change it to whatever you guys prefer. Just let me know.

josecoll:
The description of this configuration item needs to be updated in the CordaDocs to reflect the settings we have in our code as per:
https://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#configurations-hbmddl
(likewise, any change in name)

sollecitom:
Decided to be renamed to `adjustSchemas` (Mike). I'll also make sure it's properly documented.

sollecitom:
Please @mikehearn and @josecoll have another look.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -43,7 +43,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
 data class DevModeOptions(val disableCheckpointChecker: Boolean = false)
 
 data class DatabaseConfig(
-        val initDatabase: Boolean = true,
+        val adjustSchemas: Boolean = true,
```
shamsasari: Perhaps annotate this with `@OldConfig` so that the node logs a warning to switch to the new config. Actually I think `OldConfig` should throw an exception rather than logging.

sollecitom: Nobody is really using this with `H2` so far, we might still change it in a clean way.

sollecitom: We just documented it today. Not released yet.

josecoll:
I honestly don't think we should change `initDatabase`.
`adjustSchemas` is not meaningful at all.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
josecoll: Please do not change the name of this attribute.
`adjustSchemas` is not meaningful.

mikehearn:
Can you elaborate on why?

As noted elsewhere, the problem with `initDatabases` is that the term `initialization` often implies resets or erasures. If you initialise a filesystem on your hard disk, and one is already present, you can lose data. If you initialise a variable, the old contents disappear. This already confused one user who took the time to read the docs and take action based on the naming, so, clearly the name is confusing. Documenting the fact that it doesn't do what it sounds like is a poor substitute for renaming it to something better.

Now, if you can think of a better name than adjustSchemas I'm all for it. But this name has the advantage that "adjustment" does not imply "deletion" or "reset".

sollecitom:
`adjustSchemas` also seems to indicate that it can be either creating a schema or patching an existing one to make Corda work with it.

sollecitom:
As a matter of fact, `adjustSchema` seems better suited as singular than plural in my mind, but I'm not strongly opinionated enough on this.

josecoll:
OK - lets carefully analyse what our line of code does:
`.setProperty("hibernate.hbm2ddl.auto", if (databaseConfig.initDatabase) "update" else "validate")`

and re-state what the meaning of these operations (as per the Hibernate manual):
`update`: Update the database schema
`validate`: Validate the database schema
(note the `update` operation also performs **creation** if schemas are non-existent).

My preferred options would be to remove this "boolean" flag, and use a strongly typed configuration attribute that allows setting any of: none, create, create-only, drop, validate, update.

Otherwise keep as it (and default to **false**).

mikehearn:
Yes. I'm afraid I still don't see what you're getting at :( So "init" means "update or create". I don't know of a single word in English that satisfactorily captures the concept of both updating or creating, but adjust seems close enough. If initDatabase=false (adjustSchema=false) then it only validates. That seems reasonable too.

I don't see why it needs to be replaced with a more complex option, but I can see an argument for just deleting it and using the devMode setting instead. If we make the reasonable assumption that in production the DBA will want to do migrations themselves, rather than automatically at startup, and in dev the developer will never want to do migrations themselves, then it seems like the setting may be somewhat redundant to begin with. Does that make sense?

josecoll:
I'm all for simplification so removing this item completely (and assuming it is only ever used in DevMode) makes complete sense.

mikehearn:
Great. Michele thinks there might be production deployments where automatic schema adjustment is wanted, but I'm not so sure - at least for "prod" in the traditional datacenter IT sense. I can see we might want such a thing in a desktop node, but let's ignore that for now. So we seem to have agreement on removal.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -79,7 +79,9 @@ class CordappLoader private constructor(private val cordappJarPaths: List<Restri
          */
         @VisibleForTesting
         fun createDefaultWithTestPackages(configuration: NodeConfiguration, testPackages: List<String>): CordappLoader {
-            check(configuration.devMode) { "Package scanning can only occur in dev mode" }
+            if (!configuration.devMode) {
```
sollecitom: This is result of bundling `devMode` and the schema validation logic. Basically, I need to be able to run a test without Corda in `devMode`.

sollecitom:
@mikehearn and @josecoll I pushed changes according to our conversation. Please have another look.

docs/source/changelog.rst
```diff
@@ -6,9 +6,9 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
-* Corda is now able to work with H2 and Hibernate when property ``database.adjustSchemas`` is set to ``false``.
+* Corda is now able to work with H2 and Hibernate when property Hibernate validates an existing schema, rather than creating it.
```
josecoll: Remove this line and just clarify behaviour below:
eg. in `dev-mode` Hibernate will attempt to create/update database schemas.

sollecitom:
@josecoll yes, as shown here: https://github.com/corda/corda/pull/2124/files#diff-1b7638327ed5645b8095460b1749cb6bR90.

docs/source/changelog.rst
```diff
@@ -6,7 +6,8 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
-* Corda is now able to work with H2 and Hibernate when property Hibernate validates an existing schema, rather than creating it.
+* Corda is now able to work with H2 and Hibernate. In ``devMode`` Hibernate will try to create or update a schema, while
+  not in ``devMode`` Hibernate will validate the existing schema, failing if this is not found or not compliant.
 
 * Removed confusing property ``database.initDatabase``, enabling its guarded behaviour with the ``dev-mode``.
```
josecoll: Sorry for being pedantic but I asked to remove the line that commences with:
"Corda is now able to work with H2 and Hibernate."  -> this is not new, it has always been the case (so the line is misleading).

Just have a single line item that states: 
"Removed confusing property ``database.initDatabase``, enabling its guarded behaviour with the ``dev-mode``.  In ``devMode`` Hibernate will try to create or update database schemas, otherwise it will expect relevant schemas to be present in the database (pre configured via DDL scripts or equivalent), and validate these are correct.

sollecitom: Pushed again. I thought the whole point of this work was to fix a bug between H2 and Hibernate.

PR #2125, opened 2017-11-24
ENT-1151: Rework unicode block validation rules

rnicoll:


rnicoll:
I've split this up further and made full/minimal validation a parameter (with no default for safety). Minimal validation is done typically on a node and just sanity checks the name isn't totally nuts, full is something the Doorman does.

rnicoll:
Have rebased to knock the dust off this, but presuming it passes am about to merge. Any last minute comments?

PR #2126, opened 2017-11-24
Network permissioning new doc

joeldudleyr3:


docs/source/permissioning.rst
```diff
@@ -1,105 +1,99 @@
...
-The following information from the node configuration file is needed to generate the request.
+In development mode (i.e. when ``devMode = true``, see :doc:`corda-configuration-file` for more information),
+pre-configured keystores are used if the required keystores do not exist. This ensures that developers can get the
+nodes working as quickly as possible.
 
-:myLegalName: Your company's legal name as an X.500 string. X.500 allows differentiation between entities with the same
-    name as the legal name needs to be unique on the network. If another node has already been permissioned with this
-    name then the permissioning server will automatically reject the request. The request will also be rejected if it
-    violates legal name rules, see `Legal Name Constraints`_ for more information.
+However, these pre-configured keystores are not secure. For a real network, you need to create a certificate authority
```
gendal: For a real network, you need to use an existing secure certificate authority that will be used....     or create one of your own. This document explains how to create your own. However, note that in the near future it is expected that most production deployments will use an existing authority or construct one using software that will be made available in the coming months.  Until then, the documentation below can be used.  (or something like this... ie I want to signal that this is just a point in time thing)

docs/source/permissioning.rst
```diff
@@ -1,105 +1,99 @@
...
+   * We recommend X.509 v3 for forward compatibility
 
-The utility will submit the request to the doorman server and poll for a result periodically to retrieve the certificates.
-Once the request has been approved and the certificates downloaded from the server, the node will create the keystore and trust store using the certificates and the generated private key.
+* The TLS certificates must follow the `TLS v1.2 standard <https://tools.ietf.org/html/rfc5246>`_
 
-.. note:: You can exit the utility at any time if the approval process is taking longer than expected. The request process will resume on restart. 
+Creating the node certificate authority
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The node certificate authority is used to sign the node identity certificates. An intermediate CA is used instead of
```
gendal: Is the node certificate authority the root CA?  If so, this is confusing... since it makes it sound like it's specific to a given node.

docs/source/permissioning.rst
```diff
@@ -1,105 +1,99 @@
...
+2. Create a self-signed certificate for the keypair. The basic constraints extension must be set to ``true``
 
-* Remove leading and trailing whitespaces.
+   * This will be used as the root CA's certificate
 
-* Replace multiple whitespaces with a single space.
+3. Store the root CA's keypair and certificate in a keystore for later use
 
-* Normalize the string according to `NFKC normalization form <https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization>`_.
+4. Store the root CA's certificate in a Java keystore named ``truststore.jks`` using the alias ``cordarootca``
```
gendal: whose Java keystore?  Presumably the root CA is created once but the cert needs to be added to EVERY node's store?

docs/source/permissioning.rst
```diff
@@ -1,105 +1,99 @@
...
+2. Obtain a certificate for the keypair signed with the root CA key. The basic constraints extension must be set to
+   ``true``
 
-* No dollars or quote marks, although we may relax the quote mark constraint in future to handle Irish company names.
+   * This will be used as the intermediate CA's certificate
 
-Starting the Registration
--------------------------
+3. Store the intermediate CA's keypair and certificate chain (i.e. the intermediate CA certificate *and* the root CA
+   certificate) in a keystore for later use
```
gendal: an _a_ keystore?  Where? each node? just somewhere safe but not as part of any node?

docs/source/permissioning.rst
```diff
@@ -1,105 +1,99 @@
...
 
-**To start the registration**::
+Creating the node CA keystores
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    java -jar corda.jar --initial-registration --base-dir <<optional>> --config-file <<optional>>
+1. On each node, create a new keypair
 
-A ``certificates`` folder containing the keystore and trust store will be created in the base directory when the process is completed.
+2. Obtain a certificate for the keypair signed with the intermediate CA key. The basic constraints extension must be
```
gendal: we use CA in two contexts in this doc... node CA and network-level CA?  If so, perhaps disambiguate here?

joeldudleyr3:
Thanks @gendal . Was re-reading on the weekend and the two levels of CA's is confusing. I'll try and disambiguate.

But I don't understand what you mean by "an _a_ keystore"?

docs/source/permissioning.rst
gendal: sorry...  I wasn't clear. I was questioning what you meant by "in a keystore".  Any keystore?  One I have hanging around in my shed?  Or a specific one and if so which and where?

PR #2130, opened 2017-11-27
CORDA-599 IdentityService is no longer obtained lazily

andr3ej:


andr3ej:
yes that appears to be the case

PR #2131, opened 2017-11-27
Guice demo

al-r3:
Minimal demo of adding Guice to corda.

I've added it in the top most part of the code.

This is a very minimal example which shows how one can remove code from a large class using dependency injection. 
3 dependencies of NodeStartup have been passed in as dependencies, and those are provided from modules.

Node that, unlike Spring, there isn't any scanning of classes to find dependencies, modules explicitly define all the mappings and each and every provided object.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/Corda.kt
+++ b/node/src/main/kotlin/net/corda/node/Corda.kt
@@ -3,11 +3,22 @@
 
 package net.corda.node
 
+import com.google.inject.Guice
 import net.corda.node.internal.NodeStartup
+import net.corda.node.internal.NodeStartupArgumentsModule
+import net.corda.node.internal.VersionInfoModule
 
 fun main(args: Array<String>) {
     // Pass the arguments to the Node factory. In the Enterprise edition, this line is modified to point to a subclass.
     // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
     // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
-    NodeStartup(args).run()
+    val modules = listOf(NodeStartupArgumentsModule(args), VersionInfoModule())
+
+    val injector = Guice.createInjector(modules)
+
+    val nodeStartup = injector.getInstance(NodeStartup::class.java)
+
+    nodeStartup.run()
 }
+
+
--- a/node/src/main/kotlin/net/corda/node/Corda.kt
+++ b/node/src/main/kotlin/net/corda/node/Corda.kt
@@ -12,11 +12,14 @@ fun main(args: Array<String>) {
     // Pass the arguments to the Node factory. In the Enterprise edition, this line is modified to point to a subclass.
     // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
     // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
-    val modules = listOf(NodeStartupArgumentsModule(args), VersionInfoModule())
 
-    val injector = Guice.createInjector(modules)
+    /**
+     * List of installed modules.
+     */
+    val modules = listOf(NodeStartupArgumentsModule(args),
+            VersionInfoModule())
 
-    val nodeStartup = injector.getInstance(NodeStartup::class.java)
+    val nodeStartup = Guice.createInjector(modules).getInstance(NodeStartup::class.java)
 
     nodeStartup.run()
 }
--- a/node/src/main/kotlin/net/corda/node/internal/NodeArgumentParser.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeArgumentParser.kt
@@ -1,0 +1,46 @@
+package net.corda.node.internal
+
+import com.google.inject.AbstractModule
+import com.google.inject.Provides
+import joptsimple.OptionException
+import net.corda.node.ArgsParser
+import net.corda.node.CmdLineOptions
+import kotlin.system.exitProcess
+
+interface NodeArgumentParser {
+    val argsParser: ArgsParser
+    val cmdLineOptions: CmdLineOptions
+}
+
+class NodeStartupArgumentsModule(val arguments: Array<String>): AbstractModule() {
+
+    @Provides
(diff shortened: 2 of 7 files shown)


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -24,10 +25,14 @@ import java.net.InetAddress
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.*
+import javax.inject.Inject
 import kotlin.system.exitProcess
 
+
 /** This class is responsible for starting a Node from command line arguments. */
-open class NodeStartup(val args: Array<String>) {
+open class NodeStartup @Inject constructor(private val argsParser: ArgsParser,
```
sollecitom: Does this work with `internal` constructor?

node/src/main/kotlin/net/corda/node/internal/NodeArgumentParser.kt
```diff
@@ -0,0 +1,46 @@
...
+
+interface NodeArgumentParser {
+    val argsParser: ArgsParser
+    val cmdLineOptions: CmdLineOptions
+}
+
+class NodeStartupArgumentsModule(val arguments: Array<String>): AbstractModule() {
+
+    @Provides
+    fun providerNodeArgumentParser() : NodeArgumentParser = NodeArgumentParserImpl(arguments)
```
sollecitom: Is it possible to bind only the mapping between interface and implementation class, rather than the whole instance?

node/src/main/kotlin/net/corda/node/Corda.kt
```diff
@@ -3,11 +3,25 @@
...
     // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
     // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
-    NodeStartup(args).run()
-}
\ No newline at end of file
+
+    /**
+     * List of installed modules.
+     */
+    val modules = listOf(NodeStartupArgumentsModule(args),
```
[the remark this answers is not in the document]
al-r3: Not that I know of. Seems contrary to the 'guice' philosophy where you explicitly bind things.
For maps/sets there are sets binders:
https://github.com/google/guice/wiki/Multibindings

summarizing you can bind a set by having distinct modules binding an element each.

node/src/main/kotlin/net/corda/node/internal/NodeArgumentParser.kt
al-r3: There's a way which is discouraged: https://github.com/google/guice/wiki/JustInTimeBindings#implementedby

In general you want to be able to switch implementations (even if you don't have another implementation just yet).

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
al-r3: This is an interesting question, I think it depends, if you're calling Guice from the same module yes, otherwise it doesn't.

Though, Guice should not (and Guice does not want to) be the only way you construct an object, Guice simply calls out to a marked constructor that any user of the class could call if they want to.

sollecitom:
Good integration overall, it would easily cater for our DI needs.
Two observations:

- Spring implements JSR300 (Inject, Provider, Qualifier) and JSR250 (PostConstruct, PreDestroy), while Guice just the former. JSR250 is quite useful to allow active plugins and components, and to manage correct releasing of resources on application's shutdown.

- The inability to discover modules (and hence the need to know their fully qualified name to register them) would mean the need to have drop-in plugin declare their modules in a mandated way (a file with a fully qualified name we know, where they declare modules, one per line).

- From your example it looks like you have to register module instances rather than classes. This implies that you also have to hardcode the order of constructing these modules (example, if module A needs a type from module B injected, so that it can declare it as a provided instance, then module A's constructor will take that argument from B, meaning you have to know as a programmer to construct first B and then A).

- How easy would it be to construct a test Guice context from the production one, allowing replacement of some instances with Mocks?

- Spring can be used in this fashion as well, provided we would like this approach more than the discovery route.

- It's not clear how you could inject collection instances, like `Set<A>`, ` MutableList<B>`, etc. In Spring it's not only possible, but you can also use it to discover and collect all instances implementing an interface e.g., all endpoints you then wire in a web server.

sollecitom:
@mikehearn it'd be good for you to have a look at this too, and compare it with the Spring Boot example.

al-r3:
Point by point:

1 true :) Guice doesn't do lifecycle I've worked on complex app and didn't find myself needing that.
PostConstruct seems something can be done lazily/moved to the constructor
PreDestroy seems a somewhat bad idea, in general, you can guarantee that your app will do a nice shutdown anyway, so you shouldn't depend on that.

2 Guice doesn't do any magic there, you need to know all the classes you have/need. I quite like that, but you don't get to configure an app by putting together a few jars

3 You don't specifically need to know the order of construction, you can inject anything you want in a @Provides method, when you build the injector it will fail if it can't satisfy the dependencies.
A module can explicitly declare it needs a certain key (class + qualifier annotation) to be bound for it to bind its bindings:
https://google.github.io/guice/api-docs/latest/javadoc/index.html?com/google/inject/AbstractModule.html

4 There is no "context" for tests:
if they're small you don't use Guice, you simply call the constructors manually and provide the few dependencies you need manually.
For larger tests it depends, usually you have fake implementations of larger interfaces (say networking/database) then you bind those in modules thought specifically to be used for tests.
And lastly you create an injector where you mix some 'production' modules with real implementations bound and some fake implementations bound on the 'border' of what you want your test to test.

There is little/no creating an injector by "patching" an existing one, if it's doable, I've never seen it.

5 You can inject any class you want.
The easy way to inject a Set<A> is simply to bind a whole Set<A> at once, there's a trick to workaround type erasure if you want to use the bind().to() syntax, but if you use an @Provided method is trivial.

OR there are the MultiBinders: https://github.com/google/guice/wiki/Multibindings where a Set/Map is assembled by different modules providing each one or more entry in the set/map.
This lets you do a bit of a plugin based architecture but you still need to know all the types you need at compile time.
(for us this works well for serialization whitelists, to say one)

PR #2134, opened 2017-11-27
 CORDA-785: Add functions for constructing FlowLogicRef without the class

rnicoll:
Removes the need for schedulable flow states to hold a reference to the class.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
@@ -14,6 +14,7 @@ interface FlowLogicRefFactory {
      * Construct a FlowLogicRef. This is intended for cases where the calling code has the relevant class already
      * and can provide it directly.
      */
+    @Deprecated("This should be avoided, and the version which takes a class name used instead to avoid requiring the class on the classpath to deserialize calling code")
     fun create(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRef
     /**
      * Construct a FlowLogicRef. This is intended for cases where the calling code does not want to require the flow
--- a/core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
@@ -10,7 +10,16 @@ import net.corda.core.serialization.CordaSerializable
  */
 @DoNotImplement
 interface FlowLogicRefFactory {
+    /**
+     * Construct a FlowLogicRef. This is intended for cases where the calling code has the relevant class already
+     * and can provide it directly.
+     */
     fun create(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRef
+    /**
+     * Construct a FlowLogicRef. This is intended for cases where the calling code does not want to require the flow
+     * class on the classpath for all cases where the calling code is loaded.
+     */
+    fun create(flowClassName: String, vararg args: Any?): FlowLogicRef
     fun createForRPC(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRef
     fun toFlowLogic(ref: FlowLogicRef): FlowLogic<*>
 }
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
@@ -40,6 +40,15 @@ class FlowLogicRefFactoryImpl(private val classloader: ClassLoader) : SingletonS
         return createForRPC(flowClass, *args)
     }
 
+    override fun create(flowClassName: String, vararg args: Any?): FlowLogicRef {
+        val flowClass = Class.forName(flowClassName) as? Class<FlowLogic<*>>
+        if (flowClass == null) {
+            throw IllegalArgumentException("The class $flowClassName is not a subclass of FlowLogic.")
+        } else {
+            return create(flowClass, *args)
+        }
+    }
+
     override fun createForRPC(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRef {
         // TODO: This is used via RPC but it's probably better if we pass in argument names and values explicitly
         // to avoid requiring only a single constructor.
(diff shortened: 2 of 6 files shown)


core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
```diff
@@ -11,6 +11,10 @@ import net.corda.core.serialization.CordaSerializable
 @DoNotImplement
 interface FlowLogicRefFactory {
     fun create(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRef
+    fun create(flow: FlowNameAndArguments): FlowLogicRef
```
shamsasari: All these methods need docs.

core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
```diff
@@ -11,6 +11,10 @@ import net.corda.core.serialization.CordaSerializable
...
+    fun verify(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowNameAndArguments
+
+    class FlowNameAndArguments(val flowClass: String, val args: List<Pair<String, Any?>>)
```
shamsasari: Let's not use `Pair` for the public API as it's not very useful in Java. Looking at the code below is the String type the param name of the flow c'tor? Looks like it's only used for Kotlin?

shamsasari: Rename param to `flowClassName`

node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
```diff
@@ -35,42 +35,31 @@ object FlowLogicRefFactoryImpl : SingletonSerializeAsToken(), FlowLogicRefFactor
     // TODO: Replace with a per app classloader/cordapp provider/cordapp loader - this will do for now
     var classloader: ClassLoader = javaClass.classLoader
 
+    /**
+     * Construct a FlowLogicRef, intended for cases where the flow logic class is available.
+     */
```
shamsasari: This class is implementation detail and so this doc is in the wrong place. It should be in FlowLogicRefFactory

core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
```diff
@@ -11,6 +11,10 @@ import net.corda.core.serialization.CordaSerializable
 @DoNotImplement
 interface FlowLogicRefFactory {
     fun create(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRef
```
shamsasari: Based on my reading of CORDA-785, should method be deprecated?

core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
rnicoll: Unfortunately the `Pair` has to be exposed so that `FlowLogicRefFactoryImpl` can get to it, but isn't really intended for general use. There's no harm in it being visible, but it isn't the most useful data type I know.

shamsasari: Am I correct in understanding that `Pair.first` are unique, and so this can be a Map?

mkit: Can it be a data class?

node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
```diff
@@ -138,4 +127,43 @@ object FlowLogicRefFactoryImpl : SingletonSerializeAsToken(), FlowLogicRefFactor
...
+     * flow object, so intermediary nodes do not have to have a copy of the flow to verify the state object.
+     */
+    override fun verify(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRefFactory.FlowNameAndArguments {
+        if (!flowClass.isAnnotationPresent(SchedulableFlow::class.java)) {
+            throw IllegalFlowLogicException(flowClass, "because it's not a schedulable flow")
+        }
+        return verifyForRPC(flowClass, *args)
+    }
+
+    fun verifyForRPC(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRefFactory.FlowNameAndArguments {
```
mkit: Missing docs.

rnicoll:
Switched IRS to using this, and fixed up issues arising (turns out we can't actually pre-validate these, realistically, and instead we just pass in the class name).

rnicoll:
Rebased just to make sure there's no weird problems sneaking in. Any chance of a re-review @shamsasari ?

node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
```diff
@@ -42,6 +42,15 @@ object FlowLogicRefFactoryImpl : SingletonSerializeAsToken(), FlowLogicRefFactor
         return createForRPC(flowClass, *args)
     }
 
+    override fun create(flowClassName: String, vararg args: Any?): FlowLogicRef {
+        val flowClass = Class.forName(flowClassName) as? Class<FlowLogic<*>>
```
shamsasari: Use `classloader` instead of whatever `forName` uses. Also I'm not sure why it's a `var`. See if you can make it into a `val`.

shamsasari: You need to use `Class.isAssignableFrom` or `asSubclass` to correctly determine the class type.

core/src/main/kotlin/net/corda/core/flows/FlowLogicRef.kt
```diff
@@ -10,7 +10,16 @@ import net.corda.core.serialization.CordaSerializable
  */
 @DoNotImplement
 interface FlowLogicRefFactory {
+    /**
+     * Construct a FlowLogicRef. This is intended for cases where the calling code has the relevant class already
+     * and can provide it directly.
+     */
     fun create(flowClass: Class<out FlowLogic<*>>, vararg args: Any?): FlowLogicRef
```
shamsasari: Shouldn't this method be deprecated as it just encourages the situation we're trying to avoid?

rnicoll:
@rick-r3 Any comments before I merge this in?

PR #2142, opened 2017-11-28
Create a dedicated Log4J2 file for NodeExplorer and output logging configuration used.

vkolomeyko:


testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -979,7 +979,8 @@ class DriverDSL(
                     "name" to nodeConf.myLegalName,
                     "visualvm.display.name" to "corda-${nodeConf.myLegalName}",
                     Node.scanPackagesSystemProperty to cordappPackages.joinToString(Node.scanPackagesSeparator),
-                    "java.io.tmpdir" to System.getProperty("java.io.tmpdir") // Inherit from parent process
+                    "java.io.tmpdir" to System.getProperty("java.io.tmpdir"), // Inherit from parent process
+                    "log4j2.debug" to "true"
```
m4ksio: Shouldn't this be configurable? And default to false, we rather don't want to spam every single test using driver

vkolomeyko: This modification is for `startOutOfProcessNode` only.
I think it is extremely helpful to know which logging configuration is used when the node is started as a separate process since we have multiple logging configurations in the classpath.

`-Dlog4j2.debug=true` option produces around 20 Kbytes printout in StdErr, please see example attached only once per node startup
[net.corda.node.Corda.stderr.log](https://github.com/corda/corda/files/1510185/net.corda.node.Corda.stderr.log)

m4ksio: But this method is used by driver, which means lots of our integration tests depends on it. I don't think this option is useful for automatically run tests - it shines while debugging some problem - hence it can be then turned on on demand. @shamsasari opinions?

vkolomeyko: How about instead of creating a new property make `log4j2.debug` enabled if `DriverDSL.isDebug` is enabled?

shamsasari: That makes sense.

PR #2144, opened 2017-11-29
Updating docs on network map REST endpoint - GET

mkit:
Update the network map docs about the GET REST endpoint. There was a recent change to the Network Map web service that introduced the 404 response code possibility.


shamsasari:
I think @patrickkuo made recent changes to the REST endpoints.

docs/source/network-map.rst
```diff
@@ -37,6 +37,8 @@ Network Map service REST API:
 | POST           | /api/network-map/publish          | Publish new ``NodeInfo`` to the network map service, the legal identity in ``NodeInfo`` must match with the identity registered with the doorman.      |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
 | GET            | /api/network-map                  | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and NetworkParameters hash.                       |
+|                |                                   | Due to possible decoupling of the network map signing process from the network map providing service, the 404 (Not Found) response can be returned.    |
```
shamsasari: Yes you're right, no need to go into such detail. In fact I don't think you even need to mention 404 because it can only occur right at the very beginning where there's no nodes.

PR #2147, opened 2017-11-29
Moved CordaPersistence and dependent classes into internal package in…

shamsasari:
… node-api. This is so that the doorman can make use of them without needing a compile dependency to node.


node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -1,64 +1,75 @@
...
+
 class CordaPersistence(
-        val dataSource: HikariDataSource,
-        private val schemaService: SchemaService,
-        private val identityService: IdentityService,
-        databaseConfig: DatabaseConfig
+        val dataSource: DataSource,
+        databaseConfig: DatabaseConfig,
+        schemas: Set<MappedSchema>,
+        attributeConverters: List<AttributeConverter<*, *>>
```
josecoll: Can we make this parameter optional (? = null)? 
(as may not be needed at all for simple CordaPersistence)

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -1,64 +1,75 @@
...
             }
         }
     }
 
     /**
-     * Creates an instance of [DatabaseTransaction], with the given isolation level.
-     * @param isolationLevel isolation level for the transaction. If not specified the default (i.e. provided at the creation time) is used.
+     * Creates an instance of [DatabaseTransaction], with the transaction isolation level specified at the creation time.
      */
-    fun createTransaction(isolationLevel: Int): DatabaseTransaction {
```
josecoll: Good removal - shouldn't be changing the isolation level. 
I think this was added to hack around a Persistent Network Map Cache issue in `updateInfoDB`.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -116,21 +127,8 @@ class CordaPersistence(
...
-fun configureDatabase(dataSourceProperties: Properties, databaseConfig: DatabaseConfig, identityService: IdentityService, schemaService: SchemaService = NodeSchemaService(null)): CordaPersistence {
-    val config = HikariConfig(dataSourceProperties)
-    val dataSource = HikariDataSource(config)
-    val persistence = CordaPersistence.connect(dataSource, schemaService, identityService, databaseConfig)
-    // Check not in read-only mode.
-    persistence.transaction {
-        persistence.dataSource.connection.use {
-            check(!it.metaData.isReadOnly) { "Database should not be readonly." }
-        }
+        (dataSource as? AutoCloseable)?.close()
```
al-r3: This looks a bit odd, why can this be AutoCloseable or not?
Add a comment

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/HibernateConfiguration.kt
```diff
@@ -18,28 +14,27 @@ import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider
...
 import org.hibernate.type.descriptor.sql.VarbinaryTypeDescriptor
 import java.sql.Connection
 import java.util.concurrent.ConcurrentHashMap
+import javax.persistence.AttributeConverter
 
-class HibernateConfiguration(val schemaService: SchemaService, private val databaseConfig: DatabaseConfig, private val identityService: IdentityService) {
+class HibernateConfiguration(
+        schemas: Set<MappedSchema>,
+        private val databaseConfig: DatabaseConfig,
+        private val attributeConverters: List<AttributeConverter<*, *>>
```
al-r3: nit: attributeConverters can be a Collection

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
shamsasari: I can default it to an empty list, which is the better thing to do.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
shamsasari: No it's used by the doorman. I think it was a mistake to add it as the isolation level should have been set in the properties. Now that we have `DatabaseConfig` it should be easier.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/DatabaseTransaction.kt
```diff
@@ -0,0 +1,61 @@
+package net.corda.nodeapi.internal.persistence
+
+import org.hibernate.Session
+import org.hibernate.Transaction
+import rx.subjects.Subject
+import java.sql.Connection
+import java.util.*
+
+class DatabaseTransaction(
```
al-r3: Add javadoc.
Also specify why you need a ThreadLocal

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/HibernateConfiguration.kt
```diff
@@ -18,28 +14,27 @@ import org.hibernate.engine.jdbc.connections.spi.ConnectionProvider
...
 import org.hibernate.type.descriptor.java.PrimitiveByteArrayTypeDescriptor
 import org.hibernate.type.descriptor.sql.BlobTypeDescriptor
 import org.hibernate.type.descriptor.sql.VarbinaryTypeDescriptor
 import java.sql.Connection
 import java.util.concurrent.ConcurrentHashMap
+import javax.persistence.AttributeConverter
 
-class HibernateConfiguration(val schemaService: SchemaService, private val databaseConfig: DatabaseConfig, private val identityService: IdentityService) {
+class HibernateConfiguration(
+        schemas: Set<MappedSchema>,
```
josecoll: Excellent refactoring as this was previously always bootstrapping ALL the schemas defined in `NodeSchemaService` (whether they are required or not).

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
shamsasari: Done

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/HibernateConfiguration.kt
shamsasari: Done

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/DatabaseTransaction.kt
shamsasari: I just moved this class into its own file. Rather not document it when I haven't had the time to study it.

shamsasari:
The `schemaService: SchemaService = NodeSchemaService(null)` declarations are needed now that CordaPeristence doesn't hold a SchemaService. Not sure how else to do it.

Otherwise back to you @josecoll.

josecoll:
As part of this refactor / clean-up, would it be possible to also update `NodeSchemaService` to not take a `cordappLoader: CordappLoader?` constructor argument, but instead a set of schemas:
`Set<MappedSchema>`. Would help immensely with some work I am doing.

shamsasari:
Done what you asked, but had to add a second c'tor for the CordappLoader as it's required in more than one place.

josecoll:
Can you not just remove the CorDapp constructor completely, and just replace the 2 calls to it using the new default constructor and passing in `cordappLoader?.cordapps?.flatMap { it.customSchemas }` ?
(NB: `SchemaOptions` is not used anywhere afaik)

shamsasari:
Moved `cordappLoader?.cordapps?.flatMap { it.customSchemas }` into CordappLoader. So now everything is nice and clean!

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
shamsasari: Just spoke with @mkit and he confirmed that we need this as an override for specific cases. In the doorman case we need to use the highest level (serialised?) in certain situations which is not use suitable as a global level. I'll rename some stuff to make this more clear.

PR #2148, opened 2017-11-29
Fix demobench as part of network parameters work

kasiastreich:
Pregenerate notary identity using ServiceIdentityGenerator and drop network parameters file into nodes directories. Only one notary is allowed to be started through Demobench. I tested also loading profiles.


tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -67,11 +85,14 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
...
                 rpcAddress = nodeData.rpcPort.toLocalAddress(),
                 webAddress = nodeData.webPort.toLocalAddress(),
-                notary = nodeData.extraServices.filterIsInstance<NotaryService>().noneOrSingle(),
+                notary = notary,
                 h2port = nodeData.h2Port.value,
                 issuableCurrencies = nodeData.extraServices.filterIsInstance<CurrencyIssuer>().map { it.currency.toString() }
         )
 
+        if (notary != null && notaryIdentity != null && notaryIdentity?.name != nodeConfig.myLegalName)
+            throw IllegalArgumentException("Only single notary allowed")
```
shamsasari: Use `require`

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -67,11 +85,14 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
                 p2pAddress = nodeData.p2pPort.toLocalAddress(),
                 rpcAddress = nodeData.rpcPort.toLocalAddress(),
                 webAddress = nodeData.webPort.toLocalAddress(),
-                notary = nodeData.extraServices.filterIsInstance<NotaryService>().noneOrSingle(),
+                notary = notary,
                 h2port = nodeData.h2Port.value,
                 issuableCurrencies = nodeData.extraServices.filterIsInstance<CurrencyIssuer>().map { it.currency.toString() }
         )
 
+        if (notary != null && notaryIdentity != null && notaryIdentity?.name != nodeConfig.myLegalName)
```
shamsasari: If we requiring max one notary, why not simply check `notaryIdentity == null`?

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -115,22 +137,74 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
             val cordaEnv = System.getenv().toMutableMap().apply {
                 jvm.setCapsuleCacheDir(this)
             }
+            if (config.nodeConfig.notary != null)
+                makeNetworkParametersCopier(config)
             pty.run(command, cordaEnv, config.nodeDir.toString())
             log.info("Launched node: ${config.nodeConfig.myLegalName}")
+            networkParametersCopier?.install(config.nodeDir) ?: throw IllegalArgumentException("No network parameters specified")
```
shamsasari: Isn't it more of a bug in the code if we get here and `networkParametersCopier` is still null? It should just be `networkParametersCopier!!`

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -115,22 +137,74 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
             val cordaEnv = System.getenv().toMutableMap().apply {
                 jvm.setCapsuleCacheDir(this)
             }
+            if (config.nodeConfig.notary != null)
```
shamsasari: Why check if the node is a notary before creating the network parameters? I would have thought this check should just be `if (networkParametersCopier == null)`

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -115,22 +137,74 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
...
+            networkParametersCopier?.install(config.nodeDir) ?: throw IllegalArgumentException("No network parameters specified")
             return true
         } catch (e: Exception) {
             log.log(Level.SEVERE, "Failed to launch Corda: ${e.message}", e)
             return false
         }
     }
 
+    private fun makeNetworkParametersCopier(config: NodeConfigWrapper) {
+        if (notaryIdentity == null) {
```
shamsasari: Why do we need this check?

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
kasiastreich: It is because demobench let's you do different things with nodes. Like kill the notary, then start another node which can be the same notary or can be different node with notary in it, but it won't be in NetworkParameters this time.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
kasiastreich: Yes, you are right, it is a leftover code before i made networkParametersCopier nullable (and it's nullable because loading profiles in Demobench).

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -102,6 +120,7 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
 
     fun runCorda(pty: R3Pty, config: NodeConfigWrapper): Boolean {
         try {
+            require((config.nodeConfig.notary != null).xor(notaryIdentity != null)) { "There must be exactly one notary in the network" }
```
shamsasari: Add a comment that a notary can be removed and then added again.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -115,6 +134,9 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
             val cordaEnv = System.getenv().toMutableMap().apply {
                 jvm.setCapsuleCacheDir(this)
             }
+            if (networkParametersCopier == null)
+                makeNetworkParametersCopier(config)
+            networkParametersCopier!!.install(config.nodeDir)
```
chrisr3: Not really a fan of using `!!` because I think Kotlin wants to free of things like NPEs. How about:
```kotlin
(networkParametersCopier ?: makeNetworkParametersCopier(config)).install(config.nodeDir)
```
which would then require:
```kotlin
private fun makeNetworkParametersCopier(config: NodeConfigWrapper): NetworkParametersCopier {
    try {
        // etc
        val identity = getNotaryIdentity(config)
        val parametersCopier = NetworkParametersCopier(NetworkParameters(
            // etc
        ))
        notaryIdentity = identity
        networkParametersCopier = parametersCopier
        return parametersCopier
    } finally {
        // etc...
    }
}
```
which (happily) removes another `!!` here too :-).

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -124,13 +146,60 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
...
+            if (_contextSerializationEnv.get() != null) _contextSerializationEnv.set(null)
+        }
+    }
+
+    private fun initialiseSerialization() {
+        val context = if (java.lang.Boolean.getBoolean("net.corda.testing.amqp.enable")) AMQP_P2P_CONTEXT else KRYO_P2P_CONTEXT
+        _contextSerializationEnv.set(SerializationEnvironmentImpl(
+                SerializationFactoryImpl().apply {
+                    registerScheme(KryoDemoSerializationScheme)
+                    registerScheme(AMQPServerSerializationScheme())
```
chrisr3: `AMQPServerSerializationScheme()` is for the server, i.e. the node. You wanted `AMQPClientSerializationScheme()`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -124,13 +146,60 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
...
+                },
+                context))
+    }
+
+    private object KryoDemoSerializationScheme : AbstractKryoSerializationScheme() {
```
chrisr3: You wanted the pre-existing `KryoClientSerializationScheme()` object here.

kasiastreich:
@chrisr3 back to you

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeController.kt
```diff
@@ -115,6 +127,9 @@ class NodeController(check: atRuntime = ::checkExists) : Controller() {
             val cordaEnv = System.getenv().toMutableMap().apply {
                 jvm.setCapsuleCacheDir(this)
             }
+            if (networkParametersCopier == null)
```
chrisr3: You don't need this `if` statement any more because you have Elvis `?:`.

tools/demobench/src/main/kotlin/net/corda/demobench/DemoBench.kt
```diff
@@ -1,7 +1,14 @@
 package net.corda.demobench
 
 import javafx.scene.image.Image
+import net.corda.client.rpc.internal.KryoClientSerializationScheme
+import net.corda.core.serialization.internal.SerializationEnvironmentImpl
+import net.corda.core.serialization.internal.nodeSerializationEnv
 import net.corda.demobench.views.DemoBenchView
+import net.corda.nodeapi.internal.serialization.AMQP_P2P_CONTEXT
```
chrisr3: Unused import statment.

PR #2149, opened 2017-11-29
CORDA-819 Add checks to ensure TLS and Client CA cert chains to the same trusted root

patrickkuo:
* Add checks to ensure TLS and Client CA cert chains to the same trusted root
* make identity service log and ignore bad identity, instead of throwing exception
* integration tests

### DIFF
--- a/node/src/integration-test/kotlin/net/corda/node/NodeKeystoreCheckTest.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/NodeKeystoreCheckTest.kt
@@ -4,6 +4,7 @@ import net.corda.core.crypto.Crypto
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.internal.cert
 import net.corda.core.internal.div
+import net.corda.core.utilities.getOrThrow
 import net.corda.node.services.config.configureDevKeyAndTrustStores
 import net.corda.nodeapi.config.SSLConfiguration
 import net.corda.nodeapi.internal.crypto.*
@@ -11,7 +12,6 @@ import net.corda.testing.ALICE_NAME
 import net.corda.testing.driver.driver
 import org.junit.Test
 import java.nio.file.Path
-import java.util.concurrent.ExecutionException
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 import kotlin.test.assertTrue
@@ -21,11 +21,10 @@ class NodeKeystoreCheckTest {
     fun `node should throw exception if cert path doesn't chain to the trust root`() {
         driver(startNodesInProcess = true) {
             // This will fail because there are no keystore configured.
-            assertFailsWith(ExecutionException::class) {
-                startNode(customOverrides = mapOf("devMode" to false)).get()
+            assertFailsWith(IllegalArgumentException::class) {
+                startNode(customOverrides = mapOf("devMode" to false)).getOrThrow()
             }.apply {
-                assertTrue(cause is IllegalArgumentException)
-                assertTrue(cause?.message?.startsWith("Identity certificate not found. ") ?: false)
+                assertTrue(message?.startsWith("Identity certificate not found. ") ?: false)
             }
 
             // Create keystores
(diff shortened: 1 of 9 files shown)


node/src/integration-test/kotlin/net/corda/node/NodeKeystoreCheckTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+            // Self signed root
+            val badRootKeyPair = Crypto.generateKeyPair()
+            val badRoot = X509Utilities.createSelfSignedCACertificate(CordaX500Name("Bad Root", "Lodnon", "GB"), badRootKeyPair)
+            val nodeCA = keystore.getCertificateAndKeyPair(X509Utilities.CORDA_CLIENT_CA, config.keyStorePassword)
+            val badNodeCACert = X509Utilities.createCertificate(CertificateType.CLIENT_CA, badRoot, badRootKeyPair, ALICE_NAME, nodeCA.keyPair.public)
+            keystore.setKeyEntry(X509Utilities.CORDA_CLIENT_CA, nodeCA.keyPair.private, config.keyStorePassword.toCharArray(), arrayOf(badNodeCACert.cert, badRoot.cert))
+            keystore.save(config.nodeKeystore, config.keyStorePassword)
+
+            assertFailsWith(ExecutionException::class) {
```
mikehearn: EE is quite a general type - perhaps we can assert on a more specific type (the cause?) or on the message text, so we know we're testing the right path?

node/src/main/kotlin/net/corda/node/services/identity/InMemoryIdentityService.kt
```diff
@@ -67,7 +67,8 @@ class InMemoryIdentityService(identities: Iterable<PartyAndCertificate> = emptyS
                 val space = (0 until index).joinToString("") { "   " }
                 log.error("$space${certificate.toX509CertHolder().subject}")
             }
-            throw e
+            // Log the error and continue.
+            return null
```
mikehearn: Is that right?! Surely it's better to propagate the exception here?

node/src/main/kotlin/net/corda/node/services/identity/InMemoryIdentityService.kt
```diff
@@ -67,7 +67,8 @@ class InMemoryIdentityService(identities: Iterable<PartyAndCertificate> = emptyS
                 val space = (0 until index).joinToString("") { "   " }
```
mikehearn: Is the TODO above this function done now?

node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
```diff
@@ -122,7 +122,8 @@ class PersistentIdentityService(override val trustRoot: X509Certificate,
             identity.certPath.certificates.reversed().forEach {
                 log.error(it.toX509CertHolder().subject.toString())
             }
-            throw e
+            // Log the error and continue.
```
mikehearn: ditto - looks a bit dangerous?

node/src/main/kotlin/net/corda/node/services/identity/InMemoryIdentityService.kt
patrickkuo: yes thats not right, I should handle the exception in the network map cache instead.

node/src/main/kotlin/net/corda/node/services/identity/InMemoryIdentityService.kt
patrickkuo: @rnicoll ? same in `PersistentIdentityService` as well.

node/src/integration-test/kotlin/net/corda/node/NodeKeystoreCheckTest.kt
shamsasari: Use `getOrThrow()` instead of `get()` to get the actual exception.

node/src/main/kotlin/net/corda/node/services/identity/InMemoryIdentityService.kt
rnicoll: I'll do that today - I think it's fine, but the note was mostly there so I went back and re-checked it after some time away from the code, in case of oversights.

rnicoll: Yup, all looks good, do you want to remove the TODOs as part of this PR please?

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -37,13 +38,22 @@ class NetworkMapCacheImpl(
...
         networkMapCacheBase.allNodes.forEach { it.legalIdentitiesAndCerts.forEach { identityService.verifyAndRegisterIdentity(it) } }
         networkMapCacheBase.changed.subscribe { mapChange ->
             // TODO how should we handle network map removal
             if (mapChange is MapChange.Added) {
                 mapChange.node.legalIdentitiesAndCerts.forEach {
-                    identityService.verifyAndRegisterIdentity(it)
+                    try {
+                        identityService.verifyAndRegisterIdentity(it)
+                    } catch (ignore: Exception) {
+                        // Log a warning to indicate node info is not added to the network map cache.
```
rnicoll: Are we planning on soft-failing here, in the future, or is this a temporary solution until the network map is always validated? Or; if this should change later, please add a `TODO`, although the longer I spend writing this comment the more I'm realising maybe we're stuck with this because we can't refuse to start due to a mangled network map...

patrickkuo: the new network map will always validate node info, so this should never fail, this is here to make sure node don't just die in case network map got corrupted somehow.

PR #2151, opened 2017-11-29
ENT-1125 bootstrap root certificate

al-r3:
During the node registration process make nodes check that the root of the certificate is an expected certificate stored under /certificates/rootCaCert.cer

Note this PR makes the registration FAIL if there is no such file, we might want to relax that for testing/dev purposes.

Also this borrows some code from doorman (to craft a fake doorman reply), all that code should be put in the OS repo so that both doorman & tests can use it.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/config/SSLConfiguration.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/config/SSLConfiguration.kt
@@ -10,6 +10,7 @@ interface SSLConfiguration {
     val sslKeystore: Path get() = certificatesDirectory / "sslkeystore.jks"
     val nodeKeystore: Path get() = certificatesDirectory / "nodekeystore.jks"
     val trustStoreFile: Path get() = certificatesDirectory / "truststore.jks"
+    val rootCaCertFile: Path get() = certificatesDirectory / "rootCaCert.cer"
 }
 
 interface NodeSSLConfiguration : SSLConfiguration {
--- a/node-api/src/main/kotlin/net/corda/nodeapi/config/SSLConfiguration.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/config/SSLConfiguration.kt
@@ -10,10 +10,10 @@ interface SSLConfiguration {
     val sslKeystore: Path get() = certificatesDirectory / "sslkeystore.jks"
     val nodeKeystore: Path get() = certificatesDirectory / "nodekeystore.jks"
     val trustStoreFile: Path get() = certificatesDirectory / "truststore.jks"
-    val rootCaCertFile: Path get() = certificatesDirectory / "rootCaCert.cer"
 }
 
 interface NodeSSLConfiguration : SSLConfiguration {
     val baseDirectory: Path
     override val certificatesDirectory: Path get() = baseDirectory / "certificates"
+    val rootCaCertFile: Path get() = certificatesDirectory / "rootcacert.cer"
 }
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
@@ -315,7 +315,7 @@ class X509CertificateFactory {
     }
 
     // TODO migrate calls to [CertificateFactory#generateCertPath] to call this instead.
-    fun buildCertPath(vararg certificates: Certificate): CertPath {
+    fun generateCertPath(vararg certificates: Certificate): CertPath {
         return delegate.generateCertPath(certificates.asList())
     }
 }
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
@@ -313,6 +313,10 @@ class X509CertificateFactory {
     fun generateCertificate(input: InputStream): X509Certificate {
         return delegate.generateCertificate(input) as X509Certificate
     }
+
+    fun buildCertPath(vararg certificates: Certificate): CertPath {
+        return delegate.generateCertPath(certificates.asList())
+    }
 }
 
 enum class CertificateType(val keyUsage: KeyUsage, vararg val purposes: KeyPurposeId, val isCA: Boolean) {
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
@@ -314,6 +314,7 @@ class X509CertificateFactory {
         return delegate.generateCertificate(input) as X509Certificate
     }
 
+    // TODO migrate calls to [CertificateFactory#generateCertPath] to call this instead.
     fun buildCertPath(vararg certificates: Certificate): CertPath {
         return delegate.generateCertPath(certificates.asList())
     }
--- a/node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
@@ -68,8 +68,8 @@ class NetworkRegistrationHelperDriverTest {
         driver(portAllocation = portAllocation,
                 compatibilityZoneURL = compatibilityZoneUrl,
                 startNodesInProcess = true,
-                rootCertificate = rootCert) {
-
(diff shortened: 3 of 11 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/config/SSLConfiguration.kt
```diff
@@ -10,6 +10,7 @@ interface SSLConfiguration {
     val sslKeystore: Path get() = certificatesDirectory / "sslkeystore.jks"
     val nodeKeystore: Path get() = certificatesDirectory / "nodekeystore.jks"
     val trustStoreFile: Path get() = certificatesDirectory / "truststore.jks"
+    val rootCaCertFile: Path get() = certificatesDirectory / "rootCaCert.cer"
```
shamsasari: Looks like the existing convention here is to use all lowercase filenames. Best to stick with that.

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
```diff
@@ -313,6 +313,10 @@ class X509CertificateFactory {
     fun generateCertificate(input: InputStream): X509Certificate {
         return delegate.generateCertificate(input) as X509Certificate
     }
+
+    fun buildCertPath(vararg certificates: Certificate): CertPath {
```
shamsasari: I didn't add this method when I created this class since it didn't add much value. If you think it does I suggest you add a TODO to have existing uses of this class use this method.

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -151,3 +167,9 @@ class NetworkRegistrationHelper(private val config: NodeConfiguration, private v
         }
     }
 }
+
+/**
+ * Exception thrown when the doorman root certificate doesn't match the expected (out-of-band) root certificate.
+ * This usually means the has been a Man-in-the-middle attack when contacting the doorman.
+ */
+class WrongRootCaCertificateException: Exception()
```
shamsasari: This needs a description: "Root CA returned back from registration does not match what we think is the root CA. Either this is either a MITM attack or a fatal misconfiguration!"

You also need to include both certs in the exception so that node operator can see what went wrong.

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+
+/**
+ * Simple registration handler which can handle a single request, which will be given request id [REQUEST_ID].
+ */
+@Path("certificate")
+class RegistrationHandler(
+        private val certificateAndKeyPair: CertificateAndKeyPair) {
```
shamsasari: Move this back up

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -207,6 +209,8 @@ interface DriverDSLExposedInterface : CordformContext {
     }
 
     val shutdownManager: ShutdownManager
+
+    fun writeRootCaCertificateForNode(nodeName: CordaX500Name, caRootCertificate: X509CertificateHolder)
```
shamsasari: The Driver will be part of the public testing API soon and so we should be careful what we add to it. In this case we don't even need this method here, as the implementation can be moved to the test class that uses it.

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
```diff
@@ -0,0 +1,181 @@
...
+private const val REQUEST_ID = "requestId"
+
+private val x509CertificateFactory = X509CertificateFactory()
+private val portAllocation = PortAllocation.Incremental(10000)
+
+/**
+ * Driver based tests for [NetworkRegistrationHelper]
+ */
+class NetworkRegistrationHelperDriverTest {
+    val certificateAndKeyPair = createSelfKeyAndSelfSignedCertificate()
```
shamsasari: rootCertAndKeyPair

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
```diff
@@ -0,0 +1,181 @@
...
+    val caRootCertificate = certificateAndKeyPair.certificate
```
shamsasari: rootCert

node-api/src/main/kotlin/net/corda/nodeapi/config/SSLConfiguration.kt
shamsasari: Actually this shouldn't be in SSLConfiguration because this is used by other processes besides the nodes. Move it to NodeSSLConfiguration.

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
al-r3: Add the todo, there are quite a few uses of CertificateFactory#generateCertPath throughout the code base which can be shortened.

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
al-r3: Added all. (plus the file where expected is stored).

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
al-r3: Done though, if we expose the parameters to test the initial registration (and we are) we should also expose something to write the expected cert, maybe as a driver parameter.

Now if anyone wants to test the initial registration - which driver supports - they have to write the code to serialize the cert themselves

shamsasari: Yes you are right. Can we have it that a X509Certificate object is passed in if doing initial-registration and the serialisation to file is an internal implementation detail.

I'm now wondering whether the --initial-registration feature for the driver should actually just be an internal API (i.e. defined just in `DriverDSL`) for use by us. I'm less convinced that it's something needed by an app developer.

al-r3: Added it as a driver param.
As whether they need to test the initial registration, probably not. To test that in a sensible way you need to know quite a bit on how doorman works anyway.

shamsasari: Yeah we probably need to clean it up once things have settled.

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
```diff
@@ -313,6 +313,11 @@ class X509CertificateFactory {
     fun generateCertificate(input: InputStream): X509Certificate {
         return delegate.generateCertificate(input) as X509Certificate
     }
+
+    // TODO migrate calls to [CertificateFactory#generateCertPath] to call this instead.
+    fun buildCertPath(vararg certificates: Certificate): CertPath {
```
shamsasari: Perhaps keep the name the same with generateCertPath

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
```diff
@@ -0,0 +1,178 @@
...
+    fun stopServer() {
+        server.close()
+    }
+
+    @Test
+    fun `node registration correct root cert`() {
+        driver(portAllocation = portAllocation,
+                compatibilityZoneURL = compatibilityZoneUrl,
+                startNodesInProcess = true,
+                rootCertificate = rootCert) {
```
shamsasari: Move `) {` to the next line. That way the driver block doesn't look so odd

shamsasari:
@patrickkuo can you point Al on how to use the anchor verification.

patrickkuo:
you can use `X509Utilities.validateCertificateChain(trustedRoot: X509Certificate, vararg certificates: Certificate)` to validate the cert path before storing the certs.

al-r3:
I know what goes wrong. 
The notary doesn't start with initialRegistration = true (I can't pass params to the notary) so the notary uses the cert in the repository, while my node uses the cert I generate in the test

PR #2157, opened 2017-11-30
Shams master merge 291117

shamsasari:
Latest merge from master.

Also includes bump to gradle plugins version and moving NodeInfoFilesCopier into internal package.


testing/node-driver/src/main/kotlin/net/corda/testing/internal/demorun/DemoRunner.kt
```diff
@@ -35,6 +33,7 @@ fun CordformDefinition.deployNodesThen(block: () -> Unit) {
 }
 
 private fun CordformDefinition.runNodes(waitForAllNodesToFinish: Boolean, block: () -> Unit) {
+    clean()
```
al-r3: why is this needed? Add a comment in the code

PR #2162, opened 2017-12-01
Showcase of closures registered as flows.

sollecitom:
Given that people routinely ask, and we have no example of this in our tests, this is how to register flows dynamically in order to construct meaningful tests.


core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
```diff
@@ -20,6 +20,42 @@ class ReceiveMultipleFlowTests {
         mockNet.stopNodes()
     }
 
+    @Test
+    fun showcase_flows_as_closures() {
```
szymonsztuka: To be consistent with naming could the method name be change to 'showcase flows as closures' and no empty line below it? Apart form that looks good.

PR #2165, opened 2017-12-01
CORDA-806 Use non-static Rx pool when testing

andr3ej:
so that the inheritable thread local serialization env can be propagated through it without affecting subsequent tests


andr3ej:
@mnesbit @patrickkuo does this help with the problems you've been seeing?

mnesbit:
Testing it

PR #2166, opened 2017-12-01
Ensure we write a single NodeInfo when using Cordformation

al-r3:
Ensure NodeInfo is generated only once, specifically when using Cordformation to deploy a node.

Now initNodeInfo takes a NetworkMapCacheBaseInternal so that it can check the database to see if we already have a NodeInfo "for" this node.

Also removed the dependency of IdentityService on NodeInfo (it wasn't used anyway) to break the dependency loop, as initialiseDatabasePersistence needs an IdentityService


shamsasari:
Is it possible to move this to master? I'll be merging our branch soon and if this is valid on master then we can avoid the extra round-trip, otherwise I will delay approval.

al-r3:
Raised https://github.com/corda/corda/pull/2174 against master.

But the 'merge' isn't trivial (it's not difficult either).
Also that has a fix to ensure we change the nodeInfo if something changes in the conf (this doesn't)

shamsasari:
Can this fix you speak of also be applied to #2174?

al-r3:
That (aka) #2174 has it.
This doesn't, as it seems you're more keen on approving #2174

PR #2171, opened 2017-12-04
add upstart instructions

benabineri:
A user asked on Discourse about running Corda using Upstart (Ubuntu 14.04) rather than SystemD so I thought I'd add the instructions here too.

### DIFF
--- a/docs/source/deploying-a-node.rst
+++ b/docs/source/deploying-a-node.rst
@@ -77,7 +77,8 @@ handling, and ensures the Corda service is run at boot.
       * Country (``C=``) is the `ISO 3166-1 alpha-2 code <https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2>`_
    *  Change the RPC username and password
 
-.. note:: Ubuntu 16.04 and other modern Linux distributions use **SystemD**. 
+.. note:: Ubuntu 16.04 and most current Linux distributions use SystemD, so if you are running one of these
+          distributions follow the steps marked **SystemD**. 
           If you are running Ubuntu 14.04, follow the instructions for **Upstart**.
 
 8. **SystemD**: Create a ``corda.service`` file based on the example below and save it in the ``/etc/systemd/system/``
--- a/docs/source/deploying-a-node.rst
+++ b/docs/source/deploying-a-node.rst
@@ -7,9 +7,9 @@ Deploying a node
    whether they have developed and tested a CorDapp following the instructions in :doc:`generating-a-node`
    or are deploying a third-party CorDapp.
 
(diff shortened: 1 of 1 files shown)


benabineri:
@joeldudleyr3 could you check the formatting please? I couldn't work out how to build the docs locally.

docs/source/deploying-a-node.rst
```diff
@@ -27,10 +27,13 @@ handling, and ensures the Corda service is run at boot.
 3. Download the `Corda jar <https://r3.bintray.com/corda/net/corda/corda/>`_
    (under ``/VERSION_NUMBER/corda-VERSION_NUMBER.jar``) and place it in ``/opt/corda``
 
-3. Create a directory called ``plugins`` in ``/opt/corda`` and save your CorDapp jar file to it. Alternatively, download one of
+4. (Optional) Download the `Corda webserver jar <http://r3.bintray.com/corda/net/corda/corda-webserver/>`_
+   (under ``/VERSION_NUMBER/corda-VERSION_NUMBER.jar``) and place it in ``/opt/corda``
+
+5. Create a directory called ``plugins`` in ``/opt/corda`` and save your CorDapp jar file to it. Alternatively, download one of
```
wawrzek: I thought that on master 'plugins' are renamed 'cordapps'.

benabineri: I think these instructions will be "backported" to the 2.0 docs, which does not support "cordapps", whereas master supports both currently (AFAIK)

joeldudleyr3:
@benabineri I'll check the formatting

joeldudleyr3:
@benabineri I've fixed the formatting.

I feel like the layout is confusing. How about having `X.a.` and `X.b.` steps? I think you should also put a note at the top saying "These docs provide instructions for both Systemd and Upstart. We will highlight the steps where the instructions for these two systems differ. You only need to follow the instructions for one of the systems."

benabineri:
@joeldudleyr3 is that not what line 80-ish says?

benabineri:
The steps are labelled 7. SystemD, 7. Upstart, 8., 9. SystemD, 9. Upstart etc.

joeldudleyr3:
@benabineri Sorry, missed that. I still think 8a vs. 8b would be less confusing, but your call.

Thoughts on changing:
"Ubuntu 16.04 and other modern Linux distributions use SystemD. If you are running Ubuntu 14.04, follow the instructions for Upstart."

To:
"If you are running Ubuntu 16.04 and other modern Linux distributions, follow the instructions for SystemD. If you are running Ubuntu 14.04, follow the instructions for Upstart."\

Wording wasn't clear to me on first reading.

PR #2173, opened 2017-12-04
CORDA-654 Pass some key constants into MockServices

andr3ej:


mnesbit:
I don't mind the code styling here, but I am worried that this isn't getting to the root issue, which is that random made up identities are being used in our example contracts. Do we have a plan for making these identities relate to actual nodes and the public/anonymised keys of these nodes?

PR #2174, opened 2017-12-04
Create a single NodeInfo

al-r3:
Ensure NodeInfo is generated only once, specifically when using Cordformation to deploy a node.

Now initNodeInfo takes a NetworkMapCacheBaseInternal so that it can check the database to see if we already have a NodeInfo "for" this node.

Also removed the dependency of IdentityService on NodeInfo (it wasn't used anyway) to break the dependency loop, as initialiseDatabasePersistence needs an IdentityService

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -174,26 +174,33 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
         check(started == null) { "Node has already been started" }
         log.info("Generating nodeInfo ...")
         initCertificate()
-        val (keyPairs, info) = initNodeInfo()
-        val identityKeypair = keyPairs.first { it.public == info.legalIdentities.first().owningKey }
-        val serialisedNodeInfo = info.serialize()
-        val signature = identityKeypair.sign(serialisedNodeInfo)
-        // TODO: Signed data might not be sufficient for multiple identities, as it only contains one signature.
-        NodeInfoWatcher.saveToFile(configuration.baseDirectory, SignedData(serialisedNodeInfo, signature))
+        val schemaService = NodeSchemaService(cordappLoader.cordappSchemas)
+        val (identity, identityKeyPair) = obtainIdentity(notaryConfig = null)
+        initialiseDatabasePersistence(schemaService,  makeIdentityService(identity.certificate)) { database ->
+            val persistentNetworkMapCache = PersistentNetworkMapCache(database)
+            val (keyPairs, info) = initNodeInfo(persistentNetworkMapCache, identity, identityKeyPair)
+            val identityKeypair = keyPairs.first { it.public == info.legalIdentities.first().owningKey }
+            val serialisedNodeInfo = info.serialize()
+            val signature = identityKeypair.sign(serialisedNodeInfo)
+            // TODO: Signed data might not be sufficient for multiple identities, as it only contains one signature.
+            NodeInfoWatcher.saveToFile(configuration.baseDirectory, SignedData(serialisedNodeInfo, signature))
+        }
     }
 
     open fun start(): StartedNode<AbstractNode> {
         check(started == null) { "Node has already been started" }
         log.info("Node starting up ...")
         initCertificate()
-        val (keyPairs, info) = initNodeInfo()
         val schemaService = NodeSchemaService(cordappLoader.cordappSchemas)
-        val identityService = makeIdentityService(info)
+        val (identity, identityKeyPair) = obtainIdentity(notaryConfig = null)
+        val identityService = makeIdentityService(identity.certificate)
         // Do all of this in a database transaction so anything that might need a connection has one.
         val (startedImpl, schedulerService) = initialiseDatabasePersistence(schemaService, identityService) { database ->
+            val networkMapCache = NetworkMapCacheImpl(PersistentNetworkMapCache(database), identityService)
+            val (keyPairs, info) = initNodeInfo(networkMapCache, identity, identityKeyPair)
             identityService.loadIdentities(info.legalIdentitiesAndCerts)
             val transactionStorage = makeTransactionStorage(database)
-            val nodeServices = makeServices(keyPairs, schemaService, transactionStorage, database, info, identityService)
+            val nodeServices = makeServices(keyPairs, schemaService, transactionStorage, database, info, identityService, networkMapCache)
             val notaryService = makeNotaryService(nodeServices, database)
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
```diff
@@ -21,7 +21,6 @@ import java.security.cert.*
 interface IdentityService {
     val trustRoot: X509Certificate
     val trustAnchor: TrustAnchor
-    val caCertStore: CertStore
```
shamsasari: This is part of the public API now and can't be removed.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -271,13 +277,21 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
         )
-        return Pair(keyPairs, info)
+        // Check if we have already stored a version of 'our own' NodeInfo, this is to avoid regenerating it with
+        // a different timestamp.
+        networkMapCache.getNodesByLegalName(myLegalName).firstOrNull()?.let {
+            if (info.copy(serial = it.serial) == it) {
+                info = it
+            }
+        }
+        //val info =
```
shamsasari: ?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -271,13 +277,21 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+        return Pair(keyPairs.toSet(), info)
```
shamsasari: keyPairs is already a Set, and the return type doesn't expose that it's a MutableSet

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -271,13 +277,21 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+        var info = NodeInfo(
                 myAddresses(),
                 listOf(identity, myNotaryIdentity).filterNotNull(),
                 versionInfo.platformVersion,
                 platformClock.instant().toEpochMilli()
         )
-        return Pair(keyPairs, info)
+        // Check if we have already stored a version of 'our own' NodeInfo, this is to avoid regenerating it with
+        // a different timestamp.
+        networkMapCache.getNodesByLegalName(myLegalName).firstOrNull()?.let {
```
shamsasari: Can you do the look up and first and then create the new nodeinfo if needed.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -173,27 +174,33 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
         // Do all of this in a database transaction so anything that might need a connection has one.
         val (startedImpl, schedulerService) = initialiseDatabasePersistence(schemaService, identityService) { database ->
+            val networkMapCache = NetworkMapCacheImpl(PersistentNetworkMapCache(database), identityService)
+            val (keyPairs, info) = initNodeInfo(networkMapCache)
             identityService.loadIdentities(info.legalIdentitiesAndCerts)
             val transactionStorage = makeTransactionStorage(database)
             val stateLoader = StateLoaderImpl(transactionStorage)
-            val nodeServices = makeServices(keyPairs, schemaService, transactionStorage, stateLoader, database, info, identityService)
+            val nodeServices = makeServices(keyPairs, schemaService, transactionStorage, stateLoader, database, info, identityService, networkMapCache)
+            services.networkMapCache.addNode(info)
```
shamsasari: This is the same as `networkMapCache`? Can we move this adding to initNodeInfo?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -173,27 +174,33 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
-        NodeInfoWatcher.saveToFile(configuration.baseDirectory, SignedData(serialisedNodeInfo, signature))
+        val schemaService = NodeSchemaService(cordappLoader.cordappSchemas)
+        initialiseDatabasePersistence(schemaService,  makeIdentityService()) { database ->
+            val persistentNetworkMapCache = PersistentNetworkMapCache(database)
+            val (keyPairs, info) = initNodeInfo(persistentNetworkMapCache)
+            val identityKeypair = keyPairs.first { it.public == info.legalIdentities.first().owningKey }
+            val serialisedNodeInfo = info.serialize()
+            val signature = identityKeypair.sign(serialisedNodeInfo)
+            // TODO: Signed data might not be sufficient for multiple identities, as it only contains one signature.
+            NodeInfoWatcher.saveToFile(configuration.baseDirectory, SignedData(serialisedNodeInfo, signature))
```
shamsasari: What happens if generate node info is called again? It should overwrite the previous node info I think.

al-r3: generateNodeInfo() is the function which is called with the special flag to generate the nodeInfo.
So there isn't really a way of calling this "again"

Furthermore we might want to remove the flag to make the node write its NodeInfo and quit (I might need to check with @benabineri who once asked me about it) right now, with Kasia's PR it is not needed anymore by Cordformation

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: yes and done

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: removed

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
al-r3: It's always needed.
If I don't find an older one, then I need to create one.
If I find an older one, then I need to create the "new" one to see if they're different.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -620,12 +641,12 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
         }
     }
 
-    private fun makeIdentityService(info: NodeInfo): PersistentIdentityService {
+    private fun makeIdentityService(cert: X509Certificate): PersistentIdentityService {
```
shamsasari: `identityCert`

PR #2176, opened 2017-12-04
Shams merge feature network parameters

shamsasari:
The network parameters work is ready to be merged into master. The original issues with deployNodes and the demobench have been resolved.

For a lack of a better idea I've assigned this to @corda/core.


mikehearn:
Maybe squash the PR first so there is a clean set of commits to work from? I assigned Matthew and Andrzej to review.

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/SSLConfiguration.kt
```diff
@@ -15,4 +15,5 @@ interface SSLConfiguration {
 interface NodeSSLConfiguration : SSLConfiguration {
     val baseDirectory: Path
     override val certificatesDirectory: Path get() = baseDirectory / "certificates"
+    val rootCaCertFile: Path get() = certificatesDirectory / "rootcacert.cer"
```
mnesbit: This is a very confusing name. This is not the 'root' certificate, which is in the truststore. This is the node authorization ca, or node publisher certificate, or something.

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -74,10 +75,15 @@ class NetworkRegistrationHelper(private val config: NodeConfiguration, private v
             caKeyStore.addOrReplaceKey(CORDA_CLIENT_CA, keyPair.private, privateKeyPassword.toCharArray(), certificates)
             caKeyStore.deleteEntry(SELF_SIGNED_PRIVATE_KEY)
             caKeyStore.save(config.nodeKeystore, keystorePassword)
+
+            // Check the root certificate.
+            val returnedRootCa = certificates.last()
+            checkReturnedRootCaMatchesExpectedCa(returnedRootCa)
```
mnesbit: Verifying a certpath always checks that the CA is in the truststore. So why do we do this? Should this be checking the doorman certificate in some way instead?

samples/irs-demo/cordapp/build.gradle
```diff
@@ -38,7 +38,7 @@ dependencies {
     // Specify your cordapp's dependencies below, including dependent cordapps
     compile group: 'commons-io', name: 'commons-io', version: '2.5'
 
-    testCompile project(':node-driver')
+    cordaCompile project(':node-driver')
```
mnesbit: This is making the Driver part of our public API. Are we really sure about this?

shamsasari:
This is the design doc: https://r3-cev.atlassian.net/wiki/spaces/AWG/pages/127710793/New+Network+Map.

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/SSLConfiguration.kt
shamsasari: The root cert is given to the node out of band, as discussed in the design doc.

samples/irs-demo/cordapp/build.gradle
shamsasari: The driver is going to become part of the public testing API. This doesn't change that. And the other demos have this same dependency.

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/SSLConfiguration.kt
mnesbit: I am confused! Is this the doorman issued cert, or the root? The truststore containing the root is provided out of band along with their new certificate. This contains the root and will in future also need to contain revocation certs.

shamsasari: This is the root cert. It's given to the node out of band. When the node completes the signing process with the doorman and gets back its node CA it verifies that the root cert matches with the one that it was given.

mnesbit: But we will almost certainly need more certs than just the root in here. I can have some sympathy with Kostas saying that we shouldn't have a plain text password that is the same everywhere, but this doesn't actually protect anything. I would rather that we download a fixed .jks file that I can check the SHA256 hash of against a separate advertised value.

shamsasari: Passing the cert (securely) out of band does the same job, not sure why we need to add more work with hosting an advertised value. If we need more certs then they can be separate files, or zipped if we want to avoid dealing with a public truststore password. And this can be evolved without affecting public API.

constants.properties
```diff
@@ -1,4 +1,4 @@
-gradlePluginsVersion=3.0.0
+gradlePluginsVersion=3.0.2
```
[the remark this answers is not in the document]
shamsasari: Yep

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/SSLConfiguration.kt
mnesbit: It isn't one certificate! There is a yet to be defined number, which should be provided together.

shamsasari: I'm afraid I don't understand, why do we need to check against more than just the root cert?

mnesbit: There are lots of reasons why we may end up with multiple roots, e.g. to let us merge networks. Also, we need some other certs for CRL and OCSP signing, etc in the truststore. Also, this breaks existing users config. Basically I can't see any reason for this check at all and the truststore was always available by an out of band mechanism e.g. email, or public corda website.

shamsasari: I can add a TODO to remove this file and instead have the truststore delivered out of band. The reason we didn't go down that route was because of the password issue.

mnesbit: I am going to give up arguing at this point, but seriously I think this is a bad idea and not consistent with the policy used by Java, or other trusted systems. It also appears to be trying to deal with an a threat model that we don't cover, nor have any way of protecting against i.e. a compromised machine.

al-r3: this doesn't prevent network mergers, this just requires to know in advance which doorman one registers with.

For the record I fear there is code somewhere else that assumes a single root and prevents mergers - untouched by this PR.

shamsasari: The threat model here, as far as I understand it, is MITM on the doorman registration. But I admit to be out of my depth here and @mikehearn will have to chip in.

shamsasari: Spoke with @al-r3 and he'll be changing this to be the truststore. This will remove the need for this config option, as you say.

shamsasari:
I have PR #2208 for updating the changelog, in case you're looking for that.

shamsasari:
I've updated this to the latest on master and the feature branch. I've added a TODO for moving to using the truststore the root cert check, which Al is already working on.

Back to you guys, I'd like to get some movement on this please.

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
```diff
@@ -0,0 +1,111 @@
...
+        } catch (e: Exception) {
+            logger.warn("Exception parsing NodeInfo from file. $file", e)
+            null
+        }
+    }
+
+    // We need to to set serialization env, because generation of parameters is run from Cordform.
+    // KryoServerSerializationScheme is not accessible from nodeapi.
+    private fun initialiseSerialization() {
+        val context = if (java.lang.Boolean.getBoolean("net.corda.testing.amqp.enable")) AMQP_P2P_CONTEXT else KRYO_P2P_CONTEXT
```
mikehearn: Can this line be removed now? AMQP is on by default now.

shamsasari: I'm going through now to remove any dependencies to Kryo we had.

PR #2180, opened 2017-12-05
CORDA-831: Add roles to X509 certificates

rnicoll:


core/src/main/kotlin/net/corda/core/CordaOID.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.core
+
+import org.bouncycastle.asn1.ASN1ObjectIdentifier
+
+/**
+ * OIDs used for the Corda platform. Entries MUST NOT be removed from this file; if an OID is incorrectly assigned it
+ * should be marked deprecated.
+ */
+object CordaOID {
```
mnesbit: Perhaps a comment to explain this construction

core/src/main/kotlin/net/corda/core/crypto/IdentityRoleExtension.kt
```diff
@@ -0,0 +1,56 @@
...
+import java.security.cert.X509Certificate
+
+/**
+ * X.509 certificate extension for the Corda role the certificate represents.
+ */
+class IdentityRoleExtension : ASN1Encodable {
+    companion object {
+        fun getInstance(obj: Any?): IdentityRoleExtension? {
+            return when (obj) {
+                null -> null
```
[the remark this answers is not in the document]
rnicoll: So the theory was to copy the API design BouncyCastle uses, but happy to move away from that to something more Kotlin-ish.

core/src/main/kotlin/net/corda/core/crypto/IdentityRoleExtension.kt
```diff
@@ -0,0 +1,56 @@
...
+            return if (extensionData != null) {
+                val extensionString = DEROctetString.getInstance(extensionData)
+                IdentityRoleExtension.getInstance(extensionString.octets)
+            } else {
+                null
+            }
+        }
+    }
+
+    val role: Role
```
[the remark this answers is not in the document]
rnicoll: I think it's legacy from when I thought we'd have other fields in the extension, will check it works with a bare role and replace with that if so.

core/src/main/kotlin/net/corda/core/CordaOID.kt
rnicoll: Instead I've added the root and intermediate values, so they're built up in the code. Let me know if that works?

core/src/main/kotlin/net/corda/core/CordaOID.kt
```diff
@@ -0,0 +1,13 @@
...
```
shamsasari: Does this need to be public API?

rnicoll: The intent is to actually use this as a registry, as it's version controlled, so it's not just a public API, it's part of the specification.

core/src/main/kotlin/net/corda/core/CordaOID.kt
```diff
@@ -0,0 +1,13 @@
...
+    const val R3_ROOT = "1.3.6.1.4.1.50530"
+    const val CORDA_PLATFORM = R3_ROOT + ".1"
+    const val X509_EXTENSION_CORDA_ROLE = CORDA_PLATFORM + ".1"
```
shamsasari: Does this need to be public constant?

rnicoll: It's intentionally exposed so others can make these certificates with different tools, especially if they want to integrate with different HSM solutions than those we provide, but generally also just for integration into whatever PKI infrastructure customers have.

core/src/main/kotlin/net/corda/core/CordaOID.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.core
+
+import org.bouncycastle.asn1.ASN1ObjectIdentifier
```
kchalkias: this import is not required any more

node-api/src/main/kotlin/net/corda/nodeapi/internal/ServiceIdentityGenerator.kt
```diff
@@ -37,16 +39,20 @@ object ServiceIdentityGenerator {
         val notaryKey = CompositeKey.Builder().addKeys(keyPairs.map { it.public }).build(threshold)
 
         val caKeyStore = loadKeyStore(javaClass.classLoader.getResourceAsStream("certificates/cordadevcakeys.jks"), "cordacadevpass")
-        val issuer = caKeyStore.getCertificateAndKeyPair(X509Utilities.CORDA_INTERMEDIATE_CA, "cordacadevkeypass")
+        val intermediateCa = caKeyStore.getCertificateAndKeyPair(X509Utilities.CORDA_INTERMEDIATE_CA, "cordacadevkeypass")
+        // TODO: We should be saving the node CA key & cert
+        val nodeCaName = X500Name("O=${X509Utilities.CORDA_CLIENT_CA_CN},L=London,C=GB")
+        val nodeCaKeyPair = Crypto.generateKeyPair(signatureScheme = Crypto.ECDSA_SECP256K1_SHA256)
```
mnesbit: Do we mean SECP256K1 rather than SECP256R1 ? Kostas you are the expert!

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -0,0 +1,63 @@
...
+import java.security.cert.Certificate
+import java.security.cert.X509Certificate
+
+/**
+ * Describes the Corda role a certificate is used for. This is used both to verify the hierarchy of certificates is
+ * correct, and to determine which is the well known identity's certificate.
+ *
+ * @property parent the parent role of this role - must match exactly for the certificate hierarchy to be valid for use
+ * in Corda. Null indicates the parent certificate has no role.
+ * @property isIdentity true if the role is valid for use as a [Party] identity, false otherwise (the role is Corda
```
kchalkias: [Party] is not visible

core/src/main/kotlin/net/corda/core/CordaOID.kt
kchalkias: Is the R3_Root a random OID or we have registered it? if it's the official one, let's better add a comment or link.

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -0,0 +1,63 @@
...
+ * infrastructure of some kind).
+ */
+// The order of the entries in the enum MUST NOT be changed, as their ordinality is used as an identifier.
```
kchalkias: Maybe add a comment that we use 1-based numbering (so ordinal zero = 1 as CertRole).

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -0,0 +1,63 @@
...
+// TODO: Link to the specification once it has a permanent URL
+enum class CertRole(val parent: CertRole?, val isIdentity: Boolean) : ASN1Encodable {
+    /**
+     * A Doorman (intermediate CA of some kind).
+     */
+    INTERMEDIATE_CA(null, false),
```
kchalkias: Why we didn't start from Root_CA? is this the reason that we use one-based numbering (conceptually, Root_CA is zero?)

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
mnesbit: I do agree that we should not restrict our Root Cert in any way, just in case we need to make new top level roles.

node-api/src/main/kotlin/net/corda/nodeapi/internal/ServiceIdentityGenerator.kt
kchalkias: Matthew is right, we use R1 as Corda TLS default, although it's true K1 is also TLS-compatible.

kchalkias: use `X509Utilities.DEFAULT_TLS_SIGNATURE_SCHEME` Vs `Crypto.ECDSA_SECP256K1_SHA256`

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -0,0 +1,63 @@
...
+    TLS(NODE_CA, false),
+    /** A well known (publicly visible) identity */
+    WELL_KNOWN_IDENTITY(NODE_CA, true),
+    /** A confidential (limited visibility) identity */
+    CONFIDENTIAL_IDENTITY(WELL_KNOWN_IDENTITY, true);
+
+    companion object {
+        fun getInstance(id: ASN1Integer): CertRole {
+            val ordinal = id.positiveValue.toInt() - 1
+            return CertRole.values().get(ordinal)
```
kchalkias: return CertRole.values()[ordinal]

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -0,0 +1,63 @@
...
+    NODE_CA(INTERMEDIATE_CA, false),
+    /** The transport layer security certificate */
+    TLS(NODE_CA, false),
+    /** A well known (publicly visible) identity */
+    WELL_KNOWN_IDENTITY(NODE_CA, true),
+    /** A confidential (limited visibility) identity */
+    CONFIDENTIAL_IDENTITY(WELL_KNOWN_IDENTITY, true);
+
+    companion object {
+        fun getInstance(id: ASN1Integer): CertRole {
```
kchalkias: Maybe it's more secure to test if the value is positive as -1 = 1. This doesn't seem to cause any issues atm, but let's be proactive.

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -0,0 +1,63 @@
...
+            val ordinal = id.positiveValue.toInt() - 1
```
kchalkias: This might cause `ArrayIndexOutOfBoundsException` the reason is that BigInteger.positiveValue doesn't work as expected for Int overflowed values (try setting an id = Int.MAX_VALUE + something)

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
shamsasari: Can you also cache `CertRole.values()` as it does an array copy each time.

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
shamsasari: I'm actually thinking that using the ordinal isn't a good idea. Better to pass in the int to each enum value.

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
mnesbit: I think we need a SERVICE_IDENTITY role, which should NOT be a CA and which will be issued from the doorman's intermediate certificate. This does seem to be what we do with CompositeKey certificates for Notaries right now.

rnicoll: We don't make any claims about what's going on above the `INTERMEDIATE_CA`, although maybe as we're adding more flexible role parenting as part of changes to support service identities, maybe we should.

The numbering is because other ASN.1 enums seem to count from 1.

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
rnicoll: An int means we then have to resolve lookup ourselves, we'd probably just end up back where we are :-/

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -1,41 +1,63 @@
...
+        private var cachedRoles: Array<CertRole>? = null
+        @Throws(IllegalArgumentException::class)
         fun getInstance(id: ASN1Integer): CertRole {
-            val ordinal = id.positiveValue.toInt() - 1
-            return CertRole.values().get(ordinal)
+            if (cachedRoles == null) {
+                cachedRoles = CertRole.values()
+            }
+            val idVal = id.value
+            require(idVal.compareTo(BigInteger.ZERO) > 0)
```
kchalkias: What about requiring `<=BigInteger.valueOf(cachedRoles.size.toLong())` as well to avoid catching Exceptions?
Also add `{ "Invalid role ID" }` for this `require`

rnicoll: Not sure why we're avoiding catching exceptions? It means we're essentially testing the same constraint in two places, doesn't it? I mean, catching is slightly more expensive than many other things, but I'm not sure it's problematically so, and I'd rather include handling the exception either way.

kchalkias: However, >0 and <=enum.size sound like a requirement

rnicoll:
I've merged the service identity and network map identity together (because network map is a service), let me know if there's any objections to that, please.

mnesbit:
I don't think network map is a service at all I am afraid. The network map nowadays is a trusted signature over the directory file. It isn't a thing that can be looked up in the actual network map data itself. Service identities are actually used to identify clustered nodes to communicate with in flows.

node-api/src/main/kotlin/net/corda/nodeapi/internal/ServiceIdentityGenerator.kt
```diff
@@ -1,6 +1,7 @@
 package net.corda.nodeapi.internal
 
 import net.corda.core.crypto.CompositeKey
+import net.corda.core.crypto.Crypto
```
kchalkias: unused import

node-api/src/main/kotlin/net/corda/nodeapi/internal/ServiceIdentityGenerator.kt
```diff
@@ -9,6 +10,7 @@ import net.corda.core.internal.createDirectories
 import net.corda.core.internal.div
 import net.corda.core.utilities.trace
 import net.corda.nodeapi.internal.crypto.*
+import org.bouncycastle.asn1.x500.X500Name
```
kchalkias: ditto

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
```diff
@@ -322,54 +326,78 @@ class X509CertificateFactory {
...
-enum class CertificateType(val keyUsage: KeyUsage, vararg val purposes: KeyPurposeId, val isCA: Boolean) {
+enum class CertificateType(val keyUsage: KeyUsage, vararg val purposes: KeyPurposeId, val isCA: Boolean, val role: CertRole?) {
     ROOT_CA(
             KeyUsage(KeyUsage.digitalSignature or KeyUsage.keyCertSign or KeyUsage.cRLSign),
             KeyPurposeId.id_kp_serverAuth,
             KeyPurposeId.id_kp_clientAuth,
             KeyPurposeId.anyExtendedKeyUsage,
-            isCA = true
+            isCA = true,
+            role = null
```
kchalkias: I wonder if this is the only entity with `role = null`, why we didn't add a role here as well and get rid of null roles.

core/src/main/kotlin/net/corda/core/identity/PartyAndCertificate.kt
```diff
@@ -38,6 +41,23 @@ class PartyAndCertificate(val certPath: CertPath) {
     fun verify(trustAnchor: TrustAnchor): PKIXCertPathValidatorResult {
         val parameters = PKIXParameters(setOf(trustAnchor)).apply { isRevocationEnabled = false }
         val validator = CertPathValidator.getInstance("PKIX")
-        return validator.validate(certPath, parameters) as PKIXCertPathValidatorResult
+        val result = validator.validate(certPath, parameters) as PKIXCertPathValidatorResult
+        // Apply Corda-specific validity rules to the chain
+        var parentRole: CertRole? = CertRole.extract(result.trustAnchor.trustedCert)
```
kchalkias: If all roles are `null` this will pass.  Maybe we should have a test for this function to check various scenarios (e.g. all `null` except for the leaf cert)

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
rnicoll: We're intentionally leaving the space above the intermediate CA undefined, so further layers of intermediate CAs could in theory be added later. I think the first draft actually didn't specify a role on the intermediate CA, not sure why that changed.

docs/source/changelog.rst
```diff
@@ -6,8 +6,16 @@ from the previous milestone release.
...
 ----------
-* Support for external user credentials data source and password encryption [CORDA-827].
 
+* X.509 certificates now have an extension that specifies the Corda role the certificate is used for, and the role
+  hierarchy is now enforced in the validation code. See ``net.corda.core.internal.CertRole`` for the current implementation
+  until final documentation is prepared. Certificates at ``NODE_CA``, ``WELL_KNOWN_SERVICE_IDENTITY`` and above must
+  only ever by issued by network services and are therefore issuance constraints are not relevant to end users.
+  The ``TLS``, ``WELL_KNOWN_LEGAL_IDENTITY`` roles must be issued by the ``NODE_CA`` certificate issued by the
+  Doorman, and ``CONFIDENTIAL_IDENTITY`` certificates must be issued from a ``WELL_KNOWN_LEGAL_IDENTITY`` certificate.
+  For a detailed specification of the extension please see https://r3-cev.atlassian.net/wiki/spaces/AWG/pages/156860572/Certificate+identity+type+extension
```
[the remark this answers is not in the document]
rnicoll: I thought the AWG was (there was an issue with some other parts being made public, at one point). I'll find out if it's the whole space is protected or just some pages, and get that fixed. Worst case I'll add it to the docs for Corda

core/src/main/kotlin/net/corda/core/identity/PartyAndCertificate.kt
rnicoll: An all-null path is valid, it's just not a valid identity (which is why it fails when constructing the `PartyAndCertificate`). I'll add comments to make it clear this is intentional and that the two parts are tested independently, though.

core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -0,0 +1,87 @@
...
+    /** Transport layer security certificate for a node */
+    TLS(setOf(NODE_CA), false, false),
+    /** Well known (publicly visible) identity of a legal entity */
+    LEGAL_IDENTITY(setOf(INTERMEDIATE_CA, NODE_CA), true, true),
+    /** Confidential (limited visibility) identity */
+    CONFIDENTIAL_LEGAL_IDENTITY(setOf(LEGAL_IDENTITY), true, false);
+
+    companion object {
+        private var cachedRoles: Array<CertRole>? = null
+        @Throws(IllegalArgumentException::class)
```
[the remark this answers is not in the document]
rnicoll: I keep forgetting IllegalArgumentException extends RuntimeException. Fixing.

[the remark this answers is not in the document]
shamsasari: Please don't annotate `RuntimeException` or `IllegalArgumentException`. It is not common in Java to add the `throws IllegalArgumentException` declaration on methods as runtime exceptions are unchecked. Adding a KDoc is sufficient for documentation.

Further as this is an internal API there isn't any Java code calling this, so there's no benefit even if this method was throwing non-runtime exceptions.

core/src/main/kotlin/net/corda/core/identity/PartyAndCertificate.kt
rnicoll: I've added a test covering this specifically, now

rnicoll:
Addressed non-docs changes, heading into the office now and will tackle the rest once in.

rnicoll:
I think that's everything from Friday tackled. Can we merge this as-is, obviously unless there's anything critical, as it's blocking @al-r3 , and revise in a further PR?

kchalkias:
Cool, it works with `openssl` in both linux and windows by adding the new extension field in cfg or cfn files
```
[ corda_role ]
1.3.6.1.4.1.50530.1.1=ASN1:INTEGER:2
```

shamsasari:
Will be doing that @kchalkias ?

PR #2185, opened 2017-12-05
Rename NetworkisRegistrationHelperTest to NetworkRegistrationHelperTest

al-r3:


rnicoll:
@patrickkuo I'm trying to remember if there was a reason for the old name, or this has just been hit by a refactor going awry somewhere?

patrickkuo:
good spot, no idea how this happened

PR #2187, opened 2017-12-05
Wire part of network parameters

kasiastreich:
Added handling of the situation when network map is advertising different parameters hash - for now we stop the node.
Added wiring of maxMessageSize (in fact we already had that, now it's just taken from network parameters), also handling of epoch on node startup.

### DIFF
--- a/docs/source/network-map.rst
+++ b/docs/source/network-map.rst
@@ -2,7 +2,7 @@ Network Map
 ===========
 
 The network map stores a collection of ``NodeInfo`` objects, each representing another node with which the node can interact.
-There two sources from which a Corda node can retrieve ``NodeInfo`` objects:
+There are two sources from which a Corda node can retrieve ``NodeInfo`` objects:
 
 1. the REST protocol with the network map service, which also provides a publishing API,
 
@@ -25,7 +25,7 @@ Node side network map update protocol:
 
 * The Corda node will query the network map service periodically according to the ``Expires`` attribute in the HTTP header.
 
-* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameters hashes.
+* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameters hash.
 
 * The node updates its local copy of ``NodeInfos`` if it is different from the newly downloaded ``NetworkMap``.
 
@@ -36,7 +36,7 @@ Network Map service REST API:
 +================+===================================+========================================================================================================================================================+
 | POST           | /api/network-map/publish          | Publish new ``NodeInfo`` to the network map service, the legal identity in ``NodeInfo`` must match with the identity registered with the doorman.      |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /api/network-map                  | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and NetworkParameters hash.                       |
+| GET            | /api/network-map                  | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and ``NetworkParameters`` hash.                   |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
 | GET            | /api/network-map/node-info/{hash} | Retrieve ``NodeInfo`` object with the same hash.                                                                                                       |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
@@ -55,3 +55,19 @@ Nodes expect to find a serialized ``SignedData<NodeInfo>`` object, the same obje
 Whenever a node starts it writes on disk a file containing its own ``NodeInfo``, this file is called ``nodeInfo-XXX`` where ``XXX`` is a long string.
 
 Hence if an operator wants node A to see node B they can pick B's ``NodeInfo`` file from B base directory and drop it into A's ``additional-node-infos`` directory.
+
+
+Network parameters
+------------------
+Network parameters are constants that every node participating in the network needs to agree on and use for interop purposes.
+The structure is distributed as a file containing serialized ``SignedData<NetworkParameters>`` with a signature from the
+network operator. Network map advertises the hash of currently used network parameters.
+The ``NetworkParameters`` structure contains:
+ * ``minimumPlatformVersion`` -  minimum version of Corda platform that is required for nodes in the network.
+ * ``notaries`` - list of well known and trusted notary identities with information on validation type.
+ * ``maxMessageSize`` - maximum P2P message size sent over the wire in bytes.
+ * ``maxTransactionSize`` - maximum permitted transaction size in bytes.
+ * ``modifiedTime``
+ * ``epoch`` - version number of the network parameters. Starting from 1, this will always increment on each new set of parameters.
+
+The set of parameters is still under development and we may find the need to add additional fields.
--- a/docs/source/network-map.rst
+++ b/docs/source/network-map.rst
@@ -34,13 +34,13 @@ Network Map service REST API:
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
 | Request method | Path                              | Description                                                                                                                                            |
 +================+===================================+========================================================================================================================================================+
-| POST           | /api/network-map/publish          | Publish new ``NodeInfo`` to the network map service, the legal identity in ``NodeInfo`` must match with the identity registered with the doorman.      |
+| POST           | /network-map/publish              | Publish new ``NodeInfo`` to the network map service, the legal identity in ``NodeInfo`` must match with the identity registered with the doorman.      |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /api/network-map                  | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and ``NetworkParameters`` hash.                   |
+| GET            | /network-map                      | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and ``NetworkParameters`` hash.                   |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /api/network-map/node-info/{hash} | Retrieve ``NodeInfo`` object with the same hash.                                                                                                       |
+| GET            | /network-map/node-info/{hash}     | Retrieve ``NodeInfo`` object with the same hash.                                                                                                       |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /api/network-map/parameters/{hash}| Retrieve ``NetworkParameters`` object with the same hash.                                                                                              |
+| GET            | /network-map/parameters/{hash}    | Retrieve ``NetworkParameters`` object with the same hash.                                                                                              |
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
 
 TODO: Access control of the network map will be added in the future.
--- a/docs/source/network-map.rst
+++ b/docs/source/network-map.rst
@@ -25,7 +25,18 @@ Node side network map update protocol:
 
 * The Corda node will query the network map service periodically according to the ``Expires`` attribute in the HTTP header.
 
-* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameters hash.
+* The network map service returns a signed ``NetworkMap`` object which looks as follows:
+
+.. container:: codeset
+
+   .. sourcecode:: kotlin
+
+        data class NetworkMap {
+            val nodeInfoHashes: List<SecureHash>,
+            val networkParametersHash: SecureHash
+        }
+
+The object contains list of node info hashes and hash of the network parameters data structure (without the signatures).
 
 * The node updates its local copy of ``NodeInfos`` if it is different from the newly downloaded ``NetworkMap``.
 
@@ -60,14 +71,14 @@ Hence if an operator wants node A to see node B they can pick B's ``NodeInfo`` f
 Network parameters
 ------------------
 Network parameters are constants that every node participating in the network needs to agree on and use for interop purposes.
-The structure is distributed as a file containing serialized ``SignedData<NetworkParameters>`` with a signature from the
-network operator. Network map advertises the hash of currently used network parameters.
+The structure is distributed as a file containing serialized ``SignedData<NetworkParameters>`` with a signature from
+a sub-key of the compatibility zone root cert. Network map advertises the hash of currently used network parameters.
 The ``NetworkParameters`` structure contains:
  * ``minimumPlatformVersion`` -  minimum version of Corda platform that is required for nodes in the network.
  * ``notaries`` - list of well known and trusted notary identities with information on validation type.
  * ``maxMessageSize`` - maximum P2P message size sent over the wire in bytes.
  * ``maxTransactionSize`` - maximum permitted transaction size in bytes.
- * ``modifiedTime``
+ * ``modifiedTime`` - the time the network parameters were created by the CZ operator.
  * ``epoch`` - version number of the network parameters. Starting from 1, this will always increment on each new set of parameters.
 
 The set of parameters is still under development and we may find the need to add additional fields.
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
@@ -13,6 +13,7 @@ import java.security.cert.X509Certificate
 import java.time.Duration
 import java.time.Instant
 
+const val NETWORK_PARAM_FILE_PREFIX = "network-parameters"
 // TODO: Need more discussion on rather we should move this class out of internal.
 /**
  * Data class containing hash of [NetworkParameters] and network participant's [NodeInfo] hashes.
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
@@ -21,21 +21,21 @@ import java.time.Instant
 data class NetworkMap(val nodeInfoHashes: List<SecureHash>, val networkParameterHash: SecureHash)
 
 /**
- * @property minimumPlatformVersion
- * @property notaries
- * @property eventHorizon
+ * @property minimumPlatformVersion Minimal version of Corda platform that is required for nodes in the network.
+ * @property notaries List of well known and trusted notary identities with information on validation type.
  * @property maxMessageSize Maximum P2P message sent over the wire in bytes.
  * @property maxTransactionSize Maximum permitted transaction size in bytes.
  * @property modifiedTime
  * @property epoch Version number of the network parameters. Starting from 1, this will always increment on each new set
  * of parameters.
  */
-// TODO Wire up the parameters
+// TODO Add eventHorizon - how many days a node can be offline before being automatically ejected from the network.
+//  It needs separate design.
+// TODO Currently both maxTransactionSize and modifiedTime are not wired.
 @CordaSerializable
 data class NetworkParameters(
         val minimumPlatformVersion: Int,
         val notaries: List<NotaryInfo>,
-        val eventHorizon: Duration,
         val maxMessageSize: Int,
         val maxTransactionSize: Int,
         val modifiedTime: Instant,
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
@@ -10,7 +10,6 @@ import net.corda.core.serialization.deserialize
 import java.security.SignatureException
 import java.security.cert.CertPathValidatorException
 import java.security.cert.X509Certificate
-import java.time.Duration
 import java.time.Instant
 
 const val NETWORK_PARAM_FILE_PREFIX = "network-parameters"
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
@@ -22,7 +22,7 @@ const val NETWORK_PARAM_FILE_PREFIX = "network-parameters"
 data class NetworkMap(val nodeInfoHashes: List<SecureHash>, val networkParameterHash: SecureHash)
 
 /**
- * @property minimumPlatformVersion Minimal version of Corda platform that is required for nodes in the network.
+ * @property minimumPlatformVersion Minimum version of Corda platform that is required for nodes in the network.
  * @property notaries List of well known and trusted notary identities with information on validation type.
  * @property maxMessageSize Maximum P2P message sent over the wire in bytes.
  * @property maxTransactionSize Maximum permitted transaction size in bytes.
@@ -46,6 +46,8 @@ data class NetworkParameters(
         require(minimumPlatformVersion > 0) { "minimumPlatformVersion must be at least 1" }
         require(notaries.distinctBy { it.identity } == notaries) { "Duplicate notary identities" }
         require(epoch > 0) { "epoch must be at least 1" }
+        require(maxMessageSize > 0 ) { "maxMessageSize must be at least 1" }
+        require(maxTransactionSize > 0 ) { "maxTransactionSize must be at least 1" }
     }
 }
 
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
@@ -45,8 +45,8 @@ data class NetworkParameters(
         require(minimumPlatformVersion > 0) { "minimumPlatformVersion must be at least 1" }
         require(notaries.distinctBy { it.identity } == notaries) { "Duplicate notary identities" }
         require(epoch > 0) { "epoch must be at least 1" }
-        require(maxMessageSize > 0 ) { "maxMessageSize must be at least 1" }
-        require(maxTransactionSize > 0 ) { "maxTransactionSize must be at least 1" }
+        require(maxMessageSize > 0) { "maxMessageSize must be at least 1" }
+        require(maxTransactionSize > 0) { "maxTransactionSize must be at least 1" }
     }
 }
 
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
@@ -12,7 +12,7 @@ import java.security.cert.CertPathValidatorException
 import java.security.cert.X509Certificate
 import java.time.Instant
 
-const val NETWORK_PARAM_FILE_PREFIX = "network-parameters"
+const val NETWORK_PARAMS_FILE_NAME = "network-parameters"
 // TODO: Need more discussion on rather we should move this class out of internal.
 /**
  * Data class containing hash of [NetworkParameters] and network participant's [NodeInfo] hashes.
@@ -31,7 +31,7 @@ data class NetworkMap(val nodeInfoHashes: List<SecureHash>, val networkParameter
  */
 // TODO Add eventHorizon - how many days a node can be offline before being automatically ejected from the network.
 //  It needs separate design.
-// TODO Currently both maxTransactionSize and modifiedTime are not wired.
+// TODO Currently maxTransactionSize is not wired.
 @CordaSerializable
 data class NetworkParameters(
         val minimumPlatformVersion: Int,
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersCopier.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersCopier.kt
@@ -22,9 +22,9 @@ class NetworkParametersCopier(networkParameters: NetworkParameters) {
         SignedData(serialize, signature).serialize()
     }
 
-    fun install(dir: Path) {
+    fun install(dir: Path, paramsFile: String = "network-parameters") {
         try {
-            serializedNetworkParameters.open().copyTo(dir / "network-parameters")
+            serializedNetworkParameters.open().copyTo(dir / paramsFile)
         } catch (e: FileAlreadyExistsException) {
             // Leave the file untouched if it already exists
         }
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersCopier.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersCopier.kt
@@ -22,7 +22,7 @@ class NetworkParametersCopier(networkParameters: NetworkParameters) {
         SignedData(serialize, signature).serialize()
     }
 
-    fun install(dir: Path, paramsFile: String = "network-parameters") {
+    fun install(dir: Path, paramsFile: String = NETWORK_PARAM_FILE_PREFIX) {
         try {
             serializedNetworkParameters.open().copyTo(dir / paramsFile)
         } catch (e: FileAlreadyExistsException) {
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersCopier.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersCopier.kt
@@ -6,7 +6,6 @@ import net.corda.core.crypto.sign
 import net.corda.core.internal.copyTo
 import net.corda.core.internal.div
 import net.corda.core.serialization.serialize
-import net.corda.nodeapi.internal.NetworkParameters
 import java.math.BigInteger
 import java.nio.file.FileAlreadyExistsException
 import java.nio.file.Path
@@ -22,9 +21,9 @@ class NetworkParametersCopier(networkParameters: NetworkParameters) {
         SignedData(serialize, signature).serialize()
     }
 
-    fun install(dir: Path, paramsFile: String = NETWORK_PARAM_FILE_PREFIX) {
+    fun install(dir: Path) {
         try {
-            serializedNetworkParameters.open().copyTo(dir / paramsFile)
+            serializedNetworkParameters.open().copyTo(dir / NETWORK_PARAMS_FILE_NAME)
         } catch (e: FileAlreadyExistsException) {
             // Leave the file untouched if it already exists
         }
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
@@ -45,7 +45,7 @@ class NetworkParametersGenerator {
                     maxTransactionSize = 40000,
                     epoch = 1
             ))
-            nodesDirs.forEach { copier.install(it) }
+            nodesDirs.forEach(copier::install)
         } finally {
             _contextSerializationEnv.set(null)
         }
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
@@ -45,7 +45,7 @@ class NetworkParametersGenerator {
                     maxTransactionSize = 40000,
                     epoch = 1
             ))
-            nodesDirs.forEach(copier::install)
+            nodesDirs.forEach { copier.install(it) }
         } finally {
             _contextSerializationEnv.set(null)
         }
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
@@ -41,7 +41,6 @@ class NetworkParametersGenerator {
                     minimumPlatformVersion = 1,
                     notaries = notaryInfos,
                     modifiedTime = Instant.now(),
-                    eventHorizon = 10000.days,
                     maxMessageSize = 40000,
                     maxTransactionSize = 40000,
                     epoch = 1
--- a/node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
+++ b/node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
@@ -20,6 +20,7 @@ import org.junit.Test
 import java.nio.file.Path
 import kotlin.test.assertEquals
 import kotlin.test.assertFails
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 class NetworkParametersTest {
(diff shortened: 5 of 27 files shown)


node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -4,22 +4,27 @@ import net.corda.core.node.NodeInfo
...
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.Rule
 import org.junit.Test
 import java.net.URL
 
 class NetworkMapClientTest {
     private val portAllocation = PortAllocation.Incremental(10000)
+    @Rule
+    @JvmField
+    val testSerialization = SerializationEnvironmentRule(true)
```
al-r3: are the changes in this file needed at all?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -265,6 +268,13 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
      */
     protected abstract fun getRxIoScheduler(): Scheduler
 
+    // TODO NetworkParameters updates are not implemented yet. This handling is not ideal, because we simply shutdown node, which can cause some problems.
+    private fun handleNetworkParametersUpdate(newParametersHash: SecureHash) {
+        log.error("Network map is advertising different parametersHash than the ones node is currently using.\n" +
```
al-r3: The Network map is advertising a different parametersHash than the one this node is currently using.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +620,22 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
     }
 
     private fun readNetworkParameters() {
-        val file = configuration.baseDirectory / "network-parameters"
-        networkParameters = file.readAll().deserialize<SignedData<NetworkParameters>>().verified()
-        log.info(networkParameters.toString())
+        var latestEpoch = 0
+        var latestParams: NetworkParameters? = null
+        // Load network parameters with the latest known epoch.
+        for (paramFile in Files.list(configuration.baseDirectory)) {
```
al-r3: if you want you can make this more functional
Files.list(configuration.baseDirectory)
   .filter { "network-parameters" in  paramFile.toString()}
   .map { ..read/deserialize...}
   .maxBy{ it.epoch }

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +620,22 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+            if ("network-parameters" !in paramFile.toString())
```
al-r3: "network-parameters" should be a constant defined somewhere. (somewhere you can access it from Cordformation as well)

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +620,22 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+                continue
+            val params = paramFile.readAll().deserialize<SignedData<NetworkParameters>>().verified()
```
al-r3: If someone adds a file which for any reason has "network-parameters" in this will crash the node, we should just ignore un-parables files, or files for which verified() fails.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -96,6 +99,8 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
     private val executor = Executors.newSingleThreadScheduledExecutor(NamedThreadFactory("Network Map Updater Thread", Executors.defaultThreadFactory()))
     private var fileWatcherSubscription: Subscription? = null
 
+    val parametersUpdates: PublishSubject<SecureHash> = PublishSubject.create<SecureHash>()
```
al-r3: What's the intended use of this?
Currently it just emits the same value each time network map tried to update the parameters.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -53,7 +53,7 @@ class NetworkMapUpdaterTest {
 
         val scheduler = TestScheduler()
         val fileWatcher = NodeInfoWatcher(baseDir, scheduler)
-        val updater = NetworkMapUpdater(networkMapCache, fileWatcher, networkMapClient)
+        val updater = NetworkMapUpdater(networkMapCache, fileWatcher, networkMapClient, SecureHash.randomSHA256())
```
al-r3: For readability's sake add a var named NETWORK_PARAMETERS_HASH and pass that here.

testing/test-utils/src/main/kotlin/net/corda/testing/TestConstants.kt
```diff
@@ -88,3 +88,6 @@ object DummyCommandData : TypeOnlyCommandData()
 
 val DUMMY_IDENTITY_1: PartyAndCertificate get() = getTestPartyAndCertificate(DUMMY_PARTY)
 val DUMMY_PARTY: Party get() = Party(CordaX500Name(organisation = "Dummy", locality = "Madrid", country = "ES"), DUMMY_KEY_1.public)
+
+/** Maximum artemis message size. 10 MiB maximum allowed file size for attachments, including message headers. */
+val MAX_MESSAGE_SIZE: Int = 1048576
```
al-r3: can be const

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
```diff
@@ -47,6 +49,9 @@ private val portAllocation = PortAllocation.Incremental(13000)
  * Driver based tests for [NetworkRegistrationHelper]
  */
 class NetworkRegistrationHelperDriverTest {
+    @Rule
```
al-r3: same as above

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    fun tearDown() {
+        mockNet.stopNodes()
+    }
+
+    // Epoch tests
+    @Test
+    fun `choose highest epoch`() {
+        var alice = mockNet.createPartyNode(ALICE.name)
+        val aliceDirectory = alice.internals.configuration.baseDirectory
+        assertEquals(DUMMY_NOTARY.name, alice.services.networkMapCache.notaryIdentities[0].name)
```
al-r3: I imagine you're checking the notaries as a proxy for the epoch.
Add a comment saying that epoch isn't exposed by the node per se, and you're checking notary identities

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    // Notaries tests
+    @Test
+    fun `choosing notary not specified in network parameters will fail`() {
+        val fakeNotary = mockNet.createNode(MockNodeParameters(legalName = BOB_NAME, configOverrides = {
+            val notary = NotaryConfig(false)
+            doReturn(notary).whenever(it).notary}))
+        val fakeNotaryId = fakeNotary.info.chooseIdentity()
+        assert(fakeNotary.internals.configuration.notary != null)
```
al-r3: Use AssertFalse (or, even better AssertThat(...).isFalse() from assertj)

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        val alice = mockNet.createPartyNode(ALICE_NAME)
+        assert(fakeNotaryId !in alice.services.networkMapCache.notaryIdentities)
```
al-r3: same

node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
kasiastreich: Yes they are, I need to set serialization before running the server.

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
kasiastreich: The same.

al-r3: There's no server running here, can't you change Driver to initialize serialization soon than it does now?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
kasiastreich: It emits when mismatch happens, which can happen any time during node lifetime not only on startup, so I can run stop(). We will also later want to have nice handling of updates.

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
kasiastreich: And in line 65 what is initialized?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
al-r3: It doesn't seem much useful except for you to call stop, really.
In general what is the expected usage of this?

You might want to return an instance of a sealed class that has 2 implementations, one for rejection and one for acceptance of new parameters.
@shamsasari wdyt?

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -21,21 +22,21 @@ import java.time.Instant
 data class NetworkMap(val nodeInfoHashes: List<SecureHash>, val networkParameterHash: SecureHash)
 
 /**
- * @property minimumPlatformVersion
- * @property notaries
- * @property eventHorizon
+ * @property minimumPlatformVersion Minimal version of Corda platform that is required for nodes in the network.
```
shamsasari: Typo: Minimum

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -21,21 +22,21 @@ import java.time.Instant
...
  * @property modifiedTime
  * @property epoch Version number of the network parameters. Starting from 1, this will always increment on each new set
  * of parameters.
  */
-// TODO Wire up the parameters
+// TODO Add eventHorizon - how many days a node can be offline before being automatically ejected from the network.
+//  It needs separate design.
+// TODO Currently both maxTransactionSize and modifiedTime are not wired.
 @CordaSerializable
 data class NetworkParameters(
```
shamsasari: The network parameter docs need to actually be on the docsite. These classes are actually internal.

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -21,21 +22,21 @@ import java.time.Instant
...
         val minimumPlatformVersion: Int,
         val notaries: List<NotaryInfo>,
-        val eventHorizon: Duration,
         val maxMessageSize: Int,
```
shamsasari: Can you add checks in the `init` that these two max values are bigger than 0

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +622,16 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
     }
 
     private fun readNetworkParameters() {
-        val file = configuration.baseDirectory / "network-parameters"
-        networkParameters = file.readAll().deserialize<SignedData<NetworkParameters>>().verified()
-        log.info(networkParameters.toString())
+        // Load network parameters with the latest known epoch.
+        val latestParams = Files.list(configuration.baseDirectory).filter { NETWORK_PARAM_FILE_PREFIX in it.toString() }.toList()
```
shamsasari: Why do we need to worry about multiple network parameter files? Remember the design around this was changed. There will always be just one network params file.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -126,6 +131,12 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
             override fun run() {
                 val nextScheduleDelay = try {
                     val (networkMap, cacheTimeout) = networkMapClient.getNetworkMap()
+                    // TODO NetworkParameters updates are not implemented yet. Every mismatch should result in node shutdown.
+                    if (currentParametersHash != networkMap.networkParameterHash) {
+                        logger.error("Node is using parameters with hash: $currentParametersHash but network map is advertising: ${networkMap.networkParameterHash}.\n" +
+                                "Please update node to use correct network parameters file.\"")
```
shamsasari: If the network parameters hash from the network map doesn't match what we have then we have a bug somewhere. May be someone changed the node's network params file, may be there's a bug in the network map. There's not much we can do except log with this message and System.exit. There's no need to publish the network parameters hashes via rx.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +622,16 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+                .mapNotNull { try {
+                    it.readAll().deserialize<SignedData<NetworkParameters>>().verified()
+                } catch (t: Throwable) {
+                    null // We ignore all the files that are incorrect.
+                }}.maxBy { it.epoch }
+        checkNotNull(latestParams) { "Couldn't find network parameters file" }
```
shamsasari: The requirement to always have a network-parameters file in the node's directory was simply to unblock us and to work with the local, test deployments. On a real deployment the node on first startup won't be given the file, it will download it for the first time from the network map. So a missing network-parameters file is allowed as long as the CZ URL is provided in the config.

kasiastreich: ok, I will implement that. The design around that bit changed too many times.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
kasiastreich: Are you sure on that? The installation process can end up with having more than one in the directory. I thought that's why we wire epoch?

shamsasari: More than one directory? You mean node directories? Sure but all the node directories have the same network-parameters file.

kasiastreich: More than one network parameters file *in* the node's base directory.

shamsasari: No there will always be at most one file. I made this change from Mike's original design.

docs/source/network-map.rst
```diff
@@ -25,7 +25,7 @@ Node side network map update protocol:
 
 * The Corda node will query the network map service periodically according to the ``Expires`` attribute in the HTTP header.
 
-* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameters hashes.
+* The network map service returns a signed ``NetworkMap`` object, containing list of node info hashes and the network parameters hash.
```
shamsasari: and hash of the network parameters.

shamsasari: Slight complication here. The `NetworkMap` class and others are technically internal API - the classes are only needed by the node. Here in the docs we need to not refer to them as if they're public API. Rather we should describe the structures in a pseudo code format. 

```
NetworkMap {
  nodeInfoHashes: List<SecureHash>
  networkParametersHash: SecureHash
}
```

docs/source/network-map.rst
```diff
@@ -36,7 +36,7 @@ Network Map service REST API:
 +================+===================================+========================================================================================================================================================+
 | POST           | /api/network-map/publish          | Publish new ``NodeInfo`` to the network map service, the legal identity in ``NodeInfo`` must match with the identity registered with the doorman.      |
```
shamsasari: Please confirm with @patrickkuo if the endpoints still begin in /api

docs/source/network-map.rst
shamsasari: And explain that the hash is of the network parameters, and not with the signature + cert.

docs/source/network-map.rst
```diff
@@ -36,7 +36,7 @@ Network Map service REST API:
...
 +----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /api/network-map                  | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and NetworkParameters hash.                       |
+| GET            | /api/network-map                  | Retrieve ``NetworkMap`` from the server, the ``NetworkMap`` object contains list of node info hashes and ``NetworkParameters`` hash.                   |
```
shamsasari: You already describe above what the networkmap object is. You just need to say that it returns that.

docs/source/network-map.rst
```diff
@@ -55,3 +55,19 @@ Nodes expect to find a serialized ``SignedData<NodeInfo>`` object, the same obje
 Whenever a node starts it writes on disk a file containing its own ``NodeInfo``, this file is called ``nodeInfo-XXX`` where ``XXX`` is a long string.
 
 Hence if an operator wants node A to see node B they can pick B's ``NodeInfo`` file from B base directory and drop it into A's ``additional-node-infos`` directory.
+
+
+Network parameters
+------------------
+Network parameters are constants that every node participating in the network needs to agree on and use for interop purposes.
+The structure is distributed as a file containing serialized ``SignedData<NetworkParameters>`` with a signature from the
+network operator. Network map advertises the hash of currently used network parameters.
```
shamsasari: signature from a sub-key of the compatibility zone root cert.

docs/source/network-map.rst
```diff
@@ -55,3 +55,19 @@ Nodes expect to find a serialized ``SignedData<NodeInfo>`` object, the same obje
...
+The ``NetworkParameters`` structure contains:
+ * ``minimumPlatformVersion`` -  minimum version of Corda platform that is required for nodes in the network.
+ * ``notaries`` - list of well known and trusted notary identities with information on validation type.
+ * ``maxMessageSize`` - maximum P2P message size sent over the wire in bytes.
+ * ``maxTransactionSize`` - maximum permitted transaction size in bytes.
+ * ``modifiedTime``
```
shamsasari: The time the network parameters were created by the CZ operator.

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -10,9 +10,9 @@ import net.corda.core.serialization.deserialize
 import java.security.SignatureException
 import java.security.cert.CertPathValidatorException
 import java.security.cert.X509Certificate
-import java.time.Duration
 import java.time.Instant
 
+const val NETWORK_PARAM_FILE_PREFIX = "network-parameters"
```
shamsasari: NETWORK_PARAMS_FILE_NAME

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -21,21 +21,21 @@ import java.time.Instant
...
  * @property maxMessageSize Maximum P2P message sent over the wire in bytes.
  * @property maxTransactionSize Maximum permitted transaction size in bytes.
  * @property modifiedTime
  * @property epoch Version number of the network parameters. Starting from 1, this will always increment on each new set
  * of parameters.
  */
-// TODO Wire up the parameters
+// TODO Add eventHorizon - how many days a node can be offline before being automatically ejected from the network.
+//  It needs separate design.
+// TODO Currently both maxTransactionSize and modifiedTime are not wired.
```
shamsasari: There's no wiring up needed for modifiedTime (at least as I see it).

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersCopier.kt
```diff
@@ -22,9 +22,9 @@ class NetworkParametersCopier(networkParameters: NetworkParameters) {
         SignedData(serialize, signature).serialize()
     }
 
-    fun install(dir: Path) {
+    fun install(dir: Path, paramsFile: String = NETWORK_PARAM_FILE_PREFIX) {
```
shamsasari: No need to specify the filename as a parameter. AbstractNode expects it to be network-parameters so let's leave this out.

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkParametersGenerator.kt
```diff
@@ -42,12 +42,11 @@ class NetworkParametersGenerator {
                     minimumPlatformVersion = 1,
                     notaries = notaryInfos,
                     modifiedTime = Instant.now(),
-                    eventHorizon = 10000.days,
                     maxMessageSize = 40000,
                     maxTransactionSize = 40000,
                     epoch = 1
             ))
-            nodesDirs.forEach(copier::install)
+            nodesDirs.forEach { copier.install(it) }
```
shamsasari: What was wrong with the method reference?

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -1053,7 +1053,7 @@ class DriverDSL(
         val nodeInfoFilesCopier = if (compatibilityZone == null) nodeInfoFilesCopier else null
 
         nodeInfoFilesCopier?.addConfig(baseDirectory)
-        networkParameters!!.install(baseDirectory)
+        if (compatibilityZone == null) networkParameters!!.install(baseDirectory)
```
shamsasari: You don't need the null check: `networkParameters?.install(baseDirectory)`

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -818,7 +818,7 @@ class DriverDSL(
             notaryInfos += NotaryInfo(identity, type.validating)
         }
 
-        networkParameters = NetworkParametersCopier(testNetworkParameters(notaryInfos))
+        if (compatibilityZone == null) networkParameters = NetworkParametersCopier(testNetworkParameters(notaryInfos))
```
shamsasari: Can you revert this. `startCordformNodes` is a special method which is only used by DemoRunner. Can you add a check at the beginning of this method to make sure `compatibilityZone` is null.

patrickkuo:
endpoints for the network map are the same as the current ones but without `/api`

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +611,24 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
     }
 
     private fun readNetworkParameters() {
-        val file = configuration.baseDirectory / "network-parameters"
-        networkParameters = file.readAll().deserialize<SignedData<NetworkParameters>>().verified()
-        log.info(networkParameters.toString())
+        val files = Files.list(configuration.baseDirectory).filter { NETWORK_PARAM_FILE_PREFIX in it.toString() }.toList()
+        require(files.size <= 1) { "More than one network parameters file in the directory, please update only to the current one." }
```
shamsasari: Not required. There is only one network-parameters file

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +611,24 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+        val params = try {
+            // It's fine at this point if we don't have network parameters or have corrupted file, later we check if parameters can be downloaded from network map server.
+            files[0].readAll().deserialize<SignedData<NetworkParameters>>().verified()
+        } catch(t: Throwable) {
```
shamsasari: We should log a warn

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +611,24 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+            null
+        }
+        networkParameters = if (params != null) {
+            params
+        } else if (networkMapClient != null) {
+            log.info("Requesting network parameters from network map server...")
+            val (networkMap, _) = networkMapClient!!.getNetworkMap()
```
shamsasari: Can you make `networkMapClient` a lateinit var to avoid these !!

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +611,24 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+        val files = Files.list(configuration.baseDirectory).filter { NETWORK_PARAM_FILE_PREFIX in it.toString() }.toList()
+        require(files.size <= 1) { "More than one network parameters file in the directory, please update only to the current one." }
+        val params = try {
+            // It's fine at this point if we don't have network parameters or have corrupted file, later we check if parameters can be downloaded from network map server.
+            files[0].readAll().deserialize<SignedData<NetworkParameters>>().verified()
+        } catch(t: Throwable) {
+            null
+        }
+        networkParameters = if (params != null) {
+            params
```
shamsasari: `paramsFromFile`

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +611,24 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+        } else if (networkMapClient != null) {
+            log.info("Requesting network parameters from network map server...")
+            val (networkMap, _) = networkMapClient!!.getNetworkMap()
+            networkMapClient!!.getNetworkParameter(networkMap.networkParameterHash) ?: throw IllegalArgumentException("Failed loading network parameters from network map server")
```
shamsasari: We should also write to the file

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -45,6 +45,8 @@ data class NetworkParameters(
         require(minimumPlatformVersion > 0) { "minimumPlatformVersion must be at least 1" }
         require(notaries.distinctBy { it.identity } == notaries) { "Duplicate notary identities" }
         require(epoch > 0) { "epoch must be at least 1" }
+        require(maxMessageSize > 0 ) { "maxMessageSize must be at least 1" }
```
shamsasari: no whitespace before )

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,74 @@
+package net.corda.nodeapi.internal
```
shamsasari: This belongs in node

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import net.corda.testing.node.MockNodeParameters
+import net.corda.testing.node.MockServices
+import org.junit.After
+import org.junit.Test
+import java.nio.file.Path
+import kotlin.test.assertFails
+import kotlin.test.assertFalse
+
+class NetworkParametersTest {
+    private val mockNet= MockNetwork(
```
shamsasari: whitespace before =

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    // Minimum Platform Version tests
+    @Test
+    fun `node shutdowns when on lower platform version than network`() {
+        val alice = mockNet.createUnstartedNode(MockNodeParameters(legalName = ALICE.name, forcedID = 100, version = MockServices.MOCK_VERSION_INFO.copy(platformVersion = 1)))
+        val aliceDirectory = mockNet.baseDirectory(100)
+        val netParams = testNetworkParameters(
+                notaries = listOf(NotaryInfo(mockNet.defaultNotaryIdentity, true)),
+                minimumPlatformVersion = 2)
+        dropParametersToDir(aliceDirectory, netParams)
+        assertFails { alice.start() }
```
shamsasari: Check that the exception message contains the words "platform version" or similar to prevent against false-positives if the node fails for other reasons. Use `assertThatThrownBy`, which has a `hasMessageContaining` assertion method.

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    // Notaries tests
+    @Test
+    fun `choosing notary not specified in network parameters will fail`() {
+        val fakeNotary = mockNet.createNode(MockNodeParameters(legalName = BOB_NAME, configOverrides = {
+            val notary = NotaryConfig(false)
+            doReturn(notary).whenever(it).notary}))
+        val fakeNotaryId = fakeNotary.info.chooseIdentity()
+        assertFalse(fakeNotary.internals.configuration.notary == null)
```
shamsasari: Not really necessary I think. It's clear you've just specified a "notary" config just above.

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        val alice = mockNet.createPartyNode(ALICE_NAME)
+        assertFalse(fakeNotaryId in alice.services.networkMapCache.notaryIdentities)
+        assertFails {
```
shamsasari: I think the assertion on the network map cache is sufficient.

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    // Notaries tests
+    @Test
+    fun `choosing notary not specified in network parameters will fail`() {
+        val fakeNotary = mockNet.createNode(MockNodeParameters(legalName = BOB_NAME, configOverrides = {
+            val notary = NotaryConfig(false)
+            doReturn(notary).whenever(it).notary}))
+        val fakeNotaryId = fakeNotary.info.chooseIdentity()
+        assertFalse(fakeNotary.internals.configuration.notary == null)
+        val alice = mockNet.createPartyNode(ALICE_NAME)
+        assertFalse(fakeNotaryId in alice.services.networkMapCache.notaryIdentities)
```
shamsasari: `assertThat(alice.services.networkMapCache.notaryIdentities).doesNotContain(fakeNotaryId)`

node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapTest.kt
```diff
@@ -34,9 +48,19 @@ class NetworkMapTest {
         networkMapServer.close()
     }
 
+    @Test
+    fun `node corectly downloads and saves network parameters file on startup`() {
+        internalDriver(portAllocation = portAllocation, compatibilityZone = compatibilityZone, initialiseSerialization = false) {
+            val aliceDir = baseDirectory(ALICE.name)
+            startNode(providedName = ALICE.name).getOrThrow()
+            val networkParameters = Files.list(aliceDir).filter { NETWORK_PARAMS_FILE_NAME == it.fileName.toString() }.toList()[0]
```
shamsasari: `single { it.fileName.toString() == NETWORK_PARAMS_FILE_NAME }`

node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapTest.kt
```diff
@@ -34,9 +48,19 @@ class NetworkMapTest {
...
+                    .readAll().deserialize<SignedData<NetworkParameters>>().verified()
+            assertEquals(NetworkMapServer.stubNetworkParameter, networkParameters)
+        }
+    }
```
shamsasari: line separate to next test

node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapTest.kt
```diff
@@ -34,9 +48,19 @@ class NetworkMapTest {
         networkMapServer.close()
     }
 
+    @Test
+    fun `node corectly downloads and saves network parameters file on startup`() {
```
shamsasari: typo: correctly

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -136,11 +138,7 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
     protected lateinit var network: MessagingService
     protected val runOnStop = ArrayList<() -> Any?>()
     protected val _nodeReadyFuture = openFuture<Unit>()
-    protected val networkMapClient: NetworkMapClient? by lazy {
-        configuration.compatibilityZoneURL?.let {
-            NetworkMapClient(it, services.identityService.trustRoot)
-        }
-    }
+    protected lateinit var networkMapClient: NetworkMapClient
```
shamsasari: Does this need to be protected?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -192,9 +190,12 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
         log.info("Node starting up ...")
         initCertificate()
         val (keyPairs, info) = initNodeInfo()
-        readNetworkParameters()
         val schemaService = NodeSchemaService(cordappLoader.cordappSchemas)
         val identityService = makeIdentityService(info)
+        if (configuration.compatibilityZoneURL != null) {
+            networkMapClient = NetworkMapClient(configuration.compatibilityZoneURL!!, identityService.trustRoot)
```
shamsasari: Looks like `networkMapClient` is only initialised if there's a CZ URL. If so `networkMapClient` should be nullable and not lateinit

kasiastreich: You asked me to make it lateinit ;)

shamsasari: To avoid the !!, can you do something like
```
networkMapClient = configuration.compatibilityZoneURL?.let { NetworkMapClient(it, identityService.trustRoot) }
```

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -231,10 +232,10 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
             startShell(rpcOps)
             Pair(StartedNodeImpl(this, _services, info, checkpointStorage, smm, attachments, network, database, rpcOps, flowStarter, notaryService), schedulerService)
         }
-
         val networkMapUpdater = NetworkMapUpdater(services.networkMapCache,
                 NodeInfoWatcher(configuration.baseDirectory, getRxIoScheduler(), Duration.ofMillis(configuration.additionalNodeInfoPollingFrequencyMsec)),
-                networkMapClient)
+                configuration.compatibilityZoneURL?.let { networkMapClient },
```
shamsasari: This won't be needed if networkMapClient is nullable

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +611,27 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
 
     private fun readNetworkParameters() {
-        val file = configuration.baseDirectory / "network-parameters"
-        networkParameters = file.readAll().deserialize<SignedData<NetworkParameters>>().verified()
-        log.info(networkParameters.toString())
+        val files = Files.list(configuration.baseDirectory).filter { NETWORK_PARAMS_FILE_NAME == it.fileName.toString() }.toList()
+        val paramsFromFile = try {
+            // It's fine at this point if we don't have network parameters or have corrupted file, later we check if parameters can be downloaded from network map server.
+            files[0].readAll().deserialize<SignedData<NetworkParameters>>().verified()
+        } catch(t: Throwable) {
```
shamsasari: Let's just catch `Exception`. Throwable is more serious issues which we probably want to let propagate.

shamsasari: space before (

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -610,9 +611,27 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+            log.warn("Couldn't find correct network parameters file in the base directory")
+            null
+        }
+        networkParameters = if (paramsFromFile != null) {
+            paramsFromFile
+        } else if (configuration.compatibilityZoneURL != null) {
+            log.info("Requesting network parameters from network map server...")
+            val (networkMap, _) = networkMapClient.getNetworkMap()
+            val paramsFromMap = networkMapClient.getNetworkParameter(networkMap.networkParameterHash) ?: throw IllegalArgumentException("Failed loading network parameters from network map server")
```
shamsasari: `signedParams`

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -33,6 +33,9 @@ class NetworkMapUpdaterTest {
     val testSerialization = SerializationEnvironmentRule(true)
     private val jimFs = Jimfs.newFileSystem(Configuration.unix())
     private val baseDir = jimFs.getPath("/node")
+    companion object {
```
shamsasari: companion object is usually defined at the top of the class.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
shamsasari: I was probably responding to the code as it was back then. Right now it's clear lateinit doesn't make sense.

kasiastreich: This part didn't change. I will revert it back to how it was.

node-api/src/test/kotlin/net/corda/nodeapi/internal/NetworkParametersTest.kt
kasiastreich: It's to show that you cannot run the flow with notary that is not in network parameters (not only that you don't have it in cache).

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
kasiastreich: It's used in `net.corda.node.internal.Node#tryDetectIfNotPublicHost`.

shamsasari:
Please rebase this onto master. The feature branch will be deleted.

node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -21,21 +21,21 @@ import java.time.Instant
...
  * @property maxMessageSize Maximum P2P message sent over the wire in bytes.
  * @property maxTransactionSize Maximum permitted transaction size in bytes.
  * @property modifiedTime
  * @property epoch Version number of the network parameters. Starting from 1, this will always increment on each new set
  * of parameters.
  */
-// TODO Wire up the parameters
+// TODO Add eventHorizon - how many days a node can be offline before being automatically ejected from the network.
+//  It needs separate design.
+// TODO Currently maxTransactionSize is not wired.
```
shamsasari: Can you create a JIRA ticket for this.

PR #2188, opened 2017-12-05
Moved the CZ URL and node registration logic of the driver to be more…

shamsasari:
… internal, not available through the standard driver call, as these are not testing features for an app dev.

Also cleanup up some of the related tests.

### DIFF
--- a/node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
@@ -4,85 +4,96 @@ import net.corda.core.node.NodeInfo
 import net.corda.core.utilities.seconds
 import net.corda.testing.ALICE
 import net.corda.testing.BOB
+import net.corda.testing.driver.CompatibilityZoneParams
 import net.corda.testing.driver.NodeHandle
 import net.corda.testing.driver.PortAllocation
-import net.corda.testing.driver.driver
+import net.corda.testing.driver.internalDriver
 import net.corda.testing.node.network.NetworkMapServer
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.After
+import org.junit.Before
 import org.junit.Test
 import java.net.URL
 
+// TODO There is a unit test class with the same name. Rename this to something else.
 class NetworkMapClientTest {
+    private val cacheTimeout = 1.seconds
     private val portAllocation = PortAllocation.Incremental(10000)
+
+    private lateinit var networkMapServer: NetworkMapServer
+    private lateinit var compatibilityZone: CompatibilityZoneParams
+
+    @Before
+    fun start() {
+        networkMapServer = NetworkMapServer(cacheTimeout, portAllocation.nextHostAndPort())
+        val address = networkMapServer.start()
+        compatibilityZone = CompatibilityZoneParams(URL("http://$address"), rootCert = null)
+    }
+
+    @After
+    fun cleanUp() {
+        networkMapServer.close()
+    }
 
     @Test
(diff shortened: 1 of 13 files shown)


node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -4,85 +4,96 @@ import net.corda.core.node.NodeInfo
...
     private val portAllocation = PortAllocation.Incremental(10000)
 
+    private lateinit var networkMapServer: NetworkMapServer
+    private lateinit var compatibilityZone: CompatibilityZoneParams
+
+    @Before
+    fun start() {
+        networkMapServer = NetworkMapServer(cacheTimeout, portAllocation.nextHostAndPort())
+        val address = networkMapServer.start()
+        compatibilityZone = CompatibilityZoneParams(URL("http://$address"), rootCert = null)
```
al-r3: Use the 4 params constructor of URL

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
```diff
@@ -30,153 +29,108 @@ import java.net.URL
...
+        internalDriver(
+                portAllocation = portAllocation,
+                notarySpecs = emptyList(),
+                compatibilityZone = compatibilityZone
         ) {
-            assertThatThrownBy {
-                startNode(providedName = ALICE_NAME, initialRegistration = true).get()
-            }.isInstanceOf(java.nio.file.NoSuchFileException::class.java)
+            startNode(providedName = CordaX500Name("Alice", "London", "GB")).getOrThrow()
+            assertThat(registrationHandler.requests).containsKeys("Alice")
```
al-r3: Strictly speaking this is more lax than the previous implementation.
This doesn't verify that the node requests for the certificate approval status.

node/src/test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperTest.kt
```diff
@@ -92,4 +91,51 @@ class NetworkRegistrationHelperTest {
...
+        val certService = rigorousMock<NetworkRegistrationService>()
+
+        assertThatThrownBy {
+            NetworkRegistrationHelper(config, certService)
+        }.hasMessageContaining(config.rootCertFile.toString())
+    }
+
+    @Test
+    fun `root cert in response doesn't match expected`() {
+        val config = testNodeConfiguration(
```
al-r3: config and identities can be moved up to the class level. Now they're duplicate between tests.

node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
shamsasari: This is easier to use and clearer. The 4th param in the 4 param c'tor ends up being an empty string which isn't obviously clear why.

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperDriverTest.kt
shamsasari: True. I'll update the registration handler.

node/src/test/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelperTest.kt
shamsasari: Yes

shamsasari:
Comments addressed.

PR #2191, opened 2017-12-06
Make `extraCordappPackagesToScan` on Node Driver a Set.

vkolomeyko:
This is probably not the most exciting change, but having `List` feels a bit awkward when it is not the order, but rather the content that counts and passing duplicates doesn't quite make sense.

### DIFF
--- a/client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
+++ b/client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
@@ -51,7 +51,7 @@ class NodeMonitorModelTest {
     private lateinit var newNode: (CordaX500Name) -> NodeInfo
 
     private fun setup(runTest: () -> Unit) {
-        driver(extraCordappPackagesToScan = listOf("net.corda.finance")) {
+        driver(extraCordappPackagesToScan = setOf("net.corda.finance")) {
             val cashUser = User("user1", "test", permissions = setOf(
                     startFlow<CashIssueFlow>(),
                     startFlow<CashPaymentFlow>(),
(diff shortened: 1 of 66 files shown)


andr3ej:
i don't like it. yes it's logically correct, but now we're forcing java users to use a lib

rnicoll:
`com.google.common.collect.Sets`? I suppose so, but given it'll be a dependency of Corda anyway is that a problem?

vkolomeyko:
Yes, I have not changed any of the Gradle files, so dependency been already there.
@andr3ej - do you think it would be better if I take out Guava import from `VaultQueryJavaTests` and replace `Sets.newHashSet` with my own helper method that does the same?

andr3ej:
let's not complicate things further by adding another helper. sets are a much more expensive way to pass things around than lists, and cordappPackages is typically literal data so using a set doesn't even do anything in practice, and by making it a set you're kind of outsourcing logic that should be in CordappLoader. on balance i can't justify this change, but feel free to get another opinion e.g. @shamsasari

vkolomeyko:
Sure, @shamsasari - what do you think about this change?

@andr3ej 
Re: sets are a much more expensive way to pass things around than lists
Could you elaborate a bit more on that, please?
Also, earlier you've said "it's logically correct" and this is the main point of this PR is to use correct collection data structure to clearly indicate that this is a sequence of *distinct* entries.

vkolomeyko:
@shamsasari - could you have a look at this PR today, please?

@andr3ej - there were few outstanding questions from last week:
Re: sets are a much more expensive way to pass things around than lists
Could you elaborate a bit more on that, please?
Also, earlier you've said "it's logically correct" and this is the main point of this PR is to use correct collection data structure to clearly indicate that this is a sequence of *distinct* entries.
Could you please reply today?

vkolomeyko:
OK - fair enough. I will close this PR without merging.
Thanks all.

PR #2198, opened 2017-12-07
CORDA-831: Add confidential identity certificate type

rnicoll:


testing/test-utils/src/main/kotlin/net/corda/testing/TestConstants.kt
```diff
@@ -70,6 +69,20 @@ val DEV_CA: CertificateAndKeyPair by lazy {
     val caKeyStore = loadKeyStore(ClassLoader.getSystemResourceAsStream("net/corda/node/internal/certificates/cordadevcakeys.jks"), "cordacadevpass")
     caKeyStore.getCertificateAndKeyPair(X509Utilities.CORDA_INTERMEDIATE_CA, "cordacadevkeypass")
 }
+val DEV_NODE_CA: CertificateAndKeyPair by lazy {
```
patrickkuo: node ca should have name constrains in the cert to make sure it can't issue cert with a different legal entity name....

rnicoll: Oops - which means it can't be constant, of course, so I've moved the whole lot into `getTestPartyAndCertificate()`

rnicoll:
@patrickkuo This is ready for re-review

PR #2199, opened 2017-12-07
Correct identities passed to generateSpend

rnicoll:
Pass well known identities instead of confidential into generateSpend()

### DIFF
--- a/node/src/test/kotlin/net/corda/node/services/vault/NodeVaultServiceTest.kt
+++ b/node/src/test/kotlin/net/corda/node/services/vault/NodeVaultServiceTest.kt
@@ -550,8 +550,7 @@ class NodeVaultServiceTest {
 
         database.transaction {
             val moveBuilder = TransactionBuilder(notary).apply {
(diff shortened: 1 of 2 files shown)


node/src/test/kotlin/net/corda/node/services/vault/NodeVaultServiceTest.kt
```diff
@@ -547,8 +547,7 @@ class NodeVaultServiceTest {
 
         database.transaction {
             val moveBuilder = TransactionBuilder(notary).apply {
-                val changeIdentity = services.keyManagementService.freshKeyAndCert(identity, false)
-                Cash.generateSpend(services, this, Amount(1000, GBP), changeIdentity, thirdPartyIdentity)
+                Cash.generateSpend(services, this, Amount(1000, GBP), identity, thirdPartyIdentity)
```
mnesbit: This seems to issue to a well known identity

rnicoll: Have added a note around the identity used (it's just a random key in this case)

PR #2201, opened 2017-12-07
Make NodeInfo mockable

andr3ej:
no jira, experimental


mikehearn:
NodeInfo is just a data object - how comes it needs to be mocked?

andr3ej:
because only 2 of the 4 fields are actually needed in NodeSchedulerServiceTest. it will probably be easier to simply open the class, but i wanted to see how bad an interface would be

andr3ej:
actually, the real problem is NodeSchedulerServiceTest. it's instantiating the world to test not very much at all

PR #2202, opened 2017-12-07
CORDA-830 Introducing the network bootstrapper

shamsasari:
Copying of the node-info files also moved out of Cordform and into NetworkParametersGenerator (which is now called NetworkBootstrapper). This class becomes an external tool to enable deployment of nodes in a test setup on a filesystem.


docs/source/setting-up-a-corda-network.rst
```diff
@@ -9,8 +9,6 @@ order to create and validate transactions.
 There are four broader categories of functionality one such node may have. These pieces of functionality are provided
```
benabineri: three?

shamsasari: yes!

docs/source/setting-up-a-corda-network.rst
```diff
@@ -46,12 +44,38 @@ The most important fields regarding network configuration are:
 * ``rpcAddress``: The address to which Artemis will bind for RPC calls.
 * ``webAddress``: The address the webserver should bind. Note that the port must be distinct from that of ``p2pAddress`` and ``rpcAddress`` if they are on the same machine.
 
+Bootstrapping the network
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The nodes see each other using the network map. This is a collection of statically signed node-info files one for each
```
mikehearn: missing comma typo: files, one for each

docs/source/setting-up-a-corda-network.rst
```diff
@@ -46,12 +44,38 @@ The most important fields regarding network configuration are:
...
+node in the network. Most production deployments will use a highly available, secure distribution of the network map via HTTP.
+
+For test deployments where the nodes (at least initially) reside on the same filesystem, these node-info files can be
+placed directly in the node's ``additional-node-infos`` directory from where the node will pick them up and store them
+in its local network map cache. The node generates its node-info file on startup.
```
mikehearn: generates its _own_ node-info file

docs/source/setting-up-a-corda-network.rst
```diff
@@ -46,12 +44,38 @@ The most important fields regarding network configuration are:
...
+
+In addition to the network map, all the nodes on a network must use the same set of network parameters. These are a set
+of constants which guanrantee interoperability between nodes. The HTTP network map distributes the network parameters
```
mikehearn: typo: guarantee

docs/source/setting-up-a-corda-network.rst
```diff
@@ -46,12 +44,38 @@ The most important fields regarding network configuration are:
...
+which the node downloads automatically. In the absence of this the network parameters must be generated locally. This can
+be done with the network bootstrapper. This a tool that scans all the node configurations from a common directory to
+generate the network parameters file which copied to the nodes' directories. It also copies each node's node-info file
```
mikehearn: typo: which _is_ copied

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -0,0 +1,171 @@
...
+import java.util.concurrent.TimeUnit.SECONDS
+import kotlin.streams.toList
+
+/**
+ * Class to bootstrap a local network of Corda nodes on the same filesystem.
+ */
+@Suppress("UNUSED")
+class NetworkBootstrapper {
+    companion object {
+        private val nodeInfoGenCmd = listOf(
```
mikehearn: Hmm, can we make this configurable or auto-detected somehow? Once @m4ksio has switched the enterprise build to using bundled JVMs this way to start Corda will stop working. There'll be a script/exe instead.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -0,0 +1,171 @@
...
+                val nodeInfo = nodeInfoFile.readAll().deserialize<SignedData<NodeInfo>>().verified()
+                NotaryInfo(nodeInfo.notaryIdentity(), validating)
+            } else {
+                null
+            }
+        }.distinct() // We need distinct as nodes part of a distributed notary share the same notary identity
+    }
+
+    private fun installNetworkParameters(notaryInfos: List<NotaryInfo>, nodeDirs: List<Path>) {
+        // TODO Add cmd line flags for minimumPlatformVersion, maxMessageSize and maxTransactionSize
```
mikehearn: Maybe rather than command line flags, we should use a config file via the library we use. That way you can override the defaults, put them in version control, etc.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -0,0 +1,171 @@
...
+            // cluster and is shared by all the other members. This is the notary identity.
+            2 -> legalIdentities[1]
+            else -> throw IllegalArgumentException("Not sure how to get the notary identity in this scenerio: $this")
+        }
+    }
+
+    // We need to to set serialization env, because generation of parameters is run from Cordform.
+    // KryoServerSerializationScheme is not accessible from nodeapi.
+    private fun initialiseSerialization() {
+        val context = if (java.lang.Boolean.getBoolean("net.corda.testing.amqp.enable")) AMQP_P2P_CONTEXT else KRYO_P2P_CONTEXT
```
mikehearn: I think this can be taken out now. AMQP is on by default.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
shamsasari: There's a quite a few places that assume this. I'll just add a TODO

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
shamsasari: Done.

shamsasari:
Comments addressed.

mikehearn:
We need to do a bit more work on the tool. As written it starts up the node to generate its own NodeInfo file. Fine except for the auto-discovered IP address that gets put into the NodeInfo file, I think? @rick-r3 has been trying to use this to set up a test cluster, which is one of the use cases the tool is meant to satisfy.

I suspect there is an easy fix: add a config file option to override the node's idea of its own IP. In fact we used to have one, I think?

It'd also be useful if the tool could auto-generate the node directories too, based on a template or set of command line flags.

Basically can you try creating your own test cluster with this and iterate the tool until it's easy?

shamsasari:
https://r3-cev.atlassian.net/browse/CORDA-879

PR #2203, opened 2017-12-07
CORDA-840: Gradle plugins are now able to be published to artifactory.

Clintonio:
Why: Because it's needed to remove the publisher role and to make it easier to publish custom versions of gradle-plugins for feature branches.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -93,6 +93,7 @@ apply plugin: 'com.github.ben-manes.versions'
 apply plugin: 'net.corda.plugins.publish-utils'
 apply plugin: 'net.corda.plugins.cordformation'
 apply plugin: 'maven-publish'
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -80,7 +82,6 @@ plugins {
     // but the DSL has some restrictions e.g can't be used on the allprojects section. So we should revisit this if there are improvements in Gradle.
     // Version 1.0.2 of this plugin uses capsule:1.0.1
     id "us.kirchmeier.capsule" version "1.0.2"
-    id "com.jfrog.artifactory" version "4.4.18"
```
Clintonio: Note: This syntax cannot use the variable. It must be a string literal, so it was replaced with the older style.

PR #2206, opened 2017-12-07
CORDA-840: Gradle plugins are now able to be published to artifactory…

Clintonio:
…. (#2203)

Gradle plugins are now able to be published to artifactory.

This is already on master. It will enable you to use the new artifactory publish on TC to make it easier to distribute the custom gradle plugins version.


shamsasari:
We should be doing another merge with master today.

PR #2208, opened 2017-12-08
Updated the changelog to incorporate the new network map design as on…

shamsasari:
…e story.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,6 +8,48 @@ UNRELEASED
 ----------
 * Support for external user credentials data source and password encryption [CORDA-827].
 
+
+* The network map service concept has been re-designed. More information can be found in :doc:`network-map`.
+
+   * The previous design was never intended to be final but was rather a quick implementation in the earliest days of the
+     Corda project to unblock higher priority items. It sufffers from numerous disadvantages including lack of scalability,
+     as one node is expected to hold open and manage connections to every node on the network; not reliable; hard to defend
+     against DoS attacks; etc.
+
+   * There is no longer a special network map node for distributing the network map to the other nodes. Instead the network
+     map is now a collection of signed ``NodeInfo`` files distributed via HTTP.
+
+   * The ``certificateSigningService`` config has been replaced by ``compatibilityZoneURL`` which is the base URL for the
+     doorman registration and for downloading the network map. There is also an end-point for the node to publish its node-info
+     object, which the node does each time it changes. ``networkMapService`` config has been removed.
+
+   * To support local and test deployments, the node polls the ``additional-node-infos`` directory for these signed ``NodeInfo``
+     objects which are stored in its local cache. On startup the node generates its own signed file with the filename format
+     "nodeInfo-*". This can be copied to every node's ``additional-node-infos`` directory that is part of the network.
+
+   * Cordform (which is the ``deployNodes`` gradle task) does this copying automatically for the demos. The ``NetworkMap``
+     parameter is no longer needed.
+
+   * For test deployments we've introduced a bootstrapping tool (see :doc:`setting-up-a-corda-network`).
+
+   * ``extraAdvertisedServiceIds``, ``notaryNodeAddress``, ``notaryClusterAddresses`` and ``bftSMaRt`` configs have been
+     removed. The configuration of notaries has been simplified into a single ``notary`` config object. See
+     :doc:`corda-configuration-file` for more details.
+
+   * Introducing the concept of network parameters which are a set of constants which all nodes on a network must agree on
+     to correctly interop.
+
+   * The set of valid notaries has been moved to the network parameters. Notaries are no longer identified by the CN in
+     their X500 name.
+
+   * Single node notaries no longer have a second separate notary identity. Their main identity *is* their notary identity.
+     Use ``NetworkMapCache.notaryIdentities`` to get the list of available notaries.
+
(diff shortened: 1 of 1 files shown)


docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+   * The previous design was never intended to be final but was rather a quick implementation in the earliest days of the
+     Corda project to unblock higher priority items. It sufffers from numerous disadvantages including lack of scalability,
+     as one node is expected to hold open and manage connections to every node on the network; not reliable; hard to defend
+     against DoS attacks; etc.
+
+   * There is no longer a special network map node for distributing the network map to the other nodes. Instead the network
+     map is now a collection of signed ``NodeInfo`` files distributed via HTTP.
+
+   * The ``certificateSigningService`` config has been replaced by ``compatibilityZoneURL`` which is the base URL for the
+     doorman registration and for downloading this signed network map file. There is also an end-point for the node to
```
kasiastreich: "signed network map file**s**", network map is just collection of hashes

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+     publish its node-info object, which it does each time it changes. ``networkMapService`` config has been removed.
```
kasiastreich: "which is done each time the structure changes" or something more detailed, because now we have double "it" that indicates that node changes.

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+
+   * To support local and test deployments, the node polls the ``additional-node-info`` directory for these signed ``NodeInfo``
```
kasiastreich: additional-node-info**s**

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+     objects which are stored in its local cache. On startup the node generates its own signed file with the filename format
+     "nodeInfo-*". This can be copied to every node's ``additional-node-info`` directory that is part of the network.
```
kasiastreich: Same

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+
+   * Cordform does this copying automatically for the demos, and for test deployments we've introduced a bootstrapping
```
kasiastreich: I would add that it does it on `deployNodes` task.

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+     tool that automates this (see :doc:`setting-up-a-corda-network`). The ``NetworkMap`` parameter is no longer needed.
```
kasiastreich: What is `NetworkMap parameter` and when it was needed?

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+
+   * The concept of advertised services has been removed. Nodes no longer advertise services they offer. It is now expected
```
kasiastreich: That was in previous changelog. I removed `advertisedServices` for V1

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+     that parties will know beforehand the services their counterparts offer.
+
+   * ``extraAdvertisedServiceIds`` config has been removed as part of this. The only use of this config was for notaries,
```
kasiastreich: (...) as part of this work/redesign

kasiastreich: Which also reminds me, that this was pain for our own deployment, because people don't read the docs. Do we have some check on configs parsing, that we no longer specify `extraAdvertisedServiceIds`? With useful information that we changed to notary config and something like "please read the docs" ;) ?

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+     which has been simplified. ``notaryNodeAddress``, ``notaryClusterAddresses`` and ``bftSMaRt`` have been removed and
+     replaced by a single ``notary`` config object. See :doc:`corda-configuration-file` for more details.
+
+   * Introducing the concept of network parameters which are a set of constants which all nodes on a network must agree on
```
kasiastreich: Introduced? Not sure though

docs/source/changelog.rst
```diff
@@ -6,6 +6,47 @@ from the previous milestone release.
...
+     to correctly interop.
+
+   * The definition of a notary has been moved from advertised services to the network parameters. This removes the need
```
kasiastreich: It's worth mentioning that we also removed the second identity for single notaries! Looking up the notary identity should be done through `NetworkMapCache.notaryIdentities` or RPC.

docs/source/changelog.rst
shamsasari: I will rephrase to just "network map" to avoid adding too much detail

docs/source/changelog.rst
shamsasari: Yes, I think it will make sense to do that. I'll raise for master

docs/source/changelog.rst
shamsasari: Good point.

docs/source/changelog.rst
shamsasari: Removed

shamsasari:
Comments addressed.

shamsasari:
Did you mean to request changes? I don't see any comments.

PR #2210, opened 2017-12-08
[CORDA-843] ECDSA full support and entropyToKeyPair

kchalkias:
Fully supporting ECDSA (both R1 and K1) in case they are set as the default signature schemes.
This work includes:

- Ignoring a test that assumes signature idempotency (ECDSA is randomised and not deterministic)

- IRS demo uses keys via MockNode and will still use EdDSA (even if we ever change the default) as the keys are fixed in json files.

- Support entropyToKeyPair for ECDSA R1 and K1

- Unit tests for both EdDSA and ECDSA (eg EdDSA trims the entropy to the first 256 bytes, while ECDSA uses hash(entropy) if it is out of range)

- The default algorithm is still EdDSA, but this PR will allow for easier transition to ECDSA when and if required.

### DIFF
--- a/core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
+++ b/core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
@@ -14,21 +14,22 @@ import org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPrivateKey
 import org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPublicKey
 import org.bouncycastle.jce.ECNamedCurveTable
 import org.bouncycastle.jce.interfaces.ECKey
+import org.bouncycastle.jce.spec.ECNamedCurveParameterSpec
 import org.bouncycastle.pqc.jcajce.provider.sphincs.BCSphincs256PrivateKey
 import org.bouncycastle.pqc.jcajce.provider.sphincs.BCSphincs256PublicKey
 import org.junit.Assert.assertNotEquals
 import org.junit.Test
+import java.math.BigInteger
 import java.security.KeyPairGenerator
 import java.util.*
 import kotlin.test.*
 
 /**
- * Run tests for cryptographic algorithms
+ * Run tests for cryptographic algorithms.
  */
 class CryptoUtilsTest {
 
-    private val testString = "Hello World"
-    private val testBytes = testString.toByteArray()
+    private val testBytes = "Hello World".toByteArray()
 
     // key generation test
     @Test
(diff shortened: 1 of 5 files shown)


mikehearn:
There is such a thing as deterministic ECDSA - should we be using it here? Of course an HSM can do what it wants....

kchalkias:
@mikehearn I would expect that deterministic ECDSA, although similar, it should have a different OID. Also, ECDSA (deterministic or not) refers to the signature scheme itself, not key generation.
Actually `entropyToKeypair` is used for mocking and unit-tests only.

core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
```diff
@@ -14,21 +14,23 @@ import org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPrivateKey
...
 
 /**
- * Run tests for cryptographic algorithms
+ * Run tests for cryptographic algorithms.
  */
 class CryptoUtilsTest {
 
-    val testString = "Hello World"
-    val testBytes = testString.toByteArray()
+    private val testString = "Hello World"
```
rnicoll: While you're clearing this up anyway, could you mark it `const` please?

node/src/test/kotlin/net/corda/node/services/transactions/NotaryServiceTests.kt
```diff
@@ -99,6 +100,7 @@ class NotaryServiceTests {
         assertThat(ex.error).isInstanceOf(NotaryError.TimeWindowInvalid::class.java)
     }
 
+    @Ignore("Only applies to deterministic signature schemes (e.g. EdDSA) and when deterministic metadata is attached (no timestamps or nonces)")
```
rnicoll: Can we force this to use EdDSA then? I'd rather not remove/disable tests unless we have to. If we can but it's more work than this PR merits, please open a JIRA issue / link to existing issue so the work is tracked.

kchalkias: Thanks Ross, as this requires changes to `MockNode` to accept a key type, I will open a JIRA ticket and resolve it in an upcoming PR

core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
kchalkias: Removed it completely as we only use `testBytes` in our tests.

node/src/test/kotlin/net/corda/node/services/transactions/NotaryServiceTests.kt
kchalkias: Thanks for the review, here is the new ticket https://r3-cev.atlassian.net/browse/CORDA-899

PR #2223, opened 2017-12-09
Removes unnecessary R3 references from the docs.

joeldudleyr3:


docs/source/codestyle.rst
```diff
@@ -1,7 +1,7 @@
 Code style guide
 ================
 
-This document explains the coding style used in the R3 prototyping repository. You will be expected to follow these
```
ragmondo: ... r3prototyping existed until today ???

PR #2225, opened 2017-12-11
Fix VaultUpdate filtering in trackBy

exFalso:
This PR fixes the update filtering.

Previously when a user tracked UNCONSUMED states, updates with nothing produced were filtered out, making a local view of UNCONSUMED states inconsistent as updates that only consumed were swallowed.

Imo the reason for this inconsistency stems from using `VaultUpdate`s for two separate purposes:
1. As a datastructure describing what states were consumed and produced
2. As a datastructure describing *the diff between two vault query results*

I think 2 should be a different datastructure, e.g.
```kotlin
data class QueryResultDiff {
  val rowsAdded
  val rowsRemoved
}
```

The above would clarify what's going on. When the user `track`s, the stream should contain `QueryResultDiff`s rather than `VaultUpdate`s. The property of consumedness/producedness is orthogonal to a row being removed/added to a query result.
For example if the user queries CONSUMED states the diffs would always have `rowsRemoved` empty, as the set of CONSUMED states always grows.
If the user queries UNCONSUMED states then both `rowsAdded` and `rowsRemoved` may be non-empty, as the set of UNCONSUMED states may both grow and shrink.

Unfortunately the above would imply an API break, so in this PR I used `VaultUpdate`s as these diffs.
In the case of UNCONSUMED/ALL queries the updates are modified to only contain states of the queried type, but other than that the update stays the same.
In the case of CONSUMED queries the updates are additionally modified by clearing the `produced` set.

I left the old `containsType` function as is because it's public API (even though it isn't used by the filtering anymore).

### DIFF
--- a/docs/source/api-vault-query.rst
+++ b/docs/source/api-vault-query.rst
@@ -488,10 +488,21 @@ Behavioural notes
    is a separate SQL query on the underlying database, and it is entirely conceivable that state modifications are
    taking place in between and/or in parallel to paging requests. When using pagination, always check the value of the
    ``totalStatesAvailable`` (from the ``Vault.Page`` result) and adjust further paging requests appropriately.
-3. In case ``TrackBy`` is used the filters are applied to the initial query and subsequent updates will reflect the
-   diffs between such filtered queries. Note that this means that if a ``trackBy`` queries UNCONSUMED states the
-   returned snapshot will only contain UNCONSUMED states, however updates will contain both consumed and unconsumed
-   states related to the change in case a previously UNCONSUMED state got consumed.
+3. In the case where ``TrackBy`` is used, ``QueryCriteria`` filters are only applied to the initial snapshot query.
+   Subsequent updates reflect all changes since that initial filtered query.
+   For example if a ``trackBy`` queries UNCONSUMED states the returned snapshot will only contain UNCONSUMED states,
+   however updates will contain both consumed and unconsumed states related to the change in case a previously
+   UNCONSUMED state got consumed.
+
+   An example scenario:
+     Alice issues $1
+     Alice moves $1 to Bob
+     Bob moves $1 back to Alice
+   If Alice tracks the vault for UNCONSUMED states before this interchange then she will see:
+     Update($1 produced, nothing consumed)
+     Update(nothing produced, $1 consumed)
+     Update($1 produced, nothing consumed)
+   If we didn't send the second update Alice would think she has $2!
 
 Other use case scenarios
 ------------------------
--- a/docs/source/api-vault-query.rst
+++ b/docs/source/api-vault-query.rst
@@ -488,6 +488,10 @@ Behavioural notes
    is a separate SQL query on the underlying database, and it is entirely conceivable that state modifications are
    taking place in between and/or in parallel to paging requests. When using pagination, always check the value of the
    ``totalStatesAvailable`` (from the ``Vault.Page`` result) and adjust further paging requests appropriately.
+3. In case ``TrackBy`` is used the filters are applied to the initial query and subsequent updates will reflect the
+   diffs between such filtered queries. Note that this means that if a ``trackBy`` queries UNCONSUMED states the
+   returned snapshot will only contain UNCONSUMED states, however updates will contain both consumed and unconsumed
+   states related to the change in case a previously UNCONSUMED state got consumed.
 
 Other use case scenarios
 ------------------------
--- a/node-api/src/main/kotlin/net/corda/nodeapi/ArtemisTcpTransport.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/ArtemisTcpTransport.kt
@@ -51,7 +51,8 @@ class ArtemisTcpTransport {
                     // It does not use AMQP messages for its own messages e.g. topology and heartbeats.
                     // TODO further investigate how to ensure we use a well defined wire level protocol for Node to Node communications.
                     TransportConstants.PROTOCOLS_PROP_NAME to "CORE,AMQP",
-                    TransportConstants.USE_GLOBAL_WORKER_POOL_PROP_NAME to (nodeSerializationEnv != null)
+                    TransportConstants.USE_GLOBAL_WORKER_POOL_PROP_NAME to (nodeSerializationEnv != null),
+                    TransportConstants.REMOTING_THREADS_PROPNAME to (if (nodeSerializationEnv != null) -1 else 1)
             )
 
             if (config != null && enableSSL) {
--- a/node/src/integration-test/kotlin/net/corda/node/services/VaultServiceTest.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/services/VaultServiceTest.kt
@@ -1,0 +1,92 @@
+package net.corda.node.services
+
+import net.corda.core.contracts.withoutIssuer
+import net.corda.core.internal.concurrent.transpose
+import net.corda.core.messaging.startFlow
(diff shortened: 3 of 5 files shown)


josecoll:
Please make this new behaviour - as described above in the PR description - very clear in the Vault Query Corda documentation.

josecoll:
I would also recommend adding JUnit tests to clearly cover the scenarios this PR is attempting to address. 
eg. 
Consider the following scenario:
```Alice issues 2 dollars
Alice moves 1 dollar to Bob
Bob moves 1 dollar to Alice
```
The `CordaRPCOps.vaultTrackBy()` API returns the following updates for Alice for the above:
```0 consumed, 2 produced
1 consumed, 1 produced
0 consumed, 1 produced
```


Makes sense, so far so good. Now consider the following scenario:
```Alice issues 1(!) dollar
Alice moves 1 dollar to Bob
Bob moves 1 dollar to Alice
```
`CordaRPCOps.vaultTrackBy()` now returns something unexpected for Alice (well, imo):
```0 consumed, 1 produced
0 consumed, 1 produced
```
Alice now thinks she has 2 dollars, even though she only has 1! :open_mouth: What's the cause of this?

rick-r3:
@josecoll the state types will default to `UNCONSUMED` in the the `trackBy` and thus you will see all updates (consumed and produced).  Produced states are only filtered out if asking for only `CONSUMED`.  I'm correct there @exFalso ?

Thus your example is

josecoll:
I was just suggesting we update the documentation to clearly highlight the behaviour as illustrated by @exFalso original examples.

mikehearn:
This looks like the kind of behaviour change that targetVersion is useful for. We can keep the old behaviour for old apps and change it for newer apps. At least, if we know what client version the invoker of the API is .... which we should have for CorDapps but we may not have for RPC clients? We can add targetVersion to RPC in a backwards compatible way though.

exFalso:
Currently writing tests for this, I noticed that the original code has a proper bug in it as well. It doesn't filter out states from the update that aren't of the queried type, *but it casts those all the same*. So if an update happens to include Cash and something else, that something else will be case to Cash

exFalso:
@colljos added tests and a paragraph on the behaviour.
@rick-r3 yes that's correct

docs/source/api-vault-query.rst
```diff
@@ -488,6 +488,10 @@ Behavioural notes
    is a separate SQL query on the underlying database, and it is entirely conceivable that state modifications are
    taking place in between and/or in parallel to paging requests. When using pagination, always check the value of the
    ``totalStatesAvailable`` (from the ``Vault.Page`` result) and adjust further paging requests appropriately.
+3. In case ``TrackBy`` is used the filters are applied to the initial query and subsequent updates will reflect the
```
josecoll: In the case where ``TrackBy`` is used, ``QueryCriteria`` filters are only applied to the initial snapshot query. Subsequent updates reflect all changes since that initial filtered query.

docs/source/api-vault-query.rst
```diff
@@ -488,6 +488,10 @@ Behavioural notes
...
+   diffs between such filtered queries. Note that this means that if a ``trackBy`` queries UNCONSUMED states the
```
josecoll: For example, if a ...

docs/source/api-vault-query.rst
```diff
@@ -488,6 +488,10 @@ Behavioural notes
...
+   returned snapshot will only contain UNCONSUMED states, however updates will contain both consumed and unconsumed
+   states related to the change in case a previously UNCONSUMED state got consumed.
```
josecoll: It's non obvious to the reader why you would want both.
Illustrating your example with the 2 spending scenarios would really clarify the reason for this behaviour.

josecoll (changes requested):

node/src/integration-test/kotlin/net/corda/node/services/VaultServiceTest.kt
```diff
@@ -0,0 +1,92 @@
...
+import net.corda.finance.flows.CashIssueFlow
+import net.corda.finance.flows.CashPaymentFlow
+import net.corda.testing.driver.DriverDSL
+import net.corda.testing.driver.driver
+import net.corda.testing.expect
+import net.corda.testing.expectEvents
+import net.corda.testing.sequence
+import org.junit.Test
+
+class VaultServiceTest {
```
josecoll: feels like these tests should be in either `VaultQueryTests` or `VaultWithCashTests`.

josecoll:
@exFalso can you point our where the bug you mention above is fixed.

Afaik there has always been an explicit check for the correct Class Type:
```
        /** Checks whether the update contains a state of the specified type and state status */
        fun <T : ContractState> containsType(clazz: Class<T>, status: StateStatus) =
                when (status) {
                    StateStatus.UNCONSUMED -> produced.any { clazz.isAssignableFrom(it.state.data.javaClass) }
                    StateStatus.CONSUMED -> consumed.any { clazz.isAssignableFrom(it.state.data.javaClass) }
                    else -> consumed.any { clazz.isAssignableFrom(it.state.data.javaClass) }

```

exFalso:
@josecoll So in the old code that function is a simple check, it doesn't filter the update itself.

For example if I have a VaultUpdate that contains `consumed = [Cash, Obligation], produced = [Cash]` then the old filtering function returns true when `Cash` is queried, but doesn't filter out the `Obligation` state, the `VaultUpdate<ContractState>` is simply cast to `VaultUpdate<Cash>`. This doesn't throw on the server because of erasure, it will only throw on the client when they will try to use the state as Cash.

In the new code the filtering function actually does filtering (`filterTo`), it's not just a predicate

node/src/integration-test/kotlin/net/corda/node/services/VaultServiceTest.kt
exFalso: Those are unit tests, this is an integ test.. perhaps better naming is due?

docs/source/api-vault-query.rst
rick-r3: Yeah, perhaps worth re-surfacing the fact that the only filters currently applied to the updates are by contract state type and status.  We may improve the filtering later for those with an updated "target version" for their app.

rick-r3 (commented):
A couple of changes to the docs is worthwhile.

exFalso:
@josecoll updated docs, back to you

josecoll (changes requested):
This PR still needs a changelog entry to describe the behavioural change.

node-api/src/main/kotlin/net/corda/nodeapi/ArtemisTcpTransport.kt
```diff
@@ -51,7 +51,8 @@ class ArtemisTcpTransport {
                     // It does not use AMQP messages for its own messages e.g. topology and heartbeats.
                     // TODO further investigate how to ensure we use a well defined wire level protocol for Node to Node communications.
                     TransportConstants.PROTOCOLS_PROP_NAME to "CORE,AMQP",
-                    TransportConstants.USE_GLOBAL_WORKER_POOL_PROP_NAME to (nodeSerializationEnv != null)
+                    TransportConstants.USE_GLOBAL_WORKER_POOL_PROP_NAME to (nodeSerializationEnv != null),
+                    TransportConstants.REMOTING_THREADS_PROPNAME to (if (nodeSerializationEnv != null) -1 else 1)
```
josecoll: Why is this change in this PR?

mikehearn:
How much work to rebase this and get it merged? Would be good to fix it!

PR #2233, opened 2017-12-12
Fix node integration tests.

al-r3:
… used as root, so that unittests can use a cert of their liking.
Started nodes where hanging due to an exception on validating that certificate.

Also changed NodeRegistrationTest to start 2 nodes.


testing/node-driver/src/main/kotlin/net/corda/testing/node/network/NetworkMapServer.kt
```diff
@@ -49,16 +51,17 @@ class NetworkMapServer(cacheTimeout: Duration,
                     CertificateType.INTERMEDIATE_CA,
                     rootCAKeyAndCert.certificate,
                     rootCAKeyAndCert.keyPair,
-                    X500Name("CN=Corda Network Map,L=London"),
+                    CordaX500Name("Corda Network Map", "R3 Ltd", "London","GB"),
                     networkMapKey.public).cert
+            // Check that the certificate validates. Nodes will perform this check upon receiving a network map,
+            // it's better to fail here than then.
```
shamsasari: typo: there

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NodeRegistrationTest.kt
```diff
@@ -69,6 +68,8 @@ class NodeRegistrationTest {
         ) {
             startNode(providedName = CordaX500Name("Alice", "London", "GB")).getOrThrow()
             assertThat(registrationHandler.idsPolled).contains("Alice")
+            startNode(providedName = CordaX500Name("Genevieve", "London", "GB")).getOrThrow()
```
shamsasari: Thanks for this. 

Can you add a simple flow to make sure they can talk to each other. Invoking `CashIssueAndPaymentFlow` should do the trick. You'll probably need to specify "net.corda.finance" for the `extraCordappPackagesToScan` driver parameter to get it working.

al-r3:
done

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NodeRegistrationTest.kt
```diff
@@ -57,18 +65,27 @@ class NodeRegistrationTest {
...
+            assertThat(registrationHandler.idsPolled).contains(ALICE_NAME)
+            val genevieve = startNode(providedName = CordaX500Name(GENEVIEVE_NAME, "London", "GB")).getOrThrow()
+            assertThat(registrationHandler.idsPolled).contains(GENEVIEVE_NAME)
+
+            // Check we nodes communicate among themselves (and the notary).
+            val anonymous = true
+            genevieve.rpc.startFlow(::CashIssueAndPaymentFlow, 1000.DOLLARS, OpaqueBytes.of(12),
+                    alice.nodeInfo.legalIdentities.first(),
+                    anonymous,
+                    defaultNotaryIdentity)
```
shamsasari: `startFlow` returns a Future which must be waited on otherwise any exception is lost

node/src/integration-test/kotlin/net/corda/node/utilities/registration/NodeRegistrationTest.kt
```diff
@@ -133,8 +139,9 @@ class RegistrationHandler(private val rootCertAndKeyPair: CertificateAndKeyPair)
                 certificationRequest,
                 rootCertAndKeyPair.keyPair,
                 arrayOf(rootCertAndKeyPair.certificate.cert))
-        certPaths[name.organisation] = certPath
-        return Response.ok(name.organisation).build()
+        val requestId = name.organisation.replace(Regex("""\s"""),"")
```
shamsasari: I'd rather `registration` just checks there's no whitespace in the organisation field

al-r3: Why? It all seems to works if there are spaces (the default notary has spaces in its name).
It is simply on the http side that spaces cause issues, I'm not even sure if it's in code or test

shamsasari: This is just obfuscating the whitespace in the org name issue. If it's the notary that's causing the problem, then you can specify the notary name with the `notarySpecs` param.

al-r3: ??
This is removing the whitespace just to generate the request ID for the generation process. If we have a request ID with a space then the test 'doorman' returns an HTTP error.

What issue are you talking about?

shamsasari: I know. What I want is to not be doing this auto-mangling and to keep things clear, i.e. do not use X500 names with whitespaces in them.

shamsasari: To further clarify why I don't like this. The driver allows two nodes with the names "AliceCorp" and "Alice Corp" (as far as I'm aware it doesn't remove whitespaces). But this is treating both names as the same. So it's better to avoid whitespace for the purposes of these tests.

al-r3:
@shamsasari this version of the code has the test passing.
However there are changes in the node registration, see https://github.com/corda/corda/blob/5b12c5177e3d327439490cded0917972da7a0444/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt#L713
for the new alias under which the private key is saved.
Though I still miss some piece.

Also this probably breaks some test I'll need to check those.

tl;dr we need the private key stored under "identity-private-key" additionally to "cordaclientca" (maybe instead works as well)

al-r3:
Also that is not notary-specific. And there are some constants which should be folded that will lead to some interesting module juggling

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -225,6 +229,7 @@ class DriverDSLImpl(
         val configuration = config.parseAsNodeConfiguration()
 
         configuration.trustStoreFile.parent.createDirectories()
+       // if (configuration.notary != null) configuration.configureDevKeyAndTrustStores(providedName)  // needed for notaries
```
shamsasari: remove?

PR #2237, opened 2017-12-12
add noise to driver generated paths to avoid clashes

al-r3:


mnesbit:
Lets just wait to see if that makes anything else break during the build. Shouldn't, but you never know!

al-r3:
Well github won't let me merge should that happen anyway

PR #2238, opened 2017-12-13
CORDA-716 Inline testNodeConfiguration

andr3ej:
shouldn't be api, and only MockNode needs most of the fields

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
@@ -2,8 +2,6 @@
 
 package net.corda.testing.node
 
-import com.nhaarman.mockito_kotlin.doReturn
-import com.nhaarman.mockito_kotlin.whenever
 import net.corda.core.context.Actor
 import net.corda.core.context.AuthServiceId
 import net.corda.core.context.InvocationContext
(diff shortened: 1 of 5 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
```diff
@@ -48,31 +38,6 @@ fun ServiceHub.transaction(
...
-        myLegalName: CordaX500Name): NodeConfiguration {
-    abstract class MockableNodeConfiguration : NodeConfiguration // Otherwise Mockito is defeated by val getters.
-    return rigorousMock<MockableNodeConfiguration>().also {
-        doReturn(baseDirectory).whenever(it).baseDirectory
-        doReturn(myLegalName).whenever(it).myLegalName
-        doReturn("cordacadevpass").whenever(it).keyStorePassword
-        doReturn("trustpass").whenever(it).trustStorePassword
-        doReturn(emptyList<User>()).whenever(it).rpcUsers
-        doReturn(null).whenever(it).notary
-        doReturn(makeTestDataSourceProperties(myLegalName.organisation)).whenever(it).dataSourceProperties
```
andr3ej: only MockNode actually needs dataSourceProperties, and it configures its own! so this line is gone now

PR #2239, opened 2017-12-13
CORDA-353 Large Attachments

cburlinchon:
Two changes in this PR to allow importing large attachments(larger than jvm heap size).
1. NodeAttachmentService streams attachment into db whilst calculating the hash. Once we have inserted we check db contains said hash.
2. RPC now checks if we are sending or receiving a stream, if so, we make use of artemis' large message support and stream to a temp file.


mikehearn:
This PR has no reviewers assigned. Please assign some or close it until it's ready.

node-api/src/main/kotlin/net/corda/nodeapi/RPCApi.kt
```diff
@@ -144,12 +154,11 @@ object RPCApi {
                     RPCApi.ClientToServer.Tag.RPC_REQUEST -> RpcRequest(
                             clientAddress = MessageUtil.getJMSReplyTo(message),
                             methodName = message.getStringProperty(METHOD_NAME_FIELD_NAME),
-                            serialisedArguments = message.getBodyAsByteArray(),
+                            serialisedArguments = if (message.isLargeMessage) kotlin.ByteArray(0) else message.getBodyAsByteArray(),
```
exFalso: instead of doing this i'd suggest defining something like this:
```kotlin
sealed class RpcRequestPayload {
  data class Regular(val serialisedArguments: ByteArray) : RpcRequestPayload()
  data class Streamed(val stream: InputStream) : RpcRequestPayload()
}
```
replace both `serialisedArguments` and `bodyInputStream` with one of the above and use `when` to decide what to do. This way there is no ambiguity and no need for the empty bytarray

node-api/src/main/kotlin/net/corda/nodeapi/RPCApi.kt
```diff
@@ -211,7 +220,15 @@ object RPCApi {
                     RPCApi.ServerToClient.Tag.RPC_REPLY -> {
                         val id = message.invocationId(RPC_ID_FIELD_NAME, RPC_ID_TIMESTAMP_FIELD_NAME) ?: throw IllegalStateException("Cannot parse invocation id from client message.")
                         val poolWithIdContext = context.withProperty(RpcRequestOrObservableIdKey, id)
-                        RpcReply(id, message.getBodyAsByteArray().deserialize(context = poolWithIdContext))
+
+                        if(message.isLargeMessage) {
+                            val tempFile = Files.createTempFile("pre", null).toFile()
```
exFalso: I'm wary of this leaking files. `deleteOnExit` may or may not be called depending on how the jvm terminates and whether the file is being streamed to at the time of shutdown. Furthermore if we delay the deletion until jvm shutdown the client will just keep creating new files while it's running
Two alternatives:
- I wonder whether we can pipe the stream to the user instead? iirc this requires another modification in artemis as the stream is aborted when the handler finishes. Hmm, perhaps not now
- Stream to a file but create a wrapper around the returned `FileInputStream` so that the file is cleaned up once the stream finishes or the wrapper is finalised.

node/src/integration-test/kotlin/net/corda/node/services/persistence/NodeAttachmentServiceTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.io.FileOutputStream
+import java.util.concurrent.CompletableFuture.supplyAsync
+import java.util.zip.Deflater
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
+
+class NodeAttachmentServiceTest {
+
+    private fun largeAttachment(): File {
+        val file = File.createTempFile("NodeAttachmentServiceTest", null)
```
exFalso: use `@Rule` and `TemporaryFolder` instead

node/src/integration-test/kotlin/net/corda/node/services/persistence/NodeAttachmentServiceTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        return file
+    }
+
+    @Test
+    fun `import large attachment`() {
+        driver {
+            val node = startNode(rpcUsers = listOf(User("A", "A", setOf("InvokeRpc.uploadAttachment"))), maximumHeapSize = "512m").get()
+            val senderThread = supplyAsync {
+                node.rpcClientToNode().start("A", "A").use {
+                    it.proxy.uploadAttachment(FileInputStream(largeAttachment()))
```
exFalso: Actually scratch that, use `RepeatingBytesInputStream` instead of a file. Then do a roundtrip stream through the node and check whether the returned bytes are the same.

Please also add a test that does this several times in parallel

node/src/main/kotlin/net/corda/node/services/messaging/RPCServer.kt
```diff
@@ -273,22 +276,37 @@ class RPCServer(
...
                 context.invocation.pushToLoggingContext()
-                when (arguments) {
-                    is Try.Success -> {
-                        rpcExecutor!!.submit {
-                            val result = invokeRpc(context, clientToServer.methodName, arguments.value)
-                            sendReply(clientToServer.replyId, clientToServer.clientAddress, result)
-                        }
+
+                if(artemisMessage.isLargeMessage) {
+                    val tempFile = Files.createTempFile("pre", null).toFile()
```
exFalso: Same thing applies as before, except in this case we *can* hand over the stream to the RPC handler without modifying artemis as the rpc is run within the message handler. So we don't need any intermediate fs buffering

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -173,12 +180,20 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
...
+        val session = currentDBSession()
+        val criteriaBuilder = session.criteriaBuilder
+        val criteriaQuery = criteriaBuilder.createQuery(DBAttachment::class.java)
+        val root = criteriaQuery.from(DBAttachment::class.java)
+        criteriaQuery.select(root)
+        criteriaQuery.where(criteriaBuilder.equal(root.get<String>(DBAttachment::contentHash.name), contentHash.toString()))
+        val query = session.createQuery(criteriaQuery)
+        return if (query.resultList.size > 0) {
+            val bytes = query.resultList.first().content.binaryStream.readBytes()
+            AttachmentImpl(contentHash, { bytes }, checkAttachmentsOnLoad)
```
exFalso: This reads everything into memory again... is this correct? Can verify by doing the roundtrip test

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -189,45 +204,46 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
...
+        val contentHash = SecureHash.SHA256(hs.hash().asBytes())
+        val rereadAttachment = session.get(DBAttachment::class.java, id)
+
+        if (hasAttachment(contentHash)) {
+            session.delete(rereadAttachment)
+            throw java.nio.file.FileAlreadyExistsException(contentHash.toString())
         } else {
-            throw java.nio.file.FileAlreadyExistsException(id.toString())
+            checkIsAValidJAR(rereadAttachment.content.binaryStream)
+            rereadAttachment.contentHash = contentHash.toString()
```
exFalso: This is racy.. is it fine if we end up with duplicates?

exFalso (changes requested):
Couple of things

node-api/src/main/kotlin/net/corda/nodeapi/RPCApi.kt
jamesbr3: Couple of things worth mentioning about temp-file creation. The general recommendation is to:
* Never create temp files in a shared directory (i.e. somewhere that other users on the system may have access to)
* Aim to create & open temp file as an atomic operation
* Always use 'secure' random filename - i.e. that isn't guessable by an attacker. I *think* Files.createTempFile will do this automatically in Java (6+???), but not sure otoh

The aim is to avoid TOCTOU (time-of-creation, time-of-use) attacks - i.e. a race-condition where there is a small time window between file creation, and the point where you open the file for what you hope is exclusive write-access.

I'd advise double-checking where this temp-file is going to be created. And double-check the permissions on the directory to make sure it's only the corda-node user that has access.

Is Files.createTempFile.toFile() atomic? Is File.createTempFile preferable because it (I assume) keeps the file open as it returns thje File object back to you?

PR #2242, opened 2017-12-13
CORDA-833: SignedNodeInfo object for holding a signature for each identity in a NodeInfo

shamsasari:
This forms part of the network map.


node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -133,9 +129,9 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
                         // Download new node info from network map
                         try {
                             networkMapClient.getNodeInfo(it)
-                        } catch (t: Throwable) {
+                        } catch (e: Exception) {
```
rnicoll: YES!

Sorry I have a think about catching `Throwable`, always excited to see us move away from doing so.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -1,233 +1,214 @@
...
 
         updater.subscribeToNetworkMap()
         scheduler.advanceTimeBy(10, TimeUnit.SECONDS)
         // TODO: Remove sleep in unit test.
-        Thread.sleep(200)
+        Thread.sleep(2L * cacheExpiryMs)
 
         // 4 node info from network map, and 1 from file.
-        assertEquals(4, nodeInfoMap.size)
+        assertThat(nodeInfoMap).hasSize(4)
```
rnicoll: As we making a conscious decision to move to `assertThat` over `assertEquals`?

shamsasari: In this particular case `hasSize` will give more a informative error msg, i.e. what's in the list.

PR #2245, opened 2017-12-13
Clarifies that only the first two tutorials are related and form a sequence.

joeldudleyr3:


docs/source/tutorials-index.rst
```diff
@@ -1,11 +1,22 @@
...
+The Hello, World tutorials should be followed in sequence and show how to build a complete CorDapp.
+
 .. toctree::
    :maxdepth: 1
 
    hello-world-introduction
    tut-two-party-introduction
+
+The remaining tutorials cover individual platform features in isolation. They are not related to one another or to the
+Hello, World tutorials.
```
anthonykr3: Some are related, i.e. contracts + contract tests, flows + flow tests (I think) - would the second sentence be better as something like "They don't depend on the code from the hello world tutorials, and can be done in any order"

docs/source/tutorials-index.rst
```diff
@@ -1,11 +1,22 @@
 Tutorials
 =========
 
+This section is split into two parts.
+
+The Hello, World tutorials should be followed in sequence and show how to build a complete CorDapp.
```
anthonykr3: worth mentioning it being based on the example-corda-app?

PR #2249, opened 2017-12-13
Renamed TestIdentity.key to keyPair and pubkey to publicKey

shamsasari:


testing/test-utils/src/main/kotlin/net/corda/testing/CoreTestUtils.kt
```diff
@@ -81,7 +83,7 @@ fun freePort(): Int = freePortCounter.getAndAccumulate(0) { prev, _ -> 30000 + (
  */
 fun getFreeLocalPorts(hostName: String, numberToAlloc: Int): List<NetworkHostAndPort> {
     val freePort = freePortCounter.getAndAccumulate(0) { prev, _ -> 30000 + (prev - 30000 + numberToAlloc) % 10000 }
-    return (freePort..freePort + numberToAlloc - 1).map { NetworkHostAndPort(hostName, it) }
+    return (freePort until freePort + numberToAlloc).map { NetworkHostAndPort(hostName, it) }
```
andr3ej: since you're here, can you make it ``0 until numberToAlloc`` and ``freePort + it``?

shamsasari:
Not sure who else to add, it's a pretty trivial change.

shamsasari:
@corda/core

testing/test-utils/src/main/kotlin/net/corda/testing/CoreTestUtils.kt
```diff
@@ -117,11 +119,11 @@ fun getTestPartyAndCertificate(name: CordaX500Name, publicKey: PublicKey): Party
 }
 
 class TestIdentity @JvmOverloads constructor(val name: CordaX500Name, entropy: Long? = null) {
-    val key = if (entropy != null) entropyToKeyPair(BigInteger.valueOf(entropy)) else generateKeyPair()
-    val pubkey get() = key.public!!
-    val party = Party(name, pubkey)
-    val identity by lazy { getTestPartyAndCertificate(party) } // Often not needed.
-    fun ref(vararg bytes: Byte) = party.ref(*bytes)
+    val keyPair: KeyPair = if (entropy != null) entropyToKeyPair(BigInteger.valueOf(entropy)) else generateKeyPair()
```
mnesbit: Shouldn't this be by lazy?

andr3ej: the usual use-case is to associate a keypair or pubkey with a name, so no

shamsasari: I don't think so, because it's used by `publicKey` which is used quite a bit.

kchalkias:
there might some other `KEY` cases that should be renamed to `KEYPAIR`

core/src/test/kotlin/net/corda/core/serialization/TransactionSerializationTests.kt
```diff
@@ -25,9 +25,9 @@ class TransactionSerializationTests {
         val megaCorp = TestIdentity(CordaX500Name("MegaCorp", "London", "GB"))
         val MINI_CORP = TestIdentity(CordaX500Name("MiniCorp", "London", "GB")).party
         val DUMMY_NOTARY get() = dummyNotary.party
-        val DUMMY_NOTARY_KEY get() = dummyNotary.key
+        val DUMMY_NOTARY_KEY get() = dummyNotary.keyPair
```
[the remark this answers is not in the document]
andr3ej: these names only exist to keep the diff at a reviewable size when i introduced TestIdentity, ideally they should be inlined

PR #2252, opened 2017-12-14
ENT-1249 - Downgrade Artemis from v2.4 to v2.1

tlil:
Downgrade Artemis to version 2.1 to protect against rogue non-TLS connections causing OOM.
 * https://issues.apache.org/jira/browse/ARTEMIS-1559

mem-leak
mem-leak-source

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ buildscript {
     ext.capsule_version = '1.0.1'
 
     ext.asm_version = '0.5.3'
-    ext.artemis_version = '2.4.0'
+    ext.artemis_version = '2.2.0'
     ext.jackson_version = '2.9.2'
     ext.jetty_version = '9.4.7.v20170914'
     ext.jersey_version = '2.25'
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,22 @@ buildscript {
     ext.capsule_version = '1.0.1'
 
     ext.asm_version = '0.5.3'
-    ext.artemis_version = '2.2.0'
+
+    /*
+     * TODO Upgrade to version 2.4 for large message streaming support
+     *
+     * Due to a memory leak in the connection handling code in Artemis, we are
+     * temporarily downgrading to version 2.1.0 (version used prior to the 2.4
+     * bump).
+     *
+     * The memory leak essentially triggers an out-of-memory exception within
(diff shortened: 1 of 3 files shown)


mikehearn:
Is this bug reported to upstream anywhere? We need a fix in 2.4 - if we downgrade we'll lose large message streaming support again.

tlil:
@mikehearn Yes, I'm currently trying to find out if it is reported or not. If it isn't, I'll raise an issue with upstream.

I've been talking to @cburlinchon and @mnesbit. The thinking is that it would be a temporary downgrade for code freeze tomorrow. Next, we would obviously need to get the issue resolved with Artemis v2.4 as part of the large message work. Sucks, but would rather not expose vulnerabilities like this in the pre-release we cut tomorrow

tlil:
@mikehearn: **If we were to roll back**: Given the number of changes in 2.2.0 ([changelog](https://activemq.apache.org/artemis/release-notes-2.2.0.html)), would you prefer rolling back to what we were using prior to 2.4.0, which was 2.1.0?

mikehearn:
OK then. Can you put a comment in the Gradle file explaining why we're on such an old version of the broker? And ideally with a link to the upstream bug report?

tlil:
Absolutely - will put that in. Any preference as to what version? 2.2.0 is the latest version where this leak doesn't trigger an OOM. However, we were on version 2.1.0 prior to the 2.4.0 bump.

mikehearn:
Let's go back to 2.1.0 for now to keep things simple and go back to 2.4 ASAP.

tlil:
Coolie, will update accordingly. Ta

tlil:
@mikehearn @cburlinchon updated with description

jbertram:
It would be awesome to get a reproduce for this on the upstream ARTEMIS JIRA.  Is this feasible?

tlil:
Hi @jbertram – yeah, sorry, I've been out. Hopefully back next week, so I'll jump on this then. ([ARTEMIS-1559](https://issues.apache.org/jira/browse/ARTEMIS-1559?focusedCommentId=16303612&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16303612)

PR #2255, opened 2017-12-14
SignedNetworkMap verification fix

kasiastreich:
SignedNetworkMap verification should also include cert path validation,
which was probably moved away by accident, because docs say about the
exception CertPathValidatorException.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
@@ -67,7 +67,7 @@ class SignedNetworkMap(val raw: SerializedBytes<NetworkMap>, val sig: DigitalSig
     @Throws(SignatureException::class, CertPathValidatorException::class)
     fun verified(trustedRoot: X509Certificate): NetworkMap {
         sig.by.publicKey.verify(raw.bytes, sig)
-        // Assume network map cert is issued by the root.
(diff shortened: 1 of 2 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/NetworkMap.kt
```diff
@@ -63,9 +64,11 @@ class SignedNetworkMap(val raw: SerializedBytes<NetworkMap>, val sig: DigitalSig
      * @throws CertPathValidatorException if the certificate path is invalid.
      * @throws SignatureException if the signature is invalid.
      */
-    @Throws(SignatureException::class)
-    fun verified(): NetworkMap {
+    @Throws(SignatureException::class, CertPathValidatorException::class)
+    fun verified(trustedRoot: X509Certificate): NetworkMap {
         sig.by.publicKey.verify(raw.bytes, sig)
+        // Assume network map cert is issued by the root.
```
rnicoll: We're explicitly expecting them not to be issued by the root in https://github.com/corda/enterprise/pull/131 , but I think what you're checking is the network map certificate is in the same hierarchy at the default trust root, which is correct.

Could you rephrase to something like:

// Assume network map cert is under the default trust root.

kasiastreich: Addressed

PR #2259, opened 2017-12-14
CORDA-716 Configure api scanner for test modules

andr3ej:


andr3ej:
i'm not waiting until monday+, this is easily rolled back if there is a problem

chrisr3:
We also need to update the documentation to say that the `node-driver` API is now stable. The [README](https://github.com/corda/corda/tree/master/gradle-plugins/api-scanner/README.md) contains a line saying:
```
See here for Corda's public API strategy. 
```
where the [here](https://github.com/corda/corda/blob/master/docs/source/corda-api.rst) link is broken. I've corrected it for this comment :-), but the `corda-api.rst` document needs to include a section for `node-driver` now.

mikehearn:
Why is there only one new API added to the checker file? Did test-utils go in earlier? And yeah updating the docs in the same PR would have been good.

andr3ej:
@mikehearn according to the readme, the release manager updates the file when there is a release. the docs weren't updated in this pr because the plan was to merge it before the freeze and docs after.

PR #2268, opened 2017-12-15
CORDA-869: Clean up error reporting in node config/registration

rnicoll:


node/src/main/kotlin/net/corda/node/ArgsParser.kt
```diff
@@ -73,6 +74,7 @@ data class CmdLineOptions(val baseDirectory: Path,
                           val sshdServer: Boolean,
                           val justGenerateNodeInfo: Boolean,
                           val bootstrapRaftCluster: Boolean) {
+    @Throws(ConfigException::class)
```
shamsasari: We're dealing with node code here, which is not being called from Java code, so the `throws ConfigException` this annotation generates is pointless. Please remove.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -186,14 +191,8 @@ open class NodeStartup(val args: Array<String>) {
...
 
-    open protected fun loadConfigFile(cmdlineOptions: CmdLineOptions): NodeConfiguration {
-        try {
-            return cmdlineOptions.loadConfig()
-        } catch (configException: ConfigException) {
-            println("Unable to load the configuration file: ${configException.rootCause.message}")
-            throw configException
-        }
-    }
+    @Throws(ConfigException::class)
```
shamsasari: ditto

PR #2269, opened 2017-12-17
Raft notaries can share a single key pair for the service identity (i…

adagys:
…n contrast to a shared composite public key, and individual signing key pairs). This allows modifying the cluster size for a running notary service. 

Note that this does not apply to the BFT notary.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/DevIdentityGenerator.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/DevIdentityGenerator.kt
@@ -12,6 +12,9 @@ import net.corda.nodeapi.internal.config.NodeSSLConfiguration
 import net.corda.nodeapi.internal.crypto.*
 import org.slf4j.LoggerFactory
 import java.nio.file.Path
+import java.security.KeyPair
+import java.security.PublicKey
+import java.security.cert.X509Certificate
 
 /**
  * Contains utility methods for generating identities for a node.
@@ -47,37 +50,56 @@ object DevIdentityGenerator {
         return identity.party
     }
 
-    fun generateDistributedNotaryIdentity(dirs: List<Path>, notaryName: CordaX500Name, threshold: Int = 1): Party {
+    fun generateDistributedNotaryCompositeIdentity(dirs: List<Path>, notaryName: CordaX500Name, threshold: Int = 1): Party {
         require(dirs.isNotEmpty())
 
-        log.trace { "Generating identity \"$notaryName\" for nodes: ${dirs.joinToString()}" }
-        val keyPairs = (1..dirs.size).map { generateKeyPair() }
-        val compositeKey = CompositeKey.Builder().addKeys(keyPairs.map { it.public }).build(threshold)
-
+        log.trace { "Generating composite identity \"$notaryName\" for nodes: ${dirs.joinToString()}" }
         val caKeyStore = loadKeyStore(javaClass.classLoader.getResourceAsStream("certificates/cordadevcakeys.jks"), "cordacadevpass")
         val intermediateCa = caKeyStore.getCertificateAndKeyPair(X509Utilities.CORDA_INTERMEDIATE_CA, "cordacadevkeypass")
-        val rootCert = caKeyStore.getCertificate(X509Utilities.CORDA_ROOT_CA)
+        val rootCert = caKeyStore.getX509Certificate(X509Utilities.CORDA_ROOT_CA)
 
+        val keyPairs = (1..dirs.size).map { generateKeyPair() }
+        val notaryKey = CompositeKey.Builder().addKeys(keyPairs.map { it.public }).build(threshold)
         keyPairs.zip(dirs) { keyPair, nodeDir ->
-            val (serviceKeyCert, compositeKeyCert) = listOf(keyPair.public, compositeKey).map { publicKey ->
-                X509Utilities.createCertificate(
-                        CertificateType.SERVICE_IDENTITY,
-                        intermediateCa.certificate,
-                        intermediateCa.keyPair,
-                        notaryName.x500Principal,
-                        publicKey)
-            }
-            val distServKeyStoreFile = (nodeDir / "certificates").createDirectories() / "distributedService.jks"
-            val keystore = loadOrCreateKeyStore(distServKeyStoreFile, "cordacadevpass")
-            keystore.setCertificateEntry("$DISTRIBUTED_NOTARY_ALIAS_PREFIX-composite-key", compositeKeyCert)
-            keystore.setKeyEntry(
-                    "$DISTRIBUTED_NOTARY_ALIAS_PREFIX-private-key",
-                    keyPair.private,
-                    "cordacadevkeypass".toCharArray(),
-                    arrayOf(serviceKeyCert, intermediateCa.certificate, rootCert))
-            keystore.save(distServKeyStoreFile, "cordacadevpass")
+            generateCertificates(keyPair, notaryKey, intermediateCa, notaryName, nodeDir, rootCert)
         }
 
-        return Party(notaryName, compositeKey)
+        return Party(notaryName, notaryKey)
+    }
+
+    fun generateDistributedNotarySingularIdentity(dirs: List<Path>, notaryName: CordaX500Name): Party {
+        require(dirs.isNotEmpty())
+
+        log.trace { "Generating singular identity \"$notaryName\" for nodes: ${dirs.joinToString()}" }
+        val caKeyStore = loadKeyStore(javaClass.classLoader.getResourceAsStream("certificates/cordadevcakeys.jks"), "cordacadevpass")
+        val intermediateCa = caKeyStore.getCertificateAndKeyPair(X509Utilities.CORDA_INTERMEDIATE_CA, "cordacadevkeypass")
+        val rootCert = caKeyStore.getX509Certificate(X509Utilities.CORDA_ROOT_CA)
+
+        val keyPair = generateKeyPair()
+        val notaryKey = keyPair.public
+        dirs.forEach { dir ->
+            generateCertificates(keyPair, notaryKey, intermediateCa, notaryName, dir, rootCert)
+        }
+        return Party(notaryName, notaryKey)
+    }
+
+    private fun generateCertificates(keyPair: KeyPair, notaryKey: PublicKey, intermediateCa: CertificateAndKeyPair, notaryName: CordaX500Name, nodeDir: Path, rootCert: X509Certificate) {
+        val (serviceKeyCert, compositeKeyCert) = listOf(keyPair.public, notaryKey).map { publicKey ->
+            X509Utilities.createCertificate(
+                    CertificateType.SERVICE_IDENTITY,
+                    intermediateCa.certificate,
+                    intermediateCa.keyPair,
+                    notaryName.x500Principal,
+                    publicKey)
+        }
+        val distServKeyStoreFile = (nodeDir / "certificates").createDirectories() / "distributedService.jks"
+        val keystore = loadOrCreateKeyStore(distServKeyStoreFile, "cordacadevpass")
+        keystore.setCertificateEntry("$DISTRIBUTED_NOTARY_ALIAS_PREFIX-composite-key", compositeKeyCert)
+        keystore.setKeyEntry(
+                "$DISTRIBUTED_NOTARY_ALIAS_PREFIX-private-key",
+                keyPair.private,
+                "cordacadevkeypass".toCharArray(),
+                arrayOf(serviceKeyCert, intermediateCa.certificate, rootCert))
+        keystore.save(distServKeyStoreFile, "cordacadevpass")
     }
 }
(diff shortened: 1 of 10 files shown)


node/src/integration-test/kotlin/net/corda/node/services/DistributedServiceTests.kt
```diff
@@ -71,68 +75,83 @@ class DistributedServiceTests {
...
+            checkRequestsDistributedEvenly()
         }
+    }
 
-        // Now kill a notary node
-        with(notaryNodes[0].process) {
-            destroy()
-            waitFor()
+    @Test
+    fun `cluster shares a single keypair instead of using a composite public key and individual private keys`() {
```
shamsasari: This test isn't asserting this claim

adagys: True, I'll reword it

adagys:
It's a feature we need for production deployments, so we can experiment with cluster sizes

shamsasari:
Is there a design doc or JIRA ticket I can look at, because this needs more thought. How does the doorman fit into this? Who submits the CSR for the shared identity? In a production environment how is the private key shared so that the CZO is satisfied it's done securely? Is this how all Raft notaries will function?

Relying on the ServiceIdentityGenerator as we have done so far is causing us to not think about this stuff.

shamsasari:
@kchalkias can you also take a look

node-api/src/main/kotlin/net/corda/nodeapi/internal/ServiceIdentityGenerator.kt
```diff
@@ -26,20 +27,44 @@ object ServiceIdentityGenerator {
...
+        return if (singularIdentity) {
+            generateSingular(dirs, issuer, serviceName, serviceId, rootCert)
+        }
+        else generateComposite(dirs, threshold, issuer, serviceName, serviceId, rootCert)
+    }
+
+    private fun generateSingular(dirs: List<Path>, issuer: CertificateAndKeyPair, serviceName: CordaX500Name, serviceId: String, rootCert: Certificate?): Party {
+        val keyPair = generateKeyPair()
+        val notaryKey = keyPair.public
+        dirs.forEach { dir ->
```
kchalkias: We could extract method as most of the logic is similar with `generateComposite`

node/src/integration-test/kotlin/net/corda/node/services/DistributedServiceTests.kt
```diff
@@ -71,68 +75,83 @@ class DistributedServiceTests {
...
                     notarisationsPerNotary.compute(notary) { _, number -> number?.plus(1) ?: 1 }
                 }
             }
         }
 
+        // The distribution of requests should be very close to sg like 16/17/17 as by default artemis does round robin
         println("Notarisation distribution: $notarisationsPerNotary")
         require(notarisationsPerNotary.size == 3)
+        // We allow some leeway for artemis as it doesn't always produce perfect distribution
+        require(notarisationsPerNotary.values.all { it > 10 })
```
kchalkias: If the round robin is strictly followed this will pass, but if it uses a uniform distribution (eg random) with a sample of 50 items and 3 buckets, the probability that `it >10` is not negligible

adagys: This is controlled by Artemis and I don't think it's configurable to be anything else than round robin..

kchalkias: hm.. then why don't we state that `it>15`

adagys: `it>10` seems good enough. (Btw, I didn't introduce this code in this PR, just moved it)

kchalkias: ah ok, if it doesn't break we can leave it for now. Out of curiosity, I will test this in another PR, because we usually handle unit tests that include random distributions differently.

kchalkias:
@shamsasari it turns out that RAFT is used for HA purposes only and it practically reflects one entity. Under this assumption, the threat model doesn't really change if we use the same key (that entity could have done it in the background anyway). All in all, as long as we accept the threat model is not really affected (an assumption that should be documented), IMO it is indeed a nice and convenient solution to easily scale a RAFT notary.

testing/node-driver/src/main/kotlin/net/corda/testing/node/NotarySpec.kt
```diff
@@ -15,7 +15,14 @@ data class NotarySpec(
...
     abstract val clusterSize: Int
 
-    data class Raft(override val clusterSize: Int) : ClusterSpec() {
+    data class Raft(
+            override val clusterSize: Int,
+            /**
+             * If *true*, the cluster will use the same shared "singular" key pair for the service identity,
+             * if *false* – a shared composite public key with individual private keys.
+             */
+            val singularServiceIdentity: Boolean = false
```
shamsasari: Now the question I have to ask is, why even have this as an option? Surely we always want to use a shared key and never a composite key for a Raft cluster?

node-api/src/main/kotlin/net/corda/nodeapi/internal/ServiceIdentityGenerator.kt
```diff
@@ -26,29 +29,52 @@ object ServiceIdentityGenerator {
...
     fun generateToDisk(dirs: List<Path>,
                        serviceName: CordaX500Name,
                        serviceId: String,
                        threshold: Int = 1,
-                       customRootCert: X509Certificate? = null): Party {
-        log.trace { "Generating a group identity \"serviceName\" for nodes: ${dirs.joinToString()}" }
-        val keyPairs = (1..dirs.size).map { generateKeyPair() }
-        val notaryKey = CompositeKey.Builder().addKeys(keyPairs.map { it.public }).build(threshold)
+                       customRootCert: X509Certificate? = null,
+                       singularIdentity: Boolean = false): Party {
```
shamsasari: I would prefer that this be a separate method.

testing/node-driver/src/main/kotlin/net/corda/testing/node/NotarySpec.kt
adagys: It's mainly because Driver does not yet support BFT, and we want to make sure both singular and composite identities for notaries work

shamsasari: Sure, but this the `Raft` class so it won't affect BFT

adagys: Yes but we only have MockNetwork tests for BFT (which skip Artemis), and I remember there were some pain points setting it up with Driver. For the purpose of testing singular vs composite keys it doesn't matter if it's a Raft or BFT notary.

shamsasari: Sorry I'm a bit lost. This `ClusterSpec` class is only used by the driver which only supports Raft. And Raft now will use a shared private key for notarisation. So this parameter doesn't make sense.

adagys: The parameter allows testing the BFT communication path using a Raft notary (we don't care what the notary internals are in this case). This is kind of a workaround due to the fact that the BFT notary was slightly flaky when used with Driver

node-api/src/main/kotlin/net/corda/nodeapi/internal/ServiceIdentityGenerator.kt
```diff
@@ -26,29 +29,52 @@ object ServiceIdentityGenerator {
...
 
+        log.trace { "Generating a group identity \"serviceName\" for nodes: ${dirs.joinToString()}" }
         val caKeyStore = loadKeyStore(javaClass.classLoader.getResourceAsStream("certificates/cordadevcakeys.jks"), "cordacadevpass")
         val issuer = caKeyStore.getCertificateAndKeyPair(X509Utilities.CORDA_INTERMEDIATE_CA, "cordacadevkeypass")
         val rootCert = customRootCert ?: caKeyStore.getCertificate(X509Utilities.CORDA_ROOT_CA)
 
+        return if (singularIdentity) {
+            generateSingular(dirs, issuer, serviceName, serviceId, rootCert)
+        }
+        else generateComposite(dirs, threshold, issuer, serviceName, serviceId, rootCert)
```
shamsasari: braces around this as well.

testing/node-driver/src/main/kotlin/net/corda/testing/node/NotarySpec.kt
shamsasari: The problem I have with this parameter (and it's a problem elsewhere as well) is that the node-driver is a public testing API while this is an internal for our use only. Can you add a TODO here that this parameter needs to be removed before the node-driver is released as a stable module.

adagys: Make sense. Maybe then it's better to create an additional dummy distributed service cluster spec where we have this parameter, and map it to the raft notary internally?

shamsasari: Makes sense. You may want to unseal ClusterSpec so that your subclass can be in the internal package.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -362,23 +363,39 @@ class DriverDSLImpl(
...
+                            serviceName = spec.name,
+                            serviceId = NotaryService.constructId(
+                                    validating = spec.validating,
+                                    raft = true
+                            ),
+                            customRootCert = compatibilityZone?.rootCert
+                    )
+                }
+                is DummyClusterSpec -> {
+                    ServiceIdentityGenerator.generateToDiskSingular(
```
shamsasari: Shouldn't this depend on `singularServiceIdentity`?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -362,23 +363,39 @@ class DriverDSLImpl(
...
+                    ServiceIdentityGenerator.generateToDisk(
+                            dirs = listOf(baseDirectory(spec.name)),
+                            serviceName = spec.name,
+                            serviceId = "identity",
+                            customRootCert = compatibilityZone?.rootCert
+
+                    )
+                }
+                is ClusterSpec.Raft -> {
+                    ServiceIdentityGenerator.generateToDisk(
```
shamsasari: Shouldn't this be `generateToDiskSingular` as it's Raft?

adagys: Yes sorry I'm rebasing now and made a bit of a mess

adagys:
Rebased, ready for another look

node-api/src/main/kotlin/net/corda/nodeapi/internal/DevIdentityGenerator.kt
```diff
@@ -49,32 +52,51 @@ object DevIdentityGenerator {
...
+        val keyPair = generateKeyPair()
+        val notaryKey = keyPair.public
+        dirs.forEach { dir ->
+            generateCertificates(keyPair, notaryKey, intermediateCa, notaryName, dir, rootCert)
         }
+        return Party(notaryName, notaryKey)
+    }
 
-        return Party(notaryName, compositeKey)
+    private fun generateCertificates(keyPair: KeyPair, notaryKey: PublicKey, intermediateCa: CertificateAndKeyPair, notaryName: CordaX500Name, nodeDir: Path, rootCert: Certificate?) {
```
shamsasari: `rootCert` should of type `X509Certificate`

adagys: the `caKeyStore.getCertificate` returns `Certificate` so we'd need to do a cast, does it matter much?

shamsasari: Use the `getX509Certificate` extension method.

node-api/src/main/kotlin/net/corda/nodeapi/internal/DevIdentityGenerator.kt
```diff
@@ -49,32 +52,51 @@ object DevIdentityGenerator {
...
     }
 
-    fun generateDistributedNotaryIdentity(dirs: List<Path>, notaryName: CordaX500Name, threshold: Int = 1): Party {
+    fun generateDistributedNotaryCompositeIdentity(dirs: List<Path>, notaryName: CordaX500Name, threshold: Int = 1): Party {
         require(dirs.isNotEmpty())
 
-        log.trace { "Generating identity \"$notaryName\" for nodes: ${dirs.joinToString()}" }
+        log.trace { "Generating composite identity \"$notaryName\" for nodes: ${dirs.joinToString()}" }
+        val caKeyStore = loadKeyStore(javaClass.classLoader.getResourceAsStream("certificates/cordadevcakeys.jks"), "cordacadevpass")
+        val intermediateCa = caKeyStore.getCertificateAndKeyPair(X509Utilities.CORDA_INTERMEDIATE_CA, "cordacadevkeypass")
```
kchalkias: Maybe we should use const values for passwords "cordacadevkeypass" and "cordacadevpass" as they are re-used by multiple functions. I am also not sure if we should have the same password for intermediateCA and node(notary) keys especially when they are in different keystores. It should be fine for the time being, but we should open a ticket to define a pass policy.

PR #2264, opened 2017-12-15
Put IRS spring whatever back on classpath

exFalso:
this fixes the irsdemotest so it runs from intellij

long story short: the test was moved into a `web` subproject here https://github.com/corda/corda/commit/44a7d872d85e78efa0f06008bdedce27a569f1c5#diff-7609688af2331ce7e4e470a8ab25e90d
spring couldn't find the `RestWhateverThing`

PR #2271, opened 2017-12-18
Fix non-visible IRS demo bug

exFalso:
This example code was run as part of the IRS demo

The failure was intermittent because it's run on a codepath that may not notify the initiating flow in time.

### DIFF
--- a/samples/irs-demo/cordapp/build.gradle
+++ b/samples/irs-demo/cordapp/build.gradle
@@ -94,6 +94,14 @@ task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
         rpcUsers = ext.rpcUsers
         useTestClock true
     }
+    node {
+        name "O=Regulator,L=Moscow,C=RU"
+        p2pPort 10010
+        rpcPort 10011
+        cordapps = ["${project.group}:finance:$corda_release_version"]
+        rpcUsers = ext.rpcUsers
+        useTestClock true
+    }
 }
 
 task integrationTest(type: Test, dependsOn: []) {
--- a/samples/irs-demo/cordapp/src/main/kotlin/net/corda/irs/flows/AutoOfferFlow.kt
+++ b/samples/irs-demo/cordapp/src/main/kotlin/net/corda/irs/flows/AutoOfferFlow.kt
@@ -71,8 +71,8 @@ object AutoOfferFlow {
             // and because in a real life app you'd probably have more complex logic here e.g. describing why the report
             // was filed, checking that the reportee is a regulated entity and not some random node from the wrong
             // country and so on.
-            val regulator = serviceHub.identityService.partiesFromName("Regulator", true).single()
-            subFlow(ReportToRegulatorFlow(regulator, finalTx))
+            // val regulator = serviceHub.identityService.partiesFromName("Regulator", true).single()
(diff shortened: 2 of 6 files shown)


samples/irs-demo/cordapp/src/main/kotlin/net/corda/irs/flows/AutoOfferFlow.kt
```diff
@@ -71,8 +71,8 @@ object AutoOfferFlow {
             // and because in a real life app you'd probably have more complex logic here e.g. describing why the report
             // was filed, checking that the reportee is a regulated entity and not some random node from the wrong
             // country and so on.
-            val regulator = serviceHub.identityService.partiesFromName("Regulator", true).single()
-            subFlow(ReportToRegulatorFlow(regulator, finalTx))
+            // val regulator = serviceHub.identityService.partiesFromName("Regulator", true).single()
```
josecoll: shouldn't we either add a TODO (to fix at a later date) or remove completely?

exFalso: from what I understand this is here for documentation purposes (is that correct @mikehearn ?), which is why I left in the commented out code.

vkolomeyko: Should this PR not be going into `pre-release-3.0-ENT-RC01` branch?

vkolomeyko:
Should this PR not be going into pre-release-3.0-ENT-RC01 branch?

samples/irs-demo/cordapp/src/main/kotlin/net/corda/irs/flows/AutoOfferFlow.kt
mikehearn: It's supposed to actually work. What's the actual fix?

exFalso: Hmm there is no "Regulator" node... where does that come from?

mikehearn: That's odd. It should be set up in the gradle file I guess. The node doesn't have to do anything other than have the app loaded. Looking at the commit where I added this I don't see any addition to the gradle file though. I remember testing the app and seeing it work though. I wonder if I forgot to add the gradle file to the commit or something like that.

Can you try just duplicating the node block in the relevant gradle file and see if that fixes it? If not then yes comment it out and I'll take a look at fixing it.

exFalso:
@mikehearn I added the missing node (in Main.kt, build.gradle, IRSDemoTest.kt). Also cherry picked and closed https://github.com/corda/corda/pull/2264

samples/irs-demo/cordapp/build.gradle
```diff
@@ -94,6 +94,14 @@ task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
         rpcUsers = ext.rpcUsers
         useTestClock true
     }
+    node {
+        name "O=Regulator,L=Moscow,C=RU"
```
mikehearn: haha, love it

PR #2272, opened 2017-12-18
CORDA-876 MockNetwork no longer leaks serialization env if init fails

andr3ej:


testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -221,11 +222,16 @@ class MockNetwork(private val cordappPackages: List<String>,
...
-        val notaryInfos = generateNotaryIdentities()
-        // The network parameters must be serialised before starting any of the nodes
-        networkParameters = NetworkParametersCopier(testNetworkParameters(notaryInfos))
-        notaryNodes = createNotaries()
+        try {
+            filesystem.getPath("/nodes").createDirectory()
+            val notaryInfos = generateNotaryIdentities()
+            // The network parameters must be serialised before starting any of the nodes
+            networkParameters = NetworkParametersCopier(testNetworkParameters(notaryInfos))
+            notaryNodes = this.createNotaries()
```
andr3ej: "this" should be redundant, but there's an unused warning if i leave it out

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNode.kt
```diff
@@ -131,7 +131,8 @@ class MockNetwork(private val cordappPackages: List<String>,
                   servicePeerAllocationStrategy: InMemoryMessagingNetwork.ServicePeerAllocationStrategy = defaultParameters.servicePeerAllocationStrategy,
                   private val defaultFactory: (MockNodeArgs) -> MockNode = defaultParameters.defaultFactory,
                   initialiseSerialization: Boolean = defaultParameters.initialiseSerialization,
-                  private val notarySpecs: List<NotarySpec> = defaultParameters.notarySpecs) {
+                  private val notarySpecs: List<NotarySpec> = defaultParameters.notarySpecs,
+                  createNotaries: MockNetwork.() -> List<StartedNode<MockNode>> = MockNetwork.Companion::createNotariesImpl) {
```
shamsasari: We're leaking this parameter into the public API. Is there no other way to cause the init block to throw an exception? Perhaps have an internal static boolean field `shouldThrow` (I know, I know, really horrible). I'll try to think of a better suggestion.

andr3ej: oops, forgot about the api. the way i used to solve this sort of problem in java is to simply override the createNotaries method in the test. that would mean making MockNetwork open, but i'm not sure that's a big deal. what do you think?

shamsasari: I think that works.

andr3ej: Done

andr3ej:
i think i broke tests by moving that line in stopNodes, will fix in a bit

PR #2274, opened 2017-12-19
Move unspecifiedCountry to internal

andr3ej:


core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -318,3 +319,8 @@ fun ExecutorService.join() {
         // Try forever. Do not give up, tests use this method to assert the executor has no more tasks.
     }
 }
+
+@Suppress("unused")
+@VisibleForTesting
+val CordaX500Name.Companion.unspecifiedCountry
```
shamsasari: This doesn't need to be an extension property. It should simply be a `const val`

andr3ej: it needs to be clear it's a 2-letter country code and not some other country code. i almost called it unspecifiedX500Country, but no one is ever going to find it that way - this way it's auto-completable from CordaX500Name.

PR #2276, opened 2017-12-19
ENT-1302 Fix debug option in Cordformation

rick-r3:
The monitoring option was overwriting the JVM args to enable debug.  Now they coexist.


rick-r3:
Jose approved for 3.0 RC1.  This is just merging to master

PR #2277, opened 2017-12-19
CORDA-599 Two fewer lateinits

andr3ej:


andr3ej:
abandon - securityManager is perfect motivation for a non-naive approach

PR #2273, opened 2017-12-18
CORDA-891 Convert NodeSchedulerServiceTest into a unit test

andr3ej:
review #2306 first

PR #2278, opened 2017-12-19
CORDA-599 SerializeAsTokenContextImpl no longer depends on ServiceHub

andr3ej:
this keeps getting in my way


.ci/api-current.txt
```diff
@@ -2762,7 +2762,7 @@ public interface net.corda.core.serialization.SerializationWhitelist
   @org.jetbrains.annotations.NotNull public abstract net.corda.core.serialization.SerializationToken toToken(net.corda.core.serialization.SerializeAsTokenContext)
 ##
 public interface net.corda.core.serialization.SerializeAsTokenContext
-  @org.jetbrains.annotations.NotNull public abstract net.corda.core.node.ServiceHub getServiceHub()
+  @org.jetbrains.annotations.NotNull public abstract net.corda.core.node.services.AttachmentStorage getAttachments()
```
andr3ej: users have a more convenient way of getting the ServiceHub, they should never get it this way. ideally AttachmentStorage wouldn't be here either but it's too late for that

node/src/main/kotlin/net/corda/node/services/statemachine/StateMachineManagerImpl.kt
```diff
@@ -127,7 +127,7 @@ class StateMachineManagerImpl(
     // Context for tokenized services in checkpoints
     private lateinit var tokenizableServices: List<Any>
     private val serializationContext by lazy {
-        SerializeAsTokenContextImpl(tokenizableServices, SERIALIZATION_FACTORY, CHECKPOINT_CONTEXT, serviceHub)
+        SerializeAsTokenContextImpl(tokenizableServices, SERIALIZATION_FACTORY, CHECKPOINT_CONTEXT, serviceHub.attachments)
```
andr3ej: ideally SMMI should be more specific and take AttachmentStorage in its constructor, but that's another pr

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -114,10 +115,10 @@ class NodeSchedulerServiceTest : SingletonSerializeAsToken() {
                 }).whenever(it).networkMapCache
                 doReturn(myInfo).whenever(it).myInfo
                 doReturn(kms).whenever(it).keyManagementService
-                doReturn(CordappProviderImpl(CordappLoader.createWithTestPackages(listOf("net.corda.testing.contracts")), MockAttachmentStorage())).whenever(it).cordappProvider
+                doReturn(CordappProviderImpl(CordappLoader.createWithTestPackages(listOf("net.corda.testing.contracts")), attachments)).whenever(it).cordappProvider
                 doReturn(NodeVaultService(testClock, kms, validatedTransactions, database.hibernateConfig)).whenever(it).vaultService
                 doReturn(this@NodeSchedulerServiceTest).whenever(it).testReference
-
+                doReturn(attachments).whenever(it).attachments
```
andr3ej: this will all die, see #2273

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -134,7 +135,7 @@ interface SerializationContext {
     /**
      * Helper method to return a new context based on this context with the property added.
      */
-    fun withProperty(property: Any, value: Any): SerializationContext
+    fun <V : Any> withProperty(property: SerializationPropertyKey<V>, value: V): SerializationContext
```
andr3ej: technically this breaks api. do users actually use these methods? if they do i can easily revert this file, but we lose some type safety.

core/src/main/kotlin/net/corda/core/serialization/SerializationToken.kt
```diff
@@ -19,36 +16,21 @@ import net.corda.core.serialization.SingletonSerializationToken.Companion.single
...
-    fun toToken(context: SerializeAsTokenContext): SerializationToken
-}
-
-/**
- * This represents a token in the serialized stream for an instance of a type that implements [SerializeAsToken].
- */
-interface SerializationToken {
-    fun fromToken(context: SerializeAsTokenContext): Any
-}
+interface SerializeAsToken
```
andr3ej: this is really SingletonSerializeAsToken but i haven't renamed it to preserve api

core/src/main/kotlin/net/corda/core/serialization/SerializationToken.kt
```diff
@@ -58,8 +40,4 @@ class SingletonSerializationToken private constructor(private val className: Str
  * A base class for implementing large objects / components / services that need to serialize themselves to a string token
  * to indicate which instance the token is a serialized form of.
  */
-abstract class SingletonSerializeAsToken : SerializeAsToken {
-    private val token = singletonSerializationToken(javaClass)
-
-    override fun toToken(context: SerializeAsTokenContext) = token.registerWithContext(context, this)
-}
+abstract class SingletonSerializeAsToken : SerializeAsToken
```
andr3ej: keeping this as it's api, but the token juggling doesn't make sense as the SerializeAsTokenContextImpl always registers against the class name

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/SerializationScheme.kt
```diff
@@ -40,14 +42,14 @@ data class SerializationContextImpl(override val preferredSerializationVersion:
      * We need to cache the AttachmentClassLoaders to avoid too many contexts, since the class loader is part of cache key for the context.
      */
     override fun withAttachmentsClassLoader(attachmentHashes: List<SecureHash>): SerializationContext {
-        properties[attachmentsClassLoaderEnabledPropertyName] as? Boolean == true || return this
```
andr3ej: redundant cast

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/SerializationScheme.kt
```diff
@@ -40,14 +42,14 @@ data class SerializationContextImpl(override val preferredSerializationVersion:
...
-        val serializationContext = properties[serializationContextKey] as? SerializeAsTokenContextImpl ?: return this // Some tests don't set one.
+        properties[AttachmentsClassLoaderEnabledKey] == true || return this
+        val attachmentStorage = properties[AttachmentStorageKey] as AttachmentStorage
```
andr3ej: replaced dodgy abort with assertion that the AttachmentStorage is there, let's see what the tests say

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -296,9 +297,7 @@ internal fun suitableForObjectReference(type: Type): Boolean {
 /**
  * Common properties that are to be used in the [SerializationContext.properties] to alter serialization behavior/content
  */
-internal enum class CommonPropertyNames {
-    IncludeInternalInfo,
-}
```
andr3ej: no no no no no: half-baked refactorings sprayed over the codebase just cause more work for others. this is why cleanup week does not work @mikehearn

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerializationScheme.kt
```diff
@@ -74,8 +77,11 @@ abstract class AbstractKryoSerializationScheme : SerializationScheme {
     }
 
     private fun <T : Any> withContext(kryo: Kryo, context: SerializationContext, block: (Kryo) -> T): T {
-        kryo.context.ensureCapacity(context.properties.size)
+        kryo.context.ensureCapacity(context.properties.size + 1)
         context.properties.forEach { kryo.context.put(it.key, it.value) }
+        kryo.context.put(SerializationPropertiesKey, object : SerializationProperties {
+            override fun <V : Any> get(key: SerializationPropertyKey<V>) = kryo.serializationProperty(key)
+        })
```
andr3ej: the only reason this is here is to avoid creating the same SerializationProperties object lots of times

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/SerializeAsTokenSerializer.kt
```diff
@@ -6,20 +6,48 @@ import com.esotericsoftware.kryo.Serializer
...
+        val obj = fromToken(kryo, type, token)
+        return type.castIfPossible(obj) ?: throw KryoException("Token read ($token) did not return expected tokenized type: ${type.name}")
+    }
+}
+
+/** A Kryo serializer for [SerializeAsToken] implementations. */
+object SingletonSerializeAsTokenSerializer : SerializeAsTokenSerializer<SerializeAsToken, SingletonSerializationToken>(SingletonSerializationToken::class) {
+    override fun toToken(kryo: Kryo, obj: SerializeAsToken): SingletonSerializationToken {
+        val context = kryo.serializationProperty(SerializeAsTokenContextKey) ?: throw KryoException("Attempt to write a ${SerializeAsToken::class.simpleName} instance of ${obj.javaClass.name} without initialising a context")
+        context.putSingleton(obj)
```
andr3ej: previously a side-effect of obj.toToken

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/SerializationTokenTest.kt
```diff
@@ -74,15 +73,6 @@ class SerializationTokenTest {
...
     }
 
-    @Test(expected = UnsupportedOperationException::class)
-    fun `deserialize unregistered token`() {
-        val tokenizableBefore = UnitSerializeAsToken()
-        val context = serializeAsTokenContext(emptyList<Any>())
-        val testContext = this.context.withTokenContext(context)
-        val serializedBytes = tokenizableBefore.toToken(serializeAsTokenContext(emptyList<Any>())).serialize(factory, testContext)
-        serializedBytes.deserialize(factory, testContext)
-    }
```
andr3ej: bogus test was passing due to UOE on penultimate line thrown by read-only SerializeAsTokenContextImpl

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/SerializationTokenTest.kt
```diff
@@ -105,21 +95,4 @@ class SerializationTokenTest {
...
-
-    @Test(expected = KryoException::class)
-    fun `token returns unexpected type`() {
-        val tokenizableBefore = WrongTypeSerializeAsToken()
-        val context = serializeAsTokenContext(tokenizableBefore)
-        val testContext = this.context.withTokenContext(context)
-        val serializedBytes = tokenizableBefore.serialize(factory, testContext)
-        serializedBytes.deserialize(factory, testContext)
-    }
```
andr3ej: no longer possible in singleton case as the object is no longer responsible for creating the token

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/SerializationTokenTest.kt
```diff
@@ -106,20 +101,26 @@ class SerializationTokenTest {
...
     fun `token returns unexpected type`() {
         val tokenizableBefore = WrongTypeSerializeAsToken()
         val context = serializeAsTokenContext(tokenizableBefore)
         val testContext = this.context.withTokenContext(context)
         val serializedBytes = tokenizableBefore.serialize(factory, testContext)
-        serializedBytes.deserialize(factory, testContext)
+        catchThrowable { serializedBytes.deserialize(factory, testContext) }.run {
+            assertSame(KryoException::class.java, javaClass)
+        }
     }
```
andr3ej: no longer possible in singleton case so i've converted this to test the non-singleton case

andr3ej:
todo: check data persisted by master explorer can be unpersisted by this branch explorer

rick-r3:
I like the changes.  However, as you know, there's API here.  Our policy is to remain cautious and not change the API if we can avoid it.  Any chance we can have the new, more type safe elements in someway for our use but keep the old API around just in case someone used it?  The chances are really slim I think, but we keep finding people doing all kinds of weird and wonderful things out in the wild.

core/src/main/kotlin/net/corda/core/serialization/SerializationToken.kt
rick-r3: Are we able to keep the `SerializeAsToken` API the same?  There's example code that extends from `SingletonSerializeAsToken` (here https://docs.corda.net/oracles.html) and presumably there's an expectation that services could  implement all the methods themselves, for some unknown reason.

andr3ej: * i believe that example code won't break, or even need recompilation
* it's not really safe to allow users to implement toToken the way things are currently, as SerializeAsTokenContextImpl always keys against the class name, so they could have 2 services of the same class configured differently and with carefully distinguished tokens, but they would still be stored against the same key, with surprising consequences.

andr3ej:
closing due to api and as the other changes aren't justifiable on their own, but https://r3-cev.atlassian.net/browse/CORDA-957 should be fixed

PR #2280, opened 2017-12-20
Deprecate TimeWindowChecker, make TimeWindowInvalid report exact current time and transaction time window

adagys:


josecoll:
Please also add equivalent PR to `pre-release-3.0-ENT-RC01`

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -58,7 +58,9 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
 
     fun validateTimeWindow(t: TimeWindow?) {
         if (t != null && !timeWindowChecker.isValid(t))
-            throw NotaryException(NotaryError.TimeWindowInvalid)
+            throw NotaryException(
+                    NotaryError.TimeWindowInvalid(timeWindowChecker.clock.instant(), t)
```
gendal: I'm reviewing this in GitHub without easy access to the codebase so apologies if the answer to this is obvious from a cursory examination of the code but...  does .instant() return the same time that was used to validate the time window?  If so, great. If not, what happens if the transaction was too _early_ by a trivial amount when we validated it but the time is now just fine? Do we risk throwing a nonsensical exception? (eg "12:00:00.000 is outside 11:59:59.999-12:00:00.001")

node/src/main/kotlin/net/corda/node/services/transactions/BFTSMaRt.kt
```diff
@@ -247,7 +247,9 @@ object BFTSMaRt {
 
         protected fun validateTimeWindow(t: TimeWindow?) {
             if (t != null && !timeWindowChecker.isValid(t))
-                throw NotaryException(NotaryError.TimeWindowInvalid)
+                throw NotaryException(
+                        NotaryError.TimeWindowInvalid(timeWindowChecker.clock.instant(), t)
```
gendal: Question completely independent of this PR: how do we handle time windows in the BFT case? Is this code run (in principle - they could be badly behaved I guess...!) by each node in a BFTSMaRT cluster (and it's the submitting node's responsibility to assemble the answers and decide if consensus has been reached?).  Or do we assume the existence of a trusted proxy between Corda nodes and the BFT cluster?  Or something else?

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
adagys: Yes the nonsensical exception is possible, I'll fix that

node/src/main/kotlin/net/corda/node/services/transactions/BFTSMaRt.kt
adagys: it's the former, each BFT replica will return an individual response: either a signature over tx or an error. It's then the client's responsibility to aggregate the responses into a single outcome

core/src/main/kotlin/net/corda/core/node/services/TimeWindowChecker.kt
```diff
@@ -1,11 +0,0 @@
-package net.corda.core.node.services
-
-import net.corda.core.contracts.TimeWindow
-import java.time.Clock
-
-/**
- * Checks if the current instant provided by the input clock falls within the provided time-window.
- */
-class TimeWindowChecker(val clock: Clock = Clock.systemUTC()) {
```
shamsasari: Too late now to move this into internal and to change `isValid`, it's now public API.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -167,7 +168,9 @@ sealed class NotaryError {
     }
 
     /** Thrown if the time specified in the [TimeWindow] command is outside the allowed tolerance. */
-    object TimeWindowInvalid : NotaryError()
+    data class TimeWindowInvalid(val currentTime: Instant, val txTimeWindow: TimeWindow) : NotaryError() {
```
shamsasari: As much as I'd like this change to happen I don't think we can as it's API breaking. What I suggest is that we deprecate this, saying that it's no longer used, and create the new class with a different name.

adagys:
The API stability checker reports 3 changes:
-  public static final net.corda.core.flows.NotaryError$TimeWindowInvalid INSTANCE
-  @org.jetbrains.annotations.NotNull protected abstract net.corda.core.node.services.TimeWindowChecker getTimeWindowChecker()
-  public final void validateTimeWindow(net.corda.core.contracts.TimeWindow)

I believe none of those will cause issues as I just deprecated getTimeWindowChecker(), and changed validateTimeWindow() from `abstract` to `open`

shamsasari:
The one I'm not sure about is the removal of `TimeWindowInvalid.INSTANCE`. It's visible to Java users and I don't think it qualifies as something we can remove and thus break backwards compatibility. @mikehearn can confirm.

core/src/main/kotlin/net/corda/core/node/services/TimeWindowChecker.kt
```diff
@@ -6,6 +6,7 @@ import java.time.Clock
 /**
  * Checks if the current instant provided by the input clock falls within the provided time-window.
  */
+@Deprecated("This class is no longer used and will be removed soon.")
```
mikehearn: We can't remove it. It's a part of our committed API.

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -98,4 +109,15 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
         val signableData = SignableData(txId, SignatureMetadata(services.myInfo.platformVersion, Crypto.findSignatureScheme(notaryIdentityKey).schemeNumberID))
         return services.keyManagementService.sign(signableData, notaryIdentityKey)
     }
+
+    @Deprecated("This property is no longer used, will be removed in future releases") @Suppress("DEPRECATION")
```
mikehearn: Deprecation does not mean it will be removed in future releases.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -167,7 +168,9 @@ sealed class NotaryError {
     }
 
     /** Occurs when time specified in the [TimeWindow] command is outside the allowed tolerance. */
-    object TimeWindowInvalid : NotaryError()
+    data class TimeWindowInvalid(val currentTime: Instant, val txTimeWindow: TimeWindow) : NotaryError() {
```
mikehearn: This change isn't backwards compatible. It's very unlikely to cause any real issue so I'd let it slide, but there is probably an easy way to avoid the break:

```
companion object {
   // Here only for binary compatibility purposes, do not use.
    @JvmField @Deprecated val INSTANCE = TimeWindowInvalid(Instant.EPOCH, Instant.EPOCH)
}
```

Does that work?

mikehearn:
Sorry for the slow review.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
adagys:
```
-  public static final net.corda.core.flows.NotaryError$TimeWindowInvalid INSTANCE
+  @kotlin.jvm.JvmField @org.jetbrains.annotations.NotNull public static final net.corda.core.flows.NotaryError$TimeWindowInvalid INSTANCE
```
should be fine, right?

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -98,4 +109,15 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
         val signableData = SignableData(txId, SignatureMetadata(services.myInfo.platformVersion, Crypto.findSignatureScheme(notaryIdentityKey).schemeNumberID))
         return services.keyManagementService.sign(signableData, notaryIdentityKey)
     }
+
+    @Deprecated("This property is no longer used") @Suppress("DEPRECATION")
+    protected open val timeWindowChecker: TimeWindowChecker get() = throw UnsupportedOperationException("No default implementation, need to override")
+
+    @Deprecated("No longer used, might produce an inaccurate error messages", ReplaceWith("Use NotaryService.validateTimeWindow instead")) @Suppress("DEPRECATION")
+    fun validateTimeWindow(t: TimeWindow?) {
```
shamsasari: Why are we deprecating this? Surely you just want to update it to use the new version of the `TimeWindowInvalid` class.

adagys: I can make it delegate to `NotaryService.validateTimeWindow`. I introduced the static method since I want to avoid code duplication between BFT and non-BFT notaries. BFT notaries don't extend `TrustedAuthorityNotaryService`

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
mikehearn: yep

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -27,6 +28,23 @@ abstract class NotaryService : SingletonSerializeAsToken() {
...
         }
+
+        /**
+         * Checks if the current instant provided by the clock falls within the specified time window.
+         *
+         * @throws NotaryException if current time is outside the specified time window. The exception contains
+         *                         the [NotaryError.TimeWindowInvalid] error.
+         */
+        @JvmStatic
+        fun validateTimeWindow(clock: Clock, timeWindow: TimeWindow?) {
```
shamsasari: Add `@Throws` for the NotaryException

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -98,4 +111,7 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
         val signableData = SignableData(txId, SignatureMetadata(services.myInfo.platformVersion, Crypto.findSignatureScheme(notaryIdentityKey).schemeNumberID))
         return services.keyManagementService.sign(signableData, notaryIdentityKey)
     }
+
+    @Deprecated("This property is no longer used") @Suppress("DEPRECATION")
+    protected open val timeWindowChecker: TimeWindowChecker get() = throw UnsupportedOperationException("No default implementation, need to override")
```
shamsasari: The exception message here doesn't match the deprecation message. If this is no longer used then we shouldn't be overriding this, right?

adagys: This message is for the case when someone for some reason still needs to use timechecker even though it's deprecated. They will get the additional warning that it's deprecated anyway.

adagys:
@mikehearn can you override the failing api stability check pls?

shamsasari:
You need to update the `current-api.txt` file accordingly

PR #2282, opened 2017-12-21
CordFormation adopts new rpc security format in node.conf [CORDA-877]

IgorNitto:
* Cordformation now uses new format for configuring rpc access control in node configuration files    
* Added some syntactic sugar to create a list of rpc users in node dsl and used it in demos
* Modified NodeWebServer to startup with new config structure
* Incremented plugins version number


gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -283,4 +302,15 @@ open class Cordform : DefaultTask() {
 
     private fun Node.logFile(name: String): Path = this.logDirectory().resolve(name)
     private fun ProcessBuilder.addEnvironment(key: String, value: String) = this.apply { environment().put(key, value) }
+
+    // An element of node DSL describing a user.
+    // @see listOfUsers
+    internal class NodeUser(var name: String? = null) {
```
sollecitom: Why are `username` and `password` nullable with default values?

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -283,4 +302,15 @@ open class Cordform : DefaultTask() {
...
+        var password: String? = null
+        var permissions: List<String> = mutableListOf<String>()
+        fun toPropertyMap() = mapOf(
```
sollecitom: Might be good to specify that the result type here is `Map<String, Any?>` by explicitly declared it.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -95,7 +97,15 @@ class Node(private val project: Project) : CordformNode() {
     }
 
     private fun configureProperties() {
-        config = config.withValue("rpcUsers", ConfigValueFactory.fromIterable(rpcUsers))
+        if (rpcUsers != null) {
+            config = config.withValue("security", ConfigValueFactory.fromMap(mapOf(
```
sollecitom: Perhaps extract this into a separate private function.

samples/attachment-demo/build.gradle
```diff
@@ -35,12 +35,18 @@ dependencies {
 }
 
 task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
-    ext.rpcUsers = [['username': "demo", 'password': "demo", 'permissions': ["StartFlow.net.corda.attachmentdemo.AttachmentDemoFlow",
-                                                                             "InvokeRpc.wellKnownPartyFromX500Name",
-                                                                             "InvokeRpc.attachmentExists",
-                                                                             "InvokeRpc.openAttachment",
-                                                                             "InvokeRpc.uploadAttachment",
-                                                                             "InvokeRpc.internalVerifiedTransactionsFeed"]]]
+    ext.rpcUsers = listOfUsers {
```
sollecitom: I mightily prefer this version to the current one, but why are we changing it? Do we still support the current format?

webserver/src/main/kotlin/net/corda/webserver/WebServerConfig.kt
```diff
@@ -5,18 +5,23 @@ import net.corda.core.utilities.NetworkHostAndPort
...
 import net.corda.nodeapi.internal.config.getValue
+import net.corda.nodeapi.internal.config.parseAs
 import java.nio.file.Path
 
 /**
  * [baseDirectory] is not retrieved from the config file but rather from a command line argument.
  */
-class WebServerConfig(override val baseDirectory: Path, val config: Config) : NodeSSLConfiguration {
+class WebServerConfig(override val baseDirectory: Path, val config: Config)
+    : NodeSSLConfiguration {
```
sollecitom: Why was this moved to new line?

webserver/src/main/kotlin/net/corda/webserver/WebServerConfig.kt
```diff
@@ -5,18 +5,23 @@ import net.corda.core.utilities.NetworkHostAndPort
...
     override val keyStorePassword: String by config
     override val trustStorePassword: String by config
     val exportJMXto: String get() = "http"
     val useHTTPS: Boolean by config
     val myLegalName: String by config
     val rpcAddress: NetworkHostAndPort by config
     val webAddress: NetworkHostAndPort by config
-    val rpcUsers: List<User> by config
+    val runAs: User // TODO: replace with credentials supplied by a user
```
sollecitom: Can this user come from the constructor of the configuration type? This would make it far more re-usable and testable.

webserver/src/main/kotlin/net/corda/webserver/internal/NodeWebServer.kt
```diff
@@ -205,10 +205,9 @@ class NodeWebServer(val config: WebServerConfig) {
     }
 
     private fun connectLocalRpcAsNodeUser(): CordaRPCOps {
-        val rpcUser = config.rpcUsers.firstOrNull() ?: throw IllegalArgumentException("The node config has not specified any RPC users")
-        log.info("Connecting to node at ${config.rpcAddress} as $rpcUser")
+        log.info("Connecting to node at ${config.rpcAddress} as ${config.runAs}")
         val client = CordaRPCClient(config.rpcAddress)
-        val connection = client.start(rpcUser.username, rpcUser.password)
+        val connection = client.start(config.runAs.username, config.runAs.password)
```
sollecitom: It'd be great to have an extension function `client.startAs(user: User)` so that this line can become `val connection = client.startAs(config.user)`.

samples/attachment-demo/build.gradle
sollecitom: As discussed, please leave one of the demos with the old `[[user1]]` format, to "test" that it's still compatible.

IgorNitto: The current format is still there. This is simply a macro to express the same thing with a different syntax.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
IgorNitto: This is because the Gradle framework, which expect these fields to be publicly accessible AFAIK.

webserver/src/main/kotlin/net/corda/webserver/WebServerConfig.kt
IgorNitto: Sounds good. See next commit!

samples/attachment-demo/build.gradle
IgorNitto: Kept plain format in previous demo.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -54,6 +55,24 @@ open class Cordform : DefaultTask() {
...
+     *
+     * myUsers = listOfUsers {
+     *     userName1 {
+     *        password = "<password>"
+     *        permissions = ["<permission1>", ...]
+     *    }
+     *    ...
+     * }
+     */
+    fun listOfUsers(usersConfig: Closure<*>): List<Map<String, Any?>> {
```
sollecitom: Is this safe in terms of empty block or not specified?

sollecitom:
Docs and `changelog.rst`.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
IgorNitto: I checked that an empty "listOfUsers {}" will produce an empty list of users in node.conf, while using an incorrect field inside a user block will result in a compile-time error.

samples/attachment-demo/build.gradle
Clintonio: The DSL name here should be just "users". The API documentation can specify that it returns a list. This is similar to the `cordapps` block we have else where.

samples/irs-demo/cordapp/build.gradle
```diff
@@ -45,19 +45,19 @@ dependencies {
...
-                     "StartFlow.net.corda.irs.flows.AutoOfferFlow\$Requester",
-                     "StartFlow.net.corda.irs.flows.UpdateBusinessDayFlow\$Broadcast",
-                     "StartFlow.net.corda.irs.api.NodeInterestRates\$UploadFixesFlow",
-                     "InvokeRpc.vaultQueryBy",
-                     "InvokeRpc.networkMapSnapshot",
-                     "InvokeRpc.currentNodeTime",
-                     "InvokeRpc.wellKnownPartyFromX500Name"
-             ]]
-    ]
+    ext.rpcUsers = listOfUsers {
```
Clintonio: I know you're not the original author but given we want best practices in this code can you change this to "def rpcUsers = ". Putting users on the extension makes no sense here and has been cargo culted around the codebase.

IgorNitto:
At last, I had to revert the DSL element for rpcUsers for now because gradle plugins have changed in the meantime to a point that adding it back would be quite complicated and this has very low priority now.

The change is therefore restricted to the config output, which now uses the new format. I validated it using one of the samples to validate it.

PR #2286, opened 2017-12-21
Adds a caveat regarding the inability to record a transaction twice.

joeldudleyr3:


docs/source/tutorial-observer-nodes.rst
```diff
@@ -35,8 +35,14 @@ the transaction to the regulator. There are two important aspects to note here:
...
-   for. Future versions of Corda may address this issue, but for now, if you wish to both participate in the ledger
-   and also observe transactions that you can't sign for you will need to run two nodes and have two separate
-   identities.
\ No newline at end of file
+Caveats
+-------
+
+* Nodes which act as both observers and direct participants in the ledger are not supported at this time. In
+  particular, coin selection may return states which you do not have the private keys to be able to sign for. Future
+  versions of Corda may address this issue, but for now, if you wish to both participate in the ledger and also observe
```
gendal: I know this is already merged but, strictly, the comma should be after the 'but', not before it.  (Mental test: excise the parenthetical commas and the text contained within and check if the remaining sentence is coherent).

(And since we're here, I suppose we should also fix the split infinitive. How about: "... if you with to participate both in.... and ...)

PR #2290, opened 2017-12-21
CORDA-886: Removed the 60s timeout when generating node-info files in…

shamsasari:
… the network bootstrapper. Instead a warning is given.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -4,12 +4,15 @@ import com.typesafe.config.ConfigFactory
 import net.corda.cordform.CordformNode
 import net.corda.core.identity.Party
 import net.corda.core.internal.*
+import net.corda.core.internal.concurrent.fork
 import net.corda.core.node.NodeInfo
 import net.corda.core.serialization.SerializationContext
 import net.corda.core.serialization.deserialize
 import net.corda.core.serialization.internal.SerializationEnvironmentImpl
 import net.corda.core.serialization.internal._contextSerializationEnv
 import net.corda.core.utilities.ByteSequence
+import net.corda.core.utilities.getOrThrow
+import net.corda.core.utilities.seconds
 import net.corda.nodeapi.internal.SignedNodeInfo
 import net.corda.nodeapi.internal.serialization.AMQP_P2P_CONTEXT
 import net.corda.nodeapi.internal.serialization.SerializationFactoryImpl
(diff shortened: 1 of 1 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -82,15 +86,22 @@ class NetworkBootstrapper {
...
+                }
+                nodeDir.list { paths -> paths.filter { it.fileName.toString().startsWith("nodeInfo-") }.findFirst().get() }
             }
-            nodeDir.list { paths -> paths.filter { it.fileName.toString().startsWith("nodeInfo-") }.findFirst().get() }
+        }
+
+        return try {
+            future.getOrThrow(60.seconds)
+        } catch (e: TimeoutException) {
+            println("...still waiting. Something may have gone wrong with one of the nodes - check the logs")
```
joeldudleyr3: I think this message will induce a lot of people on slow machines to think that they're encountering an error.

Could we change it to, "...still waiting. If this process is taking longer than usual, check the node logs."

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -82,15 +86,22 @@ class NetworkBootstrapper {
...
-            check(process.waitFor(timeOutInSeconds, SECONDS)) {
-                "Node in ${nodeDir.fileName} took longer than ${timeOutInSeconds}s to generate its node-info - see logs in ${nodeDir / LOGS_DIR_NAME}"
-            }
-            check(process.exitValue() == 0) {
-                "Node in ${nodeDir.fileName} exited with ${process.exitValue()} when generating its node-info - see logs in ${nodeDir / LOGS_DIR_NAME}"
+        val executor = Executors.newSingleThreadExecutor()
+
+        val future = executor.fork {
+            processes.zip(nodeDirs).map { (process, nodeDir) ->
+                check(process.waitFor() == 0) {
```
joeldudleyr3: What about changing this so that each node reports back with a message once it finishes generating its node-info file? This would make the user feel like they're making progress and let them know that nothing's gone wrong.

shamsasari: I was thinking that but then I think it'll be too spammy, and I think all the nodes will finished pretty much at the same time so probably not that useful.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
shamsasari: Will do.

shamsasari:
Back to you.

PR #2291, opened 2017-12-22
Refactor database properties in mock services

szymonsztuka:
* Align the use of parameters for `makeTestDataSourceProperties` method with one in Enterprise repo.
* Move database constants (the properties names in config file) from MockServices to a separate encapsulating Util class.

### DIFF
--- a/testing/test-utils/src/main/kotlin/net/corda/testing/database/DataSourcePropertiesUtil.kt
+++ b/testing/test-utils/src/main/kotlin/net/corda/testing/database/DataSourcePropertiesUtil.kt
@@ -1,0 +1,47 @@
+package net.corda.testing.database
+
+import com.typesafe.config.Config
+import com.typesafe.config.ConfigFactory
+import net.corda.testing.database.DataSourceProperties.CONFIG_ENTRY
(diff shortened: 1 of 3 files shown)


testing/test-utils/src/main/kotlin/net/corda/testing/database/DataSourcePropertiesUtil.kt
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Creates a dataSourceProperties entry of [Config].
+ * @param url JDBC connection string, value for dataSource.url key
+ * @param driverClass JDBC Driver class name, value for dataSource.url key
+ * @param user database login, value for dataSource.user key
+ * @param password database password, value for dataSource.password key
+ */
+fun dataSourceConfig(url: String, driverClass: String, user: String, password: String) : Config {
```
shamsasari: I don't understand why you need to make this a `Config` object only for it to be converted to a `Property` using the above method. Why not simply have this method create the `Property`?

szymonsztuka:
Closed, will be obsolete because of PR #2508.

PR #2298, opened 2017-12-27
Validating the entire cert path in node registration, rather just che…

shamsasari:
…cking the root cert.

Also reduced duplicate code when creating the node CA cert path for testing.

### DIFF
--- a/node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
+++ b/node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
@@ -21,6 +21,7 @@ import net.corda.nodeapi.internal.serialization.kryo.KryoHeaderV0_1
 import net.corda.testing.ALICE_NAME
 import net.corda.testing.BOB_NAME
 import net.corda.testing.TestIdentity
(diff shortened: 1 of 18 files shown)


node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -169,10 +170,10 @@ class X509UtilitiesTest {
             override val trustStorePassword = "trustpass"
         }
 
-        val (rootCert, intermediateCa) = createRootCertAndIntermediateCa()
+        val (rootCert, intermediateCa) = createDevIntermediateCaCertPath()
```
kchalkias: `rootCert` -> `rootCa`

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -205,11 +206,11 @@ class X509UtilitiesTest {
...
```
kchalkias: ditto

shamsasari:
I will do the rename in my current branch.

PR #2302, opened 2017-12-29
[CORDA-890] Adding tests to check EdDSA compatibility with RFC8032

kchalkias:
RFC8032 _Edwards-Curve Digital Signature Algorithm (EdDSA)_ mentions various variants of EdDSA.
This PR includes unit-tests based on the test vectors provided in RFC8032 and show that i2p implementation is indeed compatible with PureEdDSA Ed25519.

### DIFF
--- a/core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
+++ b/core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
@@ -27,8 +27,8 @@ import kotlin.test.*
  */
 class CryptoUtilsTest {
 
-    val testString = "Hello World"
-    val testBytes = testString.toByteArray()
+    private val testString = "Hello World"
+    private val testBytes = testString.toByteArray()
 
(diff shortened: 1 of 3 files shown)


core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
```diff
@@ -27,8 +27,8 @@ import kotlin.test.*
  */
 class CryptoUtilsTest {
 
-    val testString = "Hello World"
-    val testBytes = testString.toByteArray()
+    private val testString = "Hello World"
+    private val testBytes = testString.toByteArray()
```
rnicoll: If you're clearing these up anyway, they should probably be `const` too.

kchalkias: thanks, just wanted to get rid of IDE warnings, but it's true the majority of tests classes need to follow your recommendation.

PR #2307, opened 2018-01-02
CORDA-892: Make cordform test use new network bootstrapper logic

anthonykr3:
Make the cordform task
- generate config files in the root dir
- extract the corda jar to the root dir
- run the network bootstrapper to create the node directories
- place the cordapps and any additional config in the generated node directories.

### DIFF
--- a/gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
+++ b/gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
@@ -22,7 +22,6 @@ import java.util.jar.JarInputStream
 @Suppress("unused")
 open class Cordform : DefaultTask() {
     private companion object {
-        @JvmStatic
         val nodeJarName = "corda.jar"
         private val defaultDirectory: Path = Paths.get("build", "nodes")
     }
--- a/gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
+++ b/gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
@@ -169,10 +169,13 @@ open class Cordform : DefaultTask() {
                 val node = node { }
                 it.accept(node)
                 node.additionalCordapps.addAll(cordapps)
+                node.rootDir(directory)
             }
-        }
-        nodes.forEach {
-            it.rootDir(directory)
+            cd.setup { nodeName -> project.projectDir.toPath().resolve(getNodeByName(nodeName)!!.nodeDir.toPath()) }
+        } else {
+            nodes.forEach {
+                it.rootDir(directory)
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -92,7 +92,10 @@ class NetworkBootstrapper {
 
     private fun extractCordaJarTo(directory: Path): Path {
         val cordaJarPath = (directory / "corda.jar")
-        Thread.currentThread().contextClassLoader.getResourceAsStream("corda.jar").copyTo(cordaJarPath)
+        if (!cordaJarPath.exists()) {
```
anthonykr3: The cordform task pulls in the network bootstrapper class directly, not via the bootstrapper jar, so it doesn't have access to the corda jar in that method. Instead the cordform task just copies it itself into the root directory and then the bootstrapper can ignore it.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -132,9 +134,26 @@ open class Cordform : DefaultTask() {
...
         installRunScript()
-        nodes.forEach(Node::build)
         bootstrapNetwork()
+        nodes.forEach(Node::build)
+    }
+
+    /**
+     * Installs the corda fat JAR to the root directory, for the network bootstrapper to use.
+     */
+    private fun installCordaJar() {
```
anthonykr3: Moved installCordaJar from the node to the task, it is only done once, the bootstrapper copies it into each node directory.

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
```diff
@@ -22,6 +22,8 @@ import java.util.jar.JarInputStream
 @Suppress("unused")
 open class Cordform : DefaultTask() {
     private companion object {
+        @JvmStatic
```
shamsasari: Not sure that we need to make this Jvm static. Also this can be private

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -24,10 +24,31 @@ class Cordformation : Plugin<Project> {
                     .single { it.name == "classpath" }
                     .find { it.name.contains("cordformation") }
             return project.rootProject.resources.text
-                    .fromArchiveEntry(archive, filePathInJar)
+                    .fromArchiveEntry(archive!!, filePathInJar)
```
shamsasari: Use `first` instead of `find` above.

shamsasari:
Do you also want to take a quick look @Clintonio ?

PR #2312, opened 2018-01-03
CORDA-882 - Better err messages when serializer encounters private property

fenryka:


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -81,10 +82,18 @@ private fun <T : Any> propertiesForSerializationFromConstructor(kotlinConstructo
...
+
         val matchingProperty = properties[name] ?:
-                throw NotSerializableException("No property matching constructor parameter named '$name' of '$clazz'. " +
-                        "If using Java, check that you have the -parameters option specified in the Java compiler. " +
-                        "Alternately, provide a proxy serializer (SerializationCustomSerializer) if recompiling isn't an option")
+                try {
+                    clazz.getDeclaredField(param.name)
+                    throw NotSerializableException("Property '$name' is private, this renders class '$clazz' unserializable")
+                }
+                catch (e: NoSuchFieldException) {
```
rick-r3: Formatting: put this after the `}` on the line above.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -81,10 +82,18 @@ private fun <T : Any> propertiesForSerializationFromConstructor(kotlinConstructo
     val rc: MutableList<PropertySerializer> = ArrayList(kotlinConstructor.parameters.size)
     for (param in kotlinConstructor.parameters) {
         val name = param.name ?: throw NotSerializableException("Constructor parameter of $clazz has no name.")
+
         val matchingProperty = properties[name] ?:
-                throw NotSerializableException("No property matching constructor parameter named '$name' of '$clazz'. " +
-                        "If using Java, check that you have the -parameters option specified in the Java compiler. " +
-                        "Alternately, provide a proxy serializer (SerializationCustomSerializer) if recompiling isn't an option")
+                try {
+                    clazz.getDeclaredField(param.name)
```
rick-r3: Field and getter are not the same thing, so I suspect that is exploiting kotlin + private behaviour.  What happens if the property is `protected` ? Also if we back with a field with a different name?  More niche, but possible.

I suspect in the tests we can rename the backing field for the kotlin property to demonstrate the latter.  Might require non-data class.  See [here](https://kotlinlang.org/docs/reference/properties.html#backing-properties).

Can kotlin reflection help here?  I'm assuming that would work, and then we can check the visibility of a property `KClass.memberProperties`, `KProperty.visibility != PUBLIC`.

Need to check how it behaves with Java beans though

fenryka: If the field is private then there is no getter! so... does that matter?

node-api/src/test/java/net/corda/nodeapi/internal/serialization/amqp/ErrorMessageTests.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Test
+    public void testJavaConstructorAnnotations() {
+        SerializerFactory factory1 = new SerializerFactory(AllWhitelist.INSTANCE, ClassLoader.getSystemClassLoader());
+        SerializationOutput ser = new SerializationOutput(factory1);
+
+
+        Assertions.assertThatThrownBy(() -> ser.serialize(new C1(1)))
+                .isInstanceOf(NotSerializableException.class)
+                .hasMessage("Property 'a' or it's getter is non public, this renders class 'class net.corda.nodeapi.internal.serialization.amqp.ErrorMessageTests$C1' unserializable -> class net.corda.nodeapi.internal.serialization.amqp.ErrorMessageTests$C1");
```
rick-r3: resolve the class name in the expected message in a similar way to the kotlin tests?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -81,10 +81,17 @@ private fun <T : Any> propertiesForSerializationFromConstructor(kotlinConstructo
...
+                    throw NotSerializableException("Property '$name' or it's getter is non public, this renders class '$clazz' unserializable")
```
gendal: its, not it's

PR #2313, opened 2018-01-03
[CORDA-894] Use Metadata's advertised scheme to identify signature algorithm

kchalkias:
- Previously `findSignatureScheme(key: PublicKey)` was used to identify the scheme type when verifying signatures. However, we could use attached Metadata's `schemeNumberID` to bypass `findSignatureScheme`. Security-wised this is more correct, as we inherently check if the Metadata contains the expected scheme. At the moment we validate public keys and param matching with the advertised metadata scheme during signing and verification.
- a descriptive log output has been added to `doVerify()`, required for debugging purposes.
- [BUG fix] fixed a mismatch Metadata issue caused when signing initial transaction and create signature with CompositeKeys (i.e., in BFT Notaries). Now the first matching key (of the `CompositeKey` is used as key_type in Metadata).
- Decongest `Crypto.kt` by moving/grouping private methods outside the Crypto object (as `internal`s), in the same way as we did for `providerMap`.
- Tests that Metadata matching works as expected.


kchalkias:
I will close this PR as it gets very big and it should be split into smaller PRs. One of them is https://github.com/corda/corda/pull/3069

PR #2315, opened 2018-01-04
Harmonises the names of the parameters across CashExitFlow constructors.

joeldudleyr3:
Previously, one constructor used `issueRef`, and the other used `issuerRef`. This is confusing when starting the flow from the CRaSH shell and using named parameters.


joeldudleyr3:
Reverted port allocation change. API stability tests failed due to missing space on agent. Re-running now.

PR #2322, opened 2018-01-05
Comparison V1-V2 - Do not merge

sollecitom:
Do not merge!


kchalkias:
This is an experimental PR, required for identifying differences between v1 and v2. It seems that we (with @sollecitom) managed to identify the source of the problem, which is related to some serialisation inconsistency between two versions and it's not a cryptography-related side-effect. We are still working on documenting the case and we are examining potential workarounds (one of them already worked).

gendal:
As discussed on Friday with you and @kchalkias, I *think* we're OK on v1 and v2 as we did not commit to wire stability (Even though we _hoped_ we delivered it)

But we WILL be committing to it from 3.0.  In particular, states and transactions persisted using 3.0 (including the dev preview) need to be deserialisable by all future versions of Corda (both OS and enterprise).  So I suspect the work you've done here could be very useful.  As you commented to me, we need to get the equivalent of the API Stability checker implemented asap.

sollecitom:
Hi Richard,

Let's discuss this on Monday. The work we're doing on docker would greatly
help us in this sense.

Enjoy the weekend,

Michele

PR #2323, opened 2018-01-05
CORDA-902 - Serialization documentation

fenryka:
As promised after yesterday

### DIFF
--- a/docs/source/serialization.rst
+++ b/docs/source/serialization.rst
@@ -285,15 +285,16 @@ serialised form
 Setter Instantiation
 ''''''''''''''''''''
 
-An alternative to constructor based initialisation Corda can also determine the important elements of an
-object by inspecting the getter and setter methods present on a class. If a class has only a default
+As an alternative to constructor based initialisation Corda can also determine the important elements of an
+object by inspecting the getter and setter methods present on a class. If a class has **only** a default
 constructor **and** properties then the serializable properties will be determined by the presence of
-both a getter and setter for that property that are both publicly visible.
+both a getter and setter for that property that are both publicly visible. I.e. the class adheres to
+the classic *idiom* of mutable JavaBeans.
 
 On deserialization, a default instance will first be created and then, in turn, the setters invoked
 on that object to populate the correct values.
 
-This is mostly relevant to certain java idioms where classes define only getters and setters. For example:
+For example:
 
 .. container:: codeset
 
--- a/docs/source/serialization.rst
+++ b/docs/source/serialization.rst
@@ -210,22 +210,108 @@ Here are the rules to adhere to for support of your own types:
 Classes
 ```````
 
+General Rules
+'''''''''''''
+
+    #.  The class must be compiled with parameter names included in the ``.class`` file.  This is the default in Kotlin
+        but must be turned on in Java (``-parameters`` command line option to ``javac``).
+    #.  The class is annotated with ``@CordaSerializable``.
+    #.  The declared types of constructor arguments, getters, and setters must be supported, and where generics are used the
+        generic parameter must be a supported type, an open wildcard (``*``), or a bounded wildcard which is currently
+        widened to an open wildcard.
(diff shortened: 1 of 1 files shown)


docs/source/serialization.rst
```diff
@@ -210,23 +210,109 @@ Here are the rules to adhere to for support of your own types:
...
+            }
+
+    .. note:: This is particularly relevant as IDE's can often point out where they believe a
+        property can be made private without knowing this can break Corda serialization. Should
+        this happen then a run time warning will be generated when the class fails to serialize
+
+Setter Instantiation
+''''''''''''''''''''
+
+An alternative to constructor based initialisation Corda can also determine the important elements of an
```
rick-r3: An -> As an

docs/source/serialization.rst
```diff
@@ -210,23 +210,109 @@ Here are the rules to adhere to for support of your own types:
...
+object by inspecting the getter and setter methods present on a class. If a class has only a default
```
rick-r3: only -> **only**

docs/source/serialization.rst
```diff
@@ -210,23 +210,109 @@ Here are the rules to adhere to for support of your own types:
...
+constructor **and** properties then the serializable properties will be determined by the presence of
+both a getter and setter for that property that are both publicly visible.
```
rick-r3: Add a sentence to say something like "i.e. The class adheres to the classic idiom of mutable JavaBeans.`

docs/source/serialization.rst
fenryka: done

docs/source/serialization.rst
fenryka: done

docs/source/serialization.rst
fenryka: done

docs/source/serialization.rst
```diff
@@ -210,23 +210,110 @@ Here are the rules to adhere to for support of your own types:
...
+used to first determine which properties of an object are to be serialised then, on deserialization, it is used to
+instantiate the object with the serialized values.
+
+This is the recommended design idiom for serializable objects in Corda as it allows for immutable state objects to
+be created
+
+    #.  A Java Bean getter for each of the properties in the constructor, with the names matching up.  For example, for a constructor
+        parameter ``foo``, there must be a getter called ``getFoo()``.  If the type of ``foo`` is boolean, the getter may
+        optionally be called ``isFoo()``.  This is why the class must be compiled with parameter names turned on.
+    #.  A constructor which takes all of the properties that you wish to record in the serialized form.  This is required in
```
gendal: This makes it sound like we're OK with people holding information in an object that doesn't get serialised.  I guess this might be OK in general... but inside state objects it feels dangerous (or maybe just confusing).  eg I could inadvertently(?) create an issuance transaction that looks fine to me (because it contains a state object I've created that contains all the info needed to make the relevant verify() function happy.  But, because the object, when deserialised in the future, isn't guaranteed to contain all the same information, maybe verify() fails then.  So there's a difference between the behaviour of the system when operating on objects that have been created directly rather than recreated via deserialisation.  Or am I misunderstanding?

docs/source/serialization.rst
```diff
@@ -210,23 +210,110 @@ Here are the rules to adhere to for support of your own types:
...
+    .. sourcecode:: kotlin
+
+        data class Example (val a: Int, val b: String) {
+            var c: Int = 20
+        }
+
+        var e = Example (10, "hello")
+        e.c = 100;
+
+        val e2 = e.serialize().deserialize() // e2.c will be 20, not 100!!!
```
gendal: eek..!  This feels like it's going to be a never-ending source of "learning opportunities" for developers.  Do we need to add a ton more debug logging? eg something that makes it really easy for developers to figure out how their objects have been serialised and what a deserialised object looks like?

PR #2331, opened 2018-01-07
Making sure none of the interactive shell code is executed if sshd is not configured

shamsasari:


shamsasari:
The local shell is not run if in dev mode or if the no-local-shell flag is provided.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -280,8 +281,10 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
      */
     protected abstract fun getRxIoScheduler(): Scheduler
 
-    open fun startShell(rpcOps: CordaRPCOps) {
-        InteractiveShell.startShell(configuration, rpcOps, securityManager, _services.identityService, _services.database)
+    private fun startShell(rpcOps: CordaRPCOps) {
+        if (configuration.sshd != null) {
+            InteractiveShell.startShell(configuration, rpcOps, securityManager, _services.identityService, _services.database)
```
m4ksio: This method needs to be called if local shell is being run - which may happen regardless of SSHD configuration.

PR #2346, opened 2018-01-09
CORDA-881: Signed network parameters has the network map cert attache…

shamsasari:
…d to it instead of just the public key.

Introduced DigitalSignatureWithCert and SignedDataWithCert as internal APIs, with the expectation that they will become public.


core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
```diff
@@ -0,0 +1,24 @@
...
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.serialization.SerializedBytes
+import net.corda.core.serialization.deserialize
+import net.corda.core.utilities.OpaqueBytes
+import java.security.cert.X509Certificate
+
+// TODO: Rename this to DigitalSignature.WithCert once we're happy for it to be public API
+/** A digital signature that identifies who the public key is owned by, and the certificate which provides prove of the identity */
+class DigitalSignatureWithCert(val by: X509Certificate, bytes: ByteArray) : DigitalSignature(bytes) {
+    fun verify(content: ByteArray): Boolean = by.publicKey.verify(content, this)
```
kasiastreich: `@Throws(SignatureException::class)` annotation needed on all 3 `verify` functions in this file.

core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun verify(content: OpaqueBytes): Boolean = by.publicKey.verify(content.bytes, this)
+}
+
+@CordaSerializable
+class SignedDataWithCert<T : Any>(val raw: SerializedBytes<T>, val sig: DigitalSignatureWithCert) {
+    fun verified(): T {
```
kasiastreich: Shouldn't we also check the cert path?

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkMap.kt
```diff
@@ -55,28 +51,8 @@ data class NetworkParameters(
...
-        X509Utilities.validateCertificateChain(trustedRoot, signature.by, trustedRoot)
-        return raw.deserialize()
-    }
-}
-
-// TODO: This class should reside in the [DigitalSignature] class.
-// TODO: Removing the val from signatureBytes causes serialisation issues
-/** A digital signature that identifies who the public key is owned by, and the certificate which provides prove of the identity */
-class DigitalSignatureWithCert(val by: X509Certificate, val signatureBytes: ByteArray) : DigitalSignature(signatureBytes)
+fun <T : Any> SignedDataWithCert<T>.verifiedNetworkMapCert(rootCert: X509Certificate): T {
```
kasiastreich: I would expect that `SignedDataWithCert` does certificate path check, not that it's done by extension function.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
```diff
@@ -106,27 +85,21 @@ class NetworkMapServer(cacheTimeout: Duration,
...
         @Consumes(MediaType.APPLICATION_OCTET_STREAM)
         fun publishNodeInfo(input: InputStream): Response {
-            return try {
-                val registrationData = input.readBytes().deserialize<SignedNodeInfo>()
-                val nodeInfo = registrationData.verified()
-                val nodeInfoHash = nodeInfo.serialize().sha256()
-                nodeInfoMap.put(nodeInfoHash, registrationData)
-                ok()
-            } catch (e: Exception) {
+            return try {val signedNodeInfo = input.readBytes().deserialize<SignedNodeInfo>()
```
kasiastreich: Move `val signedNodeInfo...` to the new line.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
```diff
@@ -106,27 +85,21 @@ class NetworkMapServer(cacheTimeout: Duration,
...
+            signedNodeInfo.verified()
+
+            nodeInfoMap[signedNodeInfo.raw.hash] = signedNodeInfo
+             ok()} catch (e: Exception) {
```
kasiastreich: Move `} catch...` to the new line.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkMap.kt
shamsasari: Adding this to `SignedDataWithCert` implies that the attached cert is always the sub-cert of the root, which is not that useful as a general API. The other alternative is to include the CertPath but that's wasteful as we already have the root cert.

core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
shamsasari: For now it's OK without as it's internal. If this becomes public API then the annotations should be added. I'll update the TODO.

shamsasari:
Comments addressed. Also updated the network map docs.

docs/source/network-map.rst
```diff
@@ -1,84 +1,82 @@
...
-+----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /network-map/parameters/{hash}    | Retrieve ``NetworkParameters`` object with the same hash.                                                                                              |
-+----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
+Network parameters are a set of values that every node participating in the network needs to agree on and use to
+correctly interoperate with each other. If the node is using the HTTP network map service then on first startup it will
+download the signed network parameters and store them in its ``network-parameters`` file.
 
-TODO: Access control of the network map will be added in the future.
+.. note:: By submitting its ``NodeInfo`` to the network map and downloading the network parameters the node operator
```
shamsasari: @gendal is it OK to say this?

docs/source/network-map.rst
```diff
@@ -1,84 +1,82 @@
...
+  is automatically agreeing to the parameter values.
 
+.. warning:: If the ``network-parameters`` file is changed and no longer matches what the network map service is advertising
+  then the node will automatically shutdown. Resolution to this to delete the incorrect file and download the network
```
gendal: Resolution to this IS to delete...  (Aside:  what happens if a new netparams file is published by the netmapsvc (maybe in error?) Every node in the network shuts down???   do we perhaps need some sort of safety catch... eg whereby a netparams file signals that a new version will be forthcoming (which a node is expected to log prominently)?  (maybe this needs to be a jira to make sure we've figured out how params changes would work in practice?) sorry is this is already all designed/implemented.

docs/source/network-map.rst
```diff
@@ -1,84 +1,82 @@
...
-+----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /network-map/parameters/{hash}    | Retrieve ``NetworkParameters`` object with the same hash.                                                                                              |
-+----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
+Network parameters are a set of values that every node participating in the network needs to agree on and use to
+correctly interoperate with each other. If the node is using the HTTP network map service then on first startup it will
+download the signed network parameters and store them in its ``network-parameters`` file.
 
-TODO: Access control of the network map will be added in the future.
+.. note:: By submitting its ``NodeInfo`` to the network map and downloading the network parameters the node operator
+  is automatically agreeing to the parameter values.
```
gendal: good question... what are they actually doing?   do we have a matching concept in a node, which is the range of parameters that the node operator is prepared to support?  and then the netparams file "forces" or "snaps" the node to those settings?   the confusion is perhaps because I don't understand what a node _does_ with this file.  Does it reconfigure itself to match them?  if not, how does an operator know in advance how to configure their node so that it's compatible?

gendal: if we don't actually have all of this built out, maybe say what the future direction is planned to be?  eg "In the current implementation, it is assumed that a node operator has configured their node in advance to comply with these parameters; in a future version they will be enforced" or something like that?

docs/source/network-map.rst
```diff
@@ -1,84 +1,82 @@
...
-            val nodeInfoHashes: List<SecureHash>,
-            val networkParametersHash: SecureHash
-        }
+The ``additional-node-infos`` directory
+---------------------------------------
 
-The object contains list of node info hashes and hash of the network parameters data structure (without the signatures).
+Along side the HTTP network map service, or as a replacement if the node isn't connected to one, the node polls the
+contents of the ``additional-node-infos`` directory located in its base directory. Each file is expected to be the same
+serialised signed ``NodeInfo`` object that's received from the network map service. These are automtically added to the
```
gendal: that IS received? or that WOULD BE received if the remote node in question had published normally (and the local node had received it)

docs/source/network-map.rst
```diff
@@ -1,84 +1,82 @@
...
+  parameters again.
 
-The ``additional-node-infos`` directory
----------------------------------------
-Each Corda node reads, and continuously polls, the files contained in a directory named ``additional-node-infos`` inside the node base directory.
+.. note:: A future release will support the notion of network parameters changes.
 
-Nodes expect to find a serialized ``SignedNodeInfo`` object, the same object which is sent to network map server.
+If the node isn't using a HTTP network map service then it's expected the signed file is provided by some other means.
+For such a scenerio there is the network bootstrapper tool which in addition to generating the network parameters file
```
gendal: scenario

docs/source/network-map.rst
shamsasari: Yes, the whole process of network parameter change has already been designed for, just not yet implemented. The upgrade process is two-step where the upgrade is announced beforehand.

In the scenario where the network map server incorrectly starts to advertise the wrong hash for the network parameters then yes all the nodes shutdown. The quick resolution would be to shutdown the server and then reboot the nodes (they will continue to use the cached parameters that they have). The reason for the shutdown feature is simply because nodes shouldn't operate if the network parameters don't line up.

docs/source/network-map.rst
shamsasari: There's no configuration required with regards to network parameters. They're decided (dictated?) by the CZO and the node operator has to agree to them. The parameters themselves slot into the node nade without any modification (e.g. the set of notaries available to CorDapps, the max msg size, etc).

I guess when the node operator signs up to the CZ and agrees to the T&Cs, they also agree to the (current) network parameters of the CZ? So perhaps this paragraph isn't needed. The reason I ask is that the node doesn't halt before in applies the network parameters and says "These are the network parameters of the CZ you've just joined. Do you agree to them?" If they say "no" then the node simply terminates there.

gendal: thanks - good to ask.  My sense is that, in the future, we would want an operator to have the option to tell its node what parameters (or ranges) it is allowed to accept... and if the params it receives match then to adopt them, otherwise halt.  with respect to current state, i guess we should definitely log what we receive (assuming we don't already!) and maybe even print them to the console?  if only as a way of 'teaching' admins that these are shortly to become a thing and begin to get people to be aware of them?

gendal: for now, perhaps document it purely technically?  ie "the node will adopt the parameters that it receives (or, if only some of them are wired in, say which ones)".   ie if the node operator doesn't like this fact they can raise it with us... but the docs can then remain silent on anything to do with legals/contracting, where we probably shouldn't be expressing a view in any case?

docs/source/network-map.rst
```diff
@@ -1,84 +1,82 @@
 Network Map
 ===========
 
-The network map stores a collection of ``NodeInfo`` objects, each representing another node with which the node can interact.
-There are two sources from which a Corda node can retrieve ``NodeInfo`` objects:
+The network map stores a collection of signed ``NodeInfo`` objects, each representing another node in the compatability zone
```
kasiastreich: Is it worth mentioning who the signer is? I mean, you cannot tamper `NodeInfo` as an operator.

docs/source/network-map.rst
shamsasari: Thanks.

docs/source/network-map.rst
shamsasari: Yep.

docs/source/network-map.rst
```diff
@@ -1,84 +1,82 @@
...
-
-        data class NetworkMap {
-            val nodeInfoHashes: List<SecureHash>,
-            val networkParametersHash: SecureHash
-        }
+The ``additional-node-infos`` directory
+---------------------------------------
 
-The object contains list of node info hashes and hash of the network parameters data structure (without the signatures).
+Along side the HTTP network map service, or as a replacement if the node isn't connected to one, the node polls the
```
kasiastreich: Alongside

shamsasari:
Back to you @kasiastreich.

docs/source/network-map.rst
```diff
@@ -1,84 +1,78 @@
...
-| GET            | /network-map/node-info/{hash}     | Retrieve ``NodeInfo`` object with the same hash.                                                                                                       |
-+----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
-| GET            | /network-map/parameters/{hash}    | Retrieve ``NetworkParameters`` object with the same hash.                                                                                              |
-+----------------+-----------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------------------+
+Network parameters
+------------------
 
-TODO: Access control of the network map will be added in the future.
+Network parameters are a set of values that every node participating in the network needs to agree on and use to
+correctly interoperate with each other. If the node is using the HTTP network map service then on first startup it will
```
kasiastreich: It's not entirely true. We can provide parameters off-line too. Then they have to match the ones advertised by network map. (If the network map is not currently available but we have network parameters we can operate network using files.)

docs/source/network-map.rst
```diff
@@ -1,84 +1,78 @@
...
-Nodes expect to find a serialized ``SignedNodeInfo`` object, the same object which is sent to network map server.
+If the node isn't using a HTTP network map service then it's expected the signed file is provided by some other means.
+For such a scenario there is the network bootstrapper tool which in addition to generating the network parameters file
+also distributes the node info files to the node directories. More information can be found in :doc:`setting-up-a-corda-network`.
 
-Whenever a node starts it writes on disk a file containing its own ``NodeInfo``, this file is called ``nodeInfo-XXX`` where ``XXX`` is a long string.
+The current set of network parameters:
 
-Hence if an operator wants node A to see node B they can pick B's ``NodeInfo`` file from B base directory and drop it into A's ``additional-node-infos`` directory.
+:minimumPlatformVersion: The minimum platform version that the nodes must be running. Any node which is below is will
```
kasiastreich: Typo: "Any node which is below will..."

docs/source/network-map.rst
```diff
@@ -1,84 +1,78 @@
...
+        have their ``NodeInfo`` submission to the network map rejected.
```
kasiastreich: Moreover if you have network parameters obtained by other means, but your platform version doesn't match -> node will shutdown.

docs/source/network-map.rst
shamsasari: That's discussed below. This is talking about first startup on a fresh node.

docs/source/network-map.rst
shamsasari: Actually, this is what happens. The node doesn't even get the chance to submit for it to be rejected!

shamsasari:
Comments addressed.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -40,42 +38,29 @@ class NetworkMapClient(compatibilityZoneURL: URL, private val trustedRoot: X509C
...
 
-    fun getNetworkParameter(networkParameterHash: SecureHash): SignedData<NetworkParameters>? {
-        val conn = URL("$networkMapUrl/network-parameter/$networkParameterHash").openHttpConnection()
-        return if (conn.responseCode == HttpURLConnection.HTTP_NOT_FOUND) {
-            null
-        } else {
-            conn.inputStream.use { it.readBytes() }.deserialize()
-        }
+    fun getNetworkParameters(networkParameterHash: SecureHash): SignedDataWithCert<NetworkParameters> {
+        return URL("$networkMapUrl/network-parameters/$networkParameterHash").openHttpConnection().responseAs()
```
kasiastreich: Is there a PR to enterprise with changed endpoints?

shamsasari: Once this PR is merged I will raise a ENT merge PR and fix everything up.

docs/source/network-map.rst
kasiastreich: You can read network params from file if for some reason network map server is not available. But I guess it's not that important to go into too much detail here.

PR #2349, opened 2018-01-10
CORDA-924 Fix IntegrationTestingTutorial

andr3ej:
* c09a5d4: the fix
* 3ce5436: another fix, turns out it wasn't significant
* c049bdc: don't drop exceptions

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
@@ -61,9 +61,7 @@ class CordaPersistence(
         DatabaseTransactionManager(this)
         // Check not in read-only mode.
         transaction {
-            dataSource.connection.use {
-                check(!it.metaData.isReadOnly) { "Database should not be readonly." }
-            }
+            check(!connection.metaData.isReadOnly) { "Database should not be readonly." }
         }
     }
 
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
@@ -1,6 +1,7 @@
 package net.corda.nodeapi.internal.persistence
 
 import net.corda.core.schemas.MappedSchema
+import net.corda.core.utilities.contextLogger
 import rx.Observable
 import rx.Subscriber
 import rx.subjects.UnicastSubject
@@ -44,6 +45,10 @@ class CordaPersistence(
         schemas: Set<MappedSchema>,
         attributeConverters: Collection<AttributeConverter<*, *>> = emptySet()
 ) : Closeable {
+    companion object {
+        private val log = contextLogger()
+    }
+
     val defaultIsolationLevel = databaseConfig.transactionIsolationLevel
     val hibernateConfig: HibernateConfiguration by lazy {
         transaction {
@@ -90,7 +95,7 @@ class CordaPersistence(
      */
     fun <T> transaction(isolationLevel: TransactionIsolationLevel, statement: DatabaseTransaction.() -> T): T {
         DatabaseTransactionManager.dataSource = this
-        return transaction(isolationLevel, 3, statement)
+        return transaction(isolationLevel, 2, statement)
     }
 
     /**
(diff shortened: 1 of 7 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -56,9 +61,7 @@ class CordaPersistence(
         DatabaseTransactionManager(this)
         // Check not in read-only mode.
         transaction {
-            dataSource.connection.use {
-                check(!it.metaData.isReadOnly) { "Database should not be readonly." }
-            }
```
andr3ej: this wasn't really broken, but we've already got a connection that will be closed for us

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -90,7 +93,7 @@ class CordaPersistence(
      */
     fun <T> transaction(isolationLevel: TransactionIsolationLevel, statement: DatabaseTransaction.() -> T): T {
         DatabaseTransactionManager.dataSource = this
-        return transaction(isolationLevel, 3, statement)
+        return transaction(isolationLevel, 2, statement)
```
andr3ej: same behaviour as before, >= 3 below is now > 2

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -611,9 +609,7 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
         if (props.isNotEmpty()) {
             val database = configureDatabase(props, configuration.database, identityService, schemaService)
             // Now log the vendor string as this will also cause a connection to be tested eagerly.
-            database.transaction {
-                log.info("Connected to ${database.dataSource.connection.metaData.databaseProductName} database.")
-            }
```
andr3ej: the dataSource.connection was never closed but that turned out to be a distinct bug

node/src/test/kotlin/net/corda/node/internal/AbstractNodeTests.kt
```diff
@@ -0,0 +1,91 @@
...
+        val log = mock<Logger>() // Don't care what happens here.
+        // Actually 10 is enough to reproduce old code hang, as pool size is 10 and we leaked 9 connections and 1 is in flight:
+        repeat(100) {
+            logVendorString(database, log)
+        }
+    }
+
+    @Test
+    fun `H2 fix is applied`() {
+        repeat(if (relaxedThoroughness) 1 else 100) {
```
andr3ej: this is too slow to run to completion in every PR

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
exFalso: this is fixed in enterprise too, this caused connection leakage

mnesbit:
I think the datasource factory changes might clash with work that Michele is doing. Please check with his team.

andr3ej:
@sollecitom is this problematic for you? i just need the hack to run before anything touches H2

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -99,34 +102,37 @@ class CordaPersistence(
...
                 val answer = transaction.statement()
                 transaction.commit()
                 return answer
             } catch (e: SQLException) {
-                transaction.rollback()
-                repetitions++
-                if (repetitions >= repetitionAttempts) {
-                    throw e
-                }
+                quietly(transaction::rollback)
```
mnesbit: I am a bit concerned that this retry logic is naive. In my experience failed SQL operations have complex vendor specific rules for if they need retrying and may need things like reconnecting. So perhaps we need a TODO here.

sollecitom:
@andr3j @mnesbit no issues on my side. A bit surprised about the need for the workaround around the hikari datasource but fine by me.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
andr3ej: it's terribly naive. to avoid a rebuild, and because todos are bad, i've raised CORDA-931 for this issue.

mikehearn:
Good debugging!

Is SynchronizedGetPutRemove really necessary? Is Collections.synchronizedMap() too aggressive or something like that?

andr3ej:
the field is of type HashMap, so it has to be a HashMap

PR #2355, opened 2018-01-11
ENT-1383 Memory weight based transaction cache

blsemo:
Make the transaction cache in DBTransactionStorage memory-we…ight based (rather than count based) so large transactions can no longer use an undue amount of memory.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
+++ b/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
@@ -43,6 +43,17 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val sshd: SSHDConfiguration?
     val database: DatabaseConfig
     val useAMQPBridges: Boolean get() = true
+    val transactionCacheSizeBytes: Long get() = defaultTransactionCacheSize
+
+    companion object {
+        // default to at least 8MB and a bit extra for larger heap sizes
+        val defaultTransactionCacheSize: Long = 8388608 + getAdditionalCacheMemory()
+
+        // add 5% of any heapsize over 300MB to the default transaction cache size
+        private fun getAdditionalCacheMemory(): Long {
+            return Math.max((Runtime.getRuntime().maxMemory() - 314572800) / 20, 0)
+        }
+    }
 }
 
 data class DevModeOptions(val disableCheckpointChecker: Boolean = false)
@@ -118,7 +129,8 @@ data class NodeConfigurationImpl(
         override val additionalNodeInfoPollingFrequencyMsec: Long = 5.seconds.toMillis(),
         override val sshd: SSHDConfiguration? = null,
         override val database: DatabaseConfig = DatabaseConfig(initialiseSchema = devMode, exportHibernateJMXStatistics = devMode),
-        override val useAMQPBridges: Boolean = true
+        override val useAMQPBridges: Boolean = true,
+        override val transactionCacheSizeBytes: Long = NodeConfiguration.defaultTransactionCacheSize
         ) : NodeConfiguration {
 
     override val exportJMXto: String get() = "http"
--- a/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
+++ b/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
@@ -14,7 +14,7 @@ import java.net.URL
 import java.nio.file.Path
 import java.util.*
 
-val Int.MB: Long get() = (this as Long) * 1024 * 1024
+val Long.MB: Long get() = this * 1024 * 1024
 
 interface NodeConfiguration : NodeSSLConfiguration {
     // myLegalName should be only used in the initial network registration, we should use the name from the certificate instead of this.
@@ -49,11 +49,11 @@ interface NodeConfiguration : NodeSSLConfiguration {
 
     companion object {
         // default to at least 8MB and a bit extra for larger heap sizes
-        val defaultTransactionCacheSize: Long = 8.MB + getAdditionalCacheMemory()
+        val defaultTransactionCacheSize: Long = 8L.MB + getAdditionalCacheMemory()
 
         // add 5% of any heapsize over 300MB to the default transaction cache size
         private fun getAdditionalCacheMemory(): Long {
-            return Math.max((Runtime.getRuntime().maxMemory() - 300.MB) / 20, 0)
+            return Math.max((Runtime.getRuntime().maxMemory() - 300L.MB) / 20, 0)
         }
     }
 }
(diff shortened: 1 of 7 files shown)


node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -43,6 +43,17 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val sshd: SSHDConfiguration?
     val database: DatabaseConfig
     val useAMQPBridges: Boolean get() = true
+    val transactionCacheSizeBytes: Long get() = defaultTransactionCacheSize
+
+    companion object {
+        // default to at least 8MB and a bit extra for larger heap sizes
+        val defaultTransactionCacheSize: Long = 8388608 + getAdditionalCacheMemory()
```
tlil: Should we make these parameters customisable through config?

Otherwise, perhaps express the sizes a bit more clearly:
```kotlin
val Int.MB: Int get() = this * 1024 * 1024
// => 
8.MB, 300.MB
```

node/src/main/kotlin/net/corda/node/services/persistence/DBTransactionStorage.kt
```diff
@@ -30,40 +38,66 @@ class DBTransactionStorage : WritableTransactionStorage, SingletonSerializeAsTok
...
+    override fun track(): DataFeed<List<SignedTransaction>, SignedTransaction> {
+        return txStorage.locked {
+            DataFeed(allPersisted().map { it.second.toSignedTx() }.toList(), updatesPublisher.bufferUntilSubscribed().wrapWithDatabaseTransaction())
+        }
+    }
 
     @VisibleForTesting
     val transactions: Iterable<SignedTransaction>
-        get() = txStorage.allPersisted().map { it.second }.toList()
+        get()
```
tlil: A couple of style guide violations – maybe hit "reformat" in IntelliJ #nitpick

node/src/main/kotlin/net/corda/node/utilities/AppendOnlyPersistentMap.kt
```diff
@@ -10,23 +12,17 @@ import java.util.*
...
  */
-class AppendOnlyPersistentMap<K, V, E, out EK>(
+abstract class AppendOnlyPersistentMapBase<K, V, E, out EK>(
         val toPersistentEntityKey: (K) -> EK,
         val fromPersistentEntity: (E) -> Pair<K, V>,
         val toPersistentEntity: (key: K, value: V) -> E,
-        val persistentEntityClass: Class<E>,
-        cacheBound: Long = 1024
-) { //TODO determine cacheBound based on entity class later or with node config allowing tuning, or using some heuristic based on heap size
+        val persistentEntityClass: Class<E>) {
```
tlil: `) {` on new line #nitpick

node/src/main/kotlin/net/corda/node/utilities/AppendOnlyPersistentMap.kt
```diff
@@ -135,3 +131,41 @@ class AppendOnlyPersistentMap<K, V, E, out EK>(
         cache.invalidateAll()
     }
 }
+
+class AppendOnlyPersistentMap<K, V, E, out EK>(toPersistentEntityKey: (K) -> EK,
```
tlil: `@JvmOverloads`?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
blsemo: Good point about the clearer expression.
They are customisable - these are just the defaults for the config.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -14,6 +14,8 @@ import java.net.URL
 import java.nio.file.Path
 import java.util.*
 
+val Int.MB: Long get() = (this as Long) * 1024 * 1024
```
tlil: Think you might be in trouble here and get an "invalid cast from Int to Long" exception. You could do something like this instead though:

```kotlin
val Int.MB: Long get() = this * 1024L * 1024L
```

tlil: Confirmed on my end:
```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
```

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -14,7 +14,7 @@ import java.net.URL
 import java.nio.file.Path
 import java.util.*
 
-val Int.MB: Long get() = (this as Long) * 1024 * 1024
+val Long.MB: Long get() = this * 1024 * 1024
```
tlil: Ah, you fixed this :) Sorry, I looked at the first commit only

node/src/main/kotlin/net/corda/node/utilities/AppendOnlyPersistentMap.kt
```diff
@@ -135,3 +132,41 @@ class AppendOnlyPersistentMap<K, V, E, out EK>(
...
+        toPersistentEntity,
+        persistentEntityClass) {
+    //TODO determine cacheBound based on entity class later or with node config allowing tuning, or using some heuristic based on heap size
+    override val cache = NonInvalidatingCache<K, Optional<V>>(
+            bound = cacheBound,
+            concurrencyLevel = 8,
+            loadFunction = { key -> Optional.ofNullable(loadValue(key)) })
+}
+
+class WeightBasedAppendOnlyPersistentMap<K, V, E, out EK>(toPersistentEntityKey: (K) -> EK,
```
tlil: Nitpick again, sorry. I believe the style guide suggests the following:

```kotlin
class WeightBasedAppendOnlyPersistentMap<K, V, E, out EK>(
        toPersistentEntityKey: (K) -> EK,
        fromPersistentEntity: (E) -> Pair<K, V>,
        toPersistentEntity: (key: K, value: V) -> E,
        persistentEntityClass: Class<E>,
        maxWeight: Long,
        weighingFunc: (K, Optional<V>) -> Int
) : AppendOnlyPersistentMapBase<K, V, E, EK>(
        toPersistentEntityKey,
        fromPersistentEntity,
        toPersistentEntity,
        persistentEntityClass
) {
```

node/src/main/kotlin/net/corda/node/utilities/AppendOnlyPersistentMap.kt
tlil: See formatting comment a few lines down

PR #2356, opened 2018-01-12
Re-enable notary idempotence test

adagys:
It's a bit dangerous leaving this test ignored

### DIFF
--- a/node/src/test/kotlin/net/corda/node/services/transactions/NotaryServiceTests.kt
+++ b/node/src/test/kotlin/net/corda/node/services/transactions/NotaryServiceTests.kt
@@ -101,7 +101,7 @@ class NotaryServiceTests {
     }
 
     @Test
-    fun `should sign identical transaction multiple times (signing is idempotent)`() {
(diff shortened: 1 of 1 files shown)


kchalkias:
I've created [a ticket](https://r3-cev.atlassian.net/browse/CORDA-899) in the past to allow it for EdDSA only or completely remove it as it applies to one algorithm only (that Cloud, HSM and FIPS users cannot even use). This assumption is very tricky, crypto agility and scalability shouldn't assume that anyway. I am not 100% sure what is the best policy here for this one. If we enable it and change the default algorithm it will break, but in the same time we advertise 5 algorithms support.

adagys:
Yes but if don't check for equality, and only that it's a valid sig over the transaction then it's now a problem right?

kchalkias:
You are right on the fact that we should consider "idempotent" boolean results only and not signature outcomes (bytes). What about adding a descriptive comment (+ the one that was previously on `Ignore` to clarify signature bytes might not be idempotent?). We might need to remove the word "idempotent" completely.

PR #2358, opened 2018-01-12
CORDA-928 Caching in the NetworkMapCache

blsemo:
cache query results via `getPeerByLegalName` and `getNodesByIdentityKey` to avoid hitting the DB hard in RPC handling.


rick-r3:
LGTM in terms of what caching we need.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -146,35 +155,36 @@ open class PersistentNetworkMapCache(
...
         logger.info("Adding node with info: $node")
         synchronized(_changed) {
-            registeredNodes[node.legalIdentities.first().owningKey]?.let {
-                if (it.serial > node.serial) {
-                    logger.info("Discarding older nodeInfo for ${node.legalIdentities.first().name}")
-                    return
-                }
-            }
-            val previousNode = registeredNodes.put(node.legalIdentities.first().owningKey, node) // TODO hack... we left the first one as special one
+            val previousNode = getNodesByLegalIdentityKey(node.legalIdentities.first().owningKey).firstOrNull()
```
exFalso: The TODO should stay i think

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -213,24 +222,9 @@ open class PersistentNetworkMapCache(
...
-                addNode(node)
-            } catch (e: Exception) {
-                logger.warn("Exception parsing network map from the database.", e)
-            }
-        }
-    }
-
     private fun updateInfoDB(nodeInfo: NodeInfo, session: Session) {
+        // invalidate cache first - this node info is to be modified, so the cache entry can't be trusted anymore
+        invalidateCaches(nodeInfo)
```
exFalso: we should invalidate *after* the update, otherwise there's a small window between the invalidation and the update where the cache is repopulated with the old value

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -241,6 +235,8 @@ open class PersistentNetworkMapCache(
     }
 
     private fun removeInfoDB(session: Session, nodeInfo: NodeInfo) {
+        // invalidate cache first - this node info is to be removed, so the cache entry can't be trusted anymore
+        invalidateCaches(nodeInfo)
```
exFalso: ditto

PR #2360, opened 2018-01-12
Rename Artemis Queue Addresses to allow future broker agility

mnesbit:
This PR renames the outgoing and inbox queue addresses to enhance broker agility.

First, by using a hash of the public key of the nodes in the address construction we reduce the length of the addresses to a range supported by more brokers. Also, with post-quantum keys this will drastically reduce the size of the string.

Also, by changing the name of the inbox to a topic tied to the node identity we disambiguate the identity and potentially in the future we will be able share the broker between nodes for performance and HA purposes.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/Node.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/Node.kt
@@ -40,6 +40,7 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import rx.Scheduler
 import rx.schedulers.Schedulers
+import java.security.PublicKey
 import java.time.Clock
 import java.util.concurrent.atomic.AtomicInteger
 import javax.management.ObjectName
@@ -166,11 +167,14 @@ open class Node(configuration: NodeConfiguration,
             VerifierType.OutOfProcess -> VerifierMessagingClient(configuration, serverAddress, services.monitoringService.metrics, networkParameters.maxMessageSize)
             VerifierType.InMemory -> null
         }
+        require(info.legalIdentities.size in 1..2) { "Currently nodes must have a primary address and optionally one serviced address" }
+        val serviceIdentity: PublicKey? = if (info.legalIdentities.size == 1) null else info.legalIdentities[1].owningKey
         return P2PMessagingClient(
                 configuration,
                 versionInfo,
                 serverAddress,
                 info.legalIdentities[0].owningKey,
+                serviceIdentity,
                 serverThread,
                 database,
                 advertisedAddress,
--- a/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
+++ b/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
@@ -21,7 +21,6 @@ import net.corda.node.utilities.AffinityExecutor
 import net.corda.node.utilities.AppendOnlyPersistentMap
 import net.corda.node.utilities.PersistentMap
 import net.corda.nodeapi.internal.ArtemisMessagingComponent.*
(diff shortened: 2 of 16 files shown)


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -151,11 +152,13 @@ open class Node(configuration: NodeConfiguration,
             VerifierType.OutOfProcess -> VerifierMessagingClient(configuration, serverAddress, services.monitoringService.metrics, networkParameters.maxMessageSize)
             VerifierType.InMemory -> null
         }
+        val serviceIdentity: PublicKey? = if (info.legalIdentities.size != 2) null else info.legalIdentities[1].owningKey
```
andr3ej: oh god. i guess this is the best you can do when your predecessor hasn't used object oriented programming

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -55,21 +54,28 @@ import javax.persistence.Lob
  * invoke methods on the provided implementation. There is more documentation on this in the docsite and the
  * CordaRPCClient class.
  *
- * @param serverAddress The address of the broker instance to connect to (might be running in the same process).
- * @param myIdentity The public key to be used as the ArtemisMQ address and queue name for the node.
- * @param nodeExecutor An executor to run received message tasks upon.
- * @param advertisedAddress The node address for inbound connections, advertised to the network map service and peers.
- * If not provided, will default to [serverAddress].
+ * @param config The configuration of the node, which is used for controlling the message redelivery options.
+ * @param versionInfo All messages from the node carry the version info and received messages are checked against this for compatability.
```
andr3ej: compatibility

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -55,21 +54,28 @@ import javax.persistence.Lob
...
+ * @param serverAddress The host and port of the Artemis broker.
+ * @param myIdentity The primary identity of the node, which defines the messaging address for externally received messages.
+ * It is also used to construct the myAddress field, which is ultimately advertised in the network map.
+ * @param serviceIdentity An optional second identity if the node is also part of a group address, for example a notary.
+ * @param nodeExecutor The received messages are marshalled onto the server executor to prevent Netty buffers leaking during fiber suspends.
```
andr3ej: thanks for documenting this

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -166,11 +167,14 @@ open class Node(configuration: NodeConfiguration,
             VerifierType.OutOfProcess -> VerifierMessagingClient(configuration, serverAddress, services.monitoringService.metrics, networkParameters.maxMessageSize)
             VerifierType.InMemory -> null
         }
+        require(info.legalIdentities.size in 1..2) { "Currently nodes must have a primary address and optionally one serviced address" }
+        val serviceIdentity: PublicKey? = if (info.legalIdentities.size == 1) null else info.legalIdentities[1].owningKey
```
exFalso: This is quite restrictive... so the two values that are fine for `legalIdentities` are `[peerIdentity]` and `[peerIdentity, serviceIdentity]`? Is this pending on the multiple identities work?

mnesbit: As I understand it this is the current limitation, but it will at least fail to start with a specific message if you break that. Hopefully, multiple identity work is coming soon.

PR #2365, opened 2018-01-15
Updates the documentation on versioning.

joeldudleyr3:


mikehearn:
Nice work Joel, really clear. Can we get Java versions of the code samples too though?

PR #2367, opened 2018-01-15
Minor: tweaks to the discussion of network parameters and network map.

mikehearn:


docs/source/network-map.rst
```diff
@@ -66,13 +66,12 @@ The current set of network parameters:
         not start.
 :notaries: List of identity and validation type (either validating or non-validating) of the notaries which are permitted
         in the compatibility zone.
-:maxMessageSize: Maximum allowed P2P message size sent over the wire in bytes. Any message larger than this will be
-        split up.
-:maxTransactionSize: Maximum permitted transaction size in bytes.
+:maxMessageSize: Maximum allowed P2P message size sent over the wire in bytes. Any message larger than this is forbidden.
```
shamsasari: If that's the case then we need to schedule this work in. Currently this parameter is passed into Artemis' minLargeMessageSize config, which from their docs implies the messages are just split up.
>Any message larger than a certain size is considered a large message. Large messages will be split up and sent in fragments. This is determined by the parameter minLargeMessageSize

I suggest reverting this and adding a TODO to change the wording when this is properly implemented.

docs/source/network-map.rst
```diff
@@ -66,13 +66,12 @@ The current set of network parameters:
...
 :modifiedTime: The time when the network parameters were last modified by the compatibility zone operator.
 :epoch: Version number of the network parameters. Starting from 1, this will always increment whenever any of the
         parameters change.
 
-.. note:: ``maxTransactionSize`` is currently not enforced in the node, but will be in a later release.
```
shamsasari: The reason for this note was that we didn't get time to remove this from the net params struct before shipping the preview release.

docs/source/network-map.rst
mikehearn: Hmm. I had figured this was wired to the 10mb limit we imposed. Wires got crossed somewhere I guess. I'll revert and put in a TODO as you suggest.

docs/source/network-map.rst
mikehearn: Right, we need to figure out the details around max sizes. Transactions are messages. So if the max message size is 10mb then the max tx size is also 10mb (or minus a bit I suppose, with headers and so on). There can be uses for tx sizes lower than the max message size e.g. limited enclave memory or avoiding pathological txns that take insane amounts of time to validate, but that's probably best handled with non-byte based measures if we find we need to go beyond max message size.

docs/source/network-map.rst
shamsasari: I had a look through the code but I couldn't see where we're implementing a hard limit on msg sizes.

PR #2368, opened 2018-01-15
Don't propagate internal notary service errors to the client

adagys:
If the notary backend fails we should return a generic "service unavailable" error instead of propagating potentially cryptic implementation-specific exceptions. This should only be used when the issue can't be solved via internal retries.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -158,7 +158,7 @@ class NotaryFlow {
  */
 data class TransactionParts(val id: SecureHash, val inputs: List<StateRef>, val timestamp: TimeWindow?, val notary: Party?)
 
-class NotaryException(val error: NotaryError) : FlowException("Error response from Notary - $error")
+class NotaryException(val error: NotaryError) : FlowException("Unable to notarise: $error")
 
 @CordaSerializable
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -158,20 +158,25 @@ class NotaryFlow {
...
-    /** Thrown if the time specified in the [TimeWindow] command is outside the allowed tolerance. */
+    /** Occurs when time specified in the [TimeWindow] command is outside the allowed tolerance. */
     object TimeWindowInvalid : NotaryError()
 
     data class TransactionInvalid(val cause: Throwable) : NotaryError() {
         override fun toString() = cause.toString()
     }
 
+    /** Occurs when the notary service cannot connect to the internal data store. */
+    object ServiceUnavailable : NotaryError() {
```
shamsasari: Let's not use `object` in public APIs as it becomes tricky to convert to a `class`. (`object` creates a public static `INSTANCE` field which is visible in Java)

shamsasari: I think this should be a "GeneralError" class. Unlike `Conflict` and `TimeWindowInvalid`, this is not a "notary" level issue which clients would want to query programmatically. Rather it's a system error and I think a string message will suffice.

PR #2369, opened 2018-01-15
Adds default endpoints for the Corda development webserver.

joeldudleyr3:
Provides a bunch of extra standard endpoints to prevent users from having to implement their own to test their CorDapps.

### DIFF
--- a/samples/attachment-demo/build.gradle
+++ b/samples/attachment-demo/build.gradle
@@ -40,7 +40,8 @@ task deployNodes(type: net.corda.plugins.Cordform, dependsOn: ['jar']) {
                                                                              "InvokeRpc.attachmentExists",
                                                                              "InvokeRpc.openAttachment",
                                                                              "InvokeRpc.uploadAttachment",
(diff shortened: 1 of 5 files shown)


joeldudleyr3:
Addressed review comments.

mikehearn:
We should really be deleting this web server, not extending it! it's deprecated and _we tell people not to use it_.

Joel, can you focus on bringing the Spring Boot template or demo or whatever into the tree? I'd like us to kill off the web server entirely soon, so this sort of change is sending developers the wrong message.

joeldudleyr3:
Can we have a discussion on this with @Clintonio and @davejh69 ?

Clinton's view is that if it's not good enough, it should be improved, not scrapped. Dave H and Clinton want a way for CorDapp front-ends to be bundled up into CorDapps (I think). The default webserver allows this.

mikehearn:
Sure, perhaps we can start a discourse thread to discuss it so others can contribute their views too.

PR #2371, opened 2018-01-16
Added conditional to disable tests when CORDA_DOCS_ONLY_BUILD envvar …

Clintonio:
…is present

Why: Reduce build time by ~40m on docs builds.


Clintonio:
We'll need to double check the tests all run before merging this one.

PR #2372, opened 2018-01-16
CORDA-929 Attachment caching

blsemo:
Cache node attachments in a read through cache. The attachment implementation can read the actual content of an attachment lazily, therefore the contents are cached in a second cache that might hold fewer entries (based on memory weight). We don't need to invalidate caches as attachments are immutable.


node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -132,7 +138,9 @@ data class NodeConfigurationImpl(
         override val sshd: SSHDConfiguration? = null,
         override val database: DatabaseConfig = DatabaseConfig(initialiseSchema = devMode, exportHibernateJMXStatistics = devMode),
         override val useAMQPBridges: Boolean = true,
-        override val transactionCacheSizeBytes: Long = NodeConfiguration.defaultTransactionCacheSize
+        override val transactionCacheSizeBytes: Long = NodeConfiguration.defaultTransactionCacheSize,
+        override val attachmentContentCacheSizeBytes: Long = NodeConfiguration.defaultAttachmentContentCacheSize,
```
shamsasari: I'm not sure this pattern of duplicating the default value makes sense. Isn't it sufficient to just specify default the default here?

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -172,11 +182,67 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
...
         val attachment = currentDBSession().get(NodeAttachmentService.DBAttachment::class.java, id.toString())
-        attachment?.let {
-            return AttachmentImpl(id, { attachment.content }, checkAttachmentsOnLoad)
+        return attachment?.content
+    }
+
+
+    private val attachmentCache = NonInvalidatingCache<SecureHash, Optional<Attachment>>(
+            attachmentCacheBound,
+            8,
```
andr3ej: i'm allergic to magic numbers, this should be a named constant

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -172,11 +182,67 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
...
+    private fun createAttachment(key: SecureHash): Attachment? {
+        val content = attachementContentCache.get(key)
+        if (content.isPresent) {
+            return AttachmentImpl(
+                    key,
+                    {
+                        attachementContentCache
+                                .get(key)
+                                .orElseThrow {
+                                    IllegalArgumentException("No attachement impl should have been created for non existent content")
```
andr3ej: another extra e

node/src/test/kotlin/net/corda/node/services/persistence/NodeAttachmentStorageTest.kt
```diff
@@ -80,6 +80,44 @@ class NodeAttachmentStorageTest {
...
+            val storage = NodeAttachmentService(MetricRegistry())
+            val id = testJar.read { storage.importAttachment(it) }
+            assertEquals(expectedHash, id)
+
+
+            assertNull(storage.openAttachment(hashB))
+            val stream = storage.openAttachment(expectedHash)!!.openAsJAR()
+            val e1 = stream.nextJarEntry!!
+            assertEquals("test1.txt", e1.name)
+            assertEquals(stream.readBytes().toString(Charset.defaultCharset()), "This is some useful content")
```
andr3ej: wrong charset, kotlin toByteArray (unlike similar jdk methods) uses utf-8

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -172,11 +182,67 @@ class NodeAttachmentService(metrics: MetricRegistry) : AttachmentStorage, Single
...
+    // to load an attachment from the db, the loader will insert a non present optional - we invalidate this
+    // immediately as we definitely want to retry whether the attachment was just delayed.
+    // On the second level, we cache Attachment implementations that use the first cache to load their content
+    // when required. As these are fairly small, we can cache quite a lot of them, this will make checking
+    // repeatedly whether an attachment exists fairly cheap. Here as well, we evict non-existent entries immediately
+    // to force a recheck if required.
+    // If repeatedly looking for non-existing attachments becomes a performance issue, this is either indicating a
+    // a problem somewhere else or this needs to be revisited.
+
+    private val attachementContentCache = NonInvalidatingWeightBasedCache<SecureHash, Optional<ByteArray>>(
```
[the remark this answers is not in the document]
blsemo: The Optional is required to handle non-existent entries - the cache does not like null pointers and will throw an `InvalidCacheLoadException`

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -41,21 +42,21 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val sshd: SSHDConfiguration?
     val database: DatabaseConfig
     val useAMQPBridges: Boolean get() = true
-    val transactionCacheSizeBytes: Long get() = defaultTransactionCacheSize
-    val attachmentContentCacheSizeBytes: Long get() = defaultAttachmentContentCacheSize
```
andr3ej: is it possible to keep these on the interface and make them do the conversion? we should use bytes throughout the codebase, as it's too confusing otherwise, and expose megabytes only for users. if that turns out to be too messy let's forget about it and just use bytes.

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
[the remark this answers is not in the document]
andr3ej: ah ok

PR #2381, opened 2018-01-17
CORDA-599 Retire test-specific node hooks

andr3ej:


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -783,13 +783,13 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
             override val myInfo: NodeInfo,
-            override val networkMapCache: NetworkMapCacheInternal
+            override val networkMapCache: NetworkMapCacheInternal,
+            override val vaultService: VaultServiceInternal
     ) : SingletonSerializeAsToken(), ServiceHubInternal, StateLoader by validatedTransactions {
         override val rpcFlows = ArrayList<Class<out FlowLogic<*>>>()
         override val stateMachineRecordedTransactionMapping = DBTransactionMappingStorage()
         override val auditService = DummyAuditService()
         override val transactionVerifierService by lazy { makeTransactionVerifierService() }
-        override val vaultService by lazy { makeVaultService(keyManagementService, validatedTransactions, database.hibernateConfig) }
```
andr3ej: turns out this lazy isn't critical, as far as the tests are concerned

node/src/main/kotlin/net/corda/node/services/keys/PersistentKeyManagementService.kt
```diff
@@ -17,6 +17,7 @@ import javax.persistence.Entity
 import javax.persistence.Id
 import javax.persistence.Lob
 
+class InitialKeys(val keys: Set<KeyPair>)
```
andr3ej: something more specific than Set to go in the LazyHub

node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
```diff
@@ -314,8 +316,17 @@ class TwoPartyTradeFlowTests(private val anonymous: Boolean) {
         return mockNet.createNode(MockNodeParameters(legalName = name), nodeFactory = { args ->
             object : MockNetwork.MockNode(args) {
                 // That constructs a recording tx storage
-                override fun makeTransactionStorage(database: CordaPersistence, transactionCacheSizeBytes: Long): WritableTransactionStorage {
-                    return RecordingTransactionStorage(database, super.makeTransactionStorage(database, transactionCacheSizeBytes))
+                private fun makeTransactionStorage(database: CordaPersistence, lhf: LazyHubFactory): RecordingTransactionStorage {
+                    val realStorage = lhf.child().let { lh ->
+                        lh.impl(DBTransactionStorage::class)
```
andr3ej: in both cases the test assumes the default type when in theory it could find it out from AbstractNode somehow. for now, i'm going to file this under "hard things should be possible".

node/src/main/kotlin/net/corda/node/services/keys/E2ETestKeyManagementService.kt
```diff
@@ -27,6 +27,9 @@ import javax.annotation.concurrent.ThreadSafe
 @ThreadSafe
 class E2ETestKeyManagementService(val identityService: IdentityServiceInternal,
                                   initialKeys: Set<KeyPair>) : SingletonSerializeAsToken(), KeyManagementService {
+    @Suppress("unused")
+    constructor(identityService: IdentityServiceInternal, initialKeys: InitialKeys) : this(identityService, initialKeys.keys)
```
Clintonio: Replace original constructor - we're the only consumer here so it makes sense to just make this the de-jure constructor.

andr3ej:
@mikehearn can you undo #2396 ? it's blocking this PR

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
mikehearn: It was there for performance reasons i.e. to try and stop the tests always initialising the vault if a specific test didn't need to. Perhaps the lazyness is always resolved immediately or something, but it isn't surprising the tests work with the lazy removed.

mikehearn:
Closing. As was made clear already we aren't going to have a DI framework, at least not unless the issue is revisited and a different decision is made then (no guarantee the outcome will be any different).

Please stop trying to refactor AbstractNode for now.

PR #2396, opened 2018-01-19
Revert "CORDA-599 RPCSecurityManager is no longer lateinit (#2347)"

mikehearn:
The decision not to introduce DI frameworks was taken in December. LazyHub can be proposed as a candidate framework next time this issue is revisited, which was already planned for the summer.


vkolomeyko:
I understood there is a fair amount of controversy around `LazyHub`.
I think it will be a sensible thing to revert it and discuss more offline. Once agreement is reached we can make a second attempt to introduce this kind of functionality.
I am approving this PR.

PR #2400, opened 2018-01-19
CORDA-915 - Replace BEANS introspector with standard reflection

fenryka:
Removes a lib dependency for a small use case

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
@@ -98,7 +98,8 @@ data class PropertyDescriptor(var field: Field?, var setter: Method?, var getter
 }
 
 object PropertyDescriptorsRegex {
-    val re = Regex("(?<type>get|set|is)(?<var>[A-Z].*)")
+    // match an uppercase letter that also has a corresponding lower case equivalent
+    val re = Regex("(?<type>get|set|is)(?<var>\\p{Lu}.*)")
 }
 
 /**
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
@@ -70,18 +70,18 @@ internal fun constructorForDeserialization(type: Type): KFunction<Any>? {
 internal fun <T : Any> propertiesForSerialization(
         kotlinConstructor: KFunction<T>?,
         type: Type,
-        factory: SerializerFactory) = PropertySerializers.make (
-            if (kotlinConstructor != null) {
-                propertiesForSerializationFromConstructor(kotlinConstructor, type, factory)
-            } else {
-                propertiesForSerializationFromAbstract(type.asClass()!!, type, factory)
-            }.sortedWith(PropertyAccessor))
+        factory: SerializerFactory) = PropertySerializers.make(
+        if (kotlinConstructor != null) {
+            propertiesForSerializationFromConstructor(kotlinConstructor, type, factory)
+        } else {
+            propertiesForSerializationFromAbstract(type.asClass()!!, type, factory)
+        }.sortedWith(PropertyAccessor))
 
 
 fun isConcrete(clazz: Class<*>): Boolean = !(clazz.isInterface || Modifier.isAbstract(clazz.modifiers))
 
 /**
- * Encapsulates the property of a class and it's potential getter and setter methods.
+ * Encapsulates the property of a class and its potential getter and setter methods.
  *
  * @property field a property of a class.
  * @property setter the method of a class that sets the field. Determined by locating
@@ -89,27 +89,44 @@ fun isConcrete(clazz: Class<*>): Boolean = !(clazz.isInterface || Modifier.isAbs
  * @property getter the method of a class that returns a fields value. Determined by
  * locating a function named getXyz for the property named in field as xyz.
  */
-data class PropertyDescriptor (var field: Field?, var setter : Method?, var getter: Method?) {
-    override fun toString() = StringBuilder("Property - ${field?.name ?: "null field"}\n").apply {
+data class PropertyDescriptor(var field: Field?, var setter: Method?, var getter: Method?) {
+    override fun toString() = StringBuilder("").apply {
+        appendln("Property - ${field?.name ?: "null field"}\n")
         appendln("  getter - ${getter?.name ?: "no getter"}")
         append("  setter - ${setter?.name ?: "no setter"}")
     }.toString()
 }
 
+object PropertyDescriptorsRegex {
+    val re = Regex("(?<type>get|set|is)(?<var>[A-Z].*)")
+}
+
 /**
  * Collate the properties of a class and match them with their getter and setter
  * methods as per a JavaBean.
+ *
+ * for a property
+ *      exampleProperty
+ *
+ * We look for methods
+ *      setExampleProperty
+ *      getExampleProperty
+ *      isExampleProperty
+ *
+ * Where setExampleProperty must return a type compatible with exampleProperty, getExampleProperty must
+ * take a single parameter of a type compatible with exampleProperty and isExampleProperty must
+ * return a boolean
  */
-fun Class<out Any?>.propertyDescriptors() : Map<String, PropertyDescriptor> {
+fun Class<out Any?>.propertyDescriptors(): Map<String, PropertyDescriptor> {
     val classProperties = mutableMapOf<String, PropertyDescriptor>()
 
-    var clazz : Class<out Any?>? = this
+    var clazz: Class<out Any?>? = this
     do {
         // get the properties declared on this instance of class
-        clazz?.declaredFields?.forEach { classProperties.put (it.name, PropertyDescriptor(it, null, null)) }
+        clazz!!.declaredFields.forEach { classProperties.put(it.name, PropertyDescriptor(it, null, null)) }
 
         // then pair them up with the declared getter and setter
-        // Note: It is possible for a class to have multple instaces of a function where the types
+        // Note: It is possible for a class to have multipleinstancess of a function where the types
         // differ. For example:
         //      interface I<out T> { val a: T }
         //      class D(override val a: String) : I<String>
@@ -117,8 +134,8 @@ fun Class<out Any?>.propertyDescriptors() : Map<String, PropertyDescriptor> {
         //      getA - returning a String (java.lang.String) and
         //      getA - returning an Object (java.lang.Object)
         // In this instance we take the most derived object
-        clazz?.declaredMethods?.map {
-            Regex("(?<type>get|set)(?<var>[A-Z].*)").find(it.name)?.apply {
+        clazz.declaredMethods?.map {
+            PropertyDescriptorsRegex.re.find(it.name)?.apply {
                 try {
                     classProperties.getValue(groups[2]!!.value.decapitalize()).apply {
                         when (groups[1]!!.value) {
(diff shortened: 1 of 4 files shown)


fenryka:
Those unit test failures might be an actual thing, will look at them on Tuesday

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -78,55 +76,135 @@ internal fun <T : Any> propertiesForSerialization(
...
+/**
+ * Encapsulates the property of a class and it's potential getter and setter methods.
+ *
+ * @property field a property of a class.
+ * @property setter the method of a class that sets the field. Determined by locating
+ * a function called setXyz on the class for the property named in field as xyz.
+ * @property getter the method of a class that returns a fields value. Determined by
+ * locating a function named getXyz for the property named in field as xyz.
+ */
+data class PropertyDescriptor (var field: Field?, var setter : Method?, var getter: Method?) {
```
andr3ej: formatting! and not just here

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -78,55 +76,135 @@ internal fun <T : Any> propertiesForSerialization(
...
+    override fun toString() = StringBuilder("Property - ${field?.name ?: "null field"}\n").apply {
+        appendln("  getter - ${getter?.name ?: "no getter"}")
```
andr3ej: this is great, i had no idea it existed. but you've used a '\n' literal above?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -78,55 +76,135 @@ internal fun <T : Any> propertiesForSerialization(
                 propertiesForSerializationFromAbstract(type.asClass()!!, type, factory)
             }.sortedWith(PropertyAccessor))
 
+
 fun isConcrete(clazz: Class<*>): Boolean = !(clazz.isInterface || Modifier.isAbstract(clazz.modifiers))
 
+/**
+ * Encapsulates the property of a class and it's potential getter and setter methods.
```
andr3ej: its

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -78,55 +76,135 @@ internal fun <T : Any> propertiesForSerialization(
...
+ * Collate the properties of a class and match them with their getter and setter
+ * methods as per a JavaBean.
+ */
+fun Class<out Any?>.propertyDescriptors() : Map<String, PropertyDescriptor> {
+    val classProperties = mutableMapOf<String, PropertyDescriptor>()
+
+    var clazz : Class<out Any?>? = this
+    do {
+        // get the properties declared on this instance of class
+        clazz?.declaredFields?.forEach { classProperties.put (it.name, PropertyDescriptor(it, null, null)) }
```
andr3ej: it's always non-null, so can you use !! as the ? suggests this code is expected to take null sometimes

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -78,55 +76,135 @@ internal fun <T : Any> propertiesForSerialization(
...
+        // Note: It is possible for a class to have multple instaces of a function where the types
+        // differ. For example:
+        //      interface I<out T> { val a: T }
+        //      class D(override val a: String) : I<String>
+        // instances of D will have both
+        //      getA - returning a String (java.lang.String) and
+        //      getA - returning an Object (java.lang.Object)
+        // In this instance we take the most derived object
+        clazz?.declaredMethods?.map {
+            Regex("(?<type>get|set)(?<var>[A-Z].*)").find(it.name)?.apply {
```
andr3ej: regex compilation expensive, should be a static

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -78,55 +76,135 @@ internal fun <T : Any> propertiesForSerialization(
...
+                try {
+                    classProperties.getValue(groups[2]!!.value.decapitalize()).apply {
```
andr3ej: i believe decapitalize uses the system locale, which will be significant for our turkish customers. i guess we want it to be deterministic?

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/GenericsTests.kt
```diff
@@ -277,4 +277,31 @@ class GenericsTests {
...
+    @Test
+    fun implemntsGeneric() {
+        open class B<out T>(open val a: T)
+        class D(override val a: String) : B<String>(a)
+
+        val factory = testDefaultFactoryNoEvolution()
+
+        val bytes = SerializationOutput(factory).serialize(D("Test"))
+
+        DeserializationInput(factory).deserialize(bytes).apply { assertEquals("Test", this.a) }
```
andr3ej: this apply is a bit gratuitous but i'll allow it

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/PrivatePropertyTests.kt
```diff
@@ -56,6 +58,61 @@ class PrivatePropertyTests {
...
+
+            fun setA(a: Int, b: Int) { this.a = a }
+            fun getA() = a
+        }
+
+        val c1 = C(33, 44)
+        Assertions.assertThatThrownBy {
+            SerializationOutput(factory).serialize(c1)
+        }.isInstanceOf(NotSerializableException::class.java).hasMessageContaining(
+                "Defined setter for parameter a takes too many arguments")
```
andr3ej: if like me you hate the fluent style because you're stuck with only the assertions it supports, try `catchThrowable{}.run{}`

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -78,55 +76,135 @@ internal fun <T : Any> propertiesForSerialization(
...
+                        when (groups[1]!!.value) {
+                            "set" -> {
+                                if (setter == null) setter = it
+                                else if (TypeToken.of(setter!!.genericReturnType).isSupertypeOf(it.genericReturnType)) {
+                                    setter = it
+                                }
+                            }
+                            "get" -> {
```
rick-r3: As well as `get`, (bean) properties can have `is` instead when they are boolean.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
fenryka: alas the string it takes at construction doesn't get treated as line breaking, will do it a slightly differnet way

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
fenryka: I was being lazy, good point

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -99,26 +100,39 @@ data class PropertyDescriptor (var field: Field?, var setter : Method?, var gett
...
+ *      isExampleProperty
+ *
+ * Where setExampleProperty must return a type compatible with exampleProperty, getExampleProperty must
+ * take a single parameter of a type compatible with exampleProperty and isExampleProperty must
+ * return a boolean
  */
-fun Class<out Any?>.propertyDescriptors() : Map<String, PropertyDescriptor> {
+fun Class<out Any?>.propertyDescriptors(): Map<String, PropertyDescriptor> {
     val classProperties = mutableMapOf<String, PropertyDescriptor>()
+    val re = Regex("(?<type>get|set|is)(?<var>[A-Z].*)")
```
andr3ej: this will still get compiled every time propertyDescriptors is called, it should be outside the fun

fenryka: done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -72,62 +70,163 @@ internal fun constructorForDeserialization(type: Type): KFunction<Any>? {
...
+ *      getExampleProperty
+ *      isExampleProperty
+ *
+ * Where setExampleProperty must return a type compatible with exampleProperty, getExampleProperty must
+ * take a single parameter of a type compatible with exampleProperty and isExampleProperty must
+ * return a boolean
+ */
+fun Class<out Any?>.propertyDescriptors(): Map<String, PropertyDescriptor> {
+    val classProperties = mutableMapOf<String, PropertyDescriptor>()
+    val re = Regex("(?<type>get|set|is)(?<var>[A-Z].*)")
```
rick-r3: In addition to Andrzej concern about compiling this each time, I'm also thinking about other char sets etc.  I wonder if matching with just the prefix get/set/is here and then doing a toUpper or something more internationalisation friendly in the code (rather than regex) ?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -71,62 +69,166 @@ internal fun constructorForDeserialization(type: Type): KFunction<Any>? {
...
+data class PropertyDescriptor(var field: Field?, var setter: Method?, var getter: Method?) {
+    override fun toString() = StringBuilder("").apply {
+        appendln("Property - ${field?.name ?: "null field"}\n")
+        appendln("  getter - ${getter?.name ?: "no getter"}")
+        append("  setter - ${setter?.name ?: "no setter"}")
+    }.toString()
+}
+
+object PropertyDescriptorsRegex {
+    val re = Regex("(?<type>get|set|is)(?<var>[A-Z].*)")
```
rick-r3: My previous comment might have got lost when this moved - apologies if duplicating.  Maybe we should not require A-Z as the first char, as that doesn't sound very internationalisation friendly.  Maybe it could be any char (non-whitespace) and we check in code rather than regex that it is upper case (or however we're supposed to do that internationalisation friendly).  e.g. `toUpper` or `isUpperCase` or something.

fenryka: can be done with a slightly better regex

fenryka: https://www.regular-expressions.info/unicode.html

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
fenryka: done

PR #2402, opened 2018-01-21
Register run on stop callbacks

nimmaj:

When writing driver based integration tests in the cordapp-template using Braid, there is a nice shutdown mechanism used by Corda to shut all of its services down without shutting down the VM.  Unfortunately there is no mechanism for Braid to know about this so Braid leaves ports open and lying around and the second test is unable to run.

This PR exposes the existing runOnStop infrastructure in AbstractNode to CordaServices via ServiceHub.  I initially put it in ServiceHubInternal but during conversation with Joel we moved it to ServiceHub.

I also wrote tests, admittedly driver based, in the cordapp-template, which worked fine.

# PR Checklist:

- [ Yes] Have you run the unit, integration and smoke tests as described here? https://docs.corda.net/head/testing.html

prtests

- [Yes ] If you added/changed public APIs, did you write/update the JavaDocs?

Probably need reviewing for consistency/correct nomenclature.

- [Yes ] If the changes are of interest to application developers, have you added them to the changelog, and potentially release notes?

- [X ] If you are contributing for the first time, please read the agreement in CONTRIBUTING.md now and add to this Pull Request that you agree to it.

I agree to the current contents of CONTRIBUTING.md.

Look forward to your feedback.

Thanks, Ben

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
@@ -354,4 +354,13 @@ interface ServiceHub : ServicesForResolution {
      * @return A new [Connection]
      */
     fun jdbcSession(): Connection
+
+    /**
+     * Allows the registration of a callback to inform services when corda is shutting down.
+     *
+     * The intent is to allow the cleaning up of resources - e.g. releasing ports.
+     *
+     * You should not rely on this to clean up executing flows - that's what quasar is for.
+     */
+    fun addRunOnStop(runOnStop: () -> Any?)
 }
--- a/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
@@ -356,11 +356,17 @@ interface ServiceHub : ServicesForResolution {
     fun jdbcSession(): Connection
 
     /**
-     * Allows the registration of a callback to inform services when corda is shutting down.
+     * Allows the registration of a callback that may inform services when the app is shutting down.
      *
      * The intent is to allow the cleaning up of resources - e.g. releasing ports.
      *
      * You should not rely on this to clean up executing flows - that's what quasar is for.
+     *
+     * Please note that the shutdown handler is not guaranteed to be called. In production the node process may crash,
+     * be killed by the operating system and other forms of fatal termination may occur that result in this code never
+     * running. So you should use this functionality only for unit/integration testing or for code that can optimise
+     * this shutdown e.g. by cleaning up things that would otherwise trigger a slow recovery process next time the
+     * node starts.
      */
-    fun addRunOnStop(runOnStop: () -> Any?)
+    fun registerUnloadHandler(runOnStop: () -> Unit)
 }
--- a/node/src/integration-test/kotlin/net/corda/node/NodePerformanceTests.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/NodePerformanceTests.kt
@@ -31,13 +31,6 @@ import java.util.*
 import java.util.concurrent.TimeUnit
 import kotlin.streams.toList
 
-
-private fun checkQuasarAgent() {
-    if (!(ManagementFactory.getRuntimeMXBean().inputArguments.any { it.contains("quasar") })) {
-        throw IllegalStateException("No quasar agent")
-    }
-}
-
 @Ignore("Run these locally")
 class NodePerformanceTests {
     @StartableByRPC
@@ -51,11 +44,6 @@ class NodePerformanceTests {
             val flowPerSecond: Double,
             val averageMs: Double
     )
-
-    @Before
-    fun before() {
-        checkQuasarAgent()
-    }
 
     @Test
     fun `empty flow per second`() {
--- a/node/src/integration-test/kotlin/net/corda/node/NodeTests.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/NodeTests.kt
@@ -1,0 +1,59 @@
+package net.corda.node
+
+import net.corda.core.node.ServiceHub
+import net.corda.core.node.services.CordaService
+import net.corda.core.serialization.SingletonSerializeAsToken
+import net.corda.core.utilities.getOrThrow
+import net.corda.core.utilities.loggerFor
+import net.corda.testing.DUMMY_BANK_A_NAME
+import net.corda.testing.driver.driver
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+import java.lang.management.ManagementFactory
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+private fun checkQuasarAgent() {
+    if (!(ManagementFactory.getRuntimeMXBean().inputArguments.any { it.contains("quasar") })) {
+        throw IllegalStateException("No quasar agent")
+    }
+}
+
+val latch = CountDownLatch(1)
+
+@CordaService
+class RunOnStopTestService(private val serviceHub: ServiceHub) : SingletonSerializeAsToken() {
+
+    companion object {
+        private val log = loggerFor<RunOnStopTestService>()
+    }
+
+    init {
+        serviceHub.addRunOnStop(this::shutdown)
+    }
+
+    fun shutdown() {
+        log.info("shutting down")
+        latch.countDown()
+    }
+
+}
+
+class NodeTests {
+
+    @Before
+    fun before() {
+        checkQuasarAgent()
+    }
+
+    @Test
+    fun `should be able to register run on stop lambda`() {
+        driver(startNodesInProcess = true, extraCordappPackagesToScan = listOf("net.corda.node"), isDebug = true) {
+            startNode(providedName = DUMMY_BANK_A_NAME).getOrThrow()
+            // just want to fall off the end of this for the mo...
+        }
+        Assert.assertTrue("Timed out waiting for AbstractNode to invoke the test service shutdown callback",latch.await(30, TimeUnit.SECONDS))
+    }
+
+}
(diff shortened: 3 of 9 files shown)


core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
```diff
@@ -354,4 +354,13 @@ interface ServiceHub : ServicesForResolution {
...
     fun jdbcSession(): Connection
+
+    /**
+     * Allows the registration of a callback to inform services when corda is shutting down.
+     *
+     * The intent is to allow the cleaning up of resources - e.g. releasing ports.
+     *
+     * You should not rely on this to clean up executing flows - that's what quasar is for.
+     */
+    fun addRunOnStop(runOnStop: () -> Any?)
```
mikehearn: Maybe `registerUnloadHandler` is a better name? Again, because that way we can change when the callback is used in future?

Also - rather than use `Any?` which implies the result might be used for something, why not use `Unit`. `Unit` is a special type in Kotlin which is used in place of `void` in Java. So it will be translated to () -> void which is a more helpful API for Java users and people using other JVM languages. I realise the internal list uses Any? because that way we can jam any function at all into it, but for public API we should be more precise in our types. I think you can assign `() -> Unit` to `() -> Any` in Kotlin so it should involve no code changes on your part.

core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
```diff
@@ -354,4 +354,13 @@ interface ServiceHub : ServicesForResolution {
      * @return A new [Connection]
      */
     fun jdbcSession(): Connection
+
+    /**
+     * Allows the registration of a callback to inform services when corda is shutting down.
```
mikehearn: This is a good api doc, but I'd add the following paragraph:

"Please note that the shutdown handler is not guaranteed to be called. In production the node process may crash, be killed by the operating system and other forms of fatal termination may occur that result in this code never running. So you should use this functionality only for unit/integration testing or for code that can optimise the shutdown e.g. by cleaning up things that would otherwise trigger a slow recovery process next time the node starts."

I would also tweak the text so it talks about shutdown of the _app_ not shutdown of the _node_. Subtle distinction without a difference in the current platform, but it ensures that we are not breaking semantic compatibility later if we start running this callback when an app is unloaded on the fly.

node/src/integration-test/kotlin/net/corda/node/NodeTests.kt
```diff
@@ -0,0 +1,59 @@
...
+import net.corda.testing.DUMMY_BANK_A_NAME
+import net.corda.testing.driver.driver
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+import java.lang.management.ManagementFactory
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+private fun checkQuasarAgent() {
```
mikehearn: I realise that this is here because you encountered this problem but it's something that should be asserted inside the driver code. Can you take this out and/or move it into the implementation of the driver DSL? The idea is right, it's just the location that's wrong.

node/src/integration-test/kotlin/net/corda/node/NodeTests.kt
```diff
@@ -0,0 +1,59 @@
...
+    if (!(ManagementFactory.getRuntimeMXBean().inputArguments.any { it.contains("quasar") })) {
+        throw IllegalStateException("No quasar agent")
+    }
+}
+
+val latch = CountDownLatch(1)
```
mikehearn: This can be moved into the companion object - at the top level it'll pollute the global namespace, interfering with code completion and slowing down the IDE.

node/src/integration-test/kotlin/net/corda/node/NodeTests.kt
```diff
@@ -0,0 +1,59 @@
...
+}
+
+class NodeTests {
+
+    @Before
+    fun before() {
+        checkQuasarAgent()
+    }
+
+    @Test
```
mikehearn: Isn't there an existing test file that this can be put in? `NodeTests` is an awfully generic name for a class that only has a single unit test in it - again, this class will show up in code completion and navigation, so it's good for it to have a specific name if it can't be merged with some other test suite.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -57,6 +57,8 @@ open class MockServices private constructor(
         private val initialIdentity: TestIdentity,
         private val moreKeys: Array<out KeyPair>
 ) : ServiceHub, StateLoader by validatedTransactions {
+
```
mikehearn: No need for these extra blank lines.

nimmaj:
Awesome comments - thanks for those.  I believe I've made all the requested updates.  The only contentious change is that I've renamed NodeTests rather than merged into another one.  The obvious candidate to merge with is NodeKeyStoreCheckTest, but merging those two together is likely to create a suite with the name NodeTests again, arguably missing the point of the comment.  I originally called my test NodeTests because I was worried about adding my test to one of the existing named tests!  Do let me know your thoughts around this. Thanks for taking the time to review.

node/src/integration-test/kotlin/net/corda/node/NodeUnloadHandlerTests.kt
```diff
@@ -0,0 +1,48 @@
...
+            // just want to fall off the end of this for the mo...
+        }
+        Assert.assertTrue("Timed out waiting for AbstractNode to invoke the test service shutdown callback",latch.await(30, TimeUnit.SECONDS))
+    }
+
+    @CordaService
+    class RunOnStopTestService(serviceHub: ServiceHub) : SingletonSerializeAsToken() {
+
+        companion object {
+            private val log = loggerFor<RunOnStopTestService>()
```
mikehearn: You can use `contextLogger()` here - it's shorter and survives moves/renames.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -888,6 +889,9 @@ fun <DI : DriverDSL, D : InternalDriverDSL, A> genericDriver(
     val serializationEnv = setGlobalSerialization(initialiseSerialization)
     val shutdownHook = addShutdownHook(driverDsl::shutdown)
     try {
+        if (!(ManagementFactory.getRuntimeMXBean().inputArguments.any { it.contains("quasar") })) {
```
mikehearn: @exFalso Does this place look right to you? I suspect we should only check for this, if in-memory mode is in use - concur?

node/src/integration-test/kotlin/net/corda/node/NodeUnloadHandlerTests.kt
nimmaj: That's much nicer.  Thanks :-)

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
nimmaj: That's interesting because where I've put it, I found I needed to add a line to the build.gradle to make the web server integration tests pass (they're broken in the current version of this PR - apols for that - i didn't realise that pushing to my repo would immediately build here).  So I would be interested in your views.

nimmaj:
Thanks.  Checked in a fix for the breaking test (that may turn out to be irrelevant) and your suggested logging change which is rather nice.  Tests seem to pass locally, though i'm struggling with dangling JVMs - don't know if that's normal?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
exFalso: @mikehearn out-of-process nodes have this argument specified explicitly elsewhere so the spawned jvms won't use the argument that's checked here. This code works anyway because we specify the agent for all tests.
Perhaps move this to `startInProcessNode()` to make it clear that it's a precondition of starting in-process nodes. This will delay the failure until a `startNode` is called, but I don't think that's an issue

node/src/integration-test/kotlin/net/corda/node/NodePerformanceTests.kt
```diff
@@ -31,13 +31,6 @@ import java.util.*
 import java.util.concurrent.TimeUnit
 import kotlin.streams.toList
 
-
-private fun checkQuasarAgent() {
-    if (!(ManagementFactory.getRuntimeMXBean().inputArguments.any { it.contains("quasar") })) {
```
exFalso: Was about to comment that this is already done here when I noticed you updated this code as well! Thorough:)

There is one more test that replicates this behaviour, `NodeStatePersistenceTests`. There this check is done to decide whether to run nodes in or out of process. I believe this is incorrect and may have been left in from a debugging session. Branching on this jvm flag means that running the test from IntelliJ may result in different behaviour than running it from gradle, so I think that code can simply be removed, it should always be out-of-process.

webserver/build.gradle
```diff
@@ -1,5 +1,6 @@
 apply plugin: 'kotlin'
 apply plugin: 'java'
+apply plugin: 'net.corda.plugins.quasar-utils'
```
exFalso: If we move the check to `startInProcessNode` this will perhaps not be needed anymore

nimmaj:
Sorry for the delay - machine was being fixed!  Ok, I've moved the code as suggested and reverted the build.gradle - feels better now - thanks for the suggestion :-)

mikehearn:
Thanks and no need to apologise for any delays, you're the one contributing and we're the one making demands here :)

It looks good to me so I approved it. Once it passes integration testing I'll merge. Thanks for the help!

PR #2404, opened 2018-01-22
CORDA-939 Make SerializationEnvironmentRule.env private so as not to expose internals.

anthonykr3:
Don't expose the internal SerializationEnvironment directly from SerializationEnvironmentRule - just expose the parts of it that are required for testing, that are part of the public api.


testing/test-utils/src/main/kotlin/net/corda/testing/core/SerializationTestHelpers.kt
```diff
@@ -42,7 +44,10 @@ class SerializationEnvironmentRule(private val inheritable: Boolean = false) : T
         }
     }
 
-    lateinit var env: SerializationEnvironment
+    private lateinit var env: SerializationEnvironment
+    lateinit var serializationFactory : SerializationFactory
+    lateinit var checkpointContext : SerializationContext
```
andr3ej: these should be val getters that delegate to env, otherwise all good

anthonykr3: I did that originally, however it made the SerializationApiInJavaTest both fail with "kotlin.UninitializedPropertyAccessException: lateinit property env has not been initialized" - I assume it is something to do with some magical initialisation logic in Kotlin that doesn't exist in Java?

andr3ej: did you try the syntax `val serializationFactory get() = env.serializationFactory`

anthonykr3: I didn't, and that works, thanks very much.

PR #2407, opened 2018-01-22
Add cmdline option for network root truststore and password

patrickkuo:


node/src/integration-test/kotlin/net/corda/node/utilities/registration/NodeRegistrationTest.kt
```diff
@@ -128,8 +128,7 @@ class NodeRegistrationTest {
                 portAllocation = portAllocation,
                 compatibilityZone = compatibilityZone,
                 initialiseSerialization = false,
-                notarySpecs = listOf(NotarySpec(notaryName)),
-                startNodesInProcess = true  // We need to run the nodes in the same process so that we can capture the correct exception
+                notarySpecs = listOf(NotarySpec(notaryName))
```
shamsasari: How is the test now passing if we're not using in-process mode? I don't think the CertPathValidatorException will be propagated in out-of-process mode. Anyway, it doesn't matter you can remove this test. `NetworkRegistrationHelperTest` has the error cases covered anyway.

patrickkuo: oh, this is not a fix for the test, I was confused with the work i have done in another PR, which I have turned on in process to see the node's log for debugging.

I will remove this since its redundant.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -806,7 +809,7 @@ class DriverDSLImpl(
                     "--base-directory=${config.corda.baseDirectory}",
                     "--logging-level=$loggingLevel",
                     "--no-local-shell").also {
-                if (extraCmdLineFlag != null) {
+                if (extraCmdLineFlag.isNotEmpty()) {
```
shamsasari: Minor: You don't actually need this.

PR #2410, opened 2018-01-22
Use message based bridge control for managing AMQP bridges.

mnesbit:
To allow out of process bridges and floats in DMZ zones change from listening for queue creation in the ArtemisMessagingServer and use an explicit pub/sub messaging protocol to manage the bridge lifecycle.

In doing this I have also removed the legacy bridge support, since there do not seem to have been any particular issues reported with the AMQP bridging solution.

The Artemis server component should now also be much less customised, which should help with broker pluggability in the future.


node-api/src/main/kotlin/net/corda/nodeapi/internal/BridgeControlMessages.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import net.corda.core.identity.CordaX500Name
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.utilities.NetworkHostAndPort
+
+@CordaSerializable
+data class BridgeEntry(val queueName: String, val targets: List<NetworkHostAndPort>, val legalNames: List<CordaX500Name>)
+
+sealed class BridgeControl {
+    @CordaSerializable
```
rick-r3: I think it would really help if we kdoc'd here when each message is expected to be sent and what it represents.  e.g. what does `NodeToBridgeSnapshot` represent and when is sent ?

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
```diff
@@ -0,0 +1,86 @@
...
+        stop()
+        bridgeManager.start()
+        val artemis = ArtemisMessagingClient(config, p2pAddress, maxMessageSize)
+        this.artemis = artemis
+        artemis.start()
+        val artemisClient = artemis.started!!
+        val artemisSession = artemisClient.session
+        val bridgeControlQueue = "$BRIDGE_CONTROL.$bridgeId"
+        artemisSession.createTemporaryQueue(BRIDGE_CONTROL, RoutingType.MULTICAST, bridgeControlQueue)
+        val control = artemisSession.createConsumer(bridgeControlQueue)
```
sollecitom: Same here, why this pattern of creating a local variable and assign that to a field?

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
```diff
@@ -0,0 +1,86 @@
...
+    private var controlConsumer: ClientConsumer? = null
+
+    companion object {
+        private val log = contextLogger()
+    }
+
+    fun start() {
+        stop()
+        bridgeManager.start()
+        val artemis = ArtemisMessagingClient(config, p2pAddress, maxMessageSize)
```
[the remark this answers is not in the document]
mnesbit: Not without breaking smart cast

node-api/src/main/kotlin/net/corda/nodeapi/internal/BridgeControlMessages.kt
mnesbit: Sure

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val bridgeManager: BridgeManager = AMQPBridgeManager(config, p2pAddress, maxMessageSize)
+    private var artemis: ArtemisMessagingClient? = null
+    private var controlConsumer: ClientConsumer? = null
+
+    companion object {
+        private val log = contextLogger()
+    }
+
+    fun start() {
+        stop()
```
sollecitom: Why calling `stop()` here? I mean, to re-start it?

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
```diff
@@ -0,0 +1,86 @@
...
+        when (controlMessage) {
+            is BridgeControl.NodeToBridgeSnapshot -> {
+                for (outQueue in controlMessage.sendQueues) {
+                    bridgeManager.deployBridge(outQueue.queueName, outQueue.targets.first(), outQueue.legalNames.toSet())
+                }
+            }
+            is BridgeControl.BridgeToNodeQuery -> {
+                log.warn("Message from Bridge $controlMessage detected on wrong topic!")
+            }
+            is BridgeControl.BridgeCreate -> {
```
sollecitom: How about `BridgeControl.Create` instead? or `CreateBridge`? Similar for delete just below.

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun start() {
+        stop()
+        bridgeManager.start()
+        val artemis = ArtemisMessagingClient(config, p2pAddress, maxMessageSize)
+        this.artemis = artemis
+        artemis.start()
+        val artemisClient = artemis.started!!
+        val artemisSession = artemisClient.session
+        val bridgeControlQueue = "$BRIDGE_CONTROL.$bridgeId"
+        artemisSession.createTemporaryQueue(BRIDGE_CONTROL, RoutingType.MULTICAST, bridgeControlQueue)
```
rick-r3: I assume this is a NOOP if the queue exists already?

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -207,11 +221,95 @@ class P2PMessagingClient(config: NodeConfiguration,
...
+            enumerateBridges(session, inbox)
         }
 
         resumeMessageRedelivery()
     }
 
+    private fun InnerState.registerBridgeControl(session: ClientSession, inbox: String) {
+        val bridgeNotifyQueue = "$BRIDGE_NOTIFY.${myIdentity.toStringShort()}"
+        session.createTemporaryQueue(BRIDGE_NOTIFY, RoutingType.MULTICAST, bridgeNotifyQueue)
+        val bridgeConsumer = session.createConsumer(bridgeNotifyQueue)
```
sollecitom: Field assignment from local variable?

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
rick-r3: Should we stop and restart or throw an if it's already started?

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -207,11 +221,95 @@ class P2PMessagingClient(config: NodeConfiguration,
...
+        bridgeNotifyConsumer = bridgeConsumer
+        bridgeConsumer.setMessageHandler { msg ->
+            val data: ByteArray = ByteArray(msg.bodySize).apply { msg.bodyBuffer.readBytes(this) }
```
sollecitom: This was used above as well, perhaps an extension function on ClientMessage to produce the BridgeControl would avoid duplication.

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -207,11 +221,95 @@ class P2PMessagingClient(config: NodeConfiguration,
...
+        fun gatherAddresses(node: NodeInfo): Sequence<BridgeEntry> {
+            val address = node.addresses.first()
+            return node.legalIdentitiesAndCerts.map {
+                val messagingAddress = NodeAddress(it.party.owningKey, address)
+                BridgeEntry(messagingAddress.queueName, listOf(messagingAddress.hostAndPort), listOf(it.party.name))
+            }.filter { artemis.started!!.session.queueQuery(SimpleString(it.queueName)).isExists }.asSequence()
+        }
+
+        fun deployBridges(node: NodeInfo) {
+            gatherAddresses(node)
```
sollecitom: Interesting formatting :)

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
mnesbit: This is pub-sub, so every node will have a queue and listener and resend the bridge list if a bridge starts up. Similarly the bridge has a control queue and listens for node starts and fresh bridge requests

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
mnesbit: I think it is better to gracefully allow multiple start

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
```diff
@@ -0,0 +1,86 @@
...
+        val control = artemisSession.createConsumer(bridgeControlQueue)
+        controlConsumer = control
+        control.setMessageHandler { msg ->
+            try {
+                processControlMessage(msg)
+            } catch (ex: Exception) {
+                log.error("Unable to process bridge control message", ex)
+            }
+        }
+        val startupMessage = BridgeControl.BridgeToNodeQuery("Test").serialize(context = SerializationDefaults.P2P_CONTEXT).bytes
```
rick-r3: Should we use something other than `Test` here to avoid confusion with tests.  e.g. `Ping` or `Hello` or `Start` etc

docs/source/changelog.rst
```diff
@@ -148,6 +148,9 @@ UNRELEASED
   However, assuming a clean reset of the artemis data and that the nodes are consistent versions,
   data persisted via the AMQP serializer will be forward compatible.
 
+* Move to a message based control of peer to peer bridge formation to allow for future out of process bridging components.
+  This removes the legacy Artemis bridges completely, so the ``useAMQPBrdges`` configuration property has been removed.
```
sollecitom: Typo: ``useAMQPBridges`` instead.

node/src/main/kotlin/net/corda/node/services/messaging/BridgeControlListener.kt
rick-r3: A comment here describing what the point of the message is.

docs/source/changelog.rst
sollecitom: Removing a property does not break parsing the node configuration I believe, am I right?

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -207,11 +221,95 @@ class P2PMessagingClient(config: NodeConfiguration,
...
+        session.createTemporaryQueue(BRIDGE_NOTIFY, RoutingType.MULTICAST, bridgeNotifyQueue)
+        val bridgeConsumer = session.createConsumer(bridgeNotifyQueue)
+        bridgeNotifyConsumer = bridgeConsumer
+        bridgeConsumer.setMessageHandler { msg ->
+            val data: ByteArray = ByteArray(msg.bodySize).apply { msg.bodyBuffer.readBytes(this) }
+            val notifyMessage = data.deserialize<BridgeControl>(context = SerializationDefaults.P2P_CONTEXT)
+            log.info(notifyMessage.toString())
+            if (notifyMessage is BridgeControl.BridgeToNodeQuery) {
+                enumerateBridges(session, inbox)
+            }
```
rick-r3: Do we want an `else` warning unexpected control message?

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -207,11 +221,95 @@ class P2PMessagingClient(config: NodeConfiguration,
...
+        val controlPacket = message.serialize(context = SerializationDefaults.P2P_CONTEXT).bytes
+        val artemisMessage = client.session.createMessage(false)
+        artemisMessage.writeBodyBufferBytes(controlPacket)
+        client.producer.send(BRIDGE_CONTROL, artemisMessage)
+    }
+
+    private fun updateBridgesOnNetworkChange(change: NetworkMapCache.MapChange) {
+        log.info("Updating bridges on network map change: ${change.node}")
+        fun gatherAddresses(node: NodeInfo): Sequence<BridgeEntry> {
+            val address = node.addresses.first()
```
rick-r3: Should we send all the addresses to the bridge and let it make the decision to use the first (and later, do something more) ?

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
rick-r3: Should we log the bridgeIdentity from the message?

mnesbit:
Any chance I could get this approved?

PR #2412, opened 2018-01-23
First approach to network parameters updates

kasiastreich:
Add two RPC methods newNetworkMapParameters and
acceptNewNetworkParameters. Implementation of client handling of network
parameters update event. Partial implementation of accepting new
parameters and installing them on the node as well as node startup with
updated parameters (no flag day yet).

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
@@ -14,6 +14,7 @@ import net.corda.core.serialization.deserialize
 import net.corda.core.serialization.serialize
 import net.corda.core.transactions.TransactionBuilder
 import net.corda.core.transactions.WireTransaction
+import net.corda.core.utilities.OpaqueBytes
 import org.bouncycastle.asn1.x500.X500Name
 import org.bouncycastle.asn1.x500.X500NameBuilder
 import org.bouncycastle.asn1.x500.style.BCStyle
@@ -308,7 +309,7 @@ val KClass<*>.packageName: String get() = java.`package`.name
 
 fun URL.openHttpConnection(): HttpURLConnection = openConnection() as HttpURLConnection
 
-fun <T: Any>  URL.post(serializedData: SerializedBytes<T>) {
+fun URL.post(serializedData: OpaqueBytes) {
     openHttpConnection().apply {
         doOutput = true
         requestMethod = "POST"
--- a/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
@@ -9,6 +9,7 @@ import net.corda.core.crypto.sha256
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.node.ServicesForResolution
 import net.corda.core.serialization.SerializationContext
+import net.corda.core.serialization.SerializedBytes
 import net.corda.core.serialization.deserialize
 import net.corda.core.serialization.serialize
 import net.corda.core.transactions.TransactionBuilder
@@ -306,6 +307,16 @@ fun TransactionBuilder.toLedgerTransaction(services: ServicesForResolution, seri
 val KClass<*>.packageName: String get() = java.`package`.name
 
 fun URL.openHttpConnection(): HttpURLConnection = openConnection() as HttpURLConnection
+
+fun <T: Any>  URL.post(serializedData: SerializedBytes<T>) {
+    openHttpConnection().apply {
+        doOutput = true
+        requestMethod = "POST"
+        setRequestProperty("Content-Type", "application/octet-stream")
+        outputStream.use { serializedData.open().copyTo(it) }
+        checkOkResponse()
+    }
+}
 
 fun HttpURLConnection.checkOkResponse() {
     if (responseCode != 200) {
--- a/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
@@ -30,6 +30,7 @@ import java.nio.charset.Charset
 import java.nio.charset.StandardCharsets.UTF_8
 import java.nio.file.*
 import java.nio.file.attribute.FileAttribute
+import java.security.KeyPair
 import java.security.PrivateKey
 import java.security.cert.X509Certificate
 import java.time.Duration
@@ -364,6 +365,10 @@ fun <T : Any> T.signWithCert(privateKey: PrivateKey, certificate: X509Certificat
     return SignedDataWithCert(serialised, DigitalSignatureWithCert(certificate, signature))
 }
 
-fun <T:Any> SerializedBytes<T>.sign(signer: (SerializedBytes<T>) -> DigitalSignature.WithKey): SignedData<T> {
+inline fun <T:Any> SerializedBytes<T>.sign(signer: (SerializedBytes<T>) -> DigitalSignature.WithKey): SignedData<T> {
     return SignedData(this, signer(this))
+}
+
+inline fun <T:Any> SerializedBytes<T>.sign(keyPair: KeyPair): SignedData<T> {
+    return SignedData(this, keyPair.private.sign(this.bytes, keyPair.public))
 }
--- a/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
@@ -3,9 +3,7 @@
 package net.corda.core.internal
 
 import net.corda.core.cordapp.CordappProvider
-import net.corda.core.crypto.Crypto
-import net.corda.core.crypto.SecureHash
-import net.corda.core.crypto.sha256
+import net.corda.core.crypto.*
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.node.ServicesForResolution
 import net.corda.core.serialization.SerializationContext
@@ -364,4 +362,8 @@ fun <T : Any> T.signWithCert(privateKey: PrivateKey, certificate: X509Certificat
     val serialised = serialize()
     val signature = Crypto.doSign(privateKey, serialised.bytes)
     return SignedDataWithCert(serialised, DigitalSignatureWithCert(certificate, signature))
+}
+
+fun <T:Any> SerializedBytes<T>.sign(signer: (SerializedBytes<T>) -> DigitalSignature.WithKey): SignedData<T> {
+    return SignedData(this, signer(this))
 }
--- a/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
@@ -365,10 +365,10 @@ fun <T : Any> T.signWithCert(privateKey: PrivateKey, certificate: X509Certificat
     return SignedDataWithCert(serialised, DigitalSignatureWithCert(certificate, signature))
 }
 
-inline fun <T:Any> SerializedBytes<T>.sign(signer: (SerializedBytes<T>) -> DigitalSignature.WithKey): SignedData<T> {
+inline fun <T : Any> SerializedBytes<T>.sign(signer: (SerializedBytes<T>) -> DigitalSignature.WithKey): SignedData<T> {
     return SignedData(this, signer(this))
 }
 
-inline fun <T:Any> SerializedBytes<T>.sign(keyPair: KeyPair): SignedData<T> {
-    return SignedData(this, keyPair.private.sign(this.bytes, keyPair.public))
+inline fun <T : Any> SerializedBytes<T>.sign(keyPair: KeyPair): SignedData<T> {
+    return SignedData(this, keyPair.sign(this.bytes))
 }
--- a/core/src/main/kotlin/net/corda/core/internal/NetworkMap.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/NetworkMap.kt
@@ -1,22 +1,37 @@
-package net.corda.nodeapi.internal.network
+package net.corda.core.internal
 
 import net.corda.core.crypto.SecureHash
 import net.corda.core.identity.Party
-import net.corda.core.internal.CertRole
-import net.corda.core.internal.SignedDataWithCert
 import net.corda.core.node.NodeInfo
 import net.corda.core.serialization.CordaSerializable
-import net.corda.nodeapi.internal.crypto.X509Utilities
-import java.security.cert.X509Certificate
 import java.time.Instant
 
 const val NETWORK_PARAMS_FILE_NAME = "network-parameters"
+const val NETWORK_PARAMS_UPDATE_FILE_NAME = "network-parameters-update"
 
 /**
  * Data class containing hash of [NetworkParameters] and network participant's [NodeInfo] hashes.
+ * @property parametersUpdate if present means that network operator scheduled update of network parameters
  */
 @CordaSerializable
-data class NetworkMap(val nodeInfoHashes: List<SecureHash>, val networkParameterHash: SecureHash)
+data class NetworkMap(
+        val nodeInfoHashes: List<SecureHash>,
+        val networkParameterHash: SecureHash,
+        val parametersUpdate: ParametersUpdate?
+)
+
+/**
+ * @property newParametersHash Hash of the new [NetworkParameters], which will be requested from network map
+ * @property description Short description of the update
+ * @property flagDay initially set to null, when most of the participants of the network agree on the new parameters the network
+ * operator decides when to switch to using new parameters
+ */
+@CordaSerializable
+data class ParametersUpdate(
+        val newParametersHash: SecureHash,
+        val description: String,
+        val flagDay: Instant?
+)
 
 /**
  * @property minimumPlatformVersion Minimum version of Corda platform that is required for nodes in the network.
@@ -50,9 +65,3 @@ data class NetworkParameters(
 
 @CordaSerializable
 data class NotaryInfo(val identity: Party, val validating: Boolean)
-
-fun <T : Any> SignedDataWithCert<T>.verifiedNetworkMapCert(rootCert: X509Certificate): T {
-    require(CertRole.extract(sig.by) == CertRole.NETWORK_MAP) { "Incorrect cert role: ${CertRole.extract(sig.by)}" }
-    X509Utilities.validateCertificateChain(rootCert, sig.by, rootCert)
-    return verified()
-}
--- a/core/src/main/kotlin/net/corda/core/internal/NetworkMap.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/NetworkMap.kt
@@ -10,8 +10,10 @@ const val NETWORK_PARAMS_FILE_NAME = "network-parameters"
 const val NETWORK_PARAMS_UPDATE_FILE_NAME = "network-parameters-update"
 
 /**
- * Data class containing hash of [NetworkParameters] and network participant's [NodeInfo] hashes.
- * @property parametersUpdate if present means that network operator scheduled update of network parameters
+ * Data structure representing the network map available from the HTTP network map service as a serialised blob.
+ * @property nodeInfoHashes list of network participant's [NodeInfo] hashes
+ * @property networkParameterHash hash of the current active [NetworkParameters]
+ * @property parametersUpdate if present means that network operator has scheduled an update of the network parameters
  */
 @CordaSerializable
 data class NetworkMap(
@@ -21,16 +23,17 @@ data class NetworkMap(
 )
 
 /**
- * @property newParametersHash Hash of the new [NetworkParameters], which will be requested from network map
+ * Data class representing scheduled network parameters update.
+ * @property newParametersHash Hash of the new [NetworkParameters] which can be requested from the network map
  * @property description Short description of the update
- * @property flagDay initially set to null, when most of the participants of the network agree on the new parameters the network
- * operator decides when to switch to using new parameters
+ * @property updateDeadline deadline by which new network parameters need to be accepted, after this date network operator
+ *          can switch to new parameters which will result in getting nodes with old parameters out of the network
  */
 @CordaSerializable
 data class ParametersUpdate(
         val newParametersHash: SecureHash,
         val description: String,
-        val flagDay: Instant?
+        val updateDeadline: Instant
 )
 
 /**
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -13,7 +13,7 @@ import net.corda.core.flows.StateMachineRunId
 import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
-import net.corda.core.internal.NetworkParameters
+import net.corda.core.node.NetworkParameters
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.services.AttachmentId
 import net.corda.core.node.services.NetworkMapCache
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -76,6 +76,10 @@ sealed class StateMachineUpdate {
 /**
  * Data class containing information about the scheduled network parameters update. The info is emitted every time node
  * receives network map with [ParametersUpdate] which wasn't seen before. For more information see: [CordaRPCOps.newNetworkMapParameters] and [CordaRPCOps.acceptNewNetworkParameters].
+ * @property hash new [NetworkParameters] hash
+ * @property parameters new [NetworkParameters] data structure
+ * @property description description of the update
+ * @property updateDeadline deadline for accepting this update using [CordaRPCOps.acceptNewNetworkParameters]
  */
 @CordaSerializable
 data class ParametersUpdateInfo(
@@ -219,16 +223,15 @@ interface CordaRPCOps : RPCOps {
     fun networkMapFeed(): DataFeed<List<NodeInfo>, NetworkMapCache.MapChange>
 
     /**
-     * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none were scheduled)
-     * and observable with future update events. Any update that occurs before the deadline of the current one automatically cancels it.
-     * Only the latest update can be accepted, because node can store only one future parameters update at any given time.
+     * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none are currently scheduled)
+     * and observable with future update events. Any update that occurs before the deadline automatically cancels the current one.
+     * Only the latest update can be accepted.
      */
     @RPCReturnsObservables
     fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
 
     /**
      * Accept network parameters with given hash, hash is obtained through [newNetworkMapParameters] method.
-     * On approval, new network parameters are saved in node's base directory as `network-parameters-update` file.
      * Information is sent back to the zone operator that the node accepted the parameters update - this process cannot be
      * undone.
      * Only parameters that are scheduled for update can be accepted, if different hash is provided this method will fail.
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -76,7 +76,7 @@ sealed class StateMachineUpdate {
 
 /**
  * Data class containing information about the scheduled network parameters update. The info is emitted every time node
- * receives network map with [ParametersUpdate] which wasn't seen before. For more information see: [CordaRPCOps.newNetworkMapParameters] and [CordaRPCOps.acceptNewNetworkParameters].
+ * receives network map with [ParametersUpdate] which wasn't seen before. For more information see: [CordaRPCOps.networkParametersFeed] and [CordaRPCOps.acceptNewNetworkParameters].
  * @property hash new [NetworkParameters] hash
  * @property parameters new [NetworkParameters] data structure
  * @property description description of the update
@@ -231,14 +231,15 @@ interface CordaRPCOps : RPCOps {
      */
     // TODO This operation should be restricted to just node admins.
     @RPCReturnsObservables
-    fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
+    fun networkParametersFeed(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
 
     /**
-     * Accept network parameters with given hash, hash is obtained through [newNetworkMapParameters] method.
+     * Accept network parameters with given hash, hash is obtained through [networkParametersFeed] method.
      * Information is sent back to the zone operator that the node accepted the parameters update - this process cannot be
      * undone.
      * Only parameters that are scheduled for update can be accepted, if different hash is provided this method will fail.
      * Note: This operation may be restricted only to node administrators.
+     * @param parametersHash hash of network parameters to accept
      * @throws IllegalArgumentException if network map advertises update with different parameters hash then the one accepted by node's operator.
      * @throws IOException if failed to send the approval to network map
      */
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -73,12 +73,16 @@ sealed class StateMachineUpdate {
     data class Removed(override val id: StateMachineRunId, val result: Try<*>) : StateMachineUpdate()
 }
 
+/**
+ * Data class containing information about the scheduled network parameters update. The info is emitted every time node
+ * receives network map with [ParametersUpdate] which wasn't seen before. For more information see: [CordaRPCOps.newNetworkMapParameters] and [CordaRPCOps.acceptNewNetworkParameters].
+ */
 @CordaSerializable
 data class ParametersUpdateInfo(
         val hash: SecureHash,
         val parameters: NetworkParameters,
         val description: String,
-        val flagDay: Instant?
+        val updateDeadline: Instant
 )
 
 @CordaSerializable
@@ -216,7 +220,8 @@ interface CordaRPCOps : RPCOps {
 
     /**
      * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none were scheduled)
-     * and observable with future update events.
+     * and observable with future update events. Any update that occurs before the deadline of the current one automatically cancels it.
+     * Only the latest update can be accepted, because node can store only one future parameters update at any given time.
      */
     @RPCReturnsObservables
     fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
@@ -225,10 +230,9 @@ interface CordaRPCOps : RPCOps {
      * Accept network parameters with given hash, hash is obtained through [newNetworkMapParameters] method.
      * On approval, new network parameters are saved in node's base directory as `network-parameters-update` file.
      * Information is sent back to the zone operator that the node accepted the parameters update - this process cannot be
-     * revoked.
+     * undone.
      * Only parameters that are scheduled for update can be accepted, if different hash is provided this method will fail.
-     * @throws IllegalArgumentException if network map advertises update with different parameters hash (node's operator needs to accept the newest version)
-     * @throws IllegalStateException if compatibility zone wasn't configured
+     * @throws IllegalArgumentException if network map advertises update with different parameters hash then the one accepted by node's operator
      */
     fun acceptNewNetworkParameters(parametersHash: SecureHash)
 
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -74,6 +74,7 @@ sealed class StateMachineUpdate {
     data class Removed(override val id: StateMachineRunId, val result: Try<*>) : StateMachineUpdate()
 }
 
+// DOCSTART 1
 /**
  * Data class containing information about the scheduled network parameters update. The info is emitted every time node
  * receives network map with [ParametersUpdate] which wasn't seen before. For more information see: [CordaRPCOps.networkParametersFeed] and [CordaRPCOps.acceptNewNetworkParameters].
@@ -89,6 +90,7 @@ data class ParametersUpdateInfo(
         val description: String,
         val updateDeadline: Instant
 )
+// DOCEND 1
 
 @CordaSerializable
 data class StateMachineTransactionMapping(val stateMachineRunId: StateMachineRunId, val transactionId: SecureHash)
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -13,6 +13,7 @@ import net.corda.core.flows.StateMachineRunId
 import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
+import net.corda.core.internal.NetworkParameters
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.services.AttachmentId
 import net.corda.core.node.services.NetworkMapCache
@@ -71,6 +72,14 @@ sealed class StateMachineUpdate {
 
     data class Removed(override val id: StateMachineRunId, val result: Try<*>) : StateMachineUpdate()
 }
+
+@CordaSerializable
+data class ParametersUpdateInfo(
+        val hash: SecureHash,
+        val parameters: NetworkParameters,
+        val description: String,
+        val flagDay: Instant?
+)
 
 @CordaSerializable
 data class StateMachineTransactionMapping(val stateMachineRunId: StateMachineRunId, val transactionId: SecureHash)
@@ -204,6 +213,23 @@ interface CordaRPCOps : RPCOps {
      */
     @RPCReturnsObservables
     fun networkMapFeed(): DataFeed<List<NodeInfo>, NetworkMapCache.MapChange>
+
+    /**
+     * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none were scheduled)
+     * and observable with future update events.
+     */
+    @RPCReturnsObservables
+    fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
+
+    /**
+     * Accept network parameters with given hash, usually hash is obtained through [newNetworkMapParameters] method.
+     * On approval, new network parameters are saved in node's base directory as `network-parameters-update` file.
+     * Information is sent back to the zone operator that the node accepted the parameters update - this process cannot be
+     * revoked.
+     * Only parameters that are scheduled for update can be accepted, if different hash is provided this method will fail.
+     */
+    @Throws(IllegalArgumentException::class)
+    fun acceptNewNetworkParameters(parametersHash: SecureHash)
 
     /**
      * Start the given flow with the given arguments. [logicType] must be annotated
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -24,6 +24,7 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.utilities.Try
 import rx.Observable
+import java.io.IOException
 import java.io.InputStream
 import java.security.PublicKey
 import java.time.Instant
@@ -226,7 +227,9 @@ interface CordaRPCOps : RPCOps {
      * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none are currently scheduled)
      * and observable with future update events. Any update that occurs before the deadline automatically cancels the current one.
      * Only the latest update can be accepted.
+     * Note: This operation may be restricted only to node administrators.
      */
+    // TODO This operation should be restricted to just node admins.
     @RPCReturnsObservables
     fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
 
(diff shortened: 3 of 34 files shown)


core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -209,6 +218,23 @@ interface CordaRPCOps : RPCOps {
...
 
+    /**
+     * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none were scheduled)
+     * and observable with future update events.
+     */
+    @RPCReturnsObservables
+    fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
+
+    /**
+     * Accept network parameters with given hash, usually hash is obtained through [newNetworkMapParameters] method.
```
shamsasari: Why "usually"?

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -209,6 +218,23 @@ interface CordaRPCOps : RPCOps {
...
+     * On approval, new network parameters are saved in node's base directory as `network-parameters-update` file.
+     * Information is sent back to the zone operator that the node accepted the parameters update - this process cannot be
+     * revoked.
+     * Only parameters that are scheduled for update can be accepted, if different hash is provided this method will fail.
+     */
+    @Throws(IllegalArgumentException::class)
```
shamsasari: Add `@throws` section in KDoc, rather than this.

(This annotation doesn't make sense for runtime exceptions)

docs/source/network-map.rst
```diff
@@ -22,6 +22,8 @@ The set of REST end-points for the network map service are as follows.
 +================+=========================================+==============================================================================================================================================+
 | POST           | /network-map/publish                    | For the node to upload its signed ``NodeInfo`` object to the network map.                                                                    |
 +----------------+-----------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------+
+| POST           | /network-map/ack-parameters             | For the node operator to acknowledge network map that new parameters were accepted for future update.                                        |
```
shamsasari: Perhaps we should rethink this. If we extend the `/network-map/publish` endpoint to expect the node info *and* the network parameters that it's on then all the node needs to do to accept the new parameters is to just re-submit its node info. 

Probably that means extending `SignedNodeInfo` to include the parameters hash. This has an added benefit for nodes using the additional-node-infos directory. They can use the parameters hash to ignore any node info files which are not on the same network parameters. Currently locally deployed nodes just assume they're all on the same set of parameters.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -167,7 +167,7 @@ class NetworkBootstrapper {
                 epoch = 1
         ), overwriteFile = true)
 
-        nodeDirs.forEach(copier::install)
+        nodeDirs.forEach { copier.install(it) }
```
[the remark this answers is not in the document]
kasiastreich: It's because I added default parameter to indicate if it's update or regular network parameters file

docs/source/network-map.rst
kasiastreich: The idea is to have information about accepted network parameters sent on `NodeInfo` republishing. I planned to add that in next PR (not a big problem though to include in this PR, but I wanted to avoid it being too big).

As to second part of your comment, I see an issue here. We will have period of two set of parameters being accepted by nodes (update period before the `flagDay`) we cannot simply reject them based on that information. Also let's remember that what is in `additional-node-infos` directory isn't always the most up-to-date view of the network.

shamsasari: >The idea is to have information about accepted network parameters sent on NodeInfo republishing. I planned to add that in next PR (not a big problem though to include in this PR, but I wanted to avoid it being too big).
But that means introducing this endpoint and then getting rid of it in the next PR.

kasiastreich: True, I realized that we don't need it yesterday and started adding republishing implementation. It has to go to `NodeInfo` itself, not `SignedNodeInfo`.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -634,23 +632,58 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
-        } else {
-            log.info("No network-parameters file found. Expecting network parameters to be available from the network map.")
-            val networkMapClient = checkNotNull(networkMapClient) {
-                "Node hasn't been configured to connect to a network map from which to get the network parameters"
+        } else null
+        networkParameters = if (advertisedParametersHash != null) {
+            // I feel it doesn't make sense, on one hand we have node starting without parameters and just accepting them by default,
+            //  on the other we have parameters update process - it needs to be unified. Say you start the node, you don't have matching parameters,
+            //  you get them from network map, but you have to run the acceptance step. TODO To be implemented in later stage?
+            if (parametersFromFile == null) downloadParameters(trustRoot, advertisedParametersHash) // Node joins for the first time.
```
shamsasari: braces

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -15,38 +16,42 @@ import net.corda.core.utilities.trace
...
 
     fun publish(signedNodeInfo: SignedNodeInfo) {
         val publishURL = URL("$networkMapUrl/publish")
         logger.trace { "Publishing NodeInfo to $publishURL." }
-        publishURL.openHttpConnection().apply {
+        doPost(publishURL, signedNodeInfo.serialize())
+        logger.trace { "Published NodeInfo to $publishURL successfully." }
+    }
+
+    private fun <T: Any> doPost(url: URL, serializedData: SerializedBytes<T>) {
```
shamsasari: rename this to `post` and make it an extension function on URL

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -180,4 +200,34 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
         }
         executor.submit(task)
     }
+
+    private fun handleUpdateNetworkParameters(update: ParametersUpdate) {
+        if (update.newParametersHash != newNetworkParameters?.first?.newParametersHash) {
```
shamsasari: invert the if-condition and return

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -209,6 +218,24 @@ interface CordaRPCOps : RPCOps {
...
+    fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
+
+    /**
+     * Accept network parameters with given hash, hash is obtained through [newNetworkMapParameters] method.
+     * On approval, new network parameters are saved in node's base directory as `network-parameters-update` file.
+     * Information is sent back to the zone operator that the node accepted the parameters update - this process cannot be
+     * revoked.
+     * Only parameters that are scheduled for update can be accepted, if different hash is provided this method will fail.
+     * @throws IllegalArgumentException if network map advertises update with different parameters hash (node's operator needs to accept the newest version)
+     * @throws IllegalStateException if compatibility zone wasn't configured
```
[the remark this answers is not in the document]
kasiastreich: Node operator can accept the old update when we started to advertised the new one.

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
```diff
@@ -372,3 +372,9 @@ enum class CertificateType(val keyUsage: KeyUsage, vararg val purposes: KeyPurpo
 }
 
 data class CertificateAndKeyPair(val certificate: X509Certificate, val keyPair: KeyPair)
+
+fun <T : Any> SignedDataWithCert<T>.verifiedNetworkMapCert(rootCert: X509Certificate): T {
```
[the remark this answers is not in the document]
kasiastreich: Don't see better place for that, can't move to `NetworkMap` because it uses ` X509Utilities.validateCertificateChain(rootCert, sig.by, rootCert)`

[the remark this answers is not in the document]
shamsasari: Why was NetworkMap moved to core?

core/src/main/kotlin/net/corda/core/internal/NetworkMap.kt
```diff
@@ -1,22 +1,40 @@
-package net.corda.nodeapi.internal.network
+package net.corda.core.internal
```
shamsasari: As discussed, since this is needed by RPC it's now part of the public API and as such doesn't belong in internal package.

shamsasari: But if it's only `NetworkParameters` that's being exposed and not the other classes, then move the other ones back. Only make public what we need.

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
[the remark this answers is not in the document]
shamsasari: X509Utilities is a general-purpose X509 util class, while this method is specific to network map code. Perhaps but it in its own class.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -307,6 +308,16 @@ val KClass<*>.packageName: String get() = java.`package`.name
 
 fun URL.openHttpConnection(): HttpURLConnection = openConnection() as HttpURLConnection
 
+fun <T: Any>  URL.post(serializedData: SerializedBytes<T>) {
```
shamsasari: Double space

shamsasari: The type parameter is not needed

shamsasari: In fact you could probably make it `OpaqueBytes`

core/src/main/kotlin/net/corda/core/node/NodeInfo.kt
```diff
@@ -17,12 +18,16 @@ import net.corda.core.utilities.NetworkHostAndPort
  *           for information on how the platform is versioned.
  * @property serial An arbitrary number incremented each time the NodeInfo is changed. This is analogous to the same
  *           concept in DNS.
+ * @property acceptedParametersHash Parameters hash accepted by this node, note that it doesn't necessarily point to the
```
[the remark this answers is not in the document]
kasiastreich: This design also has the same problem as previous one. I think the cleanest way is to have separate endpoint for accepting network parameters, where node sends signed hash of parameters. Other approaches will results either in nullable field in data structure (because of how we set up networks without network map with `--just-generate-node-info` flag) or proliferation of types - we need to sign both `NodeInfo` and accepted parameters. I will revert back to using the `ack-parameters` endpoint.

node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapTest.kt
```diff
@@ -73,6 +77,26 @@ class NetworkMapTest {
...
+                startNodesInProcess = true,
+                portAllocation = portAllocation,
+                compatibilityZone = compatibilityZone,
+                initialiseSerialization = false,
+                notarySpecs = emptyList()
+        ) {
+            val aliceDir = baseDirectory(ALICE_NAME).createDirectories()
+            val oldParameters = testNetworkParameters(emptyList(), epoch = 1)
+            NetworkParametersCopier(oldParameters).install(aliceDir)
+            NetworkParametersCopier(networkMapServer.networkParameters, update = true).install(aliceDir)
```
shamsasari: We shouldn't be writing the update file here. Rather we should be testing the whole process using the RPC interface for network parameters approval, etc.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -634,23 +630,54 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+            // I feel it doesn't make sense, on one hand we have node starting without parameters and just accepting them by default,
+            //  on the other we have parameters update process - it needs to be unified. Say you start the node, you don't have matching parameters,
+            //  you get them from network map, but you have to run the acceptance step. TODO To be implemented in later stage?
+            if (parametersFromFile == null) { // Node joins for the first time.
+                downloadParameters(trustRoot, advertisedParametersHash)
+            }
+            else if (parametersFromFile.serialize().hash == advertisedParametersHash) { // Restarted with the same parameters.
+                parametersFromFile
+            } else { // Update case.
+                val parametersUpdateFile = configuration.baseDirectory / NETWORK_PARAMS_UPDATE_FILE_NAME
```
shamsasari: Extract out the else case into a method

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -634,23 +630,54 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+            null
+        }
+        networkParameters = if (advertisedParametersHash != null) {
+            // I feel it doesn't make sense, on one hand we have node starting without parameters and just accepting them by default,
+            //  on the other we have parameters update process - it needs to be unified. Say you start the node, you don't have matching parameters,
+            //  you get them from network map, but you have to run the acceptance step. TODO To be implemented in later stage?
+            if (parametersFromFile == null) { // Node joins for the first time.
+                downloadParameters(trustRoot, advertisedParametersHash)
+            }
+            else if (parametersFromFile.serialize().hash == advertisedParametersHash) { // Restarted with the same parameters.
```
shamsasari: The `serialize` call here is slightly wasteful as in this case we've already read in `SignedDataWithCert<NetworkParameters>` which contains the serialised bytes. If it's possible instead of `networkParameters`, have `signedNetParams` and do the verification outside the if-else block.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -634,23 +630,54 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
...
+                parametersFromFile
+            } else { // Update case.
+                val parametersUpdateFile = configuration.baseDirectory / NETWORK_PARAMS_UPDATE_FILE_NAME
+                if (!parametersUpdateFile.exists()) {
+                    throw IllegalArgumentException("Node uses parameters with hash: ${parametersFromFile.serialize().hash} " +
+                            "but network map is advertising: ${advertisedParametersHash}.\n" +
+                            "Please update node to use correct network parameters file.")
+                }
+                val parameterUpdate = parametersUpdateFile.readAll().deserialize<SignedDataWithCert<NetworkParameters>>().verifiedNetworkMapCert(trustRoot)
+                if (parameterUpdate.serialize().hash != advertisedParametersHash) {
```
shamsasari: Similar point here. You've just read in the update file which has the serialised parameters.

node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
```diff
@@ -48,6 +50,18 @@ internal class CordaRPCOpsImpl(
...
 
+    override fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo> {
+        return services.networkMapUpdater.track()
+    }
+
+    override fun acceptNewNetworkParameters(parametersHash: SecureHash) {
+        services.networkMapUpdater.acceptNewNetworkParameters(
+                parametersHash,
+                // TODO When multiple identities design will be better specified this should be signature from node operator.
+                { hash -> SignedData(hash.serialize(), services.keyManagementService.sign(hash.serialize().bytes, services.myInfo.legalIdentities[0].owningKey)) }
```
shamsasari: Can you double-check with @kchalkias if we need a nonce here.

shamsasari: You're serialising `hash` twice

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -181,4 +200,48 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
         }
         executor.submit(task)
     }
+
+    private fun handleUpdateNetworkParameters(update: ParametersUpdate) {
+        if (update.newParametersHash == newNetworkParameters?.first?.newParametersHash) {
+            logger.warn("Couldn't retrive network parameters from update: $update")
```
shamsasari: This log message doesn't seem right. The retrieving is done below, right?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -181,4 +200,48 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+        val (update, newParams) = newNetworkParameters ?: throw IllegalArgumentException("Couldn't find parameters update for the hash: $parametersHash")
+        val newParametersHash = update.newParametersHash
+        if (parametersHash == newParametersHash) {
+            // The latest parameters have priority.
+            newParams.serialize()
+                    .open()
+                    .copyTo(baseDirectory / NETWORK_PARAMS_UPDATE_FILE_NAME, StandardCopyOption.REPLACE_EXISTING)
+            tryAckNetworkParametersUpdateAsync(sign(parametersHash), networkMapClient!!)
+        } else {
+            throw IllegalArgumentException("Refused to accept parameters with hash: $parametersHash because network map advertises update with hash: $newParametersHash, please check newest version")
```
shamsasari: remove colon after hash, in both instances

shamsasari: Full-stop instead of comma

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -181,4 +200,48 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+        }
+    }
+
+    private fun tryAckNetworkParametersUpdateAsync(signedParametersHash: SignedData<SecureHash>, networkMapClient: NetworkMapClient) {
```
shamsasari: Let's not have this async. Having it synchronous means any issues with the update can be passed to the RPC user, and they'll know the update wasn't successful.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -180,18 +197,72 @@ class NetworkMapUpdaterTest {
...
+        val updates = paramsFeed.updates.bufferUntilSubscribed()
+        assertEquals(null, snapshot)
+        val newParameters = testNetworkParameters(emptyList(), epoch = 2)
+        val updateDeadline = Instant.now().plus(1, ChronoUnit.DAYS)
+        scheduleParametersUpdate(newParameters, "Test update", updateDeadline)
+        updater.subscribeToNetworkMap()
+        updates.expectEvents(isStrict = false) {
+            sequence(
+                    expect { update: ParametersUpdateInfo ->
+                        require(update.updateDeadline == updateDeadline)
```
shamsasari: use asserts

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -180,18 +197,72 @@ class NetworkMapUpdaterTest {
...
+                { hash -> SignedData(hash.serialize(), keyPair.private.sign(hash.serialize().bytes, keyPair.public)) })
+        verify(networkMapClient).ackNetworkParametersUpdate(any())
+        val updateFile = baseDir / NETWORK_PARAMS_UPDATE_FILE_NAME
+        assertTrue(updateFile.exists())
+        val signedNetworkParams = updateFile.readAll().deserialize<SignedDataWithCert<NetworkParameters>>()
+        val paramsFromFile = signedNetworkParams.verifiedNetworkMapCert(DEV_ROOT_CA.certificate)
+        assertEquals(newParameters, paramsFromFile)
+    }
+
+    private fun scheduleParametersUpdate(nextParameters: NetworkParameters, description: String, updateDeadline: Instant = Instant.now()) {
```
shamsasari: Don't default to `now`. This can create flaky tests on systems which don't have fine resolution on current time.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -180,18 +197,72 @@ class NetworkMapUpdaterTest {
...
+        networkParamsMap[nextParameters.serialize().hash] = nextParameters
```
shamsasari: you're serialising twice

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -180,18 +197,72 @@ class NetworkMapUpdaterTest {
...
+    fun `ack network parameters update`() {
+        val newParameters = testNetworkParameters(emptyList(), epoch = 314)
+        scheduleParametersUpdate(newParameters, "Test update")
+        updater.subscribeToNetworkMap()
+        // TODO: Remove sleep in unit test.
+        Thread.sleep(2L * cacheExpiryMs)
+        val newHash = newParameters.serialize().hash
+        val keyPair = Crypto.generateKeyPair()
+        updater.acceptNewNetworkParameters(newHash,
+                { hash -> SignedData(hash.serialize(), keyPair.private.sign(hash.serialize().bytes, keyPair.public)) })
```
shamsasari: Perhaps we need an extension method in InternalUtils for the boilerplate of creating `SignedData`

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -180,18 +197,72 @@ class NetworkMapUpdaterTest {
...
+        verify(networkMapClient).ackNetworkParametersUpdate(any())
+        val updateFile = baseDir / NETWORK_PARAMS_UPDATE_FILE_NAME
+        assertTrue(updateFile.exists())
```
shamsasari: Not needed

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -180,18 +197,72 @@ class NetworkMapUpdaterTest {
...
             }
             on { getNodeInfo(any()) }.then {
                 nodeInfoMap[it.arguments[0]]?.verified()
             }
+            on { getNetworkParameters(any()) }.then {
+                val paramsHash: SecureHash = uncheckedCast(it.arguments[0])
+                networkParamsMap[paramsHash]?.signWithCert(networkMapCa.keyPair.private, networkMapCa.certificate)
+            }
+            on { ackNetworkParametersUpdate(any()) }.then {
+                val signedHash: SignedData<SecureHash> = uncheckedCast(it.arguments[0])
```
shamsasari: This is not doing anything

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
```diff
@@ -80,13 +85,26 @@ class NetworkMapServer(private val cacheTimeout: Duration,
         service.removeNodeInfo(nodeInfo)
     }
 
+    fun scheduleParametersUpdate(nextParameters: NetworkParameters, description: String, updateDeadline: Instant) {
+        nextNetworkParameters = nextParameters
+        parametersUpdate = ParametersUpdate(nextParameters.serialize().hash, description, updateDeadline)
+    }
+
+    fun advertiseNewParameters() {
+        checkNotNull(nextNetworkParameters) { "Schedule parameters update first" }
```
shamsasari: `networkParameters = checkNotNull(...`

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
```diff
@@ -108,10 +126,27 @@ class NetworkMapServer(private val cacheTimeout: Duration,
...
+        @POST
+        @Path("ack-parameters")
+        @Consumes(MediaType.APPLICATION_OCTET_STREAM)
+        fun ackNetworkParameters(input: InputStream): Response {
+            return try {
+                val signedParametersHash = input.readBytes().deserialize<SignedData<SecureHash>>()
+                val hash = signedParametersHash.verified()
+                latestAcceptedParametersMap[signedParametersHash.sig.by] = hash
+                ok()
+            } catch (e: Exception) {
```
shamsasari: This is a test network map and probably doesn't need correct error codes. Leaving the exception to propagate is probably sufficient.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
```diff
@@ -137,8 +172,14 @@ class NetworkMapServer(private val cacheTimeout: Duration,
         @Path("network-parameters/{var}")
         @Produces(MediaType.APPLICATION_OCTET_STREAM)
         fun getNetworkParameter(@PathParam("var") hash: String): Response {
-            require(signedNetParams.raw.hash == SecureHash.parse(hash))
-            return Response.ok(signedNetParams.serialize().bytes).build()
+            val requestedHash = SecureHash.parse(hash)
```
shamsasari: Why not simply store the valid network parameters in a map? It would make this method a one liner.

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -209,6 +222,24 @@ interface CordaRPCOps : RPCOps {
     @RPCReturnsObservables
     fun networkMapFeed(): DataFeed<List<NodeInfo>, NetworkMapCache.MapChange>
 
+    /**
+     * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none were scheduled)
+     * and observable with future update events. Any update that occurs before the deadline of the current one automatically cancels it.
+     * Only the latest update can be accepted, because node can store only one future parameters update at any given time.
+     */
+    @RPCReturnsObservables
+    fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
```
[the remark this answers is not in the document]
kasiastreich: Michele says that we don't have the concept of node admins yet.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -0,0 +1,38 @@
+package net.corda.core.node
```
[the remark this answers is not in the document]
kasiastreich: But it can't, it has to be accessible from RPC.

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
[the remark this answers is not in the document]
shamsasari: Add a TODO and update the documentation to say this operation may be restricted.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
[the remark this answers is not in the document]
shamsasari: Because `CordaRPCOps` is in core, right.

node/src/integration-test/kotlin/net/corda/node/services/network/NetworkMapTest.kt
kasiastreich: I will add separate test with RPC.

shamsasari: But then this test as is doesn't justify starting up a full node using the driver. Our integration tests are already slow, we need to be careful what we add to it. I believe you're already testing the updater with regards to the update file so there shouldn't be any need to expose the update file like this here.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
kasiastreich: On accepting network parameters, the file should be saved.

shamsasari: Right, but the reading the next line will throw an exception if it's not there.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
kasiastreich: Didn't know this one ;)

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
kasiastreich: How much does it matter if we send back information about accepting parameters? Network map server can be down, but we should still be able to install updates locally and proceed.

shamsasari: The network map server shouldn't be down (especially during this stage) and highlighting it to RPC user is necessary. We don't want them thinking the parameters have been accepted because the RPC returned successfully when it hasn't really.

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
kasiastreich: I am not doing any testing against that field, so it doesn't matter here. But I will remove default value.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
kasiastreich: It's because there is also `getNetworkMap` method that returns hash of current network parameters. When you store them in map it's hard to tell which are which without additional code, so there is not that much gain.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -365,3 +363,7 @@ fun <T : Any> T.signWithCert(privateKey: PrivateKey, certificate: X509Certificat
     val signature = Crypto.doSign(privateKey, serialised.bytes)
     return SignedDataWithCert(serialised, DigitalSignatureWithCert(certificate, signature))
 }
+
+fun <T:Any> SerializedBytes<T>.sign(signer: (SerializedBytes<T>) -> DigitalSignature.WithKey): SignedData<T> {
```
shamsasari: `inline`

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -224,24 +223,9 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
             newParams.serialize()
                     .open()
                     .copyTo(baseDirectory / NETWORK_PARAMS_UPDATE_FILE_NAME, StandardCopyOption.REPLACE_EXISTING)
-            tryAckNetworkParametersUpdateAsync(sign(parametersHash), networkMapClient!!)
+            networkMapClient!!.ackNetworkParametersUpdate(sign(parametersHash))
```
shamsasari: Can you change the `checkNotNull` above to `networkMapClient ?: throw IllegalStateException(...` to avoid this `!!`

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -222,23 +220,21 @@ class NetworkMapUpdaterTest {
...
-        scheduleParametersUpdate(newParameters, "Test update")
+        scheduleParametersUpdate(newParameters, "Test update", Instant.MIN)
         updater.subscribeToNetworkMap()
         // TODO: Remove sleep in unit test.
         Thread.sleep(2L * cacheExpiryMs)
         val newHash = newParameters.serialize().hash
         val keyPair = Crypto.generateKeyPair()
-        updater.acceptNewNetworkParameters(newHash,
-                { hash -> SignedData(hash.serialize(), keyPair.private.sign(hash.serialize().bytes, keyPair.public)) })
+        updater.acceptNewNetworkParameters(newHash, { hash -> hash.serialize().sign { keyPair.private.sign(it.bytes, keyPair.public) } })
```
shamsasari: Perhaps also have an overload of the extension method that takes in a `KeyPair` instead of the `signer` function

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -209,6 +227,28 @@ interface CordaRPCOps : RPCOps {
...
 
+    /**
+     * Returns [DataFeed] object containing information on currently scheduled parameters update (null if none are currently scheduled)
+     * and observable with future update events. Any update that occurs before the deadline automatically cancels the current one.
+     * Only the latest update can be accepted.
+     * Note: This operation may be restricted only to node administrators.
+     */
+    // TODO This operation should be restricted to just node admins.
+    @RPCReturnsObservables
+    fun newNetworkMapParameters(): DataFeed<ParametersUpdateInfo?, ParametersUpdateInfo>
```
shamsasari: `newNetworkParameters`

shamsasari: Or `networkParametersFeed` to be consistent with the other methods

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -209,6 +227,28 @@ interface CordaRPCOps : RPCOps {
...
+     * Accept network parameters with given hash, hash is obtained through [newNetworkMapParameters] method.
+     * Information is sent back to the zone operator that the node accepted the parameters update - this process cannot be
+     * undone.
+     * Only parameters that are scheduled for update can be accepted, if different hash is provided this method will fail.
+     * Note: This operation may be restricted only to node administrators.
+     * @throws IllegalArgumentException if network map advertises update with different parameters hash then the one accepted by node's operator.
+     * @throws IOException if failed to send the approval to network map
+     */
+    // TODO This operation should be restricted to just node admins.
+    fun acceptNewNetworkParameters(parametersHash: SecureHash)
```
shamsasari: Missing `@parameter` kdoc

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.time.Instant
+
+/**
+ * Network parameters are a set of values that every node participating in the zone needs to agree on and use to
+ * correctly interoperate with each other.
+ * @property minimumPlatformVersion Minimum version of Corda platform that is required for nodes in the network.
+ * @property notaries List of well known and trusted notary identities with information on validation type.
+ * @property maxMessageSize Maximum P2P message sent over the wire in bytes.
+ * @property maxTransactionSize Maximum permitted transaction size in bytes.
+ * @property modifiedTime
```
shamsasari: Fill this in

node-api/src/main/kotlin/net/corda/nodeapi/internal/SignedNodeInfo.kt
```diff
@@ -44,3 +46,11 @@ class SignedNodeInfo(val raw: SerializedBytes<NodeInfo>, val signatures: List<Di
         return nodeInfo
     }
 }
+
+inline fun NodeInfo.signNodeInfo(signer: (PublicKey, SerializedBytes<NodeInfo>) -> DigitalSignature): SignedNodeInfo {
```
shamsasari: `sign` is probably sufficient

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    private fun retrieveNetworkParameters(): NetworkParameters {
+        val advertisedParametersHash = networkMapClient?.getNetworkMap()?.networkMap?.networkParameterHash
+        val signedParametersFromFile = if (networkParamsFile.exists()) {
+            networkParamsFile.readAll().deserialize<SignedDataWithCert<NetworkParameters>>()
+        } else {
+            null
+        }
+        val parameters = if (advertisedParametersHash != null) {
+            // I feel it doesn't make sense, on one hand we have node starting without parameters and just accepting them by default,
```
shamsasari: Can you reword this into a TODO

node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
kchalkias: @shamsasari and @kasiastreich as we don't want to obfuscate something when signing, no nonces are required. Even if we had one, we should send it along with the signature, so the NMapServer can verify the clearText (hash+ nonce).
Also, we don't need a nonce to `NetworkParameters` because its hash will always be different on each update due to `epoch` change (and probably `modifiedTime`). We should ensure though that `epoch` is ALWAYS updated as going to a previous state accidentally (or not change state at all) might be prone to replay attacks (eg send an old signature). I guess we guarantee this already.

kchalkias: Note that we should never blindly sign hashes though, we should always compute the hash of data by ourselves. Because one can send a malicious data for signing. Do we take this into account?

node/src/test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -90,4 +107,39 @@ class NetworkMapClientTest {
...
+        val (networkMap) = networkMapClient.getNetworkMap()
+        assertEquals(networkMap.networkParameterHash, hash1)
+        assertEquals(networkMap.parametersUpdate?.description, description)
+        assertEquals(networkMap.parametersUpdate?.newParametersHash, hash2)
+        val params1 = networkMapClient.getNetworkParameters(hash1).verified()
+        val params2 = networkMapClient.getNetworkParameters(hash2).verified()
+        assertEquals(params1, server.networkParameters)
+        assertEquals(params2, nextParameters)
+        val keyPair = Crypto.generateKeyPair()
+        val signedHash = SignedData(hash2.serialize(), keyPair.private.sign(hash2.serialize().bytes, keyPair.public))
```
shamsasari: Use your extension method

node/src/test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -90,4 +107,39 @@ class NetworkMapClientTest {
     fun `get hostname string from http response correctly`() {
         assertEquals("test.host.name", networkMapClient.myPublicHostname())
     }
+
+    @Test
+    fun `handle parameters update`() {
+        val nextParameters = testNetworkParameters(emptyList(), epoch = 2)
+        val hash1 = server.networkParameters.serialize().hash
```
shamsasari: `originalNetworkParametersHash`

node/src/test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -90,4 +107,39 @@ class NetworkMapClientTest {
...
+        val hash2 = nextParameters.serialize().hash
```
shamsasari: `nextNetworkParametersHash`

node/src/test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -90,4 +107,39 @@ class NetworkMapClientTest {
...
+        val description = "Test parameters"
+        server.scheduleParametersUpdate(nextParameters, description, Instant.now().plus(1, ChronoUnit.DAYS))
+        val (networkMap) = networkMapClient.getNetworkMap()
+        assertEquals(networkMap.networkParameterHash, hash1)
+        assertEquals(networkMap.parametersUpdate?.description, description)
+        assertEquals(networkMap.parametersUpdate?.newParametersHash, hash2)
+        val params1 = networkMapClient.getNetworkParameters(hash1).verified()
```
shamsasari: perhaps inline these two params* with their asserts. Will make this test a little less busy

node/src/test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -90,4 +107,39 @@ class NetworkMapClientTest {
...
+        val params2 = networkMapClient.getNetworkParameters(hash2).verified()
+        assertEquals(params1, server.networkParameters)
+        assertEquals(params2, nextParameters)
+        val keyPair = Crypto.generateKeyPair()
+        val signedHash = SignedData(hash2.serialize(), keyPair.private.sign(hash2.serialize().bytes, keyPair.public))
+        networkMapClient.ackNetworkParametersUpdate(signedHash)
+    }
```
shamsasari: Perhaps verify at the end that the network map server received the ack

node/src/test/kotlin/net/corda/node/services/network/NetworkMapClientTest.kt
```diff
@@ -90,4 +107,39 @@ class NetworkMapClientTest {
...
+
+    @Test
+    fun `read correct set of parameters from file`() {
```
shamsasari: It looks like this test belongs in `NetworkParametersReaderTest`

node/src/test/kotlin/net/corda/node/services/network/NetworkMapUpdaterTest.kt
```diff
@@ -180,18 +194,73 @@ class NetworkMapUpdaterTest {
...
+        val updates = paramsFeed.updates.bufferUntilSubscribed()
+        assertEquals(null, snapshot)
+        val newParameters = testNetworkParameters(emptyList(), epoch = 2)
+        val updateDeadline = Instant.now().plus(1, ChronoUnit.DAYS)
+        scheduleParametersUpdate(newParameters, "Test update", updateDeadline)
+        updater.subscribeToNetworkMap()
+        updates.expectEvents(isStrict = false) {
+            sequence(
+                    expect { update: ParametersUpdateInfo ->
+                        assert(update.updateDeadline == updateDeadline)
```
shamsasari: Wrong assert. I meant `assertThat` or `assertEquals`!

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/network/NetworkMapServer.kt
```diff
@@ -80,13 +85,25 @@ class NetworkMapServer(private val cacheTimeout: Duration,
         service.removeNodeInfo(nodeInfo)
     }
 
+    fun scheduleParametersUpdate(nextParameters: NetworkParameters, description: String, updateDeadline: Instant) {
+        nextNetworkParameters = nextParameters
+        parametersUpdate = ParametersUpdate(nextParameters.serialize().hash, description, updateDeadline)
+    }
+
+    fun advertiseNewParameters() {
+        networkParameters  = checkNotNull(nextNetworkParameters) { "Schedule parameters update first" }
```
shamsasari: extra space

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+requires human interaction and approval of the change. The rest of the process is automatic.
+
+When the update is about to happen Network Map Service starts to advertise the additional information with the usual network map
+data. It includes new network parameters hash, description of the change and the update deadline. Node queries network map server
+for the new set of parameters and emits ``ParametersUpdateInfo`` via ``CordaRPCOps::networkParametersFeed`` method to inform
+node operator about the event.
+
+.. container:: codeset
+
+    .. sourcecode:: kotlin
```
shamsasari: Link to the code instead.

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+         */
+        @CordaSerializable
+        data class ParametersUpdateInfo(
+                val hash: SecureHash,
+                val parameters: NetworkParameters,
+                val description: String,
+                val updateDeadline: Instant
+        )
+
+Node administrator can review the change and decide if is going to accept it. The approval should be done before ``updateDeadline``.
```
shamsasari: Add that any node that doesn't approve before the deadline will be removed from the network map.

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+Any new update that occurs before the deadline automatically cancels the current one. Only the latest update can be accepted.
```
shamsasari: "If the network operator starts advertising a different set of new parameters then that new set overrides the previous new set ..."

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+To send back parameters approval to the zone operator and install them on the node it is sufficient to call RPC method
```
shamsasari: Remove "and install them on the node" as it implies the node is using them

shamsasari: Perhaps remove "sufficient"

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+``fun acceptNewNetworkParameters(parametersHash: SecureHash)`` with ``parametersHash`` from update, notice that the process cannot be undone.
```
shamsasari: "from the update. Notice that the process cannot be undone."

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+
+After certain percentage of nodes in the compatibility zone will accept the new parameters network map will start to advertise
```
shamsasari: As discussed this is not correct. This should say that the next time the node polls the network map after the deadline the advertised network parameters will be the updated ones, and that the previous network parameters are no longer valid. Also say that this point the node will automatically shutdown and it required by the node operator to bring it back up again.

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+
+Network parameters update process
+---------------------------------
+
+In case of the need to change network parameters Corda zone operator will start the update process. There are many reasons
+that may lead to this decision: we discovered that some new fields have to be added to enable smooth network interoperability or change
+of the existing compatibility constants is required due to upgrade or security reasons.
+
+To synchronize all nodes in the compatibility zone to use the new set of the network parameters two RPC methods exist. The process
+requires human interaction and approval of the change. The rest of the process is automatic.
```
shamsasari: I'm not sure about the rest of the process being automatic, as the node has to be brought back up again after shutdown.

docs/source/network-map.rst
```diff
@@ -79,3 +79,46 @@ More parameters will be added in future releases to regulate things like allowed
...
+
+When the update is about to happen Network Map Service starts to advertise the additional information with the usual network map
```
shamsasari: _the_ network map service

PR #2413, opened 2018-01-23
allow cordapp developers to select fields from x500 names

roastario:
This a simple way for cordapp developers to select a sensible way to display a x500 cert or a Corda Identity (Party)

### DIFF
--- a/core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
+++ b/core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
@@ -81,10 +81,8 @@ data class CordaX500Name(val commonName: String?,
         const val MAX_LENGTH_STATE = 64
         const val MAX_LENGTH_ORGANISATION_UNIT = 64
         const val MAX_LENGTH_COMMON_NAME = 64
-
         private val supportedAttributes = setOf(BCStyle.O, BCStyle.C, BCStyle.L, BCStyle.CN, BCStyle.ST, BCStyle.OU)
         private val countryCodes: Set<String> = ImmutableSet.copyOf(Locale.getISOCountries() + unspecifiedCountry)
-
         @JvmStatic
         fun build(principal: X500Principal): CordaX500Name {
             val x500Name = X500Name.getInstance(principal.encoded)
@@ -106,13 +104,10 @@ data class CordaX500Name(val commonName: String?,
 
             val CN = attrsMap[BCStyle.CN]?.toString()
             val OU = attrsMap[BCStyle.OU]?.toString()
-            val O = attrsMap[BCStyle.O]?.toString()
-                    ?: throw IllegalArgumentException("Corda X.500 names must include an O attribute")
-            val L = attrsMap[BCStyle.L]?.toString()
-                    ?: throw IllegalArgumentException("Corda X.500 names must include an L attribute")
+            val O = attrsMap[BCStyle.O]?.toString() ?: throw IllegalArgumentException("Corda X.500 names must include an O attribute")
+            val L = attrsMap[BCStyle.L]?.toString() ?: throw IllegalArgumentException("Corda X.500 names must include an L attribute")
             val ST = attrsMap[BCStyle.ST]?.toString()
-            val C = attrsMap[BCStyle.C]?.toString()
-                    ?: throw IllegalArgumentException("Corda X.500 names must include an C attribute")
+            val C = attrsMap[BCStyle.C]?.toString() ?: throw IllegalArgumentException("Corda X.500 names must include an C attribute")
             return CordaX500Name(CN, OU, O, L, ST, C)
         }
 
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
```diff
@@ -120,9 +123,48 @@ data class CordaX500Name(val commonName: String?,
...
-        return _x500Principal ?: X500Principal(this.x500Name.encoded).also { _x500Principal = it }
-    }
+    val x500Principal: X500Principal
+        get() {
+            return _x500Principal ?: X500Principal(this.x500Name.encoded).also { _x500Principal = it }
+        }
 
     override fun toString(): String = x500Principal.toString()
+
+    fun toDisplayString(vararg selectors: CordaX500Name.NameSelector): String {
```
mikehearn: New public functions need KDocs.

core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
```diff
@@ -120,9 +123,48 @@ data class CordaX500Name(val commonName: String?,
...
+
+    fun toDisplayString(): String {
+        return toDisplayString(
+                CordaX500Name.NameSelector.COMMON_NAME,
+                CordaX500Name.NameSelector.ORG,
+                CordaX500Name.NameSelector.ORG_UNIT,
+                CordaX500Name.NameSelector.COUNTRY)
+    }
+
+    enum class NameSelector {
```
mikehearn: This would be a public enum, but `extract` is internal - is that deliberate? Why mix them up?

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -314,6 +314,14 @@ interface CordaRPCOps : RPCOps {
 
     /** Clear all network map data from local node cache. */
     fun clearNetworkMapCache()
+
+    fun displayNameFromParty(party: Party): String
```
mikehearn: New public API requires KDocs.

node/src/main/kotlin/net/corda/node/internal/RpcAuthorisationProxy.kt
```diff
@@ -20,6 +20,21 @@ import java.security.PublicKey
 
 // TODO change to KFunction reference after Kotlin fixes https://youtrack.jetbrains.com/issue/KT-12140
 class RpcAuthorisationProxy(private val implementation: CordaRPCOps, private val context: () -> RpcAuthContext) : CordaRPCOps {
+    override fun displayNameFromParty(party: Party): String {
```
mikehearn: Can you add a // TODO above to simplify this class using reflection proxies or kotlin delegation? It's kind of ugly that we need to add this here every time we add an RPC.

core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
roastario: yes, it's meant to be deliberate in the sense that they should be able to use the enums to specify their selected fields, but not to the extend that they could use the enum directly on the x500 - as I was hoping to keep all the formatting / joining of the strings together in one place. 

Might be a bit too paranoid / defensive.

core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
```diff
@@ -120,9 +119,88 @@ data class CordaX500Name(val commonName: String?,
...
-        return _x500Principal ?: X500Principal(this.x500Name.encoded).also { _x500Principal = it }
-    }
+    val x500Principal: X500Principal
+        get() {
+            return _x500Principal ?: X500Principal(this.x500Name.encoded).also { _x500Principal = it }
+        }
 
     override fun toString(): String = x500Principal.toString()
+
+    /** This will attempt to find a unique representation of this x500 name
```
mikehearn: Newline after the double stars

Remember this text will be often seen in a web browser, so:

* Full stops are needed at the end of every sentence, even when there's a newline in the source.
* Capital letters at the start of sentences ("uniqueness is" -> "Uniqueness is")
* Try re-reading the text to make sure it flows will. "till uniqueness" isn't quite grammatical, "until the string is unique" would be better.

core/src/test/kotlin/net/corda/core/identity/CordaX500NameTest.kt
```diff
@@ -73,4 +73,70 @@ class CordaX500NameTest {
...
+
+        assertEquals("blarg", displayString1)
+        assertEquals("org, blarg", displayString2)
+        assertEquals("org, unit, BL", displayString3)
+        assertEquals("org, unit, US, Blaargh", displayString4)
+        assertEquals("org, unit, US, Blargh", displayString5)
+        assertEquals("thisIs", displayString6)
+    }
+
+    @Test(expected = IllegalStateException::class)
```
mikehearn: Maybe ISE could be more specific?

PR #25, opened 2016-12-05
core: Remove CN from X509 subjectAlternativeNames, always add hostname

exFalso:
This PR fixes some issues in the cert code. Some old code still assumed that the CN is the hostname. Should fix ci

PR #2415, opened 2018-01-23
CORDA-692 Don't include internal packages in API documentation

andr3ej:


docs/packages.md
```diff
@@ -23,7 +23,7 @@ is complete.
 # Package net.corda.core.contracts
 
 This package contains the base data types for smarts contracts implemented in Corda. To implement a new contract start
-with [Contract], or see the examples in [net.corda.finance.contracts].
```
andr3ej: this kills dokkaJavadoc

gradle.properties
```diff
@@ -1 +1,2 @@
 kotlin.incremental=true
+org.gradle.jvmargs=-Xmx2g
```
andr3ej: dokka seems to need 1.5g to run comfortably, and dokkaJavadoc 2g+. i'm not sure if this is a good idea as it means your idle gradle daemon now uses 2g+, when by default it uses 1g+. couldn't find another solution that preserves the current ./gradlew apidocs incantation.

mikehearn: :( gradle is annoying for even needing a daemon at all. Sometimes I wish I'd gone with Maven. Oh well.

The G1 GC releases memory back to the OS when it does a full GC. However it never does a full GC unless forced. If we can call System.gc()  after running Dokka then memory usage of the daemon should go back down. Wanna try it?

docs/build.gradle
```diff
@@ -5,48 +5,57 @@ dependencies {
     compile rootProject
 }
 
+def internalPackagePrefixes(sourceDirs) {
+    def prefixes = []
+    // Kotlin allows packages to deviate from the directory structure, but let's assume they don't:
```
chrisr3: This comment reminds me of [The Evil Overlord List](http://www.eviloverlord.com/lists/overlord.html), specifically Rule #25:
> No matter how well it would perform, I will never construct any sort of machinery which is completely indestructible except for one small and virtually inaccessible vulnerable spot.

andr3ej: spinoff to https://r3-cev.atlassian.net/browse/CORDA-955

gradle.properties
andr3ej: * the original problem was the 3rd run of clean apidocs would grind to a halt due to gc
* G1 helps so i've implemented it, with 1g max heap as its default appears to be greater
* System.gc has no effect, now the 5th run grinds to a halt with or without it, there is almost certainly a memory leak in dokka
* any further work on this is outside the scope of the ticket

PR #2417, opened 2018-01-24
Reverted incorrect additional advertised RPC address in NodeInfo

sollecitom:


sollecitom:
Approval won't let it get merged without green tests regardless, but let's wait :)

PR #2419, opened 2018-01-25
Exposed RPC SSL settings through Cordformation

sollecitom:
Also, fixed broken demos and smoke tests.


samples/notary-demo/build.gradle
```diff
@@ -8,7 +8,7 @@ apply plugin: 'net.corda.plugins.publish-utils'
 apply plugin: 'net.corda.plugins.cordapp'
 apply plugin: 'net.corda.plugins.cordformation'
 apply plugin: 'maven-publish'
-
+Cord
```
IgorNitto: What is "Cord"?

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -154,7 +154,9 @@ open class Node(configuration: NodeConfiguration,
         val advertisedAddress = info.addresses.single()
 
         printBasicNodeInfo("Incoming connection address", advertisedAddress.toString())
-        rpcMessagingClient = RPCMessagingClient(configuration.rpcOptions.sslConfig, rpcServerAddresses.admin, networkParameters.maxMessageSize)
```
IgorNitto: These changes done seem to be related to CordFormation, just checking if they were included deliberately.

sollecitom: GItHub doesn't show the full change: https://github.com/corda/corda/pull/2419/files/756221e9331b34194d4b782c93b12630c87e4a10#diff-c300569916d7b2a3de4fca123d704329R157

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/SslOptions.kt
```diff
@@ -0,0 +1,50 @@
+package net.corda.plugins
```
IgorNitto: I am not sure I understand why you need to re-implement in Java the DSL extension (because CordformDefinition?); there must be a way to avoid this redundancy (and if its not, then its a fundamental design flaw somewhere).

PR #2425, opened 2018-01-26
changed driver to choose random ports

ags313:
Made driver choose random ports by default, instead of starting from 10000, to reduce "port already in use" failures.


shamsasari:
Random ports is known to be flaky.

PR #2426, opened 2018-01-26
 CORDA-959 Filter unschedules and remove database activity from inside mutex.

rick-r3:
Was showing up as point of contention in performance testing.


rick-r3:
Subsequently this also fixes a couple of bugs.

PR #2432, opened 2018-01-30
Upgrade Kotlin to 1.2.20

mikehearn:
I did that thing I'm good for these days ;)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ buildscript {
      * https://issues.apache.org/jira/browse/ARTEMIS-1559
      */
     ext.artemis_version = '2.2.0'
-    ext.jackson_version = '2.9.2'
+    ext.jackson_version = '2.9.3'
     ext.jetty_version = '9.4.7.v20170914'
     ext.jersey_version = '2.25'
     ext.jolokia_version = '1.3.7'
@@ -77,6 +77,7 @@ buildscript {
         mavenLocal()
         mavenCentral()
         jcenter()
+        // This repository is needed for Dokka until 0.9.16 is released.
         maven {
             url 'https://dl.bintray.com/kotlin/kotlin-eap/'
         }
--- a/client/jackson/build.gradle
+++ b/client/jackson/build.gradle
@@ -8,11 +8,11 @@ dependencies {
     compile project(':core')
     testCompile project(':test-utils')
 
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jre8:$kotlin_version"
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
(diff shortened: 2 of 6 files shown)


mikehearn:
This is hitting some issue where the kryo serialiser isn't doing the right thing for singletons anymore.

shamsasari:
I get a similar error on master in Intellij. Making `TestSingleton` public makes the problem go away.

The `targetType.kotlin.objectInstance` call in `CordaClassResolver#registerImplicit` is throwing an `IllegalAccessException`:
```
Class kotlin.reflect.jvm.internal.KClassImpl$Data$objectInstance$2 can not access a member of class net.corda.nodeapi.internal.serialization.KryoTests$TestSingleton with modifiers "public static final"
```

I wonder where the regression occurred as on 1.1.60.

mikehearn:
The problem is a JVM access check. The reflection library used to be able to reflect over `private object Foo` and determine it was a singleton. It's no longer possible to do that, I'm not sure why but presumably some small semantic change in Kotlin reflection. So it throws an exception which is then swallowed and ignored as "not an object" even though it is, resulting in the test failure. Making the object not private is a simple fix, but there may be other cases like this. Will investigate a bit further.

shamsasari:
Yes that's what I concluded. My hunch is that they removed the `setAccessible` call in `KClassImpl$Data$objectInstance$2`.

shamsasari:
The code has no `setAccessible`:
https://github.com/JetBrains/kotlin/blob/b5fa520aa507cb6fbbfc5f6c78752ecb9c7786a5/core/reflection.jvm/src/kotlin/reflect/jvm/internal/KClassImpl.kt#L104
But looking through history this is old code and hasn't been touched recently so I don't think it's ever had `setAccessible`.

mikehearn:
I debugged it a bit yesterday. I think the test has always been flaky and the Kotlin upgrade just revealed it. The question is how did it ever pass - Kryo *is* handing back the exact same object instance when using CompatibleFieldSerializer, but it's unclear how it does that or why. I wasn't able to debug very successfully because I couldn't reproduce the passing case inside IntelliJ and I didn't want to make a custom build of Kryo with print statements in just to satisfy my curiousity.

Anyway this commit fixes it: https://github.com/corda/corda/commit/38ccd0572c3ff186a039631b84df54dcf2e54e6d

mikehearn:
@shamsasari @Clintonio Can you review this please? It got a bit involved along the way.

shamsasari:
I've already approved. The button is green!

mikehearn:
Yeah, but it changed quite a bit since then. Can I get a review of the new changes?

shamsasari:
I already did.

mikehearn:
Ah, great. Thanks.

PR #2433, opened 2018-01-30
Basic node configuration validation.

sollecitom:


m4ksio:
Why not use https://github.com/lightbend/config#schemas-and-validation builtin?

sollecitom:
@m4ksio it's not how Config from Lightbend works when it comes to validating. It uses a reference comparison, which doesn't work in our case.

sollecitom:
CORDA-988: Fail on startup if RPC settings are invalid.
https://r3-cev.atlassian.net/browse/CORDA-988

PR #2434, opened 2018-01-30
CORDA-716 - Update changelog for test package changes and provide upgrade script

anthonykr3:
As requested by @mikehearn

### DIFF
--- a/tools/scripts/upgrade-test-packages.sh
+++ b/tools/scripts/upgrade-test-packages.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-find $1 -name "*.kt" -exec sed -i "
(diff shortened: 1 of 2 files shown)


tools/scripts/upgrade-test-packages.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
anthonykr3: I wasn't sure where to put the bash script - am happy to move it if it's not in the right place.

tools/scripts/upgrade-test-packages.sh
```diff
@@ -0,0 +1,6 @@
...
+
+find $1 -name "*.kt" -exec sed -i "
```
mikehearn: Maybe *.java can be included too.

PR #2436, opened 2018-01-30
Provides more information about the default initiating subflows. Minor tweaks.

joeldudleyr3:


docs/source/api-flows.rst
```diff
@@ -508,32 +505,34 @@ An example is the ``@InitiatingFlow InitiatorFlow``/``@InitiatedBy ResponderFlow
...
+^^^^^^^^^^^^^
+Corda installs four initiating subflow pairs on each node by default:
 
-Built-in subflows
-^^^^^^^^^^^^^^^^^
+* ``FinalityFlow``/``FinalityHandler``, which should be used to notarise and record a transaction and broadcast it to
+  all relevant parties
+* ``NotaryChangeFlow``/``NotaryChangeHandler``, which should be used to change a state's notary
+* ``ContractUpgradeFlow.Initiate``/``ContractUpgradeHandler``, which should be used to change a state's contract
+* ``SwapIdentitiesFlow``/``SwapIdentitiesHandler``, which is used to exchange confidential identities with a
```
shamsasari: This is part of the confidential identities work which isn't finalised yet so this needs to be remove.

 It's unfortunate that the node registers the handlers for it. It should be commented out.

joeldudleyr3: But these confidential identities flows are already usable, just not stabilised. We can still reference unstabilised APIs in the docs, because people still have to use them. Thoughts?

shamsasari: Confidential identities are in a separate module because we've not finalised on their design and will not automatically be part of the next release.

joeldudleyr3: Really? They're part of the current release.

shamsasari: @mikehearn will have to confirm.

mikehearn: confidential-identities is mentioned in the documentation and shipped, but is not stabilised. It's also not documented itself in the docsite. So it probably doesn't make sense to refer to it across the docs yet... we should finish it off, write proper docsite pages for it, then integrate it into the rest of the docsite. But we can't do that for Corda 3.0

So I'd suggest taking this mention out.

joeldudleyr3: @mikehearn It's heavily documented - here for example: https://docs.corda.net/api-identity.html.

mikehearn: Ah! I didn't realise that, sorry. Hmm -there's no warning on that page that the API isn't stable though! We seem to have half-launched it?

I also wasn't aware that these are top-level initiating flows: how come? I need to watch Ross' talk. But I'm a bit worried that this appears to have become a part of the wire protocol before being finished.

shamsasari: >I also wasn't aware that these are top-level initiating flows: how come?

I think it's OK because the initiating `SwapIdentitiesFlow` is defined in the confidential-identities module so the only way a developer can use this is to include that module which I believe we've already marked as experimental. Perhaps we just need to make sure the docsite makes clear `SwapIdentitiesFlow` is experimental.

docs/source/running-a-node.rst
```diff
@@ -19,7 +19,7 @@ Start the nodes with ``runnodes`` by running the following command from the root
 * Linux/macOS: ``build/nodes/runnodes``
 * Windows: ``call build\nodes\runnodes.bat``
 
-.. warn:: On macOS, do not click/change focus until all the node terminal windows have opened, or some processes may
+.. warning:: On macOS, do not click/change focus until all the node terminal windows have opened, or some processes may
```
shamsasari: Curious to know what the difference is

joeldudleyr3: Only the latter works :D

docs/source/api-flows.rst
mikehearn: Yeah, making it clear those are experimental would be good. But I really meant, why is the API designed that way to begin with....

docs/source/api-flows.rst
```diff
@@ -508,32 +505,37 @@ An example is the ``@InitiatingFlow InitiatorFlow``/``@InitiatedBy ResponderFlow
...
+  counterparty
 
-Built-in subflows
-^^^^^^^^^^^^^^^^^
+.. warning:: The ``confidential-identities`` module is still not stabilised, so the
+   ``SwapIdentitiesFlow``/``SwapIdentitiesHandler`` API may change in future releases. See :doc:`corda-api`.
```
mikehearn: You should remark here also that it's not even installed unless the module is included.

PR #2437, opened 2018-01-30
Docker system (end-to-end) test

m4ksio:


constants.properties
```diff
@@ -1,8 +1,9 @@
...
 kotlinVersion=1.1.60
 platformVersion=2
 guavaVersion=21.0
 bouncycastleVersion=1.57
 typesafeConfigVersion=1.3.1
 jsr305Version=3.0.2
-artifactoryPluginVersion=4.4.18
\ No newline at end of file
+artifactoryPluginVersion=4.4.18
+snakeYamlVersion=1.19
```
sollecitom: Do we want to change version of snakeYaml in between runs?

gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -90,8 +110,9 @@ class Node(private val project: Project) : CordformNode() {
         // with loading our custom X509EdDSAEngine.
         val organizationName = name.trim().split(",").firstOrNull { it.startsWith("O=") }?.substringAfter("=")
         val dirName = organizationName ?: name
+        containerName = dirName.replace("\\s+".toRegex(), "-").toLowerCase()
         this.rootDir = rootDir.toFile()
-        nodeDir = File(this.rootDir, dirName)
+        nodeDir = File(this.rootDir, dirName.replace("\\s", ""))
```
Clintonio: Can you add the stripping to the documentation of the node name?

PR #2438, opened 2018-01-31
CORDA-939 - Dont expose FlowStateMachine via public API

anthonykr3:
Move FlowStateMachine from net.corda.core.internal to net.corda.core.flows and add the DoNotImplement attribute

I looked at other ways trying to stop it being exposed, however the fact that it is shared between net.corda.core and net.corda.node really prohibits that.

It looks like the DoNotImplement attribute is the way we have handled this situation elsewhere. I also noticed that the FlowStateMachineImpl is part of the public API, in which case hiding the FlowStateMachine doesn't seem to make that much difference anyway.


anthonykr3:
Adding @mnesbit as a reviewer as I've had to change the api-current.txt

sollecitom:
Hey @anthonykr3 please ask @exFalso to review this.

mikehearn:
`FlowStateMachineImpl` is in the node module, which isn't public API. 

I see why you've done this - we expose the internal thing in a field of FlowLogic. But that field does have a comment that says it's for internal use only. We could rename it to something more visibly internal perhaps: `__stateMachine` or something ugly and C-like, and then teach the tool that this is a special convention for "public but not stable". But I'm not sure moving this internal interface into the public API is the right way to go. It's not meant for real usage.

.ci/api-current.txt
```diff
@@ -4153,7 +4153,7 @@ public final class net.corda.testing.node.NodeTestUtils extends java.lang.Object
   @org.jetbrains.annotations.NotNull public static final net.corda.testing.dsl.LedgerDSL ledger(net.corda.core.node.ServiceHub, kotlin.jvm.functions.Function1)
   @org.jetbrains.annotations.NotNull public static final net.corda.testing.dsl.LedgerDSL ledger(net.corda.core.node.ServiceHub, net.corda.core.identity.Party, kotlin.jvm.functions.Function1)
   @org.jetbrains.annotations.NotNull public static final net.corda.core.context.InvocationContext newContext(net.corda.node.services.api.StartedNodeServices)
-  @org.jetbrains.annotations.NotNull public static final net.corda.core.internal.FlowStateMachine startFlow(net.corda.node.services.api.StartedNodeServices, net.corda.core.flows.FlowLogic)
```
anthonykr3: This API wasn't part of V2 - so this isn't an actual api stability change

anthonykr3:
Changed the test functions to just return a Future directly where possible as that is all the state machine is ever used for in tests.

@mikehearn I've implemented a "CordaInternal" attribute as discussed for the public stateFlow field on FlowLogic

@joeldudleyr3 This does change some [example code](https://github.com/corda/corda/pull/2438/files#diff-c776a491dfbac29b663b6f6af27a0035L15)

joeldudleyr3:
@anthonykr3 Returning a future directly works better for writing tests. But can't `startFlowAndReturnFuture` just be called `startFlow`? The former is a bit verbose.

anthonykr3:
I can change it to startFlow - I wasn't sure about the name, whether we needed to be explicit that we were returning a future or not, I suppose it's obvious from the message signature.

core/src/main/kotlin/net/corda/core/CordaInternal.kt
```diff
@@ -0,0 +1,15 @@
+package net.corda.core
+
+import java.lang.annotation.Inherited
+
+/**
+ * These fields are only meant to be used by Corda internally, and are not intended to be part of the public API.
+ *
+ * These methods are not part of Corda's API compatibility guarantee and applications should not use them.
```
andr3ej: i'd have this as the first sentence, as the javadoc tool uses the first sentence as a summary and this is what users really need to see

core/src/main/kotlin/net/corda/core/CordaInternal.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER)
+@MustBeDocumented
+@Inherited
```
andr3ej: does this actually have an effect? i believe that in java it makes a class annotation inherited by subclasses and that's it, but this isn't a class annotation. maybe there's some kotlin voodoo i'm unaware of.

anthonykr3: ...copy paste fail from the DoNotImplement attribute next door

Issue #2326, opened 2018-01-05
Can't name a node with the words "server" or "node" in the name!


roger3cev:
Whilst the motivation for this makes complete sense, the rules for determining whether a name should be allowed or not, should be more complicated than:

`require(!legalName.contains(it, true)) { "Character not allowed in legal names: $it" }`

:)

https://beta.companieshouse.gov.uk/search?q=node
https://beta.companieshouse.gov.uk/search?q=server

mikehearn:
Good point. That rule should be enforced in the doorman rather than as a "consensus rule" in every node.

shamsasari:
So that means removing the check
```
LegalNameValidator.validateOrganization(organisation, LegalNameValidator.Validation.MINIMAL)
```
from `CordaX500Name`.

shamsasari:
While we at it the checks
```
        require(country.length == LENGTH_COUNTRY) { "Invalid country '$country' Country code must be $LENGTH_COUNTRY letters ISO code " }
        require(country.toUpperCase() == country) { "Country code should be in upper case." }
```
can probably be removed since we doing a lookup in `Locale.getISOCountries()` anyway.

mikehearn:
Removed as long as they are still present in our doorman ID checking flow, yes.

patrickkuo:
this check is removed in corda v3, closing this issue.

PR #2439, opened 2018-01-31
Remove word restriction in corda x500 name

patrickkuo:
https://github.com/corda/corda/issues/2326

### DIFF
--- a/core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
+++ b/core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
@@ -50,9 +50,6 @@ data class CordaX500Name(val commonName: String?,
         // Legal name checks.
         LegalNameValidator.validateOrganization(organisation, LegalNameValidator.Validation.MINIMAL)
 
-        // Attribute data width checks.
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
```diff
@@ -74,7 +71,6 @@ data class CordaX500Name(val commonName: String?,
     }
 
     companion object {
-        const val LENGTH_COUNTRY = 2
```
shamsasari: Ouch, we can't remove this as it's now public API. Just deprecate it and say it's no longer used.

patrickkuo: :( ok

shamsasari:
The change log needs to be updated as well.

docs/source/changelog.rst
```diff
@@ -6,6 +6,8 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
+* Removed blacklisted word checks in corda x500 name to allow "Server" or "Node" to be use as part of the legal name.
```
shamsasari: Corda X.500

PR #2440, opened 2018-01-31
CORDA-943 - Fix trader demo

fenryka:
This is a multi issue problem
1. Fingerprinting of generics treats <T> and <?> differently, forcing
the evolver to be used when not needed

2. However, the evolver is required sometimes as generics are not
guaranteed to fingerprinting bi-directionally (thanks to type erasure of
deeply nested generic types). However, with serialization now writing
properties in a specific order, we need to ensure they're read back in
that order before applying them to an evolved constructor so as to
not corrupt the object reference cache

3. Keep some helpful debug parameters

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPSerializer.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPSerializer.kt
@@ -30,7 +30,7 @@ interface AMQPSerializer<out T> {
     /**
      * Write the given object, with declared type, to the output.
      */
-    fun writeObject(obj: Any, data: Data, type: Type, output: SerializationOutput)
+    fun writeObject(obj: Any, data: Data, type: Type, output: SerializationOutput, offset: Int = 0)
 
     /**
      * Read the given object from the input. The envelope is provided in case the schema is required.
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPSerializer.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPSerializer.kt
@@ -30,7 +30,7 @@ interface AMQPSerializer<out T> {
     /**
      * Write the given object, with declared type, to the output.
      */
-    fun writeObject(obj: Any, data: Data, type: Type, output: SerializationOutput, offset: Int = 0)
+    fun writeObject(obj: Any, data: Data, type: Type, output: SerializationOutput, debugIndent: Int = 0)
 
     /**
      * Read the given object from the input. The envelope is provided in case the schema is required.
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
@@ -1,5 +1,6 @@
 package net.corda.nodeapi.internal.serialization.amqp
 
+import com.google.common.primitives.Primitives
 import net.corda.core.internal.getStackTraceAsString
 import net.corda.core.serialization.SerializedBytes
 import net.corda.core.utilities.ByteSequence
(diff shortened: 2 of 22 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPSerializer.kt
```diff
@@ -30,7 +30,7 @@ interface AMQPSerializer<out T> {
     /**
      * Write the given object, with declared type, to the output.
      */
-    fun writeObject(obj: Any, data: Data, type: Type, output: SerializationOutput)
+    fun writeObject(obj: Any, data: Data, type: Type, output: SerializationOutput, offset: Int = 0)
```
rick-r3: If you were to refactor rename `offset` here to e.g. `debugNestingLevel` and make all the `+4`s just `+1`s, would that be okay?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
```diff
@@ -106,41 +106,47 @@ class DeserializationInput(internal val serializerFactory: SerializerFactory) {
...
-                // Skip for primitive types as they are too small and overhead of referencing them will be much higher than their content
-                if (suitableForObjectReference(objectRead.javaClass)) objectHistory.add(objectRead)
-                objectRead
+            // Store the reference in case we need it later on.
+            // Skip for primitive types as they are too small and overhead of referencing them will be much higher than their content
+            if (suitableForObjectReference(objectRead.javaClass)) {
+                objectHistory.add(objectRead)
             }
+            objectRead
+        }
```
rick-r3: Is this (class) all just reformatting except for the offset?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/Schema.kt
```diff
@@ -379,14 +381,20 @@ private fun fingerprintForType(type: Type, contextType: Type?, alreadySeen: Muta
...
-            // serialising a concrete example of the type we have A and B which are TypeVariables<*>'s but
-            // when deserializing we only have the wildcard placeholder ?, or AnyType
+            // Previously, we drew a distinction between TypeVariable, Wildcard, and AnyType, changing
+            // the signature of the fingerprinted object. This, however, doesn't work as it breaks bi-
+            // directional fingerprints. That is, fingerprinting a concrete instance of a generic
+            // type (Example<Int>), creates a different fingerprint from the generic type itself (Example<T>)
+            //
+            // On serialization Example<Int> is treated as Example<T>, a TypeVariable
+            // On deserialisation it is seen as Example<?>, A wildcard *and* a TypeVariable
+            //      Note: AnyType is a special case of WildcarcType used in other parts of the
```
rick-r3: `WildcardType`

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/Schema.kt
```diff
@@ -379,14 +381,20 @@ private fun fingerprintForType(type: Type, contextType: Type?, alreadySeen: Muta
                         fingerprintForType(paramType, type, alreadySeen, orig, factory, offset+4)
                     }
                 }
-            // Treat generic types as "any type" to prevent fingerprint mismatch. This case we fall into when
-            // looking at A and B from Example<A, B> (remember we call this function recursively). When
-            // serialising a concrete example of the type we have A and B which are TypeVariables<*>'s but
-            // when deserializing we only have the wildcard placeholder ?, or AnyType
+            // Previously, we drew a distinction between TypeVariable, Wildcard, and AnyType, changing
```
rick-r3: Should that be `WildcardType` rather than `Wildcard`?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/carpenter/AMQPSchemaExtensions.kt
```diff
@@ -124,7 +124,12 @@ fun AMQPField.getTypeAsClass(classloader: ClassLoader) = typeStrToType[Pair(type
     "string" -> String::class.java
     "binary" -> ByteArray::class.java
     "*" -> if (requires.isEmpty()) Any::class.java else classloader.loadClass(requires[0])
-    else -> classloader.loadClass(type)
+    else -> {
+        classloader.loadClass(
+                if (type.endsWith("<?>")) {
```
rick-r3: What happens with multiple generics.  e.g. If I have `Foo<S, T>`, is that encoded as `Foo<?>` or `Foo<?, ?>` ?

fenryka: This is a good point!!!

Will add a test

fenryka: fixed

PR #2441, opened 2018-01-31
Added option to enable docs only builds if a specific file is present

Clintonio:
Why: for CI


Clintonio:
Addressed comments and added to gitignore.

PR #2445, opened 2018-02-01
set network registration poll interval via http cache control header

patrickkuo:
https://r3-cev.atlassian.net/browse/CORDA-923


shamsasari:
https://github.com/corda/enterprise/pull/434

PR #2446, opened 2018-02-01
Use Java reflection for determining if a type is a Kotlin singleton.

mikehearn:
Kotlin's own reflection has a habit of throwing weird errors and does not work for private objects - thus the unit test which tests this feature has actually never worked properly, but somehow works by accident. An attempt to upgrade to the latest Kotlin revealed the issue so it must have always been unstable.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
@@ -6,14 +6,15 @@ import com.esotericsoftware.kryo.io.Output
 import com.esotericsoftware.kryo.serializers.FieldSerializer
 import com.esotericsoftware.kryo.util.DefaultClassResolver
 import com.esotericsoftware.kryo.util.Util
-import net.corda.nodeapi.internal.AttachmentsClassLoader
 import net.corda.core.serialization.ClassWhitelist
 import net.corda.core.serialization.CordaSerializable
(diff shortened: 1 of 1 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/CordaClassResolver.kt
```diff
@@ -74,10 +75,21 @@ class CordaClassResolver(serializationContext: SerializationContext) : DefaultCl
...
+            targetType.declaredFields.singleOrNull {
+                it.name == "INSTANCE" &&
+                it.type == type &&
+                Modifier.isStatic(it.modifiers) &&
+                Modifier.isFinal(it.modifiers) &&
+                Modifier.isPublic(it.modifiers)
+            }?.let {
+                it.isAccessible = true
+                type.cast(it.get(null)!!)
+            }
```
fenryka: can it be null? ?. implies it can, but we never handle that case

mikehearn: Yes, if there is no INSTANCE field at all then `singleOrNull` will return null, and we want to skip the `let` code block.

fenryka: ahhh, ok, shiny!!! I was assuming if it was null we'd want it to be something else but if null is what we're ok with it's fine

PR #2447, opened 2018-02-01
Tweak the way invocation context is integrated to reduce the pain for devs using the old API

mikehearn:
Mostly this removes a lot of deprecation warnings.


sollecitom:
@mikehearn you'll have to update the public API.

PR #2448, opened 2018-02-01
Init CRaSH shell only when it's really needed

m4ksio:


node/src/main/kotlin/net/corda/node/ArgsParser.kt
```diff
@@ -93,7 +94,9 @@ data class CmdLineOptions(val baseDirectory: Path,
                           val justGenerateNodeInfo: Boolean,
                           val bootstrapRaftCluster: Boolean) {
     fun loadConfig(): NodeConfiguration {
-        val config = ConfigHelper.loadConfig(baseDirectory, configFile).parseAsNodeConfiguration()
+        val config = ConfigHelper.loadConfig(baseDirectory, configFile, configOverrides = ConfigFactory.parseMap(
```
sollecitom: I'm really not a fan of this library. Here, conceptually, the parsing should stay untouched, and we should override after, but I'm aware this is painful with our current configuration code.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -278,7 +275,9 @@ abstract class AbstractNode(val configuration: NodeConfiguration,
     protected abstract fun getRxIoScheduler(): Scheduler
 
     open fun startShell(rpcOps: CordaRPCOps) {
-        InteractiveShell.startShell(configuration, rpcOps, securityManager, _services.identityService, _services.database)
+        if (configuration.shouldInitCrashShell()) {
```
sollecitom: Why do we check the property within this function, rather than around it?

```
open fun startShell(rpcOps: CordaRPCOps) {
    if (configuration.shouldInitCrashShell()) {
```

Compared to

```
if (configuration.shouldInitCrashShell()) {
    startShell(rpcOps)
```

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -45,6 +45,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val detectPublicIp: Boolean get() = true
     val sshd: SSHDConfiguration?
     val database: DatabaseConfig
+    val noLocalShell: Boolean get() = false
```
sollecitom: `val startLocalShell: Boolean get() = true` instead?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
m4ksio: Because than this function can be overridden in MockNode.

node/src/main/kotlin/net/corda/node/ArgsParser.kt
m4ksio: Library is fine, just they way we use it... Problem here there is very clear separation in code between config and command line args, while in fact some args are configuration entries (like this noLocalShell).

m4ksio:
It only enables shell if there is a need for it, so as long as notary node does not use local shell or SSH this should not load CRaSH

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
m4ksio: We use --no-local-shell flag, so I would like to keep it consistent

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
sollecitom: But even in the mock node, if the option is `shouldInitCrashShell() == false`, nothing should be called.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
sollecitom: Fair enough.

PR #2462, opened 2018-02-03
CORDA-978 - Only consider getters that accpet zero parameters

fenryka:
Because of inheritance, we have to consider multiple getters existing on a class for a property, however in that, we weren't excluding getters that take parameters, something we're unable to provide, so depending no the returned order sometimes we select the wrong one and it all goes terribley wrong.


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -127,36 +127,45 @@ fun Class<out Any?>.propertyDescriptors(): Map<String, PropertyDescriptor> {
         clazz!!.declaredFields.forEach { classProperties.put(it.name, PropertyDescriptor(it, null, null)) }
 
         // then pair them up with the declared getter and setter
-        // Note: It is possible for a class to have multipleinstancess of a function where the types
+        // Note: It is possible for a class to have multiple instancess of a function where the types
```
rick-r3: Do we have a test where the type has been narrowed?

PR #2465, opened 2018-02-05
CORDA-961 Wire up and enforce max transaction size

patrickkuo:
WIP - PR opened for CI test

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -3590,7 +3590,7 @@ public static final class net.corda.client.jackson.StringToMethodCallParser$Unpa
 ##
 public final class net.corda.testing.driver.Driver extends java.lang.Object
   public static final Object driver(net.corda.testing.driver.DriverParameters, kotlin.jvm.functions.Function1)
-  public static final Object driver(net.corda.testing.driver.DriverParameters, boolean, java.nio.file.Path, net.corda.testing.driver.PortAllocation, net.corda.testing.driver.PortAllocation, Map, boolean, boolean, boolean, boolean, List, List, net.corda.testing.driver.JmxPolicy, kotlin.jvm.functions.Function1)
+  public static final Object driver(net.corda.testing.driver.DriverParameters, boolean, java.nio.file.Path, net.corda.testing.driver.PortAllocation, net.corda.testing.driver.PortAllocation, Map, boolean, boolean, boolean, boolean, List, List, net.corda.testing.driver.JmxPolicy, int, kotlin.jvm.functions.Function1)
 ##
 @net.corda.core.DoNotImplement public interface net.corda.testing.driver.DriverDSL
   @org.jetbrains.annotations.NotNull public abstract java.nio.file.Path baseDirectory(net.corda.core.identity.CordaX500Name)
@@ -3605,11 +3605,12 @@ public final class net.corda.testing.driver.Driver extends java.lang.Object
 ##
 public final class net.corda.testing.driver.DriverParameters extends java.lang.Object
   public <init>()
-  public <init>(boolean, java.nio.file.Path, net.corda.testing.driver.PortAllocation, net.corda.testing.driver.PortAllocation, Map, boolean, boolean, boolean, boolean, List, List, net.corda.testing.driver.JmxPolicy)
+  public <init>(boolean, java.nio.file.Path, net.corda.testing.driver.PortAllocation, net.corda.testing.driver.PortAllocation, Map, boolean, boolean, boolean, boolean, List, List, net.corda.testing.driver.JmxPolicy, int)
   public final boolean component1()
   @org.jetbrains.annotations.NotNull public final List component10()
   @org.jetbrains.annotations.NotNull public final List component11()
   @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.JmxPolicy component12()
+  public final int component13()
   @org.jetbrains.annotations.NotNull public final java.nio.file.Path component2()
   @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.PortAllocation component3()
   @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.PortAllocation component4()
@@ -3618,13 +3619,14 @@ public final class net.corda.testing.driver.DriverParameters extends java.lang.O
   public final boolean component7()
   public final boolean component8()
   public final boolean component9()
-  @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.DriverParameters copy(boolean, java.nio.file.Path, net.corda.testing.driver.PortAllocation, net.corda.testing.driver.PortAllocation, Map, boolean, boolean, boolean, boolean, List, List, net.corda.testing.driver.JmxPolicy)
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.DriverParameters copy(boolean, java.nio.file.Path, net.corda.testing.driver.PortAllocation, net.corda.testing.driver.PortAllocation, Map, boolean, boolean, boolean, boolean, List, List, net.corda.testing.driver.JmxPolicy, int)
   public boolean equals(Object)
   @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.PortAllocation getDebugPortAllocation()
   @org.jetbrains.annotations.NotNull public final java.nio.file.Path getDriverDirectory()
   @org.jetbrains.annotations.NotNull public final List getExtraCordappPackagesToScan()
   public final boolean getInitialiseSerialization()
   @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.JmxPolicy getJmxPolicy()
+  public final int getMaxTransactionSize()
   @org.jetbrains.annotations.NotNull public final List getNotarySpecs()
   @org.jetbrains.annotations.NotNull public final net.corda.testing.driver.PortAllocation getPortAllocation()
   public final boolean getStartNodesInProcess()
@@ -3898,7 +3900,7 @@ public final class net.corda.testing.node.MockKeyManagementService extends net.c
 public class net.corda.testing.node.MockNetwork extends java.lang.Object
   public <init>(List)
   public <init>(List, net.corda.testing.node.MockNetworkParameters)
-  public <init>(List, net.corda.testing.node.MockNetworkParameters, boolean, boolean, net.corda.testing.node.InMemoryMessagingNetwork$ServicePeerAllocationStrategy, kotlin.jvm.functions.Function1, boolean, List)
+  public <init>(List, net.corda.testing.node.MockNetworkParameters, boolean, boolean, net.corda.testing.node.InMemoryMessagingNetwork$ServicePeerAllocationStrategy, kotlin.jvm.functions.Function1, boolean, List, int)
   @org.jetbrains.annotations.NotNull public final net.corda.testing.node.MockNetwork$MockNode addressToNode(net.corda.core.messaging.MessageRecipients)
   @org.jetbrains.annotations.NotNull public final java.nio.file.Path baseDirectory(int)
   @org.jetbrains.annotations.NotNull public final net.corda.node.internal.StartedNode createNode(net.corda.testing.node.MockNodeParameters)
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -280,6 +280,7 @@ public static final class net.corda.core.contracts.AmountTransfer$Companion exte
 @net.corda.core.serialization.CordaSerializable public interface net.corda.core.contracts.Attachment extends net.corda.core.contracts.NamedByHash
   public abstract void extractFile(String, java.io.OutputStream)
   @org.jetbrains.annotations.NotNull public abstract List getSigners()
+  public abstract int getSize()
   @org.jetbrains.annotations.NotNull public abstract java.io.InputStream open()
   @org.jetbrains.annotations.NotNull public abstract jar.JarInputStream openAsJAR()
 ##
@@ -347,6 +348,7 @@ public final class net.corda.core.contracts.ComponentGroupEnum extends java.lang
   @org.jetbrains.annotations.NotNull public final String getContract()
   @org.jetbrains.annotations.NotNull public net.corda.core.crypto.SecureHash getId()
   @org.jetbrains.annotations.NotNull public List getSigners()
+  public int getSize()
   @org.jetbrains.annotations.NotNull public java.io.InputStream open()
   @org.jetbrains.annotations.NotNull public jar.JarInputStream openAsJAR()
 ##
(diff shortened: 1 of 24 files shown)


core/src/main/kotlin/net/corda/core/contracts/Attachment.kt
```diff
@@ -50,4 +50,9 @@ interface Attachment : NamedByHash {
      * Can be empty, for example non-contract attachments won't be necessarily be signed.
      */
     val signers: List<Party>
+
+    /**
+     * Attachment size.
```
shamsasari: in bytes

core/src/main/kotlin/net/corda/core/internal/AbstractAttachment.kt
```diff
@@ -27,6 +27,10 @@ abstract class AbstractAttachment(dataLoader: () -> ByteArray) : Attachment {
     }
 
     protected val attachmentData: ByteArray by lazy(dataLoader)
+
+    // TODO: read file size information from metadata instead of loading the data.
+    override val size: Int by lazy { attachmentData.size }
```
shamsasari: `attachmentData` is already lazy this can get use the getter: `get() = attachmentData.size`

core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
```diff
@@ -60,6 +60,8 @@ interface ServicesForResolution : StateLoader {
 
     /** Provides access to anything relating to cordapps including contract attachment resolution and app context */
     val cordappProvider: CordappProvider
+
+    val maxTransactionSize: Int
```
shamsasari: @kasiastreich is moving the network parameters to core, so it makes sense to have that instead of this. Add a TODO in case this PR gets merged in before hers.

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -104,7 +105,8 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
             resolveIdentity: (PublicKey) -> Party?,
             resolveAttachment: (SecureHash) -> Attachment?,
             resolveStateRef: (StateRef) -> TransactionState<*>?,
-            resolveContractAttachment: (TransactionState<ContractState>) -> AttachmentId?
+            resolveContractAttachment: (TransactionState<ContractState>) -> AttachmentId?,
+            maxTransactionSize: Int
```
shamsasari: It's unfortunate that this is public API. Do we overload this with a default value (but what value?)? Or perhaps we break ABI for this as this should never been public? @mikehearn ?

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -118,7 +120,22 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
...
+        val ltx = LedgerTransaction(resolvedInputs, outputs, authenticatedArgs, attachments, id, notary, timeWindow, privacySalt)
+        checkTransactionSize(ltx, maxTransactionSize)
+        return ltx
+    }
+
+    private fun checkTransactionSize(ltx: LedgerTransaction, maxTransactionSize: Int) {
+        var remainingTransactionSize = maxTransactionSize
+        // Check attachment size first as they are most likely to go over the limit.
+        ltx.attachments.forEach {
+            remainingTransactionSize = Math.subtractExact(remainingTransactionSize, it.size)
```
shamsasari: Use `Long` instead

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -118,7 +120,22 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
...
+            require(remainingTransactionSize > 0) { "Transaction exceeded network's maximum transaction size limit : $maxTransactionSize bytes." }
+        }
+        listOf(ltx.inputs, ltx.commands, ltx.outputs).forEach {
```
shamsasari: Use a method (perhaps local) instead of creating this list

node/src/integration-test/kotlin/net/corda/node/services/statemachine/LargeTransactionsTest.kt
```diff
@@ -84,4 +85,28 @@ class LargeTransactionsTest {
...
     }
+
+    @Test
+    fun `check transaction will fail when exceed max transaction size limit`() {
+        // These 4 attachments yield a transaction that's got ~ 4mb, which will exceed the 3mb max transaction size limit
+        val bigFile1 = InputStreamAndHash.createInMemoryTestZip(1024 * 1024, 0)
+        val bigFile2 = InputStreamAndHash.createInMemoryTestZip(1024 * 1024, 1)
+        val bigFile3 = InputStreamAndHash.createInMemoryTestZip(1024 * 1024, 2)
+        val bigFile4 = InputStreamAndHash.createInMemoryTestZip(1024 * 1024, 3)
+        driver(startNodesInProcess = true, extraCordappPackagesToScan = listOf("net.corda.testing.contracts"), portAllocation = PortAllocation.RandomFree, maxTransactionSize = 3_000_000) {
```
shamsasari: Don't use random port allocation. It causes flaky tests

shamsasari: Use `internalDriver` instead which exposes `compatibilityZone` and allows you to inject in your own custom network parameters (see `NetworkMapTest`)

node/src/integration-test/kotlin/net/corda/node/services/statemachine/LargeTransactionsTest.kt
```diff
@@ -84,4 +85,28 @@ class LargeTransactionsTest {
...
+            val rpcUser = User("admin", "admin", setOf("ALL"))
+            val (alice, _) = listOf(ALICE_NAME, BOB_NAME).map { startNode(providedName = it, rpcUsers = listOf(rpcUser)) }.transpose().getOrThrow()
+            alice.rpcClientToNode().use(rpcUser.username, rpcUser.password) {
+                val hash1 = it.proxy.uploadAttachment(bigFile1.inputStream)
+                val hash2 = it.proxy.uploadAttachment(bigFile2.inputStream)
+                val hash3 = it.proxy.uploadAttachment(bigFile3.inputStream)
+                val hash4 = it.proxy.uploadAttachment(bigFile4.inputStream)
+                assertEquals(hash1, bigFile1.sha256)
+
+                assertFailsWith<UnexpectedFlowEndException> {
```
shamsasari: This exception is too vague, we need access to the original. This would be possible if using the MockNetwork. Looking at this I'm not sure we need to use the driver. Our integration tests are already lengthy as is.

node/src/integration-test/kotlin/net/corda/node/services/statemachine/LargeTransactionsTest.kt
```diff
@@ -70,7 +71,7 @@ class LargeTransactionsTest {
         val bigFile2 = InputStreamAndHash.createInMemoryTestZip(1024 * 1024 * 3, 1)
         val bigFile3 = InputStreamAndHash.createInMemoryTestZip(1024 * 1024 * 3, 2)
         val bigFile4 = InputStreamAndHash.createInMemoryTestZip(1024 * 1024 * 3, 3)
-        driver(startNodesInProcess = true, extraCordappPackagesToScan = listOf("net.corda.testing.contracts"), portAllocation = PortAllocation.RandomFree) {
+        driver(startNodesInProcess = true, extraCordappPackagesToScan = listOf("net.corda.testing.contracts"), portAllocation = PortAllocation.RandomFree, maxTransactionSize = Int.MAX_VALUE) {
```
kasiastreich: We don't need to extend driver with additional parameter, you can override network parameters.

kasiastreich:
Did you check that Demobench and samples work? I can see that we still have in Demobench fake transaction size. See `NodeController.kt`

node/src/integration-test/kotlin/net/corda/node/services/statemachine/LargeTransactionsTest.kt
patrickkuo: I am going to rewrite the test using mock network instead, saving us from changing the driver.

.ci/api-current.txt
```diff
@@ -3074,7 +3076,6 @@ public class net.corda.core.transactions.TransactionBuilder extends java.lang.Ob
   @org.jetbrains.annotations.NotNull public final net.corda.core.contracts.PrivacySalt getPrivacySalt()
   @org.jetbrains.annotations.NotNull public final Set getRequiredSigningKeys()
   public int hashCode()
-  @org.jetbrains.annotations.NotNull public final net.corda.core.transactions.LedgerTransaction toLedgerTransaction(kotlin.jvm.functions.Function1, kotlin.jvm.functions.Function1, kotlin.jvm.functions.Function1, kotlin.jvm.functions.Function1)
```
patrickkuo: need some advice on what to do with this method....

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
mikehearn: Hm, I'm not sure it was ever intended to be private.

My guess is that what we're suffering from here is lack of easy access to the current network parameters? So it'the relevant parts are being wired through every function prototype? 

If we had a global getCurrentNetParams() function, you wouldn't need to add params like these right?

patrickkuo: That's a better solution, it is a pain trying to wire in the network parameters.

core/src/main/kotlin/net/corda/core/internal/NodeProperties.kt
```diff
@@ -0,0 +1,14 @@
+package net.corda.core.internal
+
+import net.corda.core.node.NetworkParameters
+
+object NodeProperties {
+    // TODO: Use [NetworkParameters] once it's available in corda core.
```
shamsasari: Remove

core/src/main/kotlin/net/corda/core/internal/NodeProperties.kt
```diff
@@ -0,0 +1,14 @@
...
+    private var _networkParameters: NetworkParameters? = null
+
+    var networkParameters: NetworkParameters
+        get() = _networkParameters ?: throw IllegalArgumentException("Property 'networkParameters' has not been initialised.")
```
shamsasari: `get() = checkNotNull(_networkParameters) {...`

docs/source/changelog.rst
```diff
@@ -6,6 +6,9 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
+* Introduced max transaction size limit on transactions. The max transaction size parameter is set by the compatibility zone
```
shamsasari: There is an entire section below on the network map. Move this there.

docs/source/network-map.rst
```diff
@@ -71,6 +71,7 @@ The current set of network parameters:
 :maxMessageSize: Maximum allowed size in bytes of an individual message sent over the wire. Note that attachments are
         a special case and may be fragmented for streaming transfer, however, an individual transaction or flow message
         may not be larger than this value.
+:maxTransactionSize: Maximum allowed size in bytes of a transaction. Which include the size of the transaction object and its attachments.
```
shamsasari: This is the size of

node/src/test/kotlin/net/corda/node/services/transactions/MaxTransactionSizeTests.kt
```diff
@@ -0,0 +1,136 @@
...
+            val hash4 = aliceServices.attachments.importAttachment(bigFile4.inputStream)
+            assertEquals(hash1, bigFile1.sha256)
+            SendLargeTransactionFlow(notary, bob, hash1, hash2, hash3, hash4)
+        }
+        val exception = assertFailsWith<IllegalArgumentException> {
+            val future = aliceServices.startFlow(flow)
+            mockNet.runNetwork()
+            future.getOrThrow()
+        }
+        assertTrue(exception.message?.contains("Transaction exceeded network's maximum transaction size limit") ?: false)
```
shamsasari: `assertThat(exception).hasMessageContaining(...`

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -126,6 +128,10 @@ open class MockServices private constructor(
         }
     }
 
+    init {
```
shamsasari: Put this in the companion object so that it's only done once

node/src/test/kotlin/net/corda/node/services/transactions/MaxTransactionSizeTests.kt
```diff
@@ -0,0 +1,137 @@
...
+                    .addOutputState(DummyState(), DummyContract.PROGRAM_ID)
+                    .addCommand(dummyCommand(ourIdentity.owningKey))
+                    .addAttachment(hash1)
+                    .addAttachment(hash2)
+                    .addAttachment(hash3)
+                    .addAttachment(hash4)
+            val stx = serviceHub.signInitialTransaction(tx, ourIdentity.owningKey)
+            if (verify) stx.verify(serviceHub, checkSufficientSignatures = false)
+            // Send to the other side and wait for it to trigger resolution from us.
+            val bob = serviceHub.identityService.wellKnownPartyFromX500Name(otherSide.name)!!
```
shamsasari: I know you didn't introduce, but can you remove this. You can use `otherSide` directly.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -124,6 +126,10 @@ open class MockServices private constructor(
             // compiler and then the c'tor itself.
             return Throwable().stackTrace[3].className.split('.').dropLast(1).joinToString(".")
         }
+
+        init {
+            GlobalProperties.networkParameters = testNetworkParameters(emptyList())
```
shamsasari: Actually sorry I didn't think this properly. It be better that this be done in the `init` (as you had) and *also* unset in the `stopNodes` method.

We also have to do something similar for the `DriverDSLImpl` for in-process nodes.

However this is showing a design smell. The driver creates its own parameters, but only to automatically define the notaries. I think we _do_ need the ability to specific max transaction size for the driver.

patrickkuo:
all the networkParameters ref in AbstractNode are from the global variable, its a static import

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -13,13 +13,19 @@ import net.corda.core.flows.*
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.identity.PartyAndCertificate
-import net.corda.core.internal.*
+import net.corda.core.internal.FlowStateMachine
+import net.corda.core.internal.GlobalProperties.networkParameters
```
shamsasari: Ah you've made it a static import! The code is `networkParameters = NetworkParametersReader...` making `networkParameters` look like an instance field. Can you remove this static import.

PR #2466, opened 2018-02-05
CORDA-973 Refactoring for serialization compression support

andr3ej:
21b5355 don't make so many empty byte arrays
a9f4717 less byte array copying (low hanging fruit only)
3525fe0 fix trailing garbage, this might be the only broken serializer
5e31637 kryo streams OO
847e8be serialization magic OO but without breaking API


core/build.gradle
```diff
@@ -104,7 +104,7 @@ dependencies {
     // Apache JEXL: An embeddable expression evaluation library.
     // This may be temporary until we experiment with other ways to do on-the-fly contract specialisation via an API.
     compile "org.apache.commons:commons-jexl3:3.0"
-
+    compile 'commons-lang:commons-lang:2.6'
```
andr3ej: migrated from explorer, this isn't the first time i've wanted it available in core

core/src/main/kotlin/net/corda/core/utilities/ByteArrays.kt
```diff
@@ -4,44 +4,25 @@ package net.corda.core.utilities
...
  * The data of interest typically starts at position [offset] within the [bytes] and is [size] bytes long.
+ *
+ * @property offset The start position of the sequence within the byte array.
+ * @property size The number of bytes this sequence represents.
  */
 @CordaSerializable
-sealed class ByteSequence : Comparable<ByteSequence> {
-    constructor() {
-        this._bytes = COPY_BYTES
-    }
```
andr3ej: unused

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerializationScheme.kt
```diff
@@ -73,65 +71,42 @@ abstract class AbstractKryoSerializationScheme : SerializationScheme {
...
-            return pool.run { kryo ->
-                withContext(kryo, context) {
-                    if (context.objectReferencesEnabled) {
-                        uncheckedCast(kryo.readClassAndObject(input))
-                    } else {
-                        kryo.withoutReferences { uncheckedCast<Any?, T>(kryo.readClassAndObject(input)) }
-                    }
+        byteSequence.take(headerSize) == KryoHeaderV0_1 || throw KryoException("Serialized bytes header does not match expected format.")
+        return context.kryo {
+            kryoInput(byteSequence.skip(headerSize).open()) {
```
andr3ej: avoid duplicating a possibly large byte array

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -225,6 +223,7 @@ fun <T : Any> T.serialize(serializationFactory: SerializationFactory = Serializa
  * A type safe wrapper around a byte array that contains a serialised object. You can call [SerializedBytes.deserialize]
  * to get the original object back.
  */
+@Suppress("unused")
 class SerializedBytes<T : Any>(bytes: ByteArray) : OpaqueBytes(bytes) {
```
andr3ej: T is used but intellij thinks it isn't

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/SerializationScheme.kt
```diff
@@ -76,34 +70,32 @@ data class SerializationContextImpl(override val preferredSerializationVersion:
...
-        val lookupKey = byteSequence.obtainHeaderSignature() to target
-        val scheme = schemes.computeIfAbsent(lookupKey) {
-            registeredSchemes
-                    .filter { scheme -> scheme.canDeserializeVersion(it.first, it.second) }
-                    .forEach { return@computeIfAbsent it }
+    private val schemes: ConcurrentHashMap<Pair<CordaSerializationMagic, SerializationContext.UseCase>, SerializationScheme> = ConcurrentHashMap()
+
+    private fun schemeFor(byteSequence: ByteSequence, target: SerializationContext.UseCase): Pair<SerializationScheme, CordaSerializationMagic> {
+        // truncate sequence to at most magicSize, and make sure it's a copy to avoid holding onto large ByteArrays
+        val magic = CordaSerializationMagic(byteSequence.take(min(byteSequence.size, magicSize)).copyBytes())
```
andr3ej: slight behaviour change: if byteSequence is shorter than magicSize we no longer get the ByteSequence.take exception

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/custom/OpaqueBytesSubSequenceSerializer.kt
```diff
@@ -13,8 +13,6 @@ import net.corda.nodeapi.internal.serialization.amqp.SerializerFactory
 class OpaqueBytesSubSequenceSerializer(factory: SerializerFactory) :
         CustomSerializer.Proxy<OpaqueBytesSubSequence, OpaqueBytes>(OpaqueBytesSubSequence::class.java, OpaqueBytes::class.java, factory) {
     override val additionalSerializers: Iterable<CustomSerializer<out Any>> = emptyList()
-
-    override fun toProxy(obj: OpaqueBytesSubSequence): OpaqueBytes = OpaqueBytes(obj.copy().bytes)
-
+    override fun toProxy(obj: OpaqueBytesSubSequence): OpaqueBytes = OpaqueBytes(obj.copyBytes())
```
andr3ej: no one should have to think about whether `bytes` is the whole thing or a bigger array

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/Kryo.kt
```diff
@@ -186,7 +186,7 @@ object InputStreamSerializer : Serializer<InputStream>() {
                 output.writeInt(numberOfBytesRead, true)
                 output.writeBytes(buffer, 0, numberOfBytesRead)
             } else {
-                output.writeInt(0)
+                output.writeInt(0, true)
```
andr3ej: * without the boolean it writes a 4-byte int, with the boolean it writes a varint
* compression support will allow for trailing garbage in existing data

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/KryoTests.kt
```diff
@@ -36,16 +36,13 @@ class KryoTests {
         private val ALICE_PUBKEY = TestIdentity(ALICE_NAME, 70).publicKey
     }
 
-    @Rule
-    @JvmField
-    val testSerialization = SerializationEnvironmentRule()
```
andr3ej: not needed as the test doesn't rely on implicit factory/context anywhere

node/src/main/kotlin/net/corda/node/services/persistence/DBCheckpointStorage.kt
```diff
@@ -24,13 +25,13 @@ class DBCheckpointStorage : CheckpointStorage {
...
             @Column(name = "checkpoint_value")
-            var checkpoint: ByteArray = ByteArray(0)
+            var checkpoint: ByteArray = EMPTY_BYTE_ARRAY
     )
 
     override fun addCheckpoint(checkpoint: Checkpoint) {
         currentDBSession().save(DBCheckpoint().apply {
             checkpointId = checkpoint.id.toString()
-            this.checkpoint = checkpoint.serializedFiber.bytes
+            this.checkpoint = checkpoint.serializedFiber.bytes // XXX: Is copying the byte array necessary?
```
andr3ej: i don't know enough about hibernate, and implementing this without copy while keeping the guarantees of OpaqueBytes without making it hibernate-aware is definitely out of scope

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoStreams.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+internal fun Output.substitute(transform: (OutputStream) -> OutputStream) {
+    flush()
+    outputStream = transform(outputStream)
+}
+
+internal fun Input.substitute(transform: (InputStream) -> InputStream) {
+    inputStream = transform(SequenceInputStream(ByteArrayInputStream(buffer.copyOfRange(position(), limit())), inputStream))
+}
```
andr3ej: these methods will be used for compression support

core/src/main/kotlin/net/corda/core/utilities/ByteArrays.kt
rick-r3: My original thinking was that this supported anyone who extended `ByteSequence` previously.... but they can't right?  Kotlin makes the constructor private due to the `sealed`-ness?

core/src/main/kotlin/net/corda/core/utilities/ByteArrays.kt
```diff
@@ -69,23 +51,33 @@ sealed class ByteSequence : Comparable<ByteSequence> {
...
     }
 
     /**
      * Take the first n bytes of this sequence as a sub-sequence.  See [subSequence] for further semantics.
      */
-    fun take(n: Int): ByteSequence {
-        require(size >= n)
-        return subSequence(0, n)
-    }
+    fun take(n: Int) = subSequence(0, n)
```
rick-r3: Can we put explicit returns types on the methods in here?  Makes it easier to read.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerializationScheme.kt
```diff
@@ -73,65 +70,41 @@ abstract class AbstractKryoSerializationScheme : SerializationScheme {
...
-        }
-        Input(byteSequence.bytes, byteSequence.offset + headerSize, byteSequence.size - headerSize).use { input ->
-            return pool.run { kryo ->
-                withContext(kryo, context) {
-                    if (context.objectReferencesEnabled) {
-                        uncheckedCast(kryo.readClassAndObject(input))
-                    } else {
-                        kryo.withoutReferences { uncheckedCast<Any?, T>(kryo.readClassAndObject(input)) }
-                    }
+        val dataBytes = kryoMagic.consume(byteSequence) ?: throw KryoException("Serialized bytes header does not match expected format.")
```
rick-r3: Is there a potential unnecessary copy here?  Consume on `SerializedBytes` makes a copy when it calls the underlying `subSequence`.  If we called `open()` and read and discarded 8 bytes, we wouldn't copy.

Maybe we want a couple of different `consume` implementations that return `InputStream` (for Kryo) or `ByteBuffer` (for AMQP) with the first 8 bytes skipped?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoStreams.kt
```diff
@@ -0,0 +1,43 @@
...
+import java.io.*
+
+private val serializationBufferPool = LazyPool(
+        newInstance = { ByteArray(64 * 1024) })
+private val serializeOutputStreamPool = LazyPool(
+        clear = ByteArrayOutputStream::reset,
+        shouldReturnToPool = { it.size() < 256 * 1024 }, // Discard if it grew too large
+        newInstance = { ByteArrayOutputStream(64 * 1024) })
+
+internal fun <T> kryoInput(underlying: InputStream, task: Input.() -> T): T {
```
rick-r3: Nice.  We need to try this stuff in the perf cluster to see if any contention shows up due to pooling.  The other alternative is maybe a thread local style.  We'll see.  There are some other situations we need to pool or thread things that I've discovered and hacked around for testing purposes but need proper solutions: e.g. we keep creating various `SecureRandom` and `MessageDigest` etc things over and over again.  There's another JIRA for that though...

node/src/main/kotlin/net/corda/node/services/persistence/DBCheckpointStorage.kt
rick-r3: Me neither.  It probably takes another copy!  Anyway, yeah, another JIRA another day.

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -98,17 +98,15 @@ abstract class SerializationFactory {
         val currentFactory: SerializationFactory? get() = _currentFactory.get()
     }
 }
-
-typealias VersionHeader = ByteSequence
-
+typealias SerializationMagic = ByteSequence
```
rick-r3: Can we maybe call this `SerializationHeader` ? I'd also like to keep some element of versioning in the AMQP and Kryo header constant names (and byte sequence, but that hasn't changed), in case we have another version of either somewhere along the line.

core/src/main/kotlin/net/corda/core/utilities/ByteArrays.kt
andr3ej: indeed they can't - i just did a little test and the only thing you can extend from that file is OpaqueBytes

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
andr3ej: the plan is to shorten the header to 7 bytes, and then the next byte is an instruction, see #2473. current values for that byte are 0 and 1, which will both mean "read serialization data and stop". or if the byte is 2 then following it is a declared encoding that should be applied for reading the rest of the stream. i think this mini-bytecode will be quite futureproof, so the headers are unlikely to change, which is why i've renamed them magic.

core/src/main/kotlin/net/corda/core/utilities/ByteArrays.kt
andr3ej: Done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerializationScheme.kt
andr3ej: oops, yes there is. maybe i believed the kdoc on subSequence that claimed there isn't a copy, which i've now fixed. i've changed consume to return ByteBuffer, and take advantage of its equality impl.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoStreams.kt
andr3ej: also it's going via a stream now when previously the Input simply wrapped a byte array, so yeah this may need to be revisited for performance

andr3ej:
back to you @rick-r3

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
rick-r3: Aha!  Sounds like a plan.

rick-r3: What if we had a 3rd scheme?  Should it be 5 bytes and then interpret a byte at a time? 0-> Kryo (with 2 byte ignored prefix), 1-> AMQP (with 2 byte ignored prefix).  I should go look at the other PR....

PR #2467, opened 2018-02-05
CORDA-986 and CORDA-985 CompositeKey and Signature verification performance fixes

kchalkias:
- Fix double sig verification issue in  `toLedgerTransaction` and `verifyRegularTransaction`. The problem was that `checkSignaturesAreValid` was called twice.
- `CompositeKey.checkFulfilledBy` returns right after the threshold requirement is met. Previously, it required a full scan of signing keys.
- checkValidity() "knows" if it ran before; not required any more to externally check `if (!validated)`
- we now first check for missing signers, then validate signatures.
- fixed missing `checkSignaturesAreValid` from `verifyNotaryChangeTransaction
- some comments and `method signature consistency fixes (private methods)

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
@@ -11,7 +11,7 @@ import java.util.*
 
 /**
  * A tree data structure that enables the representation of composite public keys, which are used to represent
- * the signing requirements for multisignature scenarios such as RAFT notary services. A composite key is a list
+ * the signing requirements for multi-signature scenarios such as RAFT notary services. A composite key is a list
  * of leaf keys and their contributing weight, and each leaf can be a conventional single key or a composite key.
  * Keys contribute their weight to the total if they are matched by the signature.
  *
@@ -52,9 +52,19 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
             }
             return builder.build(threshold)
         }
+        // Required for sorting [children] list.
+        // TODO: node.encoded.sequence() might be expensive, consider a faster deterministic compareTo implementation
+        //      for public keys in general.
+        private val descWeightComparator = compareBy<NodeAndWeight>({ -it.weight }, { it.node.encoded.sequence() })
     }
 
-    val children: List<NodeAndWeight> = children.sorted()
+    /**
+     * To ensure a deterministic way of adding children required for equality checking, [children] list is sorted
+     * during construction. A DESC ordering in the [NodeAndWeight.weight] field will improve efficiency, because keys
+     * with bigger "weights" are the first to be checked and thus the threshold requirement might be met earlier without
+     * requiring a full [children] scan.
+     */
+    val children: List<NodeAndWeight> = children.sortedWith(descWeightComparator)
 
     init {
         // TODO: replace with the more extensive, but slower, checkValidity() test.
@@ -142,9 +152,10 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
 
         override fun compareTo(other: NodeAndWeight): Int {
             return if (weight == other.weight)
+                // TODO: this might be expensive, consider a faster deterministic compareTo implementation when weights are equal.
                 node.encoded.sequence().compareTo(other.node.encoded.sequence())
             else
-                -weight.compareTo(other.weight) // Descending order.
+                weight.compareTo(other.weight)
         }
 
         override fun toASN1Primitive(): ASN1Primitive {
--- a/core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
@@ -1,6 +1,5 @@
 package net.corda.core.crypto
 
-import net.corda.core.crypto.CompositeKey.NodeAndWeight
 import net.corda.core.serialization.CordaSerializable
 import net.corda.core.utilities.exactAdd
 import net.corda.core.utilities.sequence
@@ -106,6 +105,7 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
      * TODO: Always call this method when deserialising [CompositeKey]s.
      */
     fun checkValidity() {
+        if (validated) return
         val visitedMap = IdentityHashMap<CompositeKey, Boolean>()
         visitedMap.put(this, true)
         cycleDetection(visitedMap) // Graph cycle testing on the root node.
@@ -180,17 +180,20 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
 
     override fun getFormat() = ASN1Encoding.DER
 
-    // Extracted method from isFulfilledBy.
+    // Return true when and if the threshold requirement is met.
     private fun checkFulfilledBy(keysToCheck: Iterable<PublicKey>): Boolean {
         if (keysToCheck.any { it is CompositeKey }) return false
-        val totalWeight = children.map { (node, weight) ->
+
+        var totalWeight = 0
+        children.forEach { (node, weight) ->
             if (node is CompositeKey) {
-                if (node.checkFulfilledBy(keysToCheck)) weight else 0
+                if (node.checkFulfilledBy(keysToCheck)) totalWeight += weight
             } else {
-                if (keysToCheck.contains(node)) weight else 0
+                if (node in keysToCheck) totalWeight += weight
             }
-        }.sum()
-        return totalWeight >= threshold
+            if (totalWeight >= threshold) return true
+        }
+        return false
     }
 
     /**
@@ -201,8 +204,7 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
     fun isFulfilledBy(keysToCheck: Iterable<PublicKey>): Boolean {
         // We validate keys only when checking if they're matched, as this checks subkeys as a result.
         // Doing these checks at deserialization/construction time would result in duplicate checks.
-        if (!validated)
-            checkValidity() // TODO: remove when checkValidity() will be eventually invoked during/after deserialization.
+        checkValidity()
         return checkFulfilledBy(keysToCheck)
     }
 
--- a/core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
@@ -144,7 +144,7 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
             return if (weight == other.weight)
                 node.encoded.sequence().compareTo(other.node.encoded.sequence())
             else
-                -weight.compareTo(other.weight) // Descending ordering.
+                -weight.compareTo(other.weight) // Descending order.
         }
 
         override fun toASN1Primitive(): ASN1Primitive {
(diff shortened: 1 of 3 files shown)


kchalkias:
@rick-r3 not sure how expensive sig verification is compared to the rest of Corda operations, but we used to verify all signatures twice.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
```diff
@@ -180,17 +180,20 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
 
     override fun getFormat() = ASN1Encoding.DER
 
-    // Extracted method from isFulfilledBy.
+    // Return true when and if the threshold requirement is met.
     private fun checkFulfilledBy(keysToCheck: Iterable<PublicKey>): Boolean {
         if (keysToCheck.any { it is CompositeKey }) return false
```
shamsasari: Move this into `isFulfilledBy` so that the recursive call below doesn't do this check again.

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
```diff
@@ -165,8 +168,11 @@ data class SignedTransaction(val txBits: SerializedBytes<CoreTransaction>,
      * objects from the TransactionState.
      */
     private fun verifyRegularTransaction(checkSufficientSignatures: Boolean, services: ServiceHub) {
-        checkSignaturesAreValid()
-        if (checkSufficientSignatures) verifyRequiredSignatures()
+        if (checkSufficientSignatures) {
```
shamsasari: You can call `toLedgerTransaction` to remove the code duplication

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
```diff
@@ -106,6 +105,7 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
      * TODO: Always call this method when deserialising [CompositeKey]s.
```
shamsasari: Why don't we call `checkValidity` in `init` block? It'll save us needing the `validated` boolean as well.

kchalkias: This is debatable, that's why we have the TODO comment in `init`. As checkValidity() is more expensive, the party that owns the Composite key or someone who is just a mediator (eg forwards keys), why should they always run checkValidity() if they don't need to? I am 50-50 on this, especially if we want to improve performance.
Another example: assume you received a transaction with 10 CompositeKeys, but the threshold was satisfied after the 3rd key. You don't even need to check the signatures of the remaining keys.

shamsasari: Yeah I realised afterwards it's not immediately clear you want to always run this on construction. 

However the TODO needs to be put into its own comment as it's showing up in the Kdocs.

kchalkias: Removed this TODO, as it is mentioned in `init` already. Back to you.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
```diff
@@ -145,7 +144,7 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
             return if (weight == other.weight)
                 node.encoded.sequence().compareTo(other.node.encoded.sequence())
             else
-                weight.compareTo(other.weight)
+                -weight.compareTo(other.weight) // Descending order.
```
shamsasari: This is changing the semantics of the ordering. Was it a bug previously? There's no tests for this.

kchalkias: This one is interesting. Sorting is (and was) used to pass the equality check (and only that). The purpose was to have a deterministic way of adding elements to the list. However, I realised that a DESC sorting could improve efficiency (along with changes in `checkFulfilledBy`), because bigger "weights" are the first to be checked and thus we might surpass the threshold earlier without requiring to scan the full list.

shamsasari: The problem is that `NodeAndWeight` is now public API and this is reversing the default sort order of this class. I think the solution to this problem is to write our own private `Comparator` which is optimised for `checkFulfilledBy` and use that to insert nodes into the list.

kchalkias: Correct, this makes sense.

shamsasari: We should be able to get away with this because there's no docs for `children` and thus haven't committed to any particular ordering of the nodes.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
```diff
@@ -53,9 +52,19 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
...
+        // Required for sorting [children] list.
+        // TODO: node.encoded.sequence() might be expensive, consider a faster deterministic compareTo implementation
+        //      for public keys in general.
+        private val descWeightComparator = compareBy<NodeAndWeight>({ -it.weight }, { it.node.encoded.sequence() })
     }
 
-    val children: List<NodeAndWeight> = children.sorted()
+    /**
+     * To ensure a deterministic way of adding children required for equality checking, [children] list is sorted
+     * during construction. A DESC ordering in the [NodeAndWeight.weight] field will improve efficiency, because keys
```
shamsasari: I wouldn't document the ordering as it means we're committing to this behaviour and it'll prove problematic to change later. What we should say instead is that the order of the children may not be the same order to when they were provided in the c'tor.

kchalkias: Agreed, moved this comment to the `private val descWeightComparator` description.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
```diff
@@ -52,17 +52,17 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
...
         //      for public keys in general.
         private val descWeightComparator = compareBy<NodeAndWeight>({ -it.weight }, { it.node.encoded.sequence() })
     }
 
     /**
-     * To ensure a deterministic way of adding children required for equality checking, [children] list is sorted
-     * during construction. A DESC ordering in the [NodeAndWeight.weight] field will improve efficiency, because keys
-     * with bigger "weights" are the first to be checked and thus the threshold requirement might be met earlier without
-     * requiring a full [children] scan.
+     * Τhe order of the children may not be the same to what was provided in the constructor.
```
shamsasari: Actually it's better that this be a `@property` in the class KDoc above because otherwise it won't be seen in the c'tor doc.

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
```diff
@@ -37,7 +37,7 @@ import java.util.function.Predicate
 data class SignedTransaction(val txBits: SerializedBytes<CoreTransaction>,
                              override val sigs: List<TransactionSignature>
 ) : TransactionWithSignatures {
-    // DOCEND 1
+    // DOCEND 1.
```
shamsasari: Actually you'll need to revert this. This breaks the reference in api-transaction.rst

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
shamsasari: Actually the c'tor is private so ignore me. Instead of "constructor" you  need to refer to the builder?

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
kchalkias: Fixed

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
```diff
@@ -53,9 +52,19 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
             }
             return builder.build(threshold)
         }
+        // Required for sorting [children] list. To ensure a deterministic way of adding children required for equality
+        // checking, [children] list is sorted during construction. A DESC ordering in the [NodeAndWeight.weight] field
+        // will improve efficiency, because keys with bigger "weights" are the first to be checked and thus the
+        // threshold requirement might be met earlier without requiring a full [children] scan.
+        // TODO: node.encoded.sequence() might be expensive, consider a faster deterministic compareTo implementation
+        //      for public keys in general.
+        private val descWeightComparator = compareBy<NodeAndWeight>({ -it.weight }, { it.node.encoded.sequence() })
```
adagys: doesn't `NodeAndWeight` already implement `Comparable`?

kchalkias: true that, but it's better for `children` to have DESC ordering on weight to ensure higher weighted nodes are checked first during check fulfilment (so we can skip a full scan earlier).

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
```diff
@@ -201,8 +212,8 @@ class CompositeKey private constructor(val threshold: Int, children: List<NodeAn
     fun isFulfilledBy(keysToCheck: Iterable<PublicKey>): Boolean {
         // We validate keys only when checking if they're matched, as this checks subkeys as a result.
         // Doing these checks at deserialization/construction time would result in duplicate checks.
-        if (!validated)
-            checkValidity() // TODO: remove when checkValidity() will be eventually invoked during/after deserialization.
+        checkValidity()
+        if (keysToCheck.any { it is CompositeKey }) return false
```
adagys: maybe better to throw an exception instead of false here

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
kchalkias: We have a related a discussion with @shamsasari, we've agreed on non removing the natural ordering of the initial implementation for consistency.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
shamsasari: That would be breaking existing behaviour.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
adagys: true, but it makes the code confusing, and I'd be _very_ surprised if the ordering change would have any actual impact at the moment.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
adagys: ok, maybe we should log a warning then?

shamsasari: If we find that returning false here was the wrong decision, and apps would benefit from it throwing an exception instead, then we should create a second method which is identical to this except that it throws. But I don't see any benefit of doing that now, not until we have developer feedback.

core/src/main/kotlin/net/corda/core/crypto/CompositeKey.kt
kchalkias: So, @shamsasari wdyt?

shamsasari: I think the code is fine, but there is a valid point on whether there's any actual impact on existing use cases. I guess I assumed there was. However no point in reverting though as it is (at least theoretically) more performant.

PR #2469, opened 2018-02-05
CORDA-556: Added Cordapp Config and a sample

Clintonio:


Clintonio:
Adding @mikehearn to approve API changes.

roastario:
rebase attempt to fix failing due to RPC settings issue

.ci/api-current.txt
```diff
@@ -635,7 +635,7 @@ public static final class net.corda.core.contracts.UniqueIdentifier$Companion ex
   @org.jetbrains.annotations.NotNull public abstract List getServices()
 ##
 public final class net.corda.core.cordapp.CordappContext extends java.lang.Object
-  public <init>(net.corda.core.cordapp.Cordapp, net.corda.core.crypto.SecureHash, ClassLoader)
+  public <init>(net.corda.core.cordapp.Cordapp, net.corda.core.crypto.SecureHash, ClassLoader, net.corda.core.cordapp.CordappConfig)
```
mikehearn: We don't expect anyone to construct this themselves, do we? Can we mark the c'tor as internal? I guess that'd require a jump method as it's defined in core but used from node.

.ci/api-current.txt
```diff
@@ -4667,7 +4667,7 @@ public static final class net.corda.testing.services.MockAttachmentStorage$openA
   @org.jetbrains.annotations.NotNull public net.corda.core.crypto.SecureHash getId()
 ##
 public final class net.corda.testing.services.MockCordappProvider extends net.corda.node.internal.cordapp.CordappProviderImpl
-  public <init>(net.corda.node.internal.cordapp.CordappLoader, net.corda.core.node.services.AttachmentStorage)
+  public <init>(net.corda.node.internal.cordapp.CordappLoader, net.corda.core.node.services.AttachmentStorage, net.corda.testing.node.MockCordappConfigProvider)
```
mikehearn: Likewise - do we expect anyone to ever construct this themselves? If not can it be made internal?

Clintonio: This one may be constructed by users writing more advanced unit tests.

Clintonio: Though I wouldn't expect it, but then a few other mock classes would need to be moved into internal (MockAttachmentStorage, MockCordappConfigProvider)

.ci/api-current.txt
Clintonio: This will officially be changing the API if we make the constructor internal, but yes, we can make it internal. It was never intended to be modified. Just checking it compiles and if it does I'll commit.

Clintonio: Just checked - I'd have to extract it out as an interface and move the implementation into the same package as CordappProviderImpl.

mikehearn: Mmmm. Why can't you make the c'tor internal and then provide a public function in an internal package that proxies to it?

.ci/api-current.txt
mikehearn: OK - can we provide an overload for backwards compatibility then?

PR #2472, opened 2018-02-06
CORDA-939 - Don't expose StartedNode and AbstractNode as part of public test api

anthonykr3:
Renamed MockNetwork as InternalMockNetwork and moved to the Internal namespace. 

Created a new MockNetwork and MockNode which don't expose any internal API's

There are still some tests internally which use the InternalMockNetwork directly - that is because they inject custom AbstractNode implementations. Since AbstractNode shouldn't be implemented outside of Corda this seems like the best course of action.


.ci/api-current.txt
```diff
@@ -3667,24 +3667,25 @@ public final class net.corda.testing.driver.JmxPolicy extends java.lang.Object
   public abstract void stop()
```
anthonykr3: These changes are all to the net.cord.testing.* package, so even though it requires changes to api-current.txt, it isn't changing the API (yet) as they weren't stable in V2, and these changes will be going in V3.

anthonykr3:
@shamsasari I think you're the best person to review this?

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -91,6 +91,14 @@ sealed class NodeHandle : AutoCloseable {
             }
             onStopCallback()
         }
+        private lateinit var node: StartedNode<Node>
```
shamsasari: Can you move this above `stop`

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -91,6 +91,14 @@ sealed class NodeHandle : AutoCloseable {
...
+        val database get() = node.database
```
shamsasari: Specify the return types

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -91,6 +91,14 @@ sealed class NodeHandle : AutoCloseable {
...
+        val services get() = node.services
+        fun <T : FlowLogic<*>> registerInitiatedFlow(initiatedFlowClass: Class<T>) = node.registerInitiatedFlow(initiatedFlowClass)
```
shamsasari: For this as well (it returns a Future, right?)

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -91,6 +91,14 @@ sealed class NodeHandle : AutoCloseable {
...
+        internal fun initialise(node: StartedNode<Node>) : InProcess {
```
shamsasari: Can we instead have `InProcess` be an interface? Then we can have our internal `InProcessImpl` and pass whatever we want into it via its c'tor.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,184 @@
...
+
+/** Represents a node configuration for injection via [MockNetworkParameters] **/
+data class MockNetworkNotarySpec(val name: CordaX500Name, val validating: Boolean = true) {
+    constructor(name: CordaX500Name) : this(name, validating = true)
+}
+
+/** A class that represents an unstarted mock node for testing. Do not instantiate directly, create via a [MockNetwork] **/
+class UnstartedMockNode {
+    private lateinit var node: InternalMockNetwork.MockNode
+    val id get() = node.id
```
shamsasari: Return type declaration for public members

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,184 @@
...
+    fun setInitialiseSerialization(initialiseSerialization: Boolean) = copy(initialiseSerialization = initialiseSerialization)
+    fun setNotarySpecs(notarySpecs: List<MockNetworkNotarySpec>) = copy(notarySpecs = notarySpecs)
+}
+
+/** Represents a node configuration for injection via [MockNetworkParameters] **/
+data class MockNetworkNotarySpec(val name: CordaX500Name, val validating: Boolean = true) {
+    constructor(name: CordaX500Name) : this(name, validating = true)
+}
+
+/** A class that represents an unstarted mock node for testing. Do not instantiate directly, create via a [MockNetwork] **/
```
shamsasari: We users are not allowed to initiate directly then perhaps this needs a private c'tor and then we add an `internal` factory method?

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,184 @@
...
+     * Intialise via internal function rather than internal constructors, as internal constructors in Kotlin
+     * are visible via the Api
+     */
+    internal fun initialise(node: InternalMockNetwork.MockNode): UnstartedMockNode {
+        this.node = node
+        return this
+    }
+}
+
+/** A class that represents a started mock node for testing. Do not instantiate directly, create via a [MockNetwork] **/
```
shamsasari: ditto

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,184 @@
...
+class StartedMockNode {
+    private lateinit var node: StartedNode<InternalMockNetwork.MockNode>
+    val services get() = node.services
+    val database get() = node.database
+    val id get() = node.internals.id
+    val configuration get() = node.internals.configuration
+    val smm get() = node.smm
```
shamsasari: The StateMachineManager is internal, we're still exposing it?

testing/node-driver/src/test/kotlin/net/corda/testing/node/InternalMockNetworkTests.kt
```diff
@@ -1,15 +1,16 @@
 package net.corda.testing.node
 
 import net.corda.core.serialization.internal.effectiveSerializationEnv
+import net.corda.testing.node.internal.InternalMockNetwork
 import org.assertj.core.api.Assertions.assertThatThrownBy
 import org.junit.Test
 
-class MockNetworkTests {
+class InternalMockNetworkTests {
```
shamsasari: Can we keep the existing test class

testing/node-driver/src/integration-test/kotlin/net/corda/testing/node/internal/InternalMockNetworkIntegrationTests.kt
```diff
@@ -1,16 +1,16 @@
...
+package net.corda.testing.node.internal
 
 import net.corda.core.internal.div
 import net.corda.testing.common.internal.ProjectStructure.projectRootDir
 import net.corda.testing.node.internal.ProcessUtilities.startJavaProcess
 import org.junit.Test
 import kotlin.test.assertEquals
 
-class MockNetworkIntegrationTests {
+class InternalMockNetworkIntegrationTests {
```
shamsasari: Keep the existing test class

anthonykr3:
@shamsasari Hopefully that should better

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -33,73 +37,90 @@ import java.util.concurrent.atomic.AtomicInteger
...
      * Interface to the node's RPC system. The first RPC user will be used to login if are any, otherwise a default one
      * will be added and that will be used.
      */
-    abstract val rpc: CordaRPCOps
-    abstract val configuration: NodeConfiguration
-    abstract val webAddress: NetworkHostAndPort
-    abstract val useHTTPS: Boolean
-
+    val rpc: CordaRPCOps
+    val configuration: NodeConfiguration
```
shamsasari: What are we doing about this as this class is in `node`? I think we're going to have to move this into the implementation classes.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -33,73 +37,90 @@ import java.util.concurrent.atomic.AtomicInteger
...
+    val webAddress: NetworkHostAndPort
+    val useHTTPS: Boolean
```
shamsasari: Can we move this into the implementation classes and keep it internal? This is only used by the web server (which is sort of deprecated?) and wouldn't want it being exposed publically. I think it's just needed by the `startWebServer` method. Same applies to `webAddress`.

You might need to introduce an `InternalNodeHandle` interface.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -33,73 +37,90 @@ import java.util.concurrent.atomic.AtomicInteger
...
+    fun stop()
 
     /**
-     * Closes and stops the node.
+     * Connects to node through RPC.
+     *
+     * @param sslConfiguration specifies SSL options.
      */
-    override fun close() = stop()
+    fun rpcClientToNode(sslConfiguration: SSLConfiguration? = null): CordaRPCClient
```
shamsasari: `SSLConfiguration` is internal API. If the parameter is needed in our tests can we have an internal version of this that exposes it.

In fact, can we actually just get rid of this completely. We already have `rpc` for app devs to use, and if they really want to make their own connection then it's easy peasy to do (it's just creating an `CordaRPCClient` object).

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -33,73 +37,90 @@ import java.util.concurrent.atomic.AtomicInteger
...
+interface InProcess : NodeHandle {
+    val database: CordaPersistence
+    val services: StartedNodeServices
+    /**
+     * Register a flow that is initiated by another flow
+     */
+    fun <T : FlowLogic<*>> registerInitiatedFlow(initiatedFlowClass: Class<T>): Observable<T>
+}
+
+internal data class OutOfProcessImpl(
```
shamsasari: Move these instead into the internal package.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
shamsasari: The later half of the doc is no longer applicable.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    val id get() = node.id
+    val configuration get() = node.configuration
+    /** Start the node **/
+    fun start() = StartedMockNode.create(node.start())
+}
+
+/** A class that represents a started mock node for testing. Do not instantiate directly, create via a [MockNetwork] **/
```
shamsasari: likewise

testing/node-driver/src/test/kotlin/net/corda/testing/node/InternalMockNetworkTests.kt
shamsasari: You missed this?

docs/source/changelog.rst
```diff
@@ -178,6 +178,10 @@ UNRELEASED
 * Marked ``stateMachine`` on ``FlowLogic`` as ``CordaInternal`` to make clear that is it not part of the public api and is
   only for internal use
 
+* Created a new ``MockNetwork`` for creating mock networks for testing, and ``StartedMockNode`` and ``UnstartedMockNode``
```
shamsasari: From the point of view of app developers what has changed, just StartedMockNode and UnstartedMockNode, right? `MockNetwork` has stayed the same, no? Describing it as "new" implies a complete rewrite.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,174 @@
...
+    companion object {
+        internal fun create(node: InternalMockNetwork.MockNode): UnstartedMockNode {
+            return UnstartedMockNode(node)
+        }
+    }
+
+    val id get() = node.id
+    val configuration get() = node.configuration
+    /** Start the node **/
+    fun start() = StartedMockNode.create(node.start())
```
shamsasari: Specify the return type

testing/node-driver/src/test/kotlin/net/corda/testing/node/InternalMockNetworkTests.kt
anthonykr3: I made a public version of the MockNetworkIntegrationTests instead of the unit tests - I misread the file title - however on looking at this I'm not sure it really makes sense to make a public version of these tests as there's only one test, and it tests the initialisation, I'd have to make the public class open and add the method that this test overrides which would then just pass it through to the underlying class. Or do you just want the test renaming? I've moved into the internal package for now.

samples/irs-demo/src/integration-test/kotlin/net/corda/irs/IRSDemoTest.kt
```diff
@@ -78,7 +79,7 @@ class IRSDemoTest {
                 registerIRSModule(mapper)
                 HttpApi.fromHostAndPort(it.second, "api/irs", mapper = mapper)
             }
-            val nextFixingDates = getFixingDateObservable(nodeA.configuration)
+            val nextFixingDates = getFixingDateObservable(nodeA.nodeInfo.addresses.first())
```
shamsasari: `addresses` is the P2P addresses for receiving connections from other nodes, it's not RPC.

samples/irs-demo/src/integration-test/kotlin/net/corda/test/spring/SpringDriver.kt
```diff
@@ -96,23 +96,24 @@ data class SpringBootDriverDSL(private val driverDSL: DriverDSLImpl) : InternalD
 
     private fun startApplication(handle: NodeHandle, debugPort: Int?, clazz: Class<*>): Process {
         val className = clazz.canonicalName
+        handle.rpc
```
shamsasari: ?

testing/node-driver/src/main/kotlin/net/corda/testing/driver/internal/DriverInternal.kt
```diff
@@ -0,0 +1,75 @@
...
+import net.corda.testing.driver.OutOfProcess
+import net.corda.testing.node.User
+import rx.Observable
+import java.nio.file.Path
+
+interface NodeHandleInternal : NodeHandle {
+    val configuration: NodeConfiguration
+}
+
+internal data class OutOfProcessImpl(
```
shamsasari: Doesn't need to be `internal` now.

samples/irs-demo/src/integration-test/kotlin/net/corda/test/spring/SpringDriver.kt
anthonykr3: Don't know where that came from

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -33,73 +35,39 @@ import java.util.concurrent.atomic.AtomicInteger
...
      * will be added and that will be used.
      */
-    abstract val rpc: CordaRPCOps
-    abstract val configuration: NodeConfiguration
-    abstract val webAddress: NetworkHostAndPort
-    abstract val useHTTPS: Boolean
-
+    val rpc: CordaRPCOps
+    val p2pAddress: NetworkHostAndPort
+    val webAddress: NetworkHostAndPort
```
shamsasari: Can we move this and `useHTTPS` into `NodeHandleInternal` as well please. These are only used by the webserver and shouldn't be part of the public API.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/internal/DriverInternal.kt
```diff
@@ -0,0 +1,75 @@
...
+        override val webAddress: NetworkHostAndPort,
+        override val useHTTPS: Boolean,
+        val debugPort: Int?,
+        override val process: Process,
+        private val onStopCallback: () -> Unit
+) : OutOfProcess, NodeHandleInternal {
+    override val p2pAddress: NetworkHostAndPort get() = configuration.p2pAddress
+    override val rpcAddress: NetworkHostAndPort get() = configuration.rpcOptions.address!!
+    override val baseDirectory: Path get() = configuration.baseDirectory
+    override val rpcUsers: List<User> get() = configuration.rpcUsers.map { User(it.username, it.password, it.permissions) }
```
shamsasari: Perhaps remove the `get()` so that this doesn't create a new list each time.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/internal/DriverInternal.kt
```diff
@@ -0,0 +1,75 @@
...
+        override val configuration: NodeConfiguration,
+        override val webAddress: NetworkHostAndPort,
+        override val useHTTPS: Boolean,
+        private val nodeThread: Thread,
+        private val onStopCallback: () -> Unit,
+        private val node: StartedNode<Node>
+) : InProcess, NodeHandleInternal {
+    override val database: CordaPersistence get() = node.database
+    override val services: StartedNodeServices get() = node.services
+    override val p2pAddress: NetworkHostAndPort get() = configuration.p2pAddress
```
shamsasari: You can move all these `configuration` based properties into `NodeHandleInternal` to save on the duplicate code.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,174 @@
...
+/** A class that represents an unstarted mock node for testing. **/
+class UnstartedMockNode private constructor(private val node: InternalMockNetwork.MockNode) {
+    companion object {
+        internal fun create(node: InternalMockNetwork.MockNode): UnstartedMockNode {
+            return UnstartedMockNode(node)
+        }
+    }
+
+    val id get() = node.id
+    val configuration get() = node.configuration
```
shamsasari: This is exposing the internal `NodeConfiguration` class.

Can you also add the return type to all the public methods, especially the ones being exposed as public API. It makes it clear what we're returning, rather than inferring in our minds.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,174 @@
...
+    companion object {
+        internal fun create(node: StartedNode<InternalMockNetwork.MockNode>): StartedMockNode {
+            return StartedMockNode(node)
+        }
+    }
+
+    val services get() = node.services
+    val database get() = node.database
+    val id get() = node.internals.id
+    val configuration get() = node.internals.configuration
```
shamsasari: ditto

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -0,0 +1,174 @@
...
+     * Asks every node in order to process any queued up inbound messages. This may in turn result in nodes
+     * sending more messages to each other, thus, a typical usage is to call runNetwork with the [rounds]
+     * parameter set to -1 (the default) which simply runs as many rounds as necessary to result in network
+     * stability (no nodes sent any messages in the last round).
+     */
+    @JvmOverloads
+    fun runNetwork(rounds: Int = -1) = internalMockNetwork.runNetwork(rounds)
+
+    /** Get the base directory for the given node id. **/
+    fun baseDirectory(nodeId: Int) : Path = internalMockNetwork.baseDirectory(nodeId)
```
shamsasari: No whitespace before the colon

samples/irs-demo/src/integration-test/kotlin/net/corda/irs/IRSDemoTest.kt
```diff
@@ -78,7 +79,7 @@ class IRSDemoTest {
                 registerIRSModule(mapper)
                 HttpApi.fromHostAndPort(it.second, "api/irs", mapper = mapper)
             }
-            val nextFixingDates = getFixingDateObservable(nodeA.configuration)
+            val nextFixingDates = getFixingDateObservable(nodeA.p2pAddress)
```
shamsasari: It should be `rpcAddress`

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
anthonykr3: I've moved it, but it's used by the AttachmentDemo and the IRSDemo - I'm assuming this can't be sorted until we properly deprecate the web server?

samples/attachment-demo/src/integration-test/kotlin/net/corda/attachmentdemo/AttachmentDemoTest.kt
```diff
@@ -41,7 +42,7 @@ class AttachmentDemoTest {
 
             val recipientThread = supplyAsync {
                 CordaRPCClient(nodeB.rpcAddress).start(demoUser[0].username, demoUser[0].password).use {
-                    recipient(it.proxy, nodeB.webAddress.port)
+                    recipient(it.proxy, (nodeB as NodeHandleInternal).webAddress.port)
```
shamsasari: I just noticed that `startWebServer` above returns a `WebserverHandle` which also holds the address. Using that would be better than this cast.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: In both cases you can use `WebserverHandle.listenAddress`.

samples/attachment-demo/src/integration-test/kotlin/net/corda/attachmentdemo/AttachmentDemoTest.kt
anthonykr3: That works in the case of AttachmentDemoTest - but I don't think it works for the IRSDemo https://github.com/corda/corda/pull/2472/files/10965e9d9d7ef207b51de1ebb51953e10c583273#diff-bb844af5bb7043c4daf36931fc9e35f1R113

anthonykr3:
I think that's everything so far addressed now @shamsasari?

PR #2473, opened 2018-02-06
CORDA-973 Compression support for serialization

andr3ej:


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/SerializationFormat.kt
```diff
@@ -0,0 +1,54 @@
...
+enum class Instruction : Bytecode {
+    /** Serialization data follows, and then discard the rest of the stream (if any) as legacy data may have trailing garbage. */
+    DATA_AND_STOP,
+    /** Identical behaviour to [DATA_AND_STOP], historically used by kryo. */
+    ALT_DATA_AND_STOP,
+    /** The ordinal of a [CordaSerializationEncoding] follows, which should be used to decode the remainder of the stream. */
+    ENCODING;
+
+    companion object {
+        val vals = Vals(values())
```
rick-r3: I'm feeling like this (`Vals` and `Ord`) is less direct and harder to read than it could be.  

Can't we make it so the `Bytecode` write methods can somehow access the ordinal directly?  e.g. it can cast `this` to `Enum`.  Or can `Bytecode` have an `ordinal` method to avoid the cast (that is implemented on `Enum`)?

I can see that working for writing.  Reading is harder.  Might need a companion object in each that calls a helper method on Bytecode (passing in the enum class we're reading).

andr3ej: * Vals exists to cache the values array, which we don't want to create every time
* the ordinal trick works, but then there's no place to put the range validation. i could do it on write, but there are a few ways in which that's less elegant:
  * developers have to take care not to bypass validation in the write methods
  * the check happens every time when it only has to happen once
  * the check happens on first serialization when it could happen on enum clinit

rick-r3: Okay.  Maybe we can improve the names to improve readability then.  How about something like `OrdinalByte` for `Ord` and `BytecodeReader` for `Vals` ?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerializationScheme.kt
```diff
@@ -100,6 +112,12 @@ abstract class AbstractKryoSerializationScheme : SerializationScheme {
         return context.kryo {
             SerializedBytes(kryoOutput {
                 kryoMagic.writeTo(this)
+                context.encoding?.let { encoding ->
+                    Instruction.ENCODING.writeTo(this)
+                    (encoding as CordaSerializationEncoding).writeTo(this)
+                    substitute(encoding::wrap)
+                }
+                Instruction.ALT_DATA_AND_STOP.writeTo(this) // XXX: Change to DATA_AND_STOP?
```
rick-r3: Hmmm..... does this mean old code can't read this data? What's the compromise using the non-alt form? Is there a length check we can't enforce?  Can't spot it.

It's probably okay that old code cannot read this, since it's intended for storage but it would preclude a downgrade and I want to understand what the trade off is.

rick-r3: Okay, I'm remembering a previous version of the code.  Let's drop the XXX comment here and put an explanation on `ALT_DATA_AND_STOP` that this is here for historical "versioning" reasons, only used in Kryo and shouldn't be used in new code.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoStreams.kt
```diff
@@ -4,13 +4,30 @@ import com.esotericsoftware.kryo.io.Input
...
+}
 
 private val serializationBufferPool = LazyPool(
         newInstance = { ByteArray(64 * 1024) })
 private val serializeOutputStreamPool = LazyPool(
-        clear = ByteArrayOutputStream::reset,
+        clear = ByteBufferOutputStream::reset,
         shouldReturnToPool = { it.size() < 256 * 1024 }, // Discard if it grew too large
-        newInstance = { ByteArrayOutputStream(64 * 1024) })
+        newInstance = { ByteBufferOutputStream(64 * 1024) })
```
rick-r3: Wondering if we have / should have `.KB`, `.MB` extension methods on `Int` ?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/Bytecode.kt
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    init {
+        require(ordinal >= 0) { "The ordinal must be non-negative." }
+        require(ordinal < 128) { "Consider implementing a varint encoding." }
+    }
+}
+
+class Vals<out E>(private val values: Array<E>) {
+    fun readFrom(stream: InputStream) = values[stream.read()]
```
rick-r3: Should we add a clear message / exception handling here if we get an exception from trying to look outside the array bounds, that we read an unsupported bytecode?

Speaking of which, I wonder if the bytecode and instruction words will freak out anyone looking at this for security issues?  Maybe we can make bytecode "header byte" or something, and instruction "packet type" so they sound less executable.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
```diff
@@ -47,6 +52,23 @@ class DeserializationInput(internal val serializerFactory: SerializerFactory) {
             }
             return size + BYTES_NEEDED_TO_PEEK
         }
+
+        @Throws(NotSerializableException::class)
+        fun <T> getDataBytes(byteSequence: ByteSequence, task: (ByteBuffer) -> T): T {
```
rick-r3: `withDataBytes` rather than `getDataBytes` ? Latter makes it sound like it returns the bytes.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoStreams.kt
andr3ej: maybe, but it's not straightforward https://en.wikipedia.org/wiki/Mebibyte so descope?

rick-r3: Up to you

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/DeserializationInput.kt
andr3ej: Done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/Bytecode.kt
andr3ej: * exceptions added
* mentions of bytecode/instruction removed

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoSerializationScheme.kt
andr3ej: Done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/SerializationFormat.kt
andr3ej: Done - Septet and OrdinalReader

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/OrdinalIO.kt
```diff
@@ -0,0 +1,31 @@
+package net.corda.nodeapi.internal.serialization
+
+import java.io.EOFException
+import java.io.InputStream
+import java.io.OutputStream
+import java.nio.ByteBuffer
+
+class Septet(private val ordinal: Int) {
```
rick-r3: In still think something like `OrdinalByte` would be more meaningful.  I get that it's only 7 bits.  `OrdinalBits` maybe, which leaves room for a future extension to more than 7 bits as the upper bound check alludes to ?

`Septet` is err... not intuitive to me.

andr3ej: i like OrdinalBits, done

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoStreams.kt
```diff
@@ -4,13 +4,34 @@ import com.esotericsoftware.kryo.io.Input
...
+import java.nio.ByteBuffer
+
+class ByteBufferOutputStream(size: Int) : ByteArrayOutputStream(size) {
+    companion object {
+        private val ensureCapacity = ByteArrayOutputStream::class.java.getDeclaredMethod("ensureCapacity", Int::class.java).apply {
+            isAccessible = true
+        }
+    }
+
+    fun <T> asByteBuffer(remaining: Int, task: (ByteBuffer) -> T): T {
```
rick-r3: `asByteBuffer` doesn't feel quite right, since it doesn't return `ByteBuffer`.  We used `with...` with the `withDataBytes` function before.  `withByteBuffer` ?  Makes me wonder if we should rename `withDataBytes` to that too?

rick-r3: Not incredibly happy with `withByteBuffer` either.  `useAsByteBuffer` ?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPStreams.kt
```diff
@@ -0,0 +1,31 @@
...
+        byteBuffer // BBIS has no other state, so this is perfectly safe.
+    } else {
+        ByteBuffer.wrap(serializeOutputStreamPool.run {
+            copyTo(it)
+            it.toByteArray()
+        })
+    }
+}
+
+fun <T> OutputStream.asByteBuffer(remaining: Int, task: (ByteBuffer) -> T): T {
```
rick-r3: Maybe this needs an alternate naming too? See my similar comment elsewhere.

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -99,14 +100,22 @@ abstract class SerializationFactory {
...
+@DoNotImplement
 interface SerializationContext {
     /**
      * When serializing, use the format this header sequence represents.
      */
     val preferredSerializationVersion: SerializationMagic
+    /**
+     * If non-null, apply this encoding (typically compression) when serializing.
+     */
+    val encoding: SerializationEncoding?
```
rick-r3: can we add a `withEncoding` function too?  Now, I'm somewhat surprised we don't have a bigger API break here.  Do we know why the API checker is not flagging the `val encoding` (presumably as `getEncoding`) like it would for `withEncoding`?.... hmmm.... technically an API break but the addition of `@DoNotImplement` is 100% correct and no one is actually going to have implemented.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/kryo/KryoStreams.kt
andr3ej: * withDataBytes doesn't just provide a ByteBuffer, it first strips the header, so withByteBuffer would lose that self-documentation. it only exists as a method so i can call it from a unit test so i've added @VisibleForTesting and made it explicitly internal
* useAsByteBuffer suggests the stream is closed when the task finishes, which is not the case, so i've gone with alsoAsByteBuffer

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/AMQPStreams.kt
andr3ej: done as above

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
andr3ej: done. the purpose of @DoNotImplement is to make the checker ignore new abstract api, so no breaks are expected here

andr3ej:
correct and correct, no non-null encoding is actually used (except in tests) and only a whitelist that rejects all non-null encodings is used (except in tests)

PR #2476, opened 2018-02-07
CORDA-939 - Dont expose FlowStateMachine via public API (#2438)

anthonykr3:
From: [https://github.com/corda/corda/pull/2438](https://github.com/corda/corda/pull/2438)

* Create CordaInternal attribute for properties on public classes that are not part of the api and apply to FlowLogic.stateMachine

* Remove startFlow from public test api and replace with startFlowAndReturnFuture

* Update api-current with changed signature

* Change test used in documentation to use public test methods

* Remove the rest of the unneccessary usages of the startFlow test utility

* Remove extra whitespace

* Rename startFlowAndReturnFuture back to startFlow

* Update api

* The annotation doesn't appear unless its marked as on the actual getter and setter

* Updated docs and removed pointless attribute

* Deleted whitespace

# Conflicts:
#	.ci/api-current.txt
#	testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt


anthonykr3:
I had a nightmare trying to merge api-current.txt - @mikehearn has regenerated it in master but it doesn't look like that's been cherry picked across here yet so I've just regenerated it and checked it in, rather than manually going through and trying to pick out my changes.

fenryka:
@anthonykr3 looks like you need to re-re-add it to the PR as API stability checks are failing

fenryka:
As discussed in person, revert to the V1 api check for now, but raise a second PR prior to cutting the release that finalises on the new stable API's

anthonykr3:
Created [JIRA](https://r3-cev.atlassian.net/browse/CORDA-991) to update api-current.txt when branch is code complete.

At that point we should also back merge that api-current.txt into master.

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -418,7 +419,9 @@ abstract class FlowLogic<out T> {
      * is public only because it must be accessed across module boundaries.
      */
     var stateMachine: FlowStateMachine<*>
+        @CordaInternal
```
mikehearn: You can annotate the var directly.

testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
```diff
@@ -60,13 +61,9 @@ fun testActor(owningLegalIdentity: CordaX500Name = CordaX500Name("Test Company I
...
- * has origin [Origin.RPC] and actor with id "Only For Testing".
- */
-fun <T> StartedNodeServices.startFlow(logic: FlowLogic<T>): FlowStateMachine<T> = startFlow(logic, newContext()).getOrThrow()
-
 /**
  * Creates a new [InvocationContext] for testing purposes.
  */
 fun StartedNodeServices.newContext() = testContext(myInfo.chooseIdentity().name)
+
+fun <T> StartedNodeServices.startFlow(logic: FlowLogic<T>): CordaFuture<T> = startFlow(logic, newContext()).getOrThrow().resultFuture
```
mikehearn: Please don't remove documentation - all public APIs are meant to be documented. Try enabling the "Missing KDoc" inspection in intellij, there is an API scope pre-defined that will let you limit it to the stabilised modules.

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
anthonykr3: The problem is if you annotate the var directly it's not ending up in the api text file, so I can't filter it out, because the API is breaking it down to a getter and a setter method.

Although now the intellisense documentation isn't showing the @CordaInternal attribute, well in Kotlin anyway, I bet it would be fine in Java. I considered annotating both the property and the getter and setters but that seemed like overkill.

anthonykr3:
Created a PR on master to also add the documentation changes: [https://github.com/corda/corda/pull/2480](https://github.com/corda/corda/pull/2480)

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
mikehearn: Oh, that's odd, I thought annotating a field in Kotlin automatically propagated the annotation to the getters and setters as well. OK then.

Issue #2300, opened 2017-12-28
Multiple Observables arrived with the same ID


trferreira-BR:
Hi,

I am using V2.0 Corda and I create an Observable in web app (Spring Boot) to receive information from Node Corda by RPC and some moments It throws an exception with message 'Multiple Observables arrived with the same ID'.

That problem is possible to see because I maped in logback file an appender/logger in our web app:

`<logger name="org.apache.activemq.artemis.core" level="error" additivity="false">
      <appender-ref ref="ROLLING-FILE-APPENDER"/>
</logger>`

obs.: We identified that problem since version V1.0 Corda.

The log trace:
2017-12-19 09:27:11 - AMQ214000: Failed to call onMessage
com.esotericsoftware.kryo.KryoException: java.lang.IllegalArgumentException: Multiple Observables arrived with the same ID ObservableId(toLong=8174636548186332080)
Serialization trace:
updates (net.corda.core.messaging.DataFeed)
value (net.corda.core.utilities.Try$Success)
	at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:144)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.read(CompatibleFieldSerializer.java:147)
	at com.esotericsoftware.kryo.Kryo.readObject(Kryo.java:731)
	at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:125)
	at com.esotericsoftware.kryo.serializers.CompatibleFieldSerializer.read(CompatibleFieldSerializer.java:147)
	at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:813)
	at net.corda.nodeapi.internal.serialization.AbstractKryoSerializationScheme$deserialize$$inlined$use$lambda$1$1.invoke(SerializationScheme.kt:209)
	at net.corda.nodeapi.internal.serialization.AbstractKryoSerializationScheme$deserialize$$inlined$use$lambda$1$1.invoke(SerializationScheme.kt:152)
	at net.corda.nodeapi.internal.serialization.AbstractKryoSerializationScheme.withContext(SerializationScheme.kt:191)
	at net.corda.nodeapi.internal.serialization.AbstractKryoSerializationScheme.access$withContext(SerializationScheme.kt:152)
	at net.corda.nodeapi.internal.serialization.AbstractKryoSerializationScheme$deserialize$$inlined$use$lambda$1.execute(SerializationScheme.kt:206)
[...]

PR #2477, opened 2018-02-07
Add RPC deduplication to client and server

exFalso:
Hopefully this should fix https://github.com/corda/corda/issues/2300

I found one way of reproducing the dupes, which is kicking of the client at inopportune times, this is what I used for the test.

The dedupe cache is based on time which may result in extended memory usage under heavy load in particular in the server (and hence the node). Just something to keep in mind during perf testing

### DIFF
--- a/client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
+++ b/client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
@@ -19,6 +19,7 @@ import org.apache.activemq.artemis.api.core.SimpleString
 import org.junit.After
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
+import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 import rx.Observable
@@ -127,7 +128,7 @@ class RPCStabilityTests {
         rpcDriver {
             fun startAndCloseServer(broker: RpcBrokerHandle) {
                 startRpcServerWithBrokerRunning(
-                        configuration = RPCServerConfiguration.default.copy(consumerPoolSize = 1, producerPoolBound = 1),
+                        configuration = RPCServerConfiguration.default,
                         ops = DummyOps,
                         brokerHandle = broker
                 ).rpcServer.close()
@@ -148,7 +149,7 @@ class RPCStabilityTests {
     @Test
     fun `rpc client close doesnt leak broker resources`() {
         rpcDriver {
-            val server = startRpcServer(configuration = RPCServerConfiguration.default.copy(consumerPoolSize = 1, producerPoolBound = 1), ops = DummyOps).get()
+            val server = startRpcServer(configuration = RPCServerConfiguration.default, ops = DummyOps).get()
             RPCClient<RPCOps>(server.broker.hostAndPort!!).start(RPCOps::class.java, rpcTestUser.username, rpcTestUser.password).close()
             val initial = server.broker.getStats()
             repeat(100) {
@@ -337,11 +338,12 @@ class RPCStabilityTests {
             val request = RPCApi.ClientToServer.RpcRequest(
                     clientAddress = SimpleString(myQueue),
                     methodName = SlowConsumerRPCOps::streamAtInterval.name,
-                    serialisedArguments = listOf(10.millis, 123456).serialize(context = SerializationDefaults.RPC_SERVER_CONTEXT).bytes,
+                    serialisedArguments = listOf(10.millis, 123456).serialize(context = SerializationDefaults.RPC_SERVER_CONTEXT),
                     replyId = Trace.InvocationId.newInstance(),
                     sessionId = Trace.SessionId.newInstance()
             )
             request.writeToClientMessage(message)
+            message.putLongProperty(RPCApi.DEDUPLICATION_SEQUENCE_NUMBER_FIELD_NAME, 0)
             producer.send(message)
             session.commit()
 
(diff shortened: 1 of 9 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/ArtemisDeduplicationCache.kt
```diff
@@ -0,0 +1,20 @@
...
+import java.time.Duration
+import java.util.concurrent.TimeUnit
+
+class ArtemisDeduplicationCache(expiry: Duration) {
+    private val messageIdCache =
+            CacheBuilder.newBuilder()
+                    .expireAfterAccess(expiry.toNanos(), TimeUnit.NANOSECONDS)
+                    .build<Long, Unit>()
+
+    fun checkDuplicateMessageId(messageId: Long): Boolean {
```
tlil: Perhaps rename this method to indicate more clearly what the result is indicating? `s/check/is/` for instance #nit

exFalso: I didn't want to use `is` because the function does mutation.. although `check` doesn't warn about that either

tlil: Fair - can't really think of a better name atm anyway. Unless you did something like `checkIfMessageIdIsDuplicate()`... but not sure if that's much better.

exFalso:
Ok I bloated this PR a bit too much. Ended up rewiring the internal threading/pooling of RPC, in particular it doesn't pool artemis sessions anymore - a consequence of this is that a client kick does *not* reproduce dupes anymore, which is a good sign i think(?). Throughput is the same as before.
I moved sending into separate threads both on the client and server, this is also what defines the dupeid sequence. As a lot changed since the approval I'm requesting another pair of eyes on this (as @rick-r3 is away)

node-api/src/main/kotlin/net/corda/nodeapi/internal/DeduplicationChecker.kt
```diff
@@ -0,0 +1,29 @@
...
+            .build(WatermarkCacheLoader)
+
+    private object WatermarkCacheLoader : CacheLoader<Any, AtomicLong>() {
+        override fun load(key: Any) = AtomicLong(-1)
+    }
+
+    /**
+     * @param identity the identity that generates the sequence numbers.
+     * @param sequenceNumber the sequence number to check.
+     */
```
tlil: Add `@returns` or description at the top of the kdoc

node/src/main/kotlin/net/corda/node/services/messaging/RPCServer.kt
```diff
@@ -370,6 +407,35 @@ class RPCServer(
...
+        private val serializationContextWithObservableContext = RpcServerObservableSerializer.createContext(this)
+
+        fun sendMessage(serverToClient: RPCApi.ServerToClient) {
+            sendJobQueue.put(RpcSendJob.Send(contextDatabaseOrNull, clientAddress, serializationContextWithObservableContext, serverToClient))
+        }
+    }
+
+    private sealed class RpcSendJob {
+        data class Send(
+                // TODO HACK this is because during serialisation we subscribe to observables that may use
```
tlil: Any follow-up action to clean this up?

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
```diff
@@ -350,6 +352,46 @@ class RPCStabilityTests {
...
+    }
+    class StreamOpsImpl : StreamOps {
+        override val protocolVersion = 0
+        override fun stream(streamInterval: Duration): Observable<Long> {
+            return Observable.interval(streamInterval.toNanos(), TimeUnit.NANOSECONDS)
+        }
+    }
+    @Ignore("This is flaky as sometimes artemis delivers out of order messages after the kick")
+    @Test
+    fun `deduplication on the client side`() {
```
tlil: Would be good to have a "working" / reliable test for this before merging. This seems to be the only test introduced, and it's marked as `@Ignore` (c.f. @adagys' comment).

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
```diff
@@ -350,6 +352,46 @@ class RPCStabilityTests {
...
+    interface StreamOps : RPCOps {
+        fun stream(streamInterval: Duration): Observable<Long>
+    }
+    class StreamOpsImpl : StreamOps {
+        override val protocolVersion = 0
+        override fun stream(streamInterval: Duration): Observable<Long> {
+            return Observable.interval(streamInterval.toNanos(), TimeUnit.NANOSECONDS)
+        }
+    }
+    @Ignore("This is flaky as sometimes artemis delivers out of order messages after the kick")
```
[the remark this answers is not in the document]
exFalso: You're right! I removed this test and added two better ones. These dupe messages explicitly (double `producer.send`s)

node/src/main/kotlin/net/corda/node/services/messaging/RPCServer.kt
exFalso: Good point, I'll add a JIRA. Need @rick-r3's input on how to do this prettier.

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
exFalso: Yes, replaced this with two better tests. These dupe explicitly which isn't optimal, but test the new codepaths nevertheless. Best would be a repro of the original bug but I couldn't find a way

tlil: Thanks - I think those two tests are a good start. Just wanted to make sure we had _something_ covering the new code paths.

tlil: I heard you explained to Gavin and Matthew earlier how hard it is to repro, so I can appreciate writing a test for it is hard. I think this is good.

exFalso:
@adagys @tlil addressed comments, back you you!

PR #2478, opened 2018-02-07
CORDA-993 cleaning node depndencies

ags313:
Removing client:rpc dependency from node


node-api/src/main/kotlin/net/corda/nodeapi/RPCApi.kt
```diff
@@ -18,6 +18,7 @@ import org.apache.activemq.artemis.api.core.management.CoreNotificationType
 import org.apache.activemq.artemis.api.core.management.ManagementHelper
 import org.apache.activemq.artemis.reader.MessageUtil
 import rx.Notification
+import rx.Observable
```
sollecitom: ?

sollecitom:
@ags313 as discussed, `api-current.txt` on master is corrupted. Please regenerate it locally and push it for now. @anthonykr3 is aware and on course to fix these issues.

PR #2479, opened 2018-02-07
CORDA-992 - Make the finger printer pluggable for serialization factory

fenryka:


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializerFactory.kt
```diff
@@ -40,7 +40,13 @@ data class FactorySchemaAndDescriptor(val schemas: SerializationSchemas, val typ
 open class SerializerFactory(
         val whitelist: ClassWhitelist,
         cl: ClassLoader,
-        private val evolutionSerializerGetter: EvolutionSerializerGetterBase = EvolutionSerializerGetter()) {
+        private val evolutionSerializerGetter: EvolutionSerializerGetterBase = EvolutionSerializerGetter(),
+        val fingerPrinter: FingerPrinter = SerializerFingerPrinter()) {
+
+    init {
+        fingerPrinter.setOwner(this)
```
rick-r3: Are we able to move this to be the last thing initialised in the `SerializerFactory`  ?  Currently, if the finger printer were to look at the factory inside `setOwner` it would see a partially initialised factory.

fenryka: You can't do this at construction time as ideally you'd have teh serialization Factory create the default fingerprinter and pass itself in as a reffernce, except at that point construction of the object hasn't completed so it cna't be passed in and you can't defer the insitilsation to init as then you can't default it... basically this would totally work in C but aprently not in Java

PR #2481, opened 2018-02-07
Attach Jolokia agents in Node startup [ENT-1152]

IgorNitto:
More surgery to remove runtime dependencies on Capsule:

    Jolokia agents are loaded dynamically if configured
    Renamed exportJmxTo to jmxMonitoringHttpPort and expose it as config variable
    Updated documentation and tests


sollecitom:
Please use branch naming conventions: features/jira_id.

This shouldn't be ENT-1152 for an OS story.

Move the Jira story across projects and reopen the pr from the new branch.

sollecitom (changes requested):
As per the comment

PR #2485, opened 2018-02-08
API scanner includes all inherited methods so that when methods are m…

shamsasari:
…oved up the type hierarchy (a backwards compatible change) it only shows up as an additional line in the API file.


shamsasari:
I forgot about this. Yes it needs to be opened in the new repo.

PR #2488, opened 2018-02-08
CORDA-967 Multi-tx signature verification

kchalkias:
Add support for signature verification when we sign over a block of transactions.
According to this model, if and when block-tx signing is enabled in the future, a `TransactionSignature` might be accompanied by a partial Merkle tree, which indicates it's a tx-block signature. This might be useful if we allow block signing in the notary side.

Note that for backwards compatibility and performance purposes, if we only sign over a single transaction, then no Merkle tree computations are applied. 

Also, for advanced privacy, the leaves of a tx-block Merkle tree are `hash(txId)` and not `txId`, as shown in the following diagram.

If we sign a single tx, we sign over the `txId` itself; in this case, we don't utilise the partial tree which should be null.
tx-block signing

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
@@ -560,8 +560,7 @@ object Crypto {
     @JvmStatic
     @Throws(InvalidKeyException::class, SignatureException::class)
     fun doVerify(txId: SecureHash, transactionSignature: TransactionSignature): Boolean {
-        val hashSigned = hashSigned(txId, transactionSignature.partialMerkleTree)
-        val signableData = SignableData(hashSigned, transactionSignature.signatureMetadata)
+        val signableData = SignableData(originalSignedHash(txId, transactionSignature.partialMerkleTree), transactionSignature.signatureMetadata)
         return Crypto.doVerify(transactionSignature.by, transactionSignature.bytes, signableData.serialize().bytes)
     }
 
@@ -580,8 +579,7 @@ object Crypto {
     @JvmStatic
     @Throws(SignatureException::class)
     fun isValid(txId: SecureHash, transactionSignature: TransactionSignature): Boolean {
-        val hashSigned = hashSigned(txId, transactionSignature.partialMerkleTree)
-        val signableData = SignableData(hashSigned, transactionSignature.signatureMetadata)
+        val signableData = SignableData(originalSignedHash(txId, transactionSignature.partialMerkleTree), transactionSignature.signatureMetadata)
         return isValid(
                 findSignatureScheme(transactionSignature.by),
                 transactionSignature.by,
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/net/corda/core/crypto/TransactionSignature.kt
```diff
@@ -9,12 +9,20 @@ import java.util.*
...
+ * Note that [partialMerkleTree] should always be null when only one transaction is signed.
+ * @property bytes actual bytes of the cryptographic signature.
+ * @property by [PublicKey] of the signer.
+ * @property signatureMetadata attached [SignatureMetadata] for this signature.
+ * @property partialMerkleTree required when multi-transaction signing is utilised.
  */
 @CordaSerializable
-class TransactionSignature(bytes: ByteArray, val by: PublicKey, val signatureMetadata: SignatureMetadata) : DigitalSignature(bytes) {
+class TransactionSignature(bytes: ByteArray, val by: PublicKey, val signatureMetadata: SignatureMetadata, val partialMerkleTree: PartialMerkleTree?) : DigitalSignature(bytes) {
+    constructor(bytes: ByteArray, by: PublicKey, signatureMetadata: SignatureMetadata) : this(bytes, by, signatureMetadata, null)
```
adagys: secondary constructor needs api docs

core/src/test/kotlin/net/corda/core/crypto/TransactionSignatureTest.kt
```diff
@@ -41,4 +45,84 @@ class TransactionSignatureTest {
...
+        // All txs can be verified, as they are all included in the provided partial tree.
+        txIds.forEach {
+            assertTrue(Crypto.doVerify(it, txSignatureWithFullTree))
+        }
+    }
+
+    @Test
+    fun `Verify one-tx signature`() {
+        val keyPair = Crypto.deriveKeyPairFromEntropy(Crypto.EDDSA_ED25519_SHA512, BigInteger.valueOf(1234567890L))
+        // Deterministically create 5 txIds.
```
adagys: the comment doesn't apply here

core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
```diff
@@ -578,7 +580,8 @@ object Crypto {
     @JvmStatic
     @Throws(SignatureException::class)
     fun isValid(txId: SecureHash, transactionSignature: TransactionSignature): Boolean {
-        val signableData = SignableData(txId, transactionSignature.signatureMetadata)
+        val hashSigned = hashSigned(txId, transactionSignature.partialMerkleTree)
```
adagys: maybe call the method something like "getOriginalSignedHash"

core/src/main/kotlin/net/corda/core/crypto/TransactionSignature.kt
kchalkias: Correct. Also noticed that our code-base is missing api docs to secondary constructors in general.

kchalkias:
back to you @adagys

core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
```diff
@@ -75,7 +76,7 @@ object Crypto {
             listOf(AlgorithmIdentifier(PKCSObjectIdentifiers.rsaEncryption, null)),
             BouncyCastleProvider.PROVIDER_NAME,
             "RSA",
-            "SHA256WITHRSAEncryption",
+            "SHA256WITHRSA",
```
adagys: what's the effect of this change btw

kchalkias: It's the same OID, consider it an aesthetic change, mostly to avoid confusion of the word "encryption"
https://crypto.stackexchange.com/questions/12973/what-is-the-difference-between-sha256withrsa-and-sha256withrsaencryption-digest

PR #2489, opened 2018-02-08
Allows the webserver JAR used by each node in deployNodes to be configured.

joeldudleyr3:


gradle-plugins/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -130,7 +142,13 @@ class Node(private val project: Project) : CordformNode() {
      * Installs the corda webserver JAR to the node directory
      */
     private fun installWebserverJar() {
-        val webJar = Cordformation.verifyAndGetRuntimeJar(project, "corda-webserver")
+        // If no webserver JAR is provided, the default development webserver is used.
```
Clintonio: Can you add trace logging to both branches to state which JAR is starting.

Clintonio:
experimental/api unstable.

PR #2500, opened 2018-02-09
Experimental - BDD Test Framework

tlil:
This project illustrates how one can use Cucumber / BDD to drive and test homogeneous and heterogeneous Corda networks on a local machine. The framework has built-in support for Dockerised node dependencies so that you easily can spin up a Corda node locally that, for instance, uses a 3rd party database provider such as MS SQL Server or Postgres.

The core testing library is split out from the BDD framework to allow for use in non-BDD setups as well.

**Note**: This is an experimental module and by no means complete. If anyone wants to extend the functionality or use it in different ways, feel free to come with suggestions / modifications.


tlil:
Great - thanks, @josecoll - I'll merge this now. Let me know if there's anything I can do to help with the extensions that you're making.

PR #2501, opened 2018-02-12
CORDA-1006: Undoing the wiring of maxMessageSize as it's not correctly implemente…

shamsasari:
…d and updating the docs to clarify its status.

The network parameter was just fed into Artemis' minLargeMessageSize property which isn't the same thing.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
+++ b/core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
@@ -9,7 +9,7 @@ import java.time.Instant
  * correctly interoperate with each other.
  * @property minimumPlatformVersion Minimum version of Corda platform that is required for nodes in the network.
  * @property notaries List of well known and trusted notary identities with information on validation type.
- * @property maxMessageSize Maximum P2P message sent over the wire in bytes.
+ * @property maxMessageSize This is currently ignored. However, it will be wired up in a future release.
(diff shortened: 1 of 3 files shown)


gendal:
Don't we need this feature (albeit correctly implemented!) in order to be able to ship 3.0?  ie we need to be sure that data persisted by v3.0 can be used/understood by future versions.  If we don't have a transaction/message-size limit then people deploying on Corda 3.0 risk having data that cannot be understood by later versions.    Or are we working around this by hard-coding a relatively low limit?

shamsasari:
I think you're referring to maxTransactionSize, which is already wired-up. maxMessageSize only applies to the transitory messages on the wire.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -9,7 +9,7 @@ import java.time.Instant
  * correctly interoperate with each other.
  * @property minimumPlatformVersion Minimum version of Corda platform that is required for nodes in the network.
  * @property notaries List of well known and trusted notary identities with information on validation type.
- * @property maxMessageSize Maximum P2P message sent over the wire in bytes.
+ * @property maxMessageSize This is currently ignored, but will be wired in a future release.
```
fenryka: will be wired *up* 

Or....

This is currently ignored. However, it will be wired up in a future release.

shamsasari:
Comment addressed.

PR #2504, opened 2018-02-12
Run H2 fix test in parallel

m4ksio:
30-50 seconds saved.


node/src/test/kotlin/net/corda/node/internal/AbstractNodeTests.kt
```diff
@@ -42,12 +42,25 @@ class AbstractNodeTests {
 
     @Test
     fun `H2 fix is applied`() {
-        repeat(if (relaxedThoroughness) 1 else 100) {
+
```
andr3ej: remove all blank lines, the coding standard is vertically dense

node/src/test/kotlin/net/corda/node/internal/AbstractNodeTests.kt
```diff
@@ -42,12 +42,25 @@ class AbstractNodeTests {
...
+        val pool = Executors.newFixedThreadPool(10)
```
andr3ej: this pool is never shutdown

m4ksio:
I have successfully trigger the bug - by commenting the H2 fix and running the test a few times. It's true, that as the original it does not always manifest.

andr3ej:
it shouldn't, but it somehow does. i would like to see proof that the bug is sufficiently reliably triggered in TC before this is merged to master

node/src/test/kotlin/net/corda/node/internal/AbstractNodeTests.kt
```diff
@@ -42,12 +42,22 @@ class AbstractNodeTests {
 
     @Test
     fun `H2 fix is applied`() {
-        repeat(if (relaxedThoroughness) 1 else 100) {
+        val pool = Executors.newFixedThreadPool(5)
+        val runs = if (relaxedThoroughness) 1 else 100
+        (0 until runs).map {
             // Two "nodes" seems to be the magic number to reproduce the problem:
```
andr3ej: this comment needs updating

PR #2506, opened 2018-02-12
CORDA-1009 Remove X509EdDSAEngine dependency on X509Key

kchalkias:
Required for the deterministic core, because `sun.security.x509.X509Key` isn't going to link as it uses an internal `sun.` class.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/X509EdDSAEngine.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/X509EdDSAEngine.kt
@@ -27,12 +27,7 @@ class X509EdDSAEngine : Signature {
     override fun engineInitSign(privateKey: PrivateKey, random: SecureRandom) = engine.initSign(privateKey, random)
 
     override fun engineInitVerify(publicKey: PublicKey) {
-        val parsedKey = if (publicKey is sun.security.x509.X509Key) {
-            EdDSAPublicKey(X509EncodedKeySpec(publicKey.encoded))
-        } else {
-            publicKey
-        }
-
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/net/corda/core/internal/X509EdDSAEngine.kt
```diff
@@ -27,12 +27,11 @@ class X509EdDSAEngine : Signature {
...
 
     override fun engineInitVerify(publicKey: PublicKey) {
-        val parsedKey = if (publicKey is sun.security.x509.X509Key) {
-            EdDSAPublicKey(X509EncodedKeySpec(publicKey.encoded))
-        } else {
-            publicKey
+        val parsedKey = try {
+            publicKey as? EdDSAPublicKey ?: EdDSAPublicKey(X509EncodedKeySpec(publicKey.encoded))
+        } catch(e: Exception) {
+            throw (InvalidKeyException(e.message))
```
chrisr3: Throwing `InvalidKeyException` changes the semantics of `engineInitVerify()` for non-X.509 keys. I thought you would do something more like this:
```kotlin
val parsedKey = if (publicKey.format == "X.509" && publicKey !is EdDSAPublicKey) {
    EdDSAPublicKey(X509EncodedKeySpec(publicKey.encoded))
} else {
    publicKey
}
```

kchalkias: `format == "X.509"` is the same for every key type (RSA, EC, Sphincs) so we cannot distinguish between types based on this. `try-catch` was used because `EdDSAPublicKey(X509EncodedKeySpec(publicKey.encoded))` might returns an `InvalidKeySpecException` that does not inherit from `InvalidKeyException`, which is what we expect from `engineInitVerify`.

PR #2508, opened 2018-02-12
Use in-memory H2 for integration test

m4ksio:
Only a few integration tests really need to used on-disk database. Switching to in-memory for other cases allows us to save time!


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -109,6 +110,16 @@ class DriverDSLImpl(
     private lateinit var _notaries: CordaFuture<List<NotaryHandle>>
     override val notaryHandles: List<NotaryHandle> get() = _notaries.getOrThrow()
 
+    // While starting with inProcess mode, we need to have different names to avoid clashes
+    private var inMemoryCounter = 0
+    private fun getInMemoryConfig() = if (inMemoryDB) {
```
szymonsztuka: Could we name it 'getDatabaseConfig' or something like this? It will be easier to sync R3.Corda with this.

node/src/main/resources/reference.conf
```diff
@@ -5,9 +5,9 @@ keyStorePassword = "cordacadevpass"
 trustStorePassword = "trustpass"
 dataSourceProperties = {
     dataSourceClassName = org.h2.jdbcx.JdbcDataSource
-    "dataSource.url" = "jdbc:h2:file:"${baseDirectory}"/persistence;DB_CLOSE_ON_EXIT=FALSE;LOCK_TIMEOUT=10000;WRITE_DELAY=100;AUTO_SERVER_PORT="${h2port}
-    "dataSource.user" = sa
-    "dataSource.password" = ""
+    dataSource.url = "jdbc:h2:file:"${baseDirectory}"/persistence;DB_CLOSE_ON_EXIT=FALSE;LOCK_TIMEOUT=10000;WRITE_DELAY=100;AUTO_SERVER_PORT="${h2port}
+    dataSource.user = sa
+    dataSource.password = ""
```
szymonsztuka: Could you also remove double quotes from docs/source/example-code/src/main/resources/example-node.conf

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -109,6 +110,16 @@ class DriverDSLImpl(
...
     override val notaryHandles: List<NotaryHandle> get() = _notaries.getOrThrow()
 
+    // While starting with inProcess mode, we need to have different names to avoid clashes
+    private var inMemoryCounter = 0
+    private fun getDatabaseConfig() = if (inMemoryDB) {
+        mapOf("dataSourceProperties" to mapOf(
+                "dataSource.url" to "jdbc:h2:mem:persistence${inMemoryCounter++};DB_CLOSE_ON_EXIT=FALSE;LOCK_TIMEOUT=10000;WRITE_DELAY=100"
+        ))
+    } else {
+        emptyMap()
```
adagys: can you add a comment what db config this would default to?

mikehearn:
@m4ksio This PR was approved but never merged and now it's bit-rotted. Can you re-base and get it merged please?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -233,7 +249,7 @@ class DriverDSLImpl(
                 baseDirectory = baseDirectory(name),
                 allowMissingConfig = true,
                 configOverrides = if (overrides.hasPath("devMode")) overrides else overrides + mapOf("devMode" to true)
-        ))
+        )).enhanceForInMemoryDB()
```
szymonsztuka: Would it be better to pass amended property via `configOverrides`?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -110,6 +111,9 @@ class DriverDSLImpl(
     private lateinit var _notaries: CordaFuture<List<NotaryHandle>>
     override val notaryHandles: List<NotaryHandle> get() = _notaries.getOrThrow()
 
+    // While starting with inProcess mode, we need to have different names to avoid clashes
+    private var inMemoryCounter = 0
```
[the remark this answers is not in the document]
m4ksio: Is DriverDSL intended to be thread-safe? Variable is incremented while creating node config which usually happens in the single test

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
m4ksio: To evaluate is we can switch H2 to in-memory, we need to access the finished config first, so it's seems easier.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -126,6 +130,16 @@ class DriverDSLImpl(
 
     private val jolokiaJarPath: String by lazy { resolveJar(".*jolokia-jvm-.*-agent\\.jar$") }
 
+    private fun NodeConfig.enhanceForInMemoryDB(): NodeConfig = this.run {
```
adagys: the "enhance" bit makes it sound like magic :) maybe call it something like "overrideH2DataSource"?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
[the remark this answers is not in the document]
adagys: guess it's possible that nodes could be created in parallel (albeit unlikely). Doesn't hurt to make it atomic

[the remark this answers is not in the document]
m4ksio: Ok, by popular demand, although the rest of the class is hardly thread-safe :)

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
szymonsztuka: True, fair point

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -283,7 +284,8 @@ data class DriverParameters(
         val jmxPolicy: JmxPolicy = JmxPolicy(),
         val networkParameters: NetworkParameters = testNetworkParameters(notaries = emptyList()),
         val notaryCustomOverrides: Map<String, Any?> = emptyMap(),
-        val initialiseSerialization: Boolean = true
+        val initialiseSerialization: Boolean = true,
+        val inMemoryDB: Boolean = true
```
mikehearn: new parameters should be added to the kdoc as well

m4ksio: Added

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -268,6 +268,9 @@ fun <A> driver(defaultParameters: DriverParameters = DriverParameters(), dsl: Dr
  * @property networkParameters The network parameters to be used by all the nodes. [NetworkParameters.notaries] must be
  *     empty as notaries are defined by [notarySpecs].
  * @property notaryCustomOverrides Extra settings that need to be passed to the notary.
+ * @property initialiseSerialization Indicates whether to initialized the serialization subsystem.
+ * @property inMemoryDB Whether to use in-memory H2 for new nodes rather then on-disk (has no effect if node is configured
```
szymonsztuka: I wonder if to empahsis the functional diffrences:  

"Whether to use in-memory H2 for new nodes (the instance is created quicker however it is not persisted between the node restarts) rather ...."

m4ksio: Is this really a functional difference?

szymonsztuka: I think yes, in the same test a restarted node will have fresh database, contrary to what was before.

PR #2509, opened 2018-02-12
CORDA-1008 Samples fail to run due to wrongly configured max transaction size in the network bootstrapper

patrickkuo:
Trader demo runs fine after the fix.


patrickkuo:
good spot, integration test didn't fail so i guess its ok for now, I will change this as well to make it consistence with other configs.

PR #2510, opened 2018-02-12
CORDA-939 Modify Api Scanner to check api for internal exposures

anthonykr3:
Modify the check-api-changes.sh script to check for any api's that contain .internal modules. I've ignored exposures of nodeapi internals for now as that module isn't stable yet.

I also fixed a few more instances where internal api's were being exposed and moved the MutableClock/CordaClock classes into the public API's as @mikehearn suggested.


shamsasari:
Plus it's easier to maintain in Java code.

anthonykr3:
ScanApi doesn't do any checks or comparisons at all - all it does it iterate through the class files and output the class/method signatures of the current build.
I'm sure I could change ScanApi to do everything and get rid of the shell script but it wouldn't be done in time for the v3 release. I can raise a Jira for it and put it on the backlog?

.ci/check-api-changes.sh
```diff
@@ -31,27 +31,46 @@ if [ $removalCount -gt 0 ]; then
...
+userDiffContents=`sed '/@net.corda.core.CordaInternal/d' <<< $userDiffContents`
+
 newAbstracts=$(echo "$userDiffContents" | grep "^+" | grep "\(public\|protected\) abstract")
 abstractCount=`grep -v "^$" <<EOF | wc -l
 $newAbstracts
 EOF
 `
 
+#Get a list of any methods that expose classes in .internal. namespaces
+#Exclude nodeapi.internal as that is not stabalised yet
```
shamsasari: Make this a TODO comment

.ci/check-api-changes.sh
```diff
@@ -31,27 +31,46 @@ if [ $removalCount -gt 0 ]; then
...
+##Remove any methods that are marked with CordaInteral attribute
+userDiffContents=`sed '/@net.corda.core.CordaInternal/d' <<< $userDiffContents`
+
 newAbstracts=$(echo "$userDiffContents" | grep "^+" | grep "\(public\|protected\) abstract")
 abstractCount=`grep -v "^$" <<EOF | wc -l
 $newAbstracts
 EOF
 `
 
+#Get a list of any methods that expose classes in .internal. namespaces
```
shamsasari: Not just methods but also classes which extend/implement an internal class.

chrisr3:
> Can we not instead have the internal checks done in ScanApi? Doing this stuff in the script doesn't make sense. The same can be said of DoNotImplement.
The ScanApi task scans the class files for a given module and then outputs the results to `$module/buildDir/api`. These per-module output files are then collated into a _single_ output file in `$project/buildDir/api` by the GenerateApi task, and it is this final output file that is scanned by the script to give a comprehensive description of all the violations introduced by the PR.
How does this not make sense? Would you prefer TeamCity to fail the build on the first API violation it finds? Or perhaps on the first module with an API violation? Do you want to keep all of the stable API definitions in one place, or scattered across the build tree in multiple places?
I did check for `diff` plugins for Gradle but wasn't impressed, and a `bash` script is actually a very powerful place for this kind of task.

testing/node-driver/src/main/kotlin/net/corda/testing/node/InMemoryMessagingNetwork.kt
```diff
@@ -49,9 +50,9 @@ import kotlin.concurrent.thread
  */
 @ThreadSafe
 class InMemoryMessagingNetwork internal constructor(
-        private val sendManuallyPumped: Boolean,
-        private val servicePeerAllocationStrategy: ServicePeerAllocationStrategy = InMemoryMessagingNetwork.ServicePeerAllocationStrategy.Random(),
-        private val messagesInFlight: ReusableLatch = ReusableLatch()
+        @CordaInternal private val sendManuallyPumped: Boolean,
```
shamsasari: I think the problem here is the `internal` c'tor declaration above - the c'tor is still public at the bytecode level. Instead make this c'tor private and have an `internal` factory method.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -210,12 +211,16 @@ open class MockServices private constructor(
...
-        return cordappServices.getInstance(type) ?: throw IllegalArgumentException("Corda service ${type.name} does not exist")
+        return cordappServices.getInstance(type)
+                ?: throw IllegalArgumentException("Corda service ${type.name} does not exist")
     }
 
     override fun jdbcSession(): Connection = throw UnsupportedOperationException()
 
     override fun registerUnloadHandler(runOnStop: () -> Unit) = throw UnsupportedOperationException()
+
+    fun addMockCordapp(contractClassName: ContractClassName, attachments: MockAttachmentStorage)
```
shamsasari: Can you use a code block if this isn't a one-liner.

testing/test-utils/src/main/kotlin/net/corda/testing/internal/InternalSerializationTestHelpers.kt
```diff
@@ -1,8 +1,19 @@
...
+import net.corda.node.serialization.KryoServerSerializationScheme
+import net.corda.nodeapi.internal.serialization.*
+import net.corda.nodeapi.internal.serialization.amqp.AMQPClientSerializationScheme
+import net.corda.nodeapi.internal.serialization.amqp.AMQPServerSerializationScheme
+import net.corda.testing.core.GlobalSerializationEnvironment
 import net.corda.testing.core.SerializationEnvironmentRule
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ExecutorService
+
+internal val inVMExecutors = ConcurrentHashMap<SerializationEnvironment, ExecutorService>()
```
shamsasari: `internal` isn't really needed as it's already in an internal package.

chrisr3:
> Plus it's easier to maintain in Java code.

How so? The point of the API Stability check is to compare the API of the current build with the existing stable API definition, and look for Bad Things. Generating the output is the _less_ interesting half of the job. Do you want to implement `diff` in Java?

shamsasari:
@chrisr3 I wasn't referring to the differing component, but just to the scanning for forbidden references. But as you say my solution isn't ideal as `ScanApi` runs per-module.

.ci/check-api-changes.sh
```diff
@@ -31,27 +31,46 @@ if [ $removalCount -gt 0 ]; then
...
+#Exclude nodeapi.internal as that is not stabalised yet
+newInternalExposures=$(echo "$userDiffContents" | grep "^+" | grep -v "\.nodeapi\.internal\." | grep "\.internal\." )
```
chrisr3: There are also references to a class called `kotlin.jvm.internal.markers.KMappedMarker`. But do we really want references to `net.corda.nodeapi.internal.*` classes in our public API? How would we get rid of them when NodeAPI _does_ become stable?

gradle-plugins/api-scanner/build.gradle
```diff
@@ -5,6 +5,7 @@ apply plugin: 'com.jfrog.artifactory'
 description "Generates a summary of the artifact's public API"
 
 repositories {
+    mavenLocal()
```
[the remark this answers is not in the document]
anthonykr3: @chrisr3 I gave up on those upgrades for v3. After getting the new fast-classpath-scanner finally working, it made so many changes to the method/class signatures that it was basically incompatible without a lot more work to filter out the right bits. So we decided to park that for v3 and filter out issues in the regex ($$internal$). I left this in the build.gradle because it's just useful anyway for testing.

.ci/check-api-changes.sh
anthonykr3: The problem is, CordaPersistance is currently part of the nodeapi.internal package and there is no public version that is available to use. I think at some discussion at some point in the past, someone said 'oh, nodapi.internal isn't stabilised yet so that's not an issue' so I've excluded it. But yes - you're right, we'll have to break backwards compatability later because of it. Possibly one for @mikehearn?

gradle-plugins/api-scanner/build.gradle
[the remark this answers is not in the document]
chrisr3: Was this due to changes to the `toString()` methods in `fast-classpath-scanner` for `ClassInfo`, `MethodInfo` and `FieldInfo`? I'd be interested in learning more about this.

.ci/check-api-changes.sh
shamsasari: `CordaPersistence` is exposed by the driver and MockNetwork APIs. We definitely need to fix this before V3. Can we for now make this internal as we have done for `NodeConfiguration` (e.g. move it into `NodeHandleInternal`, etc)?

gradle-plugins/api-scanner/build.gradle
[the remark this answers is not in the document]
anthonykr3: Yeah - there have been a lot of changes - the scanner is pulling in a lot more attributes on different things, and they're all coming out in the .toString. I spent quite a bit of time trying to filter out the relevant attributes but ran out of time.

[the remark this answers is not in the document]
chrisr3: Perhaps it would be better to sub-class `ClassInfo`, 'MethodInfo` and `FieldInfo` to write "compatible" `toString()` methods? Some of the new attributes are presumably _interesting_ but it would still be better if the output format were under our control.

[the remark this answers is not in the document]
anthonykr3: Yes I think that's probably where I would have ended up

.ci/check-api-changes.sh
anthonykr3: I've hopefully removed all the nodeapi references now.

anthonykr3:
I think that's everything so far

.ci/check-api-changes.sh
```diff
@@ -26,32 +26,51 @@ EOF
 
 echo "Number of API removals/changes: "$removalCount
 if [ $removalCount -gt 0 ]; then
-    echo "$removals"
+    echo "$removals"`
```
shamsasari: Stray backtick.

core/src/main/kotlin/net/corda/core/CordaInternal.kt
```diff
@@ -6,6 +6,6 @@ package net.corda.core
  * These fields are only meant to be used by Corda internally, and are not intended to be part of the public API.
  */
 @Retention(AnnotationRetention.BINARY)
-@Target(AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER)
+@Target(AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER, AnnotationTarget.VALUE_PARAMETER)
```
shamsasari: This should be reverted as we can avoid using this annotation in that scenario by simply making the method with the "internal" parameter `internal`.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -60,7 +61,7 @@ interface OutOfProcess : NodeHandle {
 
 @DoNotImplement
 interface InProcess : NodeHandle {
-    val database: CordaPersistence
+    val databaseConnection: Connection
```
shamsasari: This is not needed as they have access to `jdbcSession()` via `services`.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/internal/DriverInternal.kt
```diff
@@ -57,7 +58,8 @@ data class InProcessImpl(
         private val onStopCallback: () -> Unit,
         private val node: StartedNode<Node>
 ) : InProcess, NodeHandleInternal {
-    override val database: CordaPersistence get() = node.database
+    val database: CordaPersistence = node.database
```
shamsasari: Our internal code can use `node` so this is probably not worth having.

testing/node-driver/src/main/kotlin/net/corda/testing/node/InMemoryMessagingNetwork.kt
```diff
@@ -48,14 +49,19 @@ import kotlin.concurrent.thread
...
         private val messagesInFlight: ReusableLatch = ReusableLatch()
 ) : SingletonSerializeAsToken() {
     companion object {
         private const val MESSAGES_LOG_NAME = "messages"
         private val log = LoggerFactory.getLogger(MESSAGES_LOG_NAME)
+
+        internal fun create(
+                sendManuallyPumped: Boolean,
+                servicePeerAllocationStrategy: ServicePeerAllocationStrategy = InMemoryMessagingNetwork.ServicePeerAllocationStrategy.Random(),
+                messagesInFlight: ReusableLatch = ReusableLatch()) = InMemoryMessagingNetwork(sendManuallyPumped, servicePeerAllocationStrategy, messagesInFlight)
```
shamsasari: Convert this to a code block

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -221,12 +221,17 @@ open class MockServices private constructor(
...
-        return cordappServices.getInstance(type) ?: throw IllegalArgumentException("Corda service ${type.name} does not exist")
+        return cordappServices.getInstance(type)
+                ?: throw IllegalArgumentException("Corda service ${type.name} does not exist")
     }
 
     override fun jdbcSession(): Connection = throw UnsupportedOperationException()
 
     override fun registerUnloadHandler(runOnStop: () -> Unit) = throw UnsupportedOperationException()
+
+    fun addMockCordapp(contractClassName: ContractClassName, attachments: MockAttachmentStorage) {
```
shamsasari: We already have `this.attachments` so the `parameter isn't needed

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -60,7 +61,7 @@ interface OutOfProcess : NodeHandle {
...
     val services: StartedNodeServices
```
shamsasari: As a side point, why the "Started" prefix?

shamsasari:
Hopefully last set of comments from me.

.ci/check-api-changes.sh
```diff
@@ -26,32 +26,51 @@ EOF
...
 
 function forUserImpl() {
-    awk '/DoNotImplement/,/^##/{ next }{ print }' $1
+    awk '/(DoNotImplement|\$\$inlined\$)/,/^##/{ next }{ print }' $1
 }
 
 userDiffContents=`diff -u <(forUserImpl $apiCurrent) <(forUserImpl $APIHOME/../build/api/api-corda-*.txt) | tail -n +3`
 
+##Remove any methods that are marked with CordaInteral attribute
+userDiffContents=`sed '/@net.corda.core.CordaInternal/d' <<< $userDiffContents`
```
chrisr3: To be honest, not printing methods annotated with `@CordaInternal` _is_ something that could be done by the API Scanner Gradle plugin. Look at `ScanApi.writeMethods()`:
```kotlin
for (MethodInfo method : methods) {
    if (isVisible(method.getAccessFlags()) // Only public and protected methods
            && isValid(method.getAccessFlags(), METHOD_MASK) // Excludes bridge and synthetic methods
            // && !hasCordaInternal(method.getAnnotationNames()) // Excludes methods annotated as @CordaInternal
            && !isKotlinInternalScope(method)) {
        writer.append("  ").println(filterAnnotationsFor(method));
    }
}

private static boolean hasCordaInternal(Collection<String> annotationNames) {
    return annotationNames.contains("net.corda.core.CordaInternal");
}
```

chrisr3:
For reference: `@DoNotImplement` was added in [CORDA-704](https://r3-cev.atlassian.net/browse/CORDA-704). We cannot modify the Gradle plugin not to output interfaces and abstract classes annotated as `@DoNotImplement` because we are still guaranteeing API stability to people who invoke these classes' methods. But we reserve the right to add new methods to these classes ourselves, which would then break Cordapps that ignored our warning not to implement them.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/internal/DriverInternal.kt
anthonykr3: `node` is private on the implementation - I don't think we can get rid of this

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
anthonykr3: I've no idea - I've not changed `StartedNodeServices`. In the MockNetwork I created StartedMockNode and UnstartedMockNode, they were originally MockNode and StartedNode<MockNode> - I was trying to differentiate between the two.

.ci/check-api-changes.sh
anthonykr3: No that makes sense.

.ci/check-api-changes.sh
```diff
@@ -31,10 +31,9 @@ if [ $removalCount -gt 0 ]; then
 fi
 
 # Adding new abstract methods could also break the API.
-# However, first exclude anything with the @DoNotImplement annotation.
-
+# However, first exclude classes marked with the @DoNotImplement annotation
 function forUserImpl() {
-    awk '/DoNotImplement/,/^##/{ next }{ print }' $1
+    awk '/(DoNotImplement)/,/^##/{ next }{ print }' $1
```
chrisr3: Do you still need these brackets around `DoNotImplement`?

anthonykr3: Nope

testing/node-driver/src/main/kotlin/net/corda/testing/driver/internal/DriverInternal.kt
shamsasari: Just make `node` public. Accessing `database` would require casting to `InProcessImpl` anyway.

PR #2519, opened 2018-02-12
Test driver default parameters removal

m4ksio:
In order to simplify future API changes, parameters-overload version of ```driver()``` method was removed. Now the full ```DriverParameters``` object is expected

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -3641,7 +3641,6 @@ public static final class net.corda.client.jackson.StringToMethodCallParser$Unpa
 ##
 public final class net.corda.testing.driver.Driver extends java.lang.Object
   public static final Object driver(net.corda.testing.driver.DriverParameters, kotlin.jvm.functions.Function1)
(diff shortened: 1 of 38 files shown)


joeldudleyr3:
@m4ksio I don't like this change. It feels like a step backwards for CorDapp developers?

m4ksio:
@joeldudleyr3 We decided that its the right way to keep the backward and forward compatibility of test driver. Otherwise adding/removing any new parameters would result in compilation problem, we would need to provide overloaded versions - think especially Java.

joeldudleyr3:
@m4ksio Removing parameters is an API break, which is what we're not doing except in exceptional circumstances anyway. Adding parameters would be fine - we'd just provide overloaded constructors, as you say.

shamsasari:
Yes removing parameters is backwards incompatible.

shamsasari:
The problem with the driver is the large number of parameter it can take. This isn't an issue in Kotlin due to the default values but it's messy and intractable in Java. Why bother adding extra overloads for any new parameters when we already have `DriverParameters`. Thus it's easier to us if we just have this `driver` entrypoint. Slightly inconvenient for Kotlin users, but doesn't affect Java users who are the majority.

joeldudleyr3:
RGB’s mantra has always been that we should make Corda easy for our users rather than ourselves. This change violates that mantra.

joeldudleyr3:
@gendal I spoke to Maks and Shams about this. We decided to drop the convenience for Kotlin users *for now*, in the knowledge that we can easily go back and re-add this API in a future release. We can reintroduce it at a later date.

PR #2522, opened 2018-02-12
CORDA-939 Remove sslConfiguration from public constructor of CordaRPCClient

anthonykr3:
Remove the sslConfiguration paraemter from the public constructor of CordaRPCClient

Have created an internal factory method for initialising the sslConfiguration - and created an internal utility in net.corda.clienty.rpc.internal.CordaRPCClientUtils for other corda modules to get instances of CordaRPCClient. I considered making the constructor private but it seemed like it was used in so many places this broke the least things.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -3918,8 +3918,20 @@ public class net.corda.testing.node.MockNetwork extends java.lang.Object
   public <init>(List, net.corda.testing.node.MockNetworkParameters)
   public <init>(List, net.corda.testing.node.MockNetworkParameters, boolean, boolean, net.corda.testing.node.InMemoryMessagingNetwork$ServicePeerAllocationStrategy, boolean, List, int)
   @org.jetbrains.annotations.NotNull public final java.nio.file.Path baseDirectory(int)
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createNode()
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createNode(net.corda.core.identity.CordaX500Name)
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createNode(net.corda.core.identity.CordaX500Name, Integer)
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createNode(net.corda.core.identity.CordaX500Name, Integer, java.math.BigInteger)
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createNode(net.corda.core.identity.CordaX500Name, Integer, java.math.BigInteger, kotlin.jvm.functions.Function1)
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createNode(net.corda.core.identity.CordaX500Name, Integer, java.math.BigInteger, kotlin.jvm.functions.Function1, net.corda.node.VersionInfo)
   @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createNode(net.corda.testing.node.MockNodeParameters)
   @org.jetbrains.annotations.NotNull public final net.corda.testing.node.StartedMockNode createPartyNode(net.corda.core.identity.CordaX500Name)
+  @org.jetbrains.annotations.NotNull public final net.corda.testing.node.UnstartedMockNode createUnstartedNode()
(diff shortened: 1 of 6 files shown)


client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -72,9 +72,20 @@ data class CordaRPCClientConfiguration(val connectionMaxRetryInterval: Duration)
  */
 class CordaRPCClient @JvmOverloads constructor(
         hostAndPort: NetworkHostAndPort,
-        configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.DEFAULT,
-        sslConfiguration: SSLConfiguration? = null
+        configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.DEFAULT
 ) {
+    companion object {
+        internal fun getClientWithSsl(
```
shamsasari: `createWithSsl`?

client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -72,9 +72,20 @@ data class CordaRPCClientConfiguration(val connectionMaxRetryInterval: Duration)
...
+                hostAndPort: NetworkHostAndPort,
+                configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.DEFAULT,
+                sslConfiguration: SSLConfiguration? = null
+        ) : CordaRPCClient{
+            return CordaRPCClient(hostAndPort, configuration).apply {
+                this.sslConfiguration = sslConfiguration
```
shamsasari: I think we can avoid this by having the default c'tor be private and which takes in this as a parameter and then the current default c'tor becomes a secondary c'tor.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/CordaRPCClientUtils.kt
```diff
@@ -0,0 +1,13 @@
+package net.corda.client.rpc.internal
+
+import net.corda.client.rpc.CordaRPCClient
+import net.corda.client.rpc.CordaRPCClientConfiguration
+import net.corda.core.utilities.NetworkHostAndPort
+import net.corda.nodeapi.internal.config.SSLConfiguration
+
+/** Utility which exposes the internal Corda RPC constructor to other internal Corda components */
+fun getCordaRPCClientWithSsl(
```
shamsasari: The double-internal trick!

.ci/api-current.txt
```diff
@@ -4166,10 +4180,12 @@ public final class net.corda.testing.node.User extends java.lang.Object
 public final class net.corda.client.rpc.CordaRPCClient extends java.lang.Object
   public <init>(net.corda.core.utilities.NetworkHostAndPort)
   public <init>(net.corda.core.utilities.NetworkHostAndPort, net.corda.client.rpc.CordaRPCClientConfiguration)
-  public <init>(net.corda.core.utilities.NetworkHostAndPort, net.corda.client.rpc.CordaRPCClientConfiguration, net.corda.nodeapi.internal.config.SSLConfiguration)
```
anthonykr3: This is actually a backards compatibility change

client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -70,11 +70,24 @@ data class CordaRPCClientConfiguration(val connectionMaxRetryInterval: Duration)
  * @param configuration An optional configuration used to tweak client behaviour.
  * @param sslConfiguration An optional [SSLConfiguration] used to enable secure communication with the server.
  */
-class CordaRPCClient @JvmOverloads constructor(
+class CordaRPCClient private constructor(
         hostAndPort: NetworkHostAndPort,
         configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.DEFAULT,
-        sslConfiguration: SSLConfiguration? = null
+        private var sslConfiguration: SSLConfiguration? = null
```
shamsasari: This doesn't need to be a property, i.e. no need for `private var`

.ci/api-current.txt
shamsasari: You mean backwards incompatible? Thankfully it's not when compared against V2: https://github.com/corda/corda/blob/release-V2/.ci/api-current.txt#L3233

anthonykr3: Cracking

PR #2527, opened 2018-02-13
CORDA-1010: Send a request signature in addition to a transaction to the notary

adagys:
I'm still working on tests, but this is ready for a first look. I made both validating and non-validating notaries enforce the notarisation request signature check.

More details are under the Jira ticked and NotarisationRequest api docs.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -1128,7 +1128,8 @@ public final class net.corda.core.flows.ContractUpgradeFlow extends java.lang.Ob
   public <init>(net.corda.core.contracts.StateAndRef, Class)
   @co.paralleluniverse.fibers.Suspendable @org.jetbrains.annotations.NotNull protected net.corda.core.flows.AbstractStateReplacementFlow$UpgradeTx assembleTx()
 ##
-public abstract class net.corda.core.flows.DataVendingFlow extends net.corda.core.flows.FlowLogic
+public class net.corda.core.flows.DataVendingFlow extends net.corda.core.flows.FlowLogic
+  public <init>(net.corda.core.flows.FlowSession, Object)
   @co.paralleluniverse.fibers.Suspendable @org.jetbrains.annotations.Nullable public Void call()
   @org.jetbrains.annotations.NotNull public final net.corda.core.flows.FlowSession getOtherSideSession()
   @org.jetbrains.annotations.NotNull public final Object getPayload()
@@ -1322,11 +1323,11 @@ public @interface net.corda.core.flows.InitiatingFlow
   @org.jetbrains.annotations.NotNull public String toString()
 ##
 @net.corda.core.serialization.CordaSerializable public static final class net.corda.core.flows.NotaryError$General extends net.corda.core.flows.NotaryError
-  public <init>(String)
-  @org.jetbrains.annotations.NotNull public final String component1()
-  @org.jetbrains.annotations.NotNull public final net.corda.core.flows.NotaryError$General copy(String)
+  public <init>(Throwable)
+  @org.jetbrains.annotations.NotNull public final Throwable component1()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.flows.NotaryError$General copy(Throwable)
   public boolean equals(Object)
-  @org.jetbrains.annotations.NotNull public final String getCause()
+  @org.jetbrains.annotations.NotNull public final Throwable getCause()
   public int hashCode()
   @org.jetbrains.annotations.NotNull public String toString()
 ##
@@ -2923,7 +2924,7 @@ public static final class net.corda.core.serialization.SingletonSerializationTok
   @org.jetbrains.annotations.NotNull public final net.corda.core.crypto.SecureHash getId()
   @org.jetbrains.annotations.NotNull public final String getReason()
 ##
-@net.corda.core.DoNotImplement public abstract class net.corda.core.transactions.CoreTransaction extends net.corda.core.transactions.BaseTransaction
+@net.corda.core.serialization.CordaSerializable @net.corda.core.DoNotImplement public abstract class net.corda.core.transactions.CoreTransaction extends net.corda.core.transactions.BaseTransaction
   public <init>()
   @org.jetbrains.annotations.NotNull public abstract List getInputs()
 ##
@@ -3165,7 +3166,7 @@ public class net.corda.core.transactions.TransactionBuilder extends java.lang.Ob
   public abstract void verifyRequiredSignatures()
   public abstract void verifySignaturesExcept(Collection)
 ##
-@net.corda.core.DoNotImplement public abstract class net.corda.core.transactions.TraversableTransaction extends net.corda.core.transactions.CoreTransaction
+@net.corda.core.serialization.CordaSerializable @net.corda.core.DoNotImplement public abstract class net.corda.core.transactions.TraversableTransaction extends net.corda.core.transactions.CoreTransaction
   public <init>(List)
   @org.jetbrains.annotations.NotNull public final List getAttachments()
   @org.jetbrains.annotations.NotNull public final List getAvailableComponentGroups()
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -1371,8 +1371,6 @@ public final class net.corda.core.flows.NotaryFlow extends java.lang.Object
   @co.paralleluniverse.fibers.Suspendable @org.jetbrains.annotations.NotNull public List call()
   @org.jetbrains.annotations.NotNull public net.corda.core.utilities.ProgressTracker getProgressTracker()
   @co.paralleluniverse.fibers.Suspendable @org.jetbrains.annotations.NotNull protected final net.corda.core.utilities.UntrustworthyData notarise(net.corda.core.identity.Party)
-  @co.paralleluniverse.fibers.Suspendable @org.jetbrains.annotations.NotNull protected net.corda.core.utilities.UntrustworthyData sendAndReceiveNonValidating(net.corda.core.identity.Party, net.corda.core.flows.FlowSession)
-  @co.paralleluniverse.fibers.Suspendable @org.jetbrains.annotations.NotNull protected net.corda.core.utilities.UntrustworthyData sendAndReceiveValidating(net.corda.core.flows.FlowSession)
   @org.jetbrains.annotations.NotNull protected final List validateResponse(net.corda.core.utilities.UntrustworthyData, net.corda.core.identity.Party)
   public static final net.corda.core.flows.NotaryFlow$Client$Companion Companion
 ##
--- a/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
@@ -25,15 +25,14 @@ import net.corda.core.transactions.SignedTransaction
  *    to ensure the serialization does not get affected by the order.
  */
 @CordaSerializable
-class NotarisationRequest(statesToConsume: List<StateRef>, private val transactionId: SecureHash) {
+class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: SecureHash) {
     companion object {
         /** Sorts in ascending order first by transaction hash, then by output index. */
         private val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
     }
 
     /** States this request specifies to be consumed. Sorted to ensure the serialized form does not get affected by the state order. */
-    @Suppress("unused")
-    private val states = statesToConsume.sortedWith(stateRefComparator)
+    val states = statesToConsume.sortedWith(stateRefComparator)
 
     /** Verifies the signature against this notarisation request. Checks that the signature is issued by the right party. */
     fun verifySignature(requestSignature: NotarisationRequestSignature, intendedSigner: Party) {
--- a/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
@@ -8,6 +8,8 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.serialization.serialize
 import net.corda.core.transactions.CoreTransaction
 import net.corda.core.transactions.SignedTransaction
+import java.security.InvalidKeyException
+import java.security.SignatureException
 
 /**
  * A notarisation request specifies a list of states to consume and the id of the consuming transaction. Its primary
@@ -38,17 +40,29 @@ class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: Se
 
     /** Verifies the signature against this notarisation request. Checks that the signature is issued by the right party. */
     fun verifySignature(requestSignature: NotarisationRequestSignature, intendedSigner: Party) {
+        val signature = requestSignature.digitalSignature
+        if (intendedSigner.owningKey != signature.by) {
+            val errorMessage = "Expected a signature by ${intendedSigner.owningKey}, but received by ${signature.by}}"
+            throw NotaryException(NotaryError.RequestSignatureInvalid(IllegalArgumentException(errorMessage)))
+        }
+        // TODO: if requestSignature was generated over an old version of NotarisationRequest, we need to be able to
+        // reserialize it in that version to get the exact same bytes. Modify the serialization logic once that's
+        // available.
+        val expectedSignedBytes = this.serialize().bytes
+        verifyCorrectBytesSigned(signature, expectedSignedBytes)
+    }
+
+    private fun verifyCorrectBytesSigned(signature: DigitalSignature.WithKey, bytes: ByteArray) {
         try {
-            val signature = requestSignature.digitalSignature
-            require(intendedSigner.owningKey == signature.by) { "Notarisation request for $transactionId not signed by the requesting party" }
-            // TODO: if requestSignature was generated over an old version of NotarisationRequest, we need to be able to
-            // reserialize it in that version to get the exact same bytes. Modify the serialization logic once that's
-            // available.
-            val expectedSignedBytes = this.serialize().bytes
-            signature.verify(expectedSignedBytes)
+            signature.verify(bytes)
         } catch (e: Exception) {
-            val error = NotaryError.RequestSignatureInvalid(e)
-            throw NotaryException(error)
+            when (e) {
+                is InvalidKeyException, is SignatureException -> {
+                    val error = NotaryError.RequestSignatureInvalid(e)
+                    throw NotaryException(error)
+                }
+                else -> throw e
+            }
         }
     }
 }
--- a/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
@@ -84,10 +84,18 @@ data class NotarisationRequestSignature(val digitalSignature: DigitalSignature.W
 data class NotarisationPayload(val transaction: Any, val requestSignature: NotarisationRequestSignature) {
     init {
         require(transaction is SignedTransaction || transaction is CoreTransaction) {
-            "Unsupported transaction type in the notarisation paylaod: ${transaction.javaClass.simpleName}"
+            "Unsupported transaction type in the notarisation payload: ${transaction.javaClass.simpleName}"
         }
     }
 
+    /**
+     * A helper for automatically casting the underlying [transaction] payload to a [SignedTransaction].
+     * Should only be used by validating notaries.
+     */
     val signedTransaction get() = transaction as SignedTransaction
+    /**
+     * A helper for automatically casting the underlying [transaction] payload to a [CoreTransaction].
+     * Should only be used by non-validating notaries.
+     */
     val coreTransaction get() = transaction as CoreTransaction
 }
--- a/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
@@ -31,8 +31,10 @@ class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: Se
         private val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
     }
 
+    private val states = statesToConsume.sortedWith(stateRefComparator)
+
     /** States this request specifies to be consumed. Sorted to ensure the serialized form does not get affected by the state order. */
-    val states = statesToConsume.sortedWith(stateRefComparator)
+    fun getStatesToConsume() = states // Getter required for AMQP serialization
 
     /** Verifies the signature against this notarisation request. Checks that the signature is issued by the right party. */
     fun verifySignature(requestSignature: NotarisationRequestSignature, intendedSigner: Party) {
--- a/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
@@ -1,0 +1,78 @@
+package net.corda.core.flows
+
+import net.corda.core.contracts.StateRef
+import net.corda.core.crypto.DigitalSignature
+import net.corda.core.crypto.SecureHash
+import net.corda.core.identity.Party
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.serialization.serialize
+import net.corda.core.transactions.CoreTransaction
+import net.corda.core.transactions.SignedTransaction
+
+/**
+ * A notarisation request specifies a list of states to consume and the id of the consuming transaction. Its primary
+ * purpose is for notarisation traceability – a signature over the notarisation request, [NotarisationRequestSignature],
+ * allows a notary to prove that a certain party requested the consumption of a particular state.
+ *
+ * While the signature must be retained, the notarisation request does not need to be transferred or stored anywhere - it
+ * can be built from a [SignedTransaction] or a [CoreTransaction]. The notary can recompute it from the committed states index.
+ *
+ * In case there is a need to prove that a party spent a particular state, the notary will:
+ * 1) Locate the consuming transaction id in the index, along with all other states consumed in the same transaction.
+ * 2) Build a [NotarisationRequest].
+ * 3) Locate the [NotarisationRequestSignature] for the transaction id. The signature will contain the signing public key.
+ * 4) Demonstrate the signature verifies against the serialized request. The provided states are always sorted internally,
+ *    to ensure the serialization does not get affected by the order.
+ */
+@CordaSerializable
+class NotarisationRequest(statesToConsume: List<StateRef>, private val transactionId: SecureHash) {
+    companion object {
+        /** Sorts in ascending order first by transaction hash, then by output index. */
+        private val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
+    }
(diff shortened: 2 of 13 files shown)


adagys:
I decided to descope the actual logging of notarisation requests - it requires some delicate notary backend api changes. We can implement that later as it doesn't affect wire stability.

kchalkias:
do we need any unit tests for this new functionality?

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -162,6 +163,12 @@ class NotaryFlow {
         @Suspendable
         abstract fun receiveAndVerifyTx(): TransactionParts
 
+        protected fun validateRequest(request: NotarisationRequest, signature: NotarisationRequestSignature) {
+            val requestingParty = otherSideSession.counterparty
+            request.verifySignature(signature, requestingParty)
+            // TODO: persist the signature for traceability
```
kchalkias: fullstop

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,53 @@
...
+        // Sorts in ascending order first by transaction hash, then by output index.
+        val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
+        val sortedInputs = inputStates.sortedWith(stateRefComparator)
+        return NotarisationRequest(sortedInputs, txId).serialize().bytes
+    }
+}
+
+/** A wrapper around a digital signature used for notarisation requests. */
+@CordaSerializable
+data class NotarisationRequestSignature(val digitalSignature: DigitalSignature.WithKey)
```
kchalkias: Why do we need this wrapper?

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,53 @@
...
+    fun verifySignature(requestSignature: NotarisationRequestSignature, intendedSigner: Party) {
+        val signature = requestSignature.digitalSignature
+        require(intendedSigner.owningKey == signature.by) { "Notarisation request for $transactionId not signed by the requesting party" }
+        val expectedSignedBytes = generateBytesToSign(inputStates, transactionId)
+        signature.verify(expectedSignedBytes)
+    }
+
+    private fun generateBytesToSign(inputStates: List<StateRef>, txId: SecureHash): ByteArray {
+        // Sorts in ascending order first by transaction hash, then by output index.
+        val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
```
kchalkias: It might be a bit faster (although conceptually not elegant) if we first compare indices then hashes.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
adagys: We don't strictly need it, but it makes the code easier to understand, and we also might want to introduce some additional metadata for the signature

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
adagys: I don't think it matters much, we're generally only talking about 10s or 100s of input states.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
kchalkias: +1 for being proactive w.r.t. extra metadata in the future. It would be much more difficult to update `DigitalSignature`.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+import net.corda.core.serialization.serialize
+import net.corda.core.transactions.CoreTransaction
+
+/**
+ * A notarisation request specifies a list of states to consume and the id of the consuming transaction. Its primary
+ * purpose is for notarisation traceability – a signature over the notarisation request, [NotarisationRequestSignature],
+ * allows a notary to prove that a certain party requested the consumption of a particular state.
+ *
+ * While the signature must be retained, the notarisation request does not need to be transferred or stored anywhere - it
+ * can be build from a [SignedTransaction] or a [CoreTransaction]. The notary can recompute it from the committed states index.
```
shamsasari: typo: built

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * In case of there is a need to prove that a party spent a particular state, the notary will:
```
shamsasari: Remove "of"

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+ * 2) Build a [NotarisationRequest].
+ * 3) Locate the [NotarisationRequestSignature] for the transaction id. The signature will contain the signing public key.
+ * 4) Demonstrate the signature verifies against the request.
+ */
+@CordaSerializable
+data class NotarisationRequest(val inputStates: List<StateRef>, val transactionId: SecureHash) {
+    fun generateSignature(serviceHub: ServiceHub): NotarisationRequestSignature {
+        val bytesToSign = generateBytesToSign(inputStates, transactionId)
+        val signature = with(serviceHub) {
+            val myLegalIdentity = serviceHub.myInfo.legalIdentitiesAndCerts.first().owningKey
```
shamsasari: You don't need to refer to `serviceHub` here

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+    fun verifySignature(requestSignature: NotarisationRequestSignature, intendedSigner: Party) {
+        val signature = requestSignature.digitalSignature
+        require(intendedSigner.owningKey == signature.by) { "Notarisation request for $transactionId not signed by the requesting party" }
+        val expectedSignedBytes = generateBytesToSign(inputStates, transactionId)
+        signature.verify(expectedSignedBytes)
+    }
+
+    private fun generateBytesToSign(inputStates: List<StateRef>, txId: SecureHash): ByteArray {
+        // Sorts in ascending order first by transaction hash, then by output index.
+        val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
```
shamsasari: Perhaps put this in the companion object.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -162,6 +167,12 @@ class NotaryFlow {
         @Suspendable
         abstract fun receiveAndVerifyTx(): TransactionParts
 
+        protected fun validateRequest(request: NotarisationRequest, signature: NotarisationRequestSignature) {
```
shamsasari: `validateNonValidatingRequest`

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+ * can be build from a [SignedTransaction] or a [CoreTransaction]. The notary can recompute it from the committed states index.
+ *
+ * In case of there is a need to prove that a party spent a particular state, the notary will:
+ * 1) Locate the consuming transaction id in the index, along with all other states consumed in the same transaction.
+ * 2) Build a [NotarisationRequest].
+ * 3) Locate the [NotarisationRequestSignature] for the transaction id. The signature will contain the signing public key.
+ * 4) Demonstrate the signature verifies against the request.
+ */
+@CordaSerializable
+data class NotarisationRequest(val inputStates: List<StateRef>, val transactionId: SecureHash) {
```
shamsasari: This can't be internal as it's exposed by `validateRequest`

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
shamsasari: This needs KDocs

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+        // Sorts in ascending order first by transaction hash, then by output index.
+        val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
+        val sortedInputs = inputStates.sortedWith(stateRefComparator)
+        return NotarisationRequest(sortedInputs, txId).serialize().bytes
+    }
+}
+
+/** A wrapper around a digital signature used for notarisation requests. */
+@CordaSerializable
+data class NotarisationRequestSignature(val digitalSignature: DigitalSignature.WithKey)
```
shamsasari: ditto

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+@CordaSerializable
+data class NotarisationPayload(val transaction: CoreTransaction, val requestSignature: NotarisationRequestSignature)
```
shamsasari: This seems to be the only thing in this file that should be internal.

node/src/main/kotlin/net/corda/node/services/transactions/NonValidatingNotaryFlow.kt
```diff
@@ -21,22 +24,29 @@ class NonValidatingNotaryFlow(otherSideSession: FlowSession, service: TrustedAut
...
-                    it.checkAllComponentsVisible(ComponentGroupEnum.INPUTS_GROUP)
-                    it.checkAllComponentsVisible(ComponentGroupEnum.TIMEWINDOW_GROUP)
-                    val notary = it.notary
-                    TransactionParts(it.id, it.inputs, it.timeWindow, notary)
-                }
-                is NotaryChangeWireTransaction -> TransactionParts(it.id, it.inputs, null, it.notary)
-                else -> {
-                    throw IllegalArgumentException("Received unexpected transaction type: ${it::class.java.simpleName}," +
-                            "expected either ${FilteredTransaction::class.java.simpleName} or ${NotaryChangeWireTransaction::class.java.simpleName}")
+        return otherSideSession.receive<NotarisationPayload>().unwrap { (transaction, requestSignature) ->
```
shamsasari: The non-validating case must always have this message flow so I wonder if `validateRequest` shouldn't actually be `receiveNotarisationPayload`.

node/src/main/kotlin/net/corda/node/services/transactions/BFTNonValidatingNotaryService.kt
```diff
@@ -74,18 +80,24 @@ class BFTNonValidatingNotaryService(
     private class ServiceFlow(val otherSideSession: FlowSession, val service: BFTNonValidatingNotaryService) : FlowLogic<Void?>() {
         @Suspendable
         override fun call(): Void? {
-            val stx = otherSideSession.receive<FilteredTransaction>().unwrap { it }
-            val signatures = commit(stx)
+            val tx = otherSideSession.receive<NotarisationPayload>().unwrap { (transaction, requestSignature) ->
+                val request = NotarisationRequest(transaction.inputs, transaction.id)
+                val requestingParty = otherSideSession.counterparty
+                request.verifySignature(requestSignature, requestingParty)
```
shamsasari: Not using `validateRequest`?

.ci/api-current.txt
```diff
@@ -1324,11 +1325,11 @@ public @interface net.corda.core.flows.InitiatingFlow
   @org.jetbrains.annotations.NotNull public String toString()
 ##
 @net.corda.core.serialization.CordaSerializable public static final class net.corda.core.flows.NotaryError$General extends net.corda.core.flows.NotaryError
```
adagys: this class has only been introduced recently, and wasn't in the Corda 2.0 release

adagys:
Ready for another look

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -210,7 +217,11 @@ sealed class NotaryError {
...
     object WrongNotary : NotaryError()
 
-    data class General(val cause: String): NotaryError() {
-        override fun toString() = cause
+    data class General(val cause: Throwable) : NotaryError() {
+        override fun toString() = cause.toString()
+    }
+
+    data class RequestSignatureInvalid(val cause: Throwable) : NotaryError() {
+        override fun toString() = "Request signature invalid: " + cause.toString()
```
kchalkias: better use String template $

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,83 @@
...
+        // Sorts in ascending order first by transaction hash, then by output index.
+        private val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
+
+        fun generateBytesToSign(inputStates: List<StateRef>, txId: SecureHash): ByteArray {
+            val sortedInputs = inputStates.sortedWith(stateRefComparator)
+            return NotarisationRequest(sortedInputs, txId).serialize().bytes
+        }
+    }
+
+    fun generateSignature(serviceHub: ServiceHub): NotarisationRequestSignature {
```
kchalkias: add kdoc on all functions in this file.

kchalkias: `@Throws(InvalidKeyException::class, SignatureException::class)`.
We should add it to `KeyManagemenentService` as well.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,83 @@
...
+ * 3) Locate the [NotarisationRequestSignature] for the transaction id. The signature will contain the signing public key.
+ * 4) Demonstrate the signature verifies against the request.
+ */
+@CordaSerializable
+data class NotarisationRequest(val inputStates: List<StateRef>, val transactionId: SecureHash) {
+    companion object {
+        // Sorts in ascending order first by transaction hash, then by output index.
+        private val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
+
+        fun generateBytesToSign(inputStates: List<StateRef>, txId: SecureHash): ByteArray {
```
kchalkias: This is the most important kdoc, as we need to describe the logic of what is actually signed (which is the evidence we are looking for).

node/src/main/kotlin/net/corda/node/services/transactions/BFTNonValidatingNotaryService.kt
```diff
@@ -132,28 +138,34 @@ class BFTNonValidatingNotaryService(
...
                 log.debug { "Error processing transaction: ${e.error}" }
                 BFTSMaRt.ReplicaResponse.Error(e.error)
             }
         }
 
+        private fun verifyRequest(commitRequest: BFTSMaRt.CommitRequest) {
+            val transaction = commitRequest.payload.coreTransaction
+            val notarisationRequest = NotarisationRequest(transaction.inputs, transaction.id)
+            notarisationRequest.verifySignature(commitRequest.payload.requestSignature, commitRequest.callerIdentity)
+            // TODO: persist the signature for traceability.
```
kchalkias: So we are not storing this? I thought this was the aim of this PR.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -78,10 +82,11 @@ class NotaryFlow {
         protected fun notarise(notaryParty: Party): UntrustworthyData<List<TransactionSignature>> {
```
shamsasari: All these protected methods need KDocs

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
shamsasari: Yes, and you don't need to call `toString()` on `cause`

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
shamsasari: It's not so important as this method is internal and will not be visible to Java code (which is where the `@Throws` is useful).

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
shamsasari: This is the only member on this file now that is exposed in the public API and so will need to be moved out of here.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -92,19 +97,20 @@ class NotaryFlow {
         }
 
         @Suspendable
-        protected open fun sendAndReceiveValidating(session: FlowSession): UntrustworthyData<List<TransactionSignature>> {
-            subFlow(SendTransactionWithRetry(session, stx))
+        protected open fun sendAndReceiveValidating(session: FlowSession, requestSignature: NotarisationRequestSignature): UntrustworthyData<List<TransactionSignature>> {
```
shamsasari: All the components to this method are either internal or private so I'm not sure it makes sense for this method to be open. An app developer can't override this without needing access to NotarisationPayload and NotarySendTransactionFlow.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
shamsasari: ^ This comment is no longer valid.

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -13,6 +13,7 @@ import net.corda.core.utilities.contextLogger
 import org.slf4j.Logger
 import java.security.PublicKey
 import java.time.Clock
+import javax.naming.ServiceUnavailableException
```
shamsasari: This exception is from JNDI and not relevant to notarisation.

node/src/main/kotlin/net/corda/node/services/transactions/NonValidatingNotaryFlow.kt
```diff
@@ -21,22 +25,30 @@ class NonValidatingNotaryFlow(otherSideSession: FlowSession, service: TrustedAut
...
-                    it.checkAllComponentsVisible(ComponentGroupEnum.INPUTS_GROUP)
-                    it.checkAllComponentsVisible(ComponentGroupEnum.TIMEWINDOW_GROUP)
-                    val notary = it.notary
-                    TransactionParts(it.id, it.inputs, it.timeWindow, notary)
-                }
-                is NotaryChangeWireTransaction -> TransactionParts(it.id, it.inputs, null, it.notary)
-                else -> {
-                    throw IllegalArgumentException("Received unexpected transaction type: ${it::class.java.simpleName}," +
-                            "expected either ${FilteredTransaction::class.java.simpleName} or ${NotaryChangeWireTransaction::class.java.simpleName}")
+        return otherSideSession.receive<NotarisationPayload>().unwrap { payload ->
```
shamsasari: Why can't this logic of receiving a `NotarisationPayload` and processing it into `validateRequest` be a helper method in `core`? You can't write your own non-validating notary service without doing these exact steps, right?

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
```diff
@@ -0,0 +1,83 @@
...
+@CordaSerializable
+data class NotarisationPayload(val transaction: Any, val requestSignature: NotarisationRequestSignature) {
+    init {
+        require(transaction is SignedTransaction || transaction is CoreTransaction)
+    }
+    val signedTransaction get() = transaction as SignedTransaction
+    val coreTransaction get() = transaction as CoreTransaction
+}
+
+fun NotaryFlow.Service.validateRequest(request: NotarisationRequest, signature: NotarisationRequestSignature) {
```
shamsasari: I see that this method is used in `MyValidatingNotaryFlow` as part of the notary demo, which implies that this method is actually a public API method right? If so it needs to be moved out of this internal package.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
kchalkias: Ok, but we should add it to kdoc, right? for the developer to handle it if required it.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
adagys: The next piece of work is to enable persisting of the signature, which will require modifying or creating new NotaryService/Uniqueness provider interfaces, so this method is likely to change or move somewhere else, that's why I don't want to make it public. Yes, it will affect custom notary implementations, but I don't think we can guarantee API stability for custom notaries yet.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
shamsasari: Documentation for our purposes? Yes it makes sense I guess.

node/src/main/kotlin/net/corda/node/services/transactions/NonValidatingNotaryFlow.kt
adagys: the processing and persisting of signature will likely require access to the notary service object future, so having it as a helper method in core wouldn't work

node/src/main/kotlin/net/corda/node/services/transactions/BFTNonValidatingNotaryService.kt
adagys: It was, but the storage part got too complex so I descoped it. This piece of work is only for wire stability

node/src/main/kotlin/net/corda/node/services/transactions/BFTNonValidatingNotaryService.kt
adagys: The BFT notary doesn't work with the existing notary interfaces, so we end up with some code duplication. In this case it might not be an issue, as the BFT notary might require different handling of the notarisation request signature.

core/src/main/kotlin/net/corda/core/internal/NotarisationRequest.kt
adagys: I moved this to an internal method, as its signature might change in the future

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
```diff
@@ -0,0 +1,69 @@
...
+ * can be built from a [SignedTransaction] or a [CoreTransaction]. The notary can recompute it from the committed states index.
+ *
+ * In case there is a need to prove that a party spent a particular state, the notary will:
+ * 1) Locate the consuming transaction id in the index, along with all other states consumed in the same transaction.
+ * 2) Build a [NotarisationRequest].
+ * 3) Locate the [NotarisationRequestSignature] for the transaction id. The signature will contain the signing public key.
+ * 4) Demonstrate the signature verifies against the request.
+ */
+@CordaSerializable
+class NotarisationRequest(statesToConsume: List<StateRef>, private val transactionId: SecureHash) {
```
fenryka: we don't *love* private vals as our support for serialising this form of class is "to be depreicated" at some point, you can add a getter for transactionId

adagys: Ok, it's not really an issue if it's public

adagys:
Comments addressed, this is ready for another look

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * In case there is a need to prove that a party spent a particular state, the notary will:
+ * 1) Locate the consuming transaction id in the index, along with all other states consumed in the same transaction.
+ * 2) Build a [NotarisationRequest].
+ * 3) Locate the [NotarisationRequestSignature] for the transaction id. The signature will contain the signing public key.
+ * 4) Demonstrate the signature verifies against the serialized request. The provided states are always sorted internally,
+ *    to ensure the serialization does not get affected by the order.
+ */
+@CordaSerializable
+class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: SecureHash) {
```
fenryka: statesToConsume needs a public getter for AMQP serialisation to work that links it to states

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
```diff
@@ -0,0 +1,79 @@
...
+    companion object {
+        /** Sorts in ascending order first by transaction hash, then by output index. */
+        private val stateRefComparator = compareBy<StateRef>({ it.txhash }, { it.index })
+    }
+
+    private val states = statesToConsume.sortedWith(stateRefComparator)
+
+    /** States this request specifies to be consumed. Sorted to ensure the serialized form does not get affected by the state order. */
+    fun getStatesToConsume() = states // Getter required for AMQP serialization
```
shamsasari: `val statesToConsume: List< StateRef > get() = states`

shamsasari: Also it's conventional for the backing field to be prefixed with an underscore, i.e.` _ statesToConsume`

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
```diff
@@ -0,0 +1,79 @@
...
+    fun verifySignature(requestSignature: NotarisationRequestSignature, intendedSigner: Party) {
+        try {
+            val signature = requestSignature.digitalSignature
+            require(intendedSigner.owningKey == signature.by) { "Notarisation request for $transactionId not signed by the requesting party" }
+            // TODO: if requestSignature was generated over an old version of NotarisationRequest, we need to be able to
+            // reserialize it in that version to get the exact same bytes. Modify the serialization logic once that's
+            // available.
+            val expectedSignedBytes = this.serialize().bytes
+            signature.verify(expectedSignedBytes)
+        } catch (e: Exception) {
```
shamsasari: This can potentially send to the other side an internal exception. Catch the exception you need, which is probably `SignatureException` and change the `require` above to explicitly throw a `NotaryException`.

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
```diff
@@ -0,0 +1,79 @@
...
+ */
+@CordaSerializable
+data class NotarisationPayload(val transaction: Any, val requestSignature: NotarisationRequestSignature) {
+    init {
+        require(transaction is SignedTransaction || transaction is CoreTransaction) {
+            "Unsupported transaction type in the notarisation paylaod: ${transaction.javaClass.simpleName}"
+        }
+    }
+
+    val signedTransaction get() = transaction as SignedTransaction
```
shamsasari: These need KDocs and the return type should probably be `SignedTransaction?` rather than throwing a ClassCastException.

PR #2443, opened 2018-01-31
Switch to using our own quasar fork with thread leak fix

cburlinchon:
Don't forget to publish the gradle plugins once merged

PR #2533, opened 2018-02-14
CORDA-939 Bump gradle plugin version

anthonykeenan:
Change the gradle plugin version to 3.0.7 due to changes to api-scanner


Clintonio (approved):
Note for anyone wondering why; this is because we both merged a 3.0.6 bump but @anthonykr3's changes are not represented in the 3.0.6 binaries.

Clintonio:
Will merge #2443 with the same plugin bump instead to bring the functionality over.

PR #2534, opened 2018-02-14
CORDA-939 Modify Api Scanner to check api for internal exposures

anthonykr3:
... and fix remaining existing internal api exposures

https://github.com/corda/corda/pull/2510


constants.properties
```diff
@@ -1,4 +1,4 @@
-gradlePluginsVersion=3.0.5
+gradlePluginsVersion=3.0.7
```
Clintonio: Can you bump this one more time? There is a 3.0.7 published already (don't use it, it's a dud release).

Clintonio: I believe it's the source of your TC issues.

fenryka:
@anthonykr3 did bumping as per @Clintonio 's suggestion help with this?

anthonykr3:
I’ve just done it now - we’ll see.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -110,6 +111,12 @@ class StartedMockNode private constructor(private val node: StartedNode<Internal
 
     /** Returns the currently live flows of type [flowClass], and their corresponding result future. */
     fun <F : FlowLogic<*>> findStateMachines(flowClass: Class<F>): List<Pair<F, CordaFuture<*>>> = node.smm.findStateMachines(flowClass)
+
+    fun <T> transaction(statement: () -> T): T {
```
mikehearn: New public API needs kdoc

finance/src/test/kotlin/net/corda/finance/flows/CashPaymentFlowTests.kt
```diff
@@ -21,16 +22,16 @@ import kotlin.test.assertEquals
...
-    private lateinit var bankOfCordaNode: StartedMockNode
+    private lateinit var bankOfCordaNode: StartedNode<MockNode>
     private lateinit var bankOfCorda: Party
-    private lateinit var aliceNode: StartedMockNode
+    private lateinit var aliceNode: StartedNode<MockNode>
 
     @Before
     fun start() {
-        mockNet = MockNetwork(servicePeerAllocationStrategy = RoundRobin(), cordappPackages = listOf("net.corda.finance.contracts.asset"))
+        mockNet = InternalMockNetwork(servicePeerAllocationStrategy = RoundRobin(), cordappPackages = listOf("net.corda.finance.contracts.asset"))
```
mikehearn: Why does the cash app need to use internal API?

mikehearn:
Thanks. I think cash contract was the only code I saw changed that didn't obviously need it, but I wonder if the other places you changed really need to use the internal API too? However those places are already internal themselves I guess, so it's no big deal.

PR #2540, opened 2018-02-15
Added "advertisedP2pAddress" in node.conf

AndreaBorzi:
This serves the purpose of splitting two different behaviour of the nodes:
- The p2p address to bind on (i.e. p2pAddress)
- The p2p address other nodes need to know to talk to you properly (i.e. advertisedP2pAddress).

The advertisedP2pAddress is what goes into the NodeInfo sent to the NetworkMap, p2pAddress is used to bind on a local interface instead.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
+++ b/node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
@@ -37,6 +37,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val bftSMaRt: BFTSMaRtConfiguration
     val notaryNodeAddress: NetworkHostAndPort?
     val notaryClusterAddresses: List<NetworkHostAndPort>
+    val advertisedP2pAddress: NetworkHostAndPort
 }
 
 data class FullNodeConfiguration(
@@ -58,6 +59,7 @@ data class FullNodeConfiguration(
         @OldConfig("artemisAddress")
         val p2pAddress: NetworkHostAndPort,
         val rpcAddress: NetworkHostAndPort?,
+        override val advertisedP2pAddress: NetworkHostAndPort,
         // TODO This field is slightly redundant as p2pAddress is sufficient to hold the address of the node's MQ broker.
         // Instead this should be a Boolean indicating whether that broker is an internal one started by the node or an external one
         val messagingServerAddress: NetworkHostAndPort?,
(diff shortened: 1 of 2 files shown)


mikehearn:
@mnesbit How does this change interact with the bridge/float work you're doing at the moment?

AndreaBorzi:
@mnesbit 

Citing the docs on p2pAddress

_In practice the ArtemisMQ messaging services bind to all local addresses on the specified port. However, note that the host is the included as the advertised entry in the NetworkMapService. **As a result the value listed here must be externally accessible when running nodes across a cluster of machines.**_

But on messagingServerAddress there's nothing which confirms that this would be the address published to the NetworkMap instead of p2pAddress. Can you confirm this?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -37,6 +37,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val bftSMaRt: BFTSMaRtConfiguration
     val notaryNodeAddress: NetworkHostAndPort?
     val notaryClusterAddresses: List<NetworkHostAndPort>
+    val advertisedP2pAddress: NetworkHostAndPort
```
mnesbit: I do understand your motivation for adding this, but the problem is we have quite a lot of build tooling that is not expecting this new property. There are also numerous mocks of the NodeConfiguration that need to implement this to stop the tests breaking. So I think getting this change to work is not at all straightforward.

AndreaBorzi: I see that a change on Node Configuration would cause your tests to fail, please note that the feature has been implemented for V2. It would be nice to have an implementation for V3 and onwards since this could be critical for production enviroments (i.e. I can't bind ports on every interface on a production machine, just to save a line of configuration).

I am not expecting this feature to be mergde/implemented right here and now, however it could be a must for many bank environments you are developing the platform for.

mnesbit: I do understand, having worked in such environments for many years. I am working on some aspects of this as part of our enterprise offering right now. Hopefully, I can filter down elements of this to open source too.

gendal:
Andrea - I just wanted to say thanks for the PR. Looks like it may be more complicated than we'd like to accept it as-is (I defer to Matthew on this!) but I really appreciate the contribution.

AndreaBorzi:
Thanks for your reviews, I close this Pull Request as you are working on the issue already.

PR #2538, opened 2018-02-14
ENT-1519: Ensure NodeInfo always has at least one address by checking in the c'tor

shamsasari:
Further, the look up of the node's own node-info from its database has been tightened to ensure there isn't more than one.

Also fixed some brittle code which was assuming exactly one address rather than at least one.

PR #2542, opened 2018-02-15
ENT-1519: Ensure NodeInfo always has at least one address by checking…

shamsasari:
… in the c'tor (#2538)

Further, the look up of the node's own node-info from its database has been tightened to ensure there isn't more than one.

Also fixed some brittle code which was assuming exactly one address rather than at least one.

(cherry picked from commit 2864ce1)

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,14 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
+
+* Added ``NetworkMapCache.getNodesByLegalName`` for querying nodes belonging to a distributed service such as a notary cluster
+  where they all share a common identity. ``NetworkMapCache.getNodeByLegalName`` has been tightened to throw if more than
+  one node with the legal name is found.
+
+* Per CorDapp configuration is now exposed. ``CordappContext`` now exposes a ``CordappConfig`` object that is populated
+  at CorDapp context creation time from a file source during runtime.
+
 * Introduced Flow Draining mode, in which a node continues executing existing flows, but does not start new. This is to support graceful node shutdown/restarts.
   In particular, when this mode is on, new flows through RPC will be rejected, scheduled flows will be ignored, and initial session messages will not be consumed.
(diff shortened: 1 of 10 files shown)


node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -153,8 +153,17 @@ open class PersistentNetworkMapCache(
         return null
     }
 
-    override fun getNodeByLegalName(name: CordaX500Name): NodeInfo? = getNodesByLegalName(name).firstOrNull()
+    override fun getNodeByLegalName(name: CordaX500Name): NodeInfo? {
+        val nodeInfos = getNodesByLegalName(name)
+        return when (nodeInfos.size) {
+            0 -> null
+            1 -> nodeInfos[0]
+            else -> throw IllegalArgumentException("More than one node found with legal name $name")
```
mikehearn: Sounds more like an ISE than an IAE, but if you define your own exception then the unit test gets simpler too (no need to test the message anymore).

docs/source/changelog.rst
```diff
@@ -6,6 +6,14 @@ from the previous milestone release.
 
 UNRELEASED
 ----------
+
+* Added ``NetworkMapCache.getNodesByLegalName`` for querying nodes belonging to a distributed service such as a notary cluster
```
mikehearn: Looks like this needs a rebase?

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
shamsasari: It's an error with the argument, not the state (db) as the name points to more than one node.

docs/source/changelog.rst
shamsasari: Just double-checked against master and it's fine.

PR #2543, opened 2018-02-15
DOCS: update AMQP serialization

fenryka:
Give an example of how to over come an issue @adagys was having

### DIFF
--- a/docs/source/serialization.rst
+++ b/docs/source/serialization.rst
@@ -399,10 +399,9 @@ Providing a public getter, as per the following example, is acceptable:
 Mismatched Class Properties / Constructor Parameters
 ````````````````````````````````````````````````````
 
-Consider an example where you wish to ensure a container property of a class is always sorted yet you wish
-to maintain immutability of the class.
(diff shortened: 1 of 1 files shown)


docs/source/serialization.rst
```diff
@@ -396,6 +396,52 @@ Providing a public getter, as per the following example, is acceptable:
           }
       }
 
+Mismatched Class Properties / Constructor Parameters
+````````````````````````````````````````````````````
+
+Consider an example where you wish to ensure a container property of a class is always sorted yet you wish
```
joeldudleyr3: "container property"? What's that?

docs/source/serialization.rst
```diff
@@ -396,6 +396,52 @@ Providing a public getter, as per the following example, is acceptable:
...
+to maintain immutability of the class.
+
+Consider this example:
```
joeldudleyr3: Duplication. Maybe change the first sentence above to "Consider the following example where..."

PR #2546, opened 2018-02-16
CORDA-1038 Update verifySignaturesExcept in api-transactions.rst

kchalkias:
Update [api-transactions.rst](https://docs.corda.net/head/api-transactions.html#verifying-the-transaction-s-signatures) to reflect new ``TransactionWithSignatures.verifySignaturesExcept`` which takes in a ``Collection`` of ``PublicKey``s.

### DIFF
--- a/docs/source/example-code/src/main/java/net/corda/docs/FlowCookbookJava.java
+++ b/docs/source/example-code/src/main/java/net/corda/docs/FlowCookbookJava.java
@@ -28,7 +28,6 @@ import java.security.GeneralSecurityException;
 import java.security.PublicKey;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


docs/source/example-code/src/main/java/net/corda/docs/FlowCookbookJava.java
```diff
@@ -540,12 +541,21 @@ protected void verifyDataRequest(@NotNull FetchDataFlow.Request.Data dataRequest
                 // DOCEND 35
 
                 // If the transaction is only partially signed, we have to pass in
-                // a list of the public keys corresponding to the missing
+                // a vararg of the public keys corresponding to the missing
```
joeldudleyr3: I don' think vararg is a concept that exists in Java. Maybe just say "...we have to pass in the public keys..."?

kchalkias: The concept exists in Java as well https://docs.oracle.com/javase/8/docs/technotes/guides/language/varargs.html

PR #2548, opened 2018-02-16
[CORDA-1016]: Fixed Demobench.

sollecitom:
[CORDA-1016]: RPC admin address is now correctly generated for nodes in Demobench,


sollecitom:
@adagys it is needed by nodes in case SSL is not enforced for RPC communication. We need to enforce SSL for the internal admin user on the Artemis broker, otherwise clients could manage queues they're not meant to; in case SSL is enforce, we can do that on the same port, but if no SSL is enforced, we require an additional port for the SSL binding.

sollecitom:
By the way, the test worked on my local, so I'm re-running it.

PR #2557, opened 2018-02-16
CORDA-1064: Fix a hang in AMQP protocol code

mnesbit:
When debugging code with active AMQP connections it is possible for the proton-j state machine to become hung and not drop the bridging socket. This is due to timeout events not flushing through the events on the executor. This change ensures the events leading to close are always run.


mikehearn:
Would this bug also be in 3.0? If so perhaps we should pull it across once merged.

mnesbit:
Yes it is in 3.0, which is why I am keen to get this in.

mnesbit:
Switch out Andras as reviewer, since he is on holiday.

PR #2563, opened 2018-02-17
Reversing the introduction of GlobalProperties (by CORDA-961)

shamsasari:
Added NetworkParameters to ServicesForResolution instead.

Also, replaced the maxTransactionSize parameter of the driver and MockNetwork with `NetworkParameters`.


core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -100,12 +100,23 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
      * @throws AttachmentResolutionException if a required attachment was not found using [resolveAttachment].
      * @throws TransactionResolutionException if an input was not found not using [resolveStateRef].
      */
+    @Deprecated("Use toLedgerTransaction(ServicesForTransaction) instead", replaceWith = ReplaceWith("toLedgerTransaction(ServicesForTransaction)"))
```
mikehearn: Hm, the ReplaceWith here won't work will it - is ServicesForTransaction a global singleton? If not then the replaced code will be broken and there isn't much reason to prefer it over doing things manually.

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -166,15 +146,15 @@ fun <A> driver(
...
+                    networkParameters = defaultParameters.networkParameters
             ),
             coerce = { it },
             dsl = dsl,
             initialiseSerialization = defaultParameters.initialiseSerialization
     )
 }
 
-/** Helper builder for configuring a [driver] from Java. */
+/** Builder for configuring a [driver] from Java. */
```
mikehearn: Rather than erase all the docs above, can we move them here so the javadocs for each value aren't lost?

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
shamsasari: Yep

testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
shamsasari: Yep

PR #2570, opened 2018-02-19
Changed the wither methods in DriverParameters and NodeParameters to …

shamsasari:
…begin with "with" and not "set"


shamsasari:
Moved this to master

PR #2581, opened 2018-02-20
CORDA-941: Instead of storing the network map contract implementation…

adagys:
… whitelist along with

the constraint, populate it when resolving inputs during WireTransaction -> LedgerTransaction
transition.


core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
```diff
@@ -27,18 +27,23 @@ data class HashAttachmentConstraint(val attachmentId: SecureHash) : AttachmentCo
...
  * See: [net.corda.core.node.NetworkParameters.whitelistedContractImplementations]
  * It allows for centralized control over the cordapps that can be used.
- *
- * @param whitelistedContractImplementations whitelisted attachment IDs by contract class name.
  */
-class WhitelistedByZoneAttachmentConstraint(private val whitelistedContractImplementations: Map<String, List<AttachmentId>>) : AttachmentConstraint {
+class WhitelistedByZoneAttachmentConstraint : AttachmentConstraint {
+    @Transient
+    /** Whitelisted attachment IDs by contract class name. */
+    private var whitelistedContractImplementations: Map<String, List<AttachmentId>>? = null
```
mikehearn: Why make only the setter public and not the getter? Can this be instead

`@Transient internal var whitelistedContractImplementations` ?

Also it can be made lateinit var, to avoid the nullability issue.

It just seems odd to write a manual public (and therefore part of the API) setter. If the internal keyword doesn't work, maybe use @CordaInternal and suppress the KDoc instead? We don't want this stuff to be public API, right?

Alternatives:

* Edit the isSatisfiedBy API and mark it as @CordaInternal internal to take it out of the API. It's technically a break, but I'll let it slide this time given the near-zero likelyhood of any app breaking.
* Handshake via a TLS slot that contains the parameters.

adagys: Yep I can hide it, or move it to an internal extension method. Just getting a feel right now whether we're happy with this approach conceptually.

I considered the first alternative, but that means the context where the constraints are verified has to have access to the network params, which requires wiring it up to LedgerTransaction & ContractUpgradeLedgerTransaction – seems a bit excessive.

Not sure I understand the second alternative (?)

adagys: Also the use of nullability is intentional, so we can throw a more meaningful exception if `isSatisfiedBy` is called before the whitelist is populated

sollecitom:
IMO we should pass that whitelist at method level. Implicit contracts and mutable variables are not easy to maintain.

Anyway, your call.

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
mikehearn: Hmm, another way to do it would be to add it to ConstraintAttachment. AFAIK it was added specifically for this PR. And looking at which, I'm not sure it's supposed to be a part of the public API at all. Can you try marking ConstraintAttachment as internal and renaming it to AttachmentForZoneWhitelistConstraint or something? AFAIK the only purpose is to tunnel info into the constraint implementation.

adagys: Ah, I see. That still means we have to wire up network parameters to `LedgerTransaction`s though

adagys: I coded it up, but I still think it's better not to touch the `LedgerTransaction`

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -39,7 +40,8 @@ data class LedgerTransaction(
         override val id: SecureHash,
         override val notary: Party?,
         val timeWindow: TimeWindow?,
-        val privacySalt: PrivacySalt
+        val privacySalt: PrivacySalt,
+        private val whitelistedContractImplementations: Map<String, List<AttachmentId>> = emptyMap()
```
mikehearn: That isn't going to work. This message is serialisable, and the c'tor is public API.

I would suggest we figure out a way to give LedgerTransaction access to the whole set of network parameters, with a hack to make the old overload default to the actual copy e.g. using a TLS slot.

adagys: One thing I'm slightly worried about is that we currently serialize `LedgerTransaction` and send it across to the external verifier. Having to serialize in the entire network parameters `just in case` some of the states might use the `WhitelistedByZone` constraint seems a bit excessive.
Re the TLS slot approach, that wouldn't be very different to `GlobalProperties` when it comes to tests etc ?

adagys:
Ready for another look

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -39,7 +41,8 @@ data class LedgerTransaction(
         override val id: SecureHash,
         override val notary: Party?,
         val timeWindow: TimeWindow?,
-        val privacySalt: PrivacySalt
+        val privacySalt: PrivacySalt,
+        private val networkParameters: NetworkParameters? = null
```
mikehearn: Why make it private?

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -88,7 +89,8 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
                 resolveIdentity = { services.identityService.partyFromKey(it) },
                 resolveAttachment = { services.attachments.openAttachment(it) },
                 resolveStateRef = { services.loadState(it) },
-                maxTransactionSize = services.networkParameters.maxTransactionSize
+                maxTransactionSize = services.networkParameters.maxTransactionSize,
```
mikehearn: maxTxSize isn't needed as a param here anymore is it, it can be subsumed by the other you've added (as that's an internal call)

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
adagys: It's technically not part of the transaction. Or do we expect contracts to be able to refer to network params?

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
adagys: good catch, fixed

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
mikehearn: Hm, OK, fair enough. We can look at exposing params to contracts later.

PR #2586, opened 2018-02-20
Updates classloader to handle classes that implement interfaces or subclass classes in other cordapps.

joeldudleyr3:
The old classloader couldn't handle cases such as the following:

1. X is defined in package 1 and implements Contract
2. Y is defined in package 2 and subclasses X

Y is a Contract, but because the classloader considered each CorDapp in isolation, it couldn't work this out.

Now, there is a single scan across all CorDapps at once, and then contracts, flows, etc. are assigned to their respective CorDapps based on their package names.

One potential issue is that if two CorDapps share a package name, all contracts, flows, etc. will be assigned to all of them. Is this an issue?

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
@@ -117,11 +117,9 @@ class CordappLoader private constructor(private val cordappJarPaths: List<Restri
                     .asSequence()
                     .map { path ->
                         if (path.protocol == "jar") {
-                            // When running tests from gradle this may be a corda module jar, so restrict to scanPackage:
                             RestrictedURL((path.openConnection() as JarURLConnection).jarFileURL, scanPackage)
                         } else {
-                            // No need to restrict as createDevCordappJar has already done that:
-                            RestrictedURL(createDevCordappJar(scanPackage, path, resource).toURL(), null)
+                            RestrictedURL(createDevCordappJar(scanPackage, path, resource).toURL(), scanPackage)
                         }
                     }
                     .toList()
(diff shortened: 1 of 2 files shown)


shamsasari:
Java 9 does not allow split packages, so when we eventually upgrade that scenario will not be possible.

joeldudleyr3:
Yes, it only appears to arise in one of the tests.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -190,18 +188,23 @@ class CordappLoader private constructor(private val cordappJarPaths: List<Restri
...
-                    findSchedulableFlows(scanResult),
-                    findServices(scanResult),
-                    findPlugins(it),
-                    findSerializers(scanResult),
-                    findCustomSchemas(scanResult),
-                    it.url)
+        return if (cordappJarPaths.isEmpty()) {
+            emptyList()
+        } else {
+            val scanResult = FastClasspathScanner().overrideClasspath(cordappJarPaths.map { it.url }).scan()
```
Clintonio: Can you split this into a new "scanAllCordapps" function? It'll likely get more augmentations in future and it'd be good to escapsulate that.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -258,12 +261,9 @@ class CordappLoader private constructor(private val cordappJarPaths: List<Restri
         return scanResult.getClassesWithSuperclass(MappedSchema::class).toSet()
     }
 
-    private val cachedScanResult = LRUMap<RestrictedURL, RestrictedScanResult>(1000)
-    private fun scanCordapp(cordappJarPath: RestrictedURL): RestrictedScanResult {
+    private fun scanCordapp(scanResult: ScanResult, cordappJarPath: RestrictedURL): RestrictedScanResult {
         logger.info("Scanning CorDapp in $cordappJarPath")
-        return cachedScanResult.computeIfAbsent(cordappJarPath, {
-            RestrictedScanResult(FastClasspathScanner().addClassLoader(appClassLoader).overrideClasspath(cordappJarPath.url).scan(), cordappJarPath.qualifiedNamePrefix)
```
Clintonio: One problem here is that eventually we're removing the cordapps from the root classloader, hence why this appClassLoader exists. Can you try to continue using the app classloader? From what I can see the issue was the fact it scanned in isolation and not the classloader.

joeldudleyr3: I removed it because the JavaDocs indicated that `addClassLoader` is ignored if `overrideClasspath` is specified. Is that not true?

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
joeldudleyr3: NP

PR #2587, opened 2018-02-20
CORDA-1092 secureRandomBytes should call getNextBytes, not generateSeed

kchalkias:
`secureRandomBytes` should call `getNextBytes()`. It affects `SecureHash.randomSHA256()` as well.


rick-r3:
@kchalkias I suspect if you rebase then the tests will pass

PR #2590, opened 2018-02-21
[CORDA-1059]: Fix Driver hanging on duplicate node names.

sollecitom:
Port of https://github.com/corda/corda/pull/2574.


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -101,6 +101,7 @@ class DriverDSLImpl(
     private val cordappPackages = extraCordappPackagesToScan + getCallerPackage()
     // Map from a nodes legal name to an observable emitting the number of nodes in its network map.
     private val countObservables = mutableMapOf<CordaX500Name, Observable<Int>>()
+    private val nodeNames = mutableSetOf<CordaX500Name>()
```
vkolomeyko: I know it is a bit late, but we could have used `java.util.concurrent.ConcurrentHashMap#newKeySet()` to avoid using `synchronized` keyword below.

sollecitom:
We can refactor it later on. Anyway, not a big deal.

PR #2592, opened 2018-02-21
[CORDA-1089]: Correctly display an error message on node startup if RPC settings are invalid

sollecitom:


sollecitom:
@fenryka the logic there is changed a bit, reflecting your point as well.

PR #2594, opened 2018-02-21
CORDA-1004 Quasar-friendly ThreadLocal solution

andr3ej:


core/src/test/kotlin/net/corda/core/flows/FastThreadLocalTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        init(3, ::Thread)
+        val threadLocal = object : FastThreadLocal<ExpensiveObj>() {
+            override fun initialValue() = ExpensiveObj()
+        }
+        assertEquals(0, runFibers(100, threadLocal::get))
+        assertEquals(100, expensiveObjCount.get())
+    }
+
+    @Test
+    fun `FastThreadLocal with FastThreadLocalThread is not fiber-local`() {
```
rick-r3: Okay, looks like we need to make all our Fiber scheduler threads inherit from `FastThreadLocalThread` if we want to use `FastThreadLocal` to avoid Quasar meddling.  We want our implementation to work inside and outside Fibers which this would then achieve.  I suppose we just need to have a quick think if there would be any reason not to do this @exFalso ?  It would also mean that we could use netty buffers correctly on Fibers, but with the downside that we might do so by accident and not notice.  We'd have to check if we suddenly saw netty stashing lots of things we were not so aware of.  There is the advantage that the internal map for these is somewhat explorable so maybe we could easily figure this out.

exFalso: I like this approach! :+1: 
I don't think it requires any `core` changes, it's fairly debuggable, and we'd have fine control over TLs vs FLs, at least the ones we create. In theory we just need to adjust the flow threadpool's factory.
If the netty-nature of these is an issue we can also just create our own version, shouldn't be too difficult.

rick-r3: Okay.  Cool.  @andr3ej Can you check to see if we change the thread factory in open source it's going to create a merge conflict?  I think the (flow) thread is part of the `AffinityExecutor` which I suspect also exists in R3 Corda and thus there's no conflict, meaning we should make the change in this PR.  We'll need to do something similar in R3 Corda as another PR if that `StateMachineManager` uses a different `Executor`, once this is merged across.

andr3ej: looks like there won't be any conflicts as the change is confined to AffinityExecutor which is still there in R3 Corda, but it will need its own PR for its custom SMM

PR #2600, opened 2018-02-21
[CORDA-1075]: Add 'bye' and 'exit' commands to Shell.

ags313:
Cosmetic, 'bye' and 'exit' in shell's 'help'.


sollecitom:
@fenryka , if you could.

PR #2603, opened 2018-02-21
CORDA-1030: waitForAllNodesToFinish can only occur in out-of-process …

shamsasari:
…mode


shamsasari:
Closing as we do want to support this in in-process mode.

PR #2614, opened 2018-02-22
CORDA-1096 - Performance when loading multiple states from the vault

blsemo:
Merge from per-release-v3

* Provide efficient `loadStates()` implementation
* Replace loops using `loadState` with calls to `loadStates`


core/src/main/kotlin/net/corda/core/node/services/TransactionStorage.kt
```diff
@@ -26,6 +24,15 @@ interface TransactionStorage : StateLoader {
         return stx.resolveBaseTransaction(this).outputs[stateRef.index]
     }
 
+    @Throws(TransactionResolutionException::class)
+    override fun loadStates(stateRefs: Set<StateRef>): Set<StateAndRef<ContractState>> {
+        return stateRefs.groupBy { it.txhash }.map {
```
thschroeter: what about `flatMapping` here instead of below?

PR #2615, opened 2018-02-22
[CORDA-2615]: Added description for configFile option in deployNodes Cordform Gradle task.

szymonsztuka:
Added description for `configFile` option in `deployNodes` Cordform Gradle task.
The option is only used when generating a node configuration and it's not present in node.conf (so it was not added to https://docs.corda.net/head/corda-configuration-file.html ).

https://github.com/corda/corda/pull/2615

### DIFF
--- a/docs/source/generating-a-node.rst
+++ b/docs/source/generating-a-node.rst
@@ -120,10 +120,10 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates 
         }
     }
 
-To extend node configuration beyond properties defined in the ``deployNodes`` task use ``configFile`` property with the path (relative or absolute) set to an additional configuration file.
+To extend node configuration beyond the properties defined in the ``deployNodes`` task use the ``configFile`` property with the path (relative or absolute) set to an additional configuration file.
 This file should follow the standard format, as per node.conf. The properties from this file will be appended to the generated node configuration.
-Additionaly, the path to the file can also be added while running the Gradle task via the ``-PconfigFile`` command line option. However, the same file will be applied to all nodes.
-Following the previous example `PartyB` node will have additional configuration options added from a file `none-b.conf`:
+The path to the file can also be added while running the Gradle task via the ``-PconfigFile`` command line option. However, the same file will be applied to all nodes.
+Following the previous example ``PartyB`` node will have additional configuration options added from a file ``none-b.conf``:
 
 .. sourcecode:: groovy
 
--- a/docs/source/generating-a-node.rst
+++ b/docs/source/generating-a-node.rst
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates 
         }
     }
 
+To extend node configuration beyond properties defined in ``deployNodes`` task use ``configFile`` property with the path (relative or absolute) to an additional configuration file.
+The file should has a standard format as node.conf one. The properties from that file will be appended as it is to the resulting node configuration.
+The path to the file can also be added while running Gradle task via ``-PconfigFile`` command line option, however the same file will be added to all nodes.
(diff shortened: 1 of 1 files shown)


docs/source/generating-a-node.rst
```diff
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates
         }
     }
 
+To extend node configuration beyond properties defined in ``deployNodes`` task use ``configFile`` property with the path (relative or absolute) to an additional configuration file.
```
fenryka: defined in *the*...

fenryka: with the path (...) *set* to an

docs/source/generating-a-node.rst
```diff
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates
...
+The file should has a standard format as node.conf one. The properties from that file will be appended as it is to the resulting node configuration.
```
fenryka: *This*. not The

*have* not has

-- This file's format should follow the same format as node.conf.

docs/source/generating-a-node.rst
```diff
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates
...
+The path to the file can also be added while running Gradle task via ``-PconfigFile`` command line option, however the same file will be added to all nodes.
```
fenryka: [full stop] However[comma]

fenryka: applied or added ?

szymonsztuka:
Wording fixed.

docs/source/generating-a-node.rst
```diff
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates
         }
     }
 
+To extend node configuration beyond properties defined in the ``deployNodes`` task use ``configFile`` property with the path (relative or absolute) set to an additional configuration file.
```
joeldudleyr3: "beyond the properties", "the ``configFile`` property".

Also, I think it would be clearer to say. "...use ``configFile`` property. This property should be set to the path (relative or absolute) of an additional configuration file."

docs/source/generating-a-node.rst
```diff
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates
...
+This file should follow the standard format, as per node.conf. The properties from this file will be appended to the generated node configuration.
```
joeldudleyr3: What standard format? Perhaps clearer to say "should be a plaintext file of ``key=value`` pairs." (if that's true, I don't know :) )

docs/source/generating-a-node.rst
```diff
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates
...
+Additionaly, the path to the file can also be added while running the Gradle task via the ``-PconfigFile`` command line option. However, the same file will be applied to all nodes.
```
joeldudleyr3: No need for "Additionally" if you're also using "also". Delete it.

docs/source/generating-a-node.rst
```diff
@@ -120,6 +120,24 @@ nodes. Here is an example ``Cordform`` task called ``deployNodes`` that creates
...
+Following the previous example `PartyB` node will have additional configuration options added from a file `none-b.conf`:
```
joeldudleyr3: Double-ticks around ``PartyB``.

Should be "``node.conf``" in double-ticks.

docs/source/generating-a-node.rst
szymonsztuka: No, it can be nested ones like e.g. ``sshd { port = 2222 } ``

PR #2616, opened 2018-02-23
Ignore tests in experimental/behave

tlil:
`behave` is not included in `pre-release-V3` - so no need for a fix on the pre-release branch.


experimental/behave/src/test/kotlin/net/corda/behave/service/SqlServerServiceTests.kt
```diff
@@ -2,12 +2,10 @@ package net.corda.behave.service
 
 import net.corda.behave.service.database.SqlServerService
 import org.assertj.core.api.Assertions.assertThat
-import org.junit.Ignore
 import org.junit.Test
 
 class SqlServerServiceTests {
 
-    @Ignore
```
josecoll: Did you mean to add this test into the mix?

tlil:
@josecoll it's just the composite ones - they're unused and not very well implemented.

experimental/behave/src/test/kotlin/net/corda/behave/service/SqlServerServiceTests.kt
tlil: Yep:

>  Only run unit tests for :experimental:behave if explicit 

So the tests will only run if you run ../../gradlew test from experimental/behave now, and not be included in the full project build.

tlil: Or that's what I'm hoping to achieve at least:

https://github.com/corda/corda/blob/9638c91a0f49d45e91ab6f2c8c595e85b759a12b/experimental/behave/build.gradle#L106-L109

josecoll: Ouch!
That is confusing for developers without an explicit note in the source code.
ie. a developer would need to look at the gradle file to understand why tests are not being run.

josecoll:
I actually found them quite useful (and would rather keep them with a sub-optimal implementation that remove them completely).

experimental/behave/src/test/kotlin/net/corda/behave/service/SqlServerServiceTests.kt
tlil: Confirmed in TeamCity:

Going from:
[Build prior to addition of `onlyIf`](https://ci-master.corda.r3cev.com/viewLog.html?order=NAME_ASC&recordsPerPage=20&filterText=net.corda.behave&status=&currentGroup=package&scope=*%23teamcity%23*%23teamcity%23*&pager.currentPage=1&buildTypeId=Corda_PullRequests&buildId=65544&tab=testsInfo)
to:
[Build after addition of `onlyIf`](https://ci-master.corda.r3cev.com/viewLog.html?order=NAME_ASC&recordsPerPage=20&filterText=net.corda.behave&status=&currentGroup=package&scope=*%23teamcity%23*%23teamcity%23*&pager.currentPage=0&buildTypeId=Corda_PullRequests&buildId=65582&tab=testsInfo)

tlil:
@josecoll Okay, I'll put them back in then. :-) They won't run as part of a TeamCity build anyway, so if they take a couple of secs longer it's not the end of the world I suppose.

experimental/behave/src/test/kotlin/net/corda/behave/service/SqlServerServiceTests.kt
tlil: I can add it to the README...? Two reasons for doing this:

(1) Some of the tests aren't written for Windows and they require a Docker installation, etc. I can update those tests to support Windows as well, but ... 
(2) We don't want to have the longer running tests slow down TeamCity as this is an experimental module.

josecoll:
Bottom line, can we not just disable ALL experimental tests from being run in the standard Unit/Integration Test TC pipeline?
(without having to make custom changes to individual gradle files in experimental projects)

experimental/behave/src/test/kotlin/net/corda/behave/service/SqlServerServiceTests.kt
josecoll: 1) we shouldn't include ANU docker tests in the standard TC pipeline
2) Experimental should be excluded from the standard TC pipeline

tlil: Yes, I ultimately agree with both (1) and (2). I think the test "team" is on it.

tlil:
@m4ksio re @josecoll's comment - if I understood you correctly, that's ultimately the plan yeah? Could we simply update TeamCity to ignore tests in `experimental/` ?

m4ksio:
We could, in fact. Or I believe it might be better to disable them, and only enable with a special flag - they are experimental for a reason.

tlil:
Right, will close this PR and open a separate one that fixes / bypasses the Nix-specific tests

PR #2617, opened 2018-02-23
Update jolokia to 1.5.0

jamesbr3:
Updating Jolokia dependency to latest version (1.5.0)

### DIFF
--- a/node/src/main/resources/build.properties
+++ b/node/src/main/resources/build.properties
@@ -2,4 +2,4 @@
 # Note: sadly, due to present limitation of IntelliJ-IDEA in processing resource files, these constants cannot be
 # imported from top-level 'constants.properties' file
 
-jolokiaAgentVersion=1.3.7
+jolokiaAgentVersion=1.5.0


josecoll:
What changes does this upgrade bring to the table?

jamesbr3:
Latest jolokia includes security fixes. Neither (fixed) issue affects Corda specifically but its good practice to apply security fixes whenever they are available.

More detail at: https://jolokia.org/  (copied below)

* The Jolokia JMX proxy mode is vulnerable to remote code execution in a specific scenario. Note that this only affects the WAR agent, which has the proxy mode enabled by default. The JVM agent is not affected as it does not support the JMX proxy mode
* An XSS attack can be performed by manipulating the mimeType query parameter to coerce the Web browser to display the JSON response directly. However, we consider the impact quite low, as Jolokia responses are not shown directly in the browser but used by a JavaScript library, which has to perform its own escaping of user provided response data. Jolokia 1.5.0 fixes this hole nevertheless.

josecoll:
Thanks for clarifying @jamesbr3

PR #2620, opened 2018-02-23
ENT-1443 Add cert role to CSR and doorman issue cert according to the cert role

patrickkuo:
Cherry picked from R3 corda


shamsasari:
https://github.com/corda/enterprise/pull/431

PR #2623, opened 2018-02-23
Move classpath to system variable to avoid "Command too long" on Windows

m4ksio:
Windows is pretty sensitive when it comes to a path lengths, so some tests can fail if the user downloaded corda into "wrong" directory. Moving classpath to a system variable seems to overcome this problem.


m4ksio:
Closing due to formal requirements not being met.

PR #2625, opened 2018-02-23
CORDA-1040 docs for FlowLogic.sleep

rick-r3:
Added some docs

### DIFF
--- a/docs/source/api-flows.rst
+++ b/docs/source/api-flows.rst
@@ -748,20 +748,31 @@ We then update the progress tracker's current step as we progress through the fl
 
 Concurrency, Locking and Waiting
 --------------------------------
-This is an advanced topic.  If you are considering something in this area, we suggest you reach out to Developer
-Relations or Corda Support to discuss your requirements.  They may even be able to discuss previously encountered patterns
-that could solve your problem, or present another way of thinking about it.
+This is an advanced topic.  Because Corda is designed to:
 
-Because Corda is designed to run many flows in parallel, may persist flows to storage and
-resurrect those flows much later, or even (in the future) migrate flows between JVMs, flows should typically not attempt
-to interact with global static transient objects, and particularly should not make use of locks.  Locks will
-significantly reduce the scalability of the node, in the best case, and if they remain locked across flow context switch
-boundaries can cause the node to deadlock.  Those context switch boundaries include sending and receiving from peers
-discussed above.
+* run many flows in parallel,
+* may persist flows to storage and resurrect those flows much later,
+* (in the future) migrate flows between JVMs,
 
-Flows should also expressly not use ``Thread.sleep()``, since this will prevent the node from processing other flows
-in the meantime.  Please consider ``ScheduledState``.  However, we appreciate that Corda support for some more advanced
-patterns is still in the future, and if there is a need for brief pauses in flows then you should use ``FlowLogic.sleep``.
-This is not to be used to create long running flows or as a substitute to using the scheduler.
-Currently the ``finance`` package uses this to make several attempts at coin selection, where necessary, when many states
-are soft locked and we wish to wait for those, or other new states in their place, to become unlocked.
+flows should should avoid use of locks and typically not even attempt to interact with objects shared between flows (except
+``ServiceHub`` and other carefully crafted services such as Oracles.  See :doc:`oracles`).
+Locks will significantly reduce the scalability of the node, in the best case, and can cause the node to deadlock if they
+remain locked across flow context switch boundaries (such as sending and receiving
+from peers discussed above, and the sleep discussed below).
+
+If you need activities that are scheduled, you should investigate the use of ``SchedulableState``.
+However, we appreciate that Corda support for some more advanced patterns is still in the future, and if there is a need
+for brief pauses in flows then you should use ``FlowLogic.sleep`` in place of where you might have used ``Thread.sleep``.
+Flows should expressly not use ``Thread.sleep``, since this will prevent the node from processing other flows
+in the meantime, significantly impairing the performance of the node.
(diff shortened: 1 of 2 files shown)


docs/source/api-flows.rst
```diff
@@ -744,3 +744,24 @@ We then update the progress tracker's current step as we progress through the fl
         :start-after: DOCSTART 18
         :end-before: DOCEND 18
         :dedent: 12
+
+
+Concurrency, Locking and Waiting
+--------------------------------
+This is an advanced topic.  If you are considering something in this area, we suggest you reach out to Developer
+Relations or Corda Support to discuss your requirements.  They may even be able to discuss previously encountered patterns
+that could solve your problem, or present another way of thinking about it.
```
joeldudleyr3: I'm not sure about this. Maybe just say: "Support for locking in threads is experimental.", then combine it with the paragraph on the dangers below?

docs/source/api-flows.rst
```diff
@@ -744,3 +744,24 @@ We then update the progress tracker's current step as we progress through the fl
...
+
+Because Corda is designed to run many flows in parallel, may persist flows to storage and
+resurrect those flows much later, or even (in the future) migrate flows between JVMs, flows should typically not attempt
+to interact with global static transient objects, and particularly should not make use of locks.  Locks will
+significantly reduce the scalability of the node, in the best case, and if they remain locked across flow context switch
+boundaries can cause the node to deadlock.  Those context switch boundaries include sending and receiving from peers
```
joeldudleyr3: Is there a less technical term we can use than "flow context switch boundaries". If not, let's explicitly define it.

docs/source/api-flows.rst
```diff
@@ -744,3 +744,24 @@ We then update the progress tracker's current step as we progress through the fl
...
+
+Concurrency, Locking and Waiting
+--------------------------------
+This is an advanced topic.  If you are considering something in this area, we suggest you reach out to Developer
+Relations or Corda Support to discuss your requirements.  They may even be able to discuss previously encountered patterns
+that could solve your problem, or present another way of thinking about it.
+
+Because Corda is designed to run many flows in parallel, may persist flows to storage and
+resurrect those flows much later, or even (in the future) migrate flows between JVMs, flows should typically not attempt
+to interact with global static transient objects, and particularly should not make use of locks.  Locks will
```
joeldudleyr3: "global static transient objects" isn't a term I'm familiar with. Could you clarify here?

docs/source/api-flows.rst
```diff
@@ -744,3 +744,24 @@ We then update the progress tracker's current step as we progress through the fl
...
+significantly reduce the scalability of the node, in the best case, and if they remain locked across flow context switch
+boundaries can cause the node to deadlock.  Those context switch boundaries include sending and receiving from peers
+discussed above.
+
+Flows should also expressly not use ``Thread.sleep()``, since this will prevent the node from processing other flows
+in the meantime.  Please consider ``ScheduledState``.  However, we appreciate that Corda support for some more advanced
```
joeldudleyr3: Maybe say: "You can implement scheduled activities in the node in two ways:", then have bullets.

docs/source/api-flows.rst
```diff
@@ -744,3 +744,24 @@ We then update the progress tracker's current step as we progress through the fl
...
+patterns is still in the future, and if there is a need for brief pauses in flows then you should use ``FlowLogic.sleep``.
```
joeldudleyr3: Not clear to me based on this why `FlowLogic.sleep` is good or bad.

docs/source/api-flows.rst
```diff
@@ -744,3 +744,24 @@ We then update the progress tracker's current step as we progress through the fl
...
+This is not to be used to create long running flows or as a substitute to using the scheduler.
+Currently the ``finance`` package uses this to make several attempts at coin selection, where necessary, when many states
```
joeldudleyr3: Maybe have this in a sub-section called "Example usages", and link to the code in question more directly?

docs/source/api-flows.rst
rick-r3: I explicitly don't really want to suggest there's an equivalence between the two.  We want people to use `SchedulableState` (turns out there's a typo above).  If they try to do something advanced / multi-threaded then `sleep` is the the last resort.

docs/source/api-flows.rst
rick-r3: I'll try and fill out some reasoning.

docs/source/api-flows.rst
rick-r3: It's defined in the next sentence.  Trying to think of better terminology but nothing comes to mind as yet.

docs/source/api-flows.rst
rick-r3: I've renamed to `objects shared between flows`

docs/source/api-flows.rst
rick-r3: Which bit?  You mean the bit about "ask dev rel"? I can remove that bit and combine with below if you like?

joeldudleyr3: Yeah - the ask dev rel/support bit. We should just put what we'd tell them here instead, so that it's mass-consumable.

docs/source/api-flows.rst
joeldudleyr3: One of the issues is that the definition is after the end of the sentence where it's used. How about "if they remain locked across flow context switch boundaries (e.g. when sending and receiving from peers) can cause the node to deadlock"?

docs/source/api-flows.rst
joeldudleyr3: NP

rick-r3:
I updated.  Can you take another look @joeldudleyr3 ?

PR #2626, opened 2018-02-23
Kdoc/comment updates

kchalkias:
Fix unrecognised references and comment updates, soft-locking input/output clarification.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
@@ -6,13 +6,14 @@ import net.corda.core.concurrent.CordaFuture
 import net.corda.core.contracts.*
 import net.corda.core.crypto.SecureHash
 import net.corda.core.flows.FlowException
+import net.corda.core.flows.FlowLogic
 import net.corda.core.identity.AbstractParty
 import net.corda.core.messaging.DataFeed
-import net.corda.core.node.services.vault.PageSpecification
-import net.corda.core.node.services.vault.QueryCriteria
-import net.corda.core.node.services.vault.Sort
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
```diff
@@ -190,43 +192,43 @@ interface VaultService {
...
     @Throws(StatesNotAvailableException::class)
     fun softLockReserve(lockId: UUID, stateRefs: NonEmptySet<StateRef>)
 
     /**
      * Release all or an explicitly specified set of [StateRef] for a given [UUID] unique identifier.
-     * A vault soft lock manager is automatically notified of a Flows that are terminated, such that any soft locked states
-     * may be released.
-     * In the case of coin selection, softLock are automatically released once previously gathered unconsumed input refs
-     * are consumed as part of cash spending.
+     * A [Vault] softLock manager is automatically notified from flows that are terminated, such that any soft locked
```
shamsasari: Why the camel case for "softLock"? It should be either as it was or "soft-lock"

core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
```diff
@@ -213,7 +213,7 @@ interface VaultService {
 
     /**
      * Release all or an explicitly specified set of [StateRef] for a given [UUID] unique identifier.
-     * A [Vault] softLock manager is automatically notified from flows that are terminated, such that any soft locked
+     * A [Vault] soft-lock manager is automatically notified from flows that are terminated, such that any soft locked
      * states may be released.
      * In the case of coin selection, softLocks are automatically released once previously gathered unconsumed
```
shamsasari: This as well

PR #2639, opened 2018-02-26
CORDA-296: added handler for connection failover

bpaunescu:
This PR is aimed to fix a couple of issues found in the RPC client side.

1) if reconnection is done after more than a few seconds, the artemis messages resulted from RPC calls during the downtime are not delivered to the server, therefore the client blocks on the call indefinitely
2) if reconnection can never be done, the client blocks indefinitely

The solution adopted is similar to what the server does, and that is to buffer outgoing requests until failover completes. In case of success, the messages are sent and the calls complete successfully. If the connection times out, the messages are discarded and the rpc calls with throw and RPCException, making sure the client won't block.

This PR also brings back the RPC connection info during node startup, as signalled in https://github.com/corda/corda/issues/2630


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -170,6 +170,7 @@ open class Node(configuration: NodeConfiguration,
         printBasicNodeInfo("Incoming connection address", advertisedAddress.toString())
         rpcServerAddresses?.let {
             rpcMessagingClient = RPCMessagingClient(configuration.rpcOptions.sslConfig, it.admin, networkParameters.maxMessageSize)
+            printBasicNodeInfo("RPC connection address", rpcServerAddresses.primary.toString())
```
mnesbit: I know this seems silly, but this almost certainly needs to be a separate tiny PR as it probably has to go into V3.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -161,6 +161,10 @@ class RPCClientProxyHandler(
                 build()
     }
 
+    // Used toto buffer client requests if the server is unavailable
```
exFalso: typo `toto`

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -161,6 +161,10 @@ class RPCClientProxyHandler(
...
+    private var bufferOutgoingRequests = false
+    private val outgoingRequestBuffer = ArrayList<RPCApi.ClientToServer.RpcRequest>()
```
exFalso: I know the server also uses an arraylist, but i wonder whether it's better to do this with an ArrayBlockingQueue (and perhaps also on the server). Currently the buffer will explode under heavy load, i think it's better to backpressure with a simple thread block eventually

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -393,6 +405,34 @@ class RPCClientProxyHandler(
             sendMessage(RPCApi.ClientToServer.ObservablesClosed(observableIds))
         }
     }
+
+    private fun handleConnectionFailover(session: ClientSession?) {
+        session!!.addFailoverListener { event ->
+            when(event) {
```
exFalso: whitespace :police_car:

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -393,6 +405,34 @@ class RPCClientProxyHandler(
...
+                    outgoingRequestBuffer.forEach { request -> sendMessage(request) }
+                    outgoingRequestBuffer.clear()
+                }
+
+                FailoverEventType.FAILOVER_FAILED -> {
+                    log.error("Could not reconnect to the RPC server. All buffered requests will be discarded and RPC calls " +
+                            "will throw an RPCException.")
+                    rpcReplyMap.values.forEach { replyFuture ->
+                        replyFuture.setException(RPCException("Could not connect to RPC server."))
+                    }
```
exFalso: We also need to `onError` all observables in the observation pool

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
bpaunescu: Which line? I can't see any whitespace that's not supposed to be there.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
bpaunescu: If a blocking queue is used, the tricky part would be handling the messages waiting to be added to the queue when the connection is restored. The queue would be emptied and waiting messages added but not sent until next failover

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -393,6 +405,36 @@ class RPCClientProxyHandler(
...
+            when (event) {
+                FailoverEventType.FAILURE_DETECTED -> {
+                    log.warn("RPC server unavailable. RPC calls are being buffered.")
+                    bufferOutgoingRequests = true
+                }
+
+                FailoverEventType.FAILOVER_COMPLETED -> {
+                    log.info("RPC server available. Draining request buffer.")
+                    bufferOutgoingRequests = false
+                    outgoingRequestBuffer.forEach { request -> sendMessage(request) }
```
mnesbit: do we need a lock over this?

node/src/main/kotlin/net/corda/node/internal/Node.kt
bpaunescu: Done in https://github.com/corda/corda/pull/2656

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
exFalso: Hmm agree, I think this is racy atm. I'm also not sure about the failover trigger guarantees.

Perhaps a good test would be:
Use a pool of 8 threads to send 1k messages each using the same single client. While they're sending trigger a failover. Check that all messages arrive *in order*

bpaunescu:
@exFalso Have reworked the buffering to overcome a failure happening while the messages are being sent, which was an issue with the previous iteration of the code. Basically, all requests are mapped using the replyId and removed from the map after the server replies. There are no issues with the buffer growing too much because each thread that uses the client will block on every RPC call until a reply is received.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
bpaunescu: Done

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -393,6 +403,36 @@ class RPCClientProxyHandler(
...
     }
+
+    private fun handleConnectionFailover(session: ClientSession?) {
+        session!!.addFailoverListener { event ->
+            when (event) {
+                FailoverEventType.FAILURE_DETECTED -> {
+                    log.warn("RPC server unavailable. RPC calls are being buffered.")
+                }
+
+                FailoverEventType.FAILOVER_COMPLETED -> {
```
exFalso: We may need to raise exceptions in observables here as well. If the failover was caused by the server restarting then it won't send observations anymore. The functionality in this PR is fine as it is, we should add this as a separate JIRA

bpaunescu: I think @thschroeter raised a bug for this.

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
```diff
@@ -253,6 +259,114 @@ class RPCStabilityTests {
...
+                    ops = ops).getOrThrow().broker.hostAndPort!!
+
+            serverFollower.unfollow()
+            val proxy = RPCClient<ThreadOps>(serverPort).start(ThreadOps::class.java, "alice", "alice").proxy
+            val expectedMap = mutableMapOf<Int, String>()
+            val resultsMap = mutableMapOf<Int, String>()
+
+            (1 until threadNo).forEach { nr ->
+                (1 until messageNo).forEach { msgNo ->
+                    expectedMap[nr] = expectedMap.getOrDefault(nr, "").plus("($nr-$msgNo)")
```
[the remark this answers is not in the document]
bpaunescu: Indeed, the strings are constructed for final assertion.

PR #2661, opened 2018-02-27
CORDA-1122 Switch AMQP bridge to use auto-commit sends and acks

blsemo:
Remove needless pinch point in the AMQP bridge where calling `session.commit()` could limit message throughput.
Auto commit sends and acks, and only call commit explicitly before a rollback to ensure a consistent state of the session in this case.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/bridging/AMQPBridgeManager.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/bridging/AMQPBridgeManager.kt
@@ -148,6 +148,8 @@ class AMQPBridgeManager(config: NodeSSLConfiguration, val artemisMessageClientFa
                             artemisMessage.acknowledge()
                         } else {
                             log.info("Rollback rejected message uuid: ${artemisMessage.getObjectProperty("_AMQ_DUPL_ID")}")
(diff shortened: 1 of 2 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/bridging/AMQPBridgeManager.kt
```diff
@@ -146,9 +146,9 @@ class AMQPBridgeManager(config: NodeSSLConfiguration, val artemisMessageClientFa
                     lock.withLock {
                         if (sendableMessage.onComplete.get() == MessageStatus.Acknowledged) {
                             artemisMessage.acknowledge()
-                            session?.commit()
                         } else {
                             log.info("Rollback rejected message uuid: ${artemisMessage.getObjectProperty("_AMQ_DUPL_ID")}")
+                            session?.commit()
```
mnesbit: Could you explain this? Not saying it is wrong but it is pretty confusing to commit and then rollback

mnesbit: Also, can we look at the @Ignored test of bridge replay logic?

PR #2675, opened 2018-02-28
[CORDA-1112] Explicit dependency on JS dep dir seems to eliminate race condition bug

m4ksio:
Content of ```bower_components``` folder is being downloaded by ```clientInstall``` step, directly into the source tree for easier development. This however seems to cause a problem with builds on fresh repos.

### DIFF
--- a/samples/irs-demo/web/build.gradle
+++ b/samples/irs-demo/web/build.gradle
@@ -79,9 +79,9 @@ jar {
     from sourceSets.test.output
     dependsOn clientInstall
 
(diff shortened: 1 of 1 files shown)


samples/irs-demo/web/build.gradle
```diff
@@ -78,6 +78,10 @@ dependencies {
 jar {
     from sourceSets.test.output
     dependsOn clientInstall
+
+	from("src/main/resources/static/js/bower_components") {
```
IgorNitto: indentation

PR #2676, opened 2018-02-28
CORDA-696: Ensure deterministic transaction id calculation for contra…

adagys:
…ct upgrade and notary change transactions.

The problem with the previous implementation is that the transaction would be deserialized with the schema specified
in the serialized form, but the calculation of the id would involve re-serializing properties using the local serialization context
which might produce a different result.

For example, loading transactions from storage and sending them over RPC (the client is still using Kryo) resulted in different transaction ids.


core/src/main/kotlin/net/corda/core/crypto/CryptoUtils.kt
```diff
@@ -223,6 +223,7 @@ fun componentHash(opaqueBytes: OpaqueBytes, privacySalt: PrivacySalt, componentG
 fun componentHash(nonce: SecureHash, opaqueBytes: OpaqueBytes): SecureHash = SecureHash.sha256Twice(nonce.bytes + opaqueBytes.bytes)
 
 /** Serialise the object and return the hash of the serialized bytes. */
+@Deprecated("Do not use, the resulting hash won't be deterministic since it's affected by the serialization context")
```
mikehearn: I'm not sure that's right - the context is fixed in the serialize call, no? P2P context? It should always be AMQP.

adagys: I should be, but for some reason it gave me a different result when called in the RPC context. Still, even if the context is AMQP, the serialized form will contain different metadata depending on version. We could use the "serialize for version x" approach, but then you need to somehow propagate that during transaction deserialization.

kchalkias: I also believe we should have such a function, which of course should be deterministic. We should also ensure the context does not change (eg adds metadata) for wire compatibility and it should be stable. AMQP is supposed to work like that, right @fenryka ?

adagys:
Need to confirm with @fenryka, but I think it's required for V3

core/src/main/kotlin/net/corda/core/crypto/CryptoUtils.kt
fenryka: There is a bunch of work for post V3 here to do, at the moment it's fine as we only have a single version (and the only metadata we pack into a blob is the protocol version) but we need the ability to serialize *as* a version being added

RPC will differ at the moment as that's still Kryo

adagys: Right, but I'm just saying that even with "serialize as version" we'd need to store some additional version metadata on the transaction itself and that's not ideal. It seems safest to perform transaction id calculation on a partially deserialized transaction - we don't have to worry about all of this.

fenryka: what metadata aside from  the serialized form version do we need, if we can pull off the header, know how it was serialized, we can deterministically recreate that

fenryka: That's true, if you don't *have* to rely on the seriliased form... why would you, save the overhead of deserialising

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -18,13 +21,20 @@ import java.security.PublicKey
...
-        override val notary: Party,
-        val legacyContractAttachmentId: SecureHash,
-        val upgradeContractClassName: ContractClassName,
-        val upgradedContractAttachmentId: SecureHash,
+        /**
+         * Contains all of the transaction components in serialized form.
+         * This is used for calculating the transaction id in a deterministic fashion, since re-serializing properties
+         * may result in a different byte sequence depending on the serialization context.
+         */
+        val serializedComponents: List<OpaqueBytes>,
```
fenryka: if this is a list of serialised blobs wouldn't SerializedBytes be better?

core/src/main/kotlin/net/corda/core/crypto/CryptoUtils.kt
mikehearn: I think there are different issues being conflated here.

1) Whether *this specific function* should be round-tripping successfully. Nothing in the API doc i.e. specification says that you are guaranteed to get the exact same hash as the hash of the bytes you deserialized _from_. The docs simply say it serialises the object and returns the hash. So I don't see why we should deprecate it - it's working as specified.

2) How best to handle bitwise exact round-tripping from bytes->objects->bytes, in the general case. Now, I believe this is best handled as a new feature post v3 in the serialisation engine, specifically, we can track metadata associated with a deserialised object using a WeakHashMap. That metadata could either be things like the version we read (small, light) or even the original byte array. _If_ the object is fully immutable, which we tell people it should be, right - although undoubtably this will break some apps - then you can just return the original bytes when asked to re-serialise instead of actually doing the operation. This approach has the advantage of being old-bug compatible too.

3) How to resolve the immediate bug you encountered to do with Kryo vs AMQP and the fact that the context doesn't seem to be respected here. Can you debug this further?

Overall I'm really keen on avoiding this change - it's big and subtle and coming in very late to the v3 release cycle. And it's a set of wire changes that seem motivated by a workaround for an unexplained bug rather than something fundamental.

adagys: 1. I don't care too much about deprecating this function and maybe let's leave it out of the discussion. I can probably just add a comment instead of deprecation. It's just that its only uses in the codebase were for id calculation and it's pretty useless if determinism isn't guaranteed.
2-3. Ok, I'll do some more investigation

Another side effect of this change is that in brings NotaryChange & ContractUpgrade transaction inline with WireTransaction in that in allows calculating the transaction id without necessarily being able to deserialize all the fields - so if we add a new property, old clients can still compute the right id. I'm not sure how important that is though, maybe @kchalkias can comment?

kchalkias: I personally like the approach of sending bytes instead of objects as it gives flexibility wrt to backwards compatibility. We had this discussion in the past, but now that implementation is closer to that of  TraversableTransaction, is it possible to reuse some of its deserialisation logic?

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -39,11 +49,15 @@ data class ContractUpgradeWireTransaction(
         get() = throw UnsupportedOperationException("ContractUpgradeWireTransaction does not contain output states, " +
                 "outputs can only be obtained from a resolved ContractUpgradeLedgerTransaction")
 
+    override val id: SecureHash by lazy {
+        serializedComponents[INPUTS.ordinal].sha256()
+                .hashConcat(serializedComponents[NOTARY.ordinal].sha256())
+                .hashConcat(hiddenComponentHash)
```
kchalkias: I think a more similar to previous logic
`sha256(inputs + notary).hashConcat(sha256Twice(privacySalt.bytes + restComponents))` might be faster with less hash invocations.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -39,11 +49,15 @@ data class ContractUpgradeWireTransaction(
...
+    }
+
     /** Hash of the list of components that are hidden in the [ContractUpgradeFilteredTransaction]. */
     private val hiddenComponentHash: SecureHash
-        get() = serializedHash(listOf(legacyContractAttachmentId, upgradeContractClassName, privacySalt))
-
-    override val id: SecureHash by lazy { serializedHash(inputs + notary).hashConcat(hiddenComponentHash) }
+        get() = serializedComponents.takeLast(3).foldRight(privacySalt.bytes.sha256()) { next, acc -> acc.hashConcat(next.sha256()) }
```
kchalkias: due to the above comment, return `sha256Twice(privacySalt.bytes + restComponents)` here.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -77,15 +96,18 @@ data class ContractUpgradeWireTransaction(
...
-        override val notary: Party,
-        val rest: SecureHash
+        val serializedComponents: List<OpaqueBytes>,
+        val hiddenComponentHash: SecureHash
 ) : CoreTransaction() {
-    override val id: SecureHash get() = serializedHash(inputs + notary).hashConcat(rest)
+    override val inputs: List<StateRef> by lazy { serializedComponents[INPUTS.ordinal].deserialize<List<StateRef>>() }
+    override val notary: Party by lazy { serializedComponents[NOTARY.ordinal].deserialize<Party>() }
+    override val id: SecureHash by lazy {
+        serializedComponents[INPUTS.ordinal].sha256().hashConcat(serializedComponents[NOTARY.ordinal].sha256()).hashConcat(hiddenComponentHash)
```
kchalkias: ditto

core/src/main/kotlin/net/corda/core/transactions/NotaryChangeTransactions.kt
```diff
@@ -39,16 +50,24 @@ data class NotaryChangeWireTransaction(
      * A privacy salt is not really required in this case, because we already used nonces in normal transactions and
      * thus input state refs will always be unique. Also, filtering doesn't apply on this type of transactions.
      */
-    override val id: SecureHash by lazy { serializedHash(inputs + notary + newNotary) }
+    override val id: SecureHash by lazy {
+        serializedComponents[INPUTS.ordinal].bytes.sha256()
+                .hashConcat(serializedComponents[NOTARY.ordinal].bytes.sha256())
+                .hashConcat(serializedComponents[NEW_NOTARY.ordinal].bytes.sha256())
+    }
```
kchalkias: similar sha256 logic with contract upgrade.

adagys:
@mikehearn so https://github.com/corda/corda/pull/2705 fixes the serialization problem. If we assume that in the future we'll use the WeakHashMap approach or something similar for deterministic re-serialization, this work isn't necessary.
The only remaining debate is whether it's useful to have these transactions structured similarly to TraversableTransaction for flexibility/backwards compatibility issues as @kchalkias mentioned.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
adagys: don't think it's very useful here, as we don't have a specific type, `SerializedBytes<Any>` is not much better than `OpaqueBytes`

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
adagys: would it make sense to combine all component bytes into one and then get the hash?

core/src/main/kotlin/net/corda/core/crypto/CryptoUtils.kt
```diff
@@ -222,7 +222,10 @@ fun componentHash(opaqueBytes: OpaqueBytes, privacySalt: PrivacySalt, componentG
 /** Return the SHA256(SHA256(nonce || serializedComponent)). */
 fun componentHash(nonce: SecureHash, opaqueBytes: OpaqueBytes): SecureHash = SecureHash.sha256Twice(nonce.bytes + opaqueBytes.bytes)
 
-/** Serialise the object and return the hash of the serialized bytes. */
+/**
+ * Serialise the object and return the hash of the serialized bytes. Note that the resulting hash is not guaranteed to
+ * be deterministic since the serialized bytes are affected by the serialization context.
+ */
```
rick-r3: This could be viewed as a little misleading.  It should be deterministic (we try to ban types that are not, like `HashMap`), but different code versions might give different values if any of the types being serialized have changed, or if the version of serialization specified by the context changes.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -18,13 +21,20 @@ import java.security.PublicKey
 /** A special transaction for upgrading the contract of a state. */
 @CordaSerializable
 data class ContractUpgradeWireTransaction(
```
rick-r3: Can we maintain API stability by using secondary constructors rather than distinct builders.  I agree the internal only aspect of the builders might be good, but we already committed to this API.

adagys: Upgrade transactions haven't been released yet, no API break

core/src/main/kotlin/net/corda/core/crypto/CryptoUtils.kt
adagys: Ok, I'll expand the comment a bit

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
kchalkias: by all you mean visible and not visible all together? how are you going to compute the tree if you then receive notary and inputs only in the ftx?

kchalkias: if you are referring to `restComponents`, yes this is byte concatenation of the non-visible components.

adagys: ah sorry I mean concatenate all visible component bytes, hash it, concatenate all hidden component bytes, hash it, and combine the hashes

kchalkias: But this is what `sha256(inputs + notary).hashConcat(sha256Twice(privacySalt.bytes + restComponents))` does, no? the only extra thing is salt concatenation in the non-visible part (and I used hash twice, as this should typically be an HMAC where key is the salt).

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
rick-r3: Okay, but `NotaryChangeWireTransaction` is released.

rick-r3:
Whilst it is a late change, it seems self contained and follows a pattern we used for `WireTransaction`.  It also avoids a potential need to upgrade it later necessitating a wait for a `minimumPlatformVersion` increase.  If a new property is added, old nodes won't include it in the transaction ID even with cached bytes etc since the choice of what to include in the `SecureHash` is in the executing node code, not in the creator of the transaction as it would now be.

 i.e. this is not required now, but might save us from such work later.  That is something that isn't clear from your last comment @adagys, when you say it isn't needed.  You mean, right now, correct?

 So we're weighing up what seems like very little downside (given the work is done) vs. a risk of a much more painful change later.  Admittedly the risk might be zero.

I'm leaning towards approving it.

mikehearn:
I think I lost track of where this is - the PR you linked to that was supposed to fix the underlying serialization issue got closed, but I'm not sure if it got replaced with a newer/better PR. 

>  If a new property is added, old nodes won't include it in the transaction ID even with cached bytes etc since the choice of what to include in the SecureHash is in the executing node code, not in the creator of the transaction as it would now be.

Right - that is a problem. The hash of the transaction is not the hash of the underlying bytes. It's actually a hash computed via a rather convoluted algorithm that would be awkward to explain, if I read the code correctly (involving serialisation of lists, hashing them, concatenating with other hashes etc). 

However in the new code added by this PR, that's the same - the id calculation would ignore any new fields added. Which isn't right either. It should be something like `override val id: SecureHash by lazy { serializedComponents.map { it.sha256() }.foldRight`, no? If we added a new field to this structure post-v3 how would it affect the id hash?

This looks like a stop-ship problem.

Anyway, I looked a bit into the preservation-of-bytes idea. We should do it anyway, but for tearable data structures specifically it seems tricky, because an inner object doesn't necessarily have a standalone byte stream to begin with. Due to the schemas, version header and so on, you can't just slice out a sub-objects bytes from a bigger object graphs original byte array and expect to get it back by re-serialising. Only a top level object would work like that. It would be a nice property if we could do that, but it'd mean a single byte stream would need to support multiple versions of the serialisation protocol at once, you'd not be able to do back-references across those boundaries etc and it doesn't seem worth it. We can already represent serialisation boundaries in code using `SerializedBytes<T>` and so it should remain.

With respect to calculating the hash in ways that we can add fields and always have them be reflected in the hash calculation, but still tear-off parts of the data structure, I guess we should be using _exactly_ the same structure as WireTransaction - with `List<ComponentGroup>` and the same approach to calculating a MerkleTree. The current approach tries to short-circuit that whilst preserving filterability/tearability, and that doesn't seem to work.

Ideally we'd have a nice abstraction for tearable data structures. It looks possible - the relevant code in WireTransaction is marked internal and could be refactored. Then users who want to make tearable states and flow messages could do so easily. But maybe that's a project for another day.

adagys:
Thanks Mike. The reason I didn't follow the `ComponentGroup` approach is that we don't allow dynamic transaction filtering for contract upgrades – the filtering is only used for non-validating notaries and will be controlled by the platform. Because of that we can simplify the component group logic and don't need Merkle trees.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -60,7 +60,7 @@ data class ContractUpgradeWireTransaction(
     }
 
     /** Required for filtering transaction components. */
-    private val nonces = (1 until serializedComponents.size).map { (Ints.toByteArray(it) + privacySalt.bytes).sha256() }
+    private val nonces = (0 until serializedComponents.size).map { (Ints.toByteArray(it) + privacySalt.bytes).sha256() }
```
kchalkias: better use a similar to `CryptoUtils.computeNonce` (of course you need only one index) to:
a) prevent length extension attacks (not that it really affects our threat model atm, but let's be proactive)
b) get rid of guava when possible

adagys:
Ready for another look

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
adagys: done

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -39,11 +50,18 @@ data class ContractUpgradeWireTransaction(
         get() = throw UnsupportedOperationException("ContractUpgradeWireTransaction does not contain output states, " +
                 "outputs can only be obtained from a resolved ContractUpgradeLedgerTransaction")
 
-    /** Hash of the list of components that are hidden in the [ContractUpgradeFilteredTransaction]. */
-    private val hiddenComponentHash: SecureHash
-        get() = serializedHash(listOf(legacyContractAttachmentId, upgradeContractClassName, privacySalt))
+    override val id: SecureHash by lazy {
+        serializedComponents.mapIndexed { index, component ->
+            SecureHash.sha256Twice(component.bytes + nonces[index].bytes)
```
kchalkias: let's better use `CryptoUtils.componentHash(nonces[index], component.bytes)`, it does the same.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -65,28 +83,69 @@ data class ContractUpgradeWireTransaction(
...
+    /** Constructs a filtered transaction: the inputs and the notary party are always visible, while the rest are hidden. */
     fun buildFilteredTransaction(): ContractUpgradeFilteredTransaction {
-        return ContractUpgradeFilteredTransaction(inputs, notary, hiddenComponentHash)
+        val totalComponents = (0 until serializedComponents.size).toSet()
+        val visibleComponents = mapOf(
+                INPUTS.ordinal to FilteredComponent(serializedComponents[INPUTS.ordinal], nonces[INPUTS.ordinal]),
+                NOTARY.ordinal to FilteredComponent(serializedComponents[NOTARY.ordinal], nonces[NOTARY.ordinal])
+        )
+        val hiddenComponents = (totalComponents - visibleComponents.keys).map { index ->
+            val hash = SecureHash.sha256Twice(serializedComponents[index].bytes + nonces[index].bytes)
```
kchalkias: ditto `componentHash`

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -65,28 +83,69 @@ data class ContractUpgradeWireTransaction(
...
+    override val notary: Party by lazy {
+        visibleComponents[NOTARY.ordinal]?.component?.deserialize<Party>()
+                ?: throw IllegalArgumentException("Notary not specified")
+    }
+    override val id: SecureHash by lazy {
+        val totalComponents = visibleComponents.size + hiddenComponents.size
+        val hashList = (0 until totalComponents).map { i ->
+            when {
+                visibleComponents.containsKey(i) -> {
+                    SecureHash.sha256Twice(visibleComponents[i]!!.component.bytes + visibleComponents[i]!!.nonce.bytes)
```
kchalkias: ditto `componentHash`

adagys:
It's actually `ContractUpgradeWireTransaction.Component`, so should be ok

mikehearn:
OK, thanks, seems the issue with new fields not being in the hash is fixed now so I'm happy.

BTW you can write it as `collection.indices.map` rather than `(0 until collection.size).map`

PR #2678, opened 2018-02-28
allow nodes to be waited for inprocess

roastario:

Adds a new "Waitable" interface within DriverDSLImpl, which allows inprocess nodes to be waited upon in the same manor as out of process nodes.


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -111,8 +112,14 @@ class DriverDSLImpl(
     private lateinit var _notaries: CordaFuture<List<NotaryHandle>>
     override val notaryHandles: List<NotaryHandle> get() = _notaries.getOrThrow()
 
+
+    interface Waitable {
+        @Throws(InterruptedException::class)
+        fun waitFor(): Int
```
fenryka: Do you ever actually use the return value here? Int seems a bit random, it just returns 1 AFAICS so why not just return Unit?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -779,8 +799,8 @@ class DriverDSLImpl(
                 }
                 node to nodeThread
             }.flatMap { nodeAndThread ->
-                addressMustBeBoundFuture(executorService, config.corda.p2pAddress).map { nodeAndThread }
-            }
+                        addressMustBeBoundFuture(executorService, config.corda.p2pAddress).map { nodeAndThread }
```
fenryka: spurious refactor? looks a bit odd now unless this is how intelliJ is formatting it

andr3ej:
adding @exFalso as this is his area

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/ShutdownManager.kt
```diff
@@ -3,7 +3,11 @@ package net.corda.testing.node.internal
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.internal.ThreadBox
 import net.corda.core.internal.concurrent.doneFuture
-import net.corda.core.utilities.*
+import net.corda.core.utilities.Try
+import net.corda.core.utilities.contextLogger
+import net.corda.core.utilities.getOrThrow
+import net.corda.core.utilities.seconds
+import java.util.concurrent.Callable
```
andr3ej: unused import

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/ShutdownManager.kt
```diff
@@ -101,4 +107,8 @@ class ShutdownManager(private val executorService: ExecutorService) {
             }
         }
     }
+
+    fun isShuttingDown(): Boolean {
```
andr3ej: nothing uses this?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -666,20 +673,33 @@ class DriverDSLImpl(
...
                 establishRpc(config, openFuture()).flatMap { rpc ->
                     allNodesConnected(rpc).map {
                         InProcessImpl(rpc.nodeInfo(), rpc, config.corda, webAddress, useHTTPS, thread, onNodeExit, node)
                     }
                 }
             }
+            state.locked {
+                processes += object : Waitable {
+                    override fun waitFor(): Int {
+                        nodeAndThreadFuture.toCompletableFuture().join().second.join()
```
andr3ej: nodeAndThreadFuture.getOrThrow().second.join()

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -666,20 +673,33 @@ class DriverDSLImpl(
...
+                        return 1;
```
andr3ej: if we're keeping the return value, this should be 0 for success. also, superfluous semicolon

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -111,8 +112,14 @@ class DriverDSLImpl(
     private lateinit var _notaries: CordaFuture<List<NotaryHandle>>
     override val notaryHandles: List<NotaryHandle> get() = _notaries.getOrThrow()
 
+
```
andr3ej: minimise blank lines throughout please, it's the coding standard

testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
```diff
@@ -152,5 +152,32 @@ class DriverTests {
...
+            it.nodeHandles.toCompletableFuture().get().forEach {
+                try {
+                    it.stop()
+                } catch (e: Exception) {
+                    e.printStackTrace()
+                }
+            }
+        }
+        driverThread.join()
+    }
```
andr3ej: there's lots i don't like about this test, most importantly there may be subtle issues with running driver in the background, so instead of bombarding you with comments i'm going to have a go at rewriting it

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
andr3ej: agreed, the only thing that uses this doesn't care about return value

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
andr3ej: i'd keep formatting as-is unless you're actually changing the code, even if it's well mangled, otherwise you're causing a mysterious merge conflict for someone down the line

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -871,8 +891,8 @@ class DriverDSLImpl(
             val index = stackTrace.indexOfLast { it.className == "net.corda.testing.driver.Driver" }
             // In this case we're dealing with the the RPCDriver or one of it's cousins which are internal and we don't care about them
             if (index == -1) return emptyList()
-            val callerPackage = Class.forName(stackTrace[index + 1].className).`package` ?:
-                    throw IllegalStateException("Function instantiating driver must be defined in a package.")
+            val callerPackage = Class.forName(stackTrace[index + 1].className).`package`
+                    ?: throw IllegalStateException("Function instantiating driver must be defined in a package.")
```
andr3ej: see other comment about formatting

testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
andr3ej: please see andr3ej-driver-waits for refactored test

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/ShutdownManager.kt
roastario: it's used by the gradle-plugins in the future.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -778,7 +796,7 @@ class DriverDSLImpl(
                 }
                 node to nodeThread
             }.flatMap { nodeAndThread ->
-                addressMustBeBoundFuture(executorService, config.corda.p2pAddress).map { nodeAndThread }
+              addressMustBeBoundFuture(executorService, config.corda.p2pAddress).map { nodeAndThread }
```
fenryka: formatting oddness, 2 space indent!

testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
```diff
@@ -153,5 +157,31 @@ class DriverTests {
         }
     }
 
+
+    @Test
+    fun `driver waits for nodes to finish`() {
+        fun NodeHandle.stopQuietly() = try {
+            stop()
+        } catch (t: Throwable) {
+            t.printStackTrace()
```
fenryka: Does this abort the test early if we end up here or having caught the exception do  we blunder on and fail in an interesting way later? Seems non obvious what "just" printing it achieves, if this is a recoverable / we don't care error a comment to say taht would be helpful. If this is actually bad shouldn't we re-throw / exit the test at this point

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -871,7 +889,7 @@ class DriverDSLImpl(
             // In this case we're dealing with the the RPCDriver or one of it's cousins which are internal and we don't care about them
             if (index == -1) return emptyList()
             val callerPackage = Class.forName(stackTrace[index + 1].className).`package` ?:
-                    throw IllegalStateException("Function instantiating driver must be defined in a package.")
+                   throw IllegalStateException("Function instantiating driver must be defined in a package.")
```
fenryka: Should probably include the class name in the error message so it's a bit more helpful, can ewasily look at the stack but I can imagine some loggers / users skimming it with their eyes and going "great, but *which* function)

testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
andr3ej: * the error i've seen is network map / node watcher / one of those things sometimes getting confused on driver exit
* we don't care as long as it doesn't interfere with the node actually stopping and driver actually exiting
* it's not up to this test to fix another part of the codebase so fine to ignore
* printStackTrace is there in case the error does become significant, so the failure reason is clear to someone running the test manually
* @roastario correct me if i'm wrong and maybe it would be worth adding a comment

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -111,10 +109,9 @@ class DriverDSLImpl(
     private lateinit var _notaries: CordaFuture<List<NotaryHandle>>
     override val notaryHandles: List<NotaryHandle> get() = _notaries.getOrThrow()
 
-
     interface Waitable {
         @Throws(InterruptedException::class)
-        fun waitFor(): Int
+        fun waitFor(): Unit
```
andr3ej: Unit can be elided but don't worry about it for this PR

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
andr3ej: class name is out of scope and can be a separate jira, but please restore the indent to how it was originally

testing/node-driver/src/integration-test/kotlin/net/corda/testing/driver/DriverTests.kt
roastario: that is correct.

roastario: this exception should be printed, I believe - it's just information for the user, it's usually the same ones about networkmap cache updates being rejected due to the executor being shutdown.

PR #2680, opened 2018-02-28
Balance object for negative amounts

amoothart:
Added a `Balance` class similar to the `Amount` class but with support holding negative numbers.

Refactored out common functionality of `Amount` into an `AbstractAmount` class which both `Amount` and `Balance` inherit from.

Unit tests, integration tests and IRS sample were tested as part of this pull request.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/Amount.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/Amount.kt
@@ -1,5 +1,6 @@
 package net.corda.core.contracts
 
+import net.corda.core.contracts.AbstractAmount.Companion.getDisplayTokenSize
 import net.corda.core.crypto.CompositeKey
 import net.corda.core.identity.Party
 import net.corda.core.serialization.CordaSerializable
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/net/corda/core/contracts/Amount.kt
```diff
@@ -483,3 +500,205 @@ class AmountTransfer<T : Any, P : Any>(val quantityDelta: Long,
...
+     */
+    operator fun times(other: Int): Balance<T> = Balance(Math.multiplyExact(quantity, other.toLong()), displayTokenSize, token)
+
+    /**
+     * This method provides a token conserving divide mechanism.
+     * @param partitions the number of balances to divide the current quantity into.
+     * @return 'partitions' separate Balance objects which sum to the same quantity as this Balance
+     * and differ by no more than a single token in size.
+     */
+    fun splitEvenly(partitions: Int): List<Balance<T>> {
```
mnesbit: This won't work for negative amounts and will need a test

core/src/main/kotlin/net/corda/core/contracts/Amount.kt
```diff
@@ -483,3 +500,205 @@ class AmountTransfer<T : Any, P : Any>(val quantityDelta: Long,
...
+ * regards [Balance] behaves like an [Amount]
+ *
+ * @property amount the number of tokens as a long value.
+ * @property displayTokenSize the nominal display unit size of a single token, potentially with trailing decimal display
+ * places if the scale parameter is non-zero.
+ * @property token the type of token this is an amount of. This is usually a singleton.
+ * @param T the type of the token, for example [Currency]. T should implement [TokenizableAssetInfo] if automatic conversion to/from a display format is required.
+ */
+@CordaSerializable
+data class Balance<T : Any>(val amount: Long,
```
mnesbit: There are lots of reasons I did not allow negative Amounts. The main problems are:
1. Negative from who's perspective? Usually a negative value means you owe money to someone else, but amount doesn't know about owners, lenders, etc. Balances are often perspective flipped according to context, which doesn't fit with the need for a single representation we can sign over.
2. Negative quantities are usually asymmetric compared to positive quantities in terms of handling. i.e. we treat debt very differently to holding liquid assets and charge/pay different rates of interest, tax, etc. 
3. Amount is tied to tokenized assets in FungibleStates. Allowing people to create states with negative amounts completely breaks the issuer logic and thus even though the total amount is conserved now I can create the appearance of more cash using just Move commands that I sign. In real life this is only possible through loans, which we pay for and have credit risk controls and overdraft limits.
4. The current cash selection and soft-locking code won't handle negative states correctly, so this isn't really usable on FungibleStates and essentially you might as well use your own Integers/Decimal values on your custom state rather than break the platform type.
5. Rounding gets a lot more complicated when we allow negative amounts, so users of conversions to/from decimal  e.g in a payment text box will find it much much harder to not end up with the outputs summing to different amounts from the inputs.
6. Proper representations of Credit and Debt really require inputs from things like interest rate and holiday calendar Oracles and thus need full smart contract and flow support. The Amount type is meant to be a simple building block. I would have been more supportive of a Balance ContractState type that holds Debt and Credit Obligations, but that is a higher order construct that should live in :finance not :core.

amoothart: Thanks @mnesbit, I'll submit a full design review to cover the problem and proposed solution.

1. The amount being tracked in balance will go from or to the entity owning the asset based on the value. However, the counterparty is not represented on the Corda network. 
1. The logic for handling negative amounts differently will be handled in Contracts and Flows right? This change would only allow for another building block which allows for the storing of negative balances and each Cordapp would need to handle the negative amount appropriately.
1. I agree that we cannot allow negative amounts in FungibleStates. The change was done in such a way that Amount != Balance and FungibleAsset continues to use Amount<Issued<T>> so it is not possible have Balances in a FungibleState.
1. Can you explain more about this concern? From point 3 negative amounts should not be possible with any cash mechanisms built into Corda.
1. Can you elaborate more on this as well. How is the rounding issue different between positive and negative numbers?
1. Interest rate and calendars may be required on a case by case basis but those should be handled external to Balance as part of the broader Cordapp. That should have no impact on this change. I am open to move this to :finance to hold the amount as a credit/debit.

PR #2681, opened 2018-02-28
CORDA-1117 Fix non deterministic port allocation issue in ExplorerSimulation

kchalkias:
Port allocation in `ExplorerSimulation` is using Driver and previous implementation was not deterministic regarding port allocation.
A fix by implementing a custom [PortAllocation] has been applied.
Docs and readme files have been updated with the correct fixed ports.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
@@ -2,7 +2,6 @@
 
 package net.corda.testing.driver
 
-import net.corda.core.CordaInternal
 import net.corda.core.DoNotImplement
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.flows.FlowLogic
@@ -17,6 +16,8 @@ import net.corda.core.utilities.getOrThrow
 import net.corda.node.internal.Node
 import net.corda.testing.common.internal.testNetworkParameters
 import net.corda.testing.core.DUMMY_NOTARY_NAME
+import net.corda.testing.driver.PortAllocation.Incremental
+import net.corda.testing.driver.internal.RandomFree
 import net.corda.testing.driver.internal.internalServices
 import net.corda.testing.node.NotarySpec
 import net.corda.testing.node.User
@@ -25,8 +26,6 @@ import net.corda.testing.node.internal.genericDriver
 import net.corda.testing.node.internal.getTimestampAsDirectoryName
 import net.corda.testing.node.internal.newContext
 import rx.Observable
-import java.net.InetSocketAddress
-import java.net.ServerSocket
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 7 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/driver/Driver.kt
```diff
@@ -107,18 +106,24 @@ data class WebserverHandle(
...
-    /** Get the next available port via [nextPort] and then return a [NetworkHostAndPort] **/
+    /** Get the next available port via [nextPort] and then return a [NetworkHostAndPort]. */
     fun nextHostAndPort() = NetworkHostAndPort("localhost", nextPort())
 
     /**
-     * An implementation of [PortAllocation] which allocates ports sequentially
+     * This method is required from custom [PortAllocation] implementations in which port allocation policy
+     * is based on the [input] object. Default implementation calls [nextHostAndPort].
+     */
+    open fun <T : Any> nextHostAndPort(input: T) = nextHostAndPort()
```
shamsasari: We shouldn't be exposing a public API method just to fix a demo.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -184,7 +184,7 @@ class DriverDSLImpl(
             startInSameProcess: Boolean?,
             maximumHeapSize: String
     ): CordaFuture<NodeHandle> {
-        val p2pAddress = portAllocation.nextHostAndPort()
+        val p2pAddress = portAllocation.nextHostAndPort(Pair(providedName, "p2p"))
```
shamsasari: This new API isn't designed right. The type of the argument is `Any` but the implementation here is clearly using `Pair<CordaX500Name, String>` and so you anyone wanting to provide their own implementation would need to know that.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -215,10 +215,10 @@ class DriverDSLImpl(
                                     customOverrides: Map<String, Any?>,
                                     startInSameProcess: Boolean? = null,
                                     maximumHeapSize: String = "200m",
-                                    p2pAddress: NetworkHostAndPort = portAllocation.nextHostAndPort()): CordaFuture<NodeHandle> {
-        val rpcAddress = portAllocation.nextHostAndPort()
-        val rpcAdminAddress = portAllocation.nextHostAndPort()
-        val webAddress = portAllocation.nextHostAndPort()
+                                    p2pAddress: NetworkHostAndPort = portAllocation.nextHostAndPort(Pair(name, "p2p"))): CordaFuture<NodeHandle> {
+        val rpcAddress = portAllocation.nextHostAndPort(Pair(name, "rpc"))
```
shamsasari: These strings should be enums

shamsasari (changes requested):
This is the wrong approach. I still think we can use `CordformDefinition` to define upfront the ports. Have a look at `BankOfCordaCordformTest` which runs the BOC demo as a CordformDefinition by calling `deployAndRunNodesThen` which allows you to run your own code after all the nodes have started.

kchalkias:
I had a solution with enums, wanted to agree first on the api. Although I feel the flexibility for custom `PortAllocation` will be useful in the future anyway, will change implementation to use `CordformDefinition`

PR #2682, opened 2018-03-01
CORDA-1130 - Round trip serialization of the TVE exceptions fail

fenryka:
This appears to be because the way they're written simply makes them
not serializable by the AMQP library. They could be more simply fixed to
map the missing parameters to properties, but then the round trip
serialization will fail as the constructors have side effects

The only way to get around this *and* to preserve the clean interface to
those calling the exceptions from user code is to allow the serializer
to access the main constructor and use a secondary for external access.

To disambiguate the signatures where needed use a dummy parameter that is
understood by the Serializer and that it knows to not include in the
serialized form. However, to enable this, this does mean adding a new
constructor method to Flow Exception

Also create a new module for common elements that node-api and core need
to access, like the serialization exception.


fenryka:
This is  currently very borked by some gradle /plugins issue that I'm looking into

fenryka:
Update from @tudor-malene this does fix the issues they were seeing

node-api-interfaces/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: 'kotlin'
+apply plugin: 'net.corda.plugins.publish-utils'
+apply plugin: 'com.jfrog.artifactory'
+
+description 'Corda node API external interfaces and annotatins'
+
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jre8:$kotlin_version"
+    compile "net.corda.plugins:cordform-common:$gradle_plugins_version"
```
chrisr3: Why do you need this dependency? I don't think you're using `cordform-common`.

fenryka: done

node-api-interfaces/src/main/kotlin/net/corda/nodeapiinterfaces/serialization/ConstructorForDeserialization.kt
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Annotation indicating a constructor to be used to reconstruct instances of a class during deserialization.
+ *
+ * NOTE: This is a duplicate of the annotation provided by th net.corda.nodeapi.internal.serialization.amqp
+ * package. We can't use that here as node-api depends on core, and we can't just move it into core because
+ * that would break the API.
+ */
+@Target(AnnotationTarget.CONSTRUCTOR)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ConstructorForDeserialization
```
chrisr3: Do you want either of these annotations appearing in the API Scanner results? Because if you don't then I _would_ have suggested annotating them both with `@CordaInternal` ... except that would then create a circular dependency on `:core`!

Maybe this should be a `corda-core-annotations` module instead and should include `@CordaInternal`? Except this would also require a change to the API Scanner!

fenryka: I don't think it matters if they appear in the api scanner does it? 

ConstructorForDeserialization is user facing as changing it could change how a class is serialised which would be a wire break (without careful thought around that)

chrisr3: Well, there has been a lot of Unhappiness with the amount of "churn" in the API Scanner output. So the question is "Do they _need_ to be there? Or can we ignore them?".

FWIW I have an experimental change to the API Scanner so that it _can_ handle any `@CordaInternal` annotation. Thank you for marching an AT-AT over my original assumptions... :wink:.

fenryka: *grins*

So API stability wise is it "bad" for the annotations to go in here?

mikehearn:
I had a chat with Kat about this - I think this case can be fixed in a much simpler and less invasive way for v3. We should do a scan to find other cases like this ahead of time, and we discussed that as well. 

We _might_ want the general feature later on,  so this work isn't necessarily wasted, but as it's more complex and will take longer to get through review I'd like to punt it for now. I'll start work on the other approach.

core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
```diff
@@ -5,46 +5,108 @@ import net.corda.core.flows.FlowException
...
+    constructor (txId: SecureHash, message: String, cause: Throwable?) : super(message, txId, cause)
+    constructor (message: String, cause: Throwable?) : super(message, cause)
 
-    class ContractConstraintRejection(txId: SecureHash, contractClass: String)
-        : TransactionVerificationException(txId, "Contract constraints failed for $contractClass", null)
+    class ContractRejection(
+            override val txId: SecureHash,
+            message: String,
+            cause: Throwable?,
+            @UnusedConstructorParameter p: SerializationOnlyParameter?)
```
rick-r3: Do we need both an annotation and a type?  We were liking the type because it shows more clearly in IDE auto-complete.  Also, the type could do with a better name perhaps.  It's a *de*serialization parameter.  Is it insane to call it `DoNotUseDeserializationOnlyConstructorDisambiguationParameter` !!??

core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
```diff
@@ -5,46 +5,108 @@ import net.corda.core.flows.FlowException
...
+            override val txId: SecureHash, message: String, @UnusedConstructorParameter p: SerializationOnlyParameter?)
+        : TransactionVerificationException(message, null) {
+
+        constructor (txId: SecureHash, contractClass: String) :
+                this(txId, "Contract constraints failed for $contractClass", null)
+    }
+
+    class MissingAttachmentRejection(
+            override val txId: SecureHash, message: String, @UnusedConstructorParameter p: SerializationOnlyParameter?)
+        : TransactionVerificationException(message, null) {
```
rick-r3: Do we need the `cause` to be set or does the exception serializer deal with that aspect?

core/src/main/kotlin/net/corda/core/flows/FlowException.kt
```diff
@@ -14,6 +15,7 @@ import net.corda.core.CordaRuntimeException
  * It is recommended a [FlowLogic] document the [FlowException] types it can throw.
  */
 open class FlowException(message: String?, cause: Throwable?) : CordaException(message, cause) {
+    constructor(message: String?, txId: SecureHash, cause: Throwable?) : this ("$message, transaction: $txId", cause)
```
rick-r3: Do we need this?  Or you thinking it's just useful?

docs/source/serialization.rst
```diff
@@ -442,6 +442,94 @@ associates it with the actual member variable.
             fun getStatesToConsume() = states
         }
 
+Constructors with Side Effects
+``````````````````````````````
+
+**The Problem**
+
+Given the Corda AMQP library uses constructors to instantiate objects it is advised to avoid using
```
rick-r3: ..AMQP **serialization** library..

docs/source/serialization.rst
```diff
@@ -442,6 +442,94 @@ associates it with the actual member variable.
...
+Clearly this is not what was intended as one would expect the two copies to match. Of course, this is
+a highly contrived example to illustrate a point. Where this is more common is in exception handling
+where the input is modified to complete some template message to allow the user of the class to
+not worry about filling that in.
+
+.. sourcecode:: kotlin
+
+    class CatException(name: String) : Exception("Cat $name was a bad girl")
+
+Without modification of the ``CatException`` class this will not behave as intended under, indeed, on
```
rick-r3: end of this line doesn't make sense to me.  remove "under, indeed, " ?

docs/source/serialization.rst
```diff
@@ -442,6 +442,94 @@ associates it with the actual member variable.
...
+
+    throw CatException("Jane")
+
+    // After a serialization round trip the result would be
+    "Cat Cat Jane was a bad girl was a bad girl"
+
+
+**The Solution**
+
+Classes must be written such that the primary constructor is the least specialised form, i.e. that which
```
rick-r3: Does it only need to be primary because it is not annotated with `@DeserializationConstructor` ?  Maybe we should talk about that, or what if they are in Java?

docs/source/serialization.rst
```diff
@@ -442,6 +442,94 @@ associates it with the actual member variable.
...
+takes a pure string and can be used when deserializing the object to instantiate the serialised string
+without alteration.
+
+Equally, a secondary constructor must be provided that can manipulate the input as desired. In cases where
+this can be done without signature clash this is easy. However, when that isn't possible (as in our
+CatException example) then the Corda Serialization framework provides the ``UnusedConstructorParameter`` annotation.
+Additionally, we recommend using the explicit type ``SerializationOnlyParameter`` to further reinforce this
+property is a simple place holder and should never be used outside of the serializer.
```
rick-r3: Perhaps talk about they type being useful for code complete visibility in IDE.  See comment at the top about whether we should also still have the annotation.  The annotation does automatically trigger my "something special going on here" senses, so not completely against it.

docs/source/serialization.rst
```diff
@@ -442,6 +442,94 @@ associates it with the actual member variable.
...
+
+Reworking our naughty cat exception:
+
+.. sourcecode:: kotlin
+
+    class CatException(msg: String, @UnusedConstructorParameter p: SerializationOnlyParameter?) : Exception(msg) {
+        constructor(name: String) : this ("Cat $name was a bad girl", null)
+    }
+
+This still allows the exception to be thrown by users as ``throw CatException(Elaine)``, preserving the clean and
```
rick-r3: Double quotes around `Elaine`?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/PropertySerializers.kt
```diff
@@ -75,6 +82,16 @@ class PrivatePropertyReader(val field: Field, parentType: Type) : PropertyReader
     }
 }
 
+/**
+ * For internal use by the Serializer only
+ *
+ * Used to specifically and purposefully remove a property from the serialised fornm/pre
```
rick-r3: "fornm/pre" ?

node-api-interfaces/src/main/kotlin/net/corda/nodeapiinterfaces/serialization/ConstructorForDeserialization.kt
rick-r3: Wait.... I got further down through the files: So there's a duplicate annotation ?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -42,11 +45,17 @@ internal fun constructorForDeserialization(type: Type): KFunction<Any>? {
...
-            } else if (preferredCandidate == null && kotlinConstructors.size == 2 && hasDefault && kotlinConstructor.parameters.isNotEmpty()) {
+            } else if (preferredCandidate == null
+                    && kotlinConstructors.size == 2
+                    && hasDefault
+                    && kotlinConstructor.parameters.isNotEmpty()) {
                 preferredCandidate = kotlinConstructor
-            } else if (kotlinConstructor.findAnnotation<ConstructorForDeserialization>() != null) {
+            } else if (
+                    kotlinConstructor.findAnnotation<ConstructorForDeserialization>() != null ||
+                    kotlinConstructor.findAnnotation<ConstructorForDeserializationCore>() != null) {
```
rick-r3: I don't like this!  Let's keep the discussion about it further up where the new annotation class is defined.

fenryka:
I'll revisit this after we ship 3.0

node-api-interfaces/src/main/kotlin/net/corda/nodeapiinterfaces/serialization/ConstructorForDeserialization.kt
chrisr3: The `@net.corda.core.CordaInternal` annotation is "magic" in that the API Scanner doesn't write the things that it annotates (just property getters and setters atm, although I'm adding annotations to that list in a separate PR) into the API file.

What I'm suggesting is that the API Scanner consider _all_ `@CordaInternal` annotations to be equally magical, regardless of their package. This would break the implicit - and rather rigid - dependency between the API Scanner and `core`.

PR #2683, opened 2018-03-01
CORDA-1076 - Improves message for structural step changes in ProgressTracker.

joeldudleyr3:


joeldudleyr3:
In relation to https://r3-cev.atlassian.net/browse/CORDA-1076. I went for a relatively minor fix, as the progress tracker is due for a rewrite anyway.

gendal:
Thanks Joel. I agree this is a little less bad but it's still weird...  the progress tracker is supposed to be telling me, in business terms, what's going on.  The other lines do that (getting signatures, preparing transaction, whatever).  What does "stepping into child progress tracker" mean?  That's telling me something about what Corda is doing behind the scenes, not what is happening in my business process.

PR #2686, opened 2018-03-01
1076 - Filters out structural steps when tracking a flow via RPC.

joeldudleyr3:
We can't filter out the structural steps by class, because `FlowLogic.track` is part of the public API and returns strings only.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/ProgressTracker.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/ProgressTracker.kt
@@ -1,3 +1,3 @@
 package net.corda.core.internal
 
-val STRUCTURAL_STEP_TEXT = "Structural step change in child of "
+val STRUCTURAL_STEP_PREFIX = "Structural step change in child of "
--- a/core/src/main/kotlin/net/corda/core/internal/ProgressTracker.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/ProgressTracker.kt
@@ -1,0 +1,3 @@
+package net.corda.core.internal
+
+val STRUCTURAL_STEP_TEXT = "Structural step change in child of "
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/net/corda/core/internal/ProgressTracker.kt
```diff
@@ -0,0 +1,3 @@
+package net.corda.core.internal
+
+val STRUCTURAL_STEP_TEXT = "Structural step change in child of "
```
andr3ej: STRUCTURAL_STEP_PREFIX would be more accurate

node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
```diff
@@ -154,7 +155,7 @@ internal class CordaRPCOpsImpl(
         return FlowProgressHandleImpl(
                 id = stateMachine.id,
                 returnValue = stateMachine.resultFuture,
-                progress = stateMachine.logic.track()?.updates ?: Observable.empty(),
+                progress = stateMachine.logic.track()?.updates?.filter { !(it.contains(STRUCTURAL_STEP_TEXT)) } ?: Observable.empty(),
```
andr3ej: startsWith, and don't need the outer set of round brackets

gendal:
Thanks.  No comment on whether this is the right fix but happy that we won't be outputting this any more!  (Separately: are we happy that ANYTHING is printed to the console of the web server when running examples with cordapp-example ? )

joeldudleyr3:
Well, it's not automatic. It's the way we've defined the web API for this CorDapp.

Are you saying remove the logging, or no web API at all?

gendal:
It was the fact that something seems to be dumping a ton of stuff to the web servers' consoles when running flows.  Maybe it's what we need for debugging, etc... I just didn't recall it from the past.  Maybe it's actually a good thing now.  No firm view sorry!

joeldudleyr3:
NP. But just to be clear, this is due to the way this CorDapp is written, rather than how Corda is written.

PR #2689, opened 2018-03-01
Fix the serialisability of TransactionVerificationException and subtypes.

mikehearn:
Fix the serialisability of TransactionVerificationException and subtypes.
Add KDocs for the exceptions in this file, and deprecate exceptions that aren't actually used.


mikehearn:
Pushed over the integration tests because the only failure was a flaky test.

PR #2691, opened 2018-03-01
Remove CordApps JARs from node classpath [CORDA-1135]

IgorNitto:


IgorNitto:
Isolated classloader changes from the packaging story. 

API compatibility in LedgerTransaction is preserved by making the assumption that if a state reference a contract, the classloader of the state class knows how to load that contract.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -869,13 +869,14 @@ internal class NetworkMapCacheEmptyException : Exception()
 fun configureDatabase(hikariProperties: Properties,
                       databaseConfig: DatabaseConfig,
                       identityService: IdentityService,
-                      schemaService: SchemaService = NodeSchemaService()): CordaPersistence {
+                      schemaService: SchemaService = NodeSchemaService(),
+                      cordappClassLoader: ClassLoader? = null): CordaPersistence {
```
sollecitom: Why default value here?

IgorNitto: So, essentially this classloader is an optional value. When equal to null, Hibernate will exclusively rely on the sys classloader. This is the setting used in unit tests, so i have given a default value to avoid changing them all.

IgorNitto:
Fixed some defects that I could only detect by manually running samples, they were mostly: 
- Deserializer not using the deserialization classloader (MapSerializer)
- Sample cordApps attempting to do reflection via sys classloader. They should only rely on the instance classloader

I patched these kind of defects where their presence was revealed by debugging samples (current smoke tests are totally fine on this). I will spent a bit more time doing manual validation and if no other problem appear  I will merge this PR.

PR #2692, opened 2018-03-01
CORDA-1134 - Don't use private serializes for all caps public properties

fenryka:


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -136,7 +136,16 @@ fun Class<out Any?>.propertyDescriptors(): Map<String, PropertyDescriptor> {
...
+        clazz = clazz.superclass
+    } while (clazz != null)
 
+    //
+    // Running as two loops rather than one as we need to ensure we have captured all of the properties
+    // before looking for interacting methods and need to cope with the class hierarchy introducing
+    // new  properties / methods
+    //
+    clazz = this
+    do {
```
rick-r3: Is there a test of the break out of this into two looks?  i.e. an example that breaks the old way.  Or does that one new test also mask this issue too?

fenryka: Why would there be, it's a behaviour change in that now we pre-load all the properties so when looking them up later for method matching we know the right case to use rather than having to guess at somethign that might not be there

fenryka: It's a change that facilitates the new behaviour rather than the actual issue (that of trying to capitalise or not)

rick-r3: I get it now.

PR #2695, opened 2018-03-01
Fix the serialisability of TransactionVerificationException and subtypes.

mikehearn:
Add KDocs for the exceptions in this file, and deprecate exceptions that aren't actually used.


fenryka:
https://r3-cev.atlassian.net/browse/CORDA-1130

fenryka:
@mikehearn you might want to liaise with @anthonykr3 about the API break here

mikehearn:
I think the API checker is flagging false positives here, something Anthony is working on.

PR #2696, opened 2018-03-01
Improve test failure logging in AMQP ack/nack test

blsemo:
Found some issues with the logging of the message sequence in the AMQP ack/nack test


mnesbit:
Fixed elsewhere

PR #2698, opened 2018-03-02
[CORDA-1136] Update startup message of P2P address and port

thschroeter:


fenryka:
Once merged to master remember to raise a PR against 3.0 :)

PR #2699, opened 2018-03-02
Corda-1136: startup messages

thschroeter:


node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -120,7 +120,7 @@ class ArtemisMessagingServer(private val config: NodeConfiguration,
         }
         // Config driven switch between legacy CORE bridges and the newer AMQP protocol bridges.
         activeMQServer.start()
-        Node.printBasicNodeInfo("Listening on port", p2pPort.toString())
+        log.info("P2P messaging server listening on port $p2pPort")
```
gendal: ah..... now this is interesting...!  So these two lines do actually say something different (in principle)?  ie in a more complex case, the "Advertised P2P address" might be how we are exposed to the outside world and so the host+port could be completely different to the port on which Artemis is listening on the node machine itself?

thschroeter: Yes, that right.

gendal: Perfect - there was indeed method in the madness :)

PR #2705, opened 2018-03-02
CORDA-1141: Kryo client serilization scheme should use AMQP for P2P c…

adagys:
…ontext.

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/KryoClientSerializationScheme.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/KryoClientSerializationScheme.kt
@@ -6,7 +6,7 @@ import net.corda.core.serialization.internal.SerializationEnvironment
 import net.corda.core.serialization.internal.SerializationEnvironmentImpl
 import net.corda.core.serialization.internal.nodeSerializationEnv
 import net.corda.core.utilities.ByteSequence
-import net.corda.nodeapi.internal.serialization.KRYO_P2P_CONTEXT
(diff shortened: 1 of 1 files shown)


rick-r3:
@fenryka Is going to finish this off or raise a replacement, since we need to deal with the potential for custom serialiser registration on the RPC client side, and it's not there yet.

mikehearn:
OK. Can you explain this in a bit more detail in a comment please? Because the change just looks wrong, doesn't it? I'm sure it's right but it's not clear to me why the Kryo part is using "AMQP_CONTEXT". I hope after we finish this work and Kryo is hiding in the corner doing just checkpointing we can simplify this code a lot, because it's quite confusing.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/KryoClientSerializationScheme.kt
```diff
@@ -43,7 +43,7 @@ class KryoClientSerializationScheme : AbstractKryoSerializationScheme() {
                         registerScheme(KryoClientSerializationScheme())
                         registerScheme(AMQPClientSerializationScheme(emptyList()))
                     },
-                    KRYO_P2P_CONTEXT,
+                    AMQP_P2P_CONTEXT,
```
fenryka: Alas it isn't this simple

fenryka (changes requested):
I'll take this over, as discussed, the problem is access to pluggable serialisers for the library, which is tricky given how RPC is enabled all over the place

PR #2707, opened 2018-03-02
Fix another non-serializable exception, add docs, fix a possible security issue.

mikehearn:


node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
```diff
@@ -41,15 +41,24 @@ class FlowLogicRefFactoryImpl(private val classloader: ClassLoader) : SingletonS
...
+            Class.forName(flowClassName, true, classloader)
+        } catch (e: ClassNotFoundException) {
+            throw IllegalFlowLogicException(flowClassName, "Flow not found: $flowClassName")
         }
+        val asSubclass = forName.asSubclass(FlowLogic::class.java)
+        return if (asSubclass != null)
+            asSubclass
+        else
+            throw IllegalFlowLogicException(flowClassName, "The class $flowClassName is not a subclass of FlowLogic.")
     }
```
fenryka: return forName.asSubclass(FlowLogic::class.java) ?:  throw IllegalFlowLogicException(flowClassName, "The class $flowClassName is not a subclass of FlowLogic.")

Would be more idiomatic, plus it's one line instead of 5

node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
```diff
@@ -93,7 +102,9 @@ class FlowLogicRefFactoryImpl(private val classloader: ClassLoader) : SingletonS
 
     override fun toFlowLogic(ref: FlowLogicRef): FlowLogic<*> {
         if (ref !is FlowLogicRefImpl) throw IllegalFlowLogicException(ref.javaClass, "FlowLogicRef was not created via correct FlowLogicRefFactory interface")
-        val klass = Class.forName(ref.flowLogicClassName, true, classloader).asSubclass(FlowLogic::class.java)
+        // We re-validate here because a FlowLogicRefImpl could have arrived via deserialisation and therefore the
```
fenryka: deserialiZation

node/src/main/kotlin/net/corda/node/services/statemachine/FlowLogicRefFactoryImpl.kt
fenryka: If it were me I'd probably drop forName and append a ".apply" to the try block with the remainder of the code within it

mikehearn: I did that at first but it resulted in a very wide line that went over 120 chars, and every possible line break was quite ugly. I can put it back if you like, but I preferred to reduce the width and increase the height here. No big deal either way.

mikehearn:
@fenryka How does it look now?

PR #2717, opened 2018-03-04
CORDA-1160: Only read node-info files if their last modified time has changed.

shamsasari:


node/src/main/kotlin/net/corda/node/services/network/NodeInfoWatcher.kt
```diff
@@ -93,42 +87,32 @@ class NodeInfoWatcher(private val nodePath: Path,
...
-                        val nodeInfo = processFile(path)?.let {
-                            if (_processedNodeInfoHashes.add(it.signed.raw.hash)) it.nodeInfo else null
+                    .filter { file ->
+                        val lastModifiedTime = file.lastModifiedTime()
+                        val previousLastModifiedTime = nodeInfoFiles[file]
+                        val newOrChangedFile = previousLastModifiedTime == null || lastModifiedTime > previousLastModifiedTime
+                        nodeInfoFiles[file] = lastModifiedTime
+                        newOrChangedFile
+                    }
+                    .mapNotNull { file ->
```
patrickkuo: Not sure what's the styling rules here but I usually won't start a new line if there is a `}` in front of the next chain method call. Not a big issue i guess.

Something like this:
```
            paths.filter { it.isRegularFile() }.filter { file ->
                val lastModifiedTime = file.lastModifiedTime()
                val previousLastModifiedTime = nodeInfoFiles[file]
                val newOrChangedFile = previousLastModifiedTime == null || lastModifiedTime > previousLastModifiedTime
                nodeInfoFiles[file] = lastModifiedTime
                newOrChangedFile
            }.mapNotNull { file ->
                logger.debug { "Reading SignedNodeInfo from $file" }
                try {
                    NodeInfoAndSigned(file.readObject())
                } catch (e: Exception) {
                    logger.warn("Unable to read SignedNodeInfo from $file", e)
                    null
                }
            }.toList()
```

shamsasari: That's a bit too cluttered. But I agree my styling isn't ideal.

PR #2727, opened 2018-03-05
Fix public API definition for Gradle Plugin 4.0.4

chrisr3:
Include `vararg` methods and order class/interface/method annotations.


andr3ej:
adding @Clintonio for artifactory stuff

Clintonio:
Except the variable name - I'm not fond of the underscores. But that's only minor and we use them everywhere else already. Up to you if you want to start on a future convention (ie; when we switch to Kotlin script) or continue with this one.

andr3ej:
i don't like them either, but this way contextUrl has the same camelcasing as the variable it's being assigned to. if we're switching to kotlin let's do things the kotlin way at that time

PR #2728, opened 2018-03-05
CORDA-1170: Define and whitelist the Artemis/AMQP application headers that are accepted by Corda

mnesbit:
Currently, we have defined in several places headers that we use in Artemis messages and we blindly copy all headers over the bridges. During review of my out of process bridge work on ENT it was pointed out that we should whitelist control these. Therefore this pushes the constants into nodeapi internal and only copies the defined headers across the AMQP bridges.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/ArtemisMessagingComponent.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/ArtemisMessagingComponent.kt
@@ -7,6 +7,9 @@ import net.corda.core.messaging.MessageRecipients
 import net.corda.core.messaging.SingleMessageRecipient
 import net.corda.core.serialization.CordaSerializable
 import net.corda.core.utilities.NetworkHostAndPort
+import net.corda.nodeapi.internal.ArtemisMessagingComponent.Companion.P2PMessagingHeaders.Type.KEY
+import org.apache.activemq.artemis.api.core.Message
+import org.apache.activemq.artemis.api.core.SimpleString
 import java.security.PublicKey
 
 /**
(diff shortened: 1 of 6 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/ArtemisMessagingComponent.kt
```diff
@@ -28,6 +31,49 @@ class ArtemisMessagingComponent {
...
+             */
+            val bridgedCertificateSubject = SimpleString("sender-subject-name")
+
+
+            object Type {
+                const val KEY = "corda_p2p_message_type"
+                const val SESSION_INIT_VALUE = "session_init"
+            }
+
+            val whiteListedHeaders: Set<String> = setOf(topicProperty.toString(),
```
exFalso: tiny thing: shouldn't it be lowercase L

node-api/src/main/kotlin/net/corda/nodeapi/internal/ArtemisMessagingComponent.kt
```diff
@@ -7,6 +7,9 @@ import net.corda.core.messaging.MessageRecipients
 import net.corda.core.messaging.SingleMessageRecipient
 import net.corda.core.serialization.CordaSerializable
 import net.corda.core.utilities.NetworkHostAndPort
+import net.corda.nodeapi.internal.ArtemisMessagingComponent.Companion.P2PMessagingHeaders.Type.KEY
```
exFalso: Weird self-import, perhaps just qualify `Type.KEY` instead?

PR #2729, opened 2018-03-05
CORDA-1006: Undoing the wiring of maxMessageSize as it's not correctl…

shamsasari:
…y implemented and updating the docs to clarify its status. (#2501)

The network parameter was just fed into Artemis' minLargeMessageSize property which isn't the same thing.

(cherry picked from commit 49f75da)


shamsasari:
Porting this onto master as well so that the ENT release also has maxMessageSize unwired.

PR #2731, opened 2018-03-05
Allow cordapps to specify flows to run on node startup

roastario:


egeek:
Any updates on this @roastario ? It would be useful to have this for some of the stuff we have been working on today

roastario:
So it's a bit more complex than I originally thought. The complexity is
around the flows being suspended and checkpointed. It's possible that some
failure scenarios would result in constant restarts which could cause 1000s
of flows to be invoked and checkpointed to the DB.

I'll have to talk to the engineering leads and work on some ideas.

Will come back asap.

mikehearn:
@roastario Can you start a thread on corda-dev about this please? I'm not sure what the background to this feature is and the PR description ..... didn't help :)

PR #2732, opened 2018-03-05
[CORDA-1156]: Output of run networkMapFeed and run networkMapSnapshot not consistent in shell (fixed)

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1156

### DIFF
--- a/node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
+++ b/node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
@@ -1,6 +1,5 @@
 package net.corda.node.shell
 
-import com.fasterxml.jackson.core.JsonFactory
 import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.core.JsonParser
 import com.fasterxml.jackson.databind.*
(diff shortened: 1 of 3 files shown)


node/src/main/kotlin/net/corda/node/shell/InteractiveShell.kt
```diff
@@ -431,25 +465,32 @@ object InteractiveShell {
...
-        val subscriber = PrintingSubscriber(printerFun, toStream)
-        uncheckedCast(observable).subscribe(subscriber)
+        if (response is DataFeed<*, *>) {
+            out.println("Snapshot:")
+            out.println(printerFun(response.snapshot))
+            out.flush()
+            out.println("Updates:")
+            return printNextElements(response.updates, printerFun, out)
+        }
+        return printNextElements(response as Observable<*>, printerFun, out)
```
szymonsztuka: Just minor suggestion, maybe reorder conditions to:
if (response is DataFeed<*, *>)  .....
if (response is Observable<*>) .....
else //case for the rest

so we avoid `response !is Observable<*> && response !is DataFeed<*, *>`

mikehearn:
Please don't make un-announced changes to user visible features like this without even mentioning it in the PR description - a big part of the shell UI changed here and it wasn't announced anywhere!

PR #2744, opened 2018-03-06
Docs numbering issue and consistency.

kchalkias:


kchalkias:
@mnesbit addressed your comment on counter party with space or underscore. Now we use **counterpart** and **counterparty** everywhere.

PR #2746, opened 2018-03-06
CORDA-1171: Fix potential privacy leak in notary double spend exceptions

adagys:
When a double-spend occurs, do not send the consuming transaction id and requesting party back to the client - this might lead to privacy leak. Only the transaction id hash is now returned.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -1373,12 +1373,10 @@ public static final class net.corda.core.flows.NotarisationRequest$Companion ext
 @net.corda.core.serialization.CordaSerializable public abstract class net.corda.core.flows.NotaryError extends java.lang.Object
 ##
 @net.corda.core.serialization.CordaSerializable public static final class net.corda.core.flows.NotaryError$Conflict extends net.corda.core.flows.NotaryError
-  public <init>(net.corda.core.crypto.SecureHash, net.corda.core.crypto.SignedData)
   @org.jetbrains.annotations.NotNull public final net.corda.core.crypto.SecureHash component1()
-  @org.jetbrains.annotations.NotNull public final net.corda.core.crypto.SignedData component2()
-  @org.jetbrains.annotations.NotNull public final net.corda.core.flows.NotaryError$Conflict copy(net.corda.core.crypto.SecureHash, net.corda.core.crypto.SignedData)
+  @org.jetbrains.annotations.NotNull public final Map component2()
+  @org.jetbrains.annotations.NotNull public final net.corda.core.flows.NotaryError$Conflict copy(net.corda.core.crypto.SecureHash, Map)
   public boolean equals(Object)
-  @org.jetbrains.annotations.NotNull public final net.corda.core.crypto.SignedData getConflict()
   @org.jetbrains.annotations.NotNull public final net.corda.core.crypto.SecureHash getTxId()
   public int hashCode()
   @org.jetbrains.annotations.NotNull public String toString()
@@ -1429,13 +1427,13 @@ public static final class net.corda.core.flows.NotaryError$TimeWindowInvalid$Com
   public static final net.corda.core.flows.NotaryError$WrongNotary INSTANCE
 ##
 @net.corda.core.serialization.CordaSerializable public final class net.corda.core.flows.NotaryException extends net.corda.core.flows.FlowException
-  public <init>(net.corda.core.flows.NotaryError)
+  public <init>(net.corda.core.flows.NotaryError, net.corda.core.crypto.SecureHash)
   @org.jetbrains.annotations.NotNull public final net.corda.core.flows.NotaryError getError()
 ##
 public final class net.corda.core.flows.NotaryFlow extends java.lang.Object
   public <init>()
 ##
-@net.corda.core.flows.InitiatingFlow public static class net.corda.core.flows.NotaryFlow$Client extends net.corda.core.flows.FlowLogic
+@net.corda.core.flows.InitiatingFlow @net.corda.core.DoNotImplement public static class net.corda.core.flows.NotaryFlow$Client extends net.corda.core.flows.FlowLogic
   public <init>(net.corda.core.transactions.SignedTransaction)
   public <init>(net.corda.core.transactions.SignedTransaction, net.corda.core.utilities.ProgressTracker)
   @co.paralleluniverse.fibers.Suspendable @org.jetbrains.annotations.NotNull public List call()
--- a/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
@@ -1,8 +1,7 @@
 package net.corda.core.flows
 
 import net.corda.core.contracts.StateRef
-import net.corda.core.crypto.DigitalSignature
-import net.corda.core.crypto.SecureHash
+import net.corda.core.crypto.*
 import net.corda.core.identity.Party
 import net.corda.core.serialization.CordaSerializable
 import net.corda.core.serialization.serialize
@@ -43,7 +42,7 @@ class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: Se
         val signature = requestSignature.digitalSignature
         if (intendedSigner.owningKey != signature.by) {
             val errorMessage = "Expected a signature by ${intendedSigner.owningKey}, but received by ${signature.by}}"
-            throw NotaryException(NotaryError.RequestSignatureInvalid(IllegalArgumentException(errorMessage)))
+            throw NotaryInternalException(NotaryError.RequestSignatureInvalid(IllegalArgumentException(errorMessage)))
         }
         // TODO: if requestSignature was generated over an old version of NotarisationRequest, we need to be able to
         // reserialize it in that version to get the exact same bytes. Modify the serialization logic once that's
@@ -59,7 +58,7 @@ class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: Se
             when (e) {
                 is InvalidKeyException, is SignatureException -> {
                     val error = NotaryError.RequestSignatureInvalid(e)
-                    throw NotaryException(error)
+                    throw NotaryInternalException(error)
                 }
                 else -> throw e
             }
@@ -99,3 +98,7 @@ data class NotarisationPayload(val transaction: Any, val requestSignature: Notar
      */
     val coreTransaction get() = transaction as CoreTransaction
 }
+
+/** Payload returned by the notary service flow to the client. */
+@CordaSerializable
+data class NotarisationResponse(val signatures: List<TransactionSignature>)
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -1,27 +1,24 @@
 package net.corda.core.flows
 
 import co.paralleluniverse.fibers.Suspendable
+import net.corda.core.DoNotImplement
 import net.corda.core.contracts.StateRef
 import net.corda.core.contracts.TimeWindow
 import net.corda.core.crypto.SecureHash
-import net.corda.core.crypto.SignedData
 import net.corda.core.crypto.TransactionSignature
-import net.corda.core.crypto.keys
 import net.corda.core.identity.Party
 import net.corda.core.internal.FetchDataFlow
 import net.corda.core.internal.generateSignature
+import net.corda.core.internal.validateSignatures
 import net.corda.core.node.services.NotaryService
 import net.corda.core.node.services.TrustedAuthorityNotaryService
-import net.corda.core.node.services.UniquenessProvider
 import net.corda.core.serialization.CordaSerializable
-import net.corda.core.transactions.CoreTransaction
 import net.corda.core.transactions.ContractUpgradeWireTransaction
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.transactions.WireTransaction
 import net.corda.core.utilities.ProgressTracker
 import net.corda.core.utilities.UntrustworthyData
 import net.corda.core.utilities.unwrap
-import java.security.SignatureException
 import java.time.Instant
 import java.util.function.Predicate
 
@@ -36,6 +33,7 @@ class NotaryFlow {
      * @throws NotaryException in case the any of the inputs to the transaction have been consumed
      *                         by another transaction or the time-window is invalid.
      */
+    @DoNotImplement
     @InitiatingFlow
     open class Client(private val stx: SignedTransaction,
                       override val progressTracker: ProgressTracker) : FlowLogic<List<TransactionSignature>>() {
@@ -68,44 +66,32 @@ class NotaryFlow {
             check(serviceHub.loadStates(stx.inputs.toSet()).all { it.state.notary == notaryParty }) {
                 "Input states must have the same Notary"
             }
-
-            try {
-                stx.resolveTransactionWithSignatures(serviceHub).verifySignaturesExcept(notaryParty.owningKey)
-            } catch (ex: SignatureException) {
-                throw NotaryException(NotaryError.TransactionInvalid(ex))
-            }
+            stx.resolveTransactionWithSignatures(serviceHub).verifySignaturesExcept(notaryParty.owningKey)
             return notaryParty
         }
 
         /** Notarises the transaction with the [notaryParty], obtains the notary's signature(s). */
         @Throws(NotaryException::class)
         @Suspendable
-        protected fun notarise(notaryParty: Party): UntrustworthyData<List<TransactionSignature>> {
-            return try {
-                val session = initiateFlow(notaryParty)
-                val requestSignature = NotarisationRequest(stx.inputs, stx.id).generateSignature(serviceHub)
-                if (serviceHub.networkMapCache.isValidatingNotary(notaryParty)) {
-                    sendAndReceiveValidating(session, requestSignature)
-                } else {
-                    sendAndReceiveNonValidating(notaryParty, session, requestSignature)
-                }
-            } catch (e: NotaryException) {
-                if (e.error is NotaryError.Conflict) {
-                    e.error.conflict.verified()
-                }
-                throw e
+        protected fun notarise(notaryParty: Party): UntrustworthyData<NotarisationResponse> {
+            val session = initiateFlow(notaryParty)
+            val requestSignature = NotarisationRequest(stx.inputs, stx.id).generateSignature(serviceHub)
+            return if (serviceHub.networkMapCache.isValidatingNotary(notaryParty)) {
+                sendAndReceiveValidating(session, requestSignature)
+            } else {
+                sendAndReceiveNonValidating(notaryParty, session, requestSignature)
             }
         }
 
         @Suspendable
-        private fun sendAndReceiveValidating(session: FlowSession, signature: NotarisationRequestSignature): UntrustworthyData<List<TransactionSignature>> {
+        private fun sendAndReceiveValidating(session: FlowSession, signature: NotarisationRequestSignature): UntrustworthyData<NotarisationResponse> {
             val payload = NotarisationPayload(stx, signature)
             subFlow(NotarySendTransactionFlow(session, payload))
             return session.receive()
         }
 
         @Suspendable
-        private fun sendAndReceiveNonValidating(notaryParty: Party, session: FlowSession, signature: NotarisationRequestSignature): UntrustworthyData<List<TransactionSignature>> {
+        private fun sendAndReceiveNonValidating(notaryParty: Party, session: FlowSession, signature: NotarisationRequestSignature): UntrustworthyData<NotarisationResponse> {
             val ctx = stx.coreTransaction
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
@@ -116,16 +102,11 @@ class NotaryFlow {
         }
 
         /** Checks that the notary's signature(s) is/are valid. */
-        protected fun validateResponse(response: UntrustworthyData<List<TransactionSignature>>, notaryParty: Party): List<TransactionSignature> {
-            return response.unwrap { signatures ->
-                signatures.forEach { validateSignature(it, stx.id, notaryParty) }
-                signatures
+        protected fun validateResponse(response: UntrustworthyData<NotarisationResponse>, notaryParty: Party): List<TransactionSignature> {
+            return response.unwrap {
+                it.validateSignatures(stx.id, notaryParty)
+                it.signatures
             }
-        }
-
-        private fun validateSignature(sig: TransactionSignature, txId: SecureHash, notaryParty: Party) {
-            check(sig.by in notaryParty.owningKey.keys) { "Invalid signer for the notary result" }
-            sig.verify(txId)
         }
 
         /**
@@ -156,11 +137,17 @@ class NotaryFlow {
             check(serviceHub.myInfo.legalIdentities.any { serviceHub.networkMapCache.isNotary(it) }) {
                 "We are not a notary on the network"
             }
-            val (id, inputs, timeWindow, notary) = receiveAndVerifyTx()
-            checkNotary(notary)
-            service.validateTimeWindow(timeWindow)
-            service.commitInputStates(inputs, id, otherSideSession.counterparty)
-            signAndSendResponse(id)
+            var txId: SecureHash? = null
+            try {
+                val parts = receiveAndVerifyTx()
+                txId = parts.id
+                checkNotary(parts.notary)
+                service.validateTimeWindow(parts.timestamp)
+                service.commitInputStates(parts.inputs, txId, otherSideSession.counterparty)
+                signTransactionAndSendResponse(txId)
+            } catch (e: NotaryInternalException) {
+                throw NotaryException(e.error, txId)
+            }
             return null
         }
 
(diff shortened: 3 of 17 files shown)


core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -203,8 +204,15 @@ class NotaryException(val error: NotaryError) : FlowException("Unable to notaris
 @CordaSerializable
 sealed class NotaryError {
     /** Occurs when one or more input states of transaction with [txId] have already been consumed by another transaction. */
-    data class Conflict(val txId: SecureHash, val conflict: SignedData<UniquenessProvider.Conflict>) : NotaryError() {
+    data class Conflict(val txId: SecureHash,
+                        val signedConflict: SignedData<DoubleSpendConflict>
+    ) : NotaryError() {
         override fun toString() = "One or more input states for transaction $txId have been used in another transaction"
+
+        @Deprecated("No longer populated due to potential privacy issues", ReplaceWith("Use signedConflict property instead"))
```
adagys: deprecating the `stateHistory` property in `UniquenessProvider.Conflict` would technically be lower impact, but `UniquenessProvided` will have to be deprecated altogether in the future (due to required changes in the `commit` interface), so I chose to deprecate `UniquenessProvider.Conflict` instead.

core/src/main/kotlin/net/corda/core/internal/NotaryUtils.kt
```diff
@@ -0,0 +1,21 @@
...
+import net.corda.core.crypto.Crypto
+import net.corda.core.crypto.SignedData
+import net.corda.core.crypto.sign
+import net.corda.core.flows.DoubleSpendConflict
+import net.corda.core.node.services.UniquenessProvider
+import net.corda.core.serialization.serialize
+
+/** Creates an empty signed UniquenessProvider.Conflict. Exists purely for backwards compatibility purposes. */
+@Suppress("DEPRECATION")
+internal val signedEmptyUniquenessConflict: SignedData<UniquenessProvider.Conflict> by lazy {
```
kchalkias: We don't need a new keyGen and signature every time, right?
Just use a static SignedData object and always return this.

adagys: It's lazy, so will only be called at most once - same effect as a static object, but we don't initialize if not needed

kchalkias: Ok, is there any reason that we should know/store the private key? Even if `lazy` it will be different each time you restart.

adagys: Nope, this is purely so we can extract the underlying object from signed data

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -203,8 +204,15 @@ class NotaryException(val error: NotaryError) : FlowException("Unable to notaris
...
+        @Suppress("DEPRECATION")
```
shamsasari: why suppress?

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -203,8 +204,15 @@ class NotaryException(val error: NotaryError) : FlowException("Unable to notaris
 @CordaSerializable
 sealed class NotaryError {
     /** Occurs when one or more input states of transaction with [txId] have already been consumed by another transaction. */
-    data class Conflict(val txId: SecureHash, val conflict: SignedData<UniquenessProvider.Conflict>) : NotaryError() {
+    data class Conflict(val txId: SecureHash,
+                        val signedConflict: SignedData<DoubleSpendConflict>
```
shamsasari: `doubleSpendConflict`?

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -82,10 +83,10 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
...
         try {
             uniquenessProvider.commit(inputs, txId, caller)
-        } catch (e: UniquenessException) {
-            val conflicts = inputs.filterIndexed { i, stateRef ->
-                val consumingTx = e.error.stateHistory[stateRef]
-                consumingTx != null && consumingTx != UniquenessProvider.ConsumingTx(txId, i, caller)
+        } catch (e: DoubleSpendException) {
+            val conflicts = inputs.filterIndexed { _, stateRef ->
+                val cause = e.error.stateConflicts[stateRef]
+                cause != null && cause.transactionIdHash != txId.sha256()
```
kchalkias: Is it practically possible to return `txId.sha256()` or this is just a precaution check?

core/src/main/kotlin/net/corda/core/node/services/UniquenessProvider.kt
```diff
@@ -18,19 +18,17 @@ interface UniquenessProvider {
 
     /** Specifies the consuming transaction for every conflicting state */
```
kchalkias: comment full stop

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
```diff
@@ -82,10 +83,10 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
     fun commitInputStates(inputs: List<StateRef>, txId: SecureHash, caller: Party) {
         try {
             uniquenessProvider.commit(inputs, txId, caller)
-        } catch (e: UniquenessException) {
-            val conflicts = inputs.filterIndexed { i, stateRef ->
-                val consumingTx = e.error.stateHistory[stateRef]
-                consumingTx != null && consumingTx != UniquenessProvider.ConsumingTx(txId, i, caller)
+        } catch (e: DoubleSpendException) {
```
shamsasari: `DoubleSpendException` can't be internal if `commit` can throw it.

core/src/main/kotlin/net/corda/core/internal/NotaryUtils.kt
```diff
@@ -0,0 +1,21 @@
...
+    val key = Crypto.generateKeyPair()
+    val emptyConflict = UniquenessProvider.Conflict(emptyMap()).serialize()
+    val signature = key.sign(emptyConflict)
+    SignedData(emptyConflict, signature)
+}
+
+/** An internal exception used for propagating the error from the notary service to the notary flow. */
+class DoubleSpendException(val error: DoubleSpendConflict) : CordaException(DoubleSpendException::class.java.name)
```
shamsasari: Just pass in null for the message.

gendal:
This feels like a big change so close to ship. Is it absolutely required for 3.0? I can't help wondering that this fundamentally changes one of the foundations of our design.  Can you confirm (if you haven't already) that this change does not change any of the assumptions underpinning our threat model?  Is @mikehearn onboard with this?

adagys:
There a three subtle changes in this PR:

1. Clients no longer receive the consuming party identity & transaction id on double spend, only the _hash_ of the transaction id. This prevents possible phishing attacks, which are otherwise hard to detect. The notary still records and persists the identity and transaction id, in addition to a request signature (TODO) - nothing has changed in that regard.
2. As a consequence of (1), transaction re-notarisation requirement has been relaxed: a notary will resign the same transaction as long as the id matches, the caller identities no longer need to mach. This has the added benefit that counterparties can obtain the notary signature even if the coordinator is crashed or did not send it back for some other reasons.
3. There's been some outstanding for BFT notary error management: previously we only returned a single error to the client, with a signature in case it's a double spend conflict. However, that would allow a malicious replica to forge the wrong error message, potentially containing wrong conflict information. Now in addition to the "outcome" error, the notary service flow also attaches a list of errors received from all/majority replicas. We also now sign over every error type. rather than just double spend conflict. These signed errors are also propagated back to the caller, so they can prove to counterparties that the notary indeed returned that particular error response.

Fixing 1 requires some low impact notary exception API changes (it's backwards compatible, but some upgrade logic might be required). Fixing 3 later would require changing the same API yet again, which I'd really like to avoid.

In general, this work does not change how transactions are validated or what error types the notary returns, it's mainly about what these error objects contain and how they are signed - which is all contained withing the notary flows/services. We could do this after 3.0, but then the backwards-compatibility impact would be higher.

core/src/main/kotlin/net/corda/core/node/services/NotaryService.kt
adagys: not exactly sure what you mean, we check here if the tx id hashes match, and don't report a conflict if so

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
adagys: even though it's deprecated, it refers to a deprecated class which results in a deprecation warning..

adagys:
Ready for another look

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -197,14 +210,39 @@ data class TransactionParts(val id: SecureHash, val inputs: List<StateRef>, val
...
+/** Exception internal to the notary service. Does not get exposed to CorDapps and flows calling [NotaryFlow.Client]. */
+class InternalNotaryException(val error: NotaryError) : FlowException("Unable to notarise: $error")
 
 /** Specifies the cause for notarisation request failure. */
 @CordaSerializable
 sealed class NotaryError {
     /** Occurs when one or more input states of transaction with [txId] have already been consumed by another transaction. */
-    data class Conflict(val txId: SecureHash, val conflict: SignedData<UniquenessProvider.Conflict>) : NotaryError() {
+    data class Conflict(
+            /** If of the transaction that was attempted to be notarised. */
```
kchalkias: **Id** of the tx

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -67,7 +70,8 @@ class PersistentUniquenessProvider : UniquenessProvider, SingletonSerializeAsTok
                         toPersistentEntityKey = { PersistentStateRef(it.txhash.toString(), it.index) },
                         fromPersistentEntity = {
                             //TODO null check will become obsolete after making DB/JPA columns not nullable
-                            var txId = it.id.txId ?: throw IllegalStateException("DB returned null SecureHash transactionId")
+                            var txId = it.id.txId
```
kchalkias: `txID` and `index` should be vals

.ci/api-current.txt
```diff
@@ -1373,10 +1373,10 @@ public static final class net.corda.core.flows.NotarisationRequest$Companion ext
 @net.corda.core.serialization.CordaSerializable public abstract class net.corda.core.flows.NotaryError extends java.lang.Object
 ##
 @net.corda.core.serialization.CordaSerializable public static final class net.corda.core.flows.NotaryError$Conflict extends net.corda.core.flows.NotaryError
-  public <init>(net.corda.core.crypto.SecureHash, net.corda.core.crypto.SignedData)
```
rick-r3: Is there not an expectation that the API remain the same?

node/src/test/kotlin/net/corda/node/services/transactions/ValidatingNotaryServiceTests.kt
```diff
@@ -80,14 +79,13 @@ class ValidatingNotaryServiceTests {
             aliceNode.services.signInitialTransaction(tx)
         }
 
-        val ex = assertFailsWith(NotaryException::class) {
+        // Expecting SignaturesMissingException instead of NotaryException, since the exception should originate from
+        // the client flow.
+        val ex = assertFailsWith(SignedTransaction.SignaturesMissingException::class) {
```
kchalkias: assertFailsWith<SignedTransaction.SignaturesMissingException> is shorter

.ci/api-current.txt
adagys: It can't unfortunately. However, this only affects notary service implementations in practice, for which we don't have stable APIs yet..

node/src/main/kotlin/net/corda/node/services/transactions/BFTNonValidatingNotaryService.kt
```diff
@@ -150,13 +150,16 @@ class BFTNonValidatingNotaryService(
...
-                if (notary !in services.myInfo.legalIdentities) throw NotaryException(NotaryError.WrongNotary)
+                if (notary !in services.myInfo.legalIdentities) throw InternalNotaryException(NotaryError.WrongNotary)
                 commitInputStates(inputs, id, callerIdentity)
                 log.debug { "Inputs committed successfully, signing $id" }
                 BFTSMaRt.ReplicaResponse.Signature(sign(id))
-            } catch (e: NotaryException) {
+            } catch (e: InternalNotaryException) {
                 log.debug { "Error processing transaction: ${e.error}" }
-                BFTSMaRt.ReplicaResponse.Error(e.error)
+                val serializedErorr = e.error.serialize()
```
kchalkias: serializedE**rr**or

adagys:
Any more feedback?

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
```diff
@@ -43,7 +42,7 @@ class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: Se
         val signature = requestSignature.digitalSignature
         if (intendedSigner.owningKey != signature.by) {
             val errorMessage = "Expected a signature by ${intendedSigner.owningKey}, but received by ${signature.by}}"
-            throw NotaryException(NotaryError.RequestSignatureInvalid(IllegalArgumentException(errorMessage)))
+            throw InternalNotaryException(NotaryError.RequestSignatureInvalid(IllegalArgumentException(errorMessage)))
```
shamsasari: This exception is not internal

shamsasari: However, how can this be internal when it's thrown by `verifySignature` which is public?

adagys: By internal, here it means that it's never visible outside the notary service. We have public helper methods since BFT and non-BFT notaries don't fit the same interfaces but require shared logic

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -68,19 +69,14 @@ class NotaryFlow {
...
-            }
+            stx.resolveTransactionWithSignatures(serviceHub).verifySignaturesExcept(notaryParty.owningKey)
             return notaryParty
         }
 
         /** Notarises the transaction with the [notaryParty], obtains the notary's signature(s). */
         @Throws(NotaryException::class)
         @Suspendable
-        protected fun notarise(notaryParty: Party): UntrustworthyData<List<TransactionSignature>> {
+        protected fun notarise(notaryParty: Party): UntrustworthyData<NotarisationResponse> {
```
shamsasari: This is API breakage. But I seem to recall you've already broken the notary API in this release. Is that correct?

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
adagys: The function of this exception is to propagate errors from uniqueness providers/verification logic up to the notary service flow

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
adagys: Yes, the wire format changes required modifying the API for custom notaries. We can't really keep it stable yet.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -197,14 +210,39 @@ data class TransactionParts(val id: SecureHash, val inputs: List<StateRef>, val
...
+class NotaryException(
+        val error: NotaryError,
+        /**
+         * Original signed responses from the notary cluster. This can be used as proof to counterparties for a particular
+         * notary response.
+         *
+         * For crash fault-tolerant notaries, the list will only contain one element, since we trust all replicas.
+         * For BFT case, it will contain signed error responses from the majority of the cluster.
+         */
+        val signedErrorResponses: List<SignedData<NotaryError>>
```
shamsasari: From your check above, you're expecting all the members of this list to be the same type as `error`. What exactly is `error` if there's more than one element in this list? Can `error` not just be the first element?

shamsasari: And why is `error` not signed?

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -197,14 +210,39 @@ data class TransactionParts(val id: SecureHash, val inputs: List<StateRef>, val
...
+class InternalNotaryException(val error: NotaryError) : FlowException("Unable to notarise: $error")
 
 /** Specifies the cause for notarisation request failure. */
 @CordaSerializable
 sealed class NotaryError {
     /** Occurs when one or more input states of transaction with [txId] have already been consumed by another transaction. */
-    data class Conflict(val txId: SecureHash, val conflict: SignedData<UniquenessProvider.Conflict>) : NotaryError() {
+    data class Conflict(
+            /** Id of the transaction that was attempted to be notarised. */
+            val txId: SecureHash,
```
shamsasari: Why does this need the tx id and not the other errors?

core/src/main/kotlin/net/corda/core/internal/NotaryUtils.kt
shamsasari: Technically the `internal` keyword isn't needed

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
adagys: the `signedErrorResponses` can contain false errors from malicious replicas as well, so we really need to check that there are enough errors of the same type to form majority. Right now we assume they are all the same, so yes, the error could be the first element.

The main reason `error` is there is so we can populate the `FlowException` error message. In theory we could derive it from the signed responses on demand.

core/src/main/kotlin/net/corda/core/flows/NotarisationRequest.kt
shamsasari: `verifySignature` needs the `@Throws` annotation

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
adagys: Hmm, good point, ideally we'd need the id for all errors, so it gets signed as well.

adagys: Otherwise a malicious replica could just pass on old signed error objects from other replicas!

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
shamsasari: Should we not do that now? I think it makes for `error` to be a getter which contractually returns the first element.

adagys: But then you can't pass it to the superclass constructor

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
adagys: I'll change the exception to have a signed over txid, notaryError pair

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
shamsasari: You mean this: "Unable to notarise: $error"? How about "Unable to notarise: ${signedErrorResponse.map { it.verified() }}"?

adagys: Hmm, yeah I guess we could take the first element as a good guess

adagys:
Thanks for the comments, Mike.

1. The problem is slightly different now than the one mentioned in the TODO. You can't retrieve the full transaction, but you can reconstruct the transaction graph, with associated consuming parties for every transaction. This way business relationships can be revealed. 
2. The biggest issue here is revealing the `requestingParty`, so we'd have to return some placeholder null party. Another danged of revealing the party is allowing someone to de-anonymise their counterparties.
3. The BFT notary error reporting affects what the notary signs, and we should in general sign over the whole error object, which isn't done at present. We could fix this in the future by introducing a separate `BFTNotaryException`, but then CorDapps would have to upgrade and always handle both types of exceptions and we'd lose notary implementation transparency.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -236,3 +274,17 @@ sealed class NotaryError {
         override fun toString() = cause.toString()
     }
 }
+
+/** Contains information about the cause of the double-spend conflict for each of the conflicting input states. */
+@CordaSerializable
+data class DoubleSpendConflict(val conflictCauses: Map<StateRef, Cause>) {
+    @CordaSerializable
+    data class Cause(
```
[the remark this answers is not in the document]
adagys: We might also return the time of the commit (from the notary's perspective). Yes, it's mainly for debugging - a double hash is useful to check if you spent this state somewhere else yourself, or you can provide it to your counterparties to check.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -197,14 +210,39 @@ data class TransactionParts(val id: SecureHash, val inputs: List<StateRef>, val
...
+         * For crash fault-tolerant notaries, the list will only contain one element, since we trust all replicas.
+         * For BFT case, it will contain signed error responses from the majority of the cluster.
+         */
+        val signedErrorResponses: List<SignedData<NotaryError>>
+) : FlowException("Unable to notarise: $error") {
+    constructor(error: NotaryError, signedErrorResponse: SignedData<NotaryError>) : this(error, listOf(signedErrorResponse))
+}
+
+/** Exception internal to the notary service. Does not get exposed to CorDapps and flows calling [NotaryFlow.Client]. */
+class InternalNotaryException(val error: NotaryError) : FlowException("Unable to notarise: $error")
```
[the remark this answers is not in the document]
adagys: it doesn't seem to apply to classes

adagys:
I thought some more about the BFT error reporting, and seems like signing error objects won't fix all problems, de-scoping it for now. This PR is now purely for fixing the potential privacy leak.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -236,3 +244,17 @@ sealed class NotaryError {
         override fun toString() = cause.toString()
     }
 }
+
+/** Contains information about the cause of the double-spend conflict for each of the conflicting input states. */
+@CordaSerializable
+data class DoubleSpendConflict(val conflictCauses: Map<StateRef, Cause>) {
```
rick-r3: There's a separate thread going around a potential new feature that @roger3cev is looking at for reference data. Not going to go into it here, but it would lead to the potential for a conflict without it technically being a double spend.  It would just be "state already consumed".  The old naming didn't use the "double spend" terminology and so it wouldn't have clashed with naming here.

Can we use "already consumed" rather than "double spend" in the naming and comments here?  It also seems more Corda-like and less bitcoin-like.

.ci/api-current.txt
rick-r3: If we are definitely going to change the API, and for a good reason (not breaking privacy), we should announce it in the change log and release notes so if anyone has been writing their own notary they know to adjust their implementation.

adagys: We already have a warning that it's still experimental, but I emphasised the API part as well.

adagys: I'll add something to the change log as a separate PR, there's some changes in previous work as well.

rick-r3: Hmmm.... but `NotaryError` is exposed to applications though, if they have written any code that looks inside `NotaryException` and then inside `Conflict`.

gendal:
Andrius just called me as this is blocking getting RC3 cut.  He explained:

* He and Rick strongly believe this change _is_ needed (for the reasons discussed above, but the primary two being that the current design allows an adversary 1) to walk _forward_ from a State to learn the IDs of transactions that consume it (transitively) and 2) to obtain the real identity of those who consumed states, when only the confidential identity may have been known before

* The alternative of serving dummy data (but not changing the API) is undesirable because any apps that rely on the current behaviour will be broken at runtime in hard-to-understand ways, so better explicitly to break them at compile-time

* If people do turn out to be using this functionality and it needs to be added back in then it can be done in a non-API-breaking way.

So I'm happy for Andrius to get this in on two conditions: 1) if this is an ABI break, as it sounds like it is, he confirms we have OTHER ABI breaks in 3.0 so this doesn't break an otherwise perfect story and 2) he confirms that should we add the function back in that doing so then would not also represent an ABI break

adagys:
So 1) This does seem like the only ABI break as far as I'm aware, 2) Yes we can add the removed information back without an ABI break.

gendal:
The ABI break is just for people who were actually using this information though, right?   And who would silently use fake data if we went with the "return a false TX ID" idea?  If so, then so be it.

I am still worried about this though... such a late-breaking change.  If @mikehearn wakes up and violently disagrees, be prepared to have to revert it and we cut a replacement RC!

mikehearn:
OK then!

PR #2748, opened 2018-03-06
[CORDA-1177]: Add jarDirs compatibility for node configuration parsing

sollecitom:


rick-r3:
Yes please.  We can't deploy perf cluster without this in R3 Corda.

PR #2749, opened 2018-03-06
[DO NOT MERGE] [CORDA-1178] Api scanner backport and update of api-current.txt

anthonykr3:
The api-scanner in the gradle-plugins repo has been modified, which adds extra elements to the api-current.txt file. By porting the version of the api-scanner in gradle-plugins to the v3 release branch, we can align the api-current.txt file between the release branch and live, which will help diagnose api compatibility breakages.

Since the 3.0 release has been cut, we will have this as the first commit in 3.1.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -14,8 +14,6 @@
   public void setMessage(String)
   public void setOriginalExceptionClassName(String)
 ##
-public @interface net.corda.core.CordaInternal
-##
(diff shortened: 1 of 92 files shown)


anthonykr3:
Updated to the most latestest latest version of the api scanner.

chrisr3:
So has been synced with release 4.0.6 of the API Scanner? I think we need a record of the synchronisation point, but apart from that it's fine.

PR #2754, opened 2018-03-06
Further eliminate references to Java target "1.6"

vkolomeyko:
Without this change the following compilation error been observed in IntelliJ:
```
net.corda.nodeapi.internal.config.ConfigUtilities#parseAs(com.typesafe.config.Config, boolean)
Kotlin: Cannot inline bytecode built with JVM target 1.8 into bytecode that is being built with JVM target 1.6. Please specify proper '-jvm-target' option
```

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -60,8 +60,8 @@
       <module name="experimental-kryo-hook_test" target="1.8" />
       <module name="experimental_main" target="1.8" />
       <module name="experimental_test" target="1.8" />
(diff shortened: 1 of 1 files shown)


vkolomeyko:
Thanks @chrisr3 for your input, especially around `Force the Caplet to target Java 6`.
Let me dig a bit deeper and update this PR.

chrisr3:
No worries, but I am also curious why this PR doesn't mention `webserver/webcapsule/build.gradle` as well.

PR #2763, opened 2018-03-07
DO NOT MERGE!!! Rog unspendable inputs

roger3cev:
**UPDATE: After some chats with Rick and Matthew, I've updated the approach, added some more unit tests and tidied up the code.** 

A prototype that adds support for unspendable input states ("UIS"). This is not a finished piece of work... It's just a prototype for the purposes of discussing the right approach.

A UIS is a `contractState` which can be added to a transaction but is not consumed when the transaction is committed to the ledger. The notary still checks to see if the UIS has been previously spent so we can ensure we have the most up-to-date version of the data. This is useful as it allows transaction creators to reference arbitrary on-ledger data in their transactions. Contract code can be used to make assertions over the relationship between the UISs and the other inputs and outputs, and the UISs can be re-used over and over again, as long as they are current. For example:

1. We have two types of state: `Trade` and `LegalEntity`. 
2. `Trade`s are bi-lateral between two `LegalEntities`. As part of a matching process we need to match the trades to `LegalEntity`s. 
3. `LegalEntity` states are issued by one node but need to be used by many nodes.
4. We can create a transaction that takes an input `Trade` with status `UNMATCHED` and an output trade with status `MATCHED` and two UIS - the `LegalEntity` states.
5. The `Trade` contract can refer to the `LegalEntity` UIS in the transaction so can check that the matched `LegalEntity`s are indeed correct.

This is useful because:

* The check is done on the ledger and is verifiable by all parties. There's no need to delegate this responsibility to a trusted third party (an Oracle).
* We can re-use states without creating long chains of provenance and breaching the confidentiality of prior users of the reference states.
* We can be assured that the most up-to-date version of the data is being used.

**How does it work?**

1. Any `ContractState` can be used as an UIS, but typically, states will be specifically created to be used as UISs - reference data, for example. As UIS can be any `ContractState`, I can use a `Cash.State` as a UIS.
 2. UIS can be sent to parties on the network as states usually would be, via the send and receive transactions flows, broadcast transaction flow, etc.
3. The UIS can be created, updated and deleted, as usual, by the parties which are allowed to, as set out in the contract code. E.g. Cash can only be spent by the owner. UISs can be created by any party but can only be updated by the creator.
4. All the transactions types have a new property called `unspendableInputs` which takes a list of `StateRef`s or `StateAndRef`s for `LedgerTransaction`s. States in the `unspendableInputs` list can be referred to during transaction verification. They will **NEVER** be consumed in a transaction, thus they can be used over and over again.
4. Parties which are allowed to can update the UIS. This consumes the state, so any parties using the _old_ version in their transactions will be notified of a conflict by a notary.
5. In the future, DDGs would publish new versions to those parties subscribed to updates. However, in the absence of DDGs, as the creator's `Party` object must be in the reference data state object, those that have out of date versions can request a new version by sending over the old `StateRef` to the `Party` defined in the state. 

**What has changed?**

1. There are some changes to the transaction types to add the `unspendableInputs` list.
2. A new transaction component group has been added.
3. The ledger DSL has been updated to handle UISs.
4. The `Service` side of the notary flow has been updated to check for conflicts involving UISs.
5. **TODO** Update notary tests and transaction tests to include unspendable inputs states.
6. **TODO** Update docs and include with guidelines for usage.

**Comments**

* This requires a platform version upgrade as parties without the `referenceInputs` property in their transaction types will not be able to verify transactions with states in the `unspendableInputs` list.
* Those that are using UISs may run into issues if the UIS is assigned to a different notary to other states. The consumers of UIS cannot change the notary of the UIS, therefore they must move all their states to the UIS notary. Contention may occur if multiple parties which to simultaneously use the same "reference data" state. 
* Chains of provenance will need to be resolved for all UIS. Transaction creators need to be aware that down-stream holders of their transactions will be able to see the UIS. We cannot selectively reveal UIS as this may potentially result in conditional transaction validity.
* Another approach can be taken: UIS can be added to transactions but not resolved or referred to in the contract code. We can do this by making the `unspendableInputs` list private in `LedgerTransaction` but expose a list of `StateRef`s for the notary. This means that transaction creators and committers can add reference data to their transactions so they can be assured of it's currentness but the trade-off is that data cannot be used to validate transactions. 
* What are the privacy concerns around this? E.g. GDPR implications.


rick-r3:
I do prefer the term Reference Input rather than unspendable input.  Latter seems a bit clumsy.

Yes, the big issue here is that verify on old nodes would not have the reference inputs available and thus the outcome could be different from a node that does have them.  Thus we can't allow their use until the minimumPlatformVersion is a version that includes them in transaction resolution and verification.  That seems like a reason to get the supporting code out there sooner rather than have it deter us.  It just needs a latch on their availability that depends on the network parameters minimumPlatformVersion such that they only start working once the entire network can handle them.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -43,11 +45,13 @@ data class ContractUpgradeWireTransaction(
     private val hiddenComponentHash: SecureHash
         get() = serializedHash(listOf(legacyContractAttachmentId, upgradeContractClassName, privacySalt))
 
-    override val id: SecureHash by lazy { serializedHash(inputs + notary).hashConcat(hiddenComponentHash) }
+    override val id: SecureHash by lazy { serializedHash(inputs + referenceInputs + notary).hashConcat(hiddenComponentHash) }
```
mnesbit: Possibly needs to be hashConcat

node/src/test/kotlin/net/corda/node/utilities/TLSAuthenticationTests.kt
```diff
@@ -340,8 +342,8 @@ class TLSAuthenticationTests {
     private fun buildTLSSockets(
             serverSocketFactory: SSLServerSocketFactory,
             clientSocketFactory: SSLSocketFactory,
-            serverPort: Int = 0, // Use 0 to get first free socket.
-            clientPort: Int = 0, // Use 0 to get first free socket.
+            serverPort: Int = 0, // Refer 0 to get first free socket.
```
mnesbit: search/replace error?

core/src/test/kotlin/net/corda/core/crypto/CryptoUtilsTest.kt
```diff
@@ -643,7 +643,7 @@ class CryptoUtilsTest {
         val keyPairEdDSA = Crypto.generateKeyPair(Crypto.EDDSA_ED25519_SHA512)
         val pubEdDSA = keyPairEdDSA.public
         assertTrue(Crypto.publicKeyOnCurve(Crypto.EDDSA_ED25519_SHA512, pubEdDSA))
-        // Use R1 curve for check.
+        // Refer R1 curve for check.
```
mnesbit: Search/replace error?

node/src/test/kotlin/net/corda/node/utilities/TLSAuthenticationTests.kt
roger3cev: Oops. Will fix that.

core/src/test/kotlin/net/corda/core/transactions/UndspendableInputTests.kt
```diff
@@ -0,0 +1,192 @@
...
+                is Create -> requireThat {
+                    "Must have no inputs" using (tx.inputs.isEmpty())
+                    "Must have only one output" using (tx.outputs.size == 1)
+                    val output = tx.outputsOfType<ExampleState>().single()
+                    val signedByCreator = command.signers.single() == output.participants.single().owningKey
+                    "Must be signed by creator" using signedByCreator
+                }
+
+                is Update -> {
+                    "Must have no inputs" using (tx.inputs.size == 1)
```
gendal: typo? must have one input I think?

PR #2767, opened 2018-03-08
CORDA-1192 - Quieten message when reflecting on a builtin Kotlin type

fenryka:


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/PropertySerializers.kt
```diff
@@ -34,7 +34,12 @@ class PublicPropertyReader(private val readMethod: Method?) : PropertyReader() {
             // is: https://youtrack.jetbrains.com/issue/KT-13077
             // TODO: Revisit this when Kotlin issue is fixed.
 
-            loggerFor<PropertySerializer>().error("Unexpected internal Kotlin error", e)
+            // So this used to report as an error, but given we serialise exceptions all the time it
+            // provides for very scary log files so move this to trace level
+            loggerFor<PropertySerializer>().let { logger ->
+                logger.warn("Using kotlin introspection on internal type ${this.declaringClass}")
```
rick-r3: But you still log at warn level?

fenryka: but it doesn't dump a humungous scary stack out unless you set at trace

PR #2772, opened 2018-03-08
CORDA-1197: Take into account last amount submitted when adding data points.

vkolomeyko:
Current logic in `CashWidget` is not handling well updates done in close succession, i.e. less than 1 second.
And such frequent updates do indeed happen, e.g. from: `net.corda.client.jfx.model.ContractStateModel#cashStates`
where `list` is modified twice.

Also use `.toDecimal()` instead of `.quantity`to have amount represented in pounds rather than in pennies.

(cherry picked from commit 952cc35)

### DIFF
--- a/tools/explorer/src/main/kotlin/net/corda/explorer/views/cordapps/cash/CashViewer.kt
+++ b/tools/explorer/src/main/kotlin/net/corda/explorer/views/cordapps/cash/CashViewer.kt
@@ -11,6 +11,7 @@ import javafx.collections.ObservableList
 import javafx.geometry.Insets
 import javafx.scene.Parent
 import javafx.scene.chart.NumberAxis
+import javafx.scene.chart.XYChart
 import javafx.scene.control.*
 import javafx.scene.input.MouseButton
 import javafx.scene.layout.BorderPane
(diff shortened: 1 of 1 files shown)


tools/explorer/src/main/kotlin/net/corda/explorer/views/cordapps/cash/CashViewer.kt
```diff
@@ -313,12 +314,26 @@ class CashViewer : CordaView("Cash") {
             linechart(null, xAxis, yAxis) {
                 series("USD") {
                     sumAmount.addListener { _, _, _ ->
+                        val lastAmount = data.last().value?.yValue
+                        val currAmount = sumAmount.value.toDecimal()
```
kchalkias: y-axis scale is not very helpful if changed from 1000 to 1100 for instance (+100),
As you may notice everything shows 1.1k
image

kchalkias: Also, when I change my preferred currency, the change is not reflected to the old data and as a result in the example below
Bob had 1100£ (at 18:36:40)
then we've changed to $ (at 18:40)
then back to £ (at 18:40:30)
_and here is the result:_
image

PR #2777, opened 2018-03-09
CORDA-1171: When a double-spend occurs, do not send the consuming tra…

adagys:
…nsaction id and requesting party back to the client - this might lead to privacy leak. Only the transaction id hash is now returned.

Merged into release branch as part of https://github.com/corda/corda/pull/2746


adagys:
Rebased without any conflicts

adagys:
Needs a rebase once https://github.com/corda/corda/pull/2776 is in

PR #2780, opened 2018-03-09
ENT-1439 Refactor Provider setup

andr3ej:


core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
```diff
@@ -11,20 +11,12 @@ class CordaSecurityProvider : Provider(PROVIDER_NAME, 0.1, "$PROVIDER_NAME secur
     }
 
     init {
-        AccessController.doPrivileged(PrivilegedAction<Unit> { setup() })
```
andr3ej: why is this needed? it looks copy-pasted from the internet

core/src/main/kotlin/net/corda/core/crypto/internal/ProviderMap.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.bouncycastle.asn1.pkcs.PrivateKeyInfo
+import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo
+import org.bouncycastle.jcajce.provider.util.AsymmetricKeyInfoConverter
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import org.bouncycastle.pqc.jcajce.provider.BouncyCastlePQCProvider
+import java.security.SecureRandom
+import java.security.Security
+
+internal val cordaSecurityProvider = CordaSecurityProvider().also {
+    Security.insertProviderAt(it, 1) // The position is 1-based.
```
andr3ej: change from addProvider should be fine as the current contents of CordaSecurityProvider are specific to it

core/src/main/kotlin/net/corda/core/crypto/internal/ProviderMap.kt
```diff
@@ -0,0 +1,45 @@
...
+    // TODO: Find a way to make JKS work with bouncy castle provider or implement our own provide so we don't have to register bouncy castle provider.
+    Security.addProvider(it)
+}
+// This map is required to defend against users that forcibly call Security.addProvider / Security.removeProvider
+// that could cause unexpected and suspicious behaviour.
+// i.e. if someone removes a Provider and then he/she adds a new one with the same name.
+// The val is private to avoid any harmful state changes.
+internal val providerMap = listOf(cordaBouncyCastleProvider, cordaSecurityProvider, BouncyCastlePQCProvider()).map { it.name to it }.toMap()
+internal val platformSecureRandomFactory: () -> SecureRandom = when {
+    SystemUtils.IS_OS_LINUX -> {
```
andr3ej: this is a little bit more general than the previous manual impl, and i reckon that's what we want

core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
```diff
@@ -199,22 +197,6 @@ object Crypto {
...
             .toMap()
 
-    // This map is required to defend against users that forcibly call Security.addProvider / Security.removeProvider
-    // that could cause unexpected and suspicious behaviour.
-    // i.e. if someone removes a Provider and then he/she adds a new one with the same name.
-    // The val is private to avoid any harmful state changes.
-    private val providerMap: Map<String, Provider> = mapOf(
-            BouncyCastleProvider.PROVIDER_NAME to getBouncyCastleProvider(),
-            CordaSecurityProvider.PROVIDER_NAME to CordaSecurityProvider(),
-            "BCPQC" to BouncyCastlePQCProvider()) // Unfortunately, provider's name is not final in BouncyCastlePQCProvider, so we explicitly set it.
```
andr3ej: seems a bit paranoid, we now use getName on the provider object

core/src/main/kotlin/net/corda/core/crypto/internal/ProviderMap.kt
```diff
@@ -0,0 +1,45 @@
...
+}.also {
+    // This registration is needed for reading back EdDSA key from java keystore.
+    // TODO: Find a way to make JKS work with bouncy castle provider or implement our own provide so we don't have to register bouncy castle provider.
+    Security.addProvider(it)
+}
+// This map is required to defend against users that forcibly call Security.addProvider / Security.removeProvider
+// that could cause unexpected and suspicious behaviour.
+// i.e. if someone removes a Provider and then he/she adds a new one with the same name.
+// The val is private to avoid any harmful state changes.
+internal val providerMap = listOf(cordaBouncyCastleProvider, cordaSecurityProvider, BouncyCastlePQCProvider()).map { it.name to it }.toMap()
```
kchalkias: wrt to `BouncyCastlePQCProvider()` due to an implementation detail in their library, `PROVIDER_NAME` is `public` `static`, but NOT `final`, thus someone might change it before you initiate the Provider object (eg. set it to "BC" and thus override the map entry for regular BouncyCastle.
Probably it's insane to consider a potential attack (or confusion) here, but maybe it's practically safer to specifically set it to `BCPQC`, than `it.name`. wdyt?

rick-r3: How do we force initialisation of this (and thus installation of our provider) very early in the boot sequence for Corda?  It will be initialised if we refer to the `Crypto` object by virtue of calling `cordaSecurityProvider.name` during initialisation of a `val`.  Do we want to be more explicit to we don't accidentally refactor it away at some point.  e.g. should we have `forceInitialise` or some such on `Crypto` and/or in an init that we call really early in node initialisation.

andr3ej: this seems a bit paranoid to me. final isn't really final, you can still change it via reflection like we do in DataSourceFactory. and if someone changed the name of this provider (which they would have to do very early on, as this code runs once at node startup), what would the impact be?

kchalkias: Just wondering if we should always write defensive code on such cases.
As mentioned above, imagine someone changes the name to "BC" (eg in one of the loaded cordapps - which is currently one of the few pluggable things in Corda, but we might favour pluggability in the future for other purposes), then both `cordaBouncyCastleProvider` and `BouncyCastlePQCProvider` share the same name, the last entry overrides the first in the `providerMap`. As a result, `cordaBouncyCastleProvider` is not in the map and everything delegates to BouncyCastlePQCProvider, which might have a faulty implementation of an algorithm that is also present on regular BouncyCastle.

I agree with the "paranoid" statement and it's true reflection can bypass almost everything, but there are people out there who will happily exaggerate and tweet anything they spot (even if it applies on very special unreal conditions; so, better play defensive, especially in the crypto stuff. Btw, there was a related tweet 3 weeks ago, trying to convince the community that our EdDSA is prone to some special side-channel attacks, fortunately we are secure against his claims - he even attached code snippet from Corda !?!?).

andr3ej: ok i've put the check back in, in a more OO way

andr3ej: done, thinking of the UUID case i've made it the first thing in main

PR #2782, opened 2018-03-09
CORDA-1142: Fix explorer to support all transaction types

adagys:
NotaryChange and ContractUpgrade transactions don't contain the output states - they're recalculated on demand from the inputs. I changed the explorer logic to resolve transaction inputs and outputs from vault updates rather than from transactions directly.

### DIFF
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
@@ -1,13 +1,14 @@
 package net.corda.client.jfx.model
 
 import javafx.beans.value.ObservableValue
+import javafx.collections.FXCollections
 import javafx.collections.ObservableMap
 import net.corda.client.jfx.utils.*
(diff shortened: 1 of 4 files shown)


client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
```diff
@@ -29,21 +35,49 @@ data class PartiallyResolvedTransaction(
...
                             InputResolution.Unresolved(stateRef)
                         } else {
-                            InputResolution.Resolved(it.tx.outRef(stateRef.index))
+                            InputResolution.Resolved(it)
+                        }
+                    }
+                },
+                outputs = if (transaction.coreTransaction is WireTransaction) {
+                    transaction.tx.outRefsOfType<ContractState>().map {
+                        EasyBind.map(SimpleObjectProperty(it)) { OutputResolution.Resolved(it) }
```
exFalso: `OutputResolution.Resolved(it).lift()` or `ReadOblyObjectWrapper(OutputResolution.Resolved(it))`

PR #2784, opened 2018-03-09
Adding public key verification to the X509Utilities.createCertificate…

mkit:
This PR adds public key verification on CSR creation preventing from usage of inconsistent public and private keys.


shamsasari:
I guess there's some more changes that need to be migrated.

PR #2792, opened 2018-03-12
Update final API for V3

adagys:


adagys:
Most of these changes are the result of https://github.com/corda/corda/pull/2746 and https://github.com/corda/corda/pull/2676. The only other change is the introduction of `ConstructorForDeserialization`

PR #2800, opened 2018-03-13
move @CordaSerializable into own module, so that no longer requires a…

roastario:
This is a simple PR to move the @CordaSerializable annotation into it's own module for now.

At the moment, if you want to use this annotation in some shared classes, you end up dragging in the entirety of corda-core. 

This allows you to depend on the corda-annotations module, and easily reduce the size of any artifacts.


corda-annotations/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+}
+
+compileKotlin {
+    kotlinOptions.jvmTarget = "1.8"
+}
+compileTestKotlin {
+    kotlinOptions.jvmTarget = "1.8"
+}
```
chrisr3: I think all you _really_ need here is:
```gradle
apply plugin: 'kotlin'
apply plugin: 'net.corda.plugins.api-scanner'

dependencies {
    compile "org.jetbrains.kotlin:kotlin-stdlib-jre8:$kotlin_version"
}
```
because you'll inherit almost everything else from the root `build.gradle`. And we should probably stick to the deprecated `kotlin-stdlib-jre8` artifact for now until we're ready to upgrade all modules simultaneously.

corda-annotations/build.gradle
```diff
@@ -13,18 +13,19 @@ group 'net.corda'
...
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jre8:$kotlin_version"
 }
 
 compileKotlin {
     kotlinOptions.jvmTarget = "1.8"
 }
 compileTestKotlin {
     kotlinOptions.jvmTarget = "1.8"
-}
```
chrisr3: You don't need these `compileKotlin` and `compileTestKotlin` objects either because they're covered by the
```gradle
tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all { .. }
```
section in the root `build.gradle`.

corda-annotations/build.gradle
```diff
@@ -13,18 +13,19 @@ group 'net.corda'
 version '3.0-SNAPSHOT'
```
chrisr3: You _definitely_ don't want this line.

corda-annotations/build.gradle
```diff
@@ -13,18 +13,19 @@ group 'net.corda'
...
 
 apply plugin: 'kotlin'
+apply plugin: 'net.corda.plugins.api-scanner'
 
 repositories {
     mavenCentral()
 }
```
chrisr3: You shouldn't need to declare any repositories because we have already done this in the top-level `build.gradle` file's `allprojects` section.

chrisr3:
The API Stability break appears to be because the API Scanner plugin is no longer recognising the `@CordaSerializable` annotation as `@Inherited`. Still investigating...

egeek:
@chrisr3 I think there might be some misunderstanding here. The problem that we discussed with @roastario is that we would like to use some of the same classes in both the CorDapp and the API layer. By doing so, the API layer ends up including all of `corda-core` due to the fact that some of the shared classes are annotated with `@CordaSerializable`.

On further investigation, its actually not just the `@CordaSerializable` that is causing us to bring in all of Corda, its the `corda-rpc` module as well.

These are the dependencies that we end up including in our API, many of those are only included because of this (dependencies are for the corda-rpc module, but it also applies to the dependencies that get included due to `@CordaSerializable`:
```
+--- net.corda:corda-rpc:corda-3.0-RC01
|    +--- net.corda:corda-core:corda-3.0-RC01
|    |    +--- org.jetbrains.kotlin:kotlin-reflect:1.1.60 -> 1.2.21
|    |    |    \--- org.jetbrains.kotlin:kotlin-stdlib:1.2.21
|    |    |         \--- org.jetbrains:annotations:13.0
|    |    +--- com.google.code.findbugs:jsr305:3.0.2
|    |    +--- org.apache.logging.log4j:log4j-core:2.9.1 -> 2.7
|    |    |    \--- org.apache.logging.log4j:log4j-api:2.7
|    |    +--- org.slf4j:jcl-over-slf4j:1.7.25 (*)
|    |    +--- org.slf4j:slf4j-api:1.7.25
|    |    +--- io.reactivex:rxjava:1.2.4 -> 1.3.4
|    |    +--- org.apache.commons:commons-jexl3:3.0
|    |    |    \--- commons-logging:commons-logging:1.2
|    |    +--- com.fasterxml.jackson.core:jackson-databind:2.9.2 -> 2.8.10 (*)
|    |    +--- net.i2p.crypto:eddsa:0.2.0
|    |    +--- org.bouncycastle:bcprov-jdk15on:1.57 -> 1.59
|    |    +--- org.bouncycastle:bcpkix-jdk15on:1.57 (*)
|    |    +--- org.hibernate.javax.persistence:hibernate-jpa-2.1-api:1.0.0.Final
|    |    +--- org.hibernate:hibernate-core:5.2.6.Final -> 5.0.12.Final
|    |    |    +--- org.jboss.logging:jboss-logging:3.3.0.Final -> 3.3.1.Final
|    |    |    +--- org.hibernate.javax.persistence:hibernate-jpa-2.1-api:1.0.0.Final
|    |    |    +--- org.javassist:javassist:3.18.1-GA -> 3.21.0-GA
|    |    |    +--- antlr:antlr:2.7.7
|    |    |    +--- org.apache.geronimo.specs:geronimo-jta_1.1_spec:1.1.1
|    |    |    +--- org.jboss:jandex:2.0.0.Final
|    |    |    +--- dom4j:dom4j:1.6.1
|    |    |    \--- org.hibernate.common:hibernate-commons-annotations:5.0.1.Final
|    |    |         \--- org.jboss.logging:jboss-logging:3.3.0.Final -> 3.3.1.Final
|    |    \--- co.paralleluniverse:quasar-core:0.7.9
|    +--- net.corda:corda-node-api:corda-3.0-RC01
|    |    +--- net.corda:corda-core:corda-3.0-RC01 (*)
|    |    +--- org.jetbrains.kotlin:kotlin-reflect:1.1.60 -> 1.2.21 (*)
|    |    +--- org.apache.commons:commons-collections4:4.1
|    |    +--- commons-beanutils:commons-beanutils:1.9.3
|    |    |    +--- commons-logging:commons-logging:1.2
|    |    |    \--- commons-collections:commons-collections:3.2.2
|    |    +--- org.apache.activemq:artemis-core-client:2.2.0
|    |    |    +--- org.jgroups:jgroups:3.6.13.Final
|    |    |    +--- org.apache.activemq:artemis-commons:2.2.0
|    |    |    |    +--- org.jboss.logging:jboss-logging:3.3.0.Final -> 3.3.1.Final
|    |    |    |    +--- io.netty:netty-all:4.1.9.Final
|    |    |    |    +--- commons-beanutils:commons-beanutils:1.9.2 -> 1.9.3 (*)
|    |    |    |    \--- com.google.guava:guava:19.0 -> 21.0
|    |    |    +--- io.netty:netty-all:4.1.9.Final
|    |    |    +--- org.apache.geronimo.specs:geronimo-json_1.0_spec:1.0-alpha-1
|    |    |    \--- org.apache.johnzon:johnzon-core:0.9.5
|    |    +--- org.apache.activemq:artemis-commons:2.2.0 (*)
|    |    +--- commons-fileupload:commons-fileupload:1.3.3
|    |    |    \--- commons-io:commons-io:2.2 -> 2.4
|    |    +--- net.corda.plugins:cordform-common:3.0.8
|    |    |    +--- com.google.code.findbugs:jsr305:3.0.2
|    |    |    \--- com.typesafe:config:1.3.1
|    |    +--- com.typesafe:config:1.3.1
|    |    +--- com.esotericsoftware:kryo:4.0.0
|    |    |    +--- com.esotericsoftware:reflectasm:1.11.3
|    |    |    |    \--- org.ow2.asm:asm:5.0.4
|    |    |    +--- com.esotericsoftware:minlog:1.3.0
|    |    |    \--- org.objenesis:objenesis:2.2
|    |    +--- de.javakaffee:kryo-serializers:0.41
|    |    |    \--- com.esotericsoftware:kryo:4.0.0 (*)
|    |    +--- org.apache.qpid:proton-j:0.21.0
|    |    +--- io.github.lukehutch:fast-classpath-scanner:2.0.21
|    |    \--- co.paralleluniverse:quasar-core:0.7.9
|    \--- co.paralleluniverse:quasar-core:0.7.9
```

What we would like to see is a way to use `@CordaSerializable` (and any similar Corda annotations), and further, to be able to leverage the Corda RPC, without including all these dependencies.

For example, JBoss logging is included, conflicting with our ability to define which logger to use in the API (which is powered by Spring Boot). Similarly Jackson conflicts with with what we are using in the API layer. Then there are other dependencies that Corda is bringing in, that are just increasing the size of our built API artefacts as we aren't using them (e.g. Hibernate).

We could update our `build.gradle` file to exclude all these transitive dependencies (e.g. with
```
compile ("net.corda:corda-rpc:$corda_release_version") {
        transitive = false
    }
```
)
but that puts us in the position of trying to manage and understand which dependencies Corda needs, and then including those directly in our build script. This is undesirable, and has the potential for us to miss dependencies, resulting in class path errors at runtime. It would be better if the points that we integrate Corda provided classes in areas outside of a CorDapp, contain only the minimum set of dependencies to make that function work.

Furthermore, I'm not sure how @roastario 's change would be incompatible with Java 9? Is it just that the package containing the annotation was missing from the exports?

chrisr3:
@egeek Java 9 insists that a package is provided by a _single_ module, i.e. jar. Moving `@CordaSerializable` into its own jar would mean that the `net.corda.core.serialization` package would be split across two modules: `corda-annotations` and `core`. This is the main objection here,

opticyclic:
Can't you just create another package for the annotations then?

roastario:
That would break all cordapps that used the old annotation. 

I'll think about this and push an updated approach.

chrisr3:
We have committed to API Stability, so `@net.core.corda.serialization.CordaSerializable` cannot simply disappear. However, didn't someone already suggest deprecating it in favour of a new `@CordaSerializable` annotation in a different package?

PR #2805, opened 2018-03-13
CORDA-1217 Upgrade guava to fix bug in caches

rick-r3:
Does not manifest single threaded, but prevent drift of versions across repos.


rick-r3:
Causes complicated unit test issue.  Closing and going down the path of porting caches to Caffeine since we had discussed doing that in the near term anyway due to at least one other issue with Guava caches.

PR #2818, opened 2018-03-14
CORDA-1217 Replace Guava caches with Caffeine

rick-r3:
Replace the buggy Guava caches with Caffeine.  https://github.com/ben-manes/caffeine

### DIFF
--- a/client/rpc/build.gradle
+++ b/client/rpc/build.gradle
@@ -69,7 +69,7 @@ dependencies {
     compile project(':node-api')
 
     // For caches rather than guava
-    compile "com.github.ben-manes.caffeine:caffeine:2.6.2"
+    compile "com.github.ben-manes.caffeine:caffeine:$caffeine_version"
 
     // Unit testing helpers.
     testCompile "org.jetbrains.kotlin:kotlin-test:$kotlin_version"
--- a/client/rpc/build.gradle
+++ b/client/rpc/build.gradle
@@ -68,6 +68,9 @@ dependencies {
     compile project(':core')
     compile project(':node-api')
 
+    // For caches rather than guava
+    compile "com.github.ben-manes.caffeine:caffeine:2.6.2"
+
     // Unit testing helpers.
     testCompile "org.jetbrains.kotlin:kotlin-test:$kotlin_version"
     testCompile "junit:junit:$junit_version"
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClient.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClient.kt
@@ -44,12 +44,6 @@ data class RPCClientConfiguration(
         val reapInterval: Duration,
         /** The number of threads to use for observations (for executing [Observable.onNext]) */
         val observationExecutorPoolSize: Int,
-        /**
-         * Determines the concurrency level of the Observable Cache. This is exposed because it implicitly determines
-         * the limit on the number of leaked observables reaped because of garbage collection per reaping.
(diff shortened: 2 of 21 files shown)


client/rpc/build.gradle
```diff
@@ -68,6 +68,9 @@ dependencies {
     compile project(':core')
     compile project(':node-api')
 
+    // For caches rather than guava
+    compile "com.github.ben-manes.caffeine:caffeine:2.6.2"
```
rick-r3: Darn.  Think I should parameterise this.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClient.kt
```diff
@@ -44,12 +44,6 @@ data class RPCClientConfiguration(
         val reapInterval: Duration,
         /** The number of threads to use for observations (for executing [Observable.onNext]) */
         val observationExecutorPoolSize: Int,
-        /**
```
rick-r3: Concurrency level is not configurable anymore.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -154,10 +155,9 @@ class RPCClientProxyHandler(
             }
             observablesToReap.locked { observables.add(observableId) }
         }
-        return CacheBuilder.newBuilder().
+        return Caffeine.newBuilder().
                 weakValues().
-                removalListener(onObservableRemove).
-                concurrencyLevel(rpcConfiguration.cacheConcurrencyLevel).
+                removalListener(onObservableRemove).executor(SameThreadExecutor.getExecutor()).
```
exFalso: I don't think we need the samethreadexecutor trick for RPC

node/src/main/kotlin/net/corda/node/services/messaging/RPCServer.kt
```diff
@@ -145,11 +146,11 @@ class RPCServer(
...
     private fun createObservableSubscriptionMap(): ObservableSubscriptionMap {
-        val onObservableRemove = RemovalListener<InvocationId, ObservableSubscription> {
-            log.debug { "Unsubscribing from Observable with id ${it.key} because of ${it.cause}" }
-            it.value.subscription.unsubscribe()
+        val onObservableRemove = RemovalListener<InvocationId, ObservableSubscription> { key, value, cause ->
+            log.debug { "Unsubscribing from Observable with id ${key} because of ${cause}" }
+            value!!.subscription.unsubscribe()
         }
-        return CacheBuilder.newBuilder().removalListener(onObservableRemove).build()
+        return Caffeine.newBuilder().removalListener(onObservableRemove).executor(SameThreadExecutor.getExecutor()).build()
```
exFalso: Same, no need for the executor

rick-r3: I think we should not introduce any additional threads (it used ForkJoinPool) though which the default would do (and this maintains prior behaviour)

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
rick-r3: I think we should not introduce any additional threads (it used ForkJoinPool) though which the default would do (and this maintains prior behaviour)

andr3ej: agreed, modifying the prior thread behaviour invites bugs

PR #2821, opened 2018-03-14
Fix `IRSSimulationTest` test

vkolomeyko:
Since prod class `IRSSimulation` loading `trade.json` as resource, `trade.json` should be located in prod resources folder of a dependant project to be successfully loaded.

Also improve error reporting.

Current failure can be seen here: https://ci-master.corda.r3cev.com/viewLog.html?buildId=70802&tab=buildResultsDiv&buildTypeId=Corda_Build_BuildZuluOpenJDK


vkolomeyko:
Thanks, @m4ksio - appreciate what you are saying.

But `IRSSimulation` is a **production** class which loads `trade.json` which is currently in the `test` area. IMHO this is wrong.

Also, having checked usages of `trade.json` it is currently only used from `IRSSimulation`. Would this work if I relocate `trade.json` under production resources folder of `network-visualiser` project?
(and return `example-irs-trade.json` where it previously was)

m4ksio:
Yes, the steps you described are reasonable!

vkolomeyko:
All done.
@m4ksio if you could re-review and approve?

PR #2822, opened 2018-03-14
Expose two static fields on SecureHash for Java.

chrisr3:
The `SecureHash.zeroHash` and `SecureHash.allOnesHash` properties must currently be accessed from Java as:
```java
SecureHash.Companion.getZeroHash()
```
and
```java
SecureHash.Companion.getAllOnesHash()
```
Make these more "Java friendly" by using `@JvmField`, and preserve API compatibility by adding explicit getter functions.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
@@ -82,13 +82,31 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
 
         /**
          * A SHA-256 hash value consisting of 32 0x00 bytes.
+         * This field provides more intuitive access from Java.
          */
-        val zeroHash = SecureHash.SHA256(ByteArray(32, { 0.toByte() }))
+        @JvmField
+        val zeroHash: SHA256 = SecureHash.SHA256(ByteArray(32, { 0.toByte() }))
+
+        /**
+         * A SHA-256 hash value consisting of 32 0x00 bytes.
+         * This function is provided for API stability.
+         */
+        @Suppress("Unused")
+        fun getZeroHash(): SHA256 = zeroHash
 
         /**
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -82,13 +82,31 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
 
         /**
          * A SHA-256 hash value consisting of 32 0x00 bytes.
+         * This field provides more intuitive access from Java.
          */
-        val zeroHash = SecureHash.SHA256(ByteArray(32, { 0.toByte() }))
+        @JvmField
+        val zeroHash: SHA256 = SecureHash.SHA256(ByteArray(32, { 0.toByte() }))
```
shamsasari: As a static constant it should be `ZERO_HASH`

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
```diff
@@ -82,13 +82,31 @@ sealed class SecureHash(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+         */
+        @Suppress("Unused")
+        fun getZeroHash(): SHA256 = zeroHash
+
+        /**
+         * A SHA-256 hash value consisting of 32 0xFF bytes.
+         * This field provides more intuitive access from Java.
+         */
+        @JvmField
+        val allOnesHash: SHA256 = SecureHash.SHA256(ByteArray(32, { 255.toByte() }))
```
shamsasari: likewise

core/src/main/kotlin/net/corda/core/crypto/SecureHash.kt
chrisr3: I think that would defeat the purpose here. From Kotlin, it is definitely `SecureHash.zeroHash` rather than `SecureHash.ZERO_HASH`.

PR #2823, opened 2018-03-14
CORDA-1208: Notary service should persist the notarisation request si…

adagys:
…gnature along with the committed input states.

This required modifying the uniqueness provider interface to accept the signature in addition to input states.
Until now the committed state log used to be stored as a map of (state reference -> (tranasction id, consuming party)).
Adding the serialized signature would mean inflating each state entry by around 700 bytes, which would be grossly inefficient.
Instead, two tables are now used: one for storing (state referece -> transaction id) map, and another for storing the notarisation
request details (transaction id, consuming party, date, signature).


shamsasari:
Sorry I have to pull out of this one. I have to get the network parameters update feature in.

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -1,62 +1,68 @@
...
+            val requestSignature: ByteArray,
 
-    @Entity
-    @javax.persistence.Table(name = "${NODE_DATABASE_PREFIX}notary_commit_log")
-    class PersistentNotaryCommit(id: PersistentStateRef, consumingTxHash: String, consumingIndex: Int, party: PersistentParty) :
-            PersistentUniqueness(id, consumingTxHash, consumingIndex, party)
+            @CreationTimestamp
+            @Temporal(TemporalType.TIMESTAMP)
+            @Column(name = "request_date")
+            var requestDate: Date? = null
```
thschroeter: Why is the `requestDate` nullable?

node/src/main/kotlin/net/corda/node/services/transactions/RaftTransactionCommitLog.kt
```diff
@@ -0,0 +1,258 @@
...
+    private val map = db.transaction { createMap() }
+
+    /** Commits the input states for the transaction as specified in the given [Commands.CommitTransaction]. */
+    fun commitTransaction(raftCommit: Commit<Commands.CommitTransaction>): Map<StateRef, SecureHash> {
+        raftCommit.use {
+            val index = it.index()
+            val conflicts = LinkedHashMap<StateRef, SecureHash>()
+            db.transaction {
+                val commitCommand = raftCommit.command()
+                logRequest(commitCommand)
```
thschroeter: We're logging all request before checking for conflicts. How do we know if the request was a double spend attempt when we just look at the request log? Do we query the states, if the `txId` is not present, we know the request was a double spend attempt?

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
adagys: I need to double check, but it's a var and will be populated automatically when persisting

node/src/main/kotlin/net/corda/node/services/transactions/RaftTransactionCommitLog.kt
adagys: With this approach we don't, but we could find out by querying the states as you say. However, my thinking was that the important thing is to be able to find out who notarised a transaction when someone presents a transction id hash on a failed notarisation attempt. Knowing whether a particular request was a double spend doesn't provide much by itself - you don't know which of the requests was valid and which one malicious (or an accident). Maybe I missed something?

thschroeter: I don't think so. Let's add this later if we need to. Thanks for you explanation.

PR #2825, opened 2018-03-14
[CORDA-2615]: Added description for configFile option in deployNodes Cordform Gradle task.

szymonsztuka:
Port of https://github.com/corda/corda/pull/2615


docs/source/generating-a-node.rst
```diff
@@ -137,6 +137,24 @@ a single node to run the network map service, by putting its name in the ``netwo
 
 .. warning:: When adding nodes, make sure that there are no port clashes!
 
+To extend node configuration beyond the properties defined in the ``deployNodes`` task use the ``configFile`` property with the path (relative or absolute) set to an additional configuration file.
+This file should follow the standard :doc:`corda-configuration-file` format, as per node.conf. The properties from this file will be appended to the generated node configuration.
```
ags313: What is the precedence?

szymonsztuka: It will be appended only, so if you add a property already created by 'deployNodes' task, it will be duplicated.

PR #2827, opened 2018-03-15
Move notary service related classes and interfaces in core to interna…

adagys:
…l, since we won't be able to stabilise the APIs for writing custom notary services any time soon (the docs already mention it).

I left out the wire protocol related classes so we don't accidentally break it.

This PR only contains moves/removals, no change in logic.


core/src/main/kotlin/net/corda/core/flows/NotaryWireFormat.kt
```diff
@@ -36,34 +34,6 @@ class NotarisationRequest(statesToConsume: List<StateRef>, val transactionId: Se
 
     /** States this request specifies to be consumed. Sorted to ensure the serialized form does not get affected by the state order. */
     val statesToConsume: List<StateRef> get() = _statesToConsumeSorted // Getter required for AMQP serialization
-
-    /** Verifies the signature against this notarisation request. Checks that the signature is issued by the right party. */
-    fun verifySignature(requestSignature: NotarisationRequestSignature, intendedSigner: Party) {
```
adagys: moved this out since it throws `NotaryInternalException`, which shouldn't be exposed outside the service flow

rick-r3:
What is the alternative to this?  I presume we could create a parallel copy of the API in `internal` and move all *our* usage to the internal one?

I'm not commenting on the merit of that, just want to know our options.

adagys:
Yeah we could do that. Or we can leave it as is and keep updating the API when we change things. But I just don't want to have the "but this is a breaking change!!! oh it's just notaries" conversation every time.

rick-r3:
Double-checking @mikehearn is okay with this?

mikehearn:
Yes, it's OK by me but it needs an explanation and a clear note in the changelog/release notes.

docs/source/changelog.rst
```diff
@@ -7,6 +7,9 @@ release, see :doc:`upgrade-notes`.
 Unreleased
 ==========
 
+* Due to ongoing work the experimental interfaces for defining custom notary services have been moved to the internal package.
+  It can still be used in CorDapps, but further changes may be required in the future.
```
rick-r3: This is just talking about the back end service?  What do we mean by `It can still be used in CorDapps, but further changes may be required in the future.`  Do we mean "If you are using these classes to implement your own notary service..." ?

adagys: Yes, I'll make it more clear

adagys:
It's moves and I also removed some deprecated backend-related methods/classes

PR #2838, opened 2018-03-16
Direct bootstrapper link instead of manual build.

joeldudleyr3:


docs/source/setting-up-a-corda-network.rst
```diff
@@ -61,11 +61,7 @@ be done with the network bootstrapper. This is a tool that scans all the node co
 generate the network parameters file which is copied to the nodes' directories. It also copies each node's node-info file
 to every other node so that they can all transact with each other.
 
-The bootstrapper tool can be built with the command:
-
-``gradlew buildBootstrapperJar``
-
-The resulting jar can be found in ``tools/bootstrapper/build/libs/``.
+The bootstrapper tool can be downloaded from http://downloads.corda.net/NetworkBootstrapper-corda-X.Y.jar, where ``X`` is the major Corda version and ``Y`` is the minor Corda version.
```
shamsasari: The file name is camel case? Do our other published binaries have the same format?

If this is the case then all references to the jar in the docs will need to change as it's currently `network-bootstrapper.jar`.

shamsasari: Personally I don't like this and would prefer that the file name be `network-boostrapper-...`

PR #2840, opened 2018-03-16
NodeVaultService::bootstrapContractStateType should not use sys classloader [CORA-1229]

IgorNitto:


IgorNitto:
Fixed the problem by reverting the classloader changes in another PR. Close this for now.

PR #2841, opened 2018-03-16
Added sign helper method to CertificateAndKeyPair for producing Signe…

shamsasari:
…dDataWithCert objects

Porting over changes from https://github.com/corda/enterprise/pull/567.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
@@ -378,10 +378,16 @@ val CordaX500Name.x500Name: X500Name
 val CordaX500Name.Companion.unspecifiedCountry
     get() = "ZZ"
 
-fun <T : Any> T.signWithCert(privateKey: PrivateKey, certificate: X509Certificate): SignedDataWithCert<T> {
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -378,10 +378,16 @@ val CordaX500Name.x500Name: X500Name
...
-fun <T : Any> T.signWithCert(privateKey: PrivateKey, certificate: X509Certificate): SignedDataWithCert<T> {
+inline fun <T : Any> T.signWithCert(signer: (SerializedBytes<T>) -> DigitalSignatureWithCert): SignedDataWithCert<T> {
     val serialised = serialize()
-    val signature = Crypto.doSign(privateKey, serialised.bytes)
-    return SignedDataWithCert(serialised, DigitalSignatureWithCert(certificate, signature))
+    return SignedDataWithCert(serialised, signer(serialised))
+}
+
+fun <T : Any> T.signWithCert(privateKey: PrivateKey, certificate: X509Certificate): SignedDataWithCert<T> {
+    return signWithCert {
```
kchalkias: Theoretically this could be one-liner without requiring the `signature` val, but if you feel it's more readable leave it as it is.

PR #2843, opened 2018-03-19
CordApps back in node and webserver classpath [CORDA-1232]

IgorNitto:
As mentioned in: https://r3-cev.atlassian.net/browse/CORDA-1232,  although classloader changes looked fine they break in some circumstances and the fix is to upgrade RPC serialization to AMQP. To unlock the next release i am temporarily re-enabling loading of cordApps via Capsule.


josecoll:
Please merge without my approval as I am not familiar with the background of this change.

PR #2844, opened 2018-03-19
Adding the X509CRL custom serializer.

mkit:
Custom serializer for the X509CRL. Required by the certificate revocation list support.
JIRA: [CORDA-1233](https://r3-cev.atlassian.net/browse/CORDA-1233)


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/custom/X509CRLSerializer.kt
```diff
@@ -0,0 +1,37 @@
+/*
```
fenryka: Does the open source code need this header? it's hardly confidential if we're putting it on GitHub under an open source licence

mkit: I will remove it. Thanks.

mkit:
Ticket created.

PR #2849, opened 2018-03-20
Fix RPC flaky test

bpaunescu:


client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
```diff
@@ -247,9 +248,18 @@ class RPCStabilityTests {
             assertEquals("pong", client.ping())
             serverFollower.shutdown()
             startRpcServer<ReconnectOps>(ops = ops, customPort = serverPort).getOrThrow()
-            Thread.sleep(1000) //wait for the server to come back up
-            val pingFuture = pool.fork(client::ping)
-            assertEquals("pong", pingFuture.getOrThrow(10.seconds))
+            var response: String? = null
```
vkolomeyko: May I suggest using `net.corda.nodeapi.EventuallyKt#eventually` ?

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -415,7 +415,6 @@ class RPCClientProxyHandler(
                     sendingEnabled = false
```
vkolomeyko: I think the plan also was to try to replace `sendingEnabled` and `lock` with single `val` of type `AtomicBoolean`?

bpaunescu: That can be done, but wouldn't have solved the flakyness of the test. While the client reconnects, any RPC call will throw. The previous version attempted to work around this by putting in a sleep() which doesn't always work well on windows. This version of the test simply calls an RPC and catches until successful or it times out.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -173,9 +172,8 @@ class RPCClientProxyHandler(
     private val deduplicationChecker = DeduplicationChecker(rpcConfiguration.deduplicationCacheExpiry)
     private val deduplicationSequenceNumber = AtomicLong(0)
 
-    private val lock = ReentrantReadWriteLock()
     @Volatile
-    private var sendingEnabled = true
+    private var sendingEnabled: AtomicBoolean = AtomicBoolean(true)
```
vkolomeyko: I think there is no need for `@Volatile` anymore and it can be `val` instead of `var`.

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
bpaunescu: It's not straightforward to use that method as it's in a different module. Doesn't seem worth adding it as a dependency just for one method. Moving the method to some common module (test-utils or test-commons maybe?) may be possible but it's not a priority now as this PR should get into the DP3 RC asap

vkolomeyko: Kind of agree - it is expanding the scope of this PR.
But as discussed offline - there is a solution to export this test dependency from `node-api` test artefacts.
I leave it up to you if you want to include it or not.

client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
```diff
@@ -32,6 +33,7 @@ import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicLong
+import kotlin.concurrent.thread
```
vkolomeyko: Minor: I do not think this import is necessary.

PR #2850, opened 2018-03-20
[CORDA-1235]: Prevent out of order consumption of initiation vs subsequent messages

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1235

### DIFF
--- a/core/src/test/kotlin/net/corda/core/flows/FlowTestsUtils.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/FlowTestsUtils.kt
@@ -43,20 +43,6 @@ inline fun <I : FlowLogic<*>, reified R : FlowLogic<*>> StartedNode<InternalMock
 }
 
 /**
- * Allows to register a flow of type [R] against an initiating flow of type [I].
- */
-inline fun <I : FlowLogic<*>, reified R> StartedNode<InternalMockNetwork.MockNode>.registerInitiatedFlowAsClosure(initiatingFlowType: KClass<I>, noinline action: FlowSession.() -> R) {
-
-    internalRegisterFlowFactory(initiatingFlowType.java, InitiatedFlowFactory.Core { session -> InitiatedFlowClosure(session, action) }, InitiatedFlowClosure::class.javaObjectType, true)
-}
-
-class InitiatedFlowClosure<out R>(private val session: FlowSession, private val action: FlowSession.() -> R) : FlowLogic<R>() {
-
-    @Suspendable
-    override fun call(): R = action(session)
-}
-
-/**
  * Allows to register a flow of type [Answer] against an initiating flow of type [I], returning a valure of type [R].
  */
 inline fun <I : FlowLogic<*>, reified R : Any> StartedNode<InternalMockNetwork.MockNode>.registerAnswer(initiatingFlowType: KClass<I>, value: R) {
--- a/core/src/test/kotlin/net/corda/core/flows/FlowTestsUtils.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/FlowTestsUtils.kt
@@ -43,6 +43,20 @@ inline fun <I : FlowLogic<*>, reified R : FlowLogic<*>> StartedNode<InternalMock
 }
 
 /**
+ * Allows to register a flow of type [R] against an initiating flow of type [I].
+ */
+inline fun <I : FlowLogic<*>, reified R> StartedNode<InternalMockNetwork.MockNode>.registerInitiatedFlowAsClosure(initiatingFlowType: KClass<I>, noinline action: FlowSession.() -> R) {
+
+    internalRegisterFlowFactory(initiatingFlowType.java, InitiatedFlowFactory.Core { session -> InitiatedFlowClosure(session, action) }, InitiatedFlowClosure::class.javaObjectType, true)
+}
+
+class InitiatedFlowClosure<out R>(private val session: FlowSession, private val action: FlowSession.() -> R) : FlowLogic<R>() {
+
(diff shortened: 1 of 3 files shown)


node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -660,30 +662,30 @@ private class P2PMessagingConsumer(
...
+        private const val initialSessionMessages = "${P2PMessagingHeaders.Type.KEY}<>'${P2PMessagingHeaders.Type.SESSION_INIT_VALUE}'"
     }
 
     private var startedFlag = false
 
     val messages: PublishSubject<ClientMessage> = PublishSubject.create<ClientMessage>()
 
-    private var initialConsumer = multiplex(queueNames, createSession, initialSessionMessages)
-    private var existingConsumer = multiplex(queueNames, createSession, existingSessionMessages)
+    private var initialOnlyConsumer = multiplex(queueNames, createSession, initialSessionMessages)
```
exFalso: Why are these vars?

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -660,30 +662,30 @@ private class P2PMessagingConsumer(
         private val drainingModeWasChangedEvents: Observable<Pair<Boolean, Boolean>>) : LifecycleSupport {
 
     private companion object {
-        private const val initialSessionMessages = "${P2PMessagingHeaders.Type.KEY}='${P2PMessagingHeaders.Type.SESSION_INIT_VALUE}'"
-        private const val existingSessionMessages = "${P2PMessagingHeaders.Type.KEY}<>'${P2PMessagingHeaders.Type.SESSION_INIT_VALUE}'"
+        private const val initialSessionMessages = "${P2PMessagingHeaders.Type.KEY}<>'${P2PMessagingHeaders.Type.SESSION_INIT_VALUE}'"
```
exFalso: Is this the correct condition? previously `initialSessionMessages` checked for equality

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -660,30 +662,30 @@ private class P2PMessagingConsumer(
...
 
         synchronized(this) {
             require(!startedFlag)
-            drainingModeWasChangedEvents.filter { change -> change.switchedOn() }.doOnNext { pauseInitial() }.subscribe()
-            drainingModeWasChangedEvents.filter { change -> change.switchedOff() }.doOnNext { resumeInitial() }.subscribe()
-            subscriptions += initialConsumer.messages.doOnNext(messages::onNext).subscribe()
-            subscriptions += existingConsumer.messages.doOnNext(messages::onNext).subscribe()
-            if (!isDrainingModeOn()) {
-                initialConsumer.start()
+            drainingModeWasChangedEvents.filter { change -> change.switchedOn() }.doOnNext { initialAndExistingConsumer.switchTo(initialOnlyConsumer) }.subscribe()
```
exFalso: The `switchedOn`/`switchedOff` thing can be changed to `distinctUntilChanged`

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
sollecitom: To see if you pay attention :D (because of a series of previous changes and we forgot to fix it, changing them to vals)

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
sollecitom: The condition is correct, the name isn't :) renamed to "existingOnlyConsumer" (while in draining mode initial session messages need not to be consumed.

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
sollecitom: No, while it could be done by concatenating the initial state with the observable and then call `distinctUntilChanged`, this way the distance from an event propagation mechanism is smaller. Also the code is more readable.

core/src/test/kotlin/net/corda/core/flows/FlowTestsUtils.kt
```diff
@@ -42,6 +42,20 @@ inline fun <I : FlowLogic<*>, reified R : FlowLogic<*>> StartedNode<InternalMock
     internalRegisterFlowFactory(initiatingFlowType.java, InitiatedFlowFactory.Core { session -> construct(session) }, R::class.javaObjectType, true)
 }
 
+/**
+ * Allows to register a flow of type [R] against an initiating flow of type [I].
+ */
+inline fun <I : FlowLogic<*>, reified R> StartedNode<InternalMockNetwork.MockNode>.registerInitiatedFlowAsClosure(initiatingFlowType: KClass<I>, noinline action: FlowSession.() -> R) {
```
exFalso: Yeah so here `action` needs to be `@Suspendable`. It's a real pain
One way I got around this in a side project is by defining a *Java* interface with the right annotation. Then you can use closure syntax because Kotlin can do SAM conversion, but *only* when interfacing with Java.. what a pain

exFalso: And I think even then Kotlin messed up sometimes (iirc when there were captures involved) and generated another method anyway.

PR #2856, opened 2018-03-21
Documents TestIdentity in the testing API docs.

joeldudleyr3:


docs/source/api-testing.rst
```diff
@@ -304,6 +304,27 @@ Contract testing
 The Corda test framework includes the ability to create a test ledger by calling the ``ledger`` function
 on an implementation of the ``ServiceHub`` interface.
 
+Test identities
+^^^^^^^^^^^^^^^
+
+You can create dummy identities to use in test transactions using the ``TestIdentity`` class. Each ``TestIdentity``
+wraps the ``CordaX500Name`` and ``KeyPair`` of a dummy identity. ``TestIdentity`` also provides helper
+methods for retrieving the identity's ``Party`` and ``PartyAndCertificate`` representations.
+
```
anthonykr3: Is it worth mentioning the `fresh` method, as that seems rather useful? Also that the `PartyAndCertificate` is retrieved via the `identity` property, and the `publicKey` property?

Issue #2454, opened 2018-02-02
Incorrect use of PersistentMap in NodeSchedulerService


ing-rafbhoedjang:
Version: Corda 2.0.0
`NodeSchedulerService.kt`

Line 271 in file `NodeSchedulerService.kt` contains the following statement in function `getScheduledFlow`:

`                    scheduledStates[scheduledState.ref] = newState`

This results in a call to `PersistentMap.set`, which means that any previous value stored in the map is *not* replaced. The new value is effectively lost. This later results in problems; see the comparison at line 262, which uses the value previously stored in the map:

    val previousState = scheduledStates[scheduledState.ref]
    if (previousState != null && previousState === scheduledState) {

The lost update to `scheduledStates` results in an old value being returned from the map. Consequently, the comparison in the if statement fails. In our use case, this then results in a scheduling loop where an old `ScheduledStateRef` keeps on being rescheduled but is never executed.

Proposed solution (**not tested**): use `scheduledStates.put(scheduledState.ref, newState)`

mikehearn:
Thanks for the report. The PersistentMap.set problem has cropped up recently in another context, that just doesn't look like a safe API choice at all. Luckily it's internal so we can tweak it easily.

mikehearn:
https://r3-cev.atlassian.net/browse/CORDA-1002
https://r3-cev.atlassian.net/browse/CORDA-1001

PR #2860, opened 2018-03-21
CORDA-1002 Add node scheduler persistence tests

anthonykr3:
Adds some extra tests to the NodeSchedulerService that test the database persistence. Related to this github issue: https://github.com/corda/corda/issues/2454.

There was no need to fix the NodeSchedulerService, as the underlying issue has been fixed by this commit: https://github.com/corda/corda/commit/9d66214f4a97710109df3baecb1244013c7d41ea however it is not immediately clear why this commit fixed it.

The `scheduledStates` were originally declared as a `PersistentMap<StateRef, ScheduledStateRef>` on the `NodeSchedulerService`. The above mentioned commit changes it to be initialised by the constructor and explicitly states the parameter type is the generic base class of type `MutableMap<StateRef, ScheduledStateRef>`.

If you assign a value to the `PersistentMap` via index (i.e. `scheduledStates[x] = y`), the `set` operator on `PersistentMap` is called which does nothing. This was the behaviour observed in this bug.

If you assign a value to a `PersistentMap` cast to a `MutableMap<StateRef, ScheduledStateRef>` it calls the `set` operator defined as an extension function in the Kotlin standard library (see [here](https://github.com/JetBrains/kotlin/blob/f957b3072b6f678f20a79dad3921aee380586f9f/libraries/stdlib/src/kotlin/collections/Maps.kt#L175)) which just calls the `put` method.

Casting the `PersistentMap` to the generic base class made the `put` method be called instead of the `set` method, which fixed the issue.

It is clear that Kotlin expects `set` and `put` to act the same way on collections, so I will alter the `PersistentMap` as part of [CORDA-1001](https://r3-cev.atlassian.net/browse/CORDA-1001) that they both work the same way.

Am happy to add more tests if there is any thing that could benefit - I think the unit tests cover most cases apart from the fact they use a map that works the way you expect a map to work which is why this issue wasn't picked up by them.

### DIFF
--- a/node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
+++ b/node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
@@ -3,6 +3,7 @@ package net.corda.node.services.events
 import com.google.common.util.concurrent.MoreExecutors
 import com.nhaarman.mockito_kotlin.*
 import net.corda.core.contracts.*
+import net.corda.core.crypto.SecureHash
 import net.corda.core.flows.FlowLogic
 import net.corda.core.flows.FlowLogicRef
 import net.corda.core.flows.FlowLogicRefFactory
@@ -11,12 +12,15 @@ import net.corda.core.internal.concurrent.openFuture
 import net.corda.core.internal.uncheckedCast
 import net.corda.core.node.ServicesForResolution
 import net.corda.core.utilities.days
+import net.corda.node.internal.configureDatabase
 import net.corda.node.services.api.FlowStarter
 import net.corda.node.services.api.NodePropertiesStore
 import net.corda.nodeapi.internal.persistence.CordaPersistence
+import net.corda.nodeapi.internal.persistence.DatabaseConfig
 import net.corda.nodeapi.internal.persistence.DatabaseTransaction
 import net.corda.testing.internal.doLookup
 import net.corda.testing.internal.rigorousMock
+import net.corda.testing.node.MockServices
 import net.corda.testing.node.TestClock
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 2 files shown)


node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -183,4 +198,92 @@ class NodeSchedulerServiceTest {
...
+        val database = configureDatabase(dataSourceProps, databaseConfig, rigorousMock())
+        val scheduler = database.transaction {
+            createScheduler(database)
+        }
+
+        val ssr1 = ScheduledStateRef(stateRef, timeInTheFuture)
+        database.transaction {
+            scheduler.scheduleStateActivity(ssr1)
+        }
+        database.dataSource.connection.commit()
```
anthonykr3: For some reason, without this commit the data isn't written - even though there is a commit in the database transaction logic.

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -183,4 +198,97 @@ class NodeSchedulerServiceTest {
...
+
+        val flowLogic = rigorousMock<FlowLogic<*>>()
+        val logicRef = rigorousMock<FlowLogicRef>()
+
+        transactionStates[stateRef] = transactionStateMock(logicRef, timeInTheFuture)
+        flows[logicRef] = flowLogic
+
+        val newDatabase = configureDatabase(dataSourceProps, DatabaseConfig(), rigorousMock())
+        val newScheduler = newDatabase.transaction {
+            System.out.println(newDatabase.dataSource.connection.metaData)
```
andr3ej: is println intentional?

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -183,4 +198,97 @@ class NodeSchedulerServiceTest {
...
+        scheduler.cancel()
+        database.close()
+
+        val flowLogic = rigorousMock<FlowLogic<*>>()
+        val logicRef = rigorousMock<FlowLogicRef>()
+
+        transactionStates[stateRef] = transactionStateMock(logicRef, timeInTheFuture)
+        flows[logicRef] = flowLogic
+
+        val newDatabase = configureDatabase(dataSourceProps, DatabaseConfig(), rigorousMock())
```
andr3ej: this is never closed, which leaks hikari threads i think

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -183,4 +198,97 @@ class NodeSchedulerServiceTest {
...
+    fun `test that if schedule is updated then the flow is invoked on the correct schedule`() {
+        val dataSourceProps = MockServices.makeTestDataSourceProperties()
+        val timeInTheFuture = mark + 1.days
+        val stateRef = StateRef(SecureHash.allOnesHash, 0)
+
+        val ssr1 = ScheduledStateRef(stateRef, mark)
+        val ssr2 = ScheduledStateRef(stateRef, timeInTheFuture)
+        val logicRef = rigorousMock<FlowLogicRef>()
+        val flowLogic = rigorousMock<FlowLogic<*>>()
+        val database = configureDatabase(dataSourceProps, databaseConfig, rigorousMock())
```
andr3ej: as above, never closed

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
andr3ej: that's weird. i'd add an XXX comment

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
anthonykr3: Nope.

PR #2862, opened 2018-03-21
CORDA-1258 - Only register custom serializers once

fenryka:


docs/source/changelog.rst
```diff
@@ -7,6 +7,10 @@ Unreleased
 Here are brief summaries of what's changed between each snapshot release. This includes guidance on how to upgrade code
 from the previous milestone release.
 
+* Fix CORDA-1258. Custom serializers were spuriously registered every time a serialization factory was fetched from the cache rather than
+  only once when it was created. Whilst registering serializers that already exist is essentially a no-op, it's a performance overhead for
+  a very frequent operation.
```
rick-r3: And it synchronizes so shows up as contended in performance tests!

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationSchemaTests.kt
```diff
@@ -0,0 +1,98 @@
...
+) : SerializerFactory(wl, cl) {
+    var registerCount = 0
+
+    override fun register(customSerializer: CustomSerializer<out Any>) {
+        ++registerCount
+        return super.register(customSerializer)
+    }
+}
+
+// Instance of our test factory counting registration attempts. Sucks its  gloal but for testing purposes this
```
rick-r3: gloal -> global

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationSchemaTests.kt
```diff
@@ -0,0 +1,98 @@
...
+
+// Test factory that lets us count the number of serializer registration attempts
+class TestSerializerFactory(
+        wl : ClassWhitelist,
+        cl : ClassLoader
+) : SerializerFactory(wl, cl) {
+    var registerCount = 0
+
+    override fun register(customSerializer: CustomSerializer<out Any>) {
+        ++registerCount
```
rick-r3: I'm thinking there's a simpler way to detect duplicate calls to that method than adding a new `UseCase`.  Can we not just set a flag in that method, and throw if called more than once.  Then a test that checks it doesn't normally throw and does throw if invoked twice (might have to make `registerCustomSerializers` open and protected, and then override it).

fenryka: Use case isn't strictly needed, but at some point having a context we can do "stuff" with that is outside the standard ones is useful, so introducing it now isn't a terrible thing.

Can't just have a flag as register is called for every serialiser, hence tracking the count later on

fenryka:
@rick-r3 addressed

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationSchemaTests.kt
fenryka: done

docs/source/changelog.rst
fenryka: done

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationSchemaTests.kt
rick-r3: I was thinking you could override `registerCustomSerializers` on the scheme.... and count how many times that is called.  Should be only once :-)

PR #2863, opened 2018-03-22
Integration test for network parameter updates and improved logging

shamsasari:
This is a backport of the relevant changes from https://github.com/corda/enterprise/pull/565 and https://github.com/corda/enterprise/pull/477.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkMap.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkMap.kt
@@ -31,7 +31,14 @@ data class NetworkMap(
         val nodeInfoHashes: List<SecureHash>,
         val networkParameterHash: SecureHash,
         val parametersUpdate: ParametersUpdate?
-)
+) {
+    override fun toString(): String {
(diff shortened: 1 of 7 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkMap.kt
```diff
@@ -31,7 +31,14 @@ data class NetworkMap(
         val nodeInfoHashes: List<SecureHash>,
         val networkParameterHash: SecureHash,
         val parametersUpdate: ParametersUpdate?
-)
+) {
+    override fun toString(): String {
+        return """${NetworkMap::class.java.simpleName}(nodeInfoHashes=
+${nodeInfoHashes.joinToString("\n")}
+networkParameterHash=$networkParameterHash
```
kasiastreich: I just noticed that, why we have "parameter" not parameters.

shamsasari: Seems like this slipped through in V3. Unless our serialisation framework allows renaming, I fear we're stuck with this.

PR #2868, opened 2018-03-23
Updated node configuration documentation to contain quoted parameters

tomtau:
Unquoted "dataSource.*" parameters may cause strange issues where reference.conf parameters override the ones specified in node.conf.

As confirmed by @szymonsztuka on Slack: 
> I suppose it more a feature of the HOCON, if reference.conf has a key name with a dot quoted and you override it by unquoted one, that the quoted take precedence (for HCOON these are different keys) a nd yes the example configuration for Postgress in docs is wrong then

Corresponding SO answer: https://stackoverflow.com/review/suggested-edits/19202559

### DIFF
--- a/docs/source/node-database.rst
+++ b/docs/source/node-database.rst
@@ -40,9 +40,9 @@ Here is an example node configuration for PostgreSQL:
 
     dataSourceProperties = {
         dataSourceClassName = "org.postgresql.ds.PGSimpleDataSource"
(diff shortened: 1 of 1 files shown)


mikehearn:
Thanks for raising the issue and a PR! This alternative PR fixes the issue in a different way, by logging an error if these cases are detected:

https://github.com/corda/corda/pull/2895

Do you think that approach is OK?

tomtau:
Yes, that approach is OK.

PR #2871, opened 2018-03-23
CORDA-1265: Upgrade to Kotlin language/API 1.2

chrisr3:
We are currently compiling against the old Kotlin 1.1 language/API definitions, despite using Kotlin 1.2.20 artifacts. Upgrade to use Kotlin 1.2 completely.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -3214,7 +3214,7 @@ public static final class net.corda.core.transactions.FilteredTransaction$Compan
   @org.jetbrains.annotations.NotNull public final List inputsOfType(Class)
   public String toString()
   public final void verify()
-  public static final net.corda.core.transactions.LedgerTransaction$Companion Companion
+  @java.lang.Deprecated public static final net.corda.core.transactions.LedgerTransaction$Companion Companion
 ##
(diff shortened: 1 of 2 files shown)


chrisr3:
Not merging this straight away because I'm not sure how this will affect compatibility between Corda V3 and R3 Corda V3.

chrisr3:
Upgrading to Kotlin 1.2 means that these uses of the `companion` object from Java are marked as `@Deprecated`:
```java
Object companion = LedgerTransaction.Companion;
int hashCode = companion.hashCode();
System.out.println(SignedTransaction.Companion);
```
Isn't this a Good Thing, seeing as both of these are **private** `companion object`s and so arguably shouldn't have been accessible in the first place? The equivalent Kotlin code refuses even to compile:
```kotlin
val companion = LedgerTransaction.Companion
val hashCode = companion.hashCode
println(SignedTransaction.Companion)
```
So I guess that what really matters is what JetBrains intends to do next: if some future version of Kotlin intends to make a `private companion object` inaccessible from Java as well (and its `class` definition is _already_ `private`) then I think we should accept the `@Deprecated` annotation rather than pollute our public API with elements that were never meant to be there in the first place.

PR #2886, opened 2018-03-27
[CORDA-1001] Modify NodeSchedulerService to use the persisted schedules directly r…

anthonykeenan:
Stop the NodeSchedulerService from using the PersistentMap

It is currently using a DB table, loading the data into a map and then loading the map into a queue. By using the db as the version of truth and only ever getting the top value from the database, the NodeSchedulerService becomes simpler and doesn't need to deal with syncronising three different lists.

This also deals with the fact that currently the cache is unbounded so a cordapp that schedules a large amount flows for the future would effectively cause a memory leak.


node/src/main/kotlin/net/corda/node/services/events/NodeSchedulerService.kt
```diff
@@ -258,7 +248,7 @@ class NodeSchedulerService(private val clock: CordaClock,
             var flowName: String? = "(unknown)"
             try {
                 database.transaction {
-                    val scheduledFlow = getScheduledFlow(scheduledState)
+                    val scheduledFlow = removeScheduledFlowFromQueue(scheduledState)
```
rick-r3: The enterprise implementation of the state machine manager (will be open sourced) needs to delegate the delete from the database to the `startFlow` which runs in a different database transaction.  That makes it much more difficult to implement the "head of queue" in the database only, since there is some transient state involved in knowing the head of the queue.

rick-r3 (changes requested):
I've made some trivial comments, and some non-trivial comments.

In the multi-threaded environment of an enterprise node it gets trickier and more subtle.  I'm thinking the way forward is to perhaps consider the enterprise case first, and do any refactor there to cover all the requirements.  We will be porting the enterprise state machine back to open source.  You could either attempt a port of just the scheduler back to open source prior to that with some bits amended or it comes over wholesale (I think no changes required) when we port the state machine across.

anthonykr3:
Closing this PR - will reopen in enterprise

PR #2887, opened 2018-03-27
ENT-1668: Don't crash on node startup when network map is not available

kasiastreich:
Don't crash on node startup when network map is not available

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
@@ -10,6 +10,7 @@ import net.corda.nodeapi.internal.network.NETWORK_PARAMS_FILE_NAME
 import net.corda.nodeapi.internal.network.NETWORK_PARAMS_UPDATE_FILE_NAME
 import net.corda.nodeapi.internal.network.SignedNetworkParameters
 import net.corda.nodeapi.internal.network.verifiedNetworkMapCert
+import java.net.ConnectException
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption
 import java.security.cert.X509Certificate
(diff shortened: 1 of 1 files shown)


gendal:
Do we continue to try in the background?  Sorry for not studying this in depth but my assumption is that a node will 1) start up even if it can't publish its NodeInfo, but will continue to try in the background and 2) start up even if it can't download the Network Map file (or any of the NodeInfos referenced by it) but will continue to try in the background.  Is that the situation we now have? (In both C3 and R3C?)

kasiastreich:
@gendal 
1. If node doesn't have network parameters file it won't startup. It can't join the network if network map is down/maybe node was misconfigured.
2. If node has network parameters saved locally (either distributed offline or from previous start) it will continue to be fully operational.

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -26,7 +27,13 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
     val networkParameters by lazy { retrieveNetworkParameters() }
 
     private fun retrieveNetworkParameters(): NetworkParameters {
-        val advertisedParametersHash = networkMapClient?.getNetworkMap()?.payload?.networkParameterHash
+        val advertisedParametersHash = try {
+            networkMapClient?.getNetworkMap()?.payload?.networkParameterHash
+        } catch(e: ConnectException) {
```
shamsasari: space after `catch`

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -26,7 +27,13 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
+            logger.info("Couldn't connect to NetworkMap")
```
shamsasari: Include the exception in the log: `"...", e)`

PR #2890, opened 2018-03-28
ENT-1663 - Inclusion of the cert path and root cert in the exception …

shamsasari:
…message when cert path validation fails.

Also, added check to node startup that --initial-registration cannot occur in devMode

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
@@ -104,7 +104,18 @@ object X509Utilities {
     fun validateCertPath(trustedRoot: X509Certificate, certPath: CertPath) {
         val params = PKIXParameters(setOf(TrustAnchor(trustedRoot, null)))
         params.isRevocationEnabled = false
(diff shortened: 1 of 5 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
```diff
@@ -104,7 +104,16 @@ object X509Utilities {
     fun validateCertPath(trustedRoot: X509Certificate, certPath: CertPath) {
         val params = PKIXParameters(setOf(TrustAnchor(trustedRoot, null)))
         params.isRevocationEnabled = false
-        CertPathValidator.getInstance("PKIX").validate(certPath, params)
+        try {
+            CertPathValidator.getInstance("PKIX").validate(certPath, params)
+        } catch (e: CertPathValidatorException) {
+            throw CertPathValidatorException(
+                    """Cert path failed to validate against root certificate. Reason=${e.reason} Offending cert index=${e.index}
+$certPath
```
kchalkias: Perhaps have `Offending cert index=${e.index}` in its own line 
and add `CertPath:` as you did with root certificate, so it's consistent and clear when reading it.

PR #2891, opened 2018-03-28
ENT-1686: Improve error message on parameters mismatch

kasiastreich:


node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -114,7 +114,8 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
             // TODO This needs special handling (node omitted update process/didn't accept new parameters or didn't restart on updateDeadline)
             logger.error("Node is using parameters with hash: $currentParametersHash but network map is " +
                     "advertising: ${networkMap.networkParameterHash}.\n" +
-                    "Please update node to use correct network parameters file.\"")
+                    "Node will shutdown now, if you accepted new network parameters it is sufficient to start it again.\n" +
```
shamsasari: I'd like this to be even less ambiguous. The node should cache the last parameters-update info and then check if the switch has occurred (by checking the update hash with the new hash) and print a more relevant message.

PR #2892, opened 2018-03-28
[CORDA-1264]: Ensure correct serialisation and masking for throwables raised by a node and propagated through RPC.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1264


node-api/src/main/kotlin/net/corda/nodeapi/exceptions/InternalNodeException.kt
```diff
@@ -0,0 +1,25 @@
+package net.corda.nodeapi.exceptions
+
+import net.corda.core.serialization.CordaSerializable
+
+/**
+ * An [Exception] to signal RPC clients that something went wrong within a Corda node.
+ */
+@CordaSerializable
+class InternalNodeException private constructor(message: String) : Exception(message) {
```
fenryka: Is this actually serializable as on deserialization we won't invoke the wrap method on the companion object?

fenryka: This is going to break with AMQP, it might work now because Kryo

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/adapters/FlowProgressHandleErrorMappingAdapter.kt
```diff
@@ -0,0 +1,22 @@
...
+import net.corda.core.mapErrors
+import net.corda.core.messaging.FlowProgressHandle
+import net.corda.core.serialization.CordaSerializable
+import net.corda.nodeapi.exceptions.InternalNodeException
+
+/**
+ * Adapter able to mask errors within a Corda node for RPC clients.
+ */
+@CordaSerializable
+data class FlowProgressHandleErrorMappingAdapter<RESULT>(private val wrapped: FlowProgressHandle<RESULT>) : FlowProgressHandle<RESULT> by wrapped {
```
fenryka: we support private vals but it's going to be deprecated so either make public or add a getter

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/InternalNodeException.kt
sollecitom: Constructor is now public.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/adapters/FlowProgressHandleErrorMappingAdapter.kt
sollecitom: Made it public.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/adapters/FlowHandleErrorMappingAdapter.kt
```diff
@@ -0,0 +1,15 @@
...
+import net.corda.core.internal.concurrent.mapError
+import net.corda.core.messaging.FlowHandle
+import net.corda.core.serialization.CordaSerializable
+import net.corda.nodeapi.exceptions.InternalNodeException
+
+/**
+ * Adapter able to mask errors within a Corda node for RPC clients.
+ */
+@CordaSerializable
+data class FlowHandleErrorMappingAdapter<RESULT>(private val wrapped: FlowHandle<RESULT>) : FlowHandle<RESULT> by wrapped {
```
sollecitom: @fenryka fixed this too.

fenryka:
As mentioned need BP to 3 branch, merge to R3C and then bp to R3C3

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/WithClientRelevantMessage.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.nodeapi.exceptions
+
+import net.corda.core.serialization.CordaSerializable
+
+/**
+ * Mark an [Exception] to be shown to RPC clients.
+ */
+@CordaSerializable
+interface WithClientRelevantMessage
```
szymonsztuka: This is like an annotation / marker, and it doesn't impose any rule while developing - i.e. someone may implement this for "non public" error. 
What about removing this interface (so we are not changing API) and move logic to InternalNodeException e.g.:
`when{
wrapped is ClientRelevantException || PermissionException || ...(all 4 implementing classes) -> { 
                    wrapped.setCause(null)
                    wrapped
} 
else -> ...`

So only in `InternalNodeException` we would decide what's public and what isn't.
However, if the need this logic somewhere else we would potentially duplicate the code this. So I see pros and cons, what do you think?

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/InternalNodeException.kt
```diff
@@ -0,0 +1,32 @@
...
+            return when {
+                wrapped is CordaRuntimeException && wrapped is WithClientRelevantMessage -> {
+                    wrapped.setCause(null)
+                    wrapped
+                }
+                else -> {
+                    when (wrapped) {
+                        is WithClientRelevantMessage -> InternalNodeException(wrapped.message ?: DEFAULT_MESSAGE)
+                        else -> InternalNodeException(DEFAULT_MESSAGE)
+                    }
```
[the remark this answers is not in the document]
sollecitom: 2 issues here:

1. Some exceptions are in the wrong module, and changing their location now is unfortunately a breaking change.
2. I need both a whitelist and the marker interface, for some serialisation-related errors should be propagated and their Java types.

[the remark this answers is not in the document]
sollecitom: Re-worked the condition to be much leaner.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/WithClientRelevantMessage.kt
sollecitom: As discussed, unfortunately not possible (in particular, `PermissionException` is in `corda-rpc-client` rather than `node-api`.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/InternalNodeException.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+class InternalNodeException(message: String) : CordaClientException(message) {
+
+    companion object {
+
+        private const val DEFAULT_MESSAGE = "Something went wrong within the Corda node."
+
+        fun defaultMessage(): String = DEFAULT_MESSAGE
+
+        fun wrap(wrapped: Throwable): Throwable {
```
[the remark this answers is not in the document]
sollecitom: Renamed to `obfuscateIfInternal`.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/adapters/FlowHandleErrorMappingAdapter.kt
```diff
@@ -0,0 +1,15 @@
...
+import net.corda.core.internal.concurrent.mapError
+import net.corda.core.messaging.FlowHandle
+import net.corda.core.serialization.CordaSerializable
+import net.corda.nodeapi.exceptions.InternalNodeException
+
+/**
+ * Adapter able to mask errors within a Corda node for RPC clients.
+ */
+@CordaSerializable
+data class FlowHandleErrorMappingAdapter<RESULT>(val wrapped: FlowHandle<RESULT>) : FlowHandle<RESULT> by wrapped {
```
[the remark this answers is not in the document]
sollecitom: Renamed to `InternalObfuscatingFlowHandle` and `InternalObfuscatingFlowProgressHandle`

PR #2905, opened 2018-04-03
[CORDA-1294]: Publish corda-shell module.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1294


tools/shell/build.gradle
```diff
@@ -90,3 +90,7 @@ task integrationTest(type: Test) {
     testClassesDirs = sourceSets.integrationTest.output.classesDirs
     classpath = sourceSets.integrationTest.runtimeClasspath
 }
+
+publish {
+    name jar.baseName
+}
```
chrisr3: You also need:
```gradle
apply plugin: 'net.corda.plugins.publish-utils'
```
to make this work.

PR #2908, opened 2018-04-03
ENT-1732: Fixing SignedNodeInfo security issue

mkit:
This fixes the security vulnerability described in [ENT-1732](https://r3-cev.atlassian.net/browse/ENT-1732).


vkolomeyko:
Perhaps, I would be a good idea to add a unit test as well...

PR #2909, opened 2018-04-03
CORDA-1276: Cleaned up creation of node-info object on node start and…

shamsasari:
… when using --just-generate-node-info


shamsasari:
Yes, I've reverted that change back to current time millis.

PR #2910, opened 2018-04-03
Corda-1286 Fix embedded shell when rpc users from remoted database

szymonsztuka:
The issue was only in dev mode as embedded shell is disable otherwise.
The problem appears to be that the CRaSH shell adds its own shell user to the list of INMEMORY RPC users, but a DB data source requires that this list is null. 

Fix by: When devMode and running embedded shell: wrap RPC Security Manager with a dev one, which intercepts authorisation for a given user and store it's credentials in memory. By the default it's "shell/shell" user, the embedded shell connects to RCP via this user.


node/src/main/kotlin/net/corda/node/internal/security/RPCSecurityManagerWithLocalShellUser.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.shiro.mgt.DefaultSecurityManager
+import org.apache.shiro.subject.SimplePrincipalCollection
+import javax.security.auth.login.FailedLoginException
+
+/**
+ * Wrapper for [RPCSecurityManager] which creates in-memory [AuthorizingSubject] for [User].
+ * Can be used to add on a specific [User] on top of the principals provided by the [RPCSecurityManager] realm.
+ * By the default grants "ALL" permissions to "shell/shell" user/password.
+ */
+class RPCSecurityManagerWithLocalShellUser(private val delegate: RPCSecurityManager, private val user: User = localShellUser()) : RPCSecurityManager by delegate {
```
sollecitom: Just one thing: please move this default value out, and require a non-null `User`. Pass the `localShellUser()` from `Node`.

PR #2920, opened 2018-04-04
Link to X509Utils. Makes it clear they are only an example.

joeldudleyr3:


docs/source/permissioning.rst
```diff
@@ -75,6 +75,10 @@ public/private keypairs and certificates. The keypairs and certificates should o
 
     * RSA with 3072-bit key size
 
+.. note:: Corda's ``X509Utilities`` show how to generate the required public/private keypairs and certificates using
+   Bouncy Castle. You can find the ``X509Utilities`` in the `Corda repository <https://github.com/corda/corda>`_, under
```
shamsasari: You don't need this second sentence. Just make the above  ``X509Utilities`` the link text.

joeldudleyr3: If we do that, we have to update it every time we change Corda version, as the link will include the branch name.

shamsasari: OK

PR #2930, opened 2018-04-05
This prevents log4j2 shutting down logging using a shutdown hook

mnesbit:
At the minute the log4j2 logging is registering a shutdown hook, which cleans up when invoked. Unfortunately, Corda uses shutdown hooks to run its own shutdown tasks, so we stop logging partly way through close.

### DIFF
--- a/config/dev/log4j2.xml
+++ b/config/dev/log4j2.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Configuration status="info">
+<Configuration status="info" shutdownHook="disable">
 
     <Properties>
         <Property name="log-path">logs</Property>


Clintonio (dismissed):

Clintonio:
It seems to pass CI so I'm assuming this doesn't cause any serious issues with shutdown. I found this bit of code;

            //shutdown log4j2
            if( LogManager.getContext() instanceof LoggerContext ) {
                logger.info("Shutting down log4j2");
                Configurator.shutdown((LoggerContext)LogManager.getContext());
            } else
                logger.warn("Unable to shutdown log4j2");

[Source](https://stackoverflow.com/questions/17400136/how-to-log-within-shutdown-hooks-with-log4j2)

We might want to add it in this PR still.

rick-r3:
@Clintonio Wouldn't that create a hard dependency on log4j2 when we are trying to indirect via SLF4J and leave the actual logging as a runtime dependency ?

Clintonio:
Ah, my understanding of this was upside down. I assumed log4j was the lower level API layer and slf4j was the implementation.

Clintonio:
I think we'd have to investigate if there's any standard shutdown option between them before merging. Will dismiss my review.

Clintonio:
https://www.slf4j.org/api/org/apache/log4j/LogManager.html

So looking at that we have [shutdown](https://www.slf4j.org/api/org/apache/log4j/LogManager.html#shutdown(\)). It's a noop by default so I'm assuming that this would be implemented by Log4J, but such an assumption needs to be verified.

mikehearn:
Is this PR stalled? Can it be closed or merged?

mnesbit:
Got stalled. I still think it is still useful, but perhaps there are better ways of doing this.

mnesbit:
Nobody seems to be shouting for this so closing

PR #2931, opened 2018-04-05
Minor: use of ByteArray.inputStream() and introduced PublicKey.hash

shamsasari:


experimental/kryo-hook/src/main/kotlin/net/corda/kryohook/KryoHook.kt
```diff
@@ -5,13 +5,25 @@ import com.esotericsoftware.kryo.Kryo
...
+import java.lang.Thread
 import java.lang.instrument.ClassFileTransformer
 import java.lang.instrument.Instrumentation
 import java.security.ProtectionDomain
 import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.atomic.AtomicInteger
+import kotlin.Any
+import kotlin.ByteArray
+import kotlin.IllegalStateException
+import kotlin.Int
```
rick-r3: What's happened here?  Seems abnormal?

PR #2934, opened 2018-04-05
Remove `errorLogPath` parameter which is not used anyway.

vkolomeyko:
StdErr is always written to working directory if such path specified.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/ProcessUtilities.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/ProcessUtilities.kt
@@ -11,7 +11,7 @@ object ProcessUtilities {
             jdwpPort: Int? = null,
             extraJvmArguments: List<String> = emptyList()
     ): Process {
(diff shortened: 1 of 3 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/ProcessUtilities.kt
```diff
@@ -34,7 +33,6 @@ object ProcessUtilities {
             classpath: String,
             jdwpPort: Int?,
             extraJvmArguments: List<String>,
-            errorLogPath: Path?,
             workingDirectory: Path?,
```
shamsasari: If `workingDirectory` is not specified then the output is redirected to any log files. Perhaps we should make this non-null.

vkolomeyko: There is a number of usages where people call this method with `workignDirectory` being `null`.
This should be fine for output redirection since we always call `inheritIO()` which "Sets the source and destination for subprocess standard I/O to be the same as those of the current Java process."

PR #2935, opened 2018-04-05
CORDA-1303 Regression: Recording a duplicate transaction attempts sec…

rick-r3:
…ond insert to vault.

Added unit test, and merged mock and real logic.


node/src/main/kotlin/net/corda/node/services/api/ServiceHubInternal.kt
```diff
@@ -47,6 +47,61 @@ interface NetworkMapCacheBaseInternal : NetworkMapCacheBase {
...
+                // solution. Then you could select subsets of states depending on where the report came from.
+                //
+                // The risk of doing this is that apps/developers may use 'canned SQL queries' not written by us that forget
+                // to add a WHERE clause for the origin column. Those queries will seem to work most of the time until
+                // they're run on an observer node and mix in irrelevant data. In the worst case this may result in
+                // erroneous data being reported to the user, which could cause security problems.
+                //
+                // Because the primary use case for recording irrelevant states is observer/regulator nodes, who are unlikely
+                // to make writes to the ledger very often or at all, we choose to punt this issue for the time being.
+                vaultService.notifyAll(statesToRecord, recordedTransactions.map { it.coreTransaction })
```
rick-r3: This line changed.... everything else is to enable the tests and to consolidate duplicate buggy code in mocks.

node/src/main/kotlin/net/corda/node/services/api/ServiceHubInternal.kt
```diff
@@ -47,6 +47,61 @@ interface NetworkMapCacheBaseInternal : NetworkMapCacheBase {
...
+                               validatedTransactions: WritableTransactionStorage,
+                               stateMachineRecordedTransactionMapping: StateMachineRecordedTransactionMappingStorage,
+                               vaultService: VaultServiceInternal) {
+
+            require(txs.any()) { "No transactions passed in for recording" }
+            val recordedTransactions = txs.filter { validatedTransactions.addTransaction(it) }
+            val stateMachineRunId = FlowStateMachineImpl.currentStateMachine()?.id
+            if (stateMachineRunId != null) {
+                recordedTransactions.forEach {
+                    stateMachineRecordedTransactionMapping.addMapping(stateMachineRunId, it.id)
```
adagys: this will throw an exception when called from mock services, can this line ever be reached?

rick-r3: It could be, but very unlikely.  It would require the test to be using mock services and the actual state machine, which doesn't seem to be a combination we have based on passing tests and slightly tricky to organise I think.  If it does happen, we'll have to implement it differently.

PR #2941, opened 2018-04-06
Do not perform deep scan when assembling Dev Cordap Jar.

vkolomeyko:
Note: It is still possible to list every necessary package explicitly.
This is important as package name of the integration test is automatically included into list of packages to scan.
E.g. In case of integration test `DistributedServiceTests` the package is `net.corda.node.services` which is quite extensive, resulting big attachments (multiple MB) to be produced
and handled. This creates a notable delay in test execution.

### DIFF
--- a/client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
+++ b/client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
@@ -11,6 +11,7 @@ import net.corda.core.flows.StateMachineRunId
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.internal.bufferUntilSubscribed
+import net.corda.core.internal.packageName
 import net.corda.core.messaging.CordaRPCOps
 import net.corda.core.messaging.StateMachineTransactionMapping
 import net.corda.core.messaging.StateMachineUpdate
@@ -54,7 +55,7 @@ class NodeMonitorModelTest {
     private lateinit var newNode: (CordaX500Name) -> NodeInfo
 
     private fun setup(runTest: () -> Unit) {
-        driver(DriverParameters(extraCordappPackagesToScan = listOf(Cash::class.java.`package`.name))) {
+        driver(DriverParameters(extraCordappPackagesToScan = listOf(Cash::class.packageName))) {
             val cashUser = User("user1", "test", permissions = setOf(
                     startFlow<CashIssueFlow>(),
                     startFlow<CashPaymentFlow>(),
--- a/client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
+++ b/client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
@@ -23,6 +23,7 @@ import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.utilities.getOrThrow
 import net.corda.finance.DOLLARS
 import net.corda.finance.USD
+import net.corda.finance.contracts.asset.Cash
 import net.corda.finance.flows.CashExitFlow
 import net.corda.finance.flows.CashIssueFlow
 import net.corda.finance.flows.CashPaymentFlow
@@ -53,7 +54,7 @@ class NodeMonitorModelTest {
     private lateinit var newNode: (CordaX500Name) -> NodeInfo
 
     private fun setup(runTest: () -> Unit) {
-        driver(DriverParameters(extraCordappPackagesToScan = listOf("net.corda.finance"))) {
+        driver(DriverParameters(extraCordappPackagesToScan = listOf(Cash::class.java.`package`.name))) {
             val cashUser = User("user1", "test", permissions = setOf(
                     startFlow<CashIssueFlow>(),
                     startFlow<CashPaymentFlow>(),
--- a/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
@@ -372,7 +372,7 @@ inline fun <reified T : ContractState> VaultService.trackBy(criteria: QueryCrite
     return _trackBy(criteria, paging, sorting, T::class.java)
 }
 
-class VaultQueryException(val description: String) : FlowException(description)
+class VaultQueryException(override val message: String) : FlowException(message)
 
 class StatesNotAvailableException(override val message: String?, override val cause: Throwable? = null) : FlowException(message, cause) {
     override fun toString() = "Soft locking error: $message"
--- a/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
@@ -372,7 +372,7 @@ inline fun <reified T : ContractState> VaultService.trackBy(criteria: QueryCrite
     return _trackBy(criteria, paging, sorting, T::class.java)
 }
 
-class VaultQueryException(description: String) : FlowException(description)
+class VaultQueryException(val description: String) : FlowException(description)
 
 class StatesNotAvailableException(override val message: String?, override val cause: Throwable? = null) : FlowException(message, cause) {
     override fun toString() = "Soft locking error: $message"
--- a/finance/src/test/kotlin/net/corda/finance/contracts/CommercialPaperTests.kt
+++ b/finance/src/test/kotlin/net/corda/finance/contracts/CommercialPaperTests.kt
@@ -88,6 +88,8 @@ class KotlinCommercialPaperLegacyTest : ICommercialPaperTestTemplate {
     override fun getContract() = CommercialPaper.CP_PROGRAM_ID
 }
 
+private val cordappPackages = listOf(Cash::class.java.`package`.name, CommercialPaper::class.java.`package`.name)
+
 @RunWith(Parameterized::class)
 class CommercialPaperTestsGeneric {
     companion object {
@@ -99,6 +101,14 @@ class CommercialPaperTestsGeneric {
         private val dummyNotary = TestIdentity(DUMMY_NOTARY_NAME, 20)
         private val alice = TestIdentity(ALICE_NAME, 70)
         private val miniCorp = TestIdentity(CordaX500Name("MiniCorp", "London", "GB"))
+
+        private fun createMockServices(firstIdentity: TestIdentity, vararg moreIdentities: TestIdentity) : MockServices {
+            return MockServices(
+                    cordappPackages,
+                    firstIdentity,
+                    makeTestIdentityService(*listOf(firstIdentity, *moreIdentities).map { it.identity }.toTypedArray()),
+                    firstIdentity.keyPair)
+        }
     }
 
     @Parameterized.Parameter
@@ -109,7 +119,7 @@ class CommercialPaperTestsGeneric {
     val testSerialization = SerializationEnvironmentRule()
 
     private val megaCorpRef = megaCorp.ref(123)
-    private val ledgerServices = MockServices(megaCorp, miniCorp)
+    private val ledgerServices = createMockServices(megaCorp, miniCorp)
 
     @Test
     fun `trade lifecycle test`() {
@@ -245,10 +255,10 @@ class CommercialPaperTestsGeneric {
         // of the dummy cash issuer.
 
         val allIdentities = arrayOf(megaCorp.identity, miniCorp.identity, dummyCashIssuer.identity, dummyNotary.identity)
-        val notaryServices = MockServices(dummyNotary)
-        val issuerServices = MockServices(dummyCashIssuer, dummyNotary)
+        val notaryServices = createMockServices(dummyNotary)
+        val issuerServices = createMockServices(dummyCashIssuer, dummyNotary)
         val (aliceDatabase, aliceServices) = makeTestDatabaseAndMockServices(
-                listOf("net.corda.finance.contracts"),
+                cordappPackages,
                 makeTestIdentityService(*allIdentities),
                 alice
         )
@@ -257,7 +267,7 @@ class CommercialPaperTestsGeneric {
         }
 
         val (megaCorpDatabase, megaCorpServices) = makeTestDatabaseAndMockServices(
-                listOf("net.corda.finance.contracts"),
+                cordappPackages,
                 makeTestIdentityService(*allIdentities),
                 megaCorp
         )
--- a/finance/src/test/kotlin/net/corda/finance/contracts/CommercialPaperTests.kt
+++ b/finance/src/test/kotlin/net/corda/finance/contracts/CommercialPaperTests.kt
@@ -4,6 +4,7 @@ import net.corda.core.contracts.*
 import net.corda.core.identity.AnonymousParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
+import net.corda.core.internal.packageName
 import net.corda.core.node.services.Vault
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.transactions.TransactionBuilder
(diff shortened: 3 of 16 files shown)


shamsasari:
What speedup are you getting with these changes?

vkolomeyko:
In terms of speed-up, I have to say "it is less than I was hoping for".

The reality is that under CI timing of integration tests run even in `master` varies quite a bit: https://ci-master.corda.r3cev.com/project.html?projectId=Corda_Build&buildTypeId=&tab=testDetails&testNameId=808452641780530985&order=START_DATE_DESC&branch_Corda_Build=__all_branches__&itemsCount=50
So it is difficult to quantify.

Anyhow, I think it is a worthwhile change as it results tighter control over attachments used in integration tests and smaller contract attachments. Also, it eliminates some of flakiness resulted from using `BiMap`.

vkolomeyko:
All,
Could you please review these changes and let me know what you think?

Thanks,
Viktor

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -137,7 +137,7 @@ class CordappLoader private constructor(private val cordappJarPaths: List<Restri
                 FileOutputStream(cordappJAR).use {
                     JarOutputStream(it).use { jos ->
                         val scanDir = File(path.toURI())
-                        scanDir.walkTopDown().forEach {
+                        scanDir.listFiles().filter { it.isFile }.forEach {
```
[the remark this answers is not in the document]
shamsasari: Isn't this changing the semantics of the public `extraCordappPackagesToScan`? You say the speedup isn't as much as you had expected, so I'm wary of this change.

Perhaps for `DistributedServiceTests` just move it into a sub-package?

client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
```diff
@@ -53,7 +54,7 @@ class NodeMonitorModelTest {
     private lateinit var newNode: (CordaX500Name) -> NodeInfo
 
     private fun setup(runTest: () -> Unit) {
-        driver(DriverParameters(extraCordappPackagesToScan = listOf("net.corda.finance"))) {
+        driver(DriverParameters(extraCordappPackagesToScan = listOf(Cash::class.java.`package`.name))) {
```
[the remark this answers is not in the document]
vkolomeyko: Agreed - this is neater, I will make this change across the board.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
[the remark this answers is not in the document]
vkolomeyko: Method name is `createDevCordappJar()` and this is indeed only used during CI or local test run - no production impact. @Clintonio - could you opine here?

With this in mind - this is relatively safe change to make.

As for performance improvements for attachements scanning, more changes in the pipeline, e.g.: #2982
But still having smaller attachments JARs would be valuable and to me is a sensible thing to do.

mikehearn:
I can set aside some time to review this if you like but I don't know much about the relevant code areas. It seems like Shams is on top of it and hopefully Clinton can help too?

Thanks for optimising the tests - it's well overdue and I'm happy to see we're really tackling it now!

vkolomeyko:
Thanks, @mikehearn - not sure why GitHub suggested your name in the list of reviewers.
I think we have a good handle on this PR, but please feel free to review this PR if you wish/have time.

There are two more PRs open by me atm which are also trying to reduce amount of classpath scanning we do.

vkolomeyko:
Rebased to the latest `master` to apply my changes on top and resolve any conflicts.
Should be all good now and ready for review.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
```diff
@@ -133,7 +130,7 @@ class CordappLoader private constructor(private val cordappJarPaths: List<Restri
                 logger.info("Generating a test-only cordapp of classes discovered in $scanPackage at $cordappJAR")
                 JarOutputStream(cordappJAR.outputStream()).use { jos ->
                     val scanDir = url.toPath()
-                    scanDir.walk { it.forEach {
+                    scanDir.toFile().listFiles().filter { it.isFile }.map { it.toPath() }.forEach {
```
vkolomeyko: The "meat" of the change.

vkolomeyko:
A very gentle reminder to review and approve this change.
It would be a shame to get this PR into conflicted state again.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
shamsasari: Apologies for the delay. It's taken longer than it should have for me to make sense of these changes.

This is API breaking. CorDapp devs who have tests using the driver with `DriverParameters.extraCordappPackagesToScan` will break if they're relying on the scanning to include sub-packages (intended feature). Otherwise asking them to list each sub-package is not very user friendly.

As I suggested before, we either move the offending tests up the package hierarchy if possible, or we provide a separate code path in the driver for our use only - this is possible with the `internalDriver`.

A third option is to reconsider this entire `createDevCordappJar` concept. The original code had the node scan the relevant packages directly. Now, we first bundle up the classes found in the packages-to-scan into a temp jar and then have the node scan that. This is wasteful and should be tackled instead.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
```diff
@@ -80,19 +88,60 @@ open class CordappProviderImpl(private val cordappLoader: CordappLoader,
...
+     *
+     *     2. contractClassNames=[net.corda.finance.contracts.asset.Cash, net.corda.finance.contracts.asset.CommodityContract, net.corda.finance.contracts.asset.Obligation, net.corda.finance.contracts.asset.OnLedgerAsset] ->
+     *          jarPath=file:/Z:/corda/finance/build/libs/corda-finance-corda-4.0-snapshot.jar
+     *
+     *  These will be represented as distinct attachments with different hashes. Since underlying storage is `HashBiMap` when look-up is the ordering is not guaranteed and given contract e.g. `net.corda.finance.contracts.asset.Cash`
+     *  can be found in either of the Jars which causes flakiness.
+     *
+     *  To mitigate that we de-dupe CorDapps based on the same `contractClassNames`.
+     */
+    private fun List<Cordapp>.deDupeSameContractClassNames(): List<CorDappLight> {
```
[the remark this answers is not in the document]
shamsasari: This seems like an unrelated issue. What's the flakiness?

But I wonder if the solution isn't to just not produce the temp cordapp jar? Based on my comment above that might be possible, no?

core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
```diff
@@ -372,7 +372,7 @@ inline fun <reified T : ContractState> VaultService.trackBy(criteria: QueryCrite
     return _trackBy(criteria, paging, sorting, T::class.java)
 }
 
-class VaultQueryException(description: String) : FlowException(description)
+class VaultQueryException(val description: String) : FlowException(description)
```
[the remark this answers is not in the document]
vkolomeyko: Sure - will rename this to `message`.
The reason why I touched this class is because with my changes one of integration tests was failing and `VaultQueryException` in its present form is not AMQP serializable.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
vkolomeyko: In terms of API breaking:
It is not a compile time break, but I do agree it changes runtime behaviour. But this is only happens if the user writes an integration test with `extraCordappPackagesToScan` and if they implicitly rely on deep scan of the packages hierarchy.
I think we can make a note of this change of behaviour in the release notes and (hopefully small number of) affected people will need to change their code.
@mikehearn - what do you think about it?
If this not acceptable then I would agree with @shamsasari and "we provide a separate code path in the driver for our use only". But this add even more API methods/parameters to Node Driver API.

In terms entire `createDevCordappJar()` concept:
I am new to this code, but my understanding that need for `createDevCordappJar()` comes from the fact that contract/flow attachment always has to be a Jar file. Now if people create a custom contract/flow as an inner class in their test - that has to be packaged into a Jar in order to be attached - this is, in my understanding, what `createDevCordappJar()` is primarily trying to address.
@Clintonio - could you confirm if my understanding of the use case is correct?

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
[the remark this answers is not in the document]
vkolomeyko: The flakiness come from the fact: that `net.corda.finance.contracts.asset.Cash` can be found in either of the Jars which causes the issue. I deliberately put verbose comment with concrete examples I observed in the debugger.

This may seem like an unrelated issue, but this is something I stumbled upon when debugging one of the integration tests with my changes. Perhaps, because now attachments are smaller and there are greater quantities of them the problem started to manifest itself more frequently.
But even without my changes, I do remember seeing in the past some flaky tests when a `Cash` contract cannot be found in attachment.

client/jfx/src/integration-test/kotlin/net/corda/client/jfx/NodeMonitorModelTest.kt
```diff
@@ -53,7 +55,7 @@ class NodeMonitorModelTest {
     private lateinit var newNode: (CordaX500Name) -> NodeInfo
 
     private fun setup(runTest: () -> Unit) {
-        driver(DriverParameters(extraCordappPackagesToScan = listOf("net.corda.finance"))) {
+        driver(DriverParameters(extraCordappPackagesToScan = listOf(Cash::class.packageName))) {
```
Clintonio: I am fine with this change but when I originally put this in place I was asked to turn it into a string to avoid reflection, though I think that's a bit of premature optimising given that these are tests.

core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
```diff
@@ -372,7 +372,7 @@ inline fun <reified T : ContractState> VaultService.trackBy(criteria: QueryCrite
     return _trackBy(criteria, paging, sorting, T::class.java)
 }
 
-class VaultQueryException(description: String) : FlowException(description)
+class VaultQueryException(override val message: String) : FlowException(message)
```
Clintonio: Why change this to an override val? This means that it becomes a part of the API rather than delegating to the underlying object's message field.

finance/src/test/kotlin/net/corda/finance/contracts/CommercialPaperTests.kt
```diff
@@ -88,6 +89,8 @@ class KotlinCommercialPaperLegacyTest : ICommercialPaperTestTemplate {
     override fun getContract() = CommercialPaper.CP_PROGRAM_ID
 }
 
+private val cordappPackages = listOf(Cash::class.packageName, CommercialPaper::class.packageName)
```
Clintonio: This would make sense to put in the companion object.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappLoader.kt
mikehearn: I'm gonna side with Shams here I'm afraid. That would be a behavioural change that isn't backwards compatible. How about moving the unit tests around? That seems like an easy thing to do?

Clintonio: In a more ideal world I'd never have had to write this code and had more time to figure out a way to do this without creating a JAR (and there are ways, time didn't permit it since we were rapidly approaching V1). 

Your understanding is correct - this does a deep scan, it was intentional and I was torn at the time as to whether this was a good idea. Performance was not my consideration, usability was, but if this is providing any meaningful speedup then I believe it to be worth it. It would be a function contract break though (not compile time API break).

My suggestion is this; instead of _replacing_ this function with a more performant one - add a new one that is more performant and allow the user to decide which they want. Or add a flag to disable recursion. That way we maintain the old functionality and add new functionality to help where it affects performance.

I'd also file a JIRA ticket to look into having some kind of test attachment store that can handle a listOf(Class<*>) so we don't need to do any runtime JAR creation.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
```diff
@@ -25,6 +25,10 @@ open class CordappProviderImpl(private val cordappLoader: CordappLoader,
 
     companion object {
         private val log = loggerFor<CordappProviderImpl>()
+
+        private data class CorDappLight(val contractClassNames: Set<String>, val jarPath: URL) {
```
Clintonio: This name should be consistent with the lower case D in code. While the name out of code is officially "CorDapp" it sort of breaks camelcase rules in code (since it isn't a cor Dapp, it's more of a play on words with "Corda Distributed App").

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
```diff
@@ -71,7 +79,7 @@ open class CordappProviderImpl(private val cordappLoader: CordappLoader,
     }
 
     override fun getContractAttachmentID(contractClassName: ContractClassName): AttachmentId? {
-        return getCordappForClass(contractClassName)?.let(this::getCordappAttachmentId)
+        return cordappAttachments.entries.find { it.value.contractClassNames.contains(contractClassName) }?.key
```
Clintonio: This should be reverted. The previous code operated at the right level of abstraction;

1. get the cordapp for this class
2. get the attachment ID for this cordapp

The replacement code delves deeper into the internal structure that may change in future whereas the function contracts for "getCordappForClass" and "getCordappAttachmentId" won't because they exist on the public API.

node/src/main/kotlin/net/corda/node/internal/cordapp/CordappProviderImpl.kt
```diff
@@ -80,19 +88,60 @@ open class CordappProviderImpl(private val cordappLoader: CordappLoader,
...
+            } else {
+                val modifiedContractsList = corDapp.contractClassNames - notedContracts
+                val modifiedCordapp = corDapp.copy(contractClassNames = modifiedContractsList)
+                log.warn("Cordapp $corDapp has it contracts list modified to: $modifiedContractsList")
+                acc += Pair(modifiedCordapp.jarPath, modifiedCordapp)
+                notedContracts += modifiedContractsList
+            }
+            Pair(acc, notedContracts)
+        }
+        return foldResult.first.values.toList()
```
Clintonio: This may solve the bug but it isn't the correct solution - we actually want to allow two cordapps with the same contract class name (consider over two versions).

It is a known issue that two cordapps can't exist with the same contract name right now but the solution isn't to automatically exclude one, it is to correct the entire cordapp system and sandbox, and there is a team forming to do that soon.

I recommend instead replacing this with a piece of analytical code which simply looks for duplicated contract classnames and warns the user that _currently_ this is unsupported. Maybe it will never be and it becomes a fatal startup error instead, but for now we aren't saying it isn't or is possible yet (until the versioning and modules teams forms). And automatically modifying cordapps on the fly is dangerous and brittle territory to enter.

The flaky test that this solves needs to be fixed in another way - especially now we know _why_ it's flaky.

vkolomeyko:
@shamsasari @mikehearn and @Clintonio - thanks a lot for spending time reviewing my work!
Given complexity around it and given limited speed-up effect it produced during integration tests execution, I am tempted to close this PR without merging into `master`.
Please let me know if you feel that this is not the right way to go and you strongly feel that we need to invest more effort in this particular change.

vkolomeyko:
I have posted a separate PR for moving `DistributedServiceTests` into a separate package: #3021.
Is there anything else we want to retain from this PR before closing it?

PR #2949, opened 2018-04-10
Fix cash selection with PostgreSQL

szymonsztuka:
Change conversion to toStringShort() instead of toBase58String() - as done for H2 Cash Selection.
Fix withIssuerRefs case - iterate via list of IssuerRefs and setBytes instead of setArray of BYTEA.
Add test for Cash Selection with issuerRef.


josecoll:
Is this fix not required for DP3 ?

sollecitom:
@josecoll we decided it's not, but if it gets in master we should attempt to put it in DP3 regardless.

PR #2951, opened 2018-04-10
Add support for different internal p2p artemis bind address/port

mnesbit:
This adds a messagingServerExternal flag to the node configuration, which controls whether to start an in-process p2p artemis. By default the messagingServerExternal setting follows the existing p2p artemis logic. Thus if messagingServerAddress is null (default) then messagingServerExternal is false and the node starts an internal artemis server bound to the p2pAddress port on all interface cards. If messagingServerAddress is specified then by default messagingServerExternal is set to true and the node will connect to an externally provided broker on the messagingServerAddress and advertise the public p2pAddress to the network map. However, now the messagingServerExternal can be forced to false in config and the node will start an internal p2p broker bound to the messagingServerAddress.

This allows situations where there is a bridge, or other port translation between the nodes public address and the local machine address/port. It also allows resolution of the interface card when operating Corda on multi-homed machines.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/Node.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/Node.kt
@@ -160,14 +160,13 @@ open class Node(configuration: NodeConfiguration,
                                       networkParameters: NetworkParameters): MessagingService {
         // Construct security manager reading users data either from the 'security' config section
         // if present or from rpcUsers list if the former is missing from config.
-        val securityManagerConfig = configuration.security?.authService ?:
-        SecurityConfiguration.AuthService.fromUsers(configuration.rpcUsers)
+        val securityManagerConfig = configuration.security?.authService ?: SecurityConfiguration.AuthService.fromUsers(configuration.rpcUsers)
 
         securityManager = with(RPCSecurityManagerImpl(securityManagerConfig)) {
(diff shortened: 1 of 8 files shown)


node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -70,7 +68,7 @@ import javax.security.auth.spi.LoginModule
  */
 @ThreadSafe
 class ArtemisMessagingServer(private val config: NodeConfiguration,
-                             private val p2pPort: Int,
+                             private val messagingServerAddress:NetworkHostAndPort,
```
andr3ej: formatting

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -167,7 +167,12 @@ open class Node(configuration: NodeConfiguration,
             if (configuration.shouldInitCrashShell()) RPCSecurityManagerWithAdditionalUser(this, localShellUser()) else this
         }
 
-        val serverAddress = configuration.messagingServerAddress ?: makeLocalMessageBroker(networkParameters)
+        if(!configuration.messagingServerExternal) {
```
andr3ej: run formatter on new code

PR #2962, opened 2018-04-12
CORDA-1337 Crypto.registerProviders

kchalkias:
+ register BCPQC as well.


core/src/main/kotlin/net/corda/core/crypto/internal/ProviderMap.kt
```diff
@@ -37,6 +37,8 @@ internal val cordaBouncyCastleProvider = BouncyCastleProvider().apply {
 }
 internal val bouncyCastlePQCProvider = BouncyCastlePQCProvider().apply {
     require(name == "BCPQC") // The constant it comes from is not final.
+}.also {
+    Security.addProvider(it)
```
patrickkuo: as discussed, I think it is better to have all the security wiring in the registerProviders method, instead of having it every where when constructing the provider.

kchalkias: Thanks @patrickkuo, as discussed, we've agreed to just have the helper function `registerProviders()` and avoid this refactoring for the time being. I've added a related TODO though.

PR #2970, opened 2018-04-16
CORDA-1344 Fix query paging in scheduled flow tests

anthonykr3:
- Fix issue with paging integration example in ScheduledFlowTests when number of results % page size != 0
- Move paging example to VaultQuery tests
- Write some tests for the example that just test that
- Write java version of pagination example and add to docs

### DIFF
--- a/node/src/test/kotlin/net/corda/node/services/events/ScheduledFlowTests.kt
+++ b/node/src/test/kotlin/net/corda/node/services/events/ScheduledFlowTests.kt
@@ -11,8 +11,6 @@ import net.corda.core.flows.SchedulableFlow
 import net.corda.core.identity.Party
 import net.corda.core.node.services.VaultService
 import net.corda.core.node.services.queryBy
-import net.corda.core.node.services.vault.DEFAULT_PAGE_NUM
-import net.corda.core.node.services.vault.PageSpecification
 import net.corda.core.node.services.vault.QueryCriteria.VaultQueryCriteria
 import net.corda.core.node.services.vault.Sort
 import net.corda.core.node.services.vault.SortAttribute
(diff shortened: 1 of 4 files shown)


anthonykr3:
Apologies, I forgot it doesn't show clearly because I moved the file.

In the while loop in the paging example it was:
`while ((pageSpec.pageSize * (pageNumber)) <= results.totalStatesAvailable)`
but it didn't work if the number of results wasn't divisible by the page size. It's now (see [here](https://github.com/corda/corda/pull/2970/files#diff-6d0ed05cbe9b95ec07f18ef8712688daR502)):
`while ((pageSpec.pageSize * (pageNumber - 1)) <= results.totalStatesAvailable)`

I was trying to write a new test in `ScheduledFlowTests`, and all my tests were failing with strange results - which I assumed was because I'd broken something in the `NodeSchedulerService` but it was actually because of this error in the paging which was being tested in the same place. Hence why I moved it out. 

Since the code is also used as part of the documentation, I also wrote a java version of the example and wrote a couple of quick unit tests to make sure my version actually worked.

node/src/test/kotlin/net/corda/node/services/events/ScheduledFlowTests.kt
```diff
@@ -141,7 +139,7 @@ class ScheduledFlowTests {
 
     @Test
     fun `run a whole batch of scheduled flows`() {
-        val N = 100
+        val N = 99
```
josecoll: Any reason this has changed from 100 to 99? (ie. is is hiding an underlying problem)

anthonykr3: Yes it was hiding the underlying problem, the test was passing as 100 = 20 * 5. If you changed the test to 99 it failed, and I left it in even though I took the paging out, just in case the error worked its way back in.

josecoll: Understood. Makes sense.

PR #2975, opened 2018-04-18
Add exponential backoff to P2P messaging retry

thschroeter:


thschroeter:
I'll add a commit to make the retry configurable.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -108,6 +111,12 @@ data class BFTSMaRtConfiguration(
     }
 }
 
+data class P2PMessagingRetryConfiguration(
```
adagys: maybe add a kdoc emphasising that this is only used for notarisation requests

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -18,7 +18,10 @@ import net.corda.tools.shell.SSHDConfiguration
 import java.net.URL
 import java.nio.file.Path
 import java.time.Duration
+import java.time.temporal.ChronoUnit
+import java.time.temporal.TemporalUnit
 import java.util.*
+import java.util.concurrent.TimeUnit
```
adagys: are these new imports used?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -35,7 +38,7 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val compatibilityZoneURL: URL?
     val certificateChainCheckPolicies: List<CertChainPolicyConfig>
     val verifierType: VerifierType
-    val messageRedeliveryDelaySeconds: Int
+    val p2pMessagingRetryConfiguration: P2PMessagingRetryConfiguration
```
adagys: could probably shorten the val name to `p2pMessagingRetry` (and the entry name in the node.conf)

thschroeter: done

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
thschroeter: removed

PR #2978, opened 2018-04-19
CORDA-1352: Add killFlow

exFalso:
Blocked on https://github.com/corda/corda/pull/2964


exFalso:
Unblocked

mikehearn:
Is this the right approach - Kotlin coroutines has quite a sophisticated approach to cancellation. Is this an urgent need or should we take a bit more time on design work?

mikehearn:
Hm, the docs don't say much about how it works (e.g. does it erase the killed checkpoints?) but OK, if we need it to prevent API drift I guess we can clarify and extend its semantics later.

PR #2979, opened 2018-04-19
Move kotlin utility helper out of public into internals.

josecoll:
As per recommendation in https://github.com/corda/enterprise/pull/757


mikehearn:
How did this pass the API checker tool? Isn't that a backwards compatibility break? You can't just remove APIs from the core module like that.

mikehearn:
Ah, never mind, it was never shipped as public API to begin with. That's how :)

PR #2981, opened 2018-04-19
CORDA-1354 Check for jars in nested folders (to add in classpath) 

kchalkias:
Also log if directory was not created / cannot be found.

### DIFF
--- a/node/src/main/java/CordaCaplet.java
+++ b/node/src/main/java/CordaCaplet.java
@@ -77,14 +77,11 @@ public class CordaCaplet extends Capsule {
         // defined as public static final fields on the Capsule class, therefore referential equality is safe.
         if (ATTR_APP_CLASS_PATH == attr) {
             T cp = super.attribute(attr);
-            // Create cordapps directory if it doesn't exist.
+
             File cordappsDir = new File(baseDir, "cordapps");
-            cordappsDir.mkdir();
-            if (!cordappsDir.exists()) {
-                log(LOG_VERBOSE, "cordapps directory cannot be found");
-                System.exit(1);
-            }
-            // Add additional directories of JARs to the classpath (at the end). e.g. for JDBC drivers.
+            // Create cordapps directory if it doesn't exist.
+            requireCordappsDirExists(cordappsDir);
+            // Add additional directories of JARs to the classpath (at the end), e.g., for JDBC drivers.
             augmentClasspath((List<Path>) cp, cordappsDir);
             try {
                 List<String> jarDirs = nodeConfig.getStringList("jarDirs");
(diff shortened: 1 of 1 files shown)


node/src/main/java/CordaCaplet.java
```diff
@@ -80,7 +78,7 @@ protected String getVarValue(String var) {
         if (ATTR_APP_CLASS_PATH == attr) {
             T cp = super.attribute(attr);
 
-            (new File(baseDir, "cordapps")).mkdir();
+            (new File(baseDir, "cordapps")).mkdir(); // TODO: shall we check if the folder has been created?
```
Clintonio: I think it'd be worth just adding it. Something that simply tries to create the directory and then if the directory doesn't exist throws an exception. It would indicate deeper problems if a node can't, say, write a directory, given how many Corda creates. Fail fast is a good thing to have.

At a bare minimum if you aren't willing to commit to that at least emit an error message if it doesn't exist after this mkdir.

node/src/main/java/CordaCaplet.java
```diff
@@ -129,23 +127,27 @@ protected String getVarValue(String var) {
 
     private void augmentClasspath(List<Path> classpath, File dir) {
```
Clintonio: I can see a case where this ends up infinitely recursing if a user decides to do a circular symlink. Though it's not worth addressing until someone actually tries it.

kchalkias: Thanks Clinton, do you think we should parameterise `augmentClasspath` with an extra flag `checkNestedJars`, just in case we want to disable it in the future?

Clintonio: For now this is fine.

node/src/main/java/CordaCaplet.java
```diff
@@ -79,10 +77,15 @@ protected String getVarValue(String var) {
...
             T cp = super.attribute(attr);
-
-            (new File(baseDir, "cordapps")).mkdir();
-            // Add additional directories of JARs to the classpath (at the end). e.g. for JDBC drivers
-            augmentClasspath((List<Path>) cp, new File(baseDir, "cordapps"));
+            // Create cordapps directory if it doesn't exist.
+            File cordappsDir = new File(baseDir, "cordapps");
+            cordappsDir.mkdir();
+            if (!cordappsDir.exists()) {
+                log(LOG_VERBOSE, "cordapps directory cannot be found");
```
Clintonio: "Cordapps dir could not be created" I believe makes more sense because we tried to create it just before.

node/src/main/java/CordaCaplet.java
```diff
@@ -79,10 +77,15 @@ protected String getVarValue(String var) {
         // defined as public static final fields on the Capsule class, therefore referential equality is safe.
         if (ATTR_APP_CLASS_PATH == attr) {
             T cp = super.attribute(attr);
-
-            (new File(baseDir, "cordapps")).mkdir();
-            // Add additional directories of JARs to the classpath (at the end). e.g. for JDBC drivers
-            augmentClasspath((List<Path>) cp, new File(baseDir, "cordapps"));
+            // Create cordapps directory if it doesn't exist.
+            File cordappsDir = new File(baseDir, "cordapps");
+            cordappsDir.mkdir();
```
Clintonio: Would be worth anding this with the conditional below (ie; if (!cordapps.mkdir && !cordappsDir.exists())

node/src/main/java/CordaCaplet.java
```diff
@@ -79,10 +77,15 @@ protected String getVarValue(String var) {
         // defined as public static final fields on the Capsule class, therefore referential equality is safe.
         if (ATTR_APP_CLASS_PATH == attr) {
             T cp = super.attribute(attr);
-
-            (new File(baseDir, "cordapps")).mkdir();
-            // Add additional directories of JARs to the classpath (at the end). e.g. for JDBC drivers
-            augmentClasspath((List<Path>) cp, new File(baseDir, "cordapps"));
+            // Create cordapps directory if it doesn't exist.
+            File cordappsDir = new File(baseDir, "cordapps");
```
Clintonio: Could you split this into a function called "requireDirectoryExists"

node/src/main/java/CordaCaplet.java
kchalkias: I used this because we are not 100% sure the directory was not there already, are we? Maybe it existed already (thus we didn't really create it), but we don't have permission to read/access it.

PR #2986, opened 2018-04-20
CORDA-1353: Notaries should reject transactions that contain too…

adagys:
… many inputs. Otherwise, it may cause a slowdown or make the service hang.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -131,19 +131,21 @@ class NotaryFlow {
      */
     // See AbstractStateReplacementFlow.Acceptor for why it's Void?
     abstract class Service(val otherSideSession: FlowSession, val service: TrustedAuthorityNotaryService) : FlowLogic<Void?>() {
+        companion object {
+            // TODO: Determine an appropriate limit and also enforce in the network parameters and the transaction builder.
+            private const val maxAllowedInputs = 10_000
+        }
 
         @Suspendable
         override fun call(): Void? {
             check(serviceHub.myInfo.legalIdentities.any { serviceHub.networkMapCache.isNotary(it) }) {
                 "We are not a notary on the network"
             }
-
             val requestPayload = otherSideSession.receive<NotarisationPayload>().unwrap { it }
             var txId: SecureHash? = null
             try {
                 val parts = validateRequest(requestPayload)
                 txId = parts.id
-                checkNotary(parts.notary)
                 service.validateTimeWindow(parts.timestamp)
                 service.commitInputStates(parts.inputs, txId, otherSideSession.counterparty, requestPayload.requestSignature)
                 signTransactionAndSendResponse(txId)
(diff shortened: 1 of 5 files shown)


node/src/test/kotlin/net/corda/node/services/transactions/NotaryServiceTests.kt
```diff
@@ -1,38 +1,27 @@
 package net.corda.node.services.transactions
```
adagys: I moved all tests from here to ValidatingNotaryServiceTests (as they we all assuming a validating notary).
Added the input spend test to both NotaryServiceTests and ValidatingNotaryServiceTests.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -131,19 +131,21 @@ class NotaryFlow {
      */
     // See AbstractStateReplacementFlow.Acceptor for why it's Void?
     abstract class Service(val otherSideSession: FlowSession, val service: TrustedAuthorityNotaryService) : FlowLogic<Void?>() {
+        companion object {
+            // TODO: Determine an appropriate limit and also enforce in the network parameters and the transaction builder.
+            private const val maxAllowedInputs = 10_000
```
adagys: 10,000 was picked because the HA notary hangs with anything above

patrickkuo: Do we know the cause of the "hangs"? 
if it fail when loading the states, is the limit independent of the size of the contract state object? Will a bigger contract state fail with a lower limit?

adagys: It's just that we commit all states atomically, so for something like MySQL this means massive SQL statements which the db can't handle.

The actual object size is less of an issue, since notaries only deal with `StateRef`s. In one of my tests, however, transaction resolution failed when trying to resolve 10,000 inputs (stack overflow)

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -131,19 +131,21 @@ class NotaryFlow {
      */
     // See AbstractStateReplacementFlow.Acceptor for why it's Void?
     abstract class Service(val otherSideSession: FlowSession, val service: TrustedAuthorityNotaryService) : FlowLogic<Void?>() {
+        companion object {
+            // TODO: Determine an appropriate limit and also enforce in the network parameters and the transaction builder.
```
shamsasari: Is there any reason why this work can't actually introduce a new network parameter and enforce the limit via that?

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -153,6 +155,16 @@ class NotaryFlow {
             return null
         }
 
+        /** Checks whether the number of input states is too large. */
+        protected fun checkInputs(inputs: List<StateRef>) {
```
shamsasari: This becomes public API. I'd rather we just have this private/internal.

adagys: Actually, it has to be called by implementations, so let's leave it public.

adagys:
are you guys happy to approve if I raise another Jira to make the input limit part of network parameters?

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
shamsasari: Can you not make the method more generic, like `checkTransaction`, or have it pass in the tx instead of the list of inputs? The worry I have is that this becomes so specific that we don't use it anymore in some future release and then have to deprecate (something we've already done in this area of the code, btw). The KDoc definitely needs to be made more general.

adagys: see https://github.com/corda/corda/pull/2827

adagys:
Created a followup task: https://r3-cev.atlassian.net/browse/CORDA-1378

PR #2987, opened 2018-04-20
CORDA-1319 Adding CRL checking for nodes

mkit:
This PR enables the CRL checking on the node side. The node configuration is extended with a new parameter called `revocationCheckConfig`, which takes 3 boolean flags configuring the SSL-level revocation check mechanism.
 @mnesbit please let me know where else the SSL configuration change is required.

### DIFF
--- a/docs/source/corda-configuration-file.rst
+++ b/docs/source/corda-configuration-file.rst
@@ -64,15 +64,10 @@ absolute path to the node's base directory.
 
     .. note:: Longer term these keys will be managed in secure hardware devices.
 
-:revocationCheckConfig:
-
-        :softFail: This is a boolean flag that when enabled (i.e. `true` value is set) the certificate revocation list (CRL) checking will use the soft fail mode.
-                   The soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
-                   If this parameter is set to `false` the rigorous CRL checking takes place, meaning that each certificate in the
-                   certificate path being checked needs to have the CRL distribution point extension set and pointing to a URL serving a valid CRL.
-        :preferCrl: When set the CRL checking is preferred over the OCSP and OCSP is considered as a fallback mechanism unless the `noFallback` flag is set.
-                    The OCSP is not supported yet, therefore, this flag should always be set.
-        :noFallback: When set the fallback mechanism is disabled. The OCSP is not supported yet, therefore, this flag should always be set.
+:crlCheckSoftFail: This is a boolean flag that when enabled (i.e. `true` value is set) the certificate revocation list (CRL) checking will use the soft fail mode.
+                  The soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
+                  If this parameter is set to `false` the rigorous CRL checking takes place, meaning that each certificate in the
+                  certificate path being checked needs to have the CRL distribution point extension set and pointing to a URL serving a valid CRL.
 
 :database: Database configuration:
 
--- a/docs/source/corda-configuration-file.rst
+++ b/docs/source/corda-configuration-file.rst
@@ -64,6 +64,16 @@ absolute path to the node's base directory.
 
     .. note:: Longer term these keys will be managed in secure hardware devices.
 
+:revocationCheckConfig:
+
+        :softFail: This is a boolean flag that when enabled (i.e. `true` value is set) the certificate revocation list (CRL) checking will use the soft fail mode.
+                   The soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
+                   If this parameter is set to `false` the rigorous CRL checking takes place, meaning that each certificate in the
+                   certificate path being checked needs to have the CRL distribution point extension set and pointing to a URL serving a valid CRL.
+        :preferCrl: When set the CRL checking is preferred over the OCSP and OCSP is considered as a fallback mechanism unless the `noFallback` flag is set.
+                    The OCSP is not supported yet, therefore, this flag should always be set.
+        :noFallback: When set the fallback mechanism is disabled. The OCSP is not supported yet, therefore, this flag should always be set.
+
 :database: Database configuration:
 
         :serverNameTablePrefix: Prefix string to apply to all the database tables. The default is no prefix.
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPServer.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPServer.kt
@@ -14,6 +14,7 @@ import io.netty.util.internal.logging.InternalLoggerFactory
 import io.netty.util.internal.logging.Slf4JLoggerFactory
 import net.corda.core.utilities.NetworkHostAndPort
 import net.corda.core.utilities.contextLogger
+import net.corda.nodeapi.internal.config.RevocationCheckConfig
 import net.corda.nodeapi.internal.protonwrapper.messages.ReceivedMessage
 import net.corda.nodeapi.internal.protonwrapper.messages.SendableMessage
 import net.corda.nodeapi.internal.protonwrapper.messages.impl.SendableMessageImpl
@@ -40,6 +41,7 @@ class AMQPServer(val hostName: String,
                  private val keyStore: KeyStore,
                  private val keyStorePrivateKeyPassword: CharArray,
                  private val trustStore: KeyStore,
+                 private val revocationCheckConfig: RevocationCheckConfig,
                  private val trace: Boolean = false) : AutoCloseable {
 
     companion object {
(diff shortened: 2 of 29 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPServer.kt
```diff
@@ -40,6 +40,7 @@ class AMQPServer(val hostName: String,
                  private val keyStore: KeyStore,
                  private val keyStorePrivateKeyPassword: CharArray,
                  private val trustStore: KeyStore,
+                 private val crlCheckSoftFail: Boolean,
```
mnesbit: Can we pick a slightly different parameter location for this. The defaulting on trace which is also a boolean and the fact I don't use that often is likely to go wrong.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/SSLHelper.kt
```diff
@@ -36,4 +40,20 @@ internal fun createServerSslHelper(keyManagerFactory: KeyManagerFactory,
...
-}
\ No newline at end of file
+}
+
+internal fun initialiseTrustStoreAndEnableCrlChecking(trustStore: KeyStore, softFail: Boolean): ManagerFactoryParameters {
+    val certPathBuilder = CertPathBuilder.getInstance("PKIX")
+    val revocationChecker = certPathBuilder.revocationChecker as PKIXRevocationChecker
+    revocationChecker.options = EnumSet.of(
+            PKIXRevocationChecker.Option.PREFER_CRLS, // Prefer CRL over OCSP
+            PKIXRevocationChecker.Option.NO_FALLBACK) // Don't fall back to OCSP checking
```
mnesbit: We do want OCSP one day. Should this be a parameter to the function too, or perhaps not even disabled?

node/src/main/kotlin/net/corda/node/services/config/SslOptions.kt
```diff
@@ -4,9 +4,19 @@ import net.corda.nodeapi.internal.config.SSLConfiguration
 import java.nio.file.Path
 import java.nio.file.Paths
 
-data class SslOptions(override val certificatesDirectory: Path, override val keyStorePassword: String, override val trustStorePassword: String) : SSLConfiguration {
+data class SslOptions(override val certificatesDirectory: Path,
```
kchalkias: As @mnesbit already mentioned, we should probably include all revocationChecker.options as a parameter on `SslOptions`.

kchalkias: Two minor comments (not related to this PR) - just mentioning it for a TODO:
- we use both `SSL` and `Ssl` for names. We should pick one of them, or even better change to `TLS`.
- we expose an internal class to en external one (`SSLConfiguration` to `SslOptions`), I thought the guidelines is we shouldn't do that.

tools/shell/src/main/kotlin/net/corda/tools/shell/ShellConfiguration.kt
```diff
@@ -23,6 +23,10 @@ data class ShellConfiguration(
     }
 }
 
-data class ShellSslOptions(override val sslKeystore: Path, override val keyStorePassword: String, override val trustStoreFile:Path, override val trustStorePassword: String) : SSLConfiguration {
+data class ShellSslOptions(override val sslKeystore: Path,
```
kchalkias: Similar minor naming consistency issues
- `sslKeystore` -> it's a path not the keystore itself.
- `trustStoreFile` -> again a path
maybe rename both to sslKeystorePath and sslTrustStorePath (or TLS vs SSL as mentioned inn another comment).

kchalkias: Is there any way to use one class type for SSL options, why do we need a different one for shell?

docs/source/corda-configuration-file.rst
```diff
@@ -64,6 +64,11 @@ absolute path to the node's base directory.
 
     .. note:: Longer term these keys will be managed in secure hardware devices.
 
+:crlCheckSoftFail: This is a boolean flag that when enabled (i.e. `true` value is set) the certificate revocation list (CRL) checking will use the soft fail mode.
```
kchalkias: Did we agree that softFail is an option anyway? In which cases does it make sense? I mean an attacker can simply cause/force a network error to bypass CRL checking.

node/src/main/kotlin/net/corda/node/services/config/SslOptions.kt
mkit: Will put the TODO comment on the first item.
The second one, maybe @shamsasari can comment on this one.

tools/shell/src/main/kotlin/net/corda/tools/shell/ShellConfiguration.kt
mkit: TODO  will be added.

docs/source/corda-configuration-file.rst
mkit: We need a soft fail as we cannot everyone to have a CRL infrastructure. This change is in the OS.

node/src/main/kotlin/net/corda/node/services/config/SslOptions.kt
shamsasari: `SslOptions` is in the node module so internal.

mikehearn:
WRT not checking other cases, I believe that's deliberate and a part of the original revocation design.  SSL only for now.

Revoking certificates used for *transactions* has rather complex and subtle semantics e.g. do transactions where you know the identity certs for keys suddenly no longer verify? But they do if they were using random keys? What about revocations that are supposed to affect uses in the past?

It leads to quite deep questions about what revocation is for and what attacks it's intended to stop. For now, preventing someone from connecting to you with a stolen key will prevent them from engaging in most tricky business (i.e. the thief cannot steal money if they cannot contact the notaries or other counterparties at all). Going further would require more thought and design effort.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/SSLHelper.kt
```diff
@@ -36,4 +41,28 @@ internal fun createServerSslHelper(keyManagerFactory: KeyManagerFactory,
     sslEngine.enabledCipherSuites = ArtemisTcpTransport.CIPHER_SUITES.toTypedArray()
     sslEngine.enableSessionCreation = true
     return SslHandler(sslEngine)
-}
\ No newline at end of file
+}
+
+internal fun initialiseTrustStoreAndEnableCrlChecking(trustStore: KeyStore, revocationCheckConfig: RevocationCheckConfig): ManagerFactoryParameters {
```
mnesbit: If the building of a revocationChecker is standard, perhaps that should be a bit more accessible than buried deep inside the AMQP wrapper library.  Perhaps move it up inside node-api, or even put it into core near teh identity check logic?

docs/source/corda-configuration-file.rst
```diff
@@ -64,6 +64,16 @@ absolute path to the node's base directory.
 
     .. note:: Longer term these keys will be managed in secure hardware devices.
 
+:revocationCheckConfig:
+
+        :softFail: This is a boolean flag that when enabled (i.e. `true` value is set) the certificate revocation list (CRL) checking will use the soft fail mode.
+                   The soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
+                   If this parameter is set to `false` the rigorous CRL checking takes place, meaning that each certificate in the
+                   certificate path being checked needs to have the CRL distribution point extension set and pointing to a URL serving a valid CRL.
+        :preferCrl: When set the CRL checking is preferred over the OCSP and OCSP is considered as a fallback mechanism unless the `noFallback` flag is set.
```
shamsasari: It doesn't make sense to add a config option for a feature that doesn't exist. The argument that this is so that nodes don't have to be upgraded when OCSP is eventually added is weak IMO. For one, as far as I understand, OCSP is not on the near roadmap, and secondly with all the other features we're adding they may have to upgrade anyway, especially if the CZ operator enforces a min platform version.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/SSLHelper.kt
mkit: It is not standard. This code is specific to the SSL communication. For identities we will need to enable it in the bouncy castle, which has a different way of configuring it.

docs/source/corda-configuration-file.rst
mkit: Confirmed with @mnesbit and @kchalkias and we are going to remove the OCSP options and have only a boolean flag for the time being.

docs/source/corda-configuration-file.rst
```diff
@@ -64,6 +64,16 @@ absolute path to the node's base directory.
 
     .. note:: Longer term these keys will be managed in secure hardware devices.
 
+:revocationCheckConfig:
```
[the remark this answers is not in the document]
mkit: 👍

docs/source/corda-configuration-file.rst
```diff
@@ -64,6 +64,16 @@ absolute path to the node's base directory.
...
+
+        :softFail: This is a boolean flag that when enabled (i.e. `true` value is set) the certificate revocation list (CRL) checking will use the soft fail mode.
```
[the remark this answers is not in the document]
mkit: 👍

PR #2989, opened 2018-04-22
ENT-1775: reworked client to handle failover in HA mode instead of Artemis

bpaunescu:
Cherry picked. RPC level failover with round-robin through list of available nodes. 
* ENT-1775: reworked client to handle failover in HA mode instead of relying on artemis

* ENT-1775: address PR comments

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
@@ -93,18 +93,33 @@ interface CordaRPCClientConfiguration {
  * [CordaRPCClientConfiguration]. While attempting failover, current and future RPC calls will throw
  * [RPCException] and previously returned observables will call onError().
  *
+ * If the client was created using a list of hosts, automatic failover will occur(the servers have to be started in HA mode)
+ *
  * @param hostAndPort The network address to connect to.
  * @param configuration An optional configuration used to tweak client behaviour.
  * @param sslConfiguration An optional [SSLConfiguration] used to enable secure communication with the server.
+ * @param haAddressPool A list of [NetworkHostAndPort] representing the addresses of servers in HA mode.
+ * The client will attempt to connect to a live server by trying each address in the list. If the servers are not in
+ * HA mode, the client will round-robin from the beginning of the list and try all servers.
  */
 class CordaRPCClient private constructor(
-        hostAndPort: NetworkHostAndPort,
-        configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.default(),
-        sslConfiguration: SSLConfiguration? = null,
-        classLoader: ClassLoader? = null
+        private val hostAndPort: NetworkHostAndPort,
+        private val configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.default(),
+        private val sslConfiguration: SSLConfiguration? = null,
+        private val classLoader: ClassLoader? = null,
+        private val haAddressPool: List<NetworkHostAndPort> = emptyList()
 ) {
     @JvmOverloads
     constructor(hostAndPort: NetworkHostAndPort, configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.default()) : this(hostAndPort, configuration, null)
+
+    /**
+     * @param haAddressPool A list of [NetworkHostAndPort] representing the addresses of servers in HA mode.
+     * The client will attempt to connect to a live server by trying each address in the list. If the servers are not in
+     * HA mode, the client will round-robin from the beginning of the list and try all servers.
+     * @param configuration An optional configuration used to tweak client behaviour.
+     */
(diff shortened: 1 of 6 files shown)


client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -115,6 +130,14 @@ class CordaRPCClient private constructor(
             return CordaRPCClient(hostAndPort, configuration, sslConfiguration)
         }
 
+        internal fun createWithSsl(
```
mikehearn: If it's internal why do you need another overload?

client/rpc/src/main/kotlin/net/corda/client/rpc/CordaRPCClient.kt
```diff
@@ -123,6 +146,15 @@ class CordaRPCClient private constructor(
         ): CordaRPCClient {
             return CordaRPCClient(hostAndPort, configuration, sslConfiguration, classLoader)
         }
+
+        internal fun createWithSslAndClassLoader(
```
mikehearn: ditto

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -173,6 +175,8 @@ class RPCClientProxyHandler(
     private val deduplicationSequenceNumber = AtomicLong(0)
 
     private val sendingEnabled = AtomicBoolean(true)
+    // used to interrupt failover thread (i.e. client is closed while failing over)
```
mikehearn: Comment style.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -403,26 +416,82 @@ class RPCClientProxyHandler(
...
+                reconnectAttempts--
+                retryInterval = minOf(maxRetryInterval, retryInterval.times(rpcConfiguration.connectionRetryIntervalMultiplier.toLong()))
+                continue
+            }
 
-                rpcReplyMap.forEach { _, replyFuture ->
-                    replyFuture.setException(RPCException("Connection failure detected."))
-                }
+            log.debug("Connected successfully using ${transport.params}")
+            log.info("RPC server available.")
```
mikehearn: Maybe log which one was picked here and how many attempts were required.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -403,26 +416,82 @@ class RPCClientProxyHandler(
...
+            cleanUpOnConnectionLoss()
+            sessionFactory?.apply {
+                connection.destroy()
+                cleanup()
+                close()
+            }
+            haFailoverThread = Thread.currentThread()
+            attemptReconnect()
+        }
+        /* Other events are not considered as reconnection is not done by Artemis */
```
mikehearn: // style

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -192,17 +196,22 @@ class RPCClientProxyHandler(
                 rpcConfiguration.reapInterval.toMillis(),
                 TimeUnit.MILLISECONDS
         )
+        // Create a session factory using the first available server. If more than one transport configuration was
+        // used when creating the server locator, every one will be tried during failover. The locator will round-robin
+        // through the available transport configurations with the starting position being generated randomly.
+        // If there's only one available, that one will be retried continuously as configured in rpcConfiguration.
+        // There is no failover on first attempt, meaning that if a connection cannot be established, the serverLocator
+        // will try another transport if it exists or throw an exception otherwise.
```
mikehearn: That's a good description of _what_ the code does but it'd be useful to have a discussion of _why_ it does it, rather than delegate to the Artemis client - I guess that's the first question that will pop into the mind of many readers. Why do we re-implement this basic functionality ourselves?

bpaunescu: I see your point. That comment explains how the code, that's been there for some time, behaves when multiple hosts are used when creating the server locator (without HA). I guess, a detailed _why_ explanation should be added to the class description

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
bpaunescu: Successful connection is logged on line 449 specifying the host used. I'll add the number of attempts. It is debug level only.

PR #3001, opened 2018-04-25
CORDA-1001 Remove unused bits of PersistentMap

anthonykeenan:
The `PersistentMap` currently has support for a map that doesn't allow duplicates. If the map is updated with duplicates, it doesn't do anything and doesn't persist the changes. That functionality isn't used however and is confusing, so I've removed it.

I've also removed the set operator - this isn't called if the `PersistentMap `is cast to a `MutableMap`, in which case the `MutableMap`'s set operator is called which calls the put function. The set operator should should just call the put operator for consistency.

I've also added some unit tests to make sure that the functions work properly and which should flag an error if someone tries to re-implement the set operator differently in the future by testing that the assignment operators work the same for a `PersistentMap` or a `MutableMap`

### DIFF
--- a/node/src/main/kotlin/net/corda/node/utilities/PersistentMap.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/PersistentMap.kt
@@ -60,7 +60,7 @@ class PersistentMap<K : Any, V, E, out EK>(
 
     override val size get() = cache.estimatedSize().toInt()
 
-    private tailrec fun set(key: K, value: V, logWarning: Boolean = true, store: (K, V) -> V?, replace: (K, V) -> Unit): Boolean {
+    private tailrec fun set(key: K, value: V): Boolean {
         var insertionAttempt = false
         var isUnique = true
         val existingInCache = cache.get(key) {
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/utilities/PersistentMap.kt
```diff
@@ -60,81 +60,33 @@ class PersistentMap<K : Any, V, E, out EK>(
...
-                    replace = { _: K, _: V -> Unit }
-            )
-
-    /**
-     * Associates the specified value with the specified key in this map and persists it.
-     * @return true if added key was unique, otherwise false
-     */
-    private fun addWithDuplicatesReplaced(key: K, value: V) =
-            set(key, value,
-                    logWarning = false,
```
anthonykeenan: logwarning was only ever set to false as far as I could see, so i've removed it

node/src/main/kotlin/net/corda/node/utilities/PersistentMap.kt
```diff
@@ -60,81 +60,33 @@ class PersistentMap<K : Any, V, E, out EK>(
...
-                    store = { k: K, v: V -> merge(k, v) },
-                    replace = { k: K, v: V -> replaceValue(k, v) }
```
anthonykeenan: since the map only works one way now, I've removed the ability to plug in new versions of merge/replace to simplify things

PR #3004, opened 2018-04-25
ENT-1540: Make sure transactions with "expired" time windows get re-notarised correctly

adagys:
Currently the time window is checked before states are being passed to a uniqueness provider. If the time window is invalid, the transaction will be rejected even if it has already been notarised, which violates idempotency.

For this reason the time window verification was moved alongside state conflict checks – this required some interface changes.

- Added expired timewindow re-notarisation tests for all types of notaries.
- As a result, BFTNotaryService tests were taking around 3min to run, so I split the test up into `BFTSMaRtTests` and `BFTNotaryServiceTests`. The latter re-uses the same cluster for all tests, shaving down the total run-time to under a minute.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -2112,7 +2112,7 @@ public final class net.corda.core.node.services.TimeWindowChecker extends java.l
 ##
 @net.corda.core.serialization.CordaSerializable public abstract class net.corda.core.node.services.TrustedAuthorityNotaryService extends net.corda.core.node.services.NotaryService
   public <init>()
-  public final void commitInputStates(List, net.corda.core.crypto.SecureHash, net.corda.core.identity.Party, net.corda.core.flows.NotarisationRequestSignature)
+  public final void commitInputStates(List, net.corda.core.crypto.SecureHash, net.corda.core.identity.Party, net.corda.core.flows.NotarisationRequestSignature, net.corda.core.contracts.TimeWindow)
   @org.jetbrains.annotations.NotNull protected org.slf4j.Logger getLog()
   @org.jetbrains.annotations.NotNull protected net.corda.core.node.services.TimeWindowChecker getTimeWindowChecker()
   @org.jetbrains.annotations.NotNull protected abstract net.corda.core.node.services.UniquenessProvider getUniquenessProvider()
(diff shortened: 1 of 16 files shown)


node/src/integration-test/kotlin/net/corda/node/services/BFTNotaryServiceTests.kt
```diff
@@ -31,87 +31,78 @@ import net.corda.testing.common.internal.testNetworkParameters
...
-        f.getOrThrow()
-    }
-
-    @Test
-    fun `detect double spend 1 faulty`() {
-        detectDoubleSpend(1)
     }
 
     @Test
-    fun `detect double spend 2 faulty`() {
```
adagys: removed this test as the number of replicas does not affect our notary logic under test

.ci/api-current.txt
```diff
@@ -2128,7 +2128,7 @@ public static final class net.corda.core.node.services.TrustedAuthorityNotarySer
   @org.jetbrains.annotations.NotNull public final net.corda.core.node.services.UniquenessProvider$Conflict getError()
 ##
 public interface net.corda.core.node.services.UniquenessProvider
-  public abstract void commit(List, net.corda.core.crypto.SecureHash, net.corda.core.identity.Party, net.corda.core.flows.NotarisationRequestSignature)
+  public abstract void commit(List, net.corda.core.crypto.SecureHash, net.corda.core.identity.Party, net.corda.core.flows.NotarisationRequestSignature, net.corda.core.contracts.TimeWindow)
```
rick-r3: There are presumably both moving internal in the move-to-internal PR?

node/src/main/kotlin/net/corda/node/services/transactions/BFTSMaRt.kt
```diff
@@ -216,25 +219,27 @@ object BFTSMaRt {
...
             log.debug { "Attempting to commit inputs for transaction: $txId" }
-
-            val conflicts = mutableMapOf<StateRef, SecureHash>()
             services.database.transaction {
                 logRequest(txId, callerName, requestSignature)
-                states.forEach { state ->
-                    commitLog[state]?.let { conflicts[state] = it }
+                val conflictingStates = LinkedHashMap<StateRef, StateConsumptionDetails>()
+                for (state in states) {
+                    commitLog[state]?.let { conflictingStates[state] = StateConsumptionDetails(it.sha256()) }
```
rick-r3: Thought for a minute there was a problem here, but you also hash again in `isConsumedByTheSameTx`

PR #3007, opened 2018-04-26
[CORDA-1342]: The shell doesn't print the return value of a started flow - fixed.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1342


tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -275,6 +275,12 @@ object InteractiveShell {
             } catch (e: InterruptedException) {
                 // TODO: When the flow framework allows us to kill flows mid-flight, do so here.
             }
+            stateObservable.returnValue.get()?.apply {
+                when (this) {
+                    is Throwable -> output.println("Flow completed exceptionally: ${this.message}", Color.red)
```
shamsasari: The exception is already printed with the rendering of the progress tracker steps

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -275,6 +275,11 @@ object InteractiveShell {
             } catch (e: InterruptedException) {
                 // TODO: When the flow framework allows us to kill flows mid-flight, do so here.
             }
+            stateObservable.returnValue.get()?.apply {
+                if (this !is Throwable) {
```
shamsasari: This won't work either because `get()` will throw an `ExecutionException`. We need to catch it and ignore it here.

Also this won't print anything if the return value is null

sollecitom: Fixing the exception handling, even if printing out the error from the progress handler subscriber is a bit unclean.

If the flow returns null, what would you like me to do?
if I put "Flow completed with <null>" it will print that for each void flow as well, better leaving out the case in which a flow returns something nullable and null?
You get a "Done" printed anyway from the steps.

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -275,11 +275,18 @@ object InteractiveShell {
...
```
shamsasari: Now this check isn't needed.

PR #3008, opened 2018-04-26
Corda Behave: compendium of minor changes and improvements

josecoll:
- build fat behave-jar so can run scenarios from shell scripts (from TC)
- additional run script to execute basic scenarios (for TC)
- default staging path shortened to "corda" (removed deps)
- toned down logging (info -> debug)
- fixed all compiler warnings
- fixed couple of bugs in startup checking steps
- base scenarios use variables declared using Examples parameterization

### DIFF
--- a/experimental/behave/build.gradle
+++ b/experimental/behave/build.gradle
@@ -88,9 +88,9 @@ dependencies {
 
     // Scenarios / End-to-End Tests
 
-    compile "info.cukes:cucumber-java8:$cucumber_version"
-    compile "info.cukes:cucumber-junit:$cucumber_version"
-    compile "info.cukes:cucumber-picocontainer:$cucumber_version"
+    behaveCompile "info.cukes:cucumber-java8:$cucumber_version"
+    behaveCompile "info.cukes:cucumber-junit:$cucumber_version"
+    behaveCompile "info.cukes:cucumber-picocontainer:$cucumber_version"
 }
 
 compileKotlin {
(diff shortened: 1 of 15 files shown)


experimental/behave/build.gradle
```diff
@@ -29,6 +29,14 @@ repositories {
 }
 
 sourceSets {
+    behave {
```
tlil: What's the difference between `behave` and `scenario` since they're partially overlapping?

experimental/behave/src/scenario/resources/features/database/connection.feature
```diff
@@ -11,4 +11,4 @@ Feature: Database - Connection
     Examples:
       | Node-Version    | Database-Type     |
       | master          | H2                |
-#     | master          | postgreSQL        |
\ No newline at end of file
+#      | master          | postgreSQL        |
```
tlil: Just remove this line for now instead of commenting it out?

experimental/behave/build.gradle
tlil: As we talked about, let's remove the `scenario` source set + the `scenarios` Gradle task further down

experimental/behave/src/scenario/resources/features/database/connection.feature
josecoll: Will add a comment at the top to remind people that PostgreSQL is supported and requires Docker to be running for the scenario test to succeed.

experimental/behave/build.gradle
josecoll: Agreed.

tlil:
The initial thinking was to keep the Behave library decoupled from the BDD stuff, so it just makes that distinction for potential future separation. I.e., no Cucumber dependencies for the `compile` target (the underlying library)

PR #3011, opened 2018-04-26
[CORDA-1379]: Allow soft trapping of unknown config keys.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1379


node-api/src/main/kotlin/net/corda/nodeapi/internal/config/ConfigUtilities.kt
```diff
@@ -91,11 +91,11 @@ fun Config.toProperties(): Properties {
             { it.value.unwrapped().toString() })
 }
 
-private fun <T : Any> Config.getValueInternal(path: String, type: KType): T {
-    return uncheckedCast(if (type.arguments.isEmpty()) getSingleValue(path, type) else getCollectionValue(path, type))
+private fun <T : Any> Config.getValueInternal(path: String, type: KType, onUnknownKeys: ((Set<String>, logger: Logger) -> Unit) = {_, _ -> }): T {
```
szymonsztuka: Would it be better to replace the anonymous one with reference to UnknownConfigKeysPolicy.IGNORE::handle ?

sollecitom: Fixed :) actually moved the default value out and passed `UnknownConfigKeysPolicy.IGNORE::handle` from the call site.

PR #3014, opened 2018-04-26
Run filterMyKeys only when required (in NodeVaultService)

kchalkias:
`filterMyKeys` might be expensive (due to KMS cache limitations among the others), so we should either compute it lazily and only when specifically required.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
+++ b/node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
@@ -161,8 +161,7 @@ class NodeVaultService(
                     filter { (_, output) ->
                         if (statesToRecord == StatesToRecord.ONLY_RELEVANT) {
                             isRelevant(output.data, myKeys.toSet())
-                        }
-                        else {
+                        } else {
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -156,12 +155,16 @@ class NodeVaultService(
...
+            val myKeys by lazy { keyManagementService.filterMyKeys(ltx.outputs.flatMap { it.data.participants.map { it.owningKey } }) }
             val (consumedStateAndRefs, producedStates) = ltx.inputs.
                     zip(ltx.outputs).
                     filter { (_, output) ->
-                        if (statesToRecord == StatesToRecord.ONLY_RELEVANT) isRelevant(output.data, myKeys.toSet())
-                        else true
+                        if (statesToRecord == StatesToRecord.ONLY_RELEVANT) {
+                            isRelevant(output.data, myKeys.toSet())
+                        }
+                        else {
```
patrickkuo: don't need newline before `else`

Issue #2674, opened 2018-02-28
Bug: Stripping issuer from Amount<Issued<T>> does not preserve displayTokenSize


dazraf:
This was found on release version 2.0.0. 
```
class Test {
  @Test
  fun `removing the issuer from an amount should preserve the displayTokenSize`() {
    val amount = Amount(1000, BigDecimal.TEN, "Widgets")
    val partyAndRef = DUMMY_BANK_A.ref(0x01)
    val issued = Issued(partyAndRef, amount.token)
    val issuedAmount = Amount(amount.quantity, amount.displayTokenSize, issued)
    val withoutIssuer = issuedAmount.withoutIssuer()
    // ^^^ this is the problem ^^^
    // in removing the issuer, the implementation ignores the display token size
    // and parses the token which defaults to 2dp
    assertEquals(amount, withoutIssuer)
  }
}
```

mikehearn:
https://github.com/corda/corda/issues/2674

joeldudleyr3:
Thanks for reporting this. Fixed by https://github.com/corda/corda/pull/3019.

PR #3019, opened 2018-04-26
CORDA-1386: Updates withoutIssuer to not strip display token size.

joeldudleyr3:


joeldudleyr3:
https://github.com/corda/corda/issues/2674 and https://r3-cev.atlassian.net/browse/CORDA-1386

PR #3021, opened 2018-04-27
CORDA-1335: Move DistributedServiceTests into a separate package to reduce size of attachments produced.

vkolomeyko:


node/src/integration-test/kotlin/net/corda/node/services/distributed/DistributedServiceTests.kt
```diff
@@ -1,4 +1,4 @@
-package net.corda.node.services
+package net.corda.node.services.distributed
```
shamsasari: Add a comment on why the package is important.

PR #3024, opened 2018-04-27
Provides large page spec when loading state refs to make vault update.

joeldudleyr3:


node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -186,9 +186,11 @@ class NodeVaultService(
     }
 
     private fun loadStates(refs: Collection<StateRef>): Collection<StateAndRef<ContractState>> {
-        return if (refs.isNotEmpty())
-            queryBy<ContractState>(QueryCriteria.VaultQueryCriteria(stateRefs = refs.toList())).states
-        else emptySet()
+        return if (refs.isNotEmpty()) {
+            val stateRefQueryCriteria = QueryCriteria.VaultQueryCriteria(stateRefs = refs.toList())
+            val pageSpec = PageSpecification(MAX_PAGE_SIZE)
```
josecoll: This is not a good idea either as can potentially lead to JVM OOM exceptions.
We need to constrain the maximum number of allowable states in a transaction (should be tied into the Network Parameters maximum transaction size).

josecoll:
Yes - and we should publish these upper bounds in our documentation.

joeldudleyr3:
Cool. Are you capturing all this in the JIRA? Link to the JIRA in the issue once you're done and I'll close this.

josecoll:
Please raise the JIRA as you have completed the PR (and initial investigation).

PR #3025, opened 2018-04-27
[CORDA-1338]: Error with VaultQuery for entity inheriting from CommonSchemaV1.FungibleState - FIX

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1338


josecoll:
Looks good - although I think the fix will only work for a single level of inheritance, right?
eg. A (root) -> B -> C (leaf)
works for querying A.a, B.b but not C.c

sollecitom:
@chrisr3 the current solution unfortunately breaks the API. Moreover, I cannot use an `inline` function with a `reified` type, because it's incompatible with Java.

We could either:

1. Add a duplicate for each function in `QueryCriteriaUtils.Builder` with an extra parameter allowing to specify the entity class. This would be compatible but it wouldn't fix the issue, it would expose another ugly way of performing that query.

2. Create a `Builder` interface and have object `QueryCriteriaUtils.Builder` implement that, also create a new `BuilderImpl` class : `Builder` who gets an entity class through the constructor.  Given that `QueryCriteriaUtils.Builder` is public API, this would force us to duplicate rather than fix, meaning that the current `QueryCriteriaUtils.Builder` object would still manifest the same issue and should be deprecated.

3. Deprecate `Builder` and rewrite an approach from scratch (similar to point 2, but less constrained).

Thoughts? Also, @mikehearn any ideas?

sollecitom:
@josecoll I'm hardly a JPA expert but I think it would work even with a longer inheritance chain.

sollecitom:
We could also pass the entity type to the `VaultCriteria` instead.

chrisr3:
Allowing the user to specify precisely the entity `Class` is probably the only way of fixing this reliably.

sollecitom:
@chrisr3 and @josecoll another look please.

josecoll:
I don't like this approach of passing in an Entity class to the constructor.
It pollutes the API and is non-intuitive.
Also, what if there are multiple entities being referenced?

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
```diff
@@ -139,18 +139,32 @@ sealed class QueryCriteria : GenericQueryCriteria<QueryCriteria, IQueryCriteriaP
...
-     override val contractStateTypes: Set<Class<out ContractState>>? = null) : CommonQueryCriteria() {
+     override val contractStateTypes: Set<Class<out ContractState>>? = null,
+     val entityClass: Class<L> = resolveEnclosingObjectFromExpression(expression)) : CommonQueryCriteria() {
         override fun visit(parser: IQueryCriteriaParser): Collection<Predicate> {
             super.visit(parser)
             return parser.parseCriteria(this)
         }
+
+        companion object {
+
```
josecoll: drop blank line

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
```diff
@@ -139,18 +139,32 @@ sealed class QueryCriteria : GenericQueryCriteria<QueryCriteria, IQueryCriteriaP
...
+     *  [entityClass] allows to manually specify [Entity] class. Needed when referencing a field declared in a mapped supertype in the [CriteriaExpression].
      *
      * Refer to [CommercialPaper.State] for a concrete example.
      */
     data class VaultCustomQueryCriteria<L : PersistentState> @JvmOverloads constructor
     (val expression: CriteriaExpression<L, Boolean>,
      override val status: Vault.StateStatus = Vault.StateStatus.UNCONSUMED,
-     override val contractStateTypes: Set<Class<out ContractState>>? = null) : CommonQueryCriteria() {
+     override val contractStateTypes: Set<Class<out ContractState>>? = null,
+     val entityClass: Class<L> = resolveEnclosingObjectFromExpression(expression)) : CommonQueryCriteria() {
```
josecoll: indentation is incorrect.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -94,6 +96,10 @@ fun <O, R> resolveEnclosingObjectFromExpression(expression: CriteriaExpression<O
 fun <O, C> resolveEnclosingObjectFromColumn(column: Column<O, C>): Class<O> = uncheckedCast(column.declaringClass)
 fun <O, C> getColumnName(column: Column<O, C>): String = column.name
 
+inline fun <reified L : PersistentState> CriteriaExpression<L, Boolean>.toCustomCriteria(status: Vault.StateStatus = Vault.StateStatus.UNCONSUMED, contractStateTypes: Set<Class<out ContractState>>? = null): QueryCriteria.VaultCustomQueryCriteria<L> {
```
josecoll: is this a public API? what is its purpose and where should it be used?

node/src/test/java/net/corda/node/services/vault/VaultQueryJavaTests.java
```diff
@@ -97,6 +98,30 @@ public void cleanUp() throws IOException {
      * Static queryBy() tests
      */
 
+    @Test
+    public void criteriaWithFieldFromMappedSuperclass() throws NoSuchFieldException {
+
```
josecoll: drop blank line

node/src/test/java/net/corda/node/services/vault/VaultQueryJavaTests.java
```diff
@@ -97,6 +98,30 @@ public void cleanUp() throws IOException {
...
+        Field currency = getField("currency", SampleCashSchemaV2.PersistentCashState.class);
+
+        CriteriaExpression.AggregateFunctionExpression<Object, Boolean> expression = sum(quantity, singletonList(currency), Sort.Direction.ASC);
+        VaultCustomQueryCriteria<SampleCashSchemaV2.PersistentCashState> criteria = new VaultCustomQueryCriteria(expression, Vault.StateStatus.UNCONSUMED, null, SampleCashSchemaV2.PersistentCashState.class);
+
+        database.transaction(tx -> vaultService.queryBy(FungibleAsset.class, criteria));
+    }
+
+    private Field getField(String fieldName, Class<?> clazz) throws NoSuchFieldException {
+
```
josecoll: drop blank line

chrisr3:
Well, if we can't deduce it programmatically then it seems to me like we have _no choice_ but to ask the user what it should be. The only question is "How?", without breaking the API.

node/src/test/kotlin/net/corda/node/services/vault/VaultQueryTests.kt
```diff
@@ -187,6 +188,19 @@ class VaultQueryTests {
     /** Generic Query tests
     (combining both FungibleState and LinearState contract types) */
 
+    @Test
```
josecoll: Can you add a test case with multiple inheritance, and validate the solution works for querying B.b and C.c
A.a <- B.b <- C.c

sollecitom:
This approach does not break the API, it introduces an overload. I agree, it's unclean to have a version that doesn't work and one that does, but I don't think it's possible to have it otherwise without breaking the API.
Also, a criteria on more than 1 entity doesn't make sense.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
sollecitom: Not sure it is, reformatted it anyway.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
sollecitom: Removed.

docs/source/api-vault-query.rst
```diff
@@ -113,6 +113,10 @@ There are four implementations of this interface which can be chained together t
        :doc:`Persistence </api-persistence>` for mechanisms of registering custom schemas for different testing
        purposes.
 
+    .. note:: ``VaultCustomQueryCriteria`` constructor can optionally specify the JPA ``Entity`` class manually.
+       This is needed when a ``CriteriaExpression`` references a field declared in a super class annotated with ``MappedSuperclass``.
+       For more information on the issue, please refer to .. _StackOverflow: https://stackoverflow.com/questions/49511406/group-a-child-class-on-transient-parent-attributes-for-a-custom-fungiblestate-sc
```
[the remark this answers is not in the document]
josecoll: Agree - I would drop the "For more information .... bla bla bla .... SO" paragraph

node/src/test/kotlin/net/corda/node/services/vault/VaultQueryTests.kt
sollecitom: If I understand what you mean, I added one.

node/src/test/kotlin/net/corda/node/services/vault/VaultQueryTests.kt
```diff
@@ -187,6 +188,34 @@ class VaultQueryTests {
...
+    fun `criteria with field from mapped superclass of superclass`() {
+        database.transaction {
+            val expression = builder {
+                SampleCashSchemaV2.PersistentCashState::quantity.sum(
+                        groupByColumns = listOf(SampleCashSchemaV2.PersistentCashState::currency, SampleCashSchemaV2.PersistentCashState::stateRef),
+                        orderBy = Sort.Direction.ASC
+                )
+            }
+            val criteria = VaultCustomQueryCriteria.of(expression)
```
josecoll: I don't see this additional API call being an improvement over the original:

```
            val expression = builder { CashSchemaV1.PersistentCashState::currency.equal(USD.currencyCode) }
            val criteria = VaultCustomQueryCriteria(expression)
            val results = vaultService.queryBy<FungibleAsset<*>>(criteria)
```

sollecitom:
It's inline, a constructor cannot.

node/src/test/kotlin/net/corda/node/services/vault/VaultQueryTests.kt
sollecitom: It's inline (saving you to pass the class explicitly by inferring it from the reified parameter), a constructor cannot.

josecoll: Please add the equivalent Java version of these new tests.

mikehearn:
Can you give me a few days on this one please.

sollecitom:
Hey @mikehearn of course, no problem. We wait for you then.

mikehearn:
https://youtrack.jetbrains.com/issue/KT-24170

JetBrains have agreed to expose the 'container' / getOwner info as public API in Kotlin 1.3, so if it satisfies our needs here or at least is more correct than before, we can use the internal API through a simple cast and mark it as a TODO to switch to the public API when we bump to the newest Kotlin.

I think that will remove the need to change our API and result in a better developer experience. Concur?

sollecitom:
@mikehearn I have applied your idea and it works.

sollecitom:
@mikehearn as per our discussion, I have overloaded the criteria API to work with a custom field type and deprecated the old functions. Please have a look.

sollecitom:
@mikehearn the API check just reports the deprecation of the `Field.<function>` variants.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -67,8 +69,19 @@ sealed class CriteriaExpression<O, out T> {
 
 @CordaSerializable
 class Column<O, out C>(val name: String, val declaringClass: Class<*>) {
+    @Deprecated("Use the one that takes a [FieldInfo] instead.")
```
mikehearn: When we deprecate something let's explain _why_, it'll make devs more understanding. Let's say:

`@Deprecated("Please provide a FieldInfo, as this avoids bugs that can occur in inheritance scenarios")`

Note that square brackets notation is Markdown that only works in KDocs, it has no impact in deprecation messages.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -67,8 +69,19 @@ sealed class CriteriaExpression<O, out T> {
...
     constructor(field: Field) : this(field.name, field.declaringClass)
-    constructor(property: KProperty1<O, C?>) : this(property.name, property.javaGetter!!.declaringClass)
+    constructor(field: FieldInfo) : this(field.name, field.entityClass)
+    constructor(property: KProperty1<O, C?>) : this(property.name, declaringClass(property))
+
+    private companion object {
+        fun <O, C> declaringClass(property: KProperty1<O, C?>): Class<*> {
+            return when (property) {
+                is CallableReference -> ((property as CallableReference).owner as KClass<*>).javaObjectType
```
mikehearn: Can you add a comment here explaining our usage of internal API, with a // TODO that links to the YouTrack bug? It's not at all obvious to anyone reading the code here why we did this, what bug it solves, why it's OK etc. A discussion would be helpful.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -217,16 +230,23 @@ object Builder {
     fun <R : Comparable<R>> compare(operator: BinaryComparisonOperator, value: R) = ColumnPredicate.BinaryComparison(operator, value)
     fun <O, R> KProperty1<O, R?>.predicate(predicate: ColumnPredicate<R>) = CriteriaExpression.ColumnPredicateExpression(Column(this), predicate)
 
-    fun <R> Field.predicate(predicate: ColumnPredicate<R>) = CriteriaExpression.ColumnPredicateExpression(Column<Any, R>(this), predicate)
+    @Deprecated("Does not support fields from a MappedSuperclass. Use equivalent on a [FieldInfo].")
```
mikehearn: Ah, this is a better deprecation message! Let's use it everywhere (without the square brackets)

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -282,56 +328,112 @@ object Builder {
...
+    @JvmOverloads
+    fun <R> FieldInfo.max(groupByColumns: List<FieldInfo>? = null, orderBy: Sort.Direction? = null) =
             functionPredicate(ColumnPredicate.AggregateFunction<R>(AggregateFunctionType.MAX), groupByColumns?.map { Column<Any, R>(it) }, orderBy)
+
+    private fun Field.info(): FieldInfo = FieldInfo(name, declaringClass)
 }
 
 inline fun <A> builder(block: Builder.() -> A) = block(Builder)
+
+class FieldInfo(val name: String, val entityClass: Class<*>)
```
mikehearn: This is new public API so it requires a KDoc with property tags.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -282,56 +328,112 @@ object Builder {
...
+
+@Throws(NoSuchFieldException::class)
+fun getField(fieldName: String, entityClass: Class<*>): FieldInfo {
```
mikehearn: Public API so it requires a KDoc.

docs/source/changelog.rst
```diff
@@ -12,6 +12,10 @@ Unreleased
 
 * Refactor RPC Server Kryo observable serializer into it's own sub module
 
+* Derived JPA entity class for ``VaultCustomQueryCriteria`` queries from the ``CallableReference.owner`` in Kotlin.
```
mikehearn: The first sentence here is a bit hard to understand, at least for me. There doesn't seem to be a verb. Maybe:

"The Vault Criteria API has been extended to take a more precise specification of which class contains a field. This primarily impacts Java users; Kotlin users need take no action. The old methods have been deprecated but still work - the new methods avoid bugs that can occur when JPA schemas inherit from each other."

mikehearn:
The plot thickens! From JetBrains:

> I've investigated this a bit more and it looks like you should ideally be simply using the property's receiver parameter's type to get what was specified in the left-hand side of the reference:
> B::foo.instanceParameter!!.type.jvmErasure However, it looks like parameters are constructed incorrectly for declarations inherited from supertypes, in that the above code returns A  now and not B  as you'd expect. We consider this a bug in Kotlin reflection which should be fixed.
>
> Until it's fixed, I suppose you can continue using getOwner  as a workaround. For inherited declarations, it returns the most specific class where the property can be located.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
sollecitom: Fixed. I know about the brackets, but thought it would still be useful to mark something as a code concept rather than an English word.

docs/source/changelog.rst
sollecitom: The verb is "Derived", but hardly native in English. Will replace with your sentence.

mikehearn: Ah, I read derived as an adjective.

PR #3027, opened 2018-04-27
SUP-171: Observable toposort for transactions

exFalso:
This adds a utility function allowing on the fly toposort of SignedTransaction observables.

Not sure if it's worth adding to the codebase, @mikehearn ?

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/TopologicalSort.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/TopologicalSort.kt
@@ -1,0 +1,117 @@
+package net.corda.core.internal
+
+import net.corda.core.contracts.StateRef
+import net.corda.core.crypto.SecureHash
+import net.corda.core.transactions.SignedTransaction
+import rx.Observable
+
+/**
+ * Provides a way to topologically sort SignedTransactions. This means that given any two transactions T1 and T2 in the
+ * list returned by [complete] if T1 is a dependency of T2 then T1 will occur earlier than T2.
+ */
+class TopologicalSort {
+    private val forwardGraph = HashMap<SecureHash, LinkedHashSet<SignedTransaction>>()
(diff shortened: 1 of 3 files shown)


mikehearn:
Maybe not add it to Core but document it and put it in a sample project e.g. trader-demo or something, so we can point people to it but not commit to it permanently.

At any rate, please add KDocs. I'm happy for this to go into the codebase but it'd be useless post-SGX so let's not commit to it ... maybe we need a separate module for things like this.

mikehearn:
Ah, Michele just told me we can use this to reduce our memory consumption in ResolveTransactionsFlow.

Let's get it into an internal package, so it's not public API, but it can still be a part of core. KDocs would still be nice. Then we can see if it helps remove the arbitrary 300 unique hash limit Michele is in the process of introducing, and the support ticket can be resolved by the reporter copy/pasting this code into their app (no upgrade needed).

mikehearn:
Is this stalled?

exFalso:
@mikehearn I moved the code into an internal package and used it from `ResolveTransactionsFlow` (it's basically the same code). Good to go?

core/src/main/kotlin/net/corda/core/internal/TopologicalSort.kt
```diff
@@ -0,0 +1,117 @@
...
+            // Note that we use a LinkedHashSet here to make the traversal deterministic (as long as the input list is)
+            forwardGraph.getOrPut(input.txhash) { LinkedHashSet() }.add(stx)
+        }
+        transactions.add(stx)
+    }
+
+    /**
+     * Return the sorted list of signed transactions.
+     */
+    fun finalise(): List<SignedTransaction> {
```
mikehearn: finalise is a special method name in Java. Can you name this method something else?

PR #3028, opened 2018-04-27
Feature to exclude CorDapps from specific MockNodes in a MockNetwork.

roger3cev:
Little feature that allows you to exclude a list of CorDapps from specified MockNodes. I've done this because, currently, all nodes get the same CorDapps and this doesn't work under certain scenarios... In my case, I subscribe to an observable inside a CordaService. This should only be done by one node but all do it as they all get the CorDapp, this results in unusual behaviour. This PR prevents that from happening.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
@@ -284,7 +284,7 @@ open class MockNetwork(
                             forcedID: Int? = null,
                             entropyRoot: BigInteger = BigInteger.valueOf(random63BitValue()),
                             configOverrides: (NodeConfiguration) -> Any? = {},
-                            cordappPackagesToExclude: List<String>): UnstartedMockNode {
+                            cordappPackagesToExclude: List<String> = emptyList()): UnstartedMockNode {
         val parameters = MockNodeParameters(forcedID, legalName, entropyRoot, configOverrides, cordappPackagesToExclude)
         return UnstartedMockNode.create(internalMockNetwork.createUnstartedNode(InternalMockNodeParameters(parameters)))
     }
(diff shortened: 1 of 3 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -33,17 +33,23 @@ import java.nio.file.Path
...
         val entropyRoot: BigInteger = BigInteger.valueOf(random63BitValue()),
-        val configOverrides: (NodeConfiguration) -> Any? = {}) {
+        val configOverrides: (NodeConfiguration) -> Any? = {},
+        val cordappPackagesToExclude: List<String> = emptyList()) {
     fun withForcedID(forcedID: Int?): MockNodeParameters = copy(forcedID = forcedID)
     fun withLegalName(legalName: CordaX500Name?): MockNodeParameters = copy(legalName = legalName)
     fun withEntropyRoot(entropyRoot: BigInteger): MockNodeParameters = copy(entropyRoot = entropyRoot)
     fun withConfigOverrides(configOverrides: (NodeConfiguration) -> Any?): MockNodeParameters = copy(configOverrides = configOverrides)
+    fun withExcludedCordappPackages(cordappPackagesToExclude: List<String>): MockNodeParameters = copy(cordappPackagesToExclude = cordappPackagesToExclude)
+    fun copy(forcedID: Int?, legalName: CordaX500Name?, entropyRoot: BigInteger, configOverrides: (NodeConfiguration) -> Any?): MockNodeParameters {
```
joeldudleyr3: I don’t understand why this was added? Why does it not include the cordapp packages to exclude?

roger3cev: It's to maintain backwards compatibility. Kotlin gives us a `copy` for free but it doesn't include the `@JvmOverloads` annotation, so we have to add the old copy function so we are compatible with old versions.

roger3cev:
I also thought about:

> I’m also wondering whether it’s better to exclude CorDapps from individual nodes, or specify a baseline via the mock network then add CorDapps on a per node basis. The latter seems more intuitive to me (eg adding an oracle cordapp to an oracle node). Did you consider this approach? If so, what determined your choice of the former? 

However doing this would require class path scanning and loading for each node (slow) and requires a larger change. I figured that as I'm the only person who has required this change to date, most probably won't require it going forward... And if they do, they can read the docs (which I will add) that explain they can exclude Cordapps from specific nodes.

roger3cev:
Changed the approach so that now it is additive as opposed to exclusive. Should be easier for developers to grok. Also added some docs.

PR #3031, opened 2018-04-27
Flow framework test optimisation

m4ksio:
I dumped all the shared helpers to a top level in one file, not sure if that's the best approach, but subclassing it's either (we use different scopes for test classes, so not much to share)


node/src/test/kotlin/net/corda/node/services/statemachine/FlowFrameworkTests.kt
```diff
@@ -687,281 +444,765 @@ class FlowFrameworkTests {
...
-        override fun call() {
-            progressTracker.currentStep = START_STEP
-            receivedPayloads = listOf(otherPartySession.receive<String>().unwrap { it })
-            progressTracker.currentStep = RECEIVED_STEP
-            if (nonTerminating) {
-                Fiber.park()
-            }
-        }
+//    @Test
+//    fun `non-FlowException thrown on other side`() {
```
exFalso: Please remove commented out code

PR #3038, opened 2018-04-30
Section in contributors.md for code contributions.

joeldudleyr3:
Also added acetheultimate to contributors list, based on https://github.com/corda/corda/pull/3036.


gendal:
Hi Joel. My take: we don't distinguish...  a valuable contribution is a valuable contribution.   Plus: can we check that everybody listed here is in the file? https://www.corda.net/community/  Plus maybe check all our own people (including devrel and services) who have contributed?

PR #3040, opened 2018-04-30
Introducing AbstractArgsParser which removes the boilerplate of print…

shamsasari:
…ing help and exiting the process on arg cmd errors.


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -49,7 +47,7 @@ open class NodeStartup(val args: Array<String>) {
             println("Corda will now exit...")
             return false
         }
-        val (argsParser, cmdlineOptions) = parseArguments()
+        val cmdlineOptions = NodeArgsParser().parse(*args)
```
patrickkuo: `parse` won't print help, this should be `parseOrExit`?

shamsasari: Ah!

shamsasari:
Fixed

PR #3044, opened 2018-05-01
CORDA-1389: Delete the root nodes directory before initialising the nodes

shamsasari:
This is to enable multiple calls to deployNodes without doing a gradle clean.


shamsasari:
This is a mirror of https://github.com/corda/corda-gradle-plugins/pull/59

PR #3046, opened 2018-05-01
Database transaction optional ability to retry on nested SQL Exception. 

szymonsztuka:
Database transaction can be set to retry failure due to any exception with the cause or the nested cause of SQLException type.

It may happen that underlying SQL Exception is wrapped by Hibernate, allow in such case to retry but only if this was requested e.g.
`val recoverAnyNestedSQLException = true`\
`val maxTransactionRecoveryAttempts = 3`\
`database.transaction(TransactionIsolationLevel.SERIALIZABLE, maxTransactionRecoveryAttempts, recoverAnyNestedSQLException) { .... }`

The change doesn't alter the existing behaviour.

The change is required for ENT-1447 and it's part of https://github.com/corda/enterprise/pull/777.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
@@ -116,10 +116,8 @@ class CordaPersistence(
      * @param isolationLevel isolation level for the transaction.
      * @param statement to be executed in the scope of this transaction.
      */
-    fun <T> transaction(isolationLevel: TransactionIsolationLevel, statement: DatabaseTransaction.() -> T): T {
-        _contextDatabase.set(this)
-        return transaction(isolationLevel, 2, statement)
-    }
+    fun <T> transaction(isolationLevel: TransactionIsolationLevel, statement: DatabaseTransaction.() -> T): T =
+            transaction(isolationLevel, 2, false, statement)
 
     /**
      * Executes given statement in the scope of transaction with the transaction level specified at the creation time.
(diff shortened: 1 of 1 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -246,3 +255,13 @@ fun <T : Any> rx.Observable<T>.wrapWithDatabaseTransaction(db: CordaPersistence?
         }
     }
 }
+
+/** Check if any nested cause is of [SQLException] type. */
+private val Throwable.hasSQLExceptionCause: Boolean
```
andr3ej: this does too much for a val, it should be a fun

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -116,27 +116,35 @@ class CordaPersistence(
...
      * @param statement to be executed in the scope of this transaction.
      */
     fun <T> transaction(statement: DatabaseTransaction.() -> T): T = transaction(defaultIsolationLevel, statement)
 
-    private fun <T> transaction(isolationLevel: TransactionIsolationLevel, recoverableFailureTolerance: Int, statement: DatabaseTransaction.() -> T): T {
+    /**
+     * Executes given statement in the scope of transaction, with the given isolation level.
+     * @param isolationLevel isolation level for the transaction.
+     * @param recoverableFailureTolerance number of transaction commit retries for SQL while SQL exception is encountered.
+     * @param recoverAnyNestedSQLException retry transaction on any SQL Exception wrapped as a clause of [Throwable].
```
andr3ej: cause not clause

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
szymonsztuka: Changed.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
szymonsztuka: Fixed.

PR #3047, opened 2018-05-01
[CORDA-1395] [CORDA-1378]: Control the max number of transaction dependencies.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1378
https://r3-cev.atlassian.net/browse/CORDA-1395

Here's we're solving a few things:

1. Fixed transactions resolution to call `NodeVaultService` in a paging way, enabling more than 200 (page size) inputs for a given transaction.

2. Fixed a problematic closure in `ResolveTransactionsFlow` that was causing a `LinkedHashMap$LinkedKeysIterator` to be captured and serialised as part of checkpoints. This caused stack overflow for a high number of transaction dependencies (`CompositeFieldSerialiser` recurs over field "before" of each Entry in the doubly linked iterator).

3. Prepared `ResolveTransactionsFlow` for a fully paging approach based on the database. A ticket for this improvement has been created.


sollecitom:
Unfortunately it was possible to avoid buffering states in memory to publish `Vault.Update`s because the type takes a `Set` and it's part of our public API.
We might need to reconsider using eager in-memory collections over reactive types in the future.

sollecitom:
The API Stability failure derives from the generated bytecode for a `private companion object` being `@java.lang.Deprecated public static final net.corda.core.transactions.LedgerTransaction$Companion Companion` while an `internal companion object` is not deprecated, so the row is different.

In any case, it's a false positive, API compatibility is *not* broken by this change.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -48,9 +48,13 @@ data class LedgerTransaction @JvmOverloads constructor(
         if (timeWindow != null) check(notary != null) { "Transactions with time-windows must be notarised" }
         checkNoNotaryChange()
         checkEncumbrancesValid()
+        check(inputs.size <= maxInputsCount) { "Transaction with ${inputs.size} inputs exceeded maximum inputs count of $maxInputsCount." }
     }
 
-    private companion object {
+    internal companion object {
+        // We might decide in the future to move this as part of the [NetworkParameters].
```
josecoll: Shouldn't we group all MAX limits in a single location?
They also need to be clearly documented to the end user.
@anvabr mentioned this at this mornings bug triage meeting.
(ideally at the Network Parameters level as you mention in this comment).

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -188,9 +188,18 @@ class NodeVaultService(
...
-        else emptySet()
+        val states = mutableListOf<StateAndRef<ContractState>>()
+        if (refs.isNotEmpty()) {
+            val refsList = refs.toList()
+            val pageSize = PageSpecification().pageSize
+            (0..refsList.size / pageSize).forEach {
+                val offset = it * pageSize
+                val limit = minOf(offset + pageSize, refsList.size)
+                val page = queryBy<ContractState>(QueryCriteria.VaultQueryCriteria(stateRefs = refsList.subList(offset, limit))).states
+                states.addAll(page)
```
adagys: since we're keeping everything in memory anyway, maybe it's simpler to use a single large page?

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
sollecitom: We discussed this with the network services team and they advised to avoid adding to the NetworkParameters just in case in a near future we won't need this limit anymore. We can always move it there later on if we decide to.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -48,9 +48,13 @@ data class LedgerTransaction @JvmOverloads constructor(
...
+        internal const val maxInputsCount = 2000
```
adagys: better to move it to BaseTransaction and the check to `checkBaseInvariants()` so it's fired when deserializing a `WireTransaction` as well

adagys: not directly related to this PR, but while you're at it could you add the `checkBaseInvariants()` check to `NotaryChangeWireTransaction` and ContractUpgradeWireTransaction` as well

docs/source/changelog.rst
```diff
@@ -6,6 +6,9 @@ release, see :doc:`upgrade-notes`.
 
 Unreleased
 ==========
+* Enforced a maximum of 2000 inputs for a LedgerTransaction.
```
adagys: can you expand a bit why this was needed

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
josecoll: Yes - fair comment.
I just think it would be useful (outside of this PR) to collate and document all the Corda "hard" limits in one place.

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
sollecitom: I disagree, asking the database to return a large number of rows is quite dangerous.

docs/source/api-transactions.rst
```diff
@@ -96,6 +96,8 @@ Since a transaction's output states do not exist until the transaction is commit
 outputs of previous transactions. Instead, we create the desired output states as ``ContractState`` instances, and
 add them to the transaction directly:
 
+.. note:: Corda supports a maximum of 2000 inputs for any given transaction.
```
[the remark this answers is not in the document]
sollecitom: Haha, my bad. Never meant for it to be here!

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -188,9 +188,18 @@ class NodeVaultService(
...
 
     private fun loadStates(refs: Collection<StateRef>): Collection<StateAndRef<ContractState>> {
-        return if (refs.isNotEmpty())
-            queryBy<ContractState>(QueryCriteria.VaultQueryCriteria(stateRefs = refs.toList())).states
-        else emptySet()
+        val states = mutableListOf<StateAndRef<ContractState>>()
+        if (refs.isNotEmpty()) {
+            val refsList = refs.toList()
+            val pageSize = PageSpecification().pageSize
+            (0..refsList.size / pageSize).forEach {
```
kchalkias: If the `pageSize` is a divisor of `refsList.size`, then one extra page is created, i.e, a list of 100 `StateRef` with `pageSize = 10`, will result to 11 pages (one extra).

sollecitom: Good catch :)

core/src/main/kotlin/net/corda/core/internal/FetchDataFlow.kt
```diff
@@ -85,6 +85,7 @@ sealed class FetchDataFlow<T : NamedByHash, in W : Any>(
             for (hash in toFetch) {
                 // We skip the validation here (with unwrap { it }) because we will do it below in validateFetchResponse.
                 // The only thing checked is the object type. It is a protocol violation to send results out of order.
+                // TODO we need to page here after large messages will work
```
kchalkias: period at the end of TODO comment.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -132,13 +132,14 @@ class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
...
             // graph we're traversing.
-            val notAlreadyFetched = nextRequests.filterNot { it in resultQ }.toSet()
+            val notAlreadyFetched = nextRequests - resultQ.keys
             nextRequests.clear()
 
             if (notAlreadyFetched.isEmpty())   // Done early.
                 break
 
             // Request the standalone transaction data (which may refer to things we don't yet have).
+            // TODO sollecitom maybe page here
```
kchalkias: I guess this will change in an upcoming commit to this PR?

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
```diff
@@ -13,6 +13,11 @@ import java.util.function.Predicate
  */
 @DoNotImplement
 abstract class BaseTransaction : NamedByHash {
+
+    companion object {
+        const val maxInputsCount = 300
```
kchalkias: add kdoc, so we know why 300.

docs/source/changelog.rst
```diff
@@ -6,6 +6,9 @@ release, see :doc:`upgrade-notes`.
 
 Unreleased
 ==========
+* Enforced a maximum of 2000 inputs for a LedgerTransaction. Needed because Corda loads transaction inputs in memory and because the ``Notary`` cannot cope with too many states in input for a transaction.
```
kchalkias: Don't forget to align this (2000) with the number we'll eventually pick.

adagys:
Hmm, strictly speaking the limit here should be 300 transaction _dependencies_ rather than inputs. If a transaction is consuming 1000 inputs and they're all issued in the same transaction, there won't be a problem. So maybe we should change the logic to check for max number of unique transaction ids in the inputs.
This doesn't apply to notaries, but notaries can handle much larger amount of inputs.

core/src/main/kotlin/net/corda/core/internal/FetchDataFlow.kt
sollecitom: I'll fix it but I don't see how this help.

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
sollecitom: Fixed.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -139,7 +153,7 @@ class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
                 break
 
             // Request the standalone transaction data (which may refer to things we don't yet have).
-            // TODO sollecitom maybe page here
+            // TODO maybe page here
```
josecoll: what does this mean? We either page or we don't .... but a TODO comment seems indecisive and unclear.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -35,6 +39,8 @@ class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
     companion object {
         private fun dependencyIDs(stx: SignedTransaction) = stx.inputs.map { it.txhash }.toSet()
 
+        private const val RESOLUTION_PAGE_SIZE = 100
```
josecoll: how do you come to this size?

docs/source/changelog.rst
sollecitom: DIdn't forget :)

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
sollecitom: We don't yet, because we need to see if the fix we have in mind will work. This is a remainder not to assume in-memory collections to work, and replace them with paging everywhere.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
sollecitom: 100 works and it's pretty safe. Sounds like a good compromise between too many calls and a huge single payload. We can always change it later on if needed.

sollecitom:
@josecoll well, the evidence is that test "`can handle max allowed inputs count for a transaction`()" in `ValidatingNotaryServiceTests` fails with 1000. Unfortunately the number does not guarantee much, because the issue depends on the allocated heap size and on the weight of transaction dependencies, but we hope to address this either as part of this PR or soon after with @exFalso .

sollecitom:
@adagys , @kchalkias  and @josecoll , comments addressed. Still not ready because we have to decide with @exFalso when to plug in the paging topological sort implementation.

josecoll:
Can you change the title of this PR and provide a brief description of what is being solved (as the original work was specifically addressing what seemed a minor issue).

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -99,10 +111,12 @@ class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
...
         }
+    }
 
-        return signedTransaction?.let {
-            result + it
-        } ?: result
+    private fun page(pageNumber: Int, pageSize: Int): Set<SecureHash> {
+        val offset = pageNumber * pageSize
+        val limit = minOf(offset + pageSize, txHashes.size)
+        return txHashes.subList(offset, limit).toSet()
```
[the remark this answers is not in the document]
sollecitom: I need to call `toSet()` or `toList()`, because the sub-list is not checkpoint-friendly.

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -12,8 +12,24 @@ import net.corda.core.serialization.CordaSerializable
  */
 @CordaSerializable
 abstract class CoreTransaction : BaseTransaction() {
+    companion object {
+        /**
+         * A conservative limit until we get rid of issues with keeping collections of [SignedTransaction]s in memory.
```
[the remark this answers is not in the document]
sollecitom: As discussed, I'll try and see whether we can remove it altogether after the @exFalso merges in the paging topological sort.

Why wouldn't it be backwards compatible anyway? We couldn't do 201 before this PR, now we can do 300, and we could increase it in the future anyway.

Happy to move that to NetworkParameters in case I cannot remove it because, as discussed, this number is the minimum supported by all the notaries in the network.

gendal:
Re the Raspberry Pi 200Mb comment...  wow... if we're really still setting Xmx so low and we're doing it solely out of some folklore that I want to be able to run Corda on my ancient Raspberry Pi, can we just agree to bump this to something sensible _immediately_?! :)

sollecitom:
@tudor-malene, @adagys  and @exFalso : after a discussion with @mikehearn, we decided to:

1. Migrate the limit on transaction dependencies into the NetworkParameters.
2. Create a JIRA issue to prevent OutOfMemoryErrors wrt the heap. The ticket will contain the proposed solution we agreed on, along with some other ideas e.g., a daemon thread that periodically checks whether there's garbage to clean from a crashed flow etc.

exFalso:
For the record: we discussed the streaming toposort with @sollecitom and decided that it's *not* the best solution for transaction resolution. Instead there is a way to do the resolution more efficiently by adding a "provisional receive" step.

We don't need to traverse the provisional batch in a topological manner, all we need to make sure is that all transactions in the batch have been verified at the end of processing. For this we can make the assumption that transactions in the provisional batch will *eventually* be verified, so we can make the assumption that "a provisional tx's inputs will either be verified at the end, OR the whole batch fails"

The above can be done quite efficiently in a memory bounded streaming manner. Toposort on the other hand always imposes an unbounded buffer requirement

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -20,8 +21,11 @@ import java.util.*
  *
  * @return a list of verified [SignedTransaction] objects, in a depth-first order.
  */
-class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
-                              private val otherSide: FlowSession) : FlowLogic<List<SignedTransaction>>() {
+class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
+                              private val otherSide: FlowSession) : FlowLogic<Unit>() {
+
+    private val txHashes = txHashesArg.toList()
```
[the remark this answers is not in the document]
sollecitom: Added. It's because we need to checkpoint it in a way that guarantees order.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -83,10 +89,16 @@ class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
...
+    override fun call() {
+        val newTxns = ArrayList<SignedTransaction>(txHashes.size)
         // Start fetching data.
-        val newTxns = downloadDependencies(txHashes)
-        fetchMissingAttachments(signedTransaction?.let { newTxns + it } ?: newTxns)
+        for (pageNumber in 0..(txHashes.size - 1) / RESOLUTION_PAGE_SIZE) {
+            val page = page(pageNumber, RESOLUTION_PAGE_SIZE)
+
+            newTxns += downloadDependencies(page)
+            val txsWithMissingAttachments = if (pageNumber == 0) signedTransaction?.let { newTxns + it } ?: newTxns else newTxns
```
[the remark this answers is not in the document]
sollecitom: This is the same logic we had before. Here we add the signedTransaction to the first page.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -99,13 +111,16 @@ class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
...
             serviceHub.recordTransactions(StatesToRecord.NONE, listOf(it))
         }
+    }
 
-        return signedTransaction?.let {
-            result + it
-        } ?: result
+    private fun page(pageNumber: Int, pageSize: Int): Set<SecureHash> {
+        val offset = pageNumber * pageSize
+        val limit = minOf(offset + pageSize, txHashes.size)
```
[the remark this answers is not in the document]
sollecitom: `min` works as well. Changed to use that, even if unsure about why it's better.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -132,13 +147,14 @@ class ResolveTransactionsFlow(private val txHashes: Set<SecureHash>,
         while (nextRequests.isNotEmpty()) {
             // Don't re-download the same tx when we haven't verified it yet but it's referenced multiple times in the
             // graph we're traversing.
-            val notAlreadyFetched = nextRequests.filterNot { it in resultQ }.toSet()
+            val notAlreadyFetched = nextRequests - resultQ.keys
```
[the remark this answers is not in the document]
sollecitom: `nextRequests` and `resultQ.keys` are both `Set`s. Specified type of `notAlreadyFetched` explicitly.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -41,6 +41,7 @@ data class ContractUpgradeWireTransaction(
 
     init {
         check(inputs.isNotEmpty()) { "A contract upgrade transaction must have inputs" }
+        checkBaseInvariants()
```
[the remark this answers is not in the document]
sollecitom: Was asked by @adagys to add `checkBaseInvariants()` to these Transaction types as part of this story.

mikehearn:
I'm not sure I fully grok that. Verifying a transaction requires its first-level dependencies to be available so the inputs can be resolved. So you have to fetch them first before a tx can be verified. You can potentially do that trick if you move one level at a time backwards.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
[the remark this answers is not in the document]
mikehearn: Ah, I asked Andrius, it's a bugfix. Should be a separate PR really as something touching core tx logic should ideally be reviewed and studied separately but I'll pass on it this time so we can get this PR in.

PR #3050, opened 2018-05-01
CORDA-1416: Upgrade version of Proton-J library

vkolomeyko:


vkolomeyko:
Hello @fenryka - to @mnesbit point, is there any *easy* way to check earlier on whether we are still binary compatible after upgrade of version for Proton-J?

fenryka:
Not really, what we need is really just blobs serialised from older versions of the code archived somewhere and tests that verify we can deserialise them

vkolomeyko:
Thanks @fenryka 

I have downloaded `Transaction` blob from HA Testing environment and created a sample integration test proving that it can be read by upgraded code.
Also, I verify that serialised representation created with new code match byte-to-byte to previous version.

@mnesbit and @fenryka - could you review my latest changes and let me know if you would like anything more to be done for this upgrade.

PR #3060, opened 2018-05-02
Reference inputs states

roger3cev:
Still to do:
* Consider the impact on encumbrances.
* Add custom notary exception for "out of data reference state".
* Add minimum platform version checks.
* Update changelog and release notes

To add in a subsequent PR:
* Flows to syndicate reference data.
* Looping retry flow to handle update/use races.


core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
```diff
@@ -22,6 +22,9 @@ abstract class TraversableTransaction(open val componentGroups: List<ComponentGr
     /** Pointers to the input states on the ledger, identified by (tx identity hash, output index). */
     override val inputs: List<StateRef> = deserialiseComponentGroup(ComponentGroupEnum.INPUTS_GROUP, { SerializedBytes<StateRef>(it).deserialize() })
 
+    /** Pointers to reference states, identified by (tx identity hash, output index). */
+    override val references: List<StateRef> = deserialiseComponentGroup(ComponentGroupEnum.REFERENCES_GROUP, { SerializedBytes<StateRef>(it).deserialize() })
```
adagys: how does one distinguish reference and normal inputs when building a filtered transaction?

core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
```diff
@@ -63,11 +63,23 @@ sealed class NotaryError {
...
          */
-        val hashOfTransactionId: SecureHash
-)
\ No newline at end of file
+        val hashOfTransactionId: SecureHash,
+
+        /**
+         * The type of consumed state: Either an reference input state or a regular input state.
+         */
+        val type: ConsumedStateType = ConsumedStateType.INPUT_STATE
```
adagys: why does the notary need to distinguish reference and normal inputs internally? if there's a conflict we can check if that state was in refs or inputs

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -158,9 +158,16 @@ data class StateRef(val txhash: SecureHash, val index: Int) {
...
-data class StateAndRef<out T : ContractState>(val state: TransactionState<T>, val ref: StateRef)
+data class StateAndRef<out T : ContractState>(val state: TransactionState<T>, val ref: StateRef) {
+    /** For adding [StateAndRef]s as references to a [TransactionBuilder]. */
+    fun referenced() = ReferencedStateAndRef(this)
+}
 // DOCEND 7
 
+/** A wrapper for a [StateAndRef] indicating that it should be added to a transaction as a reference input state. */
+@CordaSerializable
+data class ReferencedStateAndRef<out T : ContractState>(val stateAndRef: StateAndRef<T>)
```
adagys: maybe instead of this we should deprecate the `TransactionBuilder.withItems()`, and make people use specific adders

PR #3061, opened 2018-05-02
CORDA-1414: node should continue with parameters from file when network map not available

kasiastreich:
If network map is not available or returns exception on network
parameters reading (either serialization, connection, signature etc.), node should continue with parameters from file.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
@@ -29,7 +29,7 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
         val advertisedParametersHash = try {
             networkMapClient?.getNetworkMap()?.payload?.networkParameterHash
         } catch (e: Exception) {
-            logger.info("Couldn't connect to NetworkMap", e)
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -29,7 +28,7 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
     private fun retrieveNetworkParameters(): NetworkParameters {
         val advertisedParametersHash = try {
             networkMapClient?.getNetworkMap()?.payload?.networkParameterHash
-        } catch (e: ConnectException) {
+        } catch (e: Exception) {
```
shamsasari: Use IOException instead

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -29,7 +28,7 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
             logger.info("Couldn't connect to NetworkMap", e)
```
shamsasari: "Unable to download network map"

PR #3062, opened 2018-05-02
CORDA-847 - Pass serialization context down serialization call stack

fenryka:
Needed for later work on AMQP RPC where we need to set per-thread
elements on a context. Could use some magic thread local but I'd rather
it was explicit on the stack and thus easier to reason about.

Additionally, now we're passing this around we can make better use of it
in the future


node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/engine/ConnectionStateMachine.kt
```diff
@@ -351,7 +351,7 @@ internal class ConnectionStateMachine(serverMode: Boolean,
                 val connection = event.connection
                 val channel = connection?.context as? Channel
                 if (channel != null) {
-                    val appProperties = HashMap(amqpMessage.applicationProperties.value)
+                    val appProperties = HashMap(amqpMessage.applicationProperties.value as Map<String, Any?>)
```
mnesbit: This may interact with Viktor's PR

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
```diff
@@ -483,7 +483,7 @@ internal fun Type.isSubClassOf(type: Type): Boolean {
     return TypeToken.of(this).isSubtypeOf(TypeToken.of(type).rawType)
 }
 
-// ByteArrays, primitives and boxed primitives are not stored in the object history
+// ByteArrays, primtives and boxed primitives are not stored in the object history
```
mnesbit: primitives

node-api/src/test/kotlin/net/corda/nodeapi/internal/serialization/amqp/testutils/TestSerializationContext.kt
```diff
@@ -0,0 +1,17 @@
+package net.corda.nodeapi.internal.serialization.amqp.testutils
+
+import net.corda.core.serialization.SerializationContext
+import net.corda.nodeapi.internal.serialization.AllWhitelist
+import net.corda.nodeapi.internal.serialization.SerializationContextImpl
+import net.corda.nodeapi.internal.serialization.amqp.amqpMagic
+
+val serializationProperties: MutableMap<Any, Any> = mutableMapOf()
+
+val testSerializationContext = SerializationContextImpl(
```
mnesbit: How does this differ from the context used in 'normal' tests? Do we need any additional work for such tests?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializationHelper.kt
fenryka: done

fenryka:
Comments addressed @mnesbit

PR #3067, opened 2018-05-03
Register crypto providers after we init logging

cburlinchon:
New version of netty(4.1.15) clashes with our logging. We use FastThreadLocal which ends up initialising slf4j before we can set defaultLogLevel property.


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -55,6 +56,10 @@ open class NodeStartup(val args: Array<String>) {
 
         initLogging(cmdlineOptions)
 
+        // Register all cryptography [Provider]s first thing on boot.
```
kchalkias: remove “first thing on boot”.
Maybe add this comment to initLogging and explain why (i.e., the netty issue).

PR #3070, opened 2018-05-03
CORDA-1030 - allow nodes to be waited for in-process

shamsasari:
(cherry picked from commit 53bb986)


shamsasari:
Port of https://github.com/corda/corda/pull/2678

PR #3071, opened 2018-05-03
 CORDA-1363 Network registration helper should check public key in certificate before storing in keystore

patrickkuo:


node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -114,6 +115,11 @@ open class NetworkRegistrationHelper(private val config: SSLConfiguration,
             throw CertificateRequestException("Received certificate contains invalid cert role, expected '$certRole', got '$nodeCaCertRole'.")
         }
 
+        // Validate returned certificate is for the correct public key
+        if (Crypto.toSupportedPublicKey(certificates.first().publicKey) != Crypto.toSupportedPublicKey(registeringPublicKey)) {
```
kchalkias: It would be nice to have a similar check after `onSuccess`, to ensure we can read the keys/certs from our keystores.

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -114,6 +115,11 @@ open class NetworkRegistrationHelper(private val config: SSLConfiguration,
             throw CertificateRequestException("Received certificate contains invalid cert role, expected '$certRole', got '$nodeCaCertRole'.")
         }
 
+        // Validate returned certificate is for the correct public key
```
kchalkias: period at the end

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -85,14 +86,14 @@ open class NetworkRegistrationHelper(private val config: SSLConfiguration,
             requestIdStore.deleteIfExists()
             throw certificateRequestException
         }
-        validateCertificates(certificates)
+        validateCertificates(keyPair.public, certificates)
         storePrivateKeyWithCertificates(nodeKeyStore, keyPair, certificates, keyAlias)
         onSuccess(keyPair, certificates)
         // All done, clean up temp files.
         requestIdStore.deleteIfExists()
```
kchalkias: It would also be nice to print (maybe as another PR) that users are advised to backup the certificates folder. We've seen people deleting this folder, as it's not always obvious it contains public certs + keystores (i.e., private keys).

patrickkuo: i will add it to this PR

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
patrickkuo: I am assuming the keystore util will store the certs and key correctly, and we have test to assure that,reading it again from keystore to double check is not necessary?

patrickkuo:
@kchalkias are you happy with this?

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -85,14 +86,17 @@ open class NetworkRegistrationHelper(private val config: SSLConfiguration,
...
+
+        println("Successfully registered Corda node with compatibility zone, node identity certificates are stored in '${config.certificatesDirectory}', it is advised to backup the certificates folder.")
```
kchalkias: Maybe mention “keys” ... along with “certs”. Just to make sure clients understand that the certs folder includes the private keys as well.

PR #3073, opened 2018-05-03
[ENT-1825] Document details of cipher suites

kchalkias:
As requested, a page enlisting our supported cipher suites is added to the documentation.
At the moment, I've added as a submenu of "Other" menu. We could also create a menu called "References".
Note: I didn't mention Composite Keys yet as an option, because although one could theoretically use a Composite Key, we still don't have helper functions to support this during node registration.

### DIFF
--- a/docs/source/cipher-suites.rst
+++ b/docs/source/cipher-suites.rst
@@ -34,7 +34,8 @@ A Corda network has 8 types of keys and a regular node requires 4 of them:
 * The **tls** key(s) (per node)
 * The **confidential identity** key(s) (per node)
 
-We can visualise the certificate structure as follows:
+We can visualise the certificate structure as follows (for a detailed description of cert-hierarchy,
+see :doc:`permissioning`):
 
 .. image:: resources/certificate_structure.png
    :scale: 55%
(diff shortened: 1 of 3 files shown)


docs/source/cipher-suites.rst
```diff
@@ -0,0 +1,99 @@
+Cipher Suites
```
joeldudleyr3: I'd suggest changing to "Supported cipher suites", so people know when they'd want to access this page.

docs/source/cipher-suites.rst
```diff
@@ -0,0 +1,99 @@
...
+=============
+
+.. contents::
+
+The set of signature schemes supported forms a part of the consensus rules for a CORDA DLT network.
```
joeldudleyr3: Lower-case "Corda"

docs/source/cipher-suites.rst
```diff
@@ -0,0 +1,99 @@
...
+Corda has been designed to be cryptographically agile, in the sense that the available set of signature schemes is
+carefully selected based on various factors, such as provided security-level and cryptographic strength, compatibility
+with various HSM vendors, algorithm standardisation, variety of cryptographic primitives, business demand, option for
+post-quantum resistance, side channel security, efficiency and rigorous testing.
+
+Before we present the pool of supported schemes it is useful to be familiar with :doc:`key-concepts-identity`,
+:doc:`permissioning` and :doc:`api-identity`. An important design decision in Corda is its shared hierarchy
+between the TLS and Node Identity certificates.
+
+Certificate hierarchy
```
joeldudleyr3: Does this duplicate information in "Network permissioning"? Can you incorporate any additional information given here into that page, and just link to that page if necessary?

docs/source/cipher-suites.rst
```diff
@@ -0,0 +1,99 @@
...
+* The **tls** key(s) (per node)
+* The **confidential identity** key(s) (per node)
+
+We can visualise the certificate structure as follows:
+
+.. image:: resources/certificate_structure.png
+   :scale: 55%
+   :align: center
+
+Available cipher suites
```
joeldudleyr3: "Available" -> "Supported"

docs/source/other-index.rst
```diff
@@ -6,4 +6,5 @@ Other
 
    corda-repo-layout
    building-the-docs
-   json
\ No newline at end of file
+   json
+   cipher-suites
```
joeldudleyr3: This section is a graveyard. Can you add it to the "Corda network" section instead?

docs/source/cipher-suites.rst
kchalkias: The permissioning page does not really explain the 8 key types, thus I decided to include it here.

docs/source/other-index.rst
kchalkias: @anvabr suggested a new menu called references (or technical references). I remember you've removed `Glossary` the other day, so yeah we should decide what is the best place for this one.

docs/source/cipher-suites.rst
kchalkias: I can change this to Certificate/Key hierarchy or explain why it's here.

PR #3075, opened 2018-05-04
Add node scheduler db tests inside use block

anthonykeenan:
To stop leaking of resources


chrisr3:
I'm not sure that this was the cause of the intermittent failure I saw, but it's good to fix anyway.

PR #3077, opened 2018-05-04
Define a central gradle variable to capture the proton-j version

mnesbit:
As this is likely to change moderately often and be re-used in other projects I need a gradle variable


vkolomeyko:
Out of curiosity, why not to merge that?

mnesbit:
Chris R already made this change for his own reasons

PR #3080, opened 2018-05-04
ENT-1463: Prepare node-api for determination.

chrisr3:
Refactor some of the serialisation code so that we can remove non-deterministic elements.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ buildscript {
     // TODO: Upgrade gradle-capsule-plugin to a version with capsule:1.0.3
     ext.capsule_version = '1.0.1'
 
-    ext.asm_version = '0.5.3'
+    ext.asm_version = '5.0.4'
 
     /*
      * TODO Upgrade to version 2.4 for large message streaming support
--- a/build.gradle
+++ b/build.gradle
@@ -77,6 +77,9 @@ buildscript {
     ext.eaagentloader_version = '1.0.3'
     ext.jsch_version = '0.1.54'
     ext.commons_cli_version = '1.4'
+    ext.protonj_version = '0.27.1'
+    ext.snappy_version = '0.4'
+    ext.fast_classpath_scanner_version = '2.12.3'
 
     // Update 121 is required for ObjectInputFilter and at time of writing 131 was latest:
     ext.java8_minUpdateVersion = '131'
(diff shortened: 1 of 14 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializerFactory.kt
```diff
@@ -247,7 +246,7 @@ open class SerializerFactory(
                 // preserve the actual message locally
                 loggerFor<SerializerFactory>().apply {
                     error("${e.message} [hint: enable trace debugging for the stack trace]")
-                    trace(e.getStackTraceAsString())
+                    trace("", e)
```
exFalso: is this correct? @fenryka

build.gradle
```diff
@@ -21,7 +21,7 @@ buildscript {
     // TODO: Upgrade gradle-capsule-plugin to a version with capsule:1.0.3
     ext.capsule_version = '1.0.1'
 
-    ext.asm_version = '0.5.3'
```
fenryka: That's a hell of a jump

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializerFactory.kt
fenryka: If it pretty prints the exception into the log it's fine - @chrisr3  is that the intnetnio?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/carpenter/ClassCarpenter.kt
```diff
@@ -155,7 +158,7 @@ class ClassCarpenter(cl: ClassLoader = Thread.currentThread().contextClassLoader
```
fenryka: remove the jvm extension func on string as it's been replaced

build.gradle
chrisr3: This version number was never used by Corda, but it is _now_. The version number itself hasn't really changed at all - Kryo was always implicitly using ASM 5.0.4.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/carpenter/ClassCarpenter.kt
chrisr3: It has _one_ remaining user; but perhaps I can get rid of that too?

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializerFactory.kt
chrisr3: Yes. And removing the `getStackTraceAsString()` reference from `node-api-deterministic`.

PR #3086, opened 2018-05-08
ENT-1463: Declare default interface methods as Java8 default.

chrisr3:
Kotlin will implement default interface methods via a synthetic nested `$DefaultImpls` class unless we apply Kotlin 1.2.40's new `@JvmDefault` annotation. This annotation instructs the compiler to implement them as Java8 `default` methods instead on the interface class itself.

Apply this annotation to interfaces required by the `-deterministic` modules so that no `$DefaultImpls` classes are generated. Otherwise we would need to configure ProGuard explicitly to keep them all.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -173,6 +173,7 @@ allprojects {
             apiVersion = "1.2"
             jvmTarget = "1.8"
             javaParameters = true   // Useful for reflection.
+            freeCompilerArgs = ['-Xenable-jvm-default']
         }
     }
 
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/OrdinalIO.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/OrdinalIO.kt
@@ -8,9 +8,9 @@ import java.nio.ByteBuffer
 class OrdinalBits(private val ordinal: Int) {
     interface OrdinalWriter {
         val bits: OrdinalBits
-        val encodedSize get() = 1
-        fun writeTo(stream: OutputStream) = stream.write(bits.ordinal)
-        fun putTo(buffer: ByteBuffer) = buffer.put(bits.ordinal.toByte())!!
+        @JvmDefault val encodedSize get() = 1
+        @JvmDefault fun writeTo(stream: OutputStream) = stream.write(bits.ordinal)
+        @JvmDefault fun putTo(buffer: ByteBuffer) = buffer.put(bits.ordinal.toByte())!!
     }
 
     init {
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializer.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializer.kt
@@ -31,6 +31,7 @@ interface AMQPSerializer<out T> {
     /**
      * Write the given object, with declared type, to the output.
      */
+    @JvmDefault
     fun writeObject(obj: Any, data: Data, type: Type, output: SerializationOutput,
                     context: SerializationContext, debugIndent: Int = 0)
 


adagys:
isn't this technically an API breaking change?

chrisr3:
Actually, no. The synthetic class was never part of the API definition, although the API Scanner will detect the new `@JvmDefault` annotation. This just changes were Kotlin stuffs the default method implementation.

mikehearn:
Wait, I think Andrius may have a point.

Existing compiled Kotlin apps will have references to $DefaultImpls baked into them, won't they? That is, apps that worked before will stop linking, because the $DefaultImpls classes will be gone.

Is that not the case here? I'm not convinced we can use the new annotation without finding a way to preserve the old classes as forwarders.

chrisr3:
Are we guaranteeing _binary_ compatibility, though? I thought we were just promising that people could _recompile_ their apps without making any changes.

mikehearn:
We hedged a bit but so far we've provided binary compatibility, as far as I know (the API checker is really an ABI checker isn't it). It would be nice to preserve that, at least for apps running outside the deterministic sandbox that use the normal API (obviously it doesn't matter for existing code).

Let me see if I understand the details here. Currently Java code cannot use any default methods on our interfaces because Java doesn't know about the $DefaultImpls class. Kotlin code can, because it creates invokestatic calls to the hidden class.

When we apply these annotations the bytecode becomes 'regular' and the $DefaultImpls classes go away because they are no longer needed. Default method calls after a recompile become regular invokeinterface calls. After that Java code can invoke these methods.

Therefore to preserve ABI compatibility, not just source API compatibility, we'd need to do one of these things:

1. Not use the annotation. Not good, we want Java users to benefit from the default methods and resolve the issues you were working on.
2. Generate our own $DefaultImpls classes that just forward immediately to the real interface default methods (or vice-versa).
3. Rewrite the app code on the fly to substitute call sites.

On the original YouTrack issue discussing this new annotation, someone said they wrote a Gradle plugin to generate default methods that forward to $DefaultImpls classes. I've asked them if it's available or open source. 

The original work relates to determination. Why, exactly? I don't fully see the link between eliminating $DefaultImpls classes and deterministic execution, although, fixing this issue would be good for our Java users anyway.

chrisr3:
My issue with the `$DefaultImpls` classes is that ProGuard is prone to _throwing them away_ unless one of their methods is actually _used_ internally. (Not really sufficient, since `core` is a library jar.) I cannot annotate these classes because they are synthetic, and ProGuard doesn't have an option to "keep classes with a name matching [one it already intends to keep] + `$DefaultImpls`". So my options are:
- Keep _all_ `$DefaultImpls` classes (and _hope_ none of them has non-deterministic behaviour)
- Maintain a list _manually_ of every `$DefaultImpls` class we need to keep. (Brittle, and a maintenance headache.)
- Use the `@JvmDefault` annotation so that `$DefaultImpls` classes are no longer required. (_Nice!_ But Java users might notice the classes disappear from the regular build).
- Write a Gradle task that will annotate all `$DefaultImpls` class with exactly the same annotations as their "host" interface class. (It would be nice if Kotlin did that anyway - _ahem!_)

As for generating our own `$DefaultImpls` classes, we'd either need to write our own Gradle/Kotlin-Compiler plugin to do that, or rewrite the affected interfaces in Java, because the Kotlin compile will refuse to compile an interface containing a `DefaultImpls` nested class.

mikehearn:
How many interfaces are there? Is converting them to Java feasible? There's a decompiler in the "show bytecode" screen that does a decent job, albeit some variables need renaming because they don't appear in the kotlin code (result of e.g. inlining).

chrisr3:
We're only talking about 2 interfaces in `core`. There are also some interfaces in `node-api`, but they're in an `.internal.` package,

mikehearn:
Can we just port those interfaces to Java please? I'd rather not violate the ABI stability we've provided so far.

chrisr3:
We have:
```kotlin
interface Attachment : NamedByHash {
   ...
   val signers: List<Party>
   ...
}
```
as public API, which in Java would become:
```java
interface Attachment extends NamedByHash {
    ...
    List<Party> getSigners();
    ...
}
```
This is not a transparent change in Kotlin, as our own `AbstractAttachment` demonstrates:
```kotlin
abstract class AbstractAttachment(...) : Attachment {
    ...
    override val signers by lazy {
        // This fails to compile because Kotlin says it doesn't override anything.
    }
    ...
}
```
Porting this to a function remove the `lazy` nature.

chrisr3:
I've refactored this to remove the `:core` interfaces because there are obviously subtleties that I can't resolve yet.

chrisr3:
And you seem to be referring to [KT-19415](https://youtrack.jetbrains.com/issue/KT-19415).

mikehearn:
Maybe we don't need the lazyness there, or it can be done trivially by hand.

chrisr3:
I think the point is that we'd also be forcing _everyone_ who has already implemented these interfaces in Kotlin to make similar changes - and in the name of "API Stability" too.

Issue #2984, opened 2018-04-19
Unable to change the issued key to a node


abishekktwm:
Thank you for choosing to report an issue with Corda. 

When reporting an issue please make sure it contains;

* A clear description of the issue

If I issue a new certificate to a node (say by removing the existing certificate and starting the node to issue a new dev certificate) it corrupts the node database and throws the following exception

* Any logs or stack traces that you can provide (use a site like https://pastebin.com for larger logs)
```

E 13:50:32-0400 [main] internal.Node.run - Exception during node startup {}
 java.lang.IllegalArgumentException: More than one node found with legal name OU=XXX, O=XXX, L=XXX, C=XXX
	at net.corda.node.services.network.PersistentNetworkMapCache.getNodeByLegalName(PersistentNetworkMapCache.kt:161) ~[corda-node-corda-3.0-RC01.jar:?]
	at net.corda.node.services.network.NetworkMapCacheImpl.getNodeByLegalName(PersistentNetworkMapCache.kt) ~[corda-node-corda-3.0-RC01.jar:?]
	at net.corda.node.internal.AbstractNode.initNodeInfo(AbstractNode.kt:322) ~[corda-node-corda-3.0-RC01.jar:?]
	at net.corda.node.internal.AbstractNode.access$initNodeInfo(AbstractNode.kt:103) ~[corda-node-corda-3.0-RC01.jar:?]
	at net.corda.node.internal.AbstractNode$start$4.invoke(AbstractNode.kt:207) ~[corda-node-corda-3.0-RC01.jar:?]
	at net.corda.node.internal.AbstractNode$start$4.invoke(AbstractNode.kt:103) ~[corda-node-corda-3.0-RC01.jar:?]


```
* Steps to reproduce the issue
Build a cordapp using cordformation. Start a node. Stop the node. Delete the certificates/*. Start the node again.
* The version/tag/release or commit hash it occured on
M18-RC01

mikehearn:
Is that M18 of Corda? As in from last year? I'm afraid if so, we don't support that old version anymore! Also the network map and identity handling code changed completely since then.

kchalkias:
This is still possible to happen in current version of Corda. `certificates` directory holds all of the legal identity node keys. It is advisable to one get a backup of this folder in case of accidental deletion.

[...]

PR #3087, opened 2018-05-08
Better error message when a legal name clash is found.

joeldudleyr3:


patrickkuo:
do we see this often? this name duplication should be handled in network map server(hence the reason of the error is not quite right), client shouldn't receive network map with multiple machine having the same name. Something is wrong if we see this in the node. (assuming we are using network map server)

joeldudleyr3:
It is being seen here: https://github.com/corda/corda/issues/2984.

patrickkuo:
so the reason for the error in #2984 is keystore was tempered/changed, maybe it's more useful to print out the the node infos(CN and public key)

joeldudleyr3:
Can you provide an example? Not sure what you mean.

patrickkuo:
maybe something like: "More than one node found with legal name $name. NodeInfos: $nodeInfos"

the root cause of the problem is in the addNode method. We checks for public key when adding node, where we should also check the name.  We need a bug jira to rectify this.

joeldudleyr3:
@patrickkuo @mnesbit Thank you both. I’ll close this and split it into individual bug JIRAs.

PR #3092, opened 2018-05-08
Test notarisation of issue transactions with time-window

thschroeter:


mnesbit:
Not sure this tests very much as the time windows is from min time to infinity? I fully understand we don't want to create race conditions/clock skew sensitive/windows time granularity sensitive tests, but is there a better test? I am missing what this is fixing?

thschroeter:
The tests ensure the notary can handle transactions without input states. What matters in this case is the presence of the time-window not its details. (Transactions without input states and without time window aren't notarised.)

adagys:
That being said, it may be more useful to add, say, a more restricted 30s timewindow

node/src/integration-test/kotlin/net/corda/node/services/SimpleNotaryServiceTests.kt
```diff
@@ -0,0 +1,92 @@
...
+import net.corda.testing.driver.internal.InProcessImpl
+import net.corda.testing.node.ClusterSpec
+import net.corda.testing.node.NotarySpec
+import org.junit.Test
+import java.time.Instant
+import java.util.*
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class SimpleNotaryServiceTests {
```
adagys: we already have `ValidatingNotaryServiceTests`, no need to create a new class

PR #3093, opened 2018-05-08
CORDA-1393: Make Explorer GUI recover on RPC connection loss.

vkolomeyko:
Add ability for Explorer GUI to re-connect if RPC connection is down.
With these change once RPC connection is down the following happens:
1. The GUI is greyed out and no longer clickable;
2. On the background attempts are being made to re-establish connection;
3. Once connection re-established, up-to-date transactions image is received from the server;
4. The GUI is enabled back and can be worked with.
greyedexplorer2

### DIFF
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
@@ -153,7 +153,7 @@ class NodeMonitorModel {
         val retryableStateMachineUpdatesSubscription: AtomicReference<Subscription?> = AtomicReference(null)
         val subscription: Subscription = stateMachineUpdatesRaw
                 .startWith(stateMachineInfos.map { StateMachineUpdate.Added(it) })
-                .doOnError {
+                .subscribe({ retryableStateMachineUpdatesSubject.onNext(it) }, {
                     // Terminate subscription such that nothing gets past this point to downstream Observables.
                     retryableStateMachineUpdatesSubscription.get()?.unsubscribe()
                     // Flag to everyone that proxy is no longer available.
@@ -164,8 +164,7 @@ class NodeMonitorModel {
                     connection.forceClose()
                     // Perform re-connect.
                     performRpcReconnect(nodeHostAndPort, username, password)
-                }
-                .subscribe(retryableStateMachineUpdatesSubject)
+                })
 
         retryableStateMachineUpdatesSubscription.set(subscription)
         runLaterIfInitialized { proxyObservable.set(CordaRPCOpsWrapper(proxy)) }


client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -48,27 +58,84 @@ class NodeMonitorModel {
...
+                op()
+            }
+        }
+    }
+
+    /**
+     * This is needed as JavaFX listener framework attempts to call `equals()` before dispatching notification change.
+     * And calling `CordaRPCOps.equals()` results in (unhandled) remote call.
+     */
+    class CordaRPCOpsWrapper(val cordaRPCOps: CordaRPCOps) {
```
mnesbit: Worth seeing if Andras has a nicer fix/workaround?

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -48,27 +58,84 @@ class NodeMonitorModel {
...
+        override fun equals(other: Any?): Boolean {
+            return this === other
+        }
+
+        override fun hashCode(): Int {
+            throw IllegalArgumentException()
```
mnesbit: Really don't like having only equals defined. any reason the base Object hashCode doesn't work?

vkolomeyko:
Minor cosmetics change pushed as suggested by @tlil

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
vkolomeyko: @exFalso - based on the KDoc above `CordaRPCOpsWrapper`, do you have better ideas how `CordaRPCOps` can be passed through JavaFX `ObjectProperty` framework?

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
vkolomeyko: `hashCode()` most likely will work. However, using this implementation I am really trying to discourage people using the class as a key for a `HashMap`.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -82,33 +149,75 @@ class NodeMonitorModel {
...
+                logger.info("Exception upon establishing connection: " + th.message)
+                null
+            }
+
+            if(connection != null) {
+                logger.info("Connection successfully established with: $nodeHostAndPort")
+                return connection
+            }
+            // Could not connect this time round - pause before giving another try.
+            Thread.sleep(retryInterval.toMillis())
```
exFalso: This hogs an RPC thread, we should use a ScheduledExecutorService instead. Note that we should make sure that there is only ever a single retry scheduled, otherwise we may get a schedule blowup

client/jfx/src/main/kotlin/net/corda/client/jfx/utils/ChosenList.kt
```diff
@@ -21,7 +21,8 @@ import javafx.collections.ObservableListBase
  * The above will create a list that chooses and delegates to the appropriate filtered list based on the type of filter.
  */
 class ChosenList<E>(
-        private val chosenListObservable: ObservableValue<out ObservableList<out E>>
+        private val chosenListObservable: ObservableValue<out ObservableList<out E>>,
+        private val logicalName: String? = null
```
exFalso: logicalName?

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -360,7 +360,15 @@ class RPCClientProxyHandler(
             interrupt()
             join(1000)
         }
-        sessionFactory?.close()
+
+        if(notify) {
```
exFalso: whitespace :police_car:

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -360,7 +360,15 @@ class RPCClientProxyHandler(
...
+            // This is going to send remote message, see `org.apache.activemq.artemis.core.client.impl.ClientConsumerImpl.doCleanUp()`.
+            sessionFactory?.close()
+        } else {
+            // This performs a cheaper and faster version of local cleanup.
+            sessionFactory?.cleanup()
```
exFalso: I don't know whether this is sufficient. Does this clean up all sessions?

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -82,33 +149,75 @@ class NodeMonitorModel {
...
-        futurePartyUpdate.startWith(parties.map { MapChange.Added(it) }).subscribe(networkMapSubject)
-
-        proxyObservable.set(proxy)
+        val connection = establishConnectionWithRetry(nodeHostAndPort, username, password)
+        val proxy = connection.proxy
+
+        val (stateMachineInfos, stateMachineUpdatesRaw) = proxy.stateMachinesFeed()
+
+        val retryableStateMachineUpdatesSubscription: AtomicReference<Subscription?> = AtomicReference(null)
+        val subscription: Subscription = stateMachineUpdatesRaw
```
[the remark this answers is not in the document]
vkolomeyko: This manual `onNext()` feed is used for Vault subscription, please see: `net.corda.client.jfx.model.NodeMonitorModel#register()`
However in case of State Machine Feed I do special handling which performs re-connect.
Also, as discussed, in order to eliminate any sort of memory leaks I quite like to call `.unsubscribe()` hence `doOnError()` handler code is a bit more involved.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -82,33 +149,75 @@ class NodeMonitorModel {
...
+                null
+            }
+
+            if(connection != null) {
+                logger.info("Connection successfully established with: $nodeHostAndPort")
+                return connection
+            }
+            // Could not connect this time round - pause before giving another try.
+            Thread.sleep(retryInterval.toMillis())
+        } while (connection == null)
```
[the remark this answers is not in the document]
vkolomeyko: With current code it will try re-connecting infinitely, printing every attempt into the log.
We can create hard limit for such attempts say 500 before terminating Explorer GUI.
But user has always the option to close Exporer GUI if he/she doesn't want to wait any longer.

I am open to suggestions here whether we should limit the number of attempts.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -82,33 +149,75 @@ class NodeMonitorModel {
...
+                val _connection = client.start(username, password)
+                // Check connection is truly operational before returning it.
+                val nodeInfo = _connection.proxy.nodeInfo()
+                require(nodeInfo.legalIdentitiesAndCerts.isNotEmpty())
+                _connection
+            } catch( secEx: ActiveMQSecurityException) {
+                // Happens when incorrect credentials provided - no point to retry connecting.
+                throw secEx
+            }
+            catch (th: Throwable) {
```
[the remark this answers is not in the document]
vkolomeyko: Will change.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
vkolomeyko: Sure - we can create a single threaded execution service.
But why performing this on RPC thread is a bad idea? Whilst we keep doing it RPC is down anyway.
(btw it will not be always RPC thread, initial logon will probably do it in UI thread)

client/jfx/src/main/kotlin/net/corda/client/jfx/utils/ChosenList.kt
vkolomeyko: There are many `ChosenList` instances created. I created optional `logicalName` to simplify debugging of chained `Observable` calls.
This helped me getting down to the bottom of the issue.
If there is strong opposition for having this property - I do not mind removing it, but I thought it might be handy for the future.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
vkolomeyko: Will change.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
vkolomeyko: My testing (including using YourKit) shows that everything cleaned-up properly in this case.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
vkolomeyko: Yes, precisely - reference identity equality is exactly what I wanted here.
Whenever new RPC connection is established RPC Proxy will be different by reference.

In fact transition cycle for RPC proxy is the following:
null -> Ref1 -> null -> Ref2
So wrapper works very well in this case.

exFalso: I think @mnesbit meant that the default behaviour of simple `class` equals already does this, so no need to override anything

exFalso: So to be clear: use a wrapper, but don't override equals

vkolomeyko: Sure - I am with you now. Pushing the change.

PR #3102, opened 2018-05-09
Cleanup and improvements to the serialisation format of JacksonSupport (needed for CORDA-1238)

shamsasari:
Also deprecated all the public members that shouldn't have leaked into the public API.

### DIFF
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
@@ -2,56 +2,42 @@ package net.corda.client.jackson
 
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.fasterxml.jackson.annotation.JsonProperty
-import com.fasterxml.jackson.core.JsonFactory
-import com.fasterxml.jackson.core.JsonGenerator
-import com.fasterxml.jackson.core.JsonParseException
-import com.fasterxml.jackson.core.JsonParser
-import com.fasterxml.jackson.core.JsonToken
-import com.fasterxml.jackson.databind.DeserializationContext
-import com.fasterxml.jackson.databind.DeserializationFeature
-import com.fasterxml.jackson.databind.JsonDeserializer
-import com.fasterxml.jackson.databind.JsonNode
-import com.fasterxml.jackson.databind.JsonSerializer
-import com.fasterxml.jackson.databind.Module
-import com.fasterxml.jackson.databind.ObjectMapper
-import com.fasterxml.jackson.databind.SerializationFeature
-import com.fasterxml.jackson.databind.SerializerProvider
+import com.fasterxml.jackson.core.*
+import com.fasterxml.jackson.databind.*
 import com.fasterxml.jackson.databind.deser.std.NumberDeserializers
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer
 import com.fasterxml.jackson.databind.module.SimpleModule
+import com.fasterxml.jackson.databind.node.ObjectNode
 import com.fasterxml.jackson.databind.ser.std.StdSerializer
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
 import com.fasterxml.jackson.module.kotlin.KotlinModule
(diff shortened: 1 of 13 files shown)


client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
```diff
@@ -133,125 +135,172 @@ object JacksonSupport {
...
-                    mapper.partyFromKey(key) ?: throw JsonParseException(parser, "Could not find a Party with key ${key.toStringShort()}")
+                    val publicKey = parser.readValueAs<PublicKey>()
+                    mapper.partyFromKey(publicKey)
+                            ?: throw JsonParseException(parser, "Could not find a Party with key ${publicKey.toStringShort()}")
                 } else if (nameMatches.size == 1) {
                     nameMatches.first()
                 } else {
-                    throw JsonParseException(parser, "Ambiguous name match '${parser.text}': could be any of " + nameMatches.map { it.name }.joinToString(" ... or ..."))
+                    throw JsonParseException(parser, "Ambiguous name match '${parser.text}': could be any of " +
+                            nameMatches.map { it.name }.joinToString(" ... or ..."))
```
tlil: This is missing a space after `...` and before `"`, no?

client/jackson/src/test/kotlin/net/corda/client/jackson/JacksonSupportTest.kt
```diff
@@ -95,7 +88,7 @@ class JacksonSupportTest {
     @Test
     fun writeAmount() {
         val writer = mapper.writer().without(SerializationFeature.INDENT_OUTPUT)
-        assertEquals("""{"notional":"25000000.00 USD"}""", writer.writeValueAsString(Dummy(Amount.parseCurrency("$25000000"))))
+        assertEquals("""{"notional":"25000000.00 GBP"}""", writer.writeValueAsString(Dummy(Amount.parseCurrency("£25000000"))))
```
tlil: Check _both_ USD and GBP instead of replacing?

client/jackson/src/test/kotlin/net/corda/client/jackson/JacksonSupportTest.kt
```diff
@@ -119,6 +112,188 @@ class JacksonSupportTest {
 
         val writer = mapper.writer()
         // We don't particularly care about the serialized format, just need to make sure it completes successfully.
-        writer.writeValueAsString(makeDummyTx())
+        val output = writer.writeValueAsString(makeDummyTx())
+        println(output)
```
tlil: Why the `println` statement? Rather place an assertion on the output being non-empty or something?

docs/source/shell.rst
```diff
@@ -263,23 +263,44 @@ A parameter of type ``Amount<Currency>`` can be written as either:
...
+~~~~~~~~
+A parameter of type ``NodeInfo`` can be written in terms of one of its identities (see ``Party`` above)
+
+AnonymousParty
+~~~~~~~~~~~~~~
+A parameter of type ``AnonymousParty`` can be written in terms of its ``PublicKey`` (see above)
+
+NetworkHostAndPort
+~~~~~~~~~~~~~~~~~~
+A parameter of type ``NetworkHostAndPort`` can be written as a "host:port" string: ``localhost:1010``
```
tlil:
```
``"localhost:1010"``
```

client/jackson/src/test/kotlin/net/corda/client/jackson/JacksonSupportTest.kt
shamsasari: Asserting on non-empty string wouldn't be that useful. The mapper would have thrown an exception if it couldn't serialise the tx. Also, @sollecitom is updating the serialisation of txs and has a better test.

tlil: Fair enough, but why the print statement? :)

shamsasari: I was checking what the tx output was. Turns out not much! Anyway it doesn't matter, this test is deleted on master.

PR #3108, opened 2018-05-09
CORDA-1466: Fix publication of PDF developer documentation

josecoll:
Published PDF file is missing because of a missing target directory in the script move command:
https://docs.corda.net/head/_static/corda-developer-site.pdf

This fix ensures that the generated PDF file is copied to the correct place without error.

### DIFF
--- a/docs/make-docsite.sh
+++ b/docs/make-docsite.sh
@@ -12,6 +12,12 @@ else
 fi
 
 # TODO: The PDF rendering is pretty ugly and can be improved a lot.
+echo "Generating PDF document ..."
 make pdf
-mv build/pdf/corda-developer-site.pdf build/html/_static/corda-developer-site.pdf
+
+echo "Generating HTML pages ..."
 make html
+
+echo "Moving PDF file into place ..."
+mv $PWD/build/pdf/corda-developer-site.pdf $PWD/build/html/_static/corda-developer-site.pdf
+


mikehearn:
Thanks for digging into this. I couldn't figure out why CI wasn't building it properly when it worked locally.

I don't understand how this fix works though. Paths are relative to the current working directory, so prefixing a path with `$PWD/` should be a no-op. Why is it different?

josecoll:
The issue was causes by a missing target **html** directory (causing the `mv` command to fail).
Re-ordering the step sequence to run `make htm` before the `mv` solves the problem.

mikehearn:
D'oh, of course, thanks. What a brainfart.

josecoll:
Bizarrely, the OS PDF generation is failing (a ZERO byte PDF file).
https://ci-master.corda.r3cev.com/viewLog.html?buildId=82876&buildTypeId=Corda_Docs_Build&tab=buildLog&_focus=199#_state=88,199

**Enterprise however appears to generate the PDF correctly:**
https://ci-master.corda.r3cev.com/viewLog.html?buildId=82600&buildTypeId=CordaEnterprise_Docs_Build&tab=buildLog&_focus=217#_state=106,217

Final PDF document is accessible:
http://docs.corda.r3.com/website/releases/docs_head/_static/corda-developer-site.pdf

Comparing the two TeamCity build logs (OS vs ENT) it is impossible to determine why one succeeds and the other fails. Will need to investigate with further debug enabled.

PR #3111, opened 2018-05-10
Increase the number of log files retained per day

adagys:
Currently the node will keep up to 10 compressed log files during rollover, and discard old ones. This limit seems too low for production use and might lead to valuable log data loss.


adagys:
Created a ticket https://r3-cev.atlassian.net/browse/CORDA-1502

PR #3121, opened 2018-05-11
ENT-1463: Turn ClassCarpenter and SerializerFactoryFactory into interfaces.

chrisr3:
The DJVM will want to drop the Class Carpenter, so abstract it behind some interfaces. More precisely:
- `ClassCarpenter` -> `ClassCarpenterImpl`
- `SerializerFactoryFactory` -> `SerializerFactoryFactoryImpl`

with new interfaces `ClassCarpenter` and `SerializerFactoryFactory`.

Also fix up some warnings in the unit tests.


chrisr3:
You'll just have to settle for `FactoryFactory` refactoring instead :wink:

PR #3123, opened 2018-05-11
`snapshot` in corda version should be uppercase

patrickkuo:
* Artifactory refuse nightly deployment because of this

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     file("$projectDir/constants.properties").withInputStream { constants.load(it) }
 
     // Our version: bump this on release.
-    ext.corda_release_version = "corda-4.0-snapshot"
+    ext.corda_release_version = "corda-4.0-SNAPSHOT"
     // Increment this on any release that changes public APIs anywhere in the Corda platform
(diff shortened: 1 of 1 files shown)


josecoll:
I thought the Maven convention was to publish in lowercase (eg. less shouty)
@Clintonio ^^^

patrickkuo:
I think the convention is all uppercase, and TC job failed with error
`Execution failed for task ':artifactoryPublish'. org.gradle.api.UncheckedIOException: java.io.IOException: Failed to deploy file. Status code: 409 Response message: Artifactory returned the following errors: 
The repository 'corda-dev' rejected the resolution of an artifact 'corda-dev:net/corda/corda-confidential-identities/corda-4.0-snapshot/corda-confidential-identities-corda-4.0-snapshot-javadoc.jar' due to conflict in the snapshot release handling policy. Status code: 409`

Clintonio:
Yep, it's 100% uppercase. Not sure of the reason but if I had to guess it makes it more obvious that it shouldn't be used in production.

It's part of why I opposed the "CORDA-3.0" version string before. That and it's just not standard.

PR #3124, opened 2018-05-11
ENT-1850: Improve reporting of connection problems

mnesbit:
Previously the AMQPClient code in the bridges logged a lot of ambiguous exception traces and we were retrying every second. Now we do exponential backoff and I explicitly log a light summary of the provided certificate paths when we go to check them and a full detail list when we get a validation error.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
@@ -83,22 +83,32 @@ internal class AMQPChannelHandler(private val serverMode: Boolean,
                 val sslHandler = ctx.pipeline().get(SslHandler::class.java)
                 localCert = sslHandler.engine().session.localCertificates[0].x509
                 remoteCert = sslHandler.engine().session.peerCertificates[0].x509
-                try {
-                    val remoteX500Name = CordaX500Name.build(remoteCert!!.subjectX500Principal)
-                    require(allowedRemoteLegalNames == null || remoteX500Name in allowedRemoteLegalNames)
-                    log.info("handshake completed subject: $remoteX500Name")
+                val remoteX500Name = try {
+                    CordaX500Name.build(remoteCert!!.subjectX500Principal)
                 } catch (ex: IllegalArgumentException) {
-                    log.error("Invalid certificate subject", ex)
+                    log.error("Certificate subject not a valid CordaX500Name", ex)
                     ctx.close()
                     return
                 }
+                if (allowedRemoteLegalNames != null && remoteX500Name !in allowedRemoteLegalNames) {
+                    log.error("Provided certificate subject $remoteX500Name not in expected set $allowedRemoteLegalNames")
+                    ctx.close()
+                    return
+                }
+                log.info("handshake completed with subject: $remoteX500Name")
                 createAMQPEngine(ctx)
                 onOpen(Pair(ctx.channel() as SocketChannel, ConnectionChange(remoteAddress, remoteCert, true)))
             } else {
-                log.error("Handshake failure $evt")
+                log.error("Handshake failure ${evt.cause().message}")
                 ctx.close()
             }
         }
+    }
+
+    @Suppress("OverridingDeprecatedMember")
+    override fun exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable) {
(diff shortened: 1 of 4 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
```diff
@@ -83,24 +83,34 @@ internal class AMQPChannelHandler(private val serverMode: Boolean,
...
+                    log.error("Certificate subject not a valid CordaX500Name", ex)
                     ctx.close()
                     return
                 }
+                if (allowedRemoteLegalNames != null && remoteX500Name !in allowedRemoteLegalNames) {
+                    log.error("Provided certificate subject $remoteX500Name not in expected set $allowedRemoteLegalNames")
+                    ctx.close()
+                    return
+                }
+                log.info("handshake completed with subject: $remoteX500Name")
```
tlil: `s/handshake/Handshake/`

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
```diff
@@ -83,24 +83,34 @@ internal class AMQPChannelHandler(private val serverMode: Boolean,
...
                 createAMQPEngine(ctx)
                 onOpen(Pair(ctx.channel() as SocketChannel, ConnectionChange(remoteAddress, remoteCert, true)))
             } else {
-                log.error("Handshake failure $evt")
+                log.error("Handshake failure ${evt.cause().message}")
```
tlil: Does `cause()` return a `Throwable`? If so, should that be passed in as well? E.g.:

```kt
val cause = evt.cause()
log.error("Handshake failure ${cause.message}", cause)
```

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
```diff
@@ -83,24 +83,34 @@ internal class AMQPChannelHandler(private val serverMode: Boolean,
...
                 ctx.close()
             }
         }
     }
 
+    @Suppress("OverridingDeprecatedMember")
+    override fun exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable) {
+        log.warn("Closing channel due to nonrecoverable exception ${cause.message}")
```
tlil: Same as above, should `cause` be provided as a second argument to `log.warn`? Or does it just pollute the log file for no additional value in this case?

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
mnesbit: The truly horrid stacked nested cause is the thing I am trying to suppress. It isn't helpful!

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
mnesbit: No it contains no end user helpful information

tlil: Okay :+1: Makes sense.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
tlil: Yep, remember that stacktrace now that you say it. As long as it doesn't provide any useful information, that makes sense.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPChannelHandler.kt
mnesbit: What I will do is a a trace {} version that does log it all, just in case

tlil: Yep, okay. In case a future version suddenly starts producing more helpful messages/traces or something.

PR #3128, opened 2018-05-11
 CORDA-866: Implement removal of stale nodes from network - backport

kasiastreich:
* Implement removal of stale nodes from network

Add eventHorizon to NetworkParameters structure. Add republishing of
node info on 1 day intervals - it is treated by network map as heartbeat from node indicating if it's alive or not. Add removal of old node infos on network map signing.

* Add copy method to NetworkParameters data class

Add JvmOverloads annotation to the constructor, because it's data class
exposed in API


shamsasari:
Backport of https://github.com/corda/enterprise/pull/774

PR #3130, opened 2018-05-14
Abstract out of a class to reduce duplicated logic between InMemoryId…

renlulu:
…entityService and PersistentIdentityService


node/src/main/kotlin/net/corda/node/services/identity/AbstractIdentityService.kt
```diff
@@ -0,0 +1,30 @@
+package net.corda.node.services.identity
+
+import net.corda.core.identity.CordaX500Name
+import net.corda.core.identity.Party
+import net.corda.core.node.services.IdentityService
+import net.corda.core.serialization.SingletonSerializeAsToken
+import net.corda.node.services.api.IdentityServiceInternal
+
+abstract class AbstractIdentityService : SingletonSerializeAsToken(), IdentityService, IdentityServiceInternal {
```
joeldudleyr3: This is a stateless function, so we don't need abstract inheritance here. Can you change this to a top-level function?

In the future, we may abstract out and combine further functionality from the individual identity services, at which point a class would become more useful.

joeldudleyr3:
@renlulu It looks like you accidentally pulled in a bunch of extra commits - can you remove those and create a new PR with just your changes?

Also, I suggest you delete the `AbstractIdentityService` class now - it's not needed.

Does that make sense?

Issue #2262, opened 2017-12-15
Bug: `otherResults` cannot exceed 200 results even with `PageSpec`


joeldudleyr3:
This issue is reproduced here: https://github.com/joeldudleyr3/temp-for-simon.

Running the `query test` flow test will cause an exception to be thrown, requesting a page specification to be provided (even though we've already provided a page specification of `MAX_SIZE`.

This seems to be an issue whereby the page specification supplied isn't applied to `otherResults`.

joeldudleyr3:
See https://r3-cev.atlassian.net/browse/CORDA-861.

joeldudleyr3:
This has been resolved.

PR #3135, opened 2018-05-14
CORDA-861 Pagination failure with aggregate groupBy query

josecoll:
Pagination relies on a recursive call to count total results, this internal sub-query should NOT perform pagination constraint checking.


josecoll:
This also fixes also relates to the issue: https://github.com/corda/corda/issues/2262

PR #3141, opened 2018-05-14
Extracted a tool class to reduce duplicated logic between InMemoryIde…

renlulu:
…ntityService and PersistentIdentityService.


renlulu:
I certify that my contribution is in accordance with the terms of http://developercertificate.org/ ：）

PR #3142, opened 2018-05-14
Remove duplicate Netty classes from Node.

chrisr3:
Artemis includes Netty piecemeal, but we prefer to include `netty-all` instead. This means that we end up with _two_ copies of the Netty classes because Gradle is unaware of the overlap.

Fix this by excluding `netty-all` from every module. This will force us to specify precisely which pieces of Netty we wish to use.


chrisr3:
This works for Gradle, but may not work in IntelliJ:
```gradle
resolutionStrategy {
    eachDependency { details ->
        if (details.requested.group == 'io.netty') {
            details.useVersion netty_version
        }
    }
}
```

PR #3152, opened 2018-05-15
 Add configuration `notary.serviceLegalName` for clustered notaries

thschroeter:
For backwards compatibility the keystore is only checked if the `notary.serviceLegalName` is present. 

The legal name of the service can be read by the bootstrapper tool to setup networks with distributed notaries.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
@@ -536,6 +536,7 @@ class DriverDSLImpl(
             val clusterAddresses = if (clusterAddress != null) listOf(clusterAddress) else emptyList()
             val config = NotaryConfig(
                     validating = spec.validating,
+                    serviceLegalName = spec.name,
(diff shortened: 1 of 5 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -1125,4 +1126,4 @@ private fun Config.toNodeOnly(): Config {
     return if (hasPath("webAddress")) withoutPath("webAddress").withoutPath("useHTTPS") else this
 }
 
-private operator fun Config.plus(property: Pair<String, Any>) = withValue(property.first, ConfigValueFactory.fromAnyRef(property.second))
\ No newline at end of file
+private operator fun Config.plus(property: Pair<String, Any>) = withValue(property.first, ConfigValueFactory.fromAnyRef(property.second))
```
shamsasari: Is this used?

thschroeter: As far as i can tell, the plus operator is actually not used.

shamsasari: If it's not used then remove it.

PR #3155, opened 2018-05-15
[CORDA-1264}: Complete obfuscation of exceptions to client side.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1264

Ensure all exceptions are wrapped with `CordaRuntimeException` if not serialisable.


.idea/compiler.xml
```diff
@@ -4,123 +4,38 @@
     <bytecodeTargetLevel target="1.8">
```
mikehearn: Can you revert this file please? I'm not sure the contents matters much but it does cause merge conflicts.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/DuplicateAttachmentException.kt
```diff
@@ -0,0 +1,8 @@
+package net.corda.nodeapi.exceptions
+
+import net.corda.core.flows.ClientRelevantError
+
+/**
```
mikehearn: Does one line of code really justify its own file? Just move this into the same file that defines the Attachment class and I guess it's fine.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/InternalNodeException.kt
```diff
@@ -1,37 +1,18 @@
...
-        FlowException::class,
-        InvalidClassException::class,
-        RpcSerializableError::class,
-        TransactionVerificationException::class
-)
+import net.corda.core.flows.ClientRelevantError
+import net.corda.core.flows.ContextAware
 
 /**
  * An [Exception] to signal RPC clients that something went wrong within a Corda node.
```
mikehearn: Can you add some discussion here of why the exception type is so unhelpful? Talk about privacy, security etc.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/NonRpcFlowException.kt
```diff
@@ -0,0 +1,8 @@
+package net.corda.nodeapi.exceptions
+
+import net.corda.core.flows.ClientRelevantError
+
+/**
+ * Thrown to indicate that a flow was not designed for RPC and should be started from an RPC client.
```
mikehearn: Again can you move this into a different file so it's closer to the place where it's used?

node/src/main/kotlin/net/corda/node/internal/AuthenticatedRpcOpsProxy.kt
```diff
@@ -0,0 +1,50 @@
...
+            return Proxy.newProxyInstance(delegate::class.java.classLoader, arrayOf(CordaRPCOps::class.java), handler) as CordaRPCOps
+        }
+    }
+
+    private class PermissionsEnforcingInvocationHandler(override val delegate: CordaRPCOps, private val context: () -> RpcAuthContext) : InvocationHandlerTemplate {
+        override fun invoke(proxy: Any, method: Method, arguments: Array<out Any?>?) = guard(method.name, context, { super.invoke(proxy, method, arguments) })
+    }
+}
+
+private inline fun <RESULT> guard(methodName: String, context: () -> RpcAuthContext, action: () -> RESULT) = guard(methodName, emptyList(), context, action)
```
mikehearn: Does this need to be inline? i'm not sure it does.

node/src/main/kotlin/net/corda/node/internal/ExceptionMaskingRpcOpsProxy.kt
```diff
@@ -0,0 +1,127 @@
...
+import net.corda.core.setFieldToNull
+import net.corda.core.utilities.loggerFor
+import net.corda.nodeapi.exceptions.InternalNodeException
+import rx.Observable
+import java.io.InvalidClassException
+import java.lang.reflect.Method
+import java.lang.reflect.Proxy.newProxyInstance
+import kotlin.reflect.KClass
+
+// TODO sollecitom make this enterprise only as discussed with Mike
```
mikehearn: Well you just uploaded it to the open repository..... :( can you force push over this PR to get rid of it please?

node/src/test/kotlin/net/corda/node/CordaRPCOpsImplTest.kt
```diff
@@ -293,6 +292,7 @@ class CordaRPCOpsImplTest {
     @Test
     fun `attempt to start non-RPC flow`() {
         withPermissions(startFlow<NonRPCFlow>()) {
+            // TODO sollecitom fix this
```
mikehearn: Fix what?

core/src/main/kotlin/net/corda/core/flows/ClientRelevantError.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.core.flows
+
+import net.corda.core.serialization.CordaSerializable
+
+/**
+ * Allows an implementing [Throwable] to be propagated to clients.
```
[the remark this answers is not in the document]
sollecitom: Well, in theory we could send `Throwable` instances. Happy to rename it, if we assume it will only be just `Exception`s.

Also, `FlowException` is semantically quite different, being used to flag an error to be propagated to other P2P parties. Also, `FlowException` is a class rather than an interface, so it's usage it's way more inconvenient.

core/src/main/kotlin/net/corda/core/Utils.kt
```diff
@@ -71,3 +72,50 @@ fun <ELEMENT> Observable<ELEMENT>.mapErrors(transform: (Throwable) -> Throwable)
         Observable.error(transform(error))
     }
 }
+
+/**
+ * Find a field withing a class hierarchy.
```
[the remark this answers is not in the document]
sollecitom: I wasn't aware of that! Thanks!

core/src/main/kotlin/net/corda/core/flows/ContextAware.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.core.flows
+
+/**
+ * Allows types to specify additional context. Useful for exceptions or events.
```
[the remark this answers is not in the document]
sollecitom: The point here is: for exceptions, we sometimes allow an extra `errorId` field, which is there for a customer to be able to go to another party and ask for correlated logs. I could easily use the existing interface for that, but it looks useful for us to be able to specify key-value pairs of context to propagate with exception types.

Ideally, it shouldn't be part of the API, but if I don't make it part of it, it will prevent correct deserialisation on client side. I can easily mark it with `@DoNotImplement`.

Let me know.

.idea/compiler.xml
sollecitom: Not even sure why it's in version control.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/DuplicateAttachmentException.kt
sollecitom: I'll create a file for all exceptions used in `CordaRPCOps`.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/InternalNodeException.kt
sollecitom: Sure, good point.

node-api/src/main/kotlin/net/corda/nodeapi/exceptions/NonRpcFlowException.kt
sollecitom: I move this into a `RpcExceptions` file, along with others.

node/src/main/kotlin/net/corda/node/internal/AuthenticatedRpcOpsProxy.kt
sollecitom: Removed. Wasn't changing that as part of this PR (it's marked as a new line because I renamed the file), but good point anyway.

node/src/test/kotlin/net/corda/node/CordaRPCOpsImplTest.kt
sollecitom: Fixed by replacing it with `NonRpcFlowException`.

core/src/main/kotlin/net/corda/core/flows/ContextAware.kt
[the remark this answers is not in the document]
sollecitom: That said, given the need to revert `IdentifiableException`, I removed this as well.

sollecitom:
@mikehearn back to you.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -277,10 +283,39 @@ fun <T : Any> KClass<T>.objectOrNewInstance(): T {
...
+            return javaField.accessible { uncheckedCast<Any?, T>(get(receiver)) }
+        }
+        set(value) {
+            javaField.accessible {
+                set(receiver, value)
+            }
+        }
+    val name: String = javaField.name
+
+    private fun <RESULT> Field.accessible(action: Field.() -> RESULT): RESULT {
```
mikehearn: Is there much reason to put it back afterwards? From a sandboxing perspective it's still insecure because accessibility is global and another thread could race with this one.

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -290,9 +290,11 @@ interface CordaRPCOps : RPCOps {
     fun openAttachment(id: SecureHash): InputStream
 
     /** Uploads a jar to the node, returns it's hash. */
+    @Throws(java.nio.file.FileAlreadyExistsException::class)
```
mikehearn: Likewise this could be in a small fixes branch.

node/src/main/kotlin/net/corda/node/internal/AuthenticatedRpcOpsProxy.kt
```diff
@@ -0,0 +1,50 @@
+package net.corda.node.internal
```
mikehearn: Should these be inside an RPC specific package perhaps

node/src/main/kotlin/net/corda/node/internal/AuthenticatedRpcOpsProxy.kt
```diff
@@ -0,0 +1,50 @@
...
+ * Implementation of [CordaRPCOps] that checks authorisation.
+ */
+internal class AuthenticatedRpcOpsProxy(private val delegate: CordaRPCOps) : CordaRPCOps by proxy(delegate, ::rpcContext) {
+    /**
+     * Returns the RPC protocol version, which is the same the node's Platform Version. Exists since version 1 so guaranteed
+     * to be present.
+     */
+    override val protocolVersion: Int get() = delegate.nodeInfo().platformVersion
+
+    override fun <T> startFlowDynamic(logicType: Class<out FlowLogic<T>>, vararg args: Any?) = guard("startFlowDynamic", listOf(logicType), ::rpcContext) {
```
mikehearn: Are these two methods special because they're varargs? I don't grok why they have to be handled specially. A comment would be helpful

node/src/main/kotlin/net/corda/node/internal/ExceptionMaskingRpcOpsProxy.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        private fun removeDetails(error: Throwable) {
+            if (error is CordaThrowable) {
+                error.stackTrace = arrayOf<StackTraceElement>()
+                error.setCause(null)
+            }
+        }
+
+        override fun toString(): String {
```
mikehearn: What are these toString methods for?

core/src/main/kotlin/net/corda/core/flows/ClientRelevantError.kt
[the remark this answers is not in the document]
sollecitom: Moved to `net.corda.core`.

PR #3157, opened 2018-05-15
ENT-933 Add spectator and participant profiles of rigorousMock

andr3ej:


build.gradle
```diff
@@ -52,7 +52,6 @@ buildscript {
     ext.typesafe_config_version = constants.getProperty("typesafeConfigVersion")
     ext.fileupload_version = '1.3.3'
     ext.junit_version = '4.12'
-    ext.mockito_version = '2.10.0'
```
mnesbit: I do actually like having version variables, so not that keen on losing this. I know we probably have only one use site, but I keep finding myself having to re-declare dependencies in other configurations at some point.

andr3ej: whoever redeclares should add the variable

mnesbit:
Seems to fail some IRS test

andr3ej:
all green now if anyone feels like approving

PR #3162, opened 2018-05-16
CORDA-1238: Updated JacksonSupport to serialise pojos annotated with @CordaSerializable…

shamsasari:
… with the same properties as the AMQP serialisation framework.

This fixes an issue for pojos with getters that aren't c'tor-based properties. Jackson serialises these out but is then unable to deserialise the pojo back.

Had to move the property scanning code into core (internal). Also renamed PropertyAccessor.getter to serializer as it's misleading otherwise.

### DIFF
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
@@ -1,0 +1,194 @@
+@file:Suppress("DEPRECATION")
+
+package net.corda.client.jackson.internal
+
+import com.fasterxml.jackson.annotation.JsonValue
+import com.fasterxml.jackson.core.JsonGenerator
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.core.JsonToken
+import com.fasterxml.jackson.databind.*
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.annotation.JsonSerialize
+import com.fasterxml.jackson.databind.module.SimpleModule
+import com.fasterxml.jackson.databind.node.ObjectNode
+import com.fasterxml.jackson.databind.ser.BeanPropertyWriter
+import com.fasterxml.jackson.databind.ser.BeanSerializerModifier
+import net.corda.client.jackson.JacksonSupport
+import net.corda.core.contracts.Amount
+import net.corda.core.crypto.DigitalSignature
+import net.corda.core.crypto.SecureHash
+import net.corda.core.crypto.TransactionSignature
+import net.corda.core.identity.*
+import net.corda.core.internal.DigitalSignatureWithCert
+import net.corda.core.node.NodeInfo
+import net.corda.core.serialization.*
+import net.corda.core.transactions.SignedTransaction
+import net.corda.core.transactions.WireTransaction
+import net.corda.core.utilities.ByteSequence
+import net.corda.core.utilities.NetworkHostAndPort
+import net.corda.serialization.internal.AllWhitelist
+import net.corda.serialization.internal.amqp.SerializerFactory
+import net.corda.serialization.internal.amqp.constructorForDeserialization
+import net.corda.serialization.internal.amqp.createSerializerFactoryFactory
+import net.corda.serialization.internal.amqp.propertiesForSerialization
+import java.security.PublicKey
+
+class CordaModule : SimpleModule("corda-core") {
+    override fun setupModule(context: SetupContext) {
+        super.setupModule(context)
+
+        context.addBeanSerializerModifier(CordaSerializableBeanSerializerModifier())
+
+        context.setMixInAnnotations(PartyAndCertificate::class.java, PartyAndCertificateSerializerMixin::class.java)
+        context.setMixInAnnotations(NetworkHostAndPort::class.java, NetworkHostAndPortMixin::class.java)
+        context.setMixInAnnotations(CordaX500Name::class.java, CordaX500NameMixin::class.java)
+        context.setMixInAnnotations(Amount::class.java, AmountMixin::class.java)
+        context.setMixInAnnotations(AbstractParty::class.java, AbstractPartyMixin::class.java)
+        context.setMixInAnnotations(AnonymousParty::class.java, AnonymousPartyMixin::class.java)
+        context.setMixInAnnotations(Party::class.java, PartyMixin::class.java)
+        context.setMixInAnnotations(PublicKey::class.java, PublicKeyMixin::class.java)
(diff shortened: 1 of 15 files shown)


client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
```diff
@@ -0,0 +1,195 @@
...
+ */
+private object CordaSerializableBeanSerializerModifier : BeanSerializerModifier() {
+    override fun changeProperties(config: SerializationConfig,
+                                  beanDesc: BeanDescription,
+                                  beanProperties: MutableList<BeanPropertyWriter>): MutableList<BeanPropertyWriter> {
+        // TODO We're assuming here that Jackson gives us a superset of all the properties. Either confirm this or
+        // make sure the returned beanProperties are exactly the AMQP properties
+        if (beanDesc.beanClass.isAnnotationPresent(CordaSerializable::class.java)) {
+            val ctor = constructorForDeserialization(beanDesc.beanClass)
+            if (ctor != null) {
```
fenryka: combine with above with all that good Kotlin syntactic sugar

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+@ToStringSerialize
+@JsonDeserialize(using = NetworkHostAndPortDeserializer::class)
+private interface NetworkHostAndPortMixin
+
+private class NetworkHostAndPortDeserializer : JsonDeserializer<NetworkHostAndPort>() {
+    override fun deserialize(parser: JsonParser, ctxt: DeserializationContext): NetworkHostAndPort {
+        return NetworkHostAndPort.parse(parser.text)
```
fenryka: reduce to des(...) = ...

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
```diff
@@ -0,0 +1,195 @@
...
+@JsonSerialize(using = PartyAndCertificateSerializer::class)
+// TODO Add deserialization which follows the same lookup logic as Party
+private interface PartyAndCertificateSerializerMixin
+
+private class PartyAndCertificateSerializer : JsonSerializer<PartyAndCertificate>() {
+    override fun serialize(value: PartyAndCertificate, gen: JsonGenerator, serializers: SerializerProvider) {
+        gen.jsonObject {
+            writeObjectField("name", value.name)
+            writeObjectField("owningKey", value.owningKey)
+            // TODO Add configurable option to output the certPath
```
fenryka: Log Jira to track this

core/src/main/kotlin/net/corda/core/internal/SerializationScanningHelper.kt
```diff
@@ -0,0 +1,344 @@
...
+}
+
+/**
+ * Encapsulates the property of a class and its potential getter and setter methods.
+ *
+ * @property field a property of a class.
+ * @property setter the method of a class that sets the field. Determined by locating
+ * a function called setXyz on the class for the property named in field as xyz.
+ * @property getter the method of a class that returns a fields value. Determined by
+ * locating a function named getXyz for the property named in field as xyz.
```
fenryka: iser not documented in kdoc

core/src/main/kotlin/net/corda/core/internal/SerializationScanningHelper.kt
```diff
@@ -0,0 +1,344 @@
+package net.corda.core.internal
```
fenryka: If this lives in core then it should follow the other modules and be in internal/serialization/...

fenryka: equally, I'd rather it wasn't in core at all, given we already moved all / most of the serialization code out of core before.

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/PropertySerializers.kt
```diff
@@ -104,16 +104,16 @@ class EvolutionPropertyReader : PropertyReader() {
...
  * @property initialPosition where in the constructor used for serialization the property occurs.
- * @property getter a [PropertySerializer] wrapping access to the property. This will either be a
+ * @property serializer a [PropertySerializer] wrapping access to the property. This will either be a
  * method invocation on the getter or, if not publicly accessible, reflection based by temporally
  * making the property accessible.
  */
 abstract class PropertyAccessor(
         val initialPosition: Int,
-        open val getter: PropertySerializer) {
+        open val serializer: PropertySerializer) {
```
fenryka: it was called getter because semantically it's only ever used to retrieve the value, not actually serialize the object... so not sure the rename makes 100% sense (but neither am I totally aginast it, is serializer the best descritptor)

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
shamsasari: I'll be adding this in my blob inspector PR when I wire it to JacksonSupport

core/src/main/kotlin/net/corda/core/internal/SerializationScanningHelper.kt
shamsasari: I just moved this as is from node-api. I can add something in if you tell me what it does ;-)

node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/PropertySerializers.kt
shamsasari: This confused me when I looked at `PropertyAccessor` Comparator. I thought the reference to "getter" in there meant a `Method`.

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
shamsasari: How do you mean? I don't think there is a syntactic sugar applicable here.

core/src/main/kotlin/net/corda/core/internal/SerializationScanningHelper.kt
shamsasari: It's only the property scanning code. Everything else is still in node-api. What do you suggest?

shamsasari:
Comments addressed

shamsasari:
We've decided to extract out the serializaation logic from node-api and depend on that, rather than moving some of the code to core: https://github.com/corda/corda/pull/3179

PR #3171, opened 2018-05-16
Mark HA notary design as deprecated

thschroeter:
This design is superseded by the CFT notary design.

Alternatively, we could introduce an `archive` for old design docs and move it there.


mikehearn:
oh ok - maybe just delete it entirely then.

PR #3173, opened 2018-05-16
ENT-1574 Progress tracker distinguishes steps which were not processed

szymonsztuka:
* Progress Tracker doesn't set to "green tick" for a step which hasn't been set in the flow 
It marks steps as processed ("green tick") only if they were send to a renderer (set as the current step). Marking with green crossed sign for shell and as minus sign for ssh.  
See attached screenshots.

* Finality Flow sets the current step to NOTARISING only if the actual notarization is performed.

The underlying fix was for ProgressTracker descends to child steps to find stepIndexTree for more than two nested levels.
Added first a unit test which illustrates the problem , and it failed without a fix.
After the fix, the test passes:
https://ci-master.corda.r3cev.com/viewType.html?buildTypeId=Corda_PullRequests&branch_Corda_Build_PullRequests=3173&tab=buildTypeStatusDiv


szymonsztuka:
screen shot 2018-05-17 at 10 23 45 2
screen shot 2018-05-16 at 16 57 07

PR #3174, opened 2018-05-16
Bootstrap HA and BFT notaries

thschroeter:
Extending the bootstrapper to generate networks with HA and BFT notaries, based on the `notary.serviceLegalName` configuration.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -1,8 +1,10 @@
 package net.corda.nodeapi.internal.network
 
+import com.typesafe.config.Config
 import com.typesafe.config.ConfigFactory
 import net.corda.cordform.CordformNode
 import net.corda.core.contracts.ContractClassName
+import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.internal.*
 import net.corda.core.internal.concurrent.fork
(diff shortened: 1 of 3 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -61,13 +60,48 @@ class NetworkBootstrapper {
...
+    sealed class NotaryCluster {
+        data class BFT(val name: CordaX500Name) : NotaryCluster()
+        data class CFT(val name: CordaX500Name) : NotaryCluster()
+    }
+
+    data class DirectoryAndConfig(val directory: Path, val config: Config)
+
+    private fun notaryClusters(nodeDirs: List<Path>): Map<NotaryCluster, List<Path>> {
+        val clusteredNotaries = nodeDirs.flatMap { d ->
+           val c = ConfigFactory.parseFile((d / "node.conf").toFile())
```
shamsasari: We read and parse the node.conf files again in `gatherNotaryInfos`. Can you have a pre-step that goes through the dirs and parse the conf files and returns a `Map<Path, Config>`.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -61,13 +60,50 @@ class NetworkBootstrapper {
...
+    }
+
     fun bootstrap(directory: Path, cordappJars: List<Path>) {
         directory.createDirectories()
         println("Bootstrapping local network in $directory")
         generateDirectoriesIfNeeded(directory, cordappJars)
         val nodeDirs = directory.list { paths -> paths.filter { (it / "corda.jar").exists() }.toList() }
         require(nodeDirs.isNotEmpty()) { "No nodes found" }
         println("Nodes found in the following sub-directories: ${nodeDirs.map { it.fileName }}")
+        val configs = nodeDirs.map {
```
shamsasari: You could use `associateBy`

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -61,13 +60,50 @@ class NetworkBootstrapper {
...
+            it to ConfigFactory.parseFile((it / "node.conf").toFile())
+        }.toMap()
+        notaryClusters(configs).forEach { (cluster, directories) ->
+            when (cluster) {
```
shamsasari: Can you move the logic in this forEach into notaryClusters

PR #3177, opened 2018-05-17
ENT-1902: Explorer GUI from Enterprise cannot be started against remote Node

vkolomeyko:
If we are already in the JavaFX thread - perform direct invocation without postponing it.
Or else if we are operating in JavaFX thread the sequence of invocations will change.


vkolomeyko:
Gentle reminder to review and approve this PR.

PR #3180, opened 2018-05-17
Network bootstrapper check duplicate node names

stanly-johnson:

# PR Checklist:

- [x] Have you run the unit, integration and smoke tests as described here? https://docs.corda.net/head/testing.html
- [x] If you are contributing for the first time, please read the agreement in CONTRIBUTING.md now and add to this Pull Request that you agree to it.

I certify that my contribution is in accordance with the terms of http://developercertificate.org/

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -23,20 +23,22 @@ import net.corda.nodeapi.internal.ContractsJarFile
 import net.corda.nodeapi.internal.DEV_ROOT_CA
 import net.corda.nodeapi.internal.SignedNodeInfo
 import net.corda.nodeapi.internal.network.NodeInfoFilesCopier.Companion.NODE_INFO_FILE_NAME_PREFIX
-import net.corda.nodeapi.internal.serialization.CordaSerializationMagic
 import net.corda.nodeapi.internal.serialization.AMQP_P2P_CONTEXT
+import net.corda.nodeapi.internal.serialization.CordaSerializationMagic
 import net.corda.nodeapi.internal.serialization.SerializationFactoryImpl
-import net.corda.nodeapi.internal.serialization.amqp.AbstractAMQPSerializationScheme
-import net.corda.nodeapi.internal.serialization.amqp.amqpMagic
+import net.corda.nodeapi.internal.serialization.amqp.AMQPServerSerializationScheme
 import net.corda.nodeapi.internal.serialization.kryo.AbstractKryoSerializationScheme
 import net.corda.nodeapi.internal.serialization.kryo.kryoMagic
-import java.nio.file.Path
-import java.nio.file.Paths
+import java.nio.file.*
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import java.time.Instant
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeoutException
 import kotlin.streams.toList
+import kotlin.collections.HashSet
+import kotlin.collections.component1
+import kotlin.collections.component2
+import kotlin.collections.set
 
 /**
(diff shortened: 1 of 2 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -92,6 +98,23 @@ class NetworkBootstrapper {
...
 
+    /*the function checks for duplicate myLegalName in the all the *_node.conf files
+    All the myLegalName values are added to a HashSet - this helps detect duplicate values.
+    If a duplicate name is found the process is aborted with an error message
+    */
+    private fun checkForDuplicateLegalNames(directory: Path, cordappJars: List<Path>) {
+      val legalNames = HashSet<String>()
+      val confFiles = directory.list { it.filter { it.toString().endsWith("_node.conf") }.toList() }
+      for (confFile in confFiles) {
+        val legalName = confFile.readLines { it.findFirst().get() }
```
shamsasari: You're assuming the legal name entry is the first line in the conf file - that's not guaranteed to be the case. Instead parse the conf file and extract out the "myLegalName" option. Have a look at `gatherNotaryInfos` in this class.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -92,6 +98,23 @@ class NetworkBootstrapper {
         }
     }
 
+    /*the function checks for duplicate myLegalName in the all the *_node.conf files
+    All the myLegalName values are added to a HashSet - this helps detect duplicate values.
+    If a duplicate name is found the process is aborted with an error message
+    */
+    private fun checkForDuplicateLegalNames(directory: Path, cordappJars: List<Path>) {
+      val legalNames = HashSet<String>()
+      val confFiles = directory.list { it.filter { it.toString().endsWith("_node.conf") }.toList() }
```
shamsasari: The bootstrapper can work in two modes, either on a list of _node.conf files, like here, or on an existing set of node directories with the `node.conf` file in each dir. We'd want to have duplication check for both scenerios. The easy way to do that is to simply move when this method is called to after `generateDirectoriesIfNeeded` is called and look up the `node.conf` file. Again have a look at `gatherNotaryInfos` which does the same thing.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -92,6 +98,23 @@ class NetworkBootstrapper {
...
+      for (confFile in confFiles) {
+        val legalName = confFile.readLines { it.findFirst().get() }
+        if(!legalNames.add(legalName)){
+          println("Duplicate Node Found - ensure every node has a unique legal name");
+          throw Error("Duplicate Node Found");
```
shamsasari: `IllegalArgumentException` would be more correct. `Error` is for more abnormal issues that typically are not expected to occur.

Also include the name of the duplicate to aid the user.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -293,13 +316,4 @@ class NetworkBootstrapper {
         override fun rpcClientKryoPool(context: SerializationContext) = throw UnsupportedOperationException()
         override fun rpcServerKryoPool(context: SerializationContext) = throw UnsupportedOperationException()
     }
-
-    private object AMQPParametersSerializationScheme : AbstractAMQPSerializationScheme(emptyList()) {
```
shamsasari: Hmm, I wonder why we created a custom one here? Thanks for that.

stanly-johnson:
@shamsasari changes done in #3182

shamsasari:
Thanks for that. It's better to stick to the same PR when making changes, but that's OK for now. I'll close this.

PR #3183, opened 2018-05-17
ENT-933 SMM unit test

andr3ej:


core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -251,7 +251,7 @@ fun <T : Any> T.serialize(serializationFactory: SerializationFactory = Serializa
  * to get the original object back.
  */
 @Suppress("unused")
-class SerializedBytes<T : Any>(bytes: ByteArray) : OpaqueBytes(bytes) {
+class SerializedBytes<out T : Any>(bytes: ByteArray) : OpaqueBytes(bytes) {
```
andr3ej: this eliminates a cast to FlowLogic<*> in SMM

node/src/test/kotlin/net/corda/node/services/statemachine/StateMachineManagerHarness.kt
```diff
@@ -0,0 +1,172 @@
...
+            secureRandom,
+            { serialization },
+            spectator(),
+            { localSessions.removeAt(0) },
+            { flowMessaging })
+
+    init {
+        smm.start(participant())
+        onNodeReady.get()(participant())
+        verify(flowMessaging).start(handler.get())
```
andr3ej: so that verifyNoMoreInteractions succeeds at the end

andr3ej:
the enterprise merge of this will be non-trivial due to parallel changes needed to MultiThreadedStateMachineManager, so merge into open source may have to wait until i'm back from holiday on the 29th

node/src/test/kotlin/net/corda/node/services/statemachine/StateMachineManagerUnitTest.kt
```diff
@@ -0,0 +1,86 @@
...
+
+class StateMachineManagerUnitTest : StateMachineManagerHarness() {
+    @Test
+    fun `trivial flow`() {
+        assertEquals(135, TrivialLogic(135).spawn().getOrThrow())
+    }
+
+    @Test
+    fun `just receive`() {
+        Channel().run {
```
rick-r3: Should `Channel` be `Session` perhaps ?

andr3ej: that is the underlying flow framework mechanism but channel makes more sense to me when reading the high level code, as it's a bidirectional communication path between 2 parties whereas logically session could mean lots of things

andr3ej:
a) i'm guessing it's in now, as i've had to resolve some conflicts?
b) to begin with i want to reproduce the problem, in the form of a test or tests that then become acceptance criteria. i can probably do that now but at significant cost to the build time, and it won't be practical to have full rigour i.e. check all the scenarios i think of. this PR is about getting the risky question of "is it even possible for the SMM be unit tested?" out of the way, i haven't yet though about the details of how i'm going to approach the problem. it's likely that the harness will change quite a bit to accommodate real tests instead of the current demos.

mnesbit:
Guess this isn't relevant anymore

PR #3187, opened 2018-05-18
CORDA-1506: Cash selection logic fails when selection with a change from more that 2 different issuer groups

vkolomeyko:
Description will change in the future and Jira item to follow.

PR #3194, opened 2018-05-18
CORDA-1506: Optimize number of states participating in the transaction.

vkolomeyko:
This is dependant on #3187 fix merged first.

### DIFF
--- a/finance/src/main/kotlin/net/corda/finance/contracts/asset/cash/selection/AbstractCashSelection.kt
+++ b/finance/src/main/kotlin/net/corda/finance/contracts/asset/cash/selection/AbstractCashSelection.kt
@@ -151,16 +151,22 @@ abstract class AbstractCashSelection {
                         stateAndRefs.addAll(uncheckedCast(services.loadStates(stateRefs)))
                     }
 
-                    val success = stateAndRefs.isNotEmpty() && totalPennies >= amount.quantity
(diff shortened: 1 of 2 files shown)


szymonsztuka:
The current implementation would gather small coins and merge them as one bigger output which has advantages.
With the fix we would discard small coins whenever a big one can be found and spent. 

E.g. currently we need to spent 100 and we were 'unlucky' to select (in that order)  99 x £1 + £100,  and we spent 99x£1 + £1 + change £99. As the outputs we have £100 and £99 have so from 100 coins down to 2 coins in the system.  
The sub optimal selection does merge smaller coins - kind of self-cleanup which might be desirable.

vkolomeyko:
@szymonsztuka - I agree it is down to preferences.

"discard small coins" - could be one preference.
Another preference is to improve performance by not creating network traffic or producing additional DB records when we can avoid this.
This PR improvement is aiming at the latter.

This is very much up for debate what is best and of course down to usage pattern in terms of amount issued and paid.
For unit test modified it is definitely worthwhile thing as it is reducing the number of inputs involved.

@rick-r3 - could you opine on that?

rick-r3:
Can we not do this?  Coin selection and soft locking do not work in multithreaded environment and that is a much bigger problem.  The priority there is trying to avoid flows trampling on the same coins which means creating less, larger coins is an anti-pattern because it increases contention.  Potentially.

PR #3122, opened 2018-05-11
[CORDA-1269]: Changed usages of `arrayOf()` in Entities to use Kotlin 1.2 style.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1269

PR #3198, opened 2018-05-18
CORDA-1269 - Changed usages of `arrayOf()` in Entities to use Kotlin 1.2 style.

fenryka:
(#3122)


fenryka:
So this will never work for V3 as the 1.2 work wasn' bakcported so

PR #3200, opened 2018-05-21
Experimental utility code

tomtau:
As discussed with @mikehearn on the corda-dev mailing list, we wrote some generic utility code that may be useful to other Corda developers, could potentially be combined with other experimental code (e.g. @mikehearn mentioned @roastario was working on some experimental code which auto-generates boilerplate when working with linear states that evolve in predictable ways) etc., so we decided to open source it.


mikehearn:
Oh sorry, Joel just told me you signed it before. I'll merge this.

PR #3202, opened 2018-05-21
CORDA-1499 - Fix a typo in node_attchments_contracts table name.

szymonsztuka:
Database table is missing an "a" in the word attachments: `node_attchments_contracts`.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -110,6 +110,8 @@ Unreleased
   For instance, this method will check if an ECC key lies on a valid curve or if an RSA key is >= 2048bits. This might
   be required for extra key validation checks, e.g., for Doorman to check that a CSR key meets the minimum security requirements.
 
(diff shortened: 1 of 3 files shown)


mikehearn:
Can we add a check if the old name exists and if so, print out an error and refuse to start (with instructions on how to bring things back)?

If you create an `--preflight-upgrade-check` flag that runs such checks without touching anything, so it can be run at the same time the old version of the node is still running, I'll be even more impressed. We can use it as a place to put similar checks in future.

szymonsztuka:
Created a separate ticket https://r3-cev.atlassian.net/browse/CORDA-1554 for db tables validation/ dry run.

josecoll:
A moment of deja vu ... I'm sure I saw this fix ages ago ???

szymonsztuka:
I'm not aware of the former fix, maybe it was for a different table.

szymonsztuka:
Yes, Liquibase ChangeSet  in Enterprise will be added.

PR #3209, opened 2018-05-22
Reverting API breaking change in PermissionException.

sollecitom:


node-api/src/main/kotlin/net/corda/nodeapi/exceptions/RpcExceptions.kt
```diff
@@ -46,4 +47,11 @@ class OutdatedNetworkParameterHashException(old: SecureHash, new: SecureHash) :
...
  * Thrown to indicate that the command was rejected by the node, typically due to a special temporary mode.
  */
-class RejectedCommandException(message: String) : CordaRuntimeException(message), ClientRelevantError
\ No newline at end of file
+class RejectedCommandException(message: String) : CordaRuntimeException(message), ClientRelevantError
+
+/**
+ * Allows an implementing [Throwable] to be propagated to RPC clients.
+ */
+@Deprecated("Use ClientRelevantError istead.", replaceWith = ReplaceWith("ClientRelevantError"))
```
mikehearn: Typo: instead

sollecitom: Fixed, thanks! :)

PR #3216, opened 2018-05-22
CORDA-1498: serialization multiple transform bug

anthonykeenan:
When validating transforms in the serialization framework, the cyclic type check was being triggered when the same value appeared in a from and a to of a `CordaSerializationTransformRename`

For example, this code was failling the cyclic dependency check when it is a simple evolution of a paramater from A -> B -> C
```
    @CordaSerializationTransformRenames(
            CordaSerializationTransformRename(from = "A", to = "B"),
            CordaSerializationTransformRename(from = "B", to = "C")
    )
    enum class AcceptMultipleRename { C }
```

I've rewritten the check to put the dependencies into a graph, checking each property both depends on and is dependend on by no more than one other property.
Then I'm checking that the properties of the current evolution are at the end of a renaming chain (if they're in one)
Finally I'm iterating through all the nodes dependencies to see whether there are any loops.

It seems like the code was quite simple and it's suddenly got quite complicated - I couldn't think of a simpler way of doing it but I'm more than happy to be proved wrong.


serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -81,21 +81,62 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
+            }
+
+            val graph = mutableListOf<Node>()
+            // Keep a list of forward links and back links in order to build the graph in one pass
+            val forwardLinks = hashMapOf<String, Int>()
+            val reverseLinks = hashMapOf<String, Int>()
+
+            // build a dependency graph
+            @Suppress("UNCHECKED_CAST") (list as List<RenameSchemaTransform>).forEachIndexed { index, rename ->
+                forwardLinks[rename.from]?.let { throw NotSerializableException("There are multiple transformations from ${rename.from}, which is not allowed") }
```
fenryka: Probably want the actual typename in here somehow otherwise the error message isn't going to be very obvious

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -81,21 +81,62 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
+                if (!node.visited) {
+                    // Find an unvisited node
+                    var currentNode = node
+                    currentNode.visit(index)
+                    while (currentNode.next != null) {
+                        currentNode = graph[node.next!!]
+                        if (currentNode.visited) {
+                            if (currentNode.visitedBy == index) {
+                                // we have gone round in a loop
+                                throw NotSerializableException("Cyclic renames are not allowed (${currentNode.from})")
```
fenryka: Again, error message will be ambiguous

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -81,21 +81,61 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
         override fun validate(list: List<Transform>, constants: Map<String, Int>) {
-            object : Any() {
-                val from: MutableSet<String> = mutableSetOf()
-                val to: MutableSet<String> = mutableSetOf()
-            }.apply {
-                @Suppress("UNCHECKED_CAST") (list as List<RenameSchemaTransform>).forEach { rename ->
-                    if (rename.to in this.to || rename.from in this.from) {
-                        throw NotSerializableException("Cyclic renames are not allowed (${rename.to})")
-                    }
+            data class Node(val from: String, val to: String, var next: Int?, var prev: Int?, var visited: Boolean = false, var visitedBy: Int? = null) {
```
shamsasari: The extensive use of indexes is making this code hard to understand. Can `next` and `prev` not be direct references to `RenameSchemaTransform` instances?

anthonykeenan: I've made them direct node references, I agree it makes it a lot easier to read.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -81,21 +81,61 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
+                forwardLinks[rename.from] = node
+                reverseLinks[rename.to] = node
+            }
 
-                    this.to.add(rename.from)
-                    this.from.add(rename.to)
+            // Check that every property in the current type is at the end of a renaming chain, if it is in one
+            constants.keys.forEach { key ->
+                reverseLinks[key]?.let {
+                    it.next?.let { throw NotSerializableException("$key is specified as a previously evolved type, but it also exists in the current type") }
```
shamsasari: `if (reverseLinks[key]?.next != null) throw ... `

In fact you have this pattern with `NotSerializableException` several times so why not have a little helper function like `check`/`require`? I would imagine it would benefit the rest of the serialisation code as well, but that can be for later I suppose.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -81,21 +81,61 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
+
+            val graph = mutableListOf<Node>()
+            // Keep two maps of forward links and back links in order to build the graph in one pass
+            val forwardLinks = hashMapOf<String, Node>()
+            val reverseLinks = hashMapOf<String, Node>()
+
+            // build a dependency graph
+            val transforms: List<RenameSchemaTransform> = uncheckedCast(list)
+            transforms.forEach { rename ->
+                forwardLinks[rename.from]?.let { throw NotSerializableException("There are multiple transformations from ${rename.from}, which is not allowed") }
```
shamsasari: You're not making use of the `it` in the `let` so a normal `if` statement would be clearer.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -81,21 +81,61 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
+                reverseLinks[rename.to]?.let { throw NotSerializableException("There are multiple transformations to ${rename.to}, which is not allowed") }
+                val node = Node(rename.from, rename.to, forwardLinks[rename.to], reverseLinks[rename.from])
```
shamsasari: Would it be clearer if the first two params where collapsed into `rename`?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -81,21 +81,61 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
-                val from: MutableSet<String> = mutableSetOf()
-                val to: MutableSet<String> = mutableSetOf()
-            }.apply {
-                @Suppress("UNCHECKED_CAST") (list as List<RenameSchemaTransform>).forEach { rename ->
-                    if (rename.to in this.to || rename.from in this.from) {
-                        throw NotSerializableException("Cyclic renames are not allowed (${rename.to})")
-                    }
+            data class Node(val from: String, val to: String, var next: Node?, var prev: Node?, var visited: Boolean = false, var visitedBy: Node? = null) {
+                fun visit(visitedBy: Node) {
+                    visited = true
```
shamsasari: `visited` could just be a helper getter: `visitedBy != null`

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -131,5 +151,11 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
         }
 
         override fun getTypeClass(): Class<*> = TransformTypes::class.java
+
+        protected fun require(expr: Boolean, errorMessage: String) {
```
shamsasari: Probably want to use a different name to avoid confusion with the Kotlin stdlib `require`.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -131,5 +151,11 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
         }
 
         override fun getTypeClass(): Class<*> = TransformTypes::class.java
+
+        protected fun requireThat(expr: Boolean, errorMessage: String) {
```
shamsasari: The `errorMessage` param should instead be a lambda and this method inlined to avoid unnecessarily creating the exception message.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -45,25 +45,12 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
...
-                if (it.old == it.new) {
-                    throw NotSerializableException("Enum extension ${it.new} cannot default to itself")
-                }
-
-                if (constants[it.old]!! >= constants[it.new]!!) {
-                    throw NotSerializableException(
-                            "Enum extensions must default to older constants. ${it.new}[${constants[it.new]}] " +
-                                    "defaults to ${it.old}[${constants[it.old]}] which is greater")
-                }
+                requireThat(constants.contains(it.new)) {"Unknown enum constant ${it.new}"}
```
fenryka: Stuff like this makes it neater, but it's going to make backporting issues a RPITA - why are we randomly changing things that aren't broken?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -131,5 +151,11 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
         }
 
         override fun getTypeClass(): Class<*> = TransformTypes::class.java
+
+        protected inline fun requireThat(expr: Boolean, errorMessage: () -> String) {
+            if (!expr) {
+                throw NotSerializableException(errorMessage())
```
fenryka: Technically a NotSerializableException should *just* contain the class name, Not that we've done that all over hte shop but we might think about extending it if we're adding additional context

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformsSchema.kt
```diff
@@ -238,11 +240,19 @@ data class TransformsSchema(val types: Map<String, EnumMap<TransformTypes, Mutab
...
-                if (isNotEmpty()) {
-                    map[type] = this
+            try {
+                get(type, sf).apply {
+                    if (isNotEmpty()) {
+                        map[type] = this
+                    }
                 }
             }
+            catch (e: NotSerializableException ) {
```
fenryka: catch on same line as closing bracket

fenryka: also random whitespace after exception

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformsSchema.kt
```diff
@@ -238,11 +240,19 @@ data class TransformsSchema(val types: Map<String, EnumMap<TransformTypes, Mutab
...
+                val message = "Error running transforms for $type: ${e.message}"
+                logger.error(message)
+                logger.trace(e)
```
fenryka: use trace { } call

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
anthonykeenan: This bug is on the backport list anyway so that shouldn't be an issue?
I changed it at @shamsasari request and I'm not a fan of refactoring for refactoring's sake - but I actually think this makes the function that I rewrote a lot more readable. But I can change it back if you really hate it.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
anthonykeenan: So I've added a `NotSerializableDetailedException` that takes a classname and a reason (happy to change the name, my mind was drawing a blank for creative exception names) - but even though the NotSerializableException asks for a classname, it actually stores that in the `message` field of the base class, so it is slightly confusing.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
fenryka: This makes no sense, why are we throwing IO Exceptions when it isn't serializable? that's what a not serializable exception is for?

If anything it should throw a more derived instance of that

fenryka: Right, having discussed this on slack, we need to change the IOexcpetion here to throw a custom exceptionm type that can excapsualte the intention of the change... basically doing things this way is correct, but it doesn't read well and I think in 6 months time we're going to be confsued why we did it and waste time rediscovering it

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TransformTypes.kt
```diff
@@ -131,5 +151,11 @@ enum class TransformTypes(val build: (Annotation) -> Transform) : DescribedType
         }
 
         override fun getTypeClass(): Class<*> = TransformTypes::class.java
+
+        protected inline fun requireThat(expr: Boolean, errorMessage: () -> String) {
+            if (!expr) {
+                throw IOException(errorMessage())
```
fenryka: As discussed above , let smake this a derived exception type that can express our tintention

serialization/src/test/kotlin/net/corda/serialization/internal/amqp/EnumEvolvabilityTests.kt
```diff
@@ -442,6 +442,67 @@ class EnumEvolvabilityTests {
...
+
+    @Test
+    fun rejectMultipleRenameTo() {
+        data class C(val e: RejectMultipleRenameTo)
+
+        val sf = testDefaultFactory()
+        assertThatThrownBy {
+            SerializationOutput(sf).serialize(C(RejectMultipleRenameTo.A))
+        }.isInstanceOf(NotSerializableException::class.java)
+        .hasToString("Unable to serialize/deserialize net.corda.serialization.internal.amqp.EnumEvolvabilityTests\$RejectMultipleRenameTo: " +
```
fenryka: feels like embedding the class name in the string is asking for a pain if we ever move it, could probably derive it programatically, future us would be thankful

PR #3222, opened 2018-05-23
[CORDA-1526]: Ensure MDC is printed in logs only if not empty.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1526

### DIFF
--- a/config/dev/log4j2.xml
+++ b/config/dev/log4j2.xml
@@ -13,7 +13,7 @@
 
     <Appenders>
         <Console name="Console-Appender" target="SYSTEM_OUT">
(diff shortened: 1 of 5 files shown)


config/dev/log4j2.xml
```diff
@@ -13,7 +13,7 @@
 
     <Appenders>
         <Console name="Console-Appender" target="SYSTEM_OUT">
-            <PatternLayout pattern="%highlight{%level{length=1} %date{HH:mm:ssZ} [%t] %c{2}.%method - %msg %X%n}{INFO=white,WARN=red,FATAL=bright red}" />
```
tlil: Do we only want this for the console appender? What about the rolling file appender further down?

tlil: Ah, right, I missed the following from the PR description, sorry:
>Not that I left the "{}" there for files, with ease of parsing in mind.

PR #3230, opened 2018-05-24
Remove default value for `myLegalName`

thschroeter:
When the config files doesn't contain `myLegalName` the exception says:

> improperly specified input name: Vast Global MegaCorp, Ltd


node/src/main/resources/reference.conf
```diff
@@ -1,4 +1,3 @@
-myLegalName = "Vast Global MegaCorp, Ltd"
```
blsemo: What happens when no legal name is specified at all? Hopefully a really useful error message.

thschroeter: Yes, `No configuration setting found for key 'myLegalName'`.

PR #3231, opened 2018-05-24
[CORDA-1638]: Audit log of failed transactions MVP.

sollecitom:
https://r3-cev.atlassian.net/browse/ENT-1638

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
@@ -4,6 +4,7 @@ import co.paralleluniverse.fibers.Suspendable
 import net.corda.core.crypto.isFulfilledBy
 import net.corda.core.identity.Party
 import net.corda.core.identity.groupAbstractPartyByWellKnownParty
+import net.corda.core.internal.info
 import net.corda.core.transactions.LedgerTransaction
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.utilities.ProgressTracker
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
```diff
@@ -51,17 +52,24 @@ class FinalityFlow(val transaction: SignedTransaction,
...
         progressTracker.currentStep = BROADCASTING
-        for (party in parties) {
-            if (!serviceHub.myInfo.isLegalIdentity(party)) {
-                val session = initiateFlow(party)
-                subFlow(SendTransactionFlow(session, notarised))
-            }
+        val recipients = parties.filterNot(serviceHub.myInfo::isLegalIdentity)
+        logger.info("Broadcasting transaction to parties ${recipients.map { it.name }.joinToString(", ", "[", "]")}.")
+        for (party in recipients) {
+            logger.info("Sending transaction to party ${party.name}.")
```
anvabr: Probably unnecessary, we know who we are sending it to so if there is no 'received' entry it means that one has failed

core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
```diff
@@ -73,9 +81,12 @@ class FinalityFlow(val transaction: SignedTransaction,
             val notarySignatures = subFlow(NotaryFlow.Client(transaction))
             transaction + notarySignatures
         } else {
+            logger.info("No need to notarise this transaction.")
             transaction
         }
+        logger.info("Recording transaction locally.")
```
anvabr: Also probably not needed as we know what needs to happen and if it does not happen we know it did not.

core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
sollecitom: As discussed, good point, but this lowers the chance a flow might get killed in between, in which case this information would help.

core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
sollecitom: As above.

PR #3234, opened 2018-05-24
ENT-1933: make NetworkParameters serialization compatible

kasiastreich:
Waiting for Kat's serialization fix to be in, but ready to review.


node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -43,17 +45,17 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
-                readParametersUpdate(advertisedParametersHash, signedParametersFromFile.raw.hash).verifiedNetworkMapCert(trustRoot)
+                readParametersUpdate(advertisedParametersHash, signedParametersFromFile.raw.hash)
             }
         } else { // No compatibility zone configured. Node should proceed with parameters from file.
-            signedParametersFromFile?.verifiedNetworkMapCert(trustRoot) ?: throw IllegalArgumentException("Couldn't find network parameters file and compatibility zone wasn't configured/isn't reachable")
+            signedParametersFromFile ?: throw IllegalArgumentException("Couldn't find network parameters file and compatibility zone wasn't configured/isn't reachable")
         }
         logger.info("Loaded network parameters: $parameters")
-        return parameters
+        return Pair(parameters.verifiedNetworkMapCert(trustRoot), parameters.raw.hash)
```
fenryka: Data class rather than a pair would be more descripte

kasiastreich: It's only used in one place, i think it's overcomplicated to have data class for private one use.

PR #3238, opened 2018-05-25
CORDA-1524 Fix NodeSchedulerService

blsemo:
The NodeSchedulerService used to have a race condition where it could happen that two flows for the same state get started, leading to an attempted double spend.
This got cleared up by giving the scheduler an event loop on a worker thread, and doing all selection of states etc. on this single thread. The external calls now only trigger this thread to recheck what to execute rather than getting involved in any of the selection logic.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -71,6 +71,8 @@
       <module name="example-code_integrationTest" target="1.8" />
       <module name="example-code_main" target="1.8" />
       <module name="example-code_test" target="1.8" />
+      <module name="experimental-blobinspector_main" target="1.8" />
+      <module name="experimental-blobinspector_test" target="1.8" />
       <module name="experimental-kryo-hook_main" target="1.8" />
       <module name="experimental-kryo-hook_test" target="1.8" />
       <module name="experimental_main" target="1.8" />
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -71,8 +71,6 @@
       <module name="example-code_integrationTest" target="1.8" />
       <module name="example-code_main" target="1.8" />
       <module name="example-code_test" target="1.8" />
-      <module name="experimental-blobinspector_main" target="1.8" />
-      <module name="experimental-blobinspector_test" target="1.8" />
       <module name="experimental-kryo-hook_main" target="1.8" />
       <module name="experimental-kryo-hook_test" target="1.8" />
       <module name="experimental_main" target="1.8" />
--- a/node/src/main/kotlin/net/corda/node/services/events/NodeSchedulerService.kt
+++ b/node/src/main/kotlin/net/corda/node/services/events/NodeSchedulerService.kt
@@ -142,6 +142,7 @@ class NodeSchedulerService(private val clock: CordaClock,
     private class InnerState {
         var rescheduled: GuavaSettableFuture<Boolean>? = null
         var nextScheduledAction: ScheduledStateRef? = null
+        var running: Boolean = true
     }
 
     // Used to de-duplicate flow starts in case a flow is starting but the corresponding entry hasn't been removed yet
@@ -149,6 +150,7 @@ class NodeSchedulerService(private val clock: CordaClock,
     private val startingStateRefs = ConcurrentHashSet<ScheduledStateRef>()
 
     private val mutex = ThreadBox(InnerState())
+
     // We need the [StateMachineManager] to be constructed before this is called in case it schedules a flow.
     fun start() {
         schedulerTimerExecutor.execute{runLoopFunction()}
@@ -157,9 +159,10 @@ class NodeSchedulerService(private val clock: CordaClock,
     override fun scheduleStateActivity(action: ScheduledStateRef) {
         log.trace { "Schedule $action" }
         // Only increase the number of unfinished schedules if the state didn't already exist on the queue
-        val countUp = !schedulerRepo.merge(action)
+        if( !schedulerRepo.merge(action) ){
+            unfinishedSchedules.countUp()
+        }
         contextTransaction.onCommit {
-            if (countUp) unfinishedSchedules.countUp()
             mutex.locked {
                 if (action.scheduledAt < nextScheduledAction?.scheduledAt ?: Instant.MAX) {
                     // We are earliest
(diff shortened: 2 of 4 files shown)


.idea/compiler.xml
```diff
@@ -71,6 +71,8 @@
       <module name="example-code_integrationTest" target="1.8" />
       <module name="example-code_main" target="1.8" />
       <module name="example-code_test" target="1.8" />
+      <module name="experimental-blobinspector_main" target="1.8" />
```
rick-r3: Don't think this should be changed.

blsemo: Oh well, the joy of checking in `.idea/compiler.xml`

node/src/main/kotlin/net/corda/node/services/events/NodeSchedulerService.kt
```diff
@@ -141,26 +142,32 @@ class NodeSchedulerService(private val clock: CordaClock,
...
+    // any new state should trigger a reschedule immediately if nothing is scheduled, so I would not expect
+    // this to usually trigger anything.
+    private val idleWaitSeconds = 60.seconds
+
     // We need the [StateMachineManager] to be constructed before this is called in case it schedules a flow.
     fun start() {
-        mutex.locked {
-            rescheduleWakeUp()
-        }
+        schedulerTimerExecutor.execute{runLoopFunction()}
```
rick-r3: whitespace

node/src/main/kotlin/net/corda/node/services/events/NodeSchedulerService.kt
```diff
@@ -141,26 +142,32 @@ class NodeSchedulerService(private val clock: CordaClock,
...
     }
 
     override fun scheduleStateActivity(action: ScheduledStateRef) {
         log.trace { "Schedule $action" }
         // Only increase the number of unfinished schedules if the state didn't already exist on the queue
-        val countUp = !schedulerRepo.merge(action)
+        if( !schedulerRepo.merge(action) ){
```
rick-r3: whitespace

node/src/main/kotlin/net/corda/node/services/events/NodeSchedulerService.kt
```diff
@@ -186,42 +193,45 @@ class NodeSchedulerService(private val clock: CordaClock,
...
+            if (scheduledState != null) {
                 if (!awaitWithDeadline(clock, scheduledState.scheduledAt, ourRescheduledFuture)) {
                     log.trace { "Invoking as next $scheduledState" }
                     onTimeReached(scheduledState)
                 } else {
                     log.trace { "Rescheduled $scheduledState" }
                 }
             }
+            else {
+                awaitWithDeadline(clock, Instant.now() + idleWaitSeconds , ourRescheduledFuture )
```
rick-r3: Technically, `Instant.now()` should be `Clock.instant()` to work with test clocks.

node/src/main/kotlin/net/corda/node/services/events/NodeSchedulerService.kt
```diff
@@ -186,42 +193,45 @@ class NodeSchedulerService(private val clock: CordaClock,
...
-            nextScheduledAction = schedulerRepo.getLatest(startingStateRefs.size + 1).firstOrNull { !startingStateRefs.contains(it.second) }?.second
-            Pair(nextScheduledAction, rescheduled!!)
-        }
-        if (scheduledState != null) {
-            schedulerTimerExecutor.execute {
-                log.trace(schedulingAsNextFormat, scheduledState)
-                // This will block the scheduler single thread until the scheduled time (returns false) OR
-                // the Future is cancelled due to rescheduling (returns true).
+    private fun runLoopFunction(){
+        while (mutex.locked { running }){
```
rick-r3: We can eliminate the mutex locking here by making `running` `@Volatile`

blsemo: I quite like the mutex at this point - we need it anyway at the place where the flag is set (as it needs to be locked when we trigger a reschedule to pick up the shutdown), so locking it here makes sure we don't enter the loop while a shutdown is being triggered.

rick-r3: Okay.  Not too fussed.

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -200,7 +232,6 @@ class NodeSchedulerServiceTest : NodeSchedulerServiceTestBase() {
     fun `test activity due in the future and schedule another for same time`() {
         val eventA = schedule(mark + 1.days)
         val eventB = schedule(mark + 1.days)
-        assertWaitingFor(eventA)
```
rick-r3: Whether we are waiting for A or B is an implementation detail that no longer holds.

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -88,7 +120,7 @@ open class NodeSchedulerServiceTestBase {
 }
 
 class MockScheduledFlowRepository : ScheduledFlowRepository {
-    private val map = HashMap<StateRef, ScheduledStateRef>()
+    private val map = ConcurrentHashMap<StateRef, ScheduledStateRef>()
```
rick-r3: Concurrent updates mean we need concurrent map.

node/src/test/kotlin/net/corda/node/services/events/NodeSchedulerServiceTest.kt
```diff
@@ -68,16 +86,30 @@ open class NodeSchedulerServiceTestBase {
...
+        doAnswer {
+            traces += it.getArgument<ScheduledStateRef>(1)
+        }.whenever(it).trace(eq(NodeSchedulerService.schedulingAsNextFormat), any<Object>())
     }
 
-    protected fun assertWaitingFor(ssr: ScheduledStateRef, total: Int = 1) {
-        // The timeout is to make verify wait, which is necessary as we're racing the NSS thread i.e. we often get here just before the trace:
-        verify(log, timeout(5000).times(total)).trace(NodeSchedulerService.schedulingAsNextFormat, ssr)
+    protected fun assertWaitingFor(ssr: ScheduledStateRef) {
+        val endTime = System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(5)
```
rick-r3: This is slightly stricter and looks at what the last log entry is.  Also, the previous count of entries didn't make sense with the revised implementation of the scheduler.

PR #3239, opened 2018-05-25
[CORDA-1539] Minimum RSA key of 2048 bits on validatePublicKey

kchalkias:
Introduce a public function `Crypto.validatePublicKey(java.security.PublicKey)` to be called whenever a key validation is required (i.e, for CSRs to meet the security requirements). In case of an RSA key input, the minimum allowed key size is 2048 bits.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -96,6 +96,9 @@ Unreleased
 
 * Node can be shut down abruptly by ``shutdown`` function in `CordaRPCOps` or gracefully (draining flows first) through ``gracefulShutdown`` command from shell.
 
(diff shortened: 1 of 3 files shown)


docs/source/changelog.rst
```diff
@@ -102,6 +102,10 @@ Unreleased
   (even if JPA annotation nullable=false was absent).
   In case your Cordapps use this entity class to persist data in own custom tables as non Primary Key columns refer to :doc:`upgrade-notes` for upgrade instructions.
 
+* Adding a public method to check if a public key satisfies algorithm specs, `Crypto.validatePublicKey(java.security.PublicKey)`.
```
patrickkuo: maybe "check if a public key satisfies **Corda recommended** algorithm specs" ?

PR #3243, opened 2018-05-25
CORDA-1536: Fix client infinitely re-trying when incorrect endpoint specified.

vkolomeyko:
Apparently, if incorrect endpoint provided, ActiveMQNotConnectedException is thrown
which is different to ActiveMQSecurityException. Extend catch block to: ActiveMQException
to cater for all such cases.

### DIFF
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
@@ -21,7 +21,7 @@ import net.corda.core.transactions.SignedTransaction
 import net.corda.core.utilities.NetworkHostAndPort
 import net.corda.core.utilities.contextLogger
 import net.corda.core.utilities.seconds
-import org.apache.activemq.artemis.api.core.ActiveMQSecurityException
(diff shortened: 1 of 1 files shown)


client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -192,8 +192,11 @@ class NodeMonitorModel {
...
-            } catch(secEx: ActiveMQSecurityException) {
-                // Happens when incorrect credentials provided - no point to retry connecting.
+            } catch(secEx: ActiveMQException) {
+                // Happens when:
+                // * incorrect credentials provided;
+                // * incorrect endpoint specified;
+                // - no point to retry connecting.
                 throw secEx
             }
             catch(th: Throwable) {
```
mnesbit: Are there any wrapped exception causes we should consider here?

vkolomeyko: Nothing I can immediate think of.
Do we have any particular scenario in mind when a `Throwable` with nested `cause` is thrown to which we should react in a special way?

PR #3245, opened 2018-05-26
Resolve compiler warnings for Kotlin >= 1.2.40.

chrisr3:
Add `import` statements to resolve compiler warnings like:
```
w: /work/corda/finance/src/main/kotlin/net/corda/finance/flows/CashIssueFlow.kt: (49, 39): Access to this type by short name is deprecated, and soon is going to be removed. Please, add explicit qualifier or import
```

### DIFF
--- a/finance/src/main/kotlin/net/corda/finance/flows/CashPaymentFlow.kt
+++ b/finance/src/main/kotlin/net/corda/finance/flows/CashPaymentFlow.kt
@@ -11,6 +11,10 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.transactions.TransactionBuilder
 import net.corda.core.utilities.ProgressTracker
 import net.corda.finance.contracts.asset.Cash
+import net.corda.finance.flows.AbstractCashFlow.Companion.FINALISING_TX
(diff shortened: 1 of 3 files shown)


finance/src/main/kotlin/net/corda/finance/flows/CashPaymentFlow.kt
```diff
@@ -11,6 +11,10 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.transactions.TransactionBuilder
 import net.corda.core.utilities.ProgressTracker
 import net.corda.finance.contracts.asset.Cash
+import net.corda.finance.flows.AbstractCashFlow.Companion.FINALISING_TX
+import net.corda.finance.flows.AbstractCashFlow.Companion.GENERATING_ID
+import net.corda.finance.flows.AbstractCashFlow.Companion.GENERATING_TX
```
sollecitom: Are these now marked an unused an grey by Intellij? If so we should add a suppress.

chrisr3: If IntelliJ marks these imports as unused and grey then it means that you need to upgrade your Kotlin plugin.

PR #3251, opened 2018-05-29
Explicitly check the contractStateType param of the RPC vault queries…

shamsasari:
… is a ContractState class.

We lose the compile-time checks of the Class type parameter when invoking from the shell.


shamsasari:
@josecoll I've peeled @sollecitom's commit into another branch. Please approve.

PR #3254, opened 2018-05-29
[CORDA-1547]: Make log4j2.xml files more resilient to regex escaping problems.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1547


sollecitom:
@shamsasari , for reference, ported this here from your branch as discussed.

PR #3258, opened 2018-05-29
CORDA-1550 - Use gradle to generate the IntelliJ compiler.xml file

fenryka:
Rather than have some version of it checked in which keeps generating
conflicts every time something is changed when we only use it to
enforce the -parameters command line flag


build.gradle
```diff
@@ -304,6 +304,31 @@ task jacocoRootReport(type: org.gradle.testing.jacoco.tasks.JacocoReport) {
     }
 }
 
+// Since the Corda serialization framework depends on the -parameters flag being
+// used we set it for convenience by default within IntelliJ
+task writeIntellijCompilerXml() {
+    dependsOn = tasks.withType(AbstractCompile)
```
chrisr3: Firstly, you're writing Java compiler options and so `JavaCompile` is a better choice than `AbstractCompile`. But even so, I think you have this backwards. All `JavaCompile` tasks should depend on _this_ task, not vice versa.
```gradle
allprojects {
    ...
    tasks.withType(JavaCompile) {
        dependsOn writeIntellijCompilerXml
        options.compilerArgs << "-Xlint:unchecked" << "-Xlint:deprecation" << "-Xlint:-options" << "-parameters"
    }
    ...
}
```

build.gradle
```diff
@@ -304,6 +304,31 @@ task jacocoRootReport(type: org.gradle.testing.jacoco.tasks.JacocoReport) {
...
+<project version="4">
+  <component name="JavacSettings">
+    <option name="ADDITIONAL_OPTIONS_STRING" value="-parameters" />
+  </component>
+</project>"""
+    }
+
+    
+}
+
```
chrisr3: You can simplify this enormously because Gradle will help you, if you let it. By declaring what this task's output file is, you allow Gradle both to create target the directory (if it doesn't exist) and to notice when the file has/hasn't changed.
```gradle
task writeIntellijCompilerXml {
    def compilerXml = file("$rootDir/.idea/compiler.xml")
    outputs.files compilerXml

    doLast {
        println 'Generating compiler xml for IntelliJ'
        compilerXml.text = """<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="JavacSettings">
    <option name="ADDITIONAL_OPTIONS_STRING" value="-parameters" />
  </component>
</project>
"""
    }
}
```
This task also needs to be declared before the `allprojects` section so that it exists to be referenced by `JavaCompile`.

chrisr3:
I should probably also mention that IntelliJ and Gradle will fight like wildcats over your new `.idea/compiler.xml` file because it no longer includes the
```xml
<component name="CompilerConfiguration">
  <bytecodeTargetLevel target="1.8">
    <module name="api-scanner_main" target="1.8" />
    ...
  </bytecodeTargetLevel>
</component>    
```
fragment. Is there any way that you could get the `idea` plugin to generate it?

fenryka:
I'm not even sure this really works given you'd want it in place before IntelliJ read it and the order of operation there is unclear. Ideally, you'd really want to do this with a git hook... I'll have another think

mikehearn:
@fenryka any update on this?

PR #3271, opened 2018-05-30
Limit concurrency of the bootstrapper

thschroeter:
This allows users to bootstrap large Corda networks.

Once the bootstrapper has moved to the tools package we can
expose the number of processes as a command line parameter
(using Picocli).

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -61,7 +61,7 @@ class NetworkBootstrapper {
         fun main(args: Array<String>) {
             val baseNodeDirectory = requireNotNull(args.firstOrNull()) { "Expecting first argument which is the nodes' parent directory" }
             val cordappJars = if (args.size > 1) args.asList().drop(1).map { Paths.get(it) } else emptyList()
-            NetworkBootstrapper().bootstrap(Paths.get(baseNodeDirectory).toAbsolutePath().normalize(), cordappJars, Runtime.getRuntime().availableProcessors())
+            NetworkBootstrapper().bootstrap(Paths.get(baseNodeDirectory).toAbsolutePath().normalize(), cordappJars)
         }
     }
 
@@ -102,7 +102,7 @@ class NetworkBootstrapper {
         }
     }
 
-    fun bootstrap(directory: Path, cordappJars: List<Path>, numParallelProcesses: Int) {
+    fun bootstrap(directory: Path, cordappJars: List<Path>, numParallelProcesses: Int = Runtime.getRuntime().availableProcessors()) {
         directory.createDirectories()
         println("Bootstrapping local network in $directory")
         generateDirectoriesIfNeeded(directory, cordappJars)
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -19,7 +19,6 @@ import net.corda.core.serialization.deserialize
 import net.corda.core.serialization.internal.SerializationEnvironmentImpl
 import net.corda.core.serialization.internal._contextSerializationEnv
 import net.corda.core.utilities.days
-import net.corda.core.utilities.getOrThrow
 import net.corda.core.utilities.seconds
 import net.corda.nodeapi.internal.*
 import net.corda.nodeapi.internal.network.NodeInfoFilesCopier.Companion.NODE_INFO_FILE_NAME_PREFIX
@@ -32,12 +31,13 @@ import java.nio.file.Path
 import java.nio.file.Paths
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import java.time.Instant
+import java.util.*
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
-import java.util.concurrent.TimeoutException
 import kotlin.collections.component1
 import kotlin.collections.component2
 import kotlin.collections.set
+import kotlin.concurrent.schedule
 import kotlin.streams.toList
 
 /**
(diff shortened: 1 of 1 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -129,10 +131,36 @@ class NetworkBootstrapper {
...
+
+    private fun generateNodeInfo(nodeDir: Path): Path {
+        val logsDir = (nodeDir / LOGS_DIR_NAME).createDirectories()
+        val process = ProcessBuilder(nodeInfoGenCmd)
+                .directory(nodeDir.toFile())
+                .redirectErrorStream(true)
+                .redirectOutput((logsDir / "node-info-gen.log").toFile())
+                .apply { environment()["CAPSULE_CACHE_DIR"] = "../.cache" }
+                .start()
+        if (!process.waitFor(60, TimeUnit.SECONDS)) {
```
shamsasari: We need the feature that it warns the user if the brootstrapping is taking long. It was in response to users with slow machines.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -129,10 +131,36 @@ class NetworkBootstrapper {
...
+            process.destroyForcibly()
+            throw Error("Error while generating node info file. Please check the logs in $logsDir.")
```
shamsasari: `IllegalStateException`

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -129,10 +131,36 @@ class NetworkBootstrapper {
...
+        }
+        if (process.exitValue() != 0) {
+            throw Error("Error while generating node info file. Please check the logs in $logsDir.")
```
shamsasari: Use `check` instead

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -56,9 +58,10 @@ class NetworkBootstrapper {
 
         @JvmStatic
         fun main(args: Array<String>) {
+            // TODO: Use Picocli once the bootstrapper has moved into the tools package.
             val baseNodeDirectory = requireNotNull(args.firstOrNull()) { "Expecting first argument which is the nodes' parent directory" }
             val cordappJars = if (args.size > 1) args.asList().drop(1).map { Paths.get(it) } else emptyList()
-            NetworkBootstrapper().bootstrap(Paths.get(baseNodeDirectory).toAbsolutePath().normalize(), cordappJars)
+            NetworkBootstrapper().bootstrap(Paths.get(baseNodeDirectory).toAbsolutePath().normalize(), cordappJars, 4)
```
shamsasari: Default to `Runtime.availableProcessors`

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
thschroeter: What about printing "this may take a while on slow machines" upfront and then printing the progress along the way, e.g. "prepared node i of n"? Knowing the progress might be handy when bootstrapping large networks, that will take a few minutes.

Otherwise, I can implement the warning feature.

thschroeter:
@shamsasari the idea is to allow users to bootstrap large networks (with hundreds of nodes). I was doing a scalability test and couldn't use the bootstrapper because it started too many JVMs in parallel.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
thschroeter: Perhaps we should give each node info generation process two minutes instead of one, just in case?

shamsasari: I'd rather we keep the original feature. Printing "this might be slow" always isn't ideal - it should only show up for the slow machines.

shamsasari:
We don't we just hard-code the max number of threads to `availableProcessers`, or some multiple of it? For small networks it won't matter and for large ones, as you say it needs to be limited. I'm not sure of the benefit of having this as a cmd line flag.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
thschroeter: I've added the feature.

shamsasari: We don't need the timeout now right? The user can always Ctrl-C

thschroeter: I was thinking about running this in a CI environment. Should I remove the timeout or increase it to say 3 minutes?

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -56,9 +58,10 @@ class NetworkBootstrapper {
 
         @JvmStatic
         fun main(args: Array<String>) {
+            // TODO: Use Picocli once the bootstrapper has moved into the tools package.
```
shamsasari: I think it's overkill to have this a cmd line flag. What we have now is sufficient I think.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
shamsasari: Yeah 3 mins is fine.

PR #3273, opened 2018-05-30
Ent 1962 - better exception handling

bpaunescu:
It is possible for a flow to be initiated by an RPC client while the node is shutting down, specifically after SMM but before messaging service. In this case, the SMM will refuse via an IllegalArgumentException which is poor user experience. Solution is to catch and rethrow as RejectedCommandException.


sollecitom:
As discussed face to face, try-catching `IllegalArgumentException` is too broad. Throwing `RejectedCommandException` from call-site sounds safer.

bpaunescu:
Closing as OS doesn't need this fix. The SMM throws an `IllegalStateException` when starting a flow during shutdown already.

PR #3275, opened 2018-05-30
[CORDA-1557]: Remove "corda-" from 4.0 (and onwards) release version.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1557


mikehearn:
We could also look at using the axion-release Gradle plugin. It calculates version numbers from the last tag. That way we'd have fewer things to keep synchronised and GitHub releases would automatically change the version number perceived by Gradle.

PR #3277, opened 2018-05-31
CORDA-1349: Docs: improve docs on permissioning, doorman and network map.

mikehearn:


docs/source/permissioning.rst
```diff
@@ -61,31 +53,26 @@ We can visualise the permissioning structure as follows:
...
-* The certificates must follow the `X.509 standard <https://tools.ietf.org/html/rfc5280>`_
-
-   * We recommend X.509 v3 for forward compatibility
-
-* The TLS certificates must follow the `TLS v1.2 standard <https://tools.ietf.org/html/rfc5246>`_
-
-* The root network CA, doorman CA and node CA keys, as well as the node TLS
-  keys, must follow one of the following schemes:
-
+1. The certificates must follow the `X.509v3 standard <https://tools.ietf.org/html/rfc5280>`_
```
fenryka: I have memories of it being mentioned a double trailing underscore was better... but I could be imagining that

docs/source/permissioning.rst
```diff
@@ -61,31 +53,26 @@ We can visualise the permissioning structure as follows:
...
+2. The TLS certificates must follow the `TLS v1.2 standard <https://tools.ietf.org/html/rfc5246>`_
+3. The root network CA, doorman CA and node CA keys, as well as the node TLS keys, must follow one of the following schemes:
```
fenryka: will always express desire for an Oxford comma

docs/source/permissioning.rst
```diff
@@ -61,31 +53,26 @@ We can visualise the permissioning structure as follows:
...
-
     * ECDSA using the Koblitz k1 curve (secp256k1)
+    * RSA with 3072-bit key size or higher.
 
-    * RSA with 3072-bit key size
-
-.. note:: Corda's ``X509Utilities`` show how to generate the required public/private keypairs and certificates using
-   Bouncy Castle. You can find the ``X509Utilities`` in the `Corda repository <https://github.com/corda/corda>`_, under
-   ``/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt``.
+The certificates and keys will be automatically generated for you by the node on first run. But you can also generate
```
fenryka: However, you can also...

docs/source/permissioning.rst
```diff
@@ -97,98 +84,36 @@ The extension contains a single ASN.1 integer identifying the identity type the
...
+These two files should be in the node's certificate directory (``<workspace>/certificates/``), along with the network's
+own root certificates in a ``network-root-truststore.jks`` file.
 
 Connecting to a compatibility zone
 ----------------------------------
+
 To connect to a compatibility zone you need to register with their certificate signing authority (doorman) by submitting
-a certificate signing request (CSR) to obtain a valid identity for the zone.
+a certificate signing request (CSR) to obtain a valid identity for the zone. You could do this out of band, for instance
+via email or a web form, but there's also a simple request/response protocol built in to Corda.
```
fenryka: into not in to (I think)

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+become available for download and opt-in, and when they become enforced. This is again a policy decision you will
+probably choose to place some simple UI or workflow tooling around, in particular to enforce restrictions on who can
+edit the map or the parameters.
+
+Writing a doorman server
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+This step is optional because your users can obtain a signed certificate in many different ways. The doorman protocol
+is again a very simple HTTP based approach in which a node creates keys and requests a certificate, polling until it
+gets back what it expects. But you could also integrate this process with the rest of your signup process, for example,
```
fenryka: However, not but

fenryka: <full stop>For example,

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Zone parameters are stored in a file containing a Corda AMQP serialised ``SignedDataWithCert<NetworkParameters>`` object.
+It is easy to create such a file with a small Java or Kotlin program. The ``NetworkParameters`` object is a simple data
+holder that could be read from e.g. a config file, or settings from a database.
+
+Signing and saving the resulting file is just a few lines of code. A full example can be found in ``NetworkParametersCopier.kt`` in the
+source tree, but a flavour of it looks like this:
+
+.. sourcecode:: kotlin
```
fenryka: Should we provide a Java one as well. Know Richard is keen we lean more on that to avoid the impression you have to use kotling for Corda

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+Each individual parameter is documented in `the JavaDocs/KDocs for the NetworkParameters class <https://docs.corda.net/api/kotlin/corda/net.corda.core.node/-network-parameters/index.html>`_.
+
+The network map certificate is usually chained off the root certificate, and can be created according to the instructions above.
+
+Each time the zone parameters are changed, the epoch should be incremented. Epochs are essentially version numbers for
+the parameters, and they therefore cannot go backwards.
+
+Once saved, the new parameters can be served by the network map server.
+
+How to choose the parameters? This is the most complex question facing you as a new zone operator. Some settings may seem
```
fenryka: Feels like this might be a sub heading

docs/source/permissioning.rst
```diff
@@ -3,55 +3,47 @@ Network permissioning
...
 * ``sslkeystore.jks``, which stores the node’s TLS keypairs and certificates
 
-Production deployments require a secure certificate authority.
-Most production deployments will use an existing certificate authority or construct one using software that will be
-made available in the coming months. Until then, the documentation below can be used to create your own certificate
-authority.
-
-.. note:: If you are looking for information on how to connect to the existing compatibility zone go to the section: `Connecting to a compatibility zone`_
+Production deployments require a secure certificate authority. Most users will join an existing network such as the
+main Corda network or the Corda TestNet. You can also build your own networks. During development, no particular
```
gendal: i think we decided not to capitalise the N in Testnet.

docs/source/permissioning.rst
```diff
@@ -3,55 +3,47 @@ Network permissioning
...
+  certificates.
 
-* The **legal identity CAs**
+Each certificate has an X.509 extension in it that defines the certificate/key's role in the system (see below for details).
+They also use X.509 name constraints to ensure that the X.500 names that encode a human meaningful identity are propagated
+to all the child certificates properly. The following constraints are imposed:
 
-   * Node's well-known legal identity, apart from signing transactions, can also issue certificates for confidential legal identities
+* Doorman certificates are issued by a network root. Network root certs do not contain a role extension.
+* Node certificates are issued by an entity with a doorman certificate.
```
gendal: what does it mean to be an entity "with" a doorman certificate? The certificate is marked as having a doorman role? How is this manifested?

docs/source/permissioning.rst
```diff
@@ -3,55 +3,47 @@ Network permissioning
...
-  their own role flags on certificates, and it's important to verify that these are set consistently with the
-  certificate hierarchy design. As as side-effect this also acts as a secondary depth restriction on issued
-  certificates
-
-All the certificates must be issued with the custom role extension (see below).
+The structure of certificates above Doorman/Network map is intentionally left untouched, as they are not relevant to
+the identity service and therefore there is no advantage in enforcing a specific structure on those certificates. The
+certificate hierarchy consistency checks are required because nodes can issue their own certificates and can set
+their own role flags on certificates, and it's important to verify that these are set consistently with the
+certificate hierarchy design. As as side-effect this also acts as a secondary depth restriction on issued
```
gendal: As a

docs/source/permissioning.rst
gendal: Which certificates and keys will be auto generated?

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
 .. note:: You can exit the utility at any time if the approval process is taking longer than expected. The request process will resume on restart.
```
gendal: provided you restart with the same command-line flag?  Or does a node started without the --initial-registration flag know that it was in the middle of a request and resume its wait?

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+Do you need a zone?
+^^^^^^^^^^^^^^^^^^^
+
+Think twice before going this route:
+
+1. It isn't necessary for testing.
+2. It isn't necessary to add another layer of permissioning or 'know your customer' requirements onto your app.
+
+**Testing.** Creating a production-ready zone isn't necessary for testing as you can use the *network bootstrapper*
+tool to create all the certificates, keys and distribute map files needed to run many nodes. The bootstrapper can
```
gendal: Oxford comma would definitely help here....  certificates, keys, and distribute...

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+create a network locally on your desktop/laptop but it also knows how to automate cloud providers via their APIs and
+using Docker. In this way you can bring up a simulation of a real Corda network with different nodes on different
+machines in the cloud for your own testing. Testing this way has several advantages, most obviously that you avoid
+flakyness caused by synchronisation issues during network bringup. You can read more about the reasons for the
```
gendal: can we reword this? inattentive readers will interpret this as our saying corda is flaky.  How about: "... most obviously that you avoid the need to wait for information about all nodes to propagate to all other nodes as they start up before being able to begin testing; instead, with this approach, you know for sure that when a node starts it has all the information about the network it needs in order to support testing immediately.

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+   request/response flow that serves the list over the Corda protocol itself, although this requires the business
+   network to have a node for itself.
+2. Write a bit of code that downloads and caches the contents of this file on disk, and which loads it into memory in
+   the node. A good place to do this is in a class annotated with ``@CordaService``, because this class can expose
+   a ``Set<Party>`` field representing the membership of your service.
+3. In your flows use ``serviceHub.findService`` to get a reference to your ``@CordaService`` class, read the list of
+   members and at the start of each flow, throw a FlowException if the counterparty isn't in the membership list.
+
+In this way you can impose a centrally controlled ACL that all members will collectively enforce.
+
```
gendal: reiterate at this point that the production Corda network and Corda Testnet will shortly be available?

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+The network map server also distributes signed network parameter files and controls the rollout schedule for when they
+become available for download and opt-in, and when they become enforced. This is again a policy decision you will
+probably choose to place some simple UI or workflow tooling around, in particular to enforce restrictions on who can
+edit the map or the parameters.
+
+Writing a doorman server
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+This step is optional because your users can obtain a signed certificate in many different ways. The doorman protocol
```
gendal: And one such way is from the Corda Network or Corda Testnet doorman, right?  You've argued at this point that the only reason to go down this path is because people want different network params but I think there will also be people who are quite happy to adopt OUR params but who aren't _quite_ ready to join CN/CT yet.  In such cases, do we want to recommend a path that says: "If you really, REALLY, REALLY must run your own private network, be aware that you are, by definition, isolated from the main network and will soon wish you'd chosen to join the big party.  To reduce the risk of encountering this remorse, we suggest adopting the CN/CT netparams and using certificates issued by one of those networks.  This doesn't guarantee you'll subsequently be able to join the real networks (not least because they need to have run their notaries to the right standards) but there's at least hope...  whereas there is no hope if they use different params or issue their own certs

gendal: (but it does imply a small product change, I think?  since the private netmap server wouldn't be able to sign the network map it serves to the private network with the key that is associated with the CN/CT-issued certs)

docs/source/permissioning.rst
gendal: good point...!!! Java first!!

docs/source/permissioning.rst
```diff
@@ -221,3 +147,157 @@ Once the request has been approved and the certificates downloaded from the serv
...
+The network map certificate is usually chained off the root certificate, and can be created according to the instructions above.
+
+Each time the zone parameters are changed, the epoch should be incremented. Epochs are essentially version numbers for
+the parameters, and they therefore cannot go backwards.
+
+Once saved, the new parameters can be served by the network map server.
+
+How to choose the parameters? This is the most complex question facing you as a new zone operator. Some settings may seem
+straightforward and others may involve cost/benefit tradeoffs specific to your business. For example, you could choose
+to run a validating notary yourself, in which case you would (in the absence of SGX) see all the users data. Or you could
```
gendal: users'

docs/source/permissioning.rst
mikehearn: Hmm I was always taught in school *not* to put a comma at the end of a list like that. At any rate either is correct, so I'll leave it.

mikehearn:
Thanks both. I've addressed the review comments, with the exception of Richard's greater point.

I'm asking about the extent to which we can mix and match certs before things break. We didn't design for that and it'd probably complicate at least the security analysis.

docs/source/permissioning.rst
```diff
@@ -210,14 +143,188 @@ The following information from the node configuration file is needed to generate
...
+------------------------------------
+
+This section documents how to implement your own doorman and network map servers, which is the basic process required to
+create a dedicated zone. At this time we do not provide tooling to do this, because the needs of each zone are different
+and no generic, configurable doorman codebase has been written.
+
+Do you need a zone?
+^^^^^^^^^^^^^^^^^^^
+
+Think twice before going this route:
```
shamsasari: going down

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkParametersCopier.kt
```diff
@@ -19,7 +21,7 @@ class NetworkParametersCopier(
         val update: Boolean = false
 ) {
     private val copyOptions = if (overwriteFile) arrayOf(StandardCopyOption.REPLACE_EXISTING) else emptyArray()
-    private val serialisedSignedNetParams = signingCertAndKeyPair.sign(networkParameters).serialize()
+    private val serialisedSignedNetParams: SerializedBytes<SignedDataWithCert<NetworkParameters>> = signingCertAndKeyPair.sign(networkParameters).serialize()
```
shamsasari: Use the `SignedNetworkParameters` typealias

docs/source/permissioning.rst
```diff
@@ -210,14 +143,188 @@ The following information from the node configuration file is needed to generate
...
+certificates. Once the request has been approved and the certificates downloaded from the server, the node will create
+the keystore and trust store using the certificates and the generated private key.
 
-.. note:: You can exit the utility at any time if the approval process is taking longer than expected. The request process will resume on restart.
+.. note:: You can exit the utility at any time if the approval process is taking longer than expected. The request
+   process will resume on restart as long as the ``--initial-registration`` flag is specified.
 
 This process only is needed when the node connects to the network for the first time, or when the certificate expires.
+
+Creating your own compatibility zone
```
davejh69: This section seems a little awkward. We've started by describing the technical implementation of our certificate hierarchy, and how it fits into a CZ. This feels very much orthogonal to that and it veers towards being opinionated in an otherwise non-opinionated page. Feels like this should be a separate discussion. There are clear examples of places where a new CZ is required, even if they're not required for many of the reasons people might have expected.

mikehearn: Yeah, I agree it's a bit incongruous, but we *are* opinionated about this. We really want everyone to be able to connect to each other!

docs/source/permissioning.rst
```diff
@@ -97,98 +91,36 @@ The extension contains a single ASN.1 integer identifying the identity type the
 6. Well-known legal identity
 7. Confidential legal identity
 
-In a typical installation, node administrators needn't be aware of these. However, when node certificates are managed
+In a typical installation, node administrators need not be aware of these. However, when node certificates are managed
 by external tools (such as an existing PKI solution deployed within an organisation), it is important to understand
 these constraints.
```
davejh69: what constraints? We've not describe that anything is being constrained? If the certs are being constrained by the extension then what must an otherwise-valid cert not contain?

docs/source/permissioning.rst
davejh69: agreed that this is actually an opinion we want to express, but separating the technical content from the discursive would seem like a good idea. If anything it feels like we probably want to deal with the advocacy aspect up front. We describe the relationship of a node to its CZ, describe that there's a top-of-stack solution to support multiple BNs in a CZ, and then strongly suggest the use of tCN as a default CZ. If we then go into the identity hierarchy and what's required to manage that, we've already set an expectation that "default is tCN" and everything else is as much about understanding how that works as anything else (oh and, by the way, if you need the non-default then this is what you'll need to do)

docs/source/permissioning.rst
mikehearn: We've described the name constraints and the extension that defines the roles; that's what this is referring to. Can you see a way to make that sentence clearer?

davejh69: How about: "In a typical installation, node administrators need not be aware of these. However, if node certificates are to be managed by external tools, such as those provided as part of an existing PKI solution deployed within an organisation, it is important to recognise these extensions and the constraints noted above."

docs/source/permissioning.rst
mikehearn: I took another look at this, but, I can't quite convince myself that it's right to move this discussion to the top of the page. It's perfectly reasonable to want to understand permissioning *without* wanting to read about creating or not creating new zones. And I don't want to put the discussion of whether to do it in a separate page that risks people not reading it.

In conclusion I'd rather leave it as it is, though, I'd be happy to review PRs that have a concrete proposal for changing the layout or language.

mikehearn:
Can I get an approval from someone?

docs/source/permissioning.rst
```diff
@@ -268,14 +274,14 @@ The protocol is:
...
-It is easy to create such a file with a small Java or Kotlin program. The ``NetworkParameters`` object is a simple data
-holder that could be read from e.g. a config file, or settings from a database.
-
-Signing and saving the resulting file is just a few lines of code. A full example can be found in ``NetworkParametersCopier.kt`` in the
-source tree, but a flavour of it looks like this:
+Zone parameters are stored in a file containing a Corda AMQP serialised ``SignedDataWithCert<NetworkParameters>``
+object. It is easy to create such a file with a small Java or Kotlin program. The ``NetworkParameters`` object is a
+simple data holder that could be read from e.g. a config file, or settings from a database. Signing and saving the
+resulting file is just a few lines of code. A full example can be found in ``NetworkParametersCopier.kt`` in the source
+tree, but a flavour of it looks like this:
```
gendal: whenever I see something like this I think... "a vendor of software is telling me that doing thing X is a simple matter of writing some code... but isn't that the vendor's job?!!  Can we perhaps file a Jira (it's not a blocker to this PR) to create a trivial tool that does just this?

mikehearn: https://r3-cev.atlassian.net/browse/CORDA-1592

gendal: 👍

PR #3290, opened 2018-06-04
Add query function to get attachment matadata by hash id in CordaPRCOps

renlulu:

# PR Checklist:

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/Attachment.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/Attachment.kt
@@ -6,7 +6,10 @@ import net.corda.core.serialization.CordaSerializable
 import java.io.FileNotFoundException
 import java.io.InputStream
 import java.io.OutputStream
+import java.time.Instant
 import java.util.jar.JarInputStream
+import javax.persistence.Column
+import javax.persistence.Lob
 
 /**
  * An attachment is a ZIP (or an optionally signed JAR) that contains one or more files. Attachments are meant to
@@ -55,4 +58,16 @@ interface Attachment : NamedByHash {
      * Attachment size in bytes.
      */
     val size: Int
+
+}
+
+@CordaSerializable
+interface AttachmentMetadata {
+    val attId: String
+
+    val insertionDate: Instant
+
+    val uploader: String?
+
+    val filename: String?
 }
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -2,6 +2,7 @@ package net.corda.core.messaging
 
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.context.InvocationContext
+import net.corda.core.contracts.AttachmentMetadata
 import net.corda.core.contracts.ContractState
 import net.corda.core.crypto.SecureHash
 import net.corda.core.flows.FlowInitiator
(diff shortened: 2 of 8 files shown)


core/src/main/kotlin/net/corda/core/contracts/Attachment.kt
```diff
@@ -55,4 +58,16 @@ interface Attachment : NamedByHash {
      * Attachment size in bytes.
      */
     val size: Int
+
+}
+
+@CordaSerializable
+interface AttachmentMetadata {
```
joeldudleyr3: Can't we just add these as fields on `Attachment`?

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
```diff
@@ -167,7 +167,8 @@ sealed class AttachmentQueryCriteria : GenericQueryCriteria<AttachmentQueryCrite
      */
     data class AttachmentsQueryCriteria @JvmOverloads constructor (val uploaderCondition: ColumnPredicate<String>? = null,
                                                                    val filenameCondition: ColumnPredicate<String>? = null,
-                                                                   val uploadDateCondition: ColumnPredicate<Instant>? = null) : AttachmentQueryCriteria() {
+                                                                   val uploadDateCondition: ColumnPredicate<Instant>? = null,
+                                                                   val attId: ColumnPredicate<String>? = null) : AttachmentQueryCriteria() {
```
joeldudleyr3: Why query based on attachment ID? If you know the attachment ID, just grab that attachment, no?

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -297,9 +298,12 @@ interface CordaRPCOps : RPCOps {
     @Throws(java.nio.file.FileAlreadyExistsException::class)
     fun uploadAttachmentWithMetadata(jar: InputStream, uploader: String, filename: String): SecureHash
 
-    /** Queries attachments metadata */
+    /** Queries attachments hash id */
     fun queryAttachments(query: AttachmentQueryCriteria, sorting: AttachmentSort?): List<AttachmentId>
 
+    /** Queries attachments metadata */
+    fun queryAttachmentMetadata(attId: String): List<AttachmentMetadata>
```
joeldudleyr3: I think, instead of this, we should expand the `Attachment` object and provide a method here to get an attachment, maybe `getAttachment`. The user can then look at the `Attachment` themselves. You can already query based on the metadata using the `queryAttachments` method with an `AttachmentQueryCriteria`. But you'll have to check if this works (reading an input stream over RPC)

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -319,25 +320,36 @@ class NodeAttachmentService(
...
             // prepare query for execution
-            val query = session.createQuery(criteriaQuery)
+            val query = queryAttachmentsHelper(criteria, sorting)
 
             // execution
             query.resultList.map { AttachmentId.parse(it.attId) }
         }
     }
+
+    override fun getAttachmentMetadata(attachmentId: String): List<AttachmentMetadata>{
```
joeldudleyr3: Again, not sure about this method. How about we leave it out for now? You can use a `queryAttachments` then chain that with an `getAttachment`, no?

core/src/main/kotlin/net/corda/core/contracts/Attachment.kt
renlulu: thanks for your review, but I just found a little difficult adding these on Attachment, I dont want to....emmm there are too many changes to the existing code

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
renlulu: Let me think again，maybe it's not time for change

PR #3295, opened 2018-06-04
 CORDA-1494: Re-enable notarisation retries in the new state machine manager

adagys:
The original message-based retry approach does not work well with the new flow state machine due to the way sessions are handled. We decided to move the retry logic to flow-level: introduce RetryableFlow that won't have checkpoints persisted and will be restarted after a configurable timeout if it does not complete in time.

The RetryableFlow functionality will be internal for now, as it's mainly tailored for the notary client flow, and there are many subtle ways it can fail when used with arbitrary flows.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
@@ -1,0 +1,26 @@
+package net.corda.core.internal
+
+/**
+ * A marker for a flow that will return the same result if replayed from the beginning. Any side effects the flow causes
+ * must also be idempotent.
+ *
+ * Flow idempotency allows skipping persisting checkpoints, allowing better performance.
+ */
+interface IdempotentFlow
+
+/**
+ * An idempotent flow that needs to be replayed if it does not complete within a certain timeout.
+ *
+ * Example use would be the notary client flow: if the client sends a request to an HA notary cluster, it will get
+ * accepted by one of the cluster members, but the member might crash before returning a response. The client flow
+ * would be stuck waiting for that member to come back up. Retrying the notary flow will re-send the request to the
+ * next available notary cluster member.
+ *
+ * Note that any sub-flows called by a [RetryableFlow] are assumed to be [IdempotentFlow] and will NOT have checkpoints
+ * persisted. Otherwise, it wouldn't be possible to correctly reset the [RetryableFlow].
+ */
+// TODO: allow specifying retry settings per flow
+interface RetryableFlow : IdempotentFlow
+
+/** This exception is fired once the retry timeout expires. It will indicate to the flow hospital to restart the flow. */
+data class FlowRetryException(val maxRetries: Int) : Exception()
--- a/core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
@@ -21,6 +21,3 @@ interface IdempotentFlow
  */
 // TODO: allow specifying retry settings per flow
 interface RetryableFlow : IdempotentFlow
-
-/** This exception is fired once the retry timeout expires. It will indicate to the flow hospital to restart the flow. */
-data class FlowRetryException(val maxRetries: Int) : Exception()
--- a/core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
@@ -16,8 +16,8 @@ interface IdempotentFlow
  * would be stuck waiting for that member to come back up. Retrying the notary flow will re-send the request to the
  * next available notary cluster member.
  *
- * Note that any sub-flows called by a [RetryableFlow] are assumed to be [IdempotentFlow] and will NOT have checkpoints
- * persisted. Otherwise, it wouldn't be possible to correctly reset the [RetryableFlow].
+ * Note that any sub-flows called by a [TimedFlow] are assumed to be [IdempotentFlow] and will NOT have checkpoints
+ * persisted. Otherwise, it wouldn't be possible to correctly reset the [TimedFlow].
  */
 // TODO: allow specifying retry settings per flow
-interface RetryableFlow : IdempotentFlow
+interface TimedFlow : IdempotentFlow
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/Action.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/Action.kt
@@ -137,17 +137,17 @@ sealed class Action {
     data class RetryFlowFromSafePoint(val currentState: StateMachineState) : Action()
 
     /**
-     * Set a timer for retrying flow [flowId] if it does not complete in time.
+     * Schedule the flow [flowId] to be retried if it does not complete within the timeout period specified in the configuration.
      *
-     * Note that this only works with [RetryableFlow].
+     * Note that this only works with [TimedFlow].
      */
-    data class ScheduleFlowRetry(val flowId: StateMachineRunId) : Action()
+    data class ScheduleFlowTimeout(val flowId: StateMachineRunId) : Action()
 
     /**
-     * Cancel the retry timer for flow [flowId]. This must be called when a retryable flow completes to prevent
+     * Cancel the retry timeout for flow [flowId]. This must be called when a timed flow completes to prevent
      * unnecessary additional invocations.
      */
-    data class CancelFlowRetry(val flowId: StateMachineRunId) : Action()
+    data class CancelFlowTimeout(val flowId: StateMachineRunId) : Action()
 }
 
 /**
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/Action.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/Action.kt
@@ -135,6 +135,19 @@ sealed class Action {
      * Retry a flow from the last checkpoint, or if there is no checkpoint, restart the flow with the same invocation details.
      */
     data class RetryFlowFromSafePoint(val currentState: StateMachineState) : Action()
+
+    /**
+     * Set a timer for retrying flow [flowId] if it does not complete in time.
+     *
+     * Note that this only works with [RetryableFlow].
+     */
+    data class ScheduleFlowRetry(val flowId: StateMachineRunId) : Action()
+
+    /**
+     * Cancel the retry timer for flow [flowId]. This must be called when a retryable flow completes to prevent
+     * unnecessary additional invocations.
+     */
+    data class CancelFlowRetry(val flowId: StateMachineRunId) : Action()
 }
 
 /**
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
@@ -74,8 +74,8 @@ class ActionExecutorImpl(
             is Action.ExecuteAsyncOperation -> executeAsyncOperation(fiber, action)
             is Action.ReleaseSoftLocks -> executeReleaseSoftLocks(action)
             is Action.RetryFlowFromSafePoint -> executeRetryFlowFromSafePoint(action)
-            is Action.ScheduleFlowRetry -> scheduleFlowRetry(action)
-            is Action.CancelFlowRetry -> cancelFlowRetry(action)
+            is Action.ScheduleFlowTimeout -> scheduleFlowTimeout(action)
+            is Action.CancelFlowTimeout -> cancelFlowTimeout(action)
         }
     }
     private fun executeReleaseSoftLocks(action: Action.ReleaseSoftLocks) {
@@ -236,11 +236,11 @@ class ActionExecutorImpl(
         return checkpoint.serialize(context = checkpointSerializationContext)
     }
 
-    private fun cancelFlowRetry(action: Action.CancelFlowRetry) {
+    private fun cancelFlowTimeout(action: Action.CancelFlowTimeout) {
         stateMachineManager.cancelFlowTimeout(action.flowId)
     }
 
-    private fun scheduleFlowRetry(action: Action.ScheduleFlowRetry) {
+    private fun scheduleFlowTimeout(action: Action.ScheduleFlowTimeout) {
         stateMachineManager.scheduleFlowTimeout(action.flowId)
     }
 }
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
@@ -237,10 +237,10 @@ class ActionExecutorImpl(
     }
 
     private fun cancelFlowRetry(action: Action.CancelFlowRetry) {
-        stateMachineManager.cancelFlowRetry(action.flowId)
+        stateMachineManager.cancelFlowTimeout(action.flowId)
     }
 
     private fun scheduleFlowRetry(action: Action.ScheduleFlowRetry) {
-        stateMachineManager.scheduleFlowRetry(action.flowId)
+        stateMachineManager.scheduleFlowTimeout(action.flowId)
     }
 }
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
@@ -74,9 +74,10 @@ class ActionExecutorImpl(
             is Action.ExecuteAsyncOperation -> executeAsyncOperation(fiber, action)
             is Action.ReleaseSoftLocks -> executeReleaseSoftLocks(action)
             is Action.RetryFlowFromSafePoint -> executeRetryFlowFromSafePoint(action)
+            is Action.ScheduleFlowRetry -> scheduleFlowRetry(action)
+            is Action.CancelFlowRetry -> cancelFlowRetry(action)
         }
     }
-
     private fun executeReleaseSoftLocks(action: Action.ReleaseSoftLocks) {
         if (action.uuid != null) services.vaultService.softLockRelease(action.uuid)
     }
@@ -233,5 +234,13 @@ class ActionExecutorImpl(
 
     private fun serializeCheckpoint(checkpoint: Checkpoint): SerializedBytes<Checkpoint> {
         return checkpoint.serialize(context = checkpointSerializationContext)
+    }
+
+    private fun cancelFlowRetry(action: Action.CancelFlowRetry) {
+        stateMachineManager.cancelFlowRetry(action.flowId)
+    }
+
+    private fun scheduleFlowRetry(action: Action.ScheduleFlowRetry) {
+        stateMachineManager.scheduleFlowRetry(action.flowId)
     }
 }
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
@@ -555,6 +555,12 @@ class SingleThreadedStateMachineManager(
         mutex.locked { cancelRetryIfScheduled(flowId) }
     }
 
+    /**
+     * Schedules the flow [flowId] to be retried if it does not finish within a timeout value
+     * specified in the config.
+     *
+     * Assumes lock is taken on the [InnerState].
+     */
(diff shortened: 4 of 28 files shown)


adagys:
@rick-r3 we originally discussed to have retry settings defined per flow, but due to the way the subflow stack is stored we don't have access to instance vals, so I made it use global config values for now – we can change it later.

adagys:
At the moment this is also missing test with the real messaging layer

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -556,6 +547,38 @@ class SingleThreadedStateMachineManager(
...
+
+    override fun cancelFlowRetry(flowId: StateMachineRunId) {
+        mutex.locked { cancelRetryIfScheduled(flowId) }
+    }
+
+    private fun InnerState.scheduleRetry(flowId: StateMachineRunId) {
+        mutex.locked {
+            val flow = flows[flowId]
+            if (flow != null) {
+                with(serviceHub.configuration.p2pMessagingRetry) {
```
rick-r3: Perhaps add the TODO here about allowing flows to configure their own delay?

adagys: I added it next to RetryableFlow

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -556,6 +547,38 @@ class SingleThreadedStateMachineManager(
...
+    override fun scheduleFlowRetry(flowId: StateMachineRunId) {
+        mutex.locked { scheduleRetry(flowId) }
+    }
+
+    override fun cancelFlowRetry(flowId: StateMachineRunId) {
+        mutex.locked { cancelRetryIfScheduled(flowId) }
+    }
+
+    private fun InnerState.scheduleRetry(flowId: StateMachineRunId) {
+        mutex.locked {
```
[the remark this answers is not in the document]
adagys: yep, no need to lock here

node/src/main/kotlin/net/corda/node/services/statemachine/transitions/TopLevelTransition.kt
```diff
@@ -100,6 +101,9 @@ class TopLevelTransition(
                                     subFlowStack = currentState.checkpoint.subFlowStack + subFlow.value
                             )
                     )
+                    if (RetryableFlow::class.java.isAssignableFrom(event.subFlowClass)) {
+                        actions.add(Action.ScheduleFlowRetry(currentState.flowLogic.runId))
```
rick-r3: The cancel checks it's the last `RetryableFlow` but this doesn't check if it's the first.   So two flows nested would reset the timer each time (or at least overwrite the future).

rick-r3: The cancel checks it's the last `RetryableFlow` but this doesn't check if it's the first.   So two flows nested would reset the timer each time (or at least overwrite the future).  Needs a test?

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -556,6 +547,36 @@ class SingleThreadedStateMachineManager(
...
+    override fun scheduleFlowRetry(flowId: StateMachineRunId) {
+        mutex.locked { scheduleRetry(flowId) }
+    }
+
+    override fun cancelFlowRetry(flowId: StateMachineRunId) {
+        mutex.locked { cancelRetryIfScheduled(flowId) }
+    }
+
+    private fun InnerState.scheduleRetry(flowId: StateMachineRunId) {
+        val flow = flows[flowId]
```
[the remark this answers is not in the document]
rick-r3: Actually. Just put a prominent comment on them to say they happen under the lock.  You won't be able to use `alreadyLocked` in multi-threaded as it doesn't exist.... so just thinking it would keep the code more similar.

core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Example use would be the notary client flow: if the client sends a request to an HA notary cluster, it will get
+ * accepted by one of the cluster members, but the member might crash before returning a response. The client flow
+ * would be stuck waiting for that member to come back up. Retrying the notary flow will re-send the request to the
+ * next available notary cluster member.
+ *
+ * Note that any sub-flows called by a [RetryableFlow] are assumed to be [IdempotentFlow] and will NOT have checkpoints
+ * persisted. Otherwise, it wouldn't be possible to correctly reset the [RetryableFlow].
+ */
+// TODO: allow specifying retry settings per flow
```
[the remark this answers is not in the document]
adagys: Done https://r3-cev.atlassian.net/browse/ENT-2033. Well it would need a bit more moving things around since the subflow stack is kept inside the checkpoint right now and we don't want to serialize the instances in.

core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
```diff
@@ -0,0 +1,26 @@
...
+interface RetryableFlow : IdempotentFlow
+
+/** This exception is fired once the retry timeout expires. It will indicate to the flow hospital to restart the flow. */
+data class FlowRetryException(val maxRetries: Int) : Exception()
```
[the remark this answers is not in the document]
adagys: Yep, moved

adagys:
One more thing this is missing is exponenital backoff, but I'm not sure how important it is to get it back in right now

core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Example use would be the notary client flow: if the client sends a request to an HA notary cluster, it will get
+ * accepted by one of the cluster members, but the member might crash before returning a response. The client flow
+ * would be stuck waiting for that member to come back up. Retrying the notary flow will re-send the request to the
+ * next available notary cluster member.
+ *
+ * Note that any sub-flows called by a [RetryableFlow] are assumed to be [IdempotentFlow] and will NOT have checkpoints
+ * persisted. Otherwise, it wouldn't be possible to correctly reset the [RetryableFlow].
+ */
+// TODO: allow specifying retry settings per flow
+interface RetryableFlow : IdempotentFlow
```
adagys: Just thinking RetryableFlow might sound slightly confusing, as we can technically retry any flow, maybe "resettable" or "replayable" would be better?

core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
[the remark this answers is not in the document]
rick-r3: But the instances will already be on the actual stack / in the checkpoint already... anyway, we have the JIRA ticket so all good.

core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
rick-r3: Agreed.  Need to convey the timeout element better too.  `RestartAfterTimeoutFlow`?  The flow does restart rather than retrying last checkpoint, I suppose.

rick-r3:
How would we do exponential back off now or later?

rick-r3:
I can answer my own back off question.  Looking at the code, the `retryFlows` map could have a more sophisticated value (e.g. data class) that has the `Future` we have now, but also a counter of attempts.  that we increment.  Can you add a follow up JIRA ?  We should get at least some retry into the codebase as is.

core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
rick-r3: Do all code mentions of "retry" need to be come something else in this PR (more or less) if we do that?

rick-r3: Maybe it should be `TimedFlow` to keep it simple.  And throw a `TimeoutFlowException`.  And methods like `scheduleRetry` become `scheduleTimeout`.  That seems easy to follow without being a mouthful.

adagys:
Tested the backoff manually

node/src/main/kotlin/net/corda/node/services/statemachine/Action.kt
```diff
@@ -135,6 +135,19 @@ sealed class Action {
      * Retry a flow from the last checkpoint, or if there is no checkpoint, restart the flow with the same invocation details.
      */
     data class RetryFlowFromSafePoint(val currentState: StateMachineState) : Action()
+
+    /**
+     * Set a timer for retrying flow [flowId] if it does not complete in time.
+     *
+     * Note that this only works with [RetryableFlow].
+     */
+    data class ScheduleFlowRetry(val flowId: StateMachineRunId) : Action()
```
rick-r3: rename these to timeout ?

node/src/main/kotlin/net/corda/node/services/statemachine/Action.kt
```diff
@@ -135,6 +135,19 @@ sealed class Action {
...
+
+    /**
+     * Cancel the retry timer for flow [flowId]. This must be called when a retryable flow completes to prevent
+     * unnecessary additional invocations.
+     */
+    data class CancelFlowRetry(val flowId: StateMachineRunId) : Action()
```
rick-r3: rename these to timeout ?

node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
```diff
@@ -234,4 +235,12 @@ class ActionExecutorImpl(
     private fun serializeCheckpoint(checkpoint: Checkpoint): SerializedBytes<Checkpoint> {
         return checkpoint.serialize(context = checkpointSerializationContext)
     }
+
+    private fun cancelFlowRetry(action: Action.CancelFlowRetry) {
```
rick-r3: rename these to timeout ?

node/src/main/kotlin/net/corda/node/services/statemachine/ActionExecutorImpl.kt
```diff
@@ -234,4 +235,12 @@ class ActionExecutorImpl(
...
+        stateMachineManager.cancelFlowTimeout(action.flowId)
+    }
+
+    private fun scheduleFlowRetry(action: Action.ScheduleFlowRetry) {
```
rick-r3: rename these to timeout ?

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -83,18 +70,28 @@ class SingleThreadedStateMachineManager(
 
     private class Flow(val fiber: FlowStateMachineImpl<*>, val resultFuture: OpenFuture<Any?>)
 
+    private data class ScheduledRetry(
```
rick-r3: rename to `ScheduledTimeout` ?

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -747,7 +802,8 @@ class SingleThreadedStateMachineManager(
         while (true) {
             val event = flow.fiber.transientValues!!.value.eventQueue.tryReceive() ?: return
             when (event) {
-                is Event.DoRemainingWork -> {}
+                is Event.DoRemainingWork -> {
```
rick-r3: You could put this back on one line.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -124,4 +125,31 @@ object StaffedFlowHospital : FlowHospital {
...
+                        log.warn("\"Maximum number of retries reached for flow ${flowFiber.javaClass}")
+                    }
+                } else {
+                    log.warn("\"Unable to retry flow: ${flowFiber.javaClass}, it is not retryable and does not contain any retryable sub-flows.")
+                }
+            }
+            return Diagnosis.NOT_MY_SPECIALTY
+        }
+
+        private fun isRetryable(flowFiber: FlowFiber): Boolean {
```
rick-r3: `isTimed`?

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -124,4 +125,31 @@ object StaffedFlowHospital : FlowHospital {
             return exception != null && (exception is org.hibernate.exception.ConstraintViolationException || mentionsConstraintViolation(exception.cause))
         }
     }
+
+    /**
+     * Restarts [TimedFlow], keeping track of the number of retries and making sure it does not
+     * exceed the limit specified by the flow.
+     */
+    object DoctorRetry : Staff {
```
rick-r3: `DoctorTimeout`?

node/src/main/kotlin/net/corda/node/services/statemachine/transitions/TopLevelTransition.kt
```diff
@@ -95,11 +96,20 @@ class TopLevelTransition(
...
+                    }
+                    val isCurrentSubFlowRetryable = TimedFlow::class.java.isAssignableFrom(event.subFlowClass)
                     currentState = currentState.copy(
                             checkpoint = currentState.checkpoint.copy(
                                     subFlowStack = currentState.checkpoint.subFlowStack + subFlow.value
                             )
                     )
+                    // We don't schedule a retry if there already is a retryable subflow on the stack - a retry had
+                    // been scheduled already.
+                    if (isCurrentSubFlowRetryable && !containsRetryableSubFlows) {
```
rick-r3: Need to re-word from "retry" to "timed", both the helper function and `val`s

node/src/main/kotlin/net/corda/node/services/statemachine/transitions/TopLevelTransition.kt
```diff
@@ -115,35 +125,56 @@ class TopLevelTransition(
             if (checkpoint.subFlowStack.isEmpty()) {
                 freshErrorTransition(UnexpectedEventInState())
             } else {
+                val lastSubFlowClass = checkpoint.subFlowStack.last().flowClass
+                val isLastSubFlowRetryable = TimedFlow::class.java.isAssignableFrom(lastSubFlowClass)
```
rick-r3: rewording

node/src/main/kotlin/net/corda/node/services/statemachine/transitions/TopLevelTransition.kt
```diff
@@ -115,35 +125,56 @@ class TopLevelTransition(
...
+                val newSubFlowStack = checkpoint.subFlowStack.dropLast(1)
                 currentState = currentState.copy(
                         checkpoint = checkpoint.copy(
-                                subFlowStack = checkpoint.subFlowStack.subList(0, checkpoint.subFlowStack.size - 1).toList()
+                                subFlowStack = newSubFlowStack
                         )
                 )
+                if (isLastSubFlowRetryable && !containsRetryableFlows(currentState.checkpoint.subFlowStack)) {
```
rick-r3: rewording

node/src/main/kotlin/net/corda/node/services/statemachine/transitions/TopLevelTransition.kt
```diff
@@ -115,35 +125,56 @@ class TopLevelTransition(
...
+                    actions.add(Action.CancelFlowRetry(currentState.flowLogic.runId))
+                }
             }
             FlowContinuation.ProcessEvents
         }
     }
 
+    private fun containsRetryableFlows(subFlowStack: List<SubFlow>): Boolean {
```
rick-r3: rewording

node/src/main/kotlin/net/corda/node/services/statemachine/transitions/TopLevelTransition.kt
```diff
@@ -95,11 +96,20 @@ class TopLevelTransition(
             val subFlow = SubFlow.create(event.subFlowClass, event.subFlowVersion)
             when (subFlow) {
                 is Try.Success -> {
+                    val containsRetryableSubFlows = currentState.checkpoint.subFlowStack.any {
```
rick-r3: Rewording these few lines.

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -83,18 +70,28 @@ class SingleThreadedStateMachineManager(
...
         var stopping = false
         val flows = HashMap<StateMachineRunId, Flow>()
         val startedFutures = HashMap<StateMachineRunId, OpenFuture<Unit>>()
+        /** Flows scheduled to be retried if not finished within the specified timeout period. */
+        val timedFlows = HashMap<StateMachineRunId, ScheduledRetry>()
     }
 
     private val mutex = ThreadBox(InnerState())
     private val scheduler = FiberExecutorScheduler("Same thread scheduler", executor)
+    private val retryScheduler = Executors.newScheduledThreadPool(1)
```
rick-r3: `timeoutScheduler` ?

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -556,6 +556,59 @@ class SingleThreadedStateMachineManager(
...
+    /**
+     * Schedules the flow [flowId] to be retried if it does not finish within the timeout period
+     * specified in the config.
+     *
+     * Assumes lock is taken on the [InnerState].
+     */
+    private fun InnerState.scheduleTimeout(flowId: StateMachineRunId) {
+        val flow = flows[flowId]
+        if (flow != null) {
+            val scheduledRetry = timedFlows[flowId]
```
rick-r3: `scheduledTimeout`

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -556,6 +556,59 @@ class SingleThreadedStateMachineManager(
...
+            val retryCount = if (scheduledRetry != null) {
+                val retryFuture = scheduledRetry.scheduledFuture
+                if (!retryFuture.isDone) scheduledRetry.scheduledFuture.cancel(true)
+                scheduledRetry.retryCount
+            } else 0
+            val scheduledFuture = scheduleTimeoutException(flow, retryCount)
+            timedFlows[flowId] = ScheduledRetry(scheduledFuture, retryCount + 1)
+        } else {
+            logger.warn("Unable to schedule retry for flow $flowId – flow not found.")
```
rick-r3: `Unable to schedule timeout...`

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -124,4 +125,31 @@ object StaffedFlowHospital : FlowHospital {
...
+        override fun consult(flowFiber: FlowFiber, currentState: StateMachineState, newError: Throwable, history: MedicalHistory): Diagnosis {
+            if (newError is FlowTimeoutException) {
+                if (isRetryable(flowFiber)) {
+                    if (history.notDischargedForTheSameThingMoreThan(newError.maxRetries, this)) {
+                        return Diagnosis.DISCHARGE
+                    } else {
+                        log.warn("\"Maximum number of retries reached for flow ${flowFiber.javaClass}")
+                    }
+                } else {
+                    log.warn("\"Unable to retry flow: ${flowFiber.javaClass}, it is not retryable and does not contain any retryable sub-flows.")
```
rick-r3: `is not timed and does not contain timed sub-flows`

core/src/main/kotlin/net/corda/core/internal/IdempotentFlow.kt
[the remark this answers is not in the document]
exFalso: Currently the FlowLogics are stored in another "layer" of serialization in `serializedFiber`, so they would be serialized again. Also a small thing is these FlowLogics are mutable by the user, so StateMachineState wouldn't be immutable anymore, which may cause confusions e.g. when comparing two subsequent states.
I think the easiest is to just add the data to the existing SubFlow datastructure

PR #3262, opened 2018-05-30
ENT-1463, ENT-1903: Create core-deterministic and serialization-deterministic artifacts.

chrisr3:
commit 6f77838fe53d7c9565283c20bbf20f27954b27f6
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Tue May 29 23:14:23 2018 +0100

    Tidy up Gradle files.

commit 0aa958d31c6342e92ad4d6ab396db6e4a39d4fed
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Tue May 29 18:11:17 2018 +0100

    Fix EnclaveletTest with deterministic core and serialisation.

commit 732fcf37ee2219dfad373200676241d2fd90aeb3
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Sun May 27 00:21:42 2018 +0100

    Extend JarFilter to delete typealias declarations.

commit 25dbf30ed62c0c059df07782306b7f760f4cdf73
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Thu May 24 17:20:41 2018 +0100

    Test deserialising a contract and verifying it using core-deterministic.

commit f7753bf2ab588e084cb8bfaa5fd04f1a18d3aaef
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Thu May 24 11:25:49 2018 +0100

    Do not remove constructors from Kotlin annotation metadata.

commit 4ddf357b71b29775aa921aca33b4505a402a20e8
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Wed May 23 16:30:18 2018 +0100

    Add Gradle modules for deterministic rt.jar artifacts.

commit e81f462eefad2369706fd1b8447d426a71a25a03
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Wed May 23 16:22:04 2018 +0100

    Isolate reference to WeakHashMap - for deletion!

commit eea2458fbec06b28344547fdf9c191a9445fe1e7
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Thu May 17 18:01:20 2018 +0100

    Extract Kotlin metadata classes from kotlin-compiler-embeddable.
    This fixes a classpath issue that was crashing Gradle.

commit 87fdb938d83f3de6589730343c860fbbc406942e
Author: Chris Rankin <chris.rankin@r3.com>
Date:   Tue May 15 15:40:31 2018 +0100

[...]

PR #3297, opened 2018-06-04
ENT-1463, ENT-1903: Document the "determinisation" process.

chrisr3:


docs/source/deterministic-modules.rst
```diff
@@ -0,0 +1,253 @@
...
+  To build a new version of this JAR and upload it to the Artifactory, see the ``create-jdk8u`` module. This is a
+  standalone Gradle project within the Corda repository that will clone the ``deterministic-jvm8`` branch of Corda's
+  `OpenJDK repository <https://github.com/corda/openjdk>`_ and then build it. (This currently requires a C++ compiler,
+  GNU Make and a UNIX-like development environment.)
+
+Corda Modules
+  ``core-deterministic`` and ``serialization-deterministic`` are generated from Corda's ``core`` and ``serialization``
+  modules respectively using both `ProGuard <https://www.guardsquare.com/en/proguard>`_ and Corda's ``JarFilter`` Gradle
+  plugin. Corda developers configure these tools by applying Corda's ``@Deterministic`` and ``@NonDeterministic``
+  annotations to elements of ``core`` and ``serialization`` as described `here <deterministic_annotations_>`_.
```
mikehearn: need double underscore for links with ambiguous text like "here" (see my email to corda-dev on the topic of rst markup errors).

chrisr3: This is the link target for the "here" text:
```
.. _deterministic_annotations:

```
This shouldn't be ambiguous.

docs/source/deterministic-modules.rst
```diff
@@ -0,0 +1,253 @@
...
+
+    ProGuard works by calculating how much code is reachable from given "entry points", and in our case these entry
+    points are the ``@Deterministic`` classes. The unreachable classes are then discarded by ProGuard's ``shrink``
+    option.
+ #. The remaining classes may still contain non-deterministic code. However, there is no way of writing a ProGuard rule
+    explicitly to discard anything. Consider the following class:
+
+    .. sourcecode:: kotlin
+
+        @CordaSerializable
```
exFalso: Perhaps it's better to link source code using `literalinclude` so we don't have to keep this piece of code updated with `UniqueIdentifier`

docs/source/deterministic-modules.rst
```diff
@@ -0,0 +1,253 @@
...
+    any more code that has now become unreachable.
+ #. The remaining classes define our deterministic subset. However, the ``@kotlin.Metadata`` annotations on the compiled
+    Kotlin classes still contain references to all of the functions and properties that ProGuard has deleted. Therefore
+    we now use the ``JarFilter`` to delete these references, as otherwise the Kotlin compiler will pretend that the
+    deleted functions and properties are still present.
+ #. Finally, we use ProGuard again to validate our JAR against the deterministic ``rt.jar``:
+
+    .. sourcecode:: groovy
+
+        task checkDeterminism(type: ProGuardTask, dependsOn: jdkTask) {
```
exFalso: Similarly, better to use `literalinclude`. See hello-world-state.rst for an example

docs/source/deterministic-modules.rst
chrisr3: I can do that, but suspect I'll need to get _both_ #3262 _and_ #3297 merged first.

chrisr3: Although to be fair, if someone were to change `UniqueIdentifer` _not_ to have these secondary constructors then they'd break my document instead :wink:!

I don't think that this _particular_ example benefits from being linked to the actual code.

PR #3304, opened 2018-06-05
CORDA-1546: Updated the flow hospital to suspend FinalityHandler if it errors

shamsasari:
It will re-run automatically from last checkpoint on node restart, allowing the opportunity to resolve the issue, something required when dealing with contract constraint failures.

Because we're restarting mock nodes for unit testing, StaffedFlowHospital was made into a non-singleton.


node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -43,29 +47,25 @@ object StaffedFlowHospital : FlowHospital {
         log.info("Flow ${flowFiber.id} admitted to hospital in state $currentState")
         val medicalHistory = patients.computeIfAbsent(flowFiber.id) { MedicalHistory() }
         medicalHistory.records += MedicalHistory.Record.Admitted(Instant.now(), currentState.checkpoint.numberOfSuspends)
-        for ((index, error) in errors.withIndex()) {
```
rick-r3: I think you've changed the behaviour here.  Was that intentional?  Perhaps I should have had a unit test!

Previously, if any error didn't result in discharge, we would propagate.  Now if any error results in discharge, we retry.

I think we still want to iterate over the errors, one by one.  If no one handles the error, then we propagate without looking at further errors.  Maybe first decision is the chosen one if someone returns != NOT_MY_SPECIALTY.  i.e. if we want finality handler decision to trump deadlocks, we put it before it in the list of staff.  (I think deadlock trumps finality handler, actually, since it might succeed on retry).  An overall rule that discharge trumps observation or vice versa is hard to see, so I think we go with first person to handle each error.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -76,26 +76,24 @@ object StaffedFlowHospital : FlowHospital {
...
     enum class Diagnosis {
-        /**
-         * Retry from last safe point.
-         */
+        /** Retry from last safe point. */
         DISCHARGE,
-        /**
-         * Please try another member of staff.
-         */
+        /** Park and retry on node restart. */
```
rick-r3: I would describe this as `Park and await intervention`.  We'll forget to update the comment when we add further features.  e.g. it already doesn't account for `killFlow`.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -1,39 +1,43 @@
...
-            }
+        sealed class Record {
+            abstract val suspendCount: Int
 
-            class Discharged(val at: Instant, suspendCount: Int, val by: Staff, val error: Throwable) : Record(suspendCount) {
-                override fun toString() = "Discharged(at=$at, suspendCount=$suspendCount, by=$by)"
-            }
+            data class Admitted(val at: Instant, override val suspendCount: Int) : Record()
+            data class Discharged(val at: Instant, override val suspendCount: Int, val by: List<Staff>, val errors: List<Throwable>) : Record()
+            data class Observation(val at: Instant, override val suspendCount: Int, val by: List<Staff>, val errors: List<Throwable>) : Record()
```
rick-r3: I think we should call this `KeptInForObservation`.

We should also add a `NothingWeCanDo` for when we propagate errors.  Looking at it now, we'll have an unmatched `Admitted` when we propagate if we don't.  Adds a nice symmetry.

docs/source/api-contract-constraints.rst
```diff
@@ -132,6 +132,22 @@ you require the hash of the node's installed app which supplies the specified co
 hash constraints, you almost always want "whatever the current code is" and not a hard-coded hash. So this automatic
 constraint placeholder is useful.
 
+FinalityFlow
+------------
+
+It's possible to encounter contract contraint issues when notarising transactions with the ``FinalityFlow`` on a network
+containing multiple versions of the same CorDapp. This will happen when using hash contraints or with zone contraints
+if the zone whitelist has missing CorDapp versions. If a participating party fails to validate the **notarised** transaction
+then we have a scenerio where the members of the network do have an consistent view of the ledger.
```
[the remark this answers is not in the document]
shamsasari: Missing "not"!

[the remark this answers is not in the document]
shamsasari: And incorrect "an"!

docs/source/api-flows.rst
```diff
@@ -581,6 +581,16 @@ We can also choose to send the transaction to additional parties who aren't one
...
 **not** need to be called by each participant individually.
 
+Because the transaction has already been notarised and the input states consumed, if the participants when receiving the
+transaction fail to verify it, or the receiving flow (the finality handler) fails due to some other error, we then have
+the scenerio where not all parties have the correct up to date view of the ledger. To recover from this the finality handler
+is automatically sent to the flow hospital where it's suspended and retried from its last checkpoint on node restart.
+This gives the node operator the opportunity to recover from the error. Until the issue is resolved the node will continue
+to retry the flow on each startup.
+
+.. note:: It's possible to forcibly terminate the erroring finality handler using the ``killRPC`` flow but at the risk
```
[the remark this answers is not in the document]
shamsasari: I tell you, my fingers have a mind of their own.

shamsasari:
Comments addressed

node/src/test/kotlin/net/corda/node/services/FinalityHandlerTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+        assertThat(bob.getTransaction(finalisedTx.id)).isNull()
+        val finalityHandlerId = bobFinalityHandlerChanges.assertFlowNotRemoved()
+
+        val observations = bob.smm.flowHospital.patients[finalityHandlerId]!!.records.filterIsInstance<KeptInForObservation>()
+        assertThat(observations).hasSize(1)
+        assertThat(observations[0].by).contains(StaffedFlowHospital.FinalityDoctor)
+
+        bob = mockNet.restartNode(bob)
+        Thread.sleep(500)
```
rick-r3: Not a fan of this.  It's possible it will become flakey.   I don't think `mockNet.waitQuiescent()` will work instead because of the node shutdown in the middle of a flow (nodes that shutdown don't countdown by however many flows they still have in the state machine), but we could maybe provide access to manually count it down by one to simulate the effect of node restart or actually have mock nodes countdown when they are stopped.  Can't think how else to do it.  How hard is the latter, for example?

shamsasari: The purpose of the sleep is simply to ensure the FinalityHandler is run again. The problem we have here is that at node restart there's exactly on flow in the SMM (the handler) and after the flow has run and errored again there's still just one, so we have no way of knowing that it actually did run, which is the whole point of the test. `waitQuiescent` doesn't help us here. I did experiment which grepping the logs to check that the exception was thrown again, but that's even hackier I think.

shamsasari:
Got rid of the sleep by adding in a `track` function to the hospital.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -1,89 +1,153 @@
...
+class StaffedFlowHospital : FlowHospital {
+    private companion object {
+        private val log = contextLogger()
+        private val staff = listOf(DeadlockNurse, DuplicateInsertSpecialist, DoctorTimeout, FinalityDoctor)
+    }
 
-    private val staff = listOf(DeadlockNurse, DuplicateInsertSpecialist, DoctorTimeout)
+    private val mutex = ThreadBox(object {
+        val patients = HashMap<StateMachineRunId, MedicalHistory>()
+        val previouslyAdmitted = ArrayList<FlowHospital.Admission>()
```
rick-r3: This will grow indefinitely.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -1,89 +1,153 @@
...
+        val admissionsPublisher = PublishSubject.create<FlowHospital.Admission>()
```
rick-r3: Why not publish `Pair<StateMachineRunId, Record>` ?  The `Record` types are going to end up public in some way for looking in the hospital.  Maybe we want a data class rather than `Pair` just to make it a bit nicer.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -1,89 +1,153 @@
...
     // It's okay for flows to be cleaned... we fix them now!
-    override fun flowCleaned(flowFiber: FlowFiber) {}
+    override fun flowCleaned(flowFiber: FlowFiber) = Unit
 
     override fun flowRemoved(flowFiber: FlowFiber) {
-        patients.remove(flowFiber.id)
+        mutex.locked { patients.remove(flowFiber.id) }
+    }
+
+    override fun track(): DataFeed<List<FlowHospital.Admission>, FlowHospital.Admission> {
```
rick-r3: Rather than the list, just return the `patients` or something?

node/src/test/kotlin/net/corda/node/services/FinalityHandlerTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        bob.assertFinalityHandlerSentForObservation()
+        assertThat(bob.getTransaction(finalisedTx.id)).isNull()
+
+        bob = mockNet.restartNode(bob)
+        bob.assertFinalityHandlerSentForObservation()
+        assertThat(bob.getTransaction(finalisedTx.id)).isNull()
+    }
+
+    private fun StartedNode<*>.assertFinalityHandlerSentForObservation() {
+        val finalityHandlerAdmission = smm.flowHospital
```
rick-r3: Is there a race here where the event could be delivered before you subscribe if the node is fast enough?  Maybe we should first look at the snapshot returned, and if that doesn't contain the event we're looking for, then wait for the event ?

shamsasari: That's what `it.updates.startWith(it.snapshot)` below does.

shamsasari:
Removed the `FlowHospital` interface and made the `track()` method use the medical record class.

node/src/test/kotlin/net/corda/node/services/FinalityHandlerTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        // Since we've not done anything to fix the orignal error, we expect the finality handler to be sent to the hospital
+        // again on restart
+        bob.assertFlowSentForObservation(finalityHandlerId)
+        assertThat(bob.getTransaction(finalisedTx.id)).isNull()
+    }
+
+    private fun StartedNode<*>.assertFlowSentForObservation(runId: StateMachineRunId) {
+        val keptInForObservation = smm.flowHospital
+                .track()
+                .let { it.updates.startWith(it.snapshot) }
```
rick-r3: Nice!

PR #3307, opened 2018-06-05
Removed flow-library doc page as it's redundant

shamsasari:


docs/source/api-flows.rst
```diff
@@ -618,6 +618,16 @@ transaction and provide their signature if they are satisfied:
...
 
+Types of things to check include:
+
+    * Ensuring that the transaction received is the expected type, i.e. has the expected type of inputs and outputs
+    * Checking that the properties of the outputs are expected, this is in the absence of integrating reference
+      data sources to facilitate this
+    * Checking that the transaction is not incorrectly spending (perhaps maliciously) asset states, as potentially
+      the transaction creator has access to some of signer's state references
+
+Typically ``FinalityFlow`` is called after ``CollectSignaturesFlow``.
```
joeldudleyr3: Not sure whether to keep this line. If you call `FinalityFlow` before `CollectSignaturesFlow`, the flow will fail due to missing signatures. Think it would be less confusing to remove.

shamsasari: Removed

PR #3220, opened 2018-05-22
CORDA-1510 - Allow Doorman and NetworkMap to be configured independently

fenryka:
Currently only one compatabilityZoneURL can be specified, however, the
two services can be run on as separate servers. Allow nodes to be
configured in this manner

PR #3308, opened 2018-06-05
CORDA-1510 - Allow Doorman and NetworkMap to be configured independently (#3220)

fenryka:


fenryka:
This needs re-backporting in a way that doesn't bring the world back with it, i.e. re-doing rather than a cherry-pick

PR #3313, opened 2018-06-06
Blobinspector: trace level logging with --verbose

shamsasari:


shamsasari:
@tlil can you update the binary available on the downloads page with this build. Thanks.

tlil:
Will do @shamsasari :+1:

tlil:
Done

PR #3317, opened 2018-06-06
CORDA-1593 Extracting the whitelist logic and making it public

mkit:
This PR extracts the whitelist generation logic and makes it public. This is needed for the [ENT-2020](https://r3-cev.atlassian.net/browse/ENT-2020).
This change doesn't alter the existing behaviour, therefore no docs/changelog updates are needed.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/WhitelisGenerator.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/WhitelisGenerator.kt
@@ -1,0 +1,41 @@
+package net.corda.nodeapi.internal.network
+
+import net.corda.core.contracts.ContractClassName
+import net.corda.core.internal.div
+import net.corda.core.internal.exists
+import net.corda.core.internal.readAllLines
+import net.corda.core.internal.toMultiMap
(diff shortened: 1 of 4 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/WhitelisGenerator.kt
```diff
@@ -0,0 +1,41 @@
...
+
+private const val EXCLUDE_WHITELIST_FILE_NAME = "exclude_whitelist.txt"
+
+fun generateWhitelist(networkParameters: NetworkParameters?,
+                      excludeContracts: List<ContractClassName>,
+                      cordappJars: List<ContractsJar>): Map<ContractClassName, List<AttachmentId>> {
+    val existingWhitelist = networkParameters?.whitelistedContractImplementations ?: emptyMap()
+
+    if (excludeContracts.isNotEmpty()) {
+        println("Exclude contracts from whitelist: ${excludeContracts.joinToString()}")
```
tlil: Use logger instead of `println` ?

fenryka:
Also commit doesn't reference the Jira :)

PR #3321, opened 2018-06-07
Publish tools to Artifactory: Blob Inspector, Network Bootstrapper

josecoll:
Adding publication to Artifactory of following tools:
- network bootstrapper
- blob inspector

Clean-up redundant references.


settings.gradle
```diff
@@ -21,7 +21,6 @@ include 'experimental:behave'
 include 'experimental:sandbox'
 include 'experimental:quasar-hook'
 include 'experimental:kryo-hook'
-include 'experimental:blobinspector'
```
shamsasari: Doh!

tools/blobinspector/build.gradle
```diff
@@ -18,11 +20,15 @@ jar {
...
     }
-    baseName 'blobinspector'
+    archiveName = "blob-inspector-${version}.jar"
     manifest {
         attributes(
                 'Automatic-Module-Name': 'net.corda.blobinspector',
                 'Main-Class': 'net.corda.blobinspector.MainKt'
         )
     }
 }
```
shamsasari: In the ENT PR you had an extra `jar`: https://github.com/corda/enterprise/pull/943/files#diff-95f866e30c06a786cefa8898c67a95e7R30

tools/blobinspector/build.gradle
```diff
@@ -18,11 +20,15 @@ jar {
...
+
+publish {
+    name 'tools-blob-inspector'
```
shamsasari: Similarly there's a `disableDefaultJar = true` in ENT

shamsasari:
Backport of https://github.com/corda/enterprise/pull/943

tools/blobinspector/build.gradle
tlil: not anymore; https://github.com/corda/enterprise/blob/pre-release-V3/tools/blobinspector/build.gradle

tools/blobinspector/build.gradle
tlil: This seem to be different though :+1:

josecoll: Discussed with @Clintonio. 
That directive should not be included when publishing the default jar (so its more clean-up than anything else).

tools/blobinspector/build.gradle
josecoll: The ENT PR also publishes the database migration tool (which doesn't exist in OS).

tlil: I think he was referring to this jar-block :)
```
jar {
    classifier "ignore"
}
```

josecoll: OK - yes, so confirmed with @Clintonio that this additional is not required in this scenario.

PR #3327, opened 2018-06-08
CORDA-1588: Add an extra check in the attachment resolution flow to p…

adagys:
…revent duplicate attachment import if multiple transactions with the same attachment are being resolved at the same time.

Ideally we'd modify the `importAttachment` method to ignore duplicate attachments instead of throwing an exception, but that would be an api break..


adagys:
Tested both in Corda and Corda Ent with a multithreaded node – seems to fix it

PR #3329, opened 2018-06-08
ENT-1950: handle unknown peers when replying after a flow is processed

bpaunescu:
The flow hospital is used to keep track of flows that failed due to lack of peer info. SMM subscribes to NM cache and retries these flows on next update.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
@@ -12,6 +12,7 @@ import net.corda.core.flows.FlowException
 import net.corda.core.flows.FlowInfo
 import net.corda.core.flows.FlowLogic
 import net.corda.core.flows.StateMachineRunId
+import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.internal.*
 import net.corda.core.internal.concurrent.OpenFuture
@@ -147,7 +148,8 @@ class SingleThreadedStateMachineManager(
                 mutex.locked {
                     for ((_, flow) in flows) {
                         if (flowHospital.flowAffected(flow.fiber, UnknownPeerException::class.javaObjectType)) {
-                            logger.info("Scheduling flow ${flow.fiber.id} for retry.")
+                            logger.info("Received networkmap update. Scheduling flow ${flow.fiber.id}, which failed due to" +
+                                    " counterparty lookup, for retry.")
                             flow.fiber.scheduleEvent(Event.RetryFlowFromSafePoint)
                         }
                     }
@@ -412,19 +414,13 @@ class SingleThreadedStateMachineManager(
             deduplicationHandler.afterDatabaseTransaction()
             return
         }
-        val sender = serviceHub.networkMapCache.getPeerByLegalName(peer)
-        if (sender != null) {
-            when (sessionMessage) {
-                is ExistingSessionMessage -> onExistingSessionMessage(sessionMessage, deduplicationHandler, sender)
-                is InitialSessionMessage -> onSessionInit(sessionMessage, message.platformVersion, deduplicationHandler, sender)
-            }
-        } else {
-            logger.error("Unknown peer $peer in $sessionMessage")
-            throw(UnknownPeerException("Could not resolve peer $peer"))
+        when (sessionMessage) {
+            is ExistingSessionMessage -> onExistingSessionMessage(sessionMessage, deduplicationHandler, peer)
+            is InitialSessionMessage -> onSessionInit(sessionMessage, message.platformVersion, deduplicationHandler, peer)
         }
     }
 
-    private fun onExistingSessionMessage(sessionMessage: ExistingSessionMessage, deduplicationHandler: DeduplicationHandler, sender: Party) {
+    private fun onExistingSessionMessage(sessionMessage: ExistingSessionMessage, deduplicationHandler: DeduplicationHandler, peer: CordaX500Name) {
         try {
             val recipientId = sessionMessage.recipientSessionId
             val flowId = sessionToFlow[recipientId]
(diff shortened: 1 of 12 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -408,6 +420,7 @@ class SingleThreadedStateMachineManager(
             }
         } else {
             logger.error("Unknown peer $peer in $sessionMessage")
+            throw(UnknownPeerException("Could not resolve peer $peer"))
```
rick-r3: No need for parentheses.

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -141,6 +142,17 @@ class SingleThreadedStateMachineManager(
                     deliverExternalEvent(deduplicationHandler.externalCause)
                 }
             }
+
+            serviceHub.networkMapCache.changed.subscribe({
```
rick-r3: No need for parentheses when the lambda is the only parameter.

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -141,6 +142,17 @@ class SingleThreadedStateMachineManager(
...
+                mutex.locked {
+                    for ((_, flow) in flows) {
+                        if (flowHospital.flowAffected(flow.fiber, UnknownPeerException::class.javaObjectType)) {
+                            logger.info("Scheduling flow ${flow.fiber.id} for retry.")
```
rick-r3: You could say why we are retrying.

rick-r3 (changes requested):

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
rick-r3: Will this actually work?  I think this will send the exception to the Artemis handler, not the flow.

exFalso: Yeah this won't work. There's no flow to hospitalize here.
We need to delay this message delivery until the network map is updated, but we need to make sure we don't lose ordering. We could add an in-memory bounded queue which we drain on an update, and if that gets full we block delivery? Not sure

rick-r3: Is there a way we could start a flow in an errored state ?  i.e. goes straight to flow hospital before it runs a line of user flow code?

I'm thinking for initiated flows we somehow have a "lateinit" unresolved "sender" identity (no idea if that's viable - but needed because it's plumbed into the constructor etc) but then have a flag on `addAndStartFlow` that starts the flow errored rather than clean ?  I'm keen to keep building on the hospital rather than having some other kind of dead letter queues etc that we need to admin in addition to the flow hospital.  WDYT?

The other scenario is for existing sessions.  Surely if we initiated already, the sender is likely in the network map.... but if not maybe it's "easy" to delay the error until the message is almost consumed by the flow and error then).

exFalso:
There have been various discussions around this in the past couple of days. To summarize:

- We don't want to allocate resources (start flows) for unwanted peers.
- One way to achieve this is to not have the feature in this PR at all. However this has the downside that a legitimate registration will race with initial flows, causing hangs and failures. The most apparent symptom is in the notary.
- @rick-r3 suggested that another way to avoid allocating unnecessary resources is to record explicitly in the network map if a peer has been evicted, e.g. by setting their addresses to an empty list. This is effectively a tombstone that indicates we shouldn't start a flow for this peer.
- If we do this then if a peer connects and establishes a TLS connection and we don't have them in our NM cache then we know that
  1. They have a valid certificate, meaning they registered with the doorman.
  2. They haven't been evicted (unless in the rare/impossible case where we add them with no addresses from the get go).
  In other words we identified the race condition. I think in this case it is safe and justified to start a flow and wait until the next network map update.

In any case we don't think this will make GA, it requires more discussion. @mikehearn. The PR itself looks OK, although there are a couple of things I think need changing. However we shouldn't  commit more time to it before the design is decided on.

mikehearn:
Ah, sorry, I hadn't seen this - maybe make me a reviewer in future to ensure it surfaces in my dashboard.

My concern was a bit more pedestrian - there aren't supposed to be startup races between registration and flow initiation except in the production environment (where it's expected there will be human action that is driving flow starts), i.e. no different to the web where trying to link to a website so new it's still undergoing DNS propagation can fail. And if a party lookup fails, then that implies it's not there and isn't going to be there, so the flow should fail in a visible way so software can retry or surface that to the user.

exFalso:
@mikehearn I think the main issue is the way this error condition manifests, namely the flow simply hangs (although now in the notary case it will timeout eventually). I think adding this feature is a low hanging fruit for improving Corda deployment experience, even if this codepath is hit very rarely.
For example I can see this becoming a big nuisance when doing end to end testing on fresh test networks. Perhaps worth consulting devrel and/or devops for their perspective?

mnesbit:
Bogdan couldn't make this work, even though it looked simple enough at first. I would love a fix for the issues where customers make a node and with 30 seconds want to test it against the notary. This is not it unfortunately and should be closed.

bpaunescu:
@mikehearn @exFalso Indeed, this path was hit a lot during testing. However, there is a flow retry mechanism implemented that makes sure the flows don't hang and cause the node to hang as well. That mechanism didn't play well with this one which only dealt with unknown peers failure.

The general consensus was that this is no longer needed but I kept the PR open just in case. Will now close it.

PR #3337, opened 2018-06-11
CORDA-1494: Remove isTimed check from flow hospital when handling tim…

adagys:
…eout exceptions - the timeout might expire just after the TimeFlow has finished and the staff member would throw an exception.

Increase the default flow timeout value in mock network so retries don't happen – this has cause test flakiness.


thschroeter:
The `TimedFlow` import can be removed from `StaffedFlowHospital.kt`.

PR #3342, opened 2018-06-11
[CORDA-1611] Upgrade Quasar to 0.7.10

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1611


florian-f:
>  I'm surprised that we still include lib/quasar.jar in the repository.

I think it can be safely removed since it's provided by the copyQuasarJar task. If we do remove it however, we should update the documentation in the places where we tell people to use `-javaagent: lib/quasar.jar` to avoid confusion. They will wonder where they can get the jar if they haven't run one of the gradle tasks that depends on the copy task first.

prateektiwari7:
> LGTM, although I'm surprised that we still include `lib/quasar.jar` in the repository.

I think still there is lib/quasar.jar file which creates so many duplicacy. 

Can you please tell me which corda repo doesnt contain the lib/quasar.jar

prateektiwari7:
what is the alternative for not use the quasar.jar file ?

PR #3347, opened 2018-06-12
[CORDA-1612]: Fix truncated code snippets in docs PDF

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1612


docs/source/changelog.rst
```diff
@@ -7,6 +7,8 @@ release, see :doc:`upgrade-notes`.
 Unreleased
 ==========
 
+* Improved docs PDF quality. Building the docs now requires ``LaTex`` to be installed on the OS.
```
tlil: docs => documentation

docs/source/key-concepts-flows.rst
```diff
@@ -22,7 +22,7 @@ what order.
 
 Here is a visualisation of the process of agreeing a simple ledger update between Alice and Bob:
 
-.. image:: resources/flow.gif
```
tlil: We should rethink this - presumably the GIF is telling a story on the HTML doc site. Obviously needs to be static in the PDF, but in HTML it's telling a story at the moment. @joeldudleyr3 any thoughts on what we do with these two GIFs moving forward? Having them on the HTML site is fine IMHO, but doesn't translate well into a static, printable doc.

docs/source/index.rst
```diff
@@ -52,7 +50,6 @@ We look forward to seeing what you can do with Corda!
    azure-vm.rst
    aws-vm.rst
    loadtesting.rst
-
 .. toctree::
    :caption: Design docs
```
tlil: This disappears with the file renames further up; however, we _do_ want to have the design docs on the HTML site. Could we use a different TOC for HTML and PDF? Or use a topic filter for PDF?

sollecitom: Fixed

docs/make-docsite.sh
```diff
@@ -13,11 +13,10 @@ fi
 
 # TODO: The PDF rendering is pretty ugly and can be improved a lot.
```
tlil: Is this comment still true? :) It's better now, no? ;)

docs/source/key-concepts-flows.rst
sollecitom: It does.

tlil: Yes, this comment was from before you made the change. I've approved now ;P

PR #3351, opened 2018-06-12
ENT-2053: Ensure Artemis cleanup actions run on client disconnect.

mnesbit:
If the Artemis connectionTTL is not set then not all cleanup actions happen on client process death. This in particular prevents durable messages being replayed to the client, but may include other resources not being cleaned. We did try to set the connectionTTLOverride server side, but that caused periodic disconnects.

I have also modified the RPC setting of this. I am happy to debate if this should be changed in this PR, or not.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
+++ b/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
@@ -15,11 +15,7 @@ import net.corda.core.serialization.SingletonSerializeAsToken
 import net.corda.core.serialization.deserialize
 import net.corda.core.serialization.internal.nodeSerializationEnv
 import net.corda.core.serialization.serialize
-import net.corda.core.utilities.ByteSequence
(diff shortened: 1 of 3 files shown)


node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -153,17 +142,24 @@ class P2PMessagingClient(val config: NodeConfiguration,
             locator = ActiveMQClient.createServerLocatorWithoutHA(tcpTransport).apply {
                 // Never time out on our loopback Artemis connections. If we switch back to using the InVM transport this
                 // would be the default and the two lines below can be deleted.
-                connectionTTL = -1
+                connectionTTL = 60000
                 clientFailureCheckPeriod = -1
                 minLargeMessageSize = maxMessageSize + JOURNAL_HEADER_SIZE
                 isUseGlobalPools = nodeSerializationEnv != null
+            }
+            try {
```
exFalso: Remove this

PR #3354, opened 2018-06-12
Fixed double JAR publishing problem with network bootstrapper.

sollecitom:


tools/bootstrapper/build.gradle
```diff
@@ -5,7 +5,12 @@ apply plugin: 'com.jfrog.artifactory'
 description 'Network bootstrapper'
 
 configurations {
-    runtimeArtifacts
+    runtimeArtifacts.extendsFrom runtime
+    capsuleRuntime
```
Clintonio: Remove this.l

josecoll:
The naming all looks fairly inconsistent now:

corda-tools-blob-inspector (was **tools-blob-inspector**)
archiveName = "blob-inspector

corda-tools-network-bootstrapper (was **tools-network-bootstrapper**)
archiveName "tools-network-bootstrapper (was **network-bootstrapper**)

@tlil  wdyt?

tlil:
Hmm, didn't see this until now. Agree, @josecoll . We should make this more consistent.

josecoll:
So if the fix to this problem was to add

```
jar {
    baseName "corda-tools-network-bootstrapper"
}
```

@sollecitom why did you feel the need to rename the artifacts ?

sollecitom:
As discussed with you and @Clintonio , this seems to be the convention for our published artifacts. Mind that the name of the local JAR (the one that stays unpublished to even the local maven repo, and lives in the `libs` folder) is of little importance.

PR #3356, opened 2018-06-12
CORDA-1609 - Don't use reserved keyword as method name

tlil:
As reported in [CORDA-1609](https://r3-cev.atlassian.net/browse/CORDA-1609), `CordaRPCClientConfiguration.default` is not accessible from Java since `default` is a reserved keyword.

As part of the refactor made in #2831, `CordaRPCClientConfiguration` went from being a data class to an interface with a backing implementation of type `CordaRPCClientConfigurationImpl`.

This resulted in Java users having to rewrite code that was on the form:

```java
final CordaRPCClient client = new CordaRPCClient(
    nodeAddress, CordaRPCClientConfiguration.DEFAULT
);
```

to something like this:

```java
final CordaRPCClient client = new CordaRPCClient(
    nodeAddress, CordaRPCClientConfiguration.Companion.default()
);
```

However, this does not work. The user would get a compilation error because `default` is a reserved keyword in Java.

Since `CordaRPCClientConfiguration` has been made an interface, there is no easy way of introducing a static final field on the interface from Kotlin.

Consequently, I've changed the name of the function `default()` to `getDefault()`, which is now accessible though the `CordaRPCClientConfiguration.Companion` object. In other words, the user would now have to write:

```java
final CordaRPCClient client = new CordaRPCClient(
    nodeAddress, CordaRPCClientConfiguration.Companion.getDefault()
);
```

It should be noted that `getDefault()` is currently only used internally to pass in default values in `CordaRPCClient.kt` and `CordaRPCClientUtils.kt`. It dereferences `CordaRPCClientConfigurationImpl.DEFAULT`, which is in an internal package.

The latter means that in the above example, the user would actually not have to provide the parameter at all:

```java
final CordaRPCClient client = new CordaRPCClient(nodeAddress);
```

As can be seen from the definition of `CordaRPCClient`:

```kotlin
class CordaRPCClient private constructor(
        // (...)
) {
    @JvmOverloads
    constructor(hostAndPort: NetworkHostAndPort,
                configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.getDefault())
            : this(hostAndPort, configuration, null)
```

The mentioned [refactor](https://github.com/corda/corda/commit/7a077e76f0cdf457cd40033fb0da594805a5951e#diff-0948c125db93a22263eb81eaf3161c17R65) did not make it into the 3.1 release, so from an API-stability perspective, this can be applied without affecting that commitment.

### DIFF
--- a/client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
+++ b/client/rpc/src/integration-test/kotlin/net/corda/client/rpc/RPCStabilityTests.kt
@@ -106,7 +106,7 @@ class RPCStabilityTests {
                 Try.on {
                     startRpcClient<RPCOps>(
                             server.get().broker.hostAndPort!!,
-                            configuration = CordaRPCClientConfigurationImpl.default.copy(minimumServerProtocolVersion = 1)
+                            configuration = CordaRPCClientConfigurationImpl.DEFAULT.copy(minimumServerProtocolVersion = 1)
                     ).get()
                 }
             }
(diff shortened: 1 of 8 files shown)


sollecitom:
You need to update the API file.

tlil:
Indeed; I was first going off the instructions listed here: https://github.com/corda/corda/tree/master/.ci, but obviously this breaks the gate.

Will update and add Mike for approval.

tlil:
@sollecitom - i've synced up with @mikehearn on this. Plan of action is:

(1) Revert `CordaRPCClientConfiguration` back to a class so that it's backwards compatible with 3.0 and 3.1
(2) Reset `api-current.txt` to what was in 3.1
(3) @mikehearn will approve and bypass the gate to get the API checker back on track

tlil:
https://github.com/corda/corda/pull/3359

PR #3359, opened 2018-06-13
CORDA-1609 - Don't use reserved keyword as method name

tlil:
As reported in [CORDA-1609](https://r3-cev.atlassian.net/browse/CORDA-1609), `CordaRPCClientConfiguration.default` is not accessible from Java since `default` is a reserved keyword.

As part of the refactor made in #2831, `CordaRPCClientConfiguration` went from being a data class to an interface with a backing implementation of type `CordaRPCClientConfigurationImpl`.

This resulted in Java users having to rewrite code that was on the form:

```java
final CordaRPCClient client = new CordaRPCClient(
    nodeAddress, CordaRPCClientConfiguration.DEFAULT
);
```

to something like this:

```java
final CordaRPCClient client = new CordaRPCClient(
    nodeAddress, CordaRPCClientConfiguration.Companion.default()
);
```

However, this does not work. The user would get a compilation error because `default` is a reserved keyword in Java.

Since `CordaRPCClientConfiguration` has been made an interface, there is no easy way of introducing a static final field on the type from Kotlin.

Consequently, I've changed this back to using a `class` with a static field named `DEFAULT` instead of the static method `default()`.

It should be noted that `default()` / `DEFAULT` is currently only used internally to pass in default values in `CordaRPCClient.kt` and `CordaRPCClientUtils.kt`.  This means that in the above example, the user would actually not have to provide the parameter at all:

```java
final CordaRPCClient client = new CordaRPCClient(nodeAddress);
```

As can be seen from the definition of `CordaRPCClient`:

```kotlin
class CordaRPCClient private constructor(...) {
    @JvmOverloads
    constructor(
        hostAndPort: NetworkHostAndPort,
        configuration: CordaRPCClientConfiguration = CordaRPCClientConfiguration.DEFAULT
    ) : this(hostAndPort, configuration, null)
```

That said, the field/method is exposed as part of our API surface and consequently shouldn't be broken.

The mentioned [refactor](7a077e7#diff-0948c125db93a22263eb81eaf3161c17R65) did not make it into the 3.1 release, so from an API-stability perspective, the change in this PR, effectively reverting the previous refactor, can be applied without affecting our commitment to provide a backwards compatible API.


tlil:
See https://github.com/corda/corda/compare/caee24a...78f5d95 for the effective change from the API as committed in V3.1.

.ci/api-current.txt
```diff
@@ -6527,27 +6527,69 @@ public static final class net.corda.client.rpc.CordaRPCClient$Companion extends
   @NotNull
   public final net.corda.client.rpc.CordaRPCClient createWithSsl(net.corda.core.utilities.NetworkHostAndPort, net.corda.core.messaging.ClientRpcSslOptions, net.corda.client.rpc.CordaRPCClientConfiguration)
 ##
-public interface net.corda.client.rpc.CordaRPCClientConfiguration
-  public abstract int getCacheConcurrencyLevel()
+public class net.corda.client.rpc.CordaRPCClientConfiguration extends java.lang.Object
+  public <init>()
+  public <init>(java.time.Duration)
+  public <init>(java.time.Duration, int)
```
anthonykeenan: All these extra constructors are 'new' api which aren't actually _committed_ to yet, I think we should just take the version in `api-current.txt` back to the comitted version on the v3 branch, which is [here](https://github.com/corda/corda/blob/release-V3/.ci/api-current.txt#L6630)

tlil: Is that useable though? On the V3 branch, you have the following constructor:

https://github.com/corda/corda/blob/371031dd3b3c3ccc49a3f40a0fc47e9ffa0c6517/.ci/api-current.txt#L6630-L6631

Meaning that you can only set the `connectionMaxRetryInterval`, and nothing else. Besides, this is a reflection of the actual class file. I mean, we can remove `JvmOverloads` on the constructor and add the one for `init(connectionMaxRetryInterval)` explicitly, and for any further alteration rely on the user extending from `CordaRPCClientConfiguration`. But that's not any cleaner, is it?

tlil: Anthony and I discussed this on Slack, and I misunderstood. So his comment was to not add the extra constructors to the `api-current.txt` (yet) and keep it as close to what was in V3.1 as possible and leave the update to the release :+1: Will update.

.ci/api-current.txt
```diff
@@ -6527,27 +6527,23 @@ public static final class net.corda.client.rpc.CordaRPCClient$Companion extends
...
+  public boolean equals(Object)
   @NotNull
-  public abstract java.time.Duration getDeduplicationCacheExpiry()
-  public abstract int getMaxFileSize()
-  public abstract int getMaxReconnectAttempts()
-  public abstract int getMinimumServerProtocolVersion()
-  public abstract int getObservationExecutorPoolSize()
+  public java.time.Duration getConnectionMaxRetryInterval()
+  public int hashCode()
   @NotNull
```
anthonykeenan: Just a minor point - the `toString()` didn't have `@NotNull` originally
Interestingly, overriding the toString with the Kotlin String type, causes the @NotNull attribute that wasn't there before. However it wasn't part of the original api.

tlil: Yeah, due to this joyful quirk:

image

I've removed it from `.ci/api-current.txt` now.

PR #3368, opened 2018-06-14
[CORDA-1634] Destroy child processes when parent exits.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1634

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
@@ -217,7 +217,7 @@ abstract class FlowLogic<out T> {
      * is offline then message delivery will be retried until it comes back or until the message is older than the
      * network's event horizon time.
      */
-    @Deprecated("Use FlowSession.send()", level = DeprecationLevel.WARNING)
+    @Deprecated("Use FlowgSession.send()", level = DeprecationLevel.WARNING)
     @Suspendable
     open fun send(otherParty: Party, payload: Any) {
         getDeprecatedSessionForParty(otherParty).send(payload)
--- a/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
@@ -217,7 +217,7 @@ abstract class FlowLogic<out T> {
      * is offline then message delivery will be retried until it comes back or until the message is older than the
      * network's event horizon time.
      */
-    @Deprecated("Use FlowgSession.send()", level = DeprecationLevel.WARNING)
+    @Deprecated("Use FlowSession.send()", level = DeprecationLevel.WARNING)
     @Suspendable
     open fun send(otherParty: Party, payload: Any) {
         getDeprecatedSessionForParty(otherParty).send(payload)
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
@@ -653,10 +653,8 @@ class DriverDSLImpl(
             val debugPort = if (isDebug) debugPortAllocation.nextPort() else null
             val monitorPort = if (jmxPolicy.startJmxHttpServer) jmxPolicy.jmxHttpServerPortAllocation?.nextPort() else null
             val process = startOutOfProcessNode(config, quasarJarPath, debugPort, jolokiaJarPath, monitorPort, systemProperties, cordappPackages, maximumHeapSize)
-
+            addShutdownHook { process.destroy() }
             if (waitForAllNodesToFinish) {
-                // Destroy the child process when the parent terminates so it is not orphaned.
-                addShutdownHook { process.destroy() }
                 state.locked {
                     processes += object : Waitable {
                         override fun waitFor() {
@@ -664,8 +662,6 @@ class DriverDSLImpl(
                         }
                     }
                 }
-            } else {
-                shutdownManager.registerProcessShutdown(process)
             }
             val p2pReadyFuture = addressMustBeBoundFuture(executorService, config.corda.p2pAddress, process)
             return p2pReadyFuture.flatMap {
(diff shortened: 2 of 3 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -653,7 +653,10 @@ class DriverDSLImpl(
             val debugPort = if (isDebug) debugPortAllocation.nextPort() else null
             val monitorPort = if (jmxPolicy.startJmxHttpServer) jmxPolicy.jmxHttpServerPortAllocation?.nextPort() else null
             val process = startOutOfProcessNode(config, quasarJarPath, debugPort, jolokiaJarPath, monitorPort, systemProperties, cordappPackages, maximumHeapSize)
+
             if (waitForAllNodesToFinish) {
+                // Destroy the child process when the parent terminates so it is not orphaned.
+                addShutdownHook { process.destroy() }
```
vkolomeyko: I can see default setting for `waitForAllNodesToFinish` as:
```
val waitForAllNodesToFinish: Boolean = false
```
Are we sure we want to add shutdown hook only when `waitForAllNodesToFinish` explicitly set to `true` ?

florian-f: We might want to do that in the other cases as well.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -653,6 +653,7 @@ class DriverDSLImpl(
             val debugPort = if (isDebug) debugPortAllocation.nextPort() else null
             val monitorPort = if (jmxPolicy.startJmxHttpServer) jmxPolicy.jmxHttpServerPortAllocation?.nextPort() else null
             val process = startOutOfProcessNode(config, quasarJarPath, debugPort, jolokiaJarPath, monitorPort, systemProperties, cordappPackages, maximumHeapSize)
+            addShutdownHook { process.destroy() }
```
vkolomeyko: Didn't we agree to use `shutdownManager.registerProcessShutdown(process)` here ?

Also I would add a comment explaining why this is needed even when `waitForAllNodesToFinish` is `true`.

florian-f: > Didn't we agree to use shutdownManager.registerProcessShutdown(process) here ?
I tried that, it did not work as expected.

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -217,7 +217,7 @@ abstract class FlowLogic<out T> {
      * is offline then message delivery will be retried until it comes back or until the message is older than the
      * network's event horizon time.
      */
-    @Deprecated("Use FlowSession.send()", level = DeprecationLevel.WARNING)
+    @Deprecated("Use FlowgSession.send()", level = DeprecationLevel.WARNING)
```
shamsasari: Revert

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -653,6 +653,10 @@ class DriverDSLImpl(
             val debugPort = if (isDebug) debugPortAllocation.nextPort() else null
             val monitorPort = if (jmxPolicy.startJmxHttpServer) jmxPolicy.jmxHttpServerPortAllocation?.nextPort() else null
             val process = startOutOfProcessNode(config, quasarJarPath, debugPort, jolokiaJarPath, monitorPort, systemProperties, cordappPackages, maximumHeapSize)
+            // Destroy the child process when the parent exits.This is needed even when `waitForAllNodesToFinish` is
+            // true because we don't want orphaned processes in the case that the parent process is terminated by the
+            // user, for example when the `tools:explorer:runDemoNodes` gradle task is stopped with CTRL-C.
+            addShutdownHook { process.destroy() }
```
shamsasari: Instead of adding the hook here, have `ShutdownManager.registerProcessShutdown` add the hook, as an additional backup.

florian-f: > Instead of adding the hook here, have ShutdownManager.registerProcessShutdown add the hook, as an additional backup.

That's much better, thanks.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/ShutdownManager.kt
```diff
@@ -70,6 +71,7 @@ class ShutdownManager(private val executorService: ExecutorService) {
     fun registerShutdown(shutdown: () -> Unit) = registerShutdown(doneFuture(shutdown))
 
     fun registerProcessShutdown(process: Process) {
+        addShutdownHook { process.destroy() }
```
shamsasari: Actually the shutdown logic below seems better. Can you extract it out into a private method and have both `addShutdownHook` and `registerShutdown` execute it.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -653,6 +654,9 @@ class DriverDSLImpl(
             val debugPort = if (isDebug) debugPortAllocation.nextPort() else null
             val monitorPort = if (jmxPolicy.startJmxHttpServer) jmxPolicy.jmxHttpServerPortAllocation?.nextPort() else null
             val process = startOutOfProcessNode(config, quasarJarPath, debugPort, jolokiaJarPath, monitorPort, systemProperties, cordappPackages, maximumHeapSize)
+
+            shutdownManager.registerProcessShutdown(process)
```
shamsasari: Perhaps just a small comment on how Ctrl+C can cause the child processes to linger, so the shutdown register has to always added.

PR #3369, opened 2018-06-14
[CORDA-1515] Clarify distinction between non-validating and validating notaries

kchalkias:
Although our documentation discusses notary validation, there was not a clear distinction between them.
Also added that notaries act as timestamping authorities as well.

### DIFF
--- a/docs/source/key-concepts-notaries.rst
+++ b/docs/source/key-concepts-notaries.rst
@@ -4,7 +4,8 @@ Notaries
 .. topic:: Summary
 
    * *Notary clusters prevent "double-spends"*
-   * *Notary clusters may optionally also validate transactions*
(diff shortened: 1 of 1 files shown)


docs/source/key-concepts-notaries.rst
```diff
@@ -4,7 +4,8 @@ Notaries
 .. topic:: Summary
 
    * *Notary clusters prevent "double-spends"*
-   * *Notary clusters may optionally also validate transactions*
+   * *Notary clusters are also time-stamping authorities. If a transaction includes a time-window, it can only be notarised during that window*
+   * *Notary clusters may optionally also validate transactions (referenced as "validating" and "non-validating" notaries, respectively)*
```
sollecitom: The part in the parentheses is not so clear. Maybe something like "Notary clusters may optionally also validate transactions, in which case they are called "validating" notaries, as opposed to "non-validating"".

PR #3383, opened 2018-06-15
ENT-1463: Upgrade to Kotlin 1.2.50

chrisr3:
Kotlin 1.2.50 allows `@JvmDefault` also to generate the previous `$DefaultImpls` classes, and therefore preserve binary compatibility with earlier Corda releases.

The updated compiler option also requires everyone to update the IntelliJ Kotlin plugin to 1.2.50.


chrisr3:
Adding `@JvmDefault` to an interface method means that it's no longer `abstract`, and hence its implementation can be inherited. There's no _actual_ change here.

tlil:
Yep, exactly. :+1:

PR #3384, opened 2018-06-15
ENT-2053 Fix messaging TTL related settings causing disconnects

rick-r3:


rick-r3:
Thinking we won't merge this since we found a better solution that @bpaunescu is testing.

rick-r3:
This is now the better solution, although tracing back the original reasoning for these settings, it seems to be for a "demo mode".

PR #3386, opened 2018-06-15
Deterministic JVM

tlil:
### Introduction

The code in the DJVM module has not yet been integrated with the rest of the platform. It will eventually become a part of the node and enforce deterministic and secure execution of smart contract code, which is mobile and may propagate around the network without human intervention.

**NOTE**: Currently, it stands alone as an evaluation version. We want to give developers the ability to start trying it out and get used to developing deterministic code under the set of constraints that we envision will be placed on contract code in the future.

### Documentation

For details around the concept, implementation, ruleset, scope, etc., please refer to the documentation site: **docs/source/key-concepts-djvm.rst**

Related blog posts:
 * [Deterministic Contract Verification](https://medium.com/corda/deterministic-contract-verification-97feb54c141e) - Medium
 * [DJVM Behind the Scenes](https://medium.com/corda/djvm-behind-the-scenes-2ba7f5cb9275) - Medium


chrisr3 (changes requested):
I'm going to have do work through this in "bite-sized" chunks.

djvm/shell/djvm
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
tlil: Yes, this is intended to run in a Bash shell. I've not provided any equivalents for Windows as it's still possible to run through Cygwin, MinGW or similar environments. 🐧

djvm/build.gradle
```diff
@@ -0,0 +1,58 @@
+apply plugin: 'java'
+apply plugin: 'kotlin'
+
+buildscript {
+    ext.asm_version = '6.2'
```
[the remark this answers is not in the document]
tlil: Yes, you're right. Good points. It was an oversight on my part as I moved it out of the experimental section. Will fix.

djvm/src/main/kotlin/net/corda/sandbox/source/JarInputStreamIterator.kt
```diff
@@ -0,0 +1,70 @@
...
+     * Return the next class entry in the JAR stream, in the form of an input stream.
+     */
+    override fun next(): InputStream {
+        return jarInputStream!!
+    }
+
+    /**
+     * Close the JAR input stream when it has been traversed.
+     */
+    override fun close() {
```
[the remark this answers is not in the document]
tlil: yeah, a bit shorter, innit :) nice one

djvm/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+    compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
+    compile "org.slf4j:slf4j-nop:$slf4j_version"
+    compile "com.jcabi:jcabi-manifests:$jcabi_manifests_version"
+
+    // ASM: byte code manipulation library
+    compile "org.ow2.asm:asm:$asm_version"
+    compile "org.ow2.asm:asm-tree:$asm_version"
+    compile "org.ow2.asm:asm-util:$asm_version"
+    compile "org.ow2.asm:asm-commons:$asm_version"
+
```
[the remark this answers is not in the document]
tlil: Nope, don't need it. Was lingering from the experimental module. Will remove.

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassResolver.kt
```diff
@@ -0,0 +1,109 @@
...
+    private val pinnedClassesAndSandbox =
+            whitelist + pinnedClasses + Regex("^$sandboxPrefix.*$")
+
+    /**
+     * Resolve the class name from a fully qualified name.
+     */
+    fun resolve(name: String): String {
+        return when {
+            name.startsWith("[") && name.endsWith(";") -> {
+                Regex("^(\\[+)L(.*);$").replace(name) {
```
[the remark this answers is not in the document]
tlil: Indeed - oversight on my end. Fixed one in `ClassModule` as well.

djvm/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+    }
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+    }
+}
+
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+    compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
+    compile "org.slf4j:slf4j-nop:$slf4j_version"
```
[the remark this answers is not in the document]
tlil: Not yet - @exFalso, want to file a follow-up story to add logging? Or add to scope of this PR?

djvm/build.gradle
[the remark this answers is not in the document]
chrisr3: ASM 5.0.4 is a transitive dependency of Kryo. ClassCarpenter also uses ASM, but that's home-grown code and so can be updated (if necessary).

djvm/src/test/kotlin/net/corda/sandbox/TestBase.kt
```diff
@@ -0,0 +1,177 @@
+package net.corda.sandbox
+
```
chrisr3: Do we need a `TestBase` class? Aren't there alternatives to using inheritance, e.g. JUnit `@Rule` and `@ClassRule` for shared code, or even just simple helper classes?

chrisr3 (dismissed):
Now for round two :grin:.
- Nice to see lots of unit tests :smile:. Have you also tested with Kotlin-specific "package" classes, i.e. a `*Kt.class` file that doesn't actually contain a `class` definition?
- Are you still using the static version of the deterministic `rt.jar` that I sent you a while back? Because if so, you can now move to the
```
net.corda:deterministic-rt:${deterministic_rt_version}:api
```
JAR that's been uploaded to the artifactory.

docs/source/key-concepts-djvm.rst
```diff
@@ -0,0 +1,378 @@
...
+tries to catch such exceptions, as doing so would allow the user to bypass the thresholds set out in the execution
+profile.
+
+
+Only Allow Explicitly Whitelisted Runtime API
+.............................................
+
+Ensures that constant pool references are mapped against a verified subset of the Java runtime libraries. Said subset
+excludes functionality that contract code should not have access to, such as file I/O or external entropy. In future
+versions, this whitelist will be trimmed down to the bare minimum needed, essentially a subset of ``java.lang``, so
```
chrisr3: Surely a subset of `java.*` rather than `java.lang.*`? Maybe even some `javax.*` classes too? There's a fair bit of useful functionality in `java,util`, and even `java.io` contains things like `ByteArray{Input|Output}Stream`, `Closeable`, `FileNotFoundException` etc that we might want to use. And then there's `java.security.` ...

docs/source/key-concepts-djvm.rst
```diff
@@ -0,0 +1,378 @@
...
+
+
+Disallow Native Methods
+.......................
+
+Forbids native methods as these provide the user access into operating system functionality such as file handling,
+network requests, general hardware interaction, threading, *etc.* These all constitute sources of non-determinism, and
+allowing such code to be called arbitrarily from the JVM would require deterministic guarantees on the native machine
+code level. This falls out of scope for the DJVM.
+
```
chrisr3: We're going to need _some_ native methods for e.g. array handling. Nor am I convinced that we can remove classes like `sun.misc.Unsafe`, although we should obviously prevent them from being invoked _directly_ from contract code.

docs/source/key-concepts-djvm.rst
```diff
@@ -0,0 +1,378 @@
...
+Runtime Costing
+~~~~~~~~~~~~~~~
+
+The runtime accountant inserts calls to an accounting object before expensive byte code. The goal of this rewrite is to
+deterministically terminate code that has run for an unacceptably long amount of time or used an unacceptable amount of
+memory. Types of expensive byte code include method invocation, memory allocation, branching and exception throwing.
+
+The cost instrumentation strategy used is a simple one: just counting byte code that are known to be expensive to
+execute. Method size can be limited and jumps count towards the budget, so such a strategy is guaranteed to eventually
```
chrisr3: Boldly splitting infinitives that no-one has split before... :vulcan_salute:

docs/source/key-concepts-djvm.rst
```diff
@@ -0,0 +1,378 @@
...
+terminate. However it is still possible to construct byte code sequences by hand that take excessive amounts of time to
+execute. The cost instrumentation is designed to ensure that infinite loops are terminated and that if the cost of
+verifying a transaction becomes unexpectedly large (*e.g.*, contains algorithms with complexity exponential in
+transaction size) that all nodes agree precisely on when to quit. It is not intended as a protection against denial of
+service attacks. If a node is sending you transactions that appear designed to simply waste your CPU time then simply
+blocking that node is sufficient to solve the problem, given the lack of global broadcast.
+
+The budgets are separate per operation code type, so there is no unified cost model. Additionally the instrumentation is
+high overhead. A more sophisticated design would be to statically calculate byte code costs as much as possible ahead of
```
chrisr3: "... to calculate byte code costs statically..." :vulcan_salute:

djvm/build.gradle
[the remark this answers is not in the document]
exFalso: One way we could circumvent this is to make the djvm a standalone project and include it with `includeBuild` instead of `include` in settings.gradle. See https://docs.gradle.org/current/userguide/composite_builds.html

djvm/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+    // CLI utilities
+    compile "info.picocli:picocli:$picocli_version"
+
+    // Test utilities
+    testCompile "junit:junit:$junit_version"
+    testCompile "org.assertj:assertj-core:$assertj_version"
+}
+
+jar {
+    from(configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }) {
```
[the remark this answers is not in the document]
exFalso: Also, we started using a very convenient project to do the fatjaring, see for example https://github.com/corda/corda/blob/141d45c39df69591f33fbc15aeb67fd2f77672bd/experimental/avalanche/build.gradle#L40, perhaps worth switching to this

docs/source/key-concepts-djvm.rst
tlil: Copy-pasta from the tech whitepaper... Will rework this section a bit

docs/source/key-concepts-djvm.rst
tlil: Will clarify :+1:

tlil:
Re:

>Nice to see lots of unit tests 😄. Have you also tested with Kotlin-specific "package" classes, i.e. a *Kt.class file that doesn't actually contain a class definition?

Good point - I haven't so will address that.

>Are you still using the static version of the deterministic rt.jar that I sent you a while back? Because if so, you can now move to the net.corda:deterministic-rt:${deterministic_rt_version}:api. JAR that's been uploaded to the artifactory.

It's a generated whitelist based on an older version of `deterministic-rt.jar`, so I will update this to reflect the latest API from Artifactory (and make it update dynamically in the process).

docs/source/key-concepts-djvm.rst
tlil: Yeah, can update - the scope of the actual whitelist will have to be determined. The thinking was to have as much as possible also be checked by DJVM regardless of whether it exists in `rt-deterministic.jar` or not. Will update to clarify.

djvm/build.gradle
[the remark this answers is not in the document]
tlil: Done

djvm/src/main/kotlin/net/corda/sandbox/tools/Utilities.kt
```diff
@@ -0,0 +1,117 @@
...
+     * Get the path of where any generated code will be placed. Create the directory if it does not exist.
+     */
+    fun getCodePath(): Path {
+        val root = Paths.get("tmp")
+                .resolve("net")
+                .resolve("corda")
+                .resolve("sandbox")
+        Files.createDirectories(root)
+        return root
+    }
```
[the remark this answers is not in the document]
exFalso: Perhaps use `Files.createTempDirectory` instead, that will work on Windows too

djvm/src/main/kotlin/net/corda/sandbox/tools/cli/BuildCommand.kt
```diff
@@ -0,0 +1,42 @@
...
+import picocli.CommandLine.Parameters
+import java.nio.file.Path
+
+@Command(
+        name = "build",
+        description = ["Build one or more Java source files, each implementing the sandbox runnable interface " +
+                "required for execution in the deterministic sandbox."]
+)
+@Suppress("KDocMissingDocumentation")
+class BuildCommand : CommandBase() {
```
exFalso: I like this picocli library

djvm/src/main/kotlin/net/corda/sandbox/rules/implementation/AlwaysInheritFromSandboxedObject.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    override fun emit(context: EmitterContext, instruction: Instruction) = context.emit {
+        if (instruction is TypeInstruction &&
+                instruction.typeName == OBJECT_NAME) {
+            // When creating new objects, make sure the sandboxed type gets used.
+            new(SANDBOX_OBJECT_NAME, instruction.operation)
+            preventDefault()
+        }
+        if (instruction is MemberAccessInstruction &&
```
exFalso: I wonder whether it'd be beneficial to use something like `javassist` to ease bytecode reading/writing

djvm/src/main/kotlin/net/corda/sandbox/code/EmitterModule.kt
```diff
@@ -0,0 +1,125 @@
...
+ */
+@Suppress("unused")
+class EmitterModule(
+        private val methodVisitor: MethodVisitor
+) {
+
+    /**
+     * Indicates whether the default instruction in the currently processed block is to be emitted or not.
+     */
+    var emitDefaultInstruction: Boolean = true
```
exFalso: Is this essentially a global variable used while emitting code? The semantics are a bit unclear to me, it refers to emitting a "default instruction", but the use cases seem to indicate that this effects calls into superclass visitors

djvm/src/main/kotlin/net/corda/sandbox/rules/ClassRule.kt
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Representation of a rule that applies to class definitions.
+ */
+abstract class ClassRule : Rule {
+
+    /**
+     * Determine whether the rule should be applied during class rewriting. If `true`, run the rule validation
+     * function when the class is first being loaded and rewired. Otherwise, run the validation at load time within
+     * the sandbox.
```
exFalso: This isn't clear to me and it doesn't seem to be used. From what I understand all rules are run before rewriting right?

exFalso (commented):
Not done yet, just flushing a batch of commetns

djvm/src/main/kotlin/net/corda/sandbox/code/EmitterModule.kt
tlil: If this flag is unset, the superclass' visitor doesn't get called for the currently processed instruction, effectively voiding the emittance of said instruction.

In short, for each instruction, we run through the following steps:
 1. Process emitters, which each can emit instructions to the class (at the current instruction pointer)
 2. Emit original instruction, unless `emitDefaultInstruction` is false

djvm/src/main/kotlin/net/corda/sandbox/rules/ClassRule.kt
tlil: Indeed - will remove.

djvm/src/main/kotlin/net/corda/sandbox/code/EmitterModule.kt
tlil: Essentially:

https://github.com/corda/corda/blob/26212ff5894d1ed6b96c7bfa3a0a4fa46abe8477/djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassAndMemberVisitor.kt#L495-L508

djvm/src/main/kotlin/net/corda/sandbox/rules/implementation/AlwaysInheritFromSandboxedObject.kt
tlil: How do you mean? Anything in particular that you had in mind? Either way, I'd be keen to punt on that to a follow-up ticket.

djvm/src/main/kotlin/net/corda/sandbox/tools/Utilities.kt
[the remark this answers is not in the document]
tlil: This `tmp` directory is relative to the current working directory, so it works on Windows as well. In other words, we're not creating `/tmp/...` here, but rather `$(pwd)/tmp/...`. 😉

djvm/src/main/kotlin/net/corda/sandbox/source/ClassSource.kt
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The source of one or more compiled Java classes.
+ *
+ * @property qualifiedClassName The fully qualified class name.
+ * @property origin The origin of the class source, if any.
+ */
+@Suppress("unused")
+open class ClassSource protected constructor(
+        open val qualifiedClassName: String = "",
```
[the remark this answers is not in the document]
tlil: This was not used in the sources that could have multiple classes. I've removed the dependency between `ClassSource` and `PathClassSource` to avoid that confusion. For `ClassSource` it is now truly the qualified name :)

djvm/src/main/kotlin/net/corda/sandbox/references/ReferenceMap.kt
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Map from member references to all discovered call-sites / field accesses for each reference.
+ */
+class ReferenceMap(
+        private val classModule: ClassModule
+) : Iterable<EntityReference> {
+
+    private val entities = ConcurrentLinkedQueue<EntityReference>()
+
+    private val map: MutableMap<EntityReference, MutableSet<SourceLocation>> = hashMapOf()
```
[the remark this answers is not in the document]
tlil: Hmm, not quite. `entities` are used to guarantee safe traversal and processing of the collection of references whilst still allowing updates to be made. 

The fact that we're maintaining both `mapByLocation` and `map` is to speed up keying into the set of references per location and locations per reference, respectively. The former is accessible through `from()` whereas the latter can be retrieved via `get()`.

I've renamed these fields and methods to clarify.

djvm/src/main/kotlin/net/corda/sandbox/rules/implementation/AlwaysInheritFromSandboxedObject.kt
exFalso: Ah this was just a general wonderingment

djvm/src/main/kotlin/net/corda/sandbox/code/EmitterModule.kt
tlil: Yep. ASM's visitor pattern at play ;)

djvm/src/main/kotlin/net/corda/sandbox/references/ReferenceMap.kt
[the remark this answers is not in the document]
exFalso: What I meant was calling `process` drains the queue, which means subsequent `process` calls won't do the same thing, however `size` and `get` still return the same values. One way of mitigating this is to simply iterate the queue instead of draining.
But it's only used in a single place, so perhaps it's fine, I was just a bit confused trying to figure out the api contract:)

djvm/src/main/kotlin/net/corda/sandbox/source/JarInputStreamIterator.kt
```diff
@@ -0,0 +1,62 @@
...
+import java.util.jar.JarInputStream
+
+/**
+ * Iterator for traversing over a JAR stream.
+ *
+ * @param inputStream The input stream representing the JAR.
+ */
+class JarInputStreamIterator(
+        inputStream: InputStream
+) : Iterator<InputStream>, AutoCloseable {
```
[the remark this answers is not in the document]
exFalso: I meant that the type's contract indicates that we could do something like this:
```kotlin
  val streamOfStreams: Iterator<InputStream> = ...
  val threadPool: Executor = ...
  for (steam in streamOfStreams) {
    threadPool.execute { processStream(stream) }
  }
```
which would be incorrect. An `Iterator<InputStream>` indicates that the elements are independent and `Iterator` is parametric, it doesn't indicate that the underlying streams are actually linked, and also affect the enclosing `Iterator`.

In other words the `Iterator` contract is much more liberal than the structure that `nextEntry -> next` extracts.

Anyway, the only reason I'm bringing this up is because I have ptsd from another project that used stream of streams everywhere and it was completely incomprehensible :D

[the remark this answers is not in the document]
tlil: Yes, I know. Seems like that's how you roll with this Jar stream though, unless you want to constrain usage and only surface the resulting `ByteArray`s or `Class<*>`s. Instead of the stream. The latter would require the iterator class to have some knowledge of context. Let's take this convo offline. You're sitting right next to me after all. :-)

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassResolver.kt
```diff
@@ -0,0 +1,111 @@
...
+        private val sandboxPrefix: String
+) {
+
+    private val complexArrayTypeRegex = "^(\\[+)L(.*);$".toRegex()
+
+    /**
+     * Classes and packages that are either pinned or already in the sandbox.
+     */
+    private val pinnedClassesAndSandbox =
+            whitelist + pinnedClasses + Regex("^$sandboxPrefix.*$")
```
exFalso: It's still unclear to me what the relation between the whitelist and pinned classes is. In some places they seem to be used interchangably, in others it seems pinned classes are a superset of the whitelist (e.g. in Whitelist.kt), here it seems they can be separate. In SandboxExecutor it seems whitelisted classes are not loaded at all during validation, and yet in SandboxClassLoader there is an extra check to skip analysing whitelisted classes.

Can you add a comment about what the difference is somewhere please? My understanding was that pinned classes are ones that don't get rewritten and whitelisted classes are ones that don't get validated. However for example in this file `resolveName()` indicates that whitelisted classes also don't get remapped.

exFalso (commented):
Second pass done.

All in all this code is very readable! The Rule/Emitter/Provider separation is very neat. 

The only parts that were a bit unclear to me were the high-level traversal order of classes/references and the pinned class/whitelist interaction, I think these could be made more explicit. Some clarification/docs on these would be nice.

Other than that I'm happy with this PR, great stuff!

djvm/src/main/kotlin/net/corda/sandbox/SandboxRuntimeContext.kt
```diff
@@ -0,0 +1,60 @@
...
+     * A summary of the currently accumulated runtime costs (for, e.g., memory allocations, invocations, etc.).
+     */
+    val runtimeCosts = RuntimeCostSummary(configuration.executionProfile)
+
+    /**
+     * Run a set of actions within the provided sandbox context.
+     */
+    fun use(action: SandboxRuntimeContext.() -> Unit) {
+        this.apply {
+            SandboxRuntimeContext.INSTANCE = this
```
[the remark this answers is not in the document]
tlil: Nope, that should never happen. I could assert on it, but wouldn't matter anyway.

djvm/src/main/kotlin/net/corda/sandbox/analysis/AnalysisConfiguration.kt
```diff
@@ -0,0 +1,36 @@
...
+ * @property analyzeAnnotations Analyze annotations despite not being explicitly referenced.
+ * @property prefixFilters Only record messages where the originating class name matches one of the provided prefixes.
+ * If none are provided, all messages will be reported.
+ * @property classModule Module for handling evolution of a class hierarchy during analysis.
+ * @property memberModule Module for handling the specification and inspection of class members.
+ */
+class AnalysisConfiguration(
+        val whitelist: Whitelist = Whitelist.MINIMAL,
+        val pinnedClasses: Whitelist = Whitelist.PINNED_CLASSES,
+        private val sandboxPrefix: String = "sandbox/",
```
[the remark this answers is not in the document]
tlil: The binary package names are represented with slashes `/` as separators rather than dots `.`. 

And the slash could be appended in the `init` block for input without it, as you say. That said, this is a const really anyway, so will move it to being that instead.

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassAndMemberVisitor.kt
```diff
@@ -0,0 +1,540 @@
...
+     * Record a message derived from a [Throwable].
+     */
+    private fun recordMessage(exception: Throwable, context: AnalysisRuntimeContext) {
+        context.messages.add(Message.fromThrowable(exception, context.location))
+    }
+
+    /**
+     * Record a reference to a class.
+     */
+    private fun addReferenceToType(type: String) {
```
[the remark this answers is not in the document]
tlil: I'll change to `recordTypeReference`

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassAndMemberVisitor.kt
```diff
@@ -0,0 +1,540 @@
...
+        if (shouldBeProcessed(currentClass!!.name)) {
+            val classReference = ClassReference(typeName)
+            analysisContext.references.add(classReference, sourceLocation)
+        }
+    }
+
+    /**
+     * Record a reference to a class member.
+     */
+    private fun addReferenceToMember(owner: String, name: String, desc: String) {
```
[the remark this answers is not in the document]
tlil: I'll change to `recordMemberReference`

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassAndMemberVisitor.kt
```diff
@@ -0,0 +1,540 @@
...
+
+    /**
+     * Extract information about the traversed field.
+     */
+    open fun visitField(clazz: Class, field: Member) = field
+
+    /**
+     * Extract information about the traversed instruction.
+     */
+    open fun visitInstruction(method: Member, emitter: EmitterModule, instruction: Instruction) {}
```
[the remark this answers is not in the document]
tlil: It has access to the class through `method.className`

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassAndMemberVisitor.kt
```diff
@@ -0,0 +1,540 @@
...
+
+        /**
+         * Extract information about the traversed method.
+         */
+        override fun visitMethod(
+                access: Int, name: String, desc: String, signature: String?, exceptions: Array<out String>?
+        ): MethodVisitor? {
+            var visitedMember: Member? = null
+            var processMember = true
+            currentClass?.apply {
```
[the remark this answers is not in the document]
tlil: I've tidied up in this, yeah. Good call.

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassResolver.kt
```diff
@@ -0,0 +1,111 @@
...
+     */
+    fun resolveDescriptor(descriptor: String): String {
+        val outputDescriptor = StringBuilder()
+        var longName = StringBuilder()
+        var isProcessingLongName = false
+        loop@ for (char in descriptor) {
+            when {
+                char != ';' && isProcessingLongName -> {
+                    longName.append(char)
+                    continue@loop
```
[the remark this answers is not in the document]
tlil: Doesn't work since we're inside a `when` block

djvm/src/main/kotlin/net/corda/sandbox/costing/RuntimeCostSummary.kt
```diff
@@ -0,0 +1,68 @@
...
+            jumpCostThreshold = profile.jumpCostThreshold,
+            invocationCostThreshold = profile.invocationCostThreshold,
+            throwCostThreshold = profile.throwCostThreshold
+    )
+
+    /**
+     * Accumulated cost of memory allocations.
+     */
+    val allocationCost = RuntimeCost(allocationCostThreshold) {
+        "Contract [${it.name}] terminated due to over-allocation"
```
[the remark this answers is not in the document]
tlil: Yep, good point

djvm/src/main/kotlin/net/corda/sandbox/analysis/ClassAndMemberVisitor.kt
```diff
@@ -0,0 +1,540 @@
...
+
+    /**
+     * Analyze class by using the provided class reader.
+     *
+     * @param classReader An instance of the class reader to use to access the byte code.
+     * @param context The context in which to analyse the provided class.
+     * @param options Options for how to parse and process the class.
+     */
+    fun analyze(classReader: ClassReader, context: AnalysisContext, options: Int = 0) {
+        analysisContext = context
```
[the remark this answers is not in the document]
tlil: B/c I wanted to be able to reuse the visitor class to traverse a class structure in different context. Especially considering that there can be various derived implementations of it as well, e.g. for remapping, etc. Can revisit that decision at a later stage, but won't make any changes to it as part of this PR.

tlil:
Addressed most comments; will add a few more details around pinned vs whitelisted classes tomorrow, plus update the whitelist in accordance with the latest version of `deterministic-rt.jar`.

djvm/src/main/kotlin/net/corda/djvm/analysis/ClassResolver.kt
```diff
@@ -0,0 +1,112 @@
+package net.corda.djvm.analysis
+
+/**
+ * Functionality for resolving the class name of a sandboxed or sandboxable class.
```
mikehearn: I could use a bit of description here about what "resolved" means exactly.

djvm/src/main/kotlin/net/corda/djvm/analysis/Whitelist.kt
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Enumerate all the entries of the whitelist.
+     */
+    val items: Set<String>
+        get() = textEntries + entries.map { it.pattern }
+
+    companion object {
+        private val everythingRegex = setOf(".*".toRegex())
+
+        // TODO Refine to contain only the bare minimum of classes needed from java/lang
```
mikehearn: Does this TODO need to be done for the security arguments to hold? There's quite a lot in this package!

djvm/src/main/kotlin/net/corda/djvm/analysis/Whitelist.kt
```diff
@@ -0,0 +1,196 @@
...
+                everythingRegex,
+                emptySet()
+        )
+
+        /**
+         * Default whitelist.
+         */
+        val DETERMINISTIC_RUNTIME: Whitelist
+            get() {
+                // TODO This is a snapshot of what's currently in our deterministic rt.jar build
```
mikehearn: Is this done now?

djvm/src/main/kotlin/net/corda/djvm/analysis/Whitelist.kt
```diff
@@ -0,0 +1,196 @@
...
+                // The plan is to strip down this to [Whitelist.MINIMAL] and fully rely on sandbox rule verification and
+                // runtime instrumentation.
+                val jdk = Whitelist.fromResource("jdk8-deterministic.dat.gz")
+                val kotlin = Whitelist.fromResource("kotlin-deterministic.dat.gz")
```
mikehearn: Why is the Kotlin stdlib special here?

djvm/src/main/kotlin/net/corda/djvm/costing/RuntimeCostSummary.kt
```diff
@@ -0,0 +1,57 @@
+package net.corda.djvm.costing
+
+import net.corda.djvm.execution.ExecutionProfile
+
+/**
+ * This class provides a summary of the accumulated costs for the runtime metrics that are being tracked. It also keeps
```
mikehearn: I wonder if we should define a blended metric here, or whether the network parameters should represent cost limits as these individual values?

djvm/src/main/kotlin/net/corda/djvm/execution/ExecutionSummary.kt
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * The summary of the execution of a [SandboxedRunnable] in a sandbox. This class has no representation of the outcome,
+ * and is typically used when there has been a pre-mature exit from the sandbox, for instance, if an exception was
+ * thrown.
+ *
+ * @property costs The costs accumulated when running the sandboxed code.
+ */
+open class ExecutionSummary(
+        val costs: Map<String, Long> = emptyMap()
```
mikehearn: not very type safe - is it done this way to avoid classloader issues or something?

djvm/src/main/kotlin/net/corda/djvm/execution/IsolatedRunnable.kt
```diff
@@ -0,0 +1,108 @@
...
+import net.corda.djvm.rewiring.SandboxClassLoader
+import net.corda.djvm.rewiring.SandboxClassLoadingException
+import net.corda.djvm.utilities.loggerFor
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.atomic.AtomicLong
+
+/**
+ * Container for running a task in an isolated environment.
+ */
+class IsolatedRunnable(
```
mikehearn: It claims to be a Runnable but it's actually not implementing Runnable. Maybe IsolatedTask is better?

djvm/src/main/kotlin/net/corda/djvm/execution/IsolatedRunnable.kt
```diff
@@ -0,0 +1,108 @@
...
+            SandboxRuntimeContext(configuration, context.inputClasses).use {
+                output = try {
+                    action(runnable)
+                } catch (ex: Throwable) {
+                    logger.error("Exception caught in isolated runtime environment", ex)
+                    exception = ex
+                    null
+                }
+                costs = mapOf(
+                        "allocations" to runtimeCosts.allocationCost.value,
```
mikehearn: Maybe these should be defined as constants somewhere.

djvm/src/main/kotlin/net/corda/djvm/execution/IsolatedRunnable.kt
```diff
@@ -0,0 +1,108 @@
...
+        private val configuration: SandboxConfiguration,
+        private val context: AnalysisContext
+) {
+
+    /**
+     * Run an action in an isolated environment.
+     */
+    fun <T> run(action: IsolatedRunnable.() -> T?): Result<T> {
+        val runnable = this
+        val threadName = "Sandbox-$identifier-${uniqueIdentifier.getAndIncrement()}"
```
mikehearn: maybe Sandbox -> DJVM

djvm/src/main/kotlin/net/corda/djvm/execution/IsolatedRunnable.kt
```diff
@@ -0,0 +1,108 @@
...
+                        "allocations" to runtimeCosts.allocationCost.value,
+                        "invocations" to runtimeCosts.invocationCost.value,
+                        "jumps" to runtimeCosts.jumpCost.value,
+                        "throws" to runtimeCosts.throwCost.value
+                )
+            }
+            logger.trace("Exiting isolated runtime environment...")
+            completionLatch.countDown()
+        }.apply {
+            name = threadName
```
mikehearn: There's a (lower case) `thread` helper in kotlin that might avoid the apply here and be useful in general.

djvm/src/main/kotlin/net/corda/djvm/execution/IsolatedRunnable.kt
```diff
@@ -0,0 +1,108 @@
...
+        val messages = exception.let {
+            when (it) {
+                is SandboxClassLoadingException -> it.messages
+                is SandboxException -> {
+                    when (it.exception) {
+                        is SandboxClassLoadingException -> it.exception.messages
+                        else -> null
+                    }
+                }
+                is StackOverflowError -> throw StackOverflowError("Stack overflow")
```
mikehearn: why not rethrow the original exceptions?

djvm/src/main/kotlin/net/corda/djvm/execution/SandboxExecutor.kt
```diff
@@ -0,0 +1,191 @@
...
+            runnableClass: ClassSource,
+            input: TInput
+    ): ExecutionSummaryWithResult<TOutput?> {
+        // 1. We first do a breath first traversal of the class hierarchy, starting from the requested class.
+        //    The branching is defined by class references from referencesFromLocation.
+        // 2. For each class we run validation against defined rules.
+        // 3. Since this is hitting the class loader, we are also remapping and rewriting the classes using the provided
+        //    emitters and definition providers.
+        // 4. While traversing and validating, we build up another queue of references inside the reference validator.
+        // 5. We drain this queue by validating references, this time including member references.
```
mikehearn: A bit of discussion here about what "validating references" means would be good.

djvm/src/main/kotlin/net/corda/djvm/execution/SandboxExecutor.kt
```diff
@@ -0,0 +1,191 @@
...
+        // 6. For execution, we then load the top-level class, implementing the SandboxedRunnable interface, again and
+        //    and consequently hit the cache. Once loaded, we can execute the code on the spawned thread, i.e., in an
+        //    isolated environment.
+        logger.trace("Executing {} with input {}...", runnableClass, input)
```
mikehearn: A bit of discussion about possible parallelism might be good too.

djvm/src/main/kotlin/net/corda/djvm/execution/SandboxedRunnable.kt
```diff
@@ -0,0 +1,19 @@
+package net.corda.djvm.execution
+
+/**
+ * Functionality runnable by a sandbox executor.
+ */
+interface SandboxedRunnable<in TInput, out TOutput> : DiscoverableRunnable {
```
mikehearn: Isn't this a java 8 `Function`, basically?

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysInheritFromSandboxedObject.kt
```diff
@@ -0,0 +1,63 @@
...
+            new(SANDBOX_OBJECT_NAME, instruction.operation)
+            preventDefault()
+        }
+        if (instruction is MemberAccessInstruction &&
+                instruction.operation == Opcodes.INVOKESPECIAL &&
+                instruction.owner == OBJECT_NAME &&
+                instruction.memberName == CONSTRUCTOR_NAME &&
+                context.clazz.name != SANDBOX_OBJECT_NAME) {
+            // Rewrite object initialisation call so that the sandboxed constructor gets used instead.
+            loadConstant(0xfed_c0de)
```
mikehearn: A note about what this magic number is would be useful here, although I suspect the hashCode RNG should be something we can set from outside the sandbox so it's initialised from the tx bytes hash and iterates appropriately.

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysUseExactMath.kt
```diff
@@ -0,0 +1,36 @@
+package net.corda.djvm.rules.implementation
+
+import net.corda.djvm.code.Emitter
+import net.corda.djvm.code.EmitterContext
+import net.corda.djvm.code.Instruction
+import org.objectweb.asm.Opcodes.*
+
+/**
+ * Use exact integer and long arithmetic where possible.
```
mikehearn: Note here that this rule is not required for determinism, but we believe it helps make code more robust, so we enable it.

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysUseExactMath.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+@Suppress("unused")
+class AlwaysUseExactMath : Emitter {
```
mikehearn: Why does it appear unused?

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysUseStrictFloatingPointArithmetic.kt
```diff
@@ -0,0 +1,31 @@
...
+import net.corda.djvm.code.MemberDefinitionProvider
+import net.corda.djvm.references.EntityWithAccessFlag
+import net.corda.djvm.references.Member
+import net.corda.djvm.rules.MemberRule
+import net.corda.djvm.validation.RuleContext
+import org.objectweb.asm.Opcodes.ACC_STRICT
+import java.lang.reflect.Modifier
+
+/**
+ * Definition provider that ensures that all methods use strict floating-point arithmetic in the sandbox.
```
mikehearn: Can you add a note here that future JVM releases may make this pass obsolete:

https://bugs.openjdk.java.net/browse/JDK-8175916

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/DisallowFinalizerMethods.kt
```diff
@@ -0,0 +1,17 @@
...
+import net.corda.djvm.validation.RuleContext
+
+/**
+ * Rule that checks for invalid use of finalizers.
+ */
+@Suppress("unused")
+class DisallowFinalizerMethods : MemberRule() {
+
+    override fun validate(context: RuleContext, member: Member) = context.validate {
+        fail("Disallowed finalizer method") given ("${member.memberName}:${member.signature}" == "finalize:()V")
```
mikehearn: Maybe add a // TODO here to just automatically erase them. The JVM is allowed to never call a finalizer so this might increase our compatibility a bit.

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/DisallowReflection.kt
```diff
@@ -0,0 +1,30 @@
...
+import net.corda.djvm.validation.RuleContext
+
+/**
+ * Rule that checks for illegal references to reflection APIs.
+ */
+@Suppress("unused")
+class DisallowReflection : InstructionRule() {
+
+    override fun validate(context: RuleContext, instruction: Instruction) = context.validate {
+        // TODO Enable controlled use of reflection APIs
```
mikehearn: Hmm, there are more reflection APIs than just these e.g. in `sun.misc.Unsafe`. 

We should disallow access to any of the packages that are forbidden by the default Java sandbox security policy. See the `package.access` key in jre/lib/security/java.security

djvm/src/main/kotlin/net/corda/djvm/utilities/Processor.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Utility for processing a set of entries in a list matching a particular type.
+ */
+object Processor {
+
+    /**
+     * Process entries of type [T] in the provided list, using a guard around the processing of each item, catching
+     * any [Message] that might get raised.
+     */
+    inline fun <reified T> processEntriesOfType(
```
mikehearn: This looks more like an extension method on `List<*>` - but I wonder if we should refactor out a generic Corda utilities library for both djvm and node/core to share.

djvm/src/main/kotlin/sandbox/java/lang/Object.kt
```diff
@@ -0,0 +1,19 @@
+package sandbox.java.lang
+
+/**
+ * Sandboxed implementation of `java/lang/Object`.
+ */
+@Suppress("EqualsOrHashCode")
+open class Object(private val hashValue: Int) {
```
mikehearn: So, this works, but will cause crap performance in hash maps and other such things.

What we should probably do is have a per-context RNG seeded from the tx merkle root, that is iterated each time it's accessed or something. Perhaps it doesn't matter and we can just have a todo here.

hashCode here should probably chain to System.identityHashCode with a comment that in the DJ subset, System.identityHashCode is deterministic?

djvm/src/main/kotlin/sandbox/net/corda/sandbox/costing/RuntimeCostAccounter.kt
```diff
@@ -0,0 +1,105 @@
...
+            context.runtimeCosts.jumpCost.increment()
+
+    /**
+     * Record a memory allocation operation.
+     *
+     * @param typeName The class name of the object being instantiated.
+     */
+    @JvmStatic
+    fun recordAllocation(typeName: String) {
+        // TODO Derive better size estimates for complex types.
```
mikehearn: This might be useful: https://github.com/DimitrisAndreou/memory-measurer

djvm/src/main/kotlin/sandbox/net/corda/sandbox/costing/RuntimeCostAccounter.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Record an array allocation operation.
+     *
+     * @param length The number of elements in the array.
+     * @param typeName The class name of the array element type.
+     */
+    @JvmStatic
+    fun recordArrayAllocation(length: Int, typeName: String) {
+        val size = allocationCosts.getOrDefault(typeName, 16)
```
mikehearn: Need to check here that length is >= 0 otherwise I can write bytecode that arbitrarily reduces my own allocation cost.

djvm/src/main/kotlin/sandbox/net/corda/sandbox/costing/RuntimeCostAccounter.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Record an array allocation operation.
+     *
+     * @param length The number of elements in the array.
+     * @param typeSize The size of the array element type.
+     */
+    @JvmStatic
+    fun recordArrayAllocation(length: Int, typeSize: Int) {
+        context.runtimeCosts.allocationCost.increment(length * typeSize)
```
mikehearn: ditto

mikehearn (changes requested):
Incredibly clean and professionally done code, congratulations on building something quite special here.

djvm/src/main/kotlin/sandbox/net/corda/sandbox/costing/RuntimeCostAccounter.kt
tlil: Yep, good point. Also changed the accounter to use `Math.addExact()` to catch potential overflows as well.

djvm/src/main/kotlin/sandbox/java/lang/Object.kt
tlil: Yes, fully aware 😄 What you mention has been the plan all along. This work has just been the enabler for it. The value gets passed in through the emitter at the moment:

https://github.com/corda/corda/blob/d0a743f8c778749bedc7bf614786743263c15a98/djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysInheritFromSandboxedObject.kt#L38-L41

So `hashValue` could either be derived from context and injected per access on emittance (which potentially could make caching less efficient), or this class can be modified to pull in the context-based seed in a different manner.

Will clarify by either adding a TODO or provide a reference implementation of how it can be done when I push my changes.

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        // Transform the class definition and byte code in accordance with provided rules.
+        val byteCode = rewriter.rewrite(reader, context)
+
+        // Try to define the transformed class.
+        val clazz = try {
+            defineClass(resolvedName, byteCode.bytes, 0, byteCode.bytes.size)
+        } catch (exception: SecurityException) {
+            supportingClassLoader.loadClass(name)
```
exFalso: In what circumstances is this `catch` hit?

djvm/cli/src/main/kotlin/net/corda/djvm/tools/cli/TreeCommand.kt
```diff
@@ -0,0 +1,34 @@
+package net.corda.djvm.tools.cli
+
+import net.corda.djvm.tools.Utilities.workingDirectory
+import picocli.CommandLine.Command
+import java.nio.file.Files
+
+@Command(
+        name = "tree",
+        description = ["Show the hierarchy of the classes that has been created with the 'new' command."]
```
exFalso: typo has -> have

djvm/src/main/kotlin/net/corda/djvm/validation/RuleContext.kt
```diff
@@ -0,0 +1,70 @@
...
+        get() = analysisContext.location
+
+    /**
+     * The configured whitelist.
+     */
+    val whitelist: Whitelist
+        get() = analysisContext.configuration.whitelist
+
+    /**
+     * Classes that has been explicitly defined in the sandbox namespace.
```
exFalso: typo has -> have

exFalso (commented):
Small things

djvm/src/main/kotlin/net/corda/djvm/execution/ExecutionSummary.kt
tlil: I've made this strongly typed

djvm/src/main/kotlin/net/corda/djvm/execution/IsolatedRunnable.kt
tlil: Made the costs strongly typed, which removes the need for these string constants in the first place.

djvm/src/main/kotlin/net/corda/djvm/execution/IsolatedRunnable.kt
tlil: Will fix

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysUseExactMath.kt
tlil: It's just because the rules are implicitly registered at the moment. I will remove the annotation.

djvm/src/main/kotlin/sandbox/net/corda/sandbox/costing/RuntimeCostAccounter.kt
tlil: Yes, I had a look at that a while ago. Decided against using it for now since:

1) `MemoryMeasurer` requires a Java agent, and
2) `ObjectGraphMeasurer` just gives you a footprint object with number of objects, primitives and references.

To avoid inflating the scope of this initial PR too much ;), I'll leave this as future work. Will add the link as a comment to `RuntimeCostAccounter.kt` though.

djvm/src/main/kotlin/net/corda/djvm/execution/SandboxedRunnable.kt
tlil: Basically, yeah. It just adds the `DiscoverableRunnable` and explicitly allows for nullable return values from `run()` (in Kotlin land).

djvm/src/main/kotlin/net/corda/djvm/utilities/Processor.kt
tlil: Yes, I was thinking that could be useful as well. Let's do that as a follow-up to this PR though.

djvm/src/main/kotlin/net/corda/djvm/costing/RuntimeCostSummary.kt
tlil: @exFalso, thoughts? Pinned limits provided through the network parameters? I think we can do that as a follow-up to this PR regardless, though :)

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysInheritFromSandboxedObject.kt
tlil: Yes, this has been intended as a placeholder and example of how it can be passed in. I will add a comment describing that this is a magic number and that an RNG should be passed from the outside. Will clarify a bit and push an update.

djvm/src/main/kotlin/sandbox/java/lang/Object.kt
tlil: As for chaining to `System.identityHashCode()` - that's somewhat tricky unfortunately as `System` resides in `java.lang` and defining classes in that package is illegal. What did you have in mind here @mikehearn ? If I could re-define `System` through the classloader, that would make sense, but I can't.

tlil: Okay, so will:
1) Implement a `sandbox.java.lang.System`
2) Replace references in loaded classes accordingly
3) Change `sandbox.java.lang.Object.hashCode()` to call `sandbox.java.lang.System.identityHashCode(this)`

tlil:
@mikehearn @exFalso - back to you

djvm/src/main/kotlin/sandbox/net/corda/sandbox/costing/RuntimeCostAccounter.kt
mikehearn: I guess it's impossible to have a type size of zero so we should use > rather than >= here

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysReferencePinnedClasses.kt
```diff
@@ -0,0 +1,3 @@
+package net.corda.djvm.rules.implementation
+
+// TODO Use sandbox.* versions of registered pinned classes
```
mikehearn: I'm not sure I understand this comment or what impact it has on security.

djvm/src/main/kotlin/sandbox/java/lang/System.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    @Throws(java.io.IOException::class)
+    fun inheritedChannel(): java.nio.channels.Channel? {
+        throw IOException()
+    }
+
+    fun setSecurityManager(manager: java.lang.SecurityManager) {}
+
+    fun getSecurityManager(): java.lang.SecurityManager? {
+        throw NotImplementedError()
```
mikehearn: This should just return NULL to indicate "no security manager in use". The standard pattern for Java sandbox permission checks is to call getSecurityManager and then if it's not null, use it.

djvm/src/main/kotlin/sandbox/java/lang/System.kt
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    fun gc() {}
+
+    fun runFinalization() {}
+
+    fun runFinalizersOnExit(flag: Boolean) {}
+
+    fun load(path: String) {
+        throw NotImplementedError("Eh eh, no can do!")
```
mikehearn: Maybe these can be empty too, so libraries that try to load native code but in which we never hit the codepath that needs them still work.

mikehearn (changes requested):

mikehearn:
Looks super good now, nearly there.

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysReferencePinnedClasses.kt
tlil: This file was removed in a previous commit. And the type size comment above has already been fixed.

tlil: Ah, GitHub fooled me. I see what you meant about type size. Fixed

mikehearn (approved):

mikehearn:
giphy

PR #3394, opened 2018-06-18
[CORDA-1597] Make the NodeArgsParser fail on unrecognized (non-option…

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1597

### DIFF
--- a/node/src/test/kotlin/net/corda/node/NodeArgsParserTest.kt
+++ b/node/src/test/kotlin/net/corda/node/NodeArgsParserTest.kt
@@ -178,4 +178,18 @@ class NodeArgsParserTest {
             assertThat(cmdLineOptions.unknownConfigKeysPolicy).isEqualTo(onUnknownConfigKeyPolicy)
         }
     }
+
+    @Test
+    fun `invalid argument`() {
+        assertThatExceptionOfType(IllegalArgumentException::class.java).isThrownBy {
+            parser.parse("foo")
+        }.withMessageContaining("Unrecognized argument(s): foo")
+    }
+
+    @Test
+    fun `invalid arguments`() {
+        assertThatExceptionOfType(IllegalArgumentException::class.java).isThrownBy {
+            parser.parse("foo", "bar")
+        }.withMessageContaining("Unrecognized argument(s): foo, bar")
+    }
 }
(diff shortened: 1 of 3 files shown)


node/src/test/kotlin/net/corda/node/NodeArgsParserTest.kt
```diff
@@ -178,4 +178,18 @@ class NodeArgsParserTest {
             assertThat(cmdLineOptions.unknownConfigKeysPolicy).isEqualTo(onUnknownConfigKeyPolicy)
         }
     }
+
+    @Test
+    fun `invalid argument`() {
+        assertThatExceptionOfType(IllegalArgumentException::class.java).isThrownBy {
```
anthonykeenan: Should it throw an exception? Or should it exit with a message?

For example if you run `java -jar corda.jar --unknown-option` you get `unknown-option is not a recognised option`, followed by the help output - surely that would be a much more pleasant user experience?`

florian-f: The `IllegalArgumentException` is thrown by `require`, which I chose to make it consistent with other cases. 
We could throw an `OptionException` instead, this would be caught `AbstractArgsParser.parseOrExit` and result in the behaviour you described. I agree that the user experience would be better.

florian-f: OK so we can't throw an `OptionException` there because the constructor is protected. 
Also, we might want to have the same behaviour for all the other cases that we check with `require` in the NodeArgsParser. My suggestion is to catch `IllegalArgumentException`s as well in `parseOrExit` and treat them the same way we treat `OptionException`s.

anthonykeenan: Ok, so it's better, for 
`java -jar corda.jar foo bar`
It returns `Unrecognized argument(s): foo, bar` and the help text underneath.
If I run
`java -jar corda.jar -foo -bar` however I just get `f is not a recognized option` and then the help text. It's not the end of the world but is that easily fixable?

florian-f: What happens in the case of `-foo` is that jopts interprets it as a cluster of short options, so it treats it as `-f -o -o`. I'm not sure we can change this behaviour without modifying jopts itself, and without breaking short option clustering. 

Interestingly, you get different behaviour with `-config-file` (single hyphen), because the `handleShortOptionToken` method in OptionParser only executes the cluster-logic if it doesn't recognize the option (long _or_ short).

anthonykeenan: No that's fine then, the error message is a lot better than it was and it's showing the help rather than doing strange things so I'm happy with that.

gendal: thanks for fighting the good fight, @anthonykeenan! throwing exceptions in the faces of our beloved users would be a terrible thing to do!

PR #3399, opened 2018-06-19
CORDA-1645: Checkpoint before calling an idempotent sub-flow.

adagys:
When an idempotent sub-flow causes a flow restart, the flow will be
replayed from the last checkpoint before the sub-flow invocation.
However, the logic between the last checkpoint and the sub-flow invocation
may contain side-effects which shouldn't be replayed. Thus we need to persist
a checkpoint just before an idempotent sub-flow is invoked.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/FlowIORequest.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/FlowIORequest.kt
@@ -90,5 +90,6 @@ sealed class FlowIORequest<out R : Any> {
      * Indicates that no actual IO request occurred, and the flow should be resumed immediately.
      * This is used for performing explicit checkpointing anywhere in a flow.
      */
+    // TODO: consider using an empty FlowAsyncOperation instead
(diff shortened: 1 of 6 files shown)


adagys:
Yeah that would make sense. Although in general we'd want to skip the checkpoint if an async operation completes immediately – I have a task for that noted somewhere. So maybe it's better to have a separate IO req type for this.

exFalso:
We can't check in advance whether an async operation completes immediately though. I think it's better if we always checkpoint and be explicit whether we want to skip in the skipCheckpoint flag thing. Perhaps we could also indicate this in the operation class somehow?

exFalso:
Another thing is that this does only one transition and executeasync does 2. Not sure whether it matters though

adagys:
You're right, that won't work with the way it's implemented right now. In any case, given the time constraints I think let's leave this discussion until after GA. Having an explicit IO req type makes it easier to reason about, I'd say, and we can always refactor it later.

PR #3400, opened 2018-06-19
CORDA-1603: Class carpenter able to synthesis Cash.State objects

shamsasari:
This required a minor tweak to allow classes that implement generic interfaces and also to remove the check on non-AMQP fields in interfaces.


shamsasari:
Added a "lenient" mode to the carpenter.

shamsasari:
`fromContext` method removed.

fenryka:
LGTM now :)

PR #3403, opened 2018-06-19
[CORDA-1638]: Node crashes in --initial-registration polling mode if doorman returns a transient HTTP error (fix).

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1638

### DIFF
--- a/node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
@@ -16,12 +16,15 @@ import net.corda.nodeapi.internal.crypto.X509Utilities.CORDA_ROOT_CA
 import org.bouncycastle.asn1.x500.X500Name
 import org.bouncycastle.openssl.jcajce.JcaPEMWriter
 import org.bouncycastle.util.io.pem.PemObject
+import java.io.IOException
 import java.io.StringWriter
 import java.nio.file.Path
 import java.security.KeyPair
 import java.security.KeyStore
 import java.security.PublicKey
 import java.security.cert.X509Certificate
+import java.time.Duration
+import javax.naming.ServiceUnavailableException
 
 /**
  * Helper for managing the node registration process, which checks for any existing certificates and requests them if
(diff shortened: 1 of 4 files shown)


node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -160,12 +164,22 @@ open class NetworkRegistrationHelper(private val config: SSLConfiguration,
...
-            if (certificates != null) {
-                return certificates
+            try {
+                val (pollInterval, certificates) = certService.retrieveCertificates(requestId)
+                if (certificates != null) {
+                    return certificates
+                }
+                Thread.sleep(pollInterval.toMillis())
+            } catch (e: ServiceUnavailableException) {
+                idlePeriodDuration = computeNextIdleDoormanConnectionPollInterval.invoke(idlePeriodDuration)
```
patrickkuo: i think `computeNextIdleDoormanConnectionPollInterval` is too long ... maybe just `nextIdleDuration`?  Also can remove invoke and use () directly

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -254,4 +271,21 @@ class NodeRegistrationHelper(private val config: NodeConfiguration, certService:
...
+
+        init {
+            require(times > 0)
+        }
+
+        private var counter = times
+
+        override fun invoke(@Suppress("UNUSED_PARAMETER") previousPeriod: Duration?): Duration? {
+            synchronized(this) {
+                val nextPeriod = if (counter > 0) period else null
```
patrickkuo: can simplify to ` return if (counter-- > 0) period else null` ?

sollecitom:
@patrickkuo , fixed. I'll wait for the tests to finish before pushing.

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
sollecitom: Not sure it's "simpler", but changed anyway.

PR #3404, opened 2018-06-19
CORDA-1645: Fix compilation issue

adagys:


adagys:
Wonder if we can bypass the integration tests at least

PR #3407, opened 2018-06-20
 Create Artemis p2p.inbound addresses before starting the broker

thschroeter:


thschroeter:
I'm adding `queueConfigurations` to the `SecureArtemisConfiguration` as well.

node/src/main/kotlin/net/corda/node/services/messaging/ArtemisMessagingServer.kt
```diff
@@ -49,7 +54,8 @@ import javax.security.auth.login.AppConfigurationEntry.LoginModuleControlFlag.RE
 @ThreadSafe
 class ArtemisMessagingServer(private val config: NodeConfiguration,
                              private val messagingServerAddress: NetworkHostAndPort,
-                             private val maxMessageSize: Int) : ArtemisBroker, SingletonSerializeAsToken() {
+                             private val maxMessageSize: Int,
+                             private val identities: List<PublicKey>) : ArtemisBroker, SingletonSerializeAsToken() {
```
bpaunescu: Can we add another constructor without identities that uses this one with the emptylList() ?

thschroeter: What about adding a default? `private val identities: List<PublicKey> = emptyList()`?

bpaunescu:
Looks ok to me, but think it would be good to have @exFalso or @rick-r3 review on this.

rick-r3:
Looks good.

PR #3408, opened 2018-06-20
CORDA-1610: Retain progress tracker during flow retry.

adagys:
Make sure the same progress tracker object is re-used in the restarted flow so subscribers can keep receiving progress updates.


adagys:
Yep.. not pretty, but I don't see a better way of doing it without breaking the API :/

PR #3409, opened 2018-06-20
ENT-1741 Build standalone shell JAR distribution

szymonsztuka:
`tools:shell` module is split into two in order to publish the library and the tool separately:
1) common shell code remains in `tools:shell` and it's used by the `node` module for the embedded shell
2) Standalone Shell tool moved to `tools:shell-cli` and it depends on `shell`, the tool was not published so far

Artefacts being produced: 
1) The module `tools:shell` as it was before ArtifactId `corda-shell`
2) New `tools:shell-cli` is published as ArtifactId `corda-tools-shell-cli`


sollecitom:
Is this for GA @szymonsztuka ?

szymonsztuka:
@sollecitom yes, it can be merged after the diff

PR #3412, opened 2018-06-21
add simple network verification cordapp

roastario:


tlil:
Nice - perhaps it would be useful to extend so that the "test notary" flow could take a notary name as input (for networks with multiple notaries)?

Perhaps also having a simple flow for testing P2P between two participants only as well?

roastario:
The TestCommsFlow does  check p2p comms. It does a Send/Receive of a simple string.

tlil:
Yes, I know. I meant having a way to switch between doing **test-all** vs **test-particular-node** as it were.

samples/network-verifier/build.gradle
```diff
@@ -0,0 +1,49 @@
+apply plugin: 'kotlin'
+apply plugin: 'java'
+apply plugin: 'net.corda.plugins.cordapp'
+apply plugin: 'net.corda.plugins.cordformation'
+
+dependencies {
+    cordaCompile project(":core")
+    cordaCompile project(":node-api")
```
mnesbit: I thought we had our own cordapp configuration space?? That meant we didn't package up with the framework??
So I think this is wrong for the non-rpc bit and thus you will need two sub projects.

samples/network-verifier/src/main/kotlin/net/corda/verification/TestCommsFlow.kt
```diff
@@ -0,0 +1,81 @@
...
+                    val initiateFlow = initiateFlow(it)
+                    initiateFlow.receive<String>().unwrap { it }
+                }.toList().also {
+                    progressTracker.currentStep = RECIEVED_ALL
+                }
+        val tx = TransactionBuilder(notary = serviceHub.networkMapCache.notaryIdentities.first())
+        tx.addOutputState(CommsTestState(responses, serviceHub.myInfo.legalIdentities.first()), CommsTestContract::class.qualifiedName!!)
+        tx.addCommand(CommsTestCommand, serviceHub.myInfo.legalIdentities.first().owningKey)
+        val signedTx = serviceHub.signInitialTransaction(tx)
+        subFlow(FinalityFlow(signedTx))
```
ivanr3: What is the point of doing this? This tx wouldn't go to notary as this is an issuance neither it would go to other parties as we are the only signer.

samples/network-verifier/src/main/kotlin/net/corda/verification/TestCommsFlow.kt
```diff
@@ -0,0 +1,81 @@
...
+        val responses = serviceHub.networkMapCache.allNodes.map {
+            it.legalIdentities.first()
+        }.filterNot {
+            it in serviceHub.myInfo.legalIdentities
+        }.filterNot {
+            it in serviceHub.networkMapCache.notaryIdentities
+        }.filter(::matchesX500)
+                .map {
+                    val initiateFlow = initiateFlow(it)
+                    initiateFlow.receive<String>().unwrap { it }
```
ivanr3: Its a shame that we can't specify  timeout for the .send() . receive() calls. Should we expose some information about which nodes have replied? For example we can have a singleton service with Set<Party> repliedParties and then expose it to the web api? So we have visibility over what are the problematic nodes?

samples/network-verifier/src/main/kotlin/net/corda/verification/TestCommsFlow.kt
```diff
@@ -0,0 +1,81 @@
...
+    @Suspendable
+    override fun call(): List<String> {
+        progressTracker.currentStep = SENDING
+        val responses = serviceHub.networkMapCache.allNodes.map {
+            it.legalIdentities.first()
+        }.filterNot {
+            it in serviceHub.myInfo.legalIdentities
+        }.filterNot {
+            it in serviceHub.networkMapCache.notaryIdentities
+        }.filter(::matchesX500)
```
ivanr3: This would complete successfully if I test against a non-existing party. Would make sense to throw an exception if the party doesn't exist at all.

samples/network-verifier/src/main/kotlin/net/corda/verification/TestNotaryFlow.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    object ISSUING : ProgressTracker.Step("ISSUING")
+    object ISSUED : ProgressTracker.Step("ISSUED")
+    object DESTROYING : ProgressTracker.Step("DESTROYING")
+    object FINALIZED : ProgressTracker.Step("FINALIZED")
+
+    override val progressTracker: ProgressTracker = ProgressTracker(ISSUING, ISSUED, DESTROYING, FINALIZED)
+
+
+    override fun call(): String {
```
ivanr3: Missing @suspendable

PR #3414, opened 2018-06-21
Setting up a bot Testing Environment

DanielCordell:
I'm testing the Bot Pokey, which will auto-merge all green pull requests.

These branches will be deleted after the test is over.

@GavT will accept this, and then we will test if this pull request gets merged by the bot.


GavT:
Approved

PR #3418, opened 2018-06-21
ENT-1463, ENT-1903: Document new minimum JDK version.

chrisr3:


gendal:
Just to double-check... this isn't a GA change, right?  Either way, do we need pre-emptively to update the upgrade/release notes in master so that when we ship a version with this dependency change we remember to clearly communicate it to people?

chrisr3:
This change aligns the documentation in `master` with its current JDK requirement (c.f. root `build.gradle`), AFAIK, GA doesn't include my new Gradle plugin and so doesn't need to update its minimum JDK.

PR #3422, opened 2018-06-22
CORDA-1661 Reverting DEV certificates

mkit:
This PR reverts the DEV certificates to the state as they were at Corda OS 3.0 release


patrickkuo:
just to confirm, it's the key get reverted right? the certificate is newly created and will have the new attributes.

mkit:
Yes. Certificates are regenerated. You can see it by looking into the intermediate certificate and the extensions there. There is a new extension for the CRL distribution point.

patrickkuo:
both root and intermediate key are the same as the ones in v3.0.

The subject name of the intermediate cert has changes to Doorman CA, can you confirm this won't cause any problem when validating cert chain

it will be good to have a small unit test to confirm this will pass the check

mnesbit:
I think it is safer if the path for the test certificates is regression-test rather than old-certificates, otherwise people might use them.

mkit:
@mnesbit Renamed.

PR #30, opened 2016-12-05
Added a note in the getting setup doc about JavaFX and non-Oracle JVMs

shamsasari:

PR #27, opened 2016-12-05
Bank of Corda demo - Issuer of Cash

josecoll:
Resolve BankOfCorda through NMS in protocol

Fixes following Integration testing.

Register custom RPC Kryo classes.

Protocol -> Flow renaming

Bank of Corda demo - Issuer of Cash

Resolve BankOfCorda through NMS in protocol

Fixes following Integration testing.

Protocol -> Flow renaming

Addressed all comments in PR review.

Removed bank lines.

Updated minor inconsistency in README.md

All protocol references changed to flow.

changed protocol -> flow in TODO comment.

changed startProtocolPermission -> startFlowPermission in README.md

Added transaction id to IssuerFlow Success response.
Removed explicit call to record Cash Move transaction (as already recorded in subflow)

Removed quasar dependency.

Addressed comment in PR.

Updated to use CompositeKey.

Added arguments to pass in Currency and Amount.

Updated run configurations to pass in Currency and Amount values

Added additional parameter to IssuerFlow request: issueToPartyReference

Added Vault updates verification in RPC Integration test.

Fixed RPC Integration test (Vault assertions)

Issue #18, opened 2016-12-02
node:compileKotlin JDBCHashMap.kt compilation error


y12studio:
I'm getting a compilation error.

```
$ git clone https://github.com/corda/corda && cd corda
$ ./gradlew build --info

...
Executing task ':node:compileKotlin' (up-to-date check took 0.027 secs) due to:
  No history is available.
All input files are considered out-of-date for incremental task ':node:compileKotlin'.
e: /home/lin/git/corda/node/src/main/kotlin/net/corda/node/utilities/JDBCHashMap.kt: (208, 27): Inherited platform declarations clash: The following declarations have the same JVM signature (remove(Ljava/lang/Object;Ljava/lang/Object;)Z):
    fun remove(key: Any, value: Any): Boolean
    fun remove(key: Int, value: MutableList<AbstractJDBCHashMap.NotReallyMutableEntry<K, V>>): Boolean
e: /home/lin/git/corda/node/src/main/kotlin/net/corda/node/utilities/JDBCHashMap.kt: (208, 27): Inherited platform declarations clash: The following declarations have the same JVM signature (getOrDefault(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;):
    fun getOrDefault(key: Any!, defaultValue: MutableList<AbstractJDBCHashMap.NotReallyMutableEntry<K, V>>!): MutableList<AbstractJDBCHashMap.NotReallyMutableEntry<K, V>>!
    fun getOrDefault(key: Int, defaultValue: MutableList<AbstractJDBCHashMap.NotReallyMutableEntry<K, V>>): MutableList<AbstractJDBCHashMap.NotReallyMutableEntry<K, V>>
[KOTLIN] deleting /home/lin/git/corda/node/build/kotlin-classes/main on error
:node:compileKotlin FAILED
:node:compileKotlin (Thread[Daemon worker Thread 4,5,main]) completed. Took 8.533 secs.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':node:compileKotlin'.
> Compilation error. See log for more details

* Try:
Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output.

BUILD FAILED

Total time: 9.99 secs
Stopped 0 compiler daemon(s).
[...]

PR #15, opened 2016-12-01
Fix double spending of inputs issue on ledger level in test dsl.

kasiastreich:
Simple fix with tracking of used inputs on a ledger level.

PR #3427, opened 2018-06-23
CORDA-1662 - Corda Serialization Evolution breaksdown with Java classes

fenryka:
Nullability logic was relying on annotations that Kotlin applies by
default but is left to the developer in Javaland. Change this around
so it works for both.

In Kotlin, the property must be nullable, in Java, it can't be a
primitive.


mikehearn:
Is this logic really right? Why does it need to be Kotlin specific? Isn't the rule:

* If there are no annotations, it's nullable.
* If there's a NonNull annotation, it's not nullable.
* If there's a package level ParametersAreNonnullByDefault annotation, it's non-nullable unless annotated as Nullable

AFAIK the Kotlin compiler does annotate everything with nullability annotations, so we should just be able to use them.

fenryka:
I can re-write the reflection in a more Java centric way.... which probably isn't a bad thing give nwhose going to be using it most

fenryka:
I can't think of a nicer way of writing this than exists...

Kotlin doesn't apply specific nullability annotations to types, it embeds that info in the kotlin meta data annotation so you'd need to do a specific heck for Kotlin anyway... and you'd also need to check for primitives and java

Either a class is a Java one and it's nullable and may or may not have an annotation applied, but even if it's annotated as nullable it can be set as null and it may be a primitive type. Note, the logic here isn't that someone could feasibly add a nullable primitive, but that in a generic way the serializer can try to deserialize an older instance of the state with that new value missing. If it's nullable then we can continue as ever, if it's not or it's a primitive we have to error and be able to decide that in a generic way.

Then, on the flip side, if it is a Kotlin object we can rely on the metadata helper returning what we want

I can re-write this a number of ways but I don't see it getting any cleaner

mikehearn:
I'm still confused, sorry.

The rule "if there's no annotation it's nullable" seems to cover the Java case.

mikehearn:
Also Kotlin does use annotations. Look at the "Show Kotlin Bytecode" window. E.g. here's a c'tor that has a nullable parameter:

```
  public <init>(Lnet/corda/core/schemas/PersistentStateRef;)V
    @Lorg/jetbrains/annotations/Nullable;() // invisible, parameter 0
```

and one that has a non-nullable parameter:

```
  public <init>(Lnet/corda/core/contracts/StateRef;)V
    @Lorg/jetbrains/annotations/NotNull;() // invisible, parameter 0
```

fenryka:
@mikehearn   Then reflection is working oddly as somethign like

`constructor.javaConstructor?.parameters?.forEach {
                println ("  * ${it.name} - ${it.type}")
                it.annotations?.forEach {
                    println ("    * $it")
                }
            }`

When reflecting on a Java class will print out any nullable annotations applied: such as

`static class N1 {
        private String word;
        private @Nullable Integer wibble;

        public N1(String word, @Nullable Integer wibble) {
            this.word = word;
            this.wibble = wibble;
        }
}`

Where as it won't for a Kotlin class

fenryka:
@mikehearn the Kotlin annotations aren't retained at runtime

shamsasari:
```

**Comment by shamsasari** [comment]
```
data class TestData(val a: String?, val b: String)

    val ctor = TestData::class.java.constructors[0]
    ctor.parameters.forEach {
        println("${it.name} ${it.declaredAnnotations.joinToString()}")
    }
}
```
prints just
```
a 
b 
```

Examining with javap we get
```
  public net.corda.serialization.internal.amqp.TestData(java.lang.String, java.lang.String);
    descriptor: (Ljava/lang/String;Ljava/lang/String;)V
    flags: ACC_PUBLIC
    Code:
      stack=2, locals=3, args_size=3
         0: aload_2
         1: ldc           #21                 // String b
         3: invokestatic  #27                 // Method kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
         6: aload_0
         7: invokespecial #30                 // Method java/lang/Object."<init>":()V
        10: aload_0
        11: aload_1
        12: putfield      #11                 // Field a:Ljava/lang/String;
        15: aload_0
        16: aload_2
        17: putfield      #18                 // Field b:Ljava/lang/String;
        20: return
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      21     0  this   Lnet/corda/serialization/internal/amqp/TestData;
            0      21     1     a   Ljava/lang/String;
            0      21     2     b   Ljava/lang/String;
      LineNumberTable:
        line 3: 6
    MethodParameters:
      Name                           Flags
      a
      b
    RuntimeInvisibleParameterAnnotations:
      0:
        0: #7()
      1:
        0: #15()
```
```
   #7 = Utf8               Lorg/jetbrains/annotations/Nullable;
  #15 = Utf8               Lorg/jetbrains/annotations/NotNull;
```

So yes, the kotlin compiler is not paying nice with annotation visibility.

File: serialization/src/main/kotlin/net/corda/serialization/internal/amqp/EvolutionSerializer.kt [hunk 1]
```diff
@@ -117,12 +118,25 @@ abstract class EvolutionSerializer(
...
+            // case it can't be referenced
+            val isKotlin = (new.type.javaClass.declaredAnnotations.any {
+                        it.annotationClass.qualifiedName == "kotlin.Metadata"
+            })
+
             constructor.parameters.withIndex().forEach {
                 readersAsSerialized[it.value.name!!]?.apply {
                     this.resultsIndex = it.index
-                } ?: if (!it.value.type.isMarkedNullable) {
+                } ?: // If there is no value in the byte stream to map to the parameter of the constructor
```
shamsasari: This is difficult to make sense of. Can you convert the elvis to a normal if-statement

fenryka:
@shamsasari ... good stuff... the alternative is what I did and read the Kotlin annotation code, you can see they've set it to CLASS and not RUNTIME

Which is vexing

fenryka:
```Kotlin
@Documented
@Retention(RetentionPolicy.CLASS)
@Target({ElementType.METHOD, ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE})
public @interface Nullable {
  String value() default "";
}
```

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaEvolutionTests.java
```diff
@@ -0,0 +1,134 @@
...
+
+        SerializedBytes<Object> bytes = new SerializationOutput(factory).serialize(
+                n, TestSerializationContext.testSerializationContext);
+
+        new FileOutputStream(new File(getPath(resource).toUri())).write(bytes.getBytes());
+        */
+
+        N1 n2 = new DeserializationInput(factory).deserialize(
+                new SerializedBytes<>(Files.readAllBytes(
+                        Paths.get(this.getClass().getResource(resource).getPath()))),
```
shamsasari: If you rebase onto master you'll have access to `writeTestResource()` and `readTestResource()`, which will remove all of the boilerplate.

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaEvolutionTests.java
```diff
@@ -0,0 +1,134 @@
...
+        // Uncomment to regenerate the base state of the test
+        /*
+        N2 n = new N2(w);
+
+        SerializedBytes<Object> bytes = new SerializationOutput(factory).serialize(
+                n, TestSerializationContext.testSerializationContext);
+
+        new FileOutputStream(new File(getPath(resource).toUri())).write(bytes.getBytes());
+        */
+        Assertions.assertThatThrownBy(() -> new DeserializationInput(factory).deserialize(
```
shamsasari: Static import `assertThatThrownBy`

shamsasari:
Or just look at the Kotlin code!

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/EvolutionSerializer.kt
fenryka: done

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaEvolutionTests.java
fenryka: Hopefully, they're callable from Java in a nice way....

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializationHelper.kt
```diff
@@ -540,3 +540,18 @@ fun hasCordaSerializable(type: Class<*>): Boolean {
             || type.interfaces.any(::hasCordaSerializable)
             || (type.superclass != null && hasCordaSerializable(type.superclass))
 }
+
+fun isJavaPrimitive(type: Class<*>) = type in JavaPrimitiveTypes.primativeTypes
+
+private object JavaPrimitiveTypes {
+    val primativeTypes = hashSetOf<Class<*>>(
```
shamsasari: Can't we use `Class.isPrimitive`?

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaEvolutionTests.java
```diff
@@ -0,0 +1,104 @@
...
+
+        public N2(String word, float wibble) {
+            this.word = word;
+            this.wibble = wibble;
+        }
+        public String getWord() { return word; }
+        public float getWibble() { return wibble; }
+    }
+
+    SerializerFactory factory = new SerializerFactory(
```
shamsasari: Use testDefaultFactory instead

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializationHelper.kt
fenryka: that's really an AMQP primitive vs a Java primtive.... the main difference is String

shamsasari: It is Java primitives: https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isPrimitive--

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaEvolutionTests.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class JavaEvolutionTests {
+    @Rule
+    public final ExpectedException exception = ExpectedException.none();
+
+    // Class as it was when it was serialized and written to disk. Uncomment
+    // if the test referencing the object needs regeneraThrowableting.
```
shamsasari: regeneraThrowableting

Issue #3148, opened 2018-05-15
Remove duplicated code between `PersistentIdentityService` and `InMemoryIdentityService`


joeldudleyr3:
**Issue:** There is a **lot** of duplicated code between `PersistentIdentityService` and `InMemoryIdentityService`.

**Proposed fix:** Refactor duplicated code out into `IdentityServiceUtil`.

**Where:** https://github.com/corda/corda/tree/master/node/src/main/kotlin/net/corda/node/services/identity

stanly-johnson:
@joeldudleyr3 I'll work on this 👍

lankydan:
has this been done? If not, I can have a look at it.

joeldudleyr3:
I don't think so - go ahead :) @stanly-johnson - let us know if you're still working on this

stanly-johnson:
No, it's not done. @lankydan go ahead :)

PR #3428, opened 2018-06-24
Remove duplicated code between PersistentIdentityService and InMemoryIdentityService

lankydan:
Move duplicated code in InMemoryIdentityService and PersistentIdentityService to IdentityServiceUtil

Turned IdentityServiceUtil into a class with static functions
Moved duplicated code to IdentityServiceUtil
InMemoryIdentityService and PersistentIdentityService now delegate to IdentityServiceUtil where logic is shared


lankydan:
i think i failed to tidy up my branch properly which is why there are so many commits...

I also wonder if some of this code could be instead moved to default methods within IdentityService instead of IdentityServiceUtil.

joeldudleyr3:
No worries about the number of commits. We can squash them together when we merge.

@shamsasari Would you be able to take a look at this?

shamsasari:
https://github.com/corda/corda/issues/3148

joeldudleyr3:
Shams, work has stopped on the other PR. I'll close it.

lankydan:
sure I i'll make those changes when I can, wanted a second opinion on it since not everyone likes using default methods :)

lankydan:
Below is the new summary of my changes

Move duplicated code in InMemoryIdentityService and PersistentIdentityService into default functions within IdentityService

- Moved duplicated code into IdentityService using default function
- Move verifyIdentity into IdentityServiceUtil
- PersistentIdentityService normally relies on its default functions but puts them into transactions are calls them via super call
- Remove overriden functions from both PersistentIdentityService and InMemoryIdentityService that can instead rely on the default functions
- Change the throws declaration on assertOwnership to the correct exception

core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
```diff
@@ -63,7 +71,7 @@ interface IdentityService {
      * @param key The owning [PublicKey] of the [Party].
      * @return Returns a [Party] with a matching owningKey if known, else returns null.
      */
-    fun partyFromKey(key: PublicKey): Party?
+    fun partyFromKey(key: PublicKey) = certificateFromKey(key)?.party
```
shamsasari: Always specify the return type for public methods, especially ones which are API.

core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
```diff
@@ -37,10 +38,17 @@ interface IdentityService {
      * Asserts that an anonymous party maps to the given full party, by looking up the certificate chain associated with
      * the anonymous party and resolving it back to the given full party.
      *
-     * @throws IllegalStateException if the anonymous party is not owned by the full party.
+     * @throws UnknownAnonymousPartyException if the anonymous party is not owned by the full party.
```
shamsasari: This is an interesting one. The two implementations always threw  `UnknownAnonymousPartyException `, so any user relying on this would have caught this and not `IllegalStateException`. I think this is the correct change, rather than having this method throw `IllegalStateException`. 

However, can you add an entry to `changelog.rst`, something like "Docs for IdentityService. assertOwnership updated to correctly state that an UnknownAnonymousPartyException is thrown rather than IllegalStateException."

core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
```diff
@@ -102,7 +124,8 @@ interface IdentityService {
      * @return the well known identity.
      * @throws IllegalArgumentException
      */
-    fun requireWellKnownPartyFromAnonymous(party: AbstractParty): Party
+    fun requireWellKnownPartyFromAnonymous(party: AbstractParty) = wellKnownPartyFromAnonymous(party)
```
shamsasari: Convert to block body

node/src/main/kotlin/net/corda/node/services/identity/IdentityServiceUtil.kt
```diff
@@ -2,23 +2,72 @@ package net.corda.node.services.identity
...
+import java.security.InvalidAlgorithmParameterException
+import java.security.cert.CertPathValidatorException
+import java.security.cert.CertificateExpiredException
+import java.security.cert.CertificateNotYetValidException
+import java.security.cert.TrustAnchor
 
+/**
+ * Contains utility methods for use in IdentityService classes
+ */
+class IdentityServiceUtil {
```
shamsasari: We have a `IdentityServiceInternal` interface. You can put these methods there.

node/src/main/kotlin/net/corda/node/services/identity/InMemoryIdentityService.kt
```diff
@@ -46,27 +44,12 @@ class InMemoryIdentityService(identities: List<out PartyAndCertificate> = emptyL
     @Throws(CertificateExpiredException::class, CertificateNotYetValidException::class, InvalidAlgorithmParameterException::class)
     override fun verifyAndRegisterIdentity(identity: PartyAndCertificate): PartyAndCertificate? {
```
shamsasari: Would it make more sense to implement this method in `IdentityServiceInternal` but also introducing `registerIdentity` as an abstract method there as well?

lankydan:
ill look at these requested changes when i can, busy, busy, busy. But after having a quick look through I think I agree with them.

docs/source/changelog.rst
```diff
@@ -6,6 +6,11 @@ release, see :doc:`upgrade-notes`.
 
 Unreleased
 ----------
+* Removed IdentityServiceUtil and added extra default functions to the IdentityServiceInternal interface.
```
shamsasari: The changelog is more for the app developers and how the changes may affect them (i.e. you!) Your refactoring here will not be visible so no need to mentioned it here.

node/src/main/kotlin/net/corda/node/services/api/IdentityServiceInternal.kt
```diff
@@ -1,11 +1,76 @@
...
+                log.warn("$space${certificate.subjectX500Principal}")
+            }
+            throw e
+        }
+        // Ensure we record the first identity of the same name, first
+        val wellKnownCert = identityCertChain.single { CertRole.extract(it)?.isWellKnown ?: false }
+        if (wellKnownCert != identity.certificate) {
+            val idx = identityCertChain.lastIndexOf(wellKnownCert)
+            val firstPath = X509Utilities.buildCertPath(identityCertChain.slice(idx until identityCertChain.size))
+            verifyIdentity(trustAnchor, PartyAndCertificate(firstPath))
```
shamsasari: Looking at the original `verifyAndRegisterIdentity`, because it does "registering" as well, the recursive call that occurs here also registers this `firstPath` cert path. However we're no longer doing that. This can be easily fixed by calling `registerIdentity` at the end of this method, matching what the original `verifyAndRegisterIdentity` does.

In fact this method can be the implementation of `verifyAndRegisterIdentity`, with the two implementations just having to implement `registerIdentity`.

lankydan:
good catch

PR #3430, opened 2018-06-25
IdentitySyncFlow - loadStates of an input only if it is in local storage

cxyzhang0:
This change makes the IdentitySyncFlow usable in more scenarios. 
For example, refer to the issue per https://github.com/corda/obligation-cordapp/issues/5 .
In that example, A has an obligation to B and both A and B are anonymous.  B transfers it to anonymous C.  To make A and C know each other's identity, we use ISF. But it will throw an TransactionResolutionException when C trying to loadStates since C does not have the input in the local storage.
The proposed fix is to first check if an input is in the local storage and load it if it is.

# PR Checklist:

Yes
N/A


cxyzhang0:
For some reason, git pull and push brought in many unrelated commits.  So close this PR and open #3701 as its replacement.

PR #3431, opened 2018-06-25
IdentitySyncFlowWrapper

cxyzhang0:
A simple wrapper of the IdentitySyncFlow by making it standalone with its own flow session.  That makes it easy to use in Responder flows by simplifying messages back and forth.  See https://github.com/corda/obligation-cordapp/issues/5
It is nothing new, and actually it is just a copy of the wrapper in IdentitySyncFlowTests.  We just expose it to the developers.  Especially, it will be very useful for the obligation transfer use case.
The name of the object is subject to change based on reviewer suggestions.

# PR Checklist:

Yes.  And there are new tests in IdentitySyncFlowWrapperTests.kt.
Also, tested in a version of obligation-cordapp - mock network tests of the TransferObligationFlow using the wrapper passed.  But test via the obligation-cordapp's web UI failed with "Counterparty sent session rejection message at unexpected time with message class net.corda.confidential.IdentitySyncFlowWrapper$Initiator is not registered".  
Is there a special way to register Corda provided flows?  BTW, if I implement the wrapper in my CorDapp, I don't have the above registration issue.
If approved, we will update the docs to introduce the wrapper's usage.
Will do upon approval.


cxyzhang0:
Interesting - I just did the following on my fork:
git fetch upstream master
git checkout master
git merge upstream/master
git checkout identitysyncwrapper
git rebase master
git push
and got all the recent commits by others added to the PR.
I hope that does not mess up the review.

shamsasari:
Hey @cxyzhang0 apologies for the delay in getting back to you.

Unfortunately this PR now has a whole bunch of commits from master in it. I think the mistake was to merge master into your feature branch. Instead just

git fetch upstream master
git merge upstream/master

while on your feature branch should have done the trick. If you're unable to fix this it might be easier just to raise a new PR. Thanks.

cxyzhang0:
@shamsasari I don't understand why "All nodes will end up listening for identity sync requests, regardless of weather they're involved with the counterparty in a transaction or not".  I thought explicit otherParty is passed in to the flow so only me and the otherParty are exchanging identities.  Right?  Did I miss something?

The use case is described in corda/obligation-cordapp#5 (https://github.com/corda/obligation-cordapp/issues/5), for transferring obligation:  A has an obligation to B. Now B transfers it to C.  In the initiating transfer flow, B can sync identities with A and C, but how can A and C sync their identities with each other? One way is to do it in the transfer responder flow where A and C each call the proposed IdentitySyncFlowWrapper to the other.  I can implement the wrapper inside the Obligation CorDapp, but thought it might be useful for other use cases.  Hence the PR for the Corda main.

If there is a better approach, please let me know.

Thank you very much for your time.

shamsasari:
The plan is for the confidential identities feature to be incorporated into core Corda at some point once it's fully baked. Once that happens any flow that's `@InitiatedBy` will automatically give the node a listener for requests but with no checks whether the requester is authorised. For example, `SwapIdentitiesFlow` is like this and I'm not sure it's correct - any node can get another node to issue anonymous identities on demand with no controls (so we probably need to fix that).

So coming back to `IdentitySyncFlow` I assumed the same issue. While not as serious I think it still applies. The receiving listener node (via `IdentitySyncFlowWrapper.Receive`) will happily learn the full identities of any anonymous identities that's sent to it. Because it's `@InitiatedBy` the flow has no context to determine whether it should be engaging with the counterparty at all. What if they're mistakenly leaking the identities (or perhaps even maliciously) and now all of a sudden the node has de-anonymised identities it should not have, and it had no way of blocking it?

This is why I suggested this workflow should exist within the specific CorDapp, because then the  `@InitiatedBy` flow can first do some app-specific checks to make sure it has an existing relationship with the counterparty. In the Obligation CorDapp example, the `Initiator` flow could first send the partially signed transaction of the obligation move to A so that it can see that 1) it's involved and 2) the move has been approved by B and so it's OK to be syncing identities.

Does that make sense, or have I missed the point and overthinking this? Let me know.

cxyzhang0:
@shamsasari I think I understand your concerns. I appreciate your patience in explaining. The wrapper can be implemented on a CorDapp by CordDapp basis without unduly exposing any potential risks by the platform (Corda core).  I will close this PR and I can move forward without it.

PR #3432, opened 2018-06-25
[CORDA-1663]: Node won't start if cordapps generate states prior to deletion (fix).

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1663

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
+++ b/node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
@@ -567,7 +567,7 @@ class NodeVaultService(
             }
         }
         if (unknownTypes.isNotEmpty()) {
-            log.warn("There are unknown contract state types in the vault, which will prevent these states from being used. Names are ${unknownTypes.joinToString(", ", "[", "]")}.")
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -519,13 +552,24 @@ class NodeVaultService(
...
+            concreteType?.let {
+                val contractInterfaces = deriveContractInterfaces(it)
+                contractInterfaces.map {
+                    val contractInterface = contractInterfaceToConcreteTypes.getOrPut(it.name) { mutableSetOf() }
+                    contractInterface.add(it.name)
+                }
+            }
+        }
+        if (unknownTypes.isNotEmpty()) {
+            log.warn("There are unknown contract state types in the vault, which will prevent these states from being used. Names are ${unknownTypes.joinToString(", ", "[", "]")}.")
```
adagys: maybe expand the comment with something like "the relevant CorDapps must be loaded for these states to be used"

sollecitom: Done

PR #3434, opened 2018-06-25
Use an expiry time of a century instead of a decade for dev certs.

mikehearn:
This is effectively the same as not expiring (I plan to have died of an over-generous lunch habit, by the time I'm 134).

### DIFF
--- a/docs/source/permissioning.rst
+++ b/docs/source/permissioning.rst
@@ -76,6 +76,11 @@ required public/private keypairs and certificates using Bouncy Castle. You can f
 repository <https://github.com/corda/corda>`__, under
 ``/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt``.
 
+.. note:: Corda allows any expiry time to be set in identity certificates. However, we recommend that the expiry time
+   is set so far in the future that the certificate will effectively never expire. Certificate expiration has been
+   responsible for causing many outages in the past, and is not an effective way to phase out old algorithms either.
+   It's better and safer not to use it.
+
 Certificate role extension
 --------------------------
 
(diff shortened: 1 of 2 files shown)


mnesbit:
This won't do anything as the built in root and doorman certs are pregenerated at 10 years. We are in fact reverting those too right now as we have complaints from customers when we changed the keys.

jamesbr3:
It's not clear what certificates this is applicable to? I'm also not sure what 100-year certificate expiry is supposed to achieve that 10-years is doing already? (even 10 years is too long IMO). 

It is good practice to set certificate expiry to short periods. This forces organisations to put good process in place to ensure certs are renewed. Maximum validity period for TLS certs is 2 years for example (this is industry standard now).

mnesbit:
I can at least confirm that the code generates a new cert with ASN1 encoding of the end date as GeneralTime, not UTCTime, as per the spec. However, this PR has no unit tests and still won't actually affect any DEV certificates due to the precoded parent certs being used.

mikehearn:
Dev certs aren't that important. But I thought this value was used when generating new certs for submission to a doorman - where is the code to generate the identity certs if it's not this ?

mikehearn:
@jamesbr3 see here for rationale:

https://medium.com/corda/certificate-revocation-and-expiry-5e7f4295e44e

mnesbit:
This is the code to generate the certs, but the parent certs are baked into test resources to ensure a consistent root for everyone (https://github.com/corda/corda/tree/master/node-api/src/main/resources/certificates). Talk to michel about this as he effectively owns them at this point.
Parent cert time windows always override child certs time windows.

mikehearn:
Ah, of course. Expiry of parent certs automatically kills the child certs.

So we're basically doomed w.r.t. changing expiry times at this point because we already minted the root certs for the Corda network (and dev certs don't matter)? If that's the case I'll close this PR.

PR #3440, opened 2018-06-26
CORDA-1610: In addition to injecting the old progress tracker after f…

adagys:
…low retry, attach and re-subscribe to child trackers as well.


m4ksio:
Looks like a proper improvement.

PR #3453, opened 2018-06-27
[CORDA-1671] Check if config file exists before trying to parse it.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1671

### DIFF
--- a/node/src/main/java/CordaCaplet.java
+++ b/node/src/main/java/CordaCaplet.java
@@ -35,7 +35,7 @@ public class CordaCaplet extends Capsule {
             ConfigParseOptions parseOptions = ConfigParseOptions.defaults().setAllowMissing(false);
             Config defaultConfig = ConfigFactory.parseResources("reference.conf", parseOptions);
             Config baseDirectoryConfig = ConfigFactory.parseMap(Collections.singletonMap("baseDirectory", baseDir));
-            Config nodeConfig = ConfigFactory.parseFile(configFile, parseOptions);
+            Config nodeConfig = configFile.exists() ? ConfigFactory.parseFile(configFile, parseOptions) : ConfigFactory.empty();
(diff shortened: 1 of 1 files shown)


node/src/main/java/CordaCaplet.java
```diff
@@ -35,7 +35,7 @@ private Config parseConfigFile(List<String> args) {
             ConfigParseOptions parseOptions = ConfigParseOptions.defaults().setAllowMissing(false);
             Config defaultConfig = ConfigFactory.parseResources("reference.conf", parseOptions);
             Config baseDirectoryConfig = ConfigFactory.parseMap(Collections.singletonMap("baseDirectory", baseDir));
-            Config nodeConfig = ConfigFactory.parseFile(configFile, parseOptions);
+            Config nodeConfig = configFile.exists() ? ConfigFactory.parseFile(configFile, parseOptions) : ConfigFactory.empty();
```
rick-r3: What happens when we do not specify "-version" and start a node without a config file?  Does it later complain that something mandatory is missing?

I'm thinking this is fine if a node actually still won't start without a config file, and this just solves the issue for "-version".

florian-f: > What happens when we do not specify "-version" and start a node without a config file? Does it later complain that something mandatory is missing?

It will fail at a later point and the node won't start up, throwing a FileNotFoundException. There's a different issue about making this more user-friendly, but I think we shelved that because the command line args parsing is supposed to be overhauled anyways.

rick-r3 (approved):

mikehearn:
I think this PR does the same thing and might work better:

https://github.com/corda/corda/pull/3512

florian-f:
Closing this, see https://github.com/corda/corda/pull/3512

PR #3454, opened 2018-06-27
CORDA-1660: Wiring up the CordaRPCClient class loader to the p2p serialisation context

shamsasari:
This is to allow the standalone shell to be able to receive WireTransactions containing Cash.State objects.

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/serialization/amqp/AMQPClientSerializationScheme.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/serialization/amqp/AMQPClientSerializationScheme.kt
@@ -3,6 +3,7 @@ package net.corda.client.rpc.internal.serialization.amqp
 import net.corda.core.cordapp.Cordapp
 import net.corda.core.serialization.ClassWhitelist
 import net.corda.core.serialization.SerializationContext
+import net.corda.core.serialization.SerializationContext.*
 import net.corda.core.serialization.SerializationCustomSerializer
 import net.corda.core.serialization.internal.SerializationEnvironment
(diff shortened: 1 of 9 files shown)


client/rpc/src/main/kotlin/net/corda/client/rpc/internal/serialization/amqp/AMQPClientSerializationScheme.kt
```diff
@@ -29,25 +30,26 @@ class AMQPClientSerializationScheme(
...
+                    rpcServerContext = AMQP_RPC_SERVER_CONTEXT
+            )
         }
     }
 
-    override fun canDeserializeVersion(magic: CordaSerializationMagic, target: SerializationContext.UseCase) =
-        magic == amqpMagic && (
-            target == SerializationContext.UseCase.RPCClient || target == SerializationContext.UseCase.P2P)
+    override fun canDeserializeVersion(magic: CordaSerializationMagic, target: SerializationContext.UseCase): Boolean {
+        return magic == amqpMagic && (target == UseCase.RPCClient || target == UseCase.P2P)
```
fenryka: pointless refactor

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactory.kt
```diff
@@ -74,11 +74,11 @@ open class SerializerFactory(
 
     @DeleteForDJVM
     constructor(whitelist: ClassWhitelist,
-                classLoader: ClassLoader,
+                carpenterClassLoader: ClassLoader,
```
fenryka: technically this is the class loader *for* the capetner, not *its* classloader... probably not an important difference

PR #3457, opened 2018-06-27
CORDA-1686: Make Node Explorer release its RPC connection on shutdown.

chrisr3:
While we have already made the offending RPC thread into a daemon, Node Explorer should really also be closing its RPC connection when it shuts down.

### DIFF
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
@@ -49,7 +49,7 @@ class NodeMonitorModel : AutoCloseable {
     private val stateMachineTransactionMappingSubject = PublishSubject.create<StateMachineTransactionMapping>()
     private val progressTrackingSubject = PublishSubject.create<ProgressTrackingEvent>()
     private val networkMapSubject = PublishSubject.create<MapChange>()
-    private lateinit var rpcConnection: CordaRPCConnection
+    private var rpcConnection: CordaRPCConnection? = null
 
     val stateMachineUpdates: Observable<StateMachineUpdate> = stateMachineUpdatesSubject
     val vaultUpdates: Observable<Vault.Update<ContractState>> = vaultUpdatesSubject
@@ -89,12 +89,10 @@ class NodeMonitorModel : AutoCloseable {
      * Disconnects from the Corda node for a clean client shutdown.
      */
     override fun close() {
-        if (::rpcConnection.isInitialized) {
-            try {
-                rpcConnection.forceClose()
-            } catch (e: Exception) {
-                logger.error("Error closing RPC connection to node", e)
-            }
+        try {
+            rpcConnection?.forceClose()
+        } catch (e: Exception) {
+            logger.error("Error closing RPC connection to node", e)
         }
     }
 
@@ -159,8 +157,10 @@ class NodeMonitorModel : AutoCloseable {
     }
 
     private fun performRpcReconnect(nodeHostAndPort: NetworkHostAndPort, username: String, password: String, shouldRetry: Boolean): List<StateMachineInfo> {
-        rpcConnection = establishConnectionWithRetry(nodeHostAndPort, username, password, shouldRetry)
-        val proxy = rpcConnection.proxy
+        val proxy = establishConnectionWithRetry(nodeHostAndPort, username, password, shouldRetry).let { connection ->
+            rpcConnection = connection
+            connection.proxy
+        }
 
         val (stateMachineInfos, stateMachineUpdatesRaw) = proxy.stateMachinesFeed()
 
@@ -175,7 +175,7 @@ class NodeMonitorModel : AutoCloseable {
                     // It is good idea to close connection to properly mark the end of it. During re-connect we will create a new
                     // client and a new connection, so no going back to this one. Also the server might be down, so we are
                     // force closing the connection to avoid propagation of notification to the server side.
-                    rpcConnection.forceClose()
+                    rpcConnection?.forceClose()
                     // Perform re-connect.
                     performRpcReconnect(nodeHostAndPort, username, password, shouldRetry = true)
                 })
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
@@ -90,7 +90,7 @@ class NodeMonitorModel : AutoCloseable {
      */
     override fun close() {
         try {
-            rpcConnection?.forceClose()
+            rpcConnection?.notifyServerAndClose()
         } catch (e: Exception) {
             logger.error("Error closing RPC connection to node", e)
         }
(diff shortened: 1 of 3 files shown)


tools/explorer/src/main/kotlin/net/corda/explorer/views/LoginView.kt
```diff
@@ -64,12 +67,13 @@ class LoginView : View(WINDOW_TITLE) {
...
                     }.showAndWait().get()
                     if (button == ButtonType.OK) {
+                        nodeModel?.close()
                         exitProcess(0)
                     }
                 }
             }
         }.showAndWait().get()
-        if (status != LoginStatus.loggedIn) login()
+        return if (status == LoginStatus.loggedIn) nodeModel else login()
```
vkolomeyko: Should we add tail recursion here?

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -175,18 +188,17 @@ class NodeMonitorModel {
     }
 
     private fun establishConnectionWithRetry(nodeHostAndPort: NetworkHostAndPort, username: String, password: String, shouldRetry: Boolean): CordaRPCConnection {
-
         val retryInterval = 5.seconds
 
+        val client = CordaRPCClient(
```
vkolomeyko: I deliberately put `client` creation inside `do {} while` loop to ensure that on every iteration we create *brand* new connection and have no legacy from previous connection attempt.

Unless you have a strong preference for having `client` outside of the loop, I would want to keep it that way.

chrisr3: As I understand it, the client is meant to be re-used like this. Perhaps refer this to @exfalso?

tools/explorer/src/main/kotlin/net/corda/explorer/views/LoginView.kt
chrisr3: I've refactored the existing tail recursion. I didn't want to change how this function works completely.

chrisr3:
Agreed, there's no need to promote this into GA.

tools/explorer/src/main/kotlin/net/corda/explorer/views/LoginView.kt
vkolomeyko: `tailrec fun login(): NodeMonitorModel?` seems to just work...
My concern that in case of multiple login attempts we might be unnecessarily growing the stack.

chrisr3: Ah, I see. I was unaware of Kotlin's `tailrec` keyword.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
vkolomeyko: I can see that current implementation of `CordaRPCClient` is just connectivity parameters container.
And even calling `start` it returns a brand new `CordaRPCConnection` every time in the current implementation.
However, implementation may change in the future and `CordaRPCClient` may start to have some state.

I defer to @exFalso to opine, but my view that it is safer to create a branch new instance of `CordaRPCClient` on every loop such that any references rooted from previous `CordaRPCClient` are correctly GC'd. Especially given that we can easily do that in the current code and  `CordaRPCClient` is relatively cheap to construct and performance in the method in question is not really a concern.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -49,6 +49,7 @@ class NodeMonitorModel {
     private val stateMachineTransactionMappingSubject = PublishSubject.create<StateMachineTransactionMapping>()
     private val progressTrackingSubject = PublishSubject.create<ProgressTrackingEvent>()
     private val networkMapSubject = PublishSubject.create<MapChange>()
+    private lateinit var rpcConnection: CordaRPCConnection
```
exFalso: This won't work if the client is reconnecting (cannot double-assign to lateinit)

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
exFalso: Either is fine. The only state `CordaRPCClient` holds is the serialization scheme initialisation, so the current code saves a bit on that. The client itself will never have "important" state, that's all in the Connection

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
chrisr3: It seems very strange that you can't re-assign a `var` - by definition! And this test doesn't explode:
```kotlin
    private lateinit var foo: String

    @Test
    fun reassignLateInit() {
        foo = "Hello"
        foo = "And again!"
        foo = "For the hat-trick!"
    }
```
However, I've removed the `lateinit` anyway.

vkolomeyko:
@exFalso - Are you sure about `lateinit` ?
E.g. this test passes fine:
```
class LateInitTest {
    private lateinit var variable: String

    @Test
    fun test() {
        variable = "foo"
        variable = "bar"
        assertEquals("bar", variable)
    }
}
```

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -49,6 +49,7 @@ class NodeMonitorModel {
     private val stateMachineTransactionMappingSubject = PublishSubject.create<StateMachineTransactionMapping>()
     private val progressTrackingSubject = PublishSubject.create<ProgressTrackingEvent>()
     private val networkMapSubject = PublishSubject.create<MapChange>()
+    private var rpcConnection: CordaRPCConnection? = null
```
vkolomeyko: I think we can use `lateinit`
This will allow using `CordaRPCConnection` instead of `CordaRPCConnection?`

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
vkolomeyko: If either is fine - I would prefer to have a version with least changes to existing code :-)
But I am flexible.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
```diff
@@ -84,6 +85,17 @@ class NodeMonitorModel {
      */
     class CordaRPCOpsWrapper(val cordaRPCOps: CordaRPCOps)
 
+    /**
+     * Disconnects from the Corda node for a clean client shutdown.
+     */
+    override fun close() {
+        try {
+            rpcConnection?.forceClose()
```
vkolomeyko: Shall we not use `notifyServerAndClose` here? - seems to be a more polite version of closing connection which ensures that server side all the necessary clean-up promptly.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/NodeMonitorModel.kt
exFalso: My bad, i was sure this was the behaviour!

PR #3443, opened 2018-06-26
Specify Notary in `CashIssueAndPaymentFlow` and `PaymentRequest`

thschroeter:
Ensure the explorer works with networks with multiple notaries.

PR #3461, opened 2018-06-27
Fixed broken smoke test caused by no notaries in the network parameters

shamsasari:
For now the workaround is to manually select the notary. The proper solution is to add notaries to the network parameters in smoke tests.


shamsasari:
Broken by https://github.com/corda/corda/pull/3443

PR #3464, opened 2018-06-28
Link to where the scripts can be found

joeldudleyr3:


joeldudleyr3:
I chose to leave it in two separate parts because you have to clone the repo first, and then navigate to the folder to run the scripts. Thoughts?

chrisr3:
Can't you do both, i.e. turn `tools/demobench` into a clickable link?

joeldudleyr3:
Updated. LMKWYT

PR #3468, opened 2018-06-28
CORDA-1693: Ensure that DemoBench's RPC connections terminate on shutdown.

chrisr3:
On rare occasions, DemoBench has not shut down in a timely manner because `NodeRPC` timer threads were left running. Therefore:
- Turn these timer threads into daemons.
- Interrupt the timer threads on shutdown, in case they are stuck (e.g.) waiting for network I/O.
- Ensure that all `NodeTerminalView` resources are destroyed on close, despite any exceptions.

Also fix whitespace issues and some compiler warnings.


vkolomeyko:
Probably not for GA as it is rather late in the cycle.

PR #3473, opened 2018-06-28
[WIP] Fix flaky test

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1698


florian-f:
The API stability check failed because MockServices now has TestClock instead of Clock.
I spoke to @tudor-malene , he agreed that this change would not break anything for users. Updated api-current.txt to reflect the change.

PR #3490, opened 2018-07-02
Document node command-line options

joeldudleyr3:


docs/source/running-a-node.rst
```diff
@@ -43,6 +43,26 @@ Optionally run the node's webserver as well by opening a terminal window in the
 
 .. warning:: The node webserver is for testing purposes only and will be removed soon.
 
+Command-line options
+~~~~~~~~~~~~~~~~~~~~
+The node can optionally be started with the following command-line options:
+
+* ``base-directory`` - the path to the node's base directory
```
shamsasari: To me this is duplication. Running with `--help` will display all this anyway, with some extra information like default values. Why not just paste that output here?

joeldudleyr3: I don't see it as duplication, since that information isn't available via the docsite.

Are you saying that you want me to document default values as well? Happy to do that in this PR or in a separate one :)

shamsasari: I disagree with the idea of documenting the command line flags separately when that documentation is already available with the `--help` flag. With the blob inspector I just mentioned that the `--help` flag is available. As a compromise you could take the help output of the corda.jar and just paste it here as a code block.

PR #3498, opened 2018-07-03
[CORDA-1659]: Improve handling/logging of failed address binding.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1659


node/src/main/kotlin/net/corda/node/services/rpc/ArtemisRpcBroker.kt
```diff
@@ -44,7 +46,14 @@ internal class ArtemisRpcBroker internal constructor(
 
     override fun start() {
         logger.debug("Artemis RPC broker is starting.")
-        server.start()
+        try {
+            server.start()
+        } catch (e: Errors.NativeIoException) {
```
anthonykeenan: On Windows the error that gets thrown is a `java.net.BindException` rather than a `io.netty.channel.unix.Errors.NativeIoException`, which I'm assuming is because I'm not using unix - this is why the tests are failing on my machine.

anthonykeenan:
I've pushed a commit that fixes the issues on my machine - see what you think. `IOException` is the lowest common denominator between the *nix/windows exceptions

PR #3500, opened 2018-07-03
Some code tidy up ahead of post-GA development.

mnesbit:
This code rationalises the explosion in constructor parameters to the AMQP protocol wrappers.
Also uses correct nested scopes for the MDC based context logging as recommended by Michele.
Also, handles appropriately the SSL Handshake Timeout, although sadly that does require spotting the message string coming out of Netty.


mnesbit:
Also note, I will have to merge this to ENT as there is additonal code there which needs fixes

PR #3502, opened 2018-07-03
CORDA-1516 - Adding notary and privacy-salt to key-concepts-transactions

kchalkias:


kchalkias:
Thanks Joel, any idea on the ideal place for describing Privacy Salt? There was a question the other day on how we deal with the nonces and potential low-entropy components.

kchalkias:
Also, maybe we should at least add the notary, because in Committing transactions we haven't mentioned the notaries yet (and we might give the false impression that it's not part of the tx, but it only belongs to the states) - btw, I received this question a couple of times. _Where does the notary belong to?_

docs/source/key-concepts-transactions.rst
```diff
@@ -115,9 +115,11 @@ As well as input states and output states, transactions contain:
 * Commands
 * Attachments
 * Time-Window
+* Notary
 
-For example, a transaction where Alice pays off £5 of an IOU with Bob using a £5 cash payment, supported by two
-attachments and a time-window, may look as follows:
+For example, a transaction which should be notarised by NotaryClusterA only if received inside the predefined time-window
```
joeldudleyr3: This sentence has become a little complex. How about:

"For example, suppose we have a transaction where Alice uses a £5 cash payment to pay off £5 of an IOU with Bob. This transaction has two supporting attachments and will only be notarised by NotaryClusterA if the notary pool receives it within the specified time-window. This transaction would look as follows:"

joeldudleyr3: RGB prefers "pool" to "cluster" ("cluster" implies trust among members).

docs/source/key-concepts-transactions.rst
```diff
@@ -184,3 +186,13 @@ In some cases, we want a transaction proposed to only be approved during a certa
 
 In such cases, we can add a *time-window* to the transaction. Time-windows specify the time window during which the
 transaction can be committed. We discuss time-windows in the section on :doc:`key-concepts-time-windows`.
+
+Notary
+^^^^^^
+A notary cluster is a network service that provides uniqueness consensus by attesting that, for a given transaction,
```
joeldudleyr3: RGB prefers "pool" to "cluster" ("cluster" implies trust among members).

docs/source/key-concepts-transactions.rst
```diff
@@ -184,3 +186,13 @@ In some cases, we want a transaction proposed to only be approved during a certa
...
+it has not already signed other transactions that consume any of the proposed transaction’s input states.
+The notary cluster provides the point of finality in the system.
```
joeldudleyr3: RGB prefers "pool" to "cluster" ("cluster" implies trust among members).

kchalkias:
back to you @joeldudleyr3

PR #3504, opened 2018-07-03
CORDA-1712 - Make documentation clearer for Network Bootstrapper

anthonykeenan:
Tried to add clarity to the bootstrapper documentation around

a) bootstrapping with cordapps
b) whitelisting contracts
c) regenerating the whitelist

### DIFF
--- a/docs/source/network-bootstrapper.rst
+++ b/docs/source/network-bootstrapper.rst
@@ -45,8 +45,6 @@ will generate directories containing three nodes: ``notary``, ``partya`` and ``p
 that comes with the bootstrapper. If a different version of Corda is required then simply place that ``corda.jar`` file
 alongside the configuration files in the directory.
 
-The directory can also contain CorDapp JARs which will be copied to each node's ``cordapps`` directory.
-
 You can also have the node directories containing their "node.conf" files already laid out. The previous example would be:
 
 .. sourcecode:: none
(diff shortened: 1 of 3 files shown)


docs/source/network-bootstrapper.rst
```diff
@@ -61,33 +61,34 @@ You can also have the node directories containing their "node.conf" files alread
 
 Similarly, each node directory may contain its own ``corda.jar``, which the bootstrapper will use instead.
 
-Synchronisation
-~~~~~~~~~~~~~~~
+Providing CorDapps to the Network Bootstrapper
```
shamsasari: You should delete line 48 above "The directory can also contain CorDapp JARs ..."

docs/source/network-bootstrapper.rst
```diff
@@ -99,3 +100,79 @@ For example:
...
+        ├── node.conf
+        ├── network-parameters      // The contracts from cordapp-b are appended to the whitelist in network-parameters
+        └── cordapps
+            ├── cordapp-a.jar
+            └── cordapp-b.jar       // The updated cordapp is placed in the nodes cordapp directory
+
+.. note:: The whitelist can only ever be appended to. Once added a contract implementation can never be removed.
+
+.. note:: The bootstrapper is provided for test deployments and can only generate information for nodes collected on
+    the same machine. If a network needs to be updated using the bootstrapper once deployed, the nodes will need
```
shamsasari: This seems to repeat the paragraph starting on line 115

docs/source/network-bootstrapper.rst
```diff
@@ -99,3 +100,79 @@ For example:
...
+Synchronisation
+~~~~~~~~~~~~~~~
+
+This tool only bootstraps a network. It cannot dynamically update if a new node needs to join the network or if an existing
+one has changed something in their node-info, e.g. their P2P address. For this the new node-info file will need to be placed
+in the other nodes' ``additional-node-infos`` directory. A simple way to do this is to use `rsync <https://en.wikipedia.org/wiki/Rsync>`_.
+However, if it's known beforehand the set of nodes that will eventually form part of the network then all the node directories
+can be pre-generated in the bootstrap and only started when needed.
+
+Running the bootstrapper again on the same network will allow a new node to be added or an existing one to have its updated
```
blsemo: Is that not a contradiction with line 109? Can it be used to update a network or not?
Also, it would  be good to have a description what happens when you run it on a fresh set of directories, and what happens when you run it on an existing network (e.g. any modifications to node.conf files are lost?)
Also, how does it interact with dev mode, and where do certs come from when not in dev mode?

docs/source/network-bootstrapper.rst
```diff
@@ -99,3 +100,79 @@ For example:
...
+node-info re-distributed. However this comes at the expense of having to temporarily collect the node directories back
+together again under a common parent directory.
```
blsemo: This "collecting back together" is mentioned several times on this page, but is only mentioned as an abstract concept. A concrete description how to bootstrap/update a distributed network would be very helpful.

docs/source/network-bootstrapper.rst
```diff
@@ -99,3 +100,79 @@ For example:
 
     net.corda.finance.contracts.asset.Cash
     net.corda.finance.contracts.asset.CommercialPaper
+
+.. _network-bootstrapper-synchronisation:
+
+Synchronisation
+~~~~~~~~~~~~~~~
+
```
blsemo: I don't fully understand the title "Synchronisation" here - what is being synchronised?

anthonykeenan:
I've had another iteration to try and address your issues @blsemo

docs/source/network-bootstrapper.rst
```diff
@@ -99,3 +98,155 @@ For example:
 
     net.corda.finance.contracts.asset.Cash
     net.corda.finance.contracts.asset.CommercialPaper
+
+Modifying a bootstrapped network
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The network bootstrapper is provided as a development tool for setting up Corda networks for development and testing.
+There is some limited functionality which can be used to make changes to a network, but for anything more complicated consider
+using a :doc:`network-map` server.
```
shamsasari: Does this suggestion contradict https://docs.corda.net/head/permissioning.html#creating-your-own-compatibility-zone?

docs/source/network-bootstrapper.rst
```diff
@@ -99,3 +98,155 @@ For example:
...
+
+The bootstrapper will generate a directory and the ``node-info`` file for PartyB, and will also make sure a copy of each
+nodes' ``node-info`` file is in the ``additional-node-info`` directory of every node. Any other files in the existing nodes,
+such a generated keys, will be unaffected.
+
+.. note:: The bootstrapper is provided for test deployments and can only generate information for nodes collected on
+    the same machine. If a network needs to be updated using the bootstrapper once deployed, the nodes will need
+    collecting back together.
+
+Modifying an existing node on the network
```
shamsasari: I don't think this warrants its own section now. Under the "Modifying" section above, just have an additional paragraph at the end which states if you can't move the node directories back then consider using rsync or have pre-generated nodes, etc.

docs/source/network-bootstrapper.rst
anthonykeenan: Yes... because it looks like there is only partial documentation for the network-bootstrappers new cloud functionality. I've created a jira to address that: https://r3-cev.atlassian.net/browse/CORDA-1746 but I think it's out of scope for this.

PR #3511, opened 2018-07-04
Clear some cut-and-paste cruft out of our Gradle files.

chrisr3:
The experimental modules are part of the main project and so inherit repositories, plugins and properties.

### DIFF
--- a/experimental/avalanche/build.gradle
+++ b/experimental/avalanche/build.gradle
@@ -1,41 +1,23 @@
 buildscript {
-    ext.kotlin_version = '1.2.40'
-
     repositories {
-        mavenCentral()
         jcenter()
     }
 
     dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.3'
+        classpath "com.github.jengelman.gradle.plugins:shadow:$shadow_version"
     }
 }
 
-plugins {
-    id "org.jetbrains.kotlin.jvm"
-    id 'com.github.johnrengelman.shadow' version '2.0.3'
-    id 'java'
-    id 'application'
-}
-repositories {
-    mavenCentral()
-}
+apply plugin: 'kotlin'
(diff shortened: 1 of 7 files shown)


experimental/avalanche/build.gradle
```diff
@@ -1,41 +1,15 @@
...
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.3'
-    }
-}
-
 plugins {
     id "org.jetbrains.kotlin.jvm"
-    id 'com.github.johnrengelman.shadow' version '2.0.3'
-    id 'java'
+    id 'com.github.johnrengelman.shadow' version '2.0.4'
```
Clintonio: I think we're using this else where right? If we are could you standardise it in the ext part of the root build.gradle?

experimental/avalanche/build.gradle
```diff
@@ -1,41 +1,15 @@
...
-    kotlinOptions {
-        jvmTarget = "1.8"
-    }
-}
-compileTestKotlin {
-    kotlinOptions {
-        jvmTarget = "1.8"
-    }
+    compile "info.picocli:picocli:3.0.1"
+    testCompile group: 'junit', name: 'junit', version: junit_version
```
Clintonio: We should standardise on one style - string or map.

Clintonio:
Also call it what it is - cargo culting.

experimental/avalanche/build.gradle
chrisr3: It's there already - but this is the Plugins DSL and I cannot use it here :man_facepalming:

experimental/avalanche/build.gradle
chrisr3: I'll standardise on "string" here.

experimental/avalanche/build.gradle
Clintonio: Can we just kill the DSL? It's sub-standard for us right now?

PR #3514, opened 2018-07-04
Enabled warnings as errors

shamsasari:


build.gradle
```diff
@@ -242,6 +242,12 @@ allprojects {
             exclude module: 'isolated'
         }
     }
+
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+        kotlinOptions {
+            allWarningsAsErrors = true
```
chrisr3: Don't we already have a `KotlinCompile` block? Why not add the `allWarningsAsError=true` line to that instead?

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -48,7 +48,7 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
     @DeleteForDJVM
     constructor(componentGroups: List<ComponentGroup>) : this(componentGroups, PrivacySalt())
 
-    @Deprecated("Required only in some unit-tests and for backwards compatibility purposes.", ReplaceWith("WireTransaction(val componentGroups: List<ComponentGroup>, override val privacySalt: PrivacySalt)"), DeprecationLevel.WARNING)
+    @Deprecated("Required only for backwards compatibility purposes.", ReplaceWith("WireTransaction(val componentGroups: List<ComponentGroup>, override val privacySalt: PrivacySalt)"), DeprecationLevel.WARNING)
```
chrisr3: This is actually used by `EnclaveletTest`.

build.gradle
shamsasari: Good spot.

shamsasari:
Successor to https://github.com/corda/corda/pull/2516

PR #3518, opened 2018-07-04
Reverted removal of deduplicationSeed parameter

shamsasari:
I accidentally removed the `deduplicationSeed` parameter of `Checkpoint.create`,  which is used by the multi-threaded SMM


shamsasari:
Looks like `deduplicationSeed` isn't used at all.

PR #3519, opened 2018-07-04
Add support for hex and base64 inputs to the blob inspector

mikehearn:
Also, tidy up a couple of minor things in the docs and an error message.

### DIFF
--- a/docs/source/tools-index.rst
+++ b/docs/source/tools-index.rst
@@ -1,10 +1,12 @@
 Tools
 =====
 
+Corda provides various command line and GUI tools to help you as you work. Along with the three below, you may also
+wish to try the :doc:`blob-inspector`.
+
 .. toctree::
    :maxdepth: 1
 
    network-bootstrapper
-   blob-inspector
    demobench
    node-explorer
--- a/tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
+++ b/tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
@@ -2,6 +2,7 @@ package net.corda.blobinspector
 
 import com.fasterxml.jackson.core.JsonFactory
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
+import com.google.common.io.BaseEncoding
 import com.jcabi.manifests.Manifests
 import net.corda.client.jackson.JacksonSupport
 import net.corda.core.internal.isRegularFile
@@ -43,18 +44,21 @@ fun main(args: Array<String>) {
 }
 
 @Command(
-        name = "Blob Inspector",
+        name = "blob-inspector",
         versionProvider = CordaVersionProvider::class,
-        mixinStandardHelpOptions = true, // add --help and --version options,
+        mixinStandardHelpOptions = true,   // add --help and --version options,
         showDefaultValues = true,
-        description = ["Inspect AMQP serialised binary blobs"]
+        description = ["Convert AMQP serialised binary blobs to text"]
 )
 class BlobInspector : Runnable {
     @Parameters(index = "0", paramLabel = "SOURCE", description = ["URL or file path to the blob"], converter = [SourceConverter::class])
-    var source: URL? = null
+    private var source: URL? = null
 
     @Option(names = ["--format"], paramLabel = "type", description = ["Output format. Possible values: [YAML, JSON]"])
-    private var formatType: FormatType = FormatType.YAML
+    private var formatType: OutputFormatType = OutputFormatType.YAML
+
+    @Option(names = ["--input-format"], paramLabel = "type", description = ["Input format. If the file can't be decoded with the given value it's auto-detected, so you should never normally need to specify this. Possible values: [BINARY, HEX, BASE64]"])
+    private var inputFormatType: InputFormatType = InputFormatType.BINARY
 
     @Option(names = ["--full-parties"],
             description = ["Display the owningKey and certPath properties of Party and PartyAndReference objects respectively"])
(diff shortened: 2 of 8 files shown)


tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -98,6 +102,31 @@ class Main : Runnable {
...
+    private fun parseToBinary(format: InputFormatType, inputBytes: ByteArray): ByteArray? {
+        try {
+            val bytes = when (format) {
+                InputFormatType.BINARY -> inputBytes
+                InputFormatType.HEX -> BaseEncoding.base16().decode(String(inputBytes).trim().toUpperCase())
+                InputFormatType.BASE64 -> BaseEncoding.base64().decode(String(inputBytes).trim())
+            }
+            require(bytes.size > amqpMagic.size) { "Insufficient bytes for AMQP blob" }
+            return if (bytes.copyOf(amqpMagic.size).contentEquals(amqpMagic.bytes))
+                bytes
```
shamsasari: add info log that this format type was chosen

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -69,15 +73,15 @@ class Main : Runnable {
             System.setProperty("logLevel", "trace")
         }
 
-        val bytes = source!!.readBytes().run {
-            require(size > amqpMagic.size) { "Insufficient bytes for AMQP blob" }
-            sequence()
+        val inputBytes = source!!.readBytes()
+        val bytes = parseToBinaryRelaxed(inputFormatType, inputBytes)
+        if (bytes == null) {
+            println("Error: this input does not appear to be encoded in Corda's AMQP extended format, sorry.")
```
shamsasari: Use `require`  - the exception handling in `main` does the rest.

docs/source/tools-index.rst
```diff
@@ -1,10 +1,12 @@
 Tools
 =====
 
+Corda provides various command line and GUI tools to help you as you work. Along with the three below, you may also
```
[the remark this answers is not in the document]
mikehearn: It looked kind of weird when viewing the page because both sections of the sidebar expand simultaneously if they're the same page, it's supposed to be a tree rather than a DAG I guess. But it's not that important. I can put it back if you like.

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
mikehearn: I threw it manually to keep the flow typing.

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -98,6 +99,32 @@ class Main : Runnable {
...
+        // Try the format the user gave us first, then try the others.
+        return parseToBinary(format, inputBytes) ?: parseToBinary(InputFormatType.HEX, inputBytes)
+        ?: parseToBinary(InputFormatType.BASE64, inputBytes) ?: parseToBinary(InputFormatType.BINARY, inputBytes)
+    }
+
+    private fun parseToBinary(format: InputFormatType, inputBytes: ByteArray): ByteArray? {
+        try {
+            val bytes = when (format) {
+                InputFormatType.BINARY -> inputBytes
+                InputFormatType.HEX -> BaseEncoding.base16().decode(String(inputBytes).trim().toUpperCase())
```
kchalkias: We already have such decoders in `EncodingUtilis`, I don't think importing guava's `BaseEncoding` is required at this stage. Thus, you could probably rewrite those to: 
`InputFormatType.HEX -> String(inputBytes).trim().hexToByteArray()`
and
`InputFormatType.BASE64 -> String(inputBytes).trim().base64ToByteArray()`

mikehearn:
@kchalkias Thanks, good catch, I fixed that to use our own encoding utils and merged with Shams' work. I can now decode the text that the customer sent me on slack. PTAL!

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -43,18 +45,21 @@ fun main(args: Array<String>) {
...
-        mixinStandardHelpOptions = true, // add --help and --version options,
+        mixinStandardHelpOptions = true,   // add --help and --version options,
         showDefaultValues = true,
-        description = ["Inspect AMQP serialised binary blobs"]
+        description = ["Convert AMQP serialised binary blobs to text"]
 )
 class BlobInspector : Runnable {
     @Parameters(index = "0", paramLabel = "SOURCE", description = ["URL or file path to the blob"], converter = [SourceConverter::class])
-    var source: URL? = null
+    private var source: URL? = null
```
shamsasari: This needs to be public.

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -105,6 +107,32 @@ class BlobInspector : Runnable {
         }
     }
 
+    private fun parseToBinaryRelaxed(format: InputFormatType, inputBytes: ByteArray): ByteArray? {
+        // Try the format the user gave us first, then try the others.
+        return parseToBinary(format, inputBytes) ?: parseToBinary(InputFormatType.HEX, inputBytes)
```
shamsasari: It was clearer when you had it one per line

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -105,6 +107,32 @@ class BlobInspector : Runnable {
...
+        ?: parseToBinary(InputFormatType.BASE64, inputBytes) ?: parseToBinary(InputFormatType.BINARY, inputBytes)
+    }
+
+    private fun parseToBinary(format: InputFormatType, inputBytes: ByteArray): ByteArray? {
+        try {
+            val bytes = when (format) {
+                InputFormatType.BINARY -> inputBytes
+                InputFormatType.HEX -> String(inputBytes).trim().toUpperCase().hexToByteArray()
```
kchalkias: `toUpperCase()` is not required, `hexToByteArray` works for mixed-case, we have a related test
`decoding lowercase and mixed HEX` in `EncodingUtilsTest`.

PR #3525, opened 2018-07-05
CORDA-1332 Reference input states

roger3cev:
Original PR: https://github.com/corda/corda/pull/3060
Design doc: https://github.com/corda/enterprise/blob/93dfa7bf493c8b83bfdbdafddad889bf614416a4/docs/source/design/reference-states/design.md

We probably still need to think about:
* How minimum platform version checks should be implemented
* How encumbrances work with reference states
* Currently there's no restriction over which states can be referenced but should references be restricted to only linear states? 
* If ownable states can be reference states, then reference states may potentially leak information about who owns assets (Kostas)

Resolved issues:

* **[ADDED THIS AS AKNOWN LIMITATION]** How using > 2 reference states tied to different notaries in a transaction. The implication being that the user of the reference states likely cannot change the notary. Anyway, to do so would introduce a race condition. Currently, `TransactionBuilder` throws if reference states assigned to different notaries have been added. Not a problem now but will be later on...
* **[MOVED THE ASYNC OPERATION TO AN INTERNAL PACKAGE]** This PR is currently blocked on `FlowAsyncOperation` being an internal API (See API stability check failure) I'm not sure how to build the `WithReferencedStatesFlow` without using this API.


core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
```diff
@@ -65,11 +65,23 @@ sealed class NotaryError {
...
+         */
+        val type: ConsumedStateType = ConsumedStateType.INPUT_STATE
+) {
+    fun copy(hashOfTransactionId: SecureHash): StateConsumptionDetails {
+        return StateConsumptionDetails(hashOfTransactionId, type)
+    }
+}
+
+@CordaSerializable
+enum class ConsumedStateType { INPUT_STATE, REFERENCE_INPUT_STATE }
```
kchalkias: Hm.. I am not sure we should call this `ConsumedStateType ` as referenced-states are not consumed, right?

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -145,19 +148,48 @@ open class TransactionBuilder(
         toLedgerTransaction(services).verify()
     }
 
-    open fun addInputState(stateAndRef: StateAndRef<*>): TransactionBuilder {
+    private fun checkNotary(stateAndRef: StateAndRef<*>) {
         val notary = stateAndRef.state.notary
         require(notary == this.notary) { "Input state requires notary \"$notary\" which does not match the transaction notary \"${this.notary}\"." }
+    }
+
+    private fun checkForInputsAndReferencesOverlap() {
```
kchalkias: Probably if `checkNoDuplicateInputs` in `BaseTransaction` includes this check, then this is not required here. Otherwise, similarly you should check on each `addState` if there is a duplicate as well, not only if they overlap.

kchalkias:
As discussed, we should also agree if we will deal with the potential leak that 

> One can tell if you still own an asset by just adding it as a reference state.

core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
roger3cev: Hmm good point. I just added something here as it meant less changes to the existing code.

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
```diff
@@ -32,13 +34,14 @@ abstract class BaseTransaction : NamedByHash {
...
     private fun checkNotarySetIfInputsPresent() {
-        if (inputs.isNotEmpty()) {
+        if (inputs.isNotEmpty() || references.isNotEmpty()) {
             check(notary != null) { "The notary must be specified explicitly for any transaction that has inputs" }
         }
     }
 
     private fun checkNoDuplicateInputs() {
         check(inputs.size == inputs.toSet().size) { "Duplicate input states detected" }
+        check(references.size == references.toSet().size) { "Duplicate reference states detected" }
```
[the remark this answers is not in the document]
roger3cev: Yes, there is another function which checks this in transaction builder.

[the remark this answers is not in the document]
kchalkias: It should be here though, right?

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
roger3cev: Yes good point.

core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
kchalkias: yeah, literally we should call all of them input states, and the current inputs should be called consumed and the referenced_inputs should be called referenced, but this would imply api changes. Ok!

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
```diff
@@ -82,7 +82,17 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
      * The minimum amount of information needed to notarise a transaction. Note that this does not include
      * any sensitive transaction details.
      */
-    protected data class TransactionParts(val id: SecureHash, val inputs: List<StateRef>, val timestamp: TimeWindow?, val notary: Party?)
+    protected data class TransactionParts @JvmOverloads constructor(
```
adagys: not public api, jvmoverloads not strictly necessary

core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
adagys: is this used anywhere? I don't think we need it at all

adagys:
Don't remember if we discussed this already, but need to think if this has any impact on notary change & contract upgrade transaction types.

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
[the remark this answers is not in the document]
roger3cev: Yes, makes more sense to put it here. See my comment below about when we want errors to occur? When states are added to the builder or when the `TransactionBuilder` is converted to a `WireTransaction`?

roger3cev:
OK perhaps we can get some time this week to discuss. Cheers

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
kchalkias: IMO, we cannot remove any checks from `WireTransaction.init` function anyway, because we should also check during deserialisation for valid txs. So anything before that is duplicated work, right? I can see some redundancy already, i.e., we check for null notary on null time-window, but this will be checked again on the WireTx constructor.
Do we need a fail-fast solution at this stage? I am not sure it will be a common case anyway + I believe this shouldn't be builder's task (as we can always evolve WireTx rules, but forget to update the builder). To my mind it seems like a duplicated effort that could also lead to inconsistency. That said, I personally prefer the WireTx constructor approach.

core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
mikehearn: Maybe move it into the flow as an inner enum class then, to avoid putting it into the top level namespace.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -65,8 +65,8 @@ class NotaryFlow {
         protected fun checkTransaction(): Party {
             val notaryParty = stx.notary ?: throw IllegalStateException("Transaction does not specify a Notary")
             check(serviceHub.networkMapCache.isNotary(notaryParty)) { "$notaryParty is not a notary on the network" }
-            check(serviceHub.loadStates(stx.inputs.toSet()).all { it.state.notary == notaryParty }) {
-                "Input states must have the same Notary"
+            check(serviceHub.loadStates(stx.inputs.toSet() + stx.references.toSet()).all { it.state.notary == notaryParty }) {
```
mikehearn: is this check happening after we ensured a state isn't both consumed and ref state?

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,61 @@
+package net.corda.core.flows
+
+import co.paralleluniverse.fibers.Suspendable
+import net.corda.core.contracts.StateRef
+
+/**
+ * Given a flow which uses [ReferenceState]s, the [WithReferencedStatesFlow] will execute the the flow as a subFlow.
+ * If the flow fails due to a [NotaryError.Conflict] for a [ReferenceState], then it will be suspended until the
+ * state refs for the [ReferenceState]s are updated.
```
mikehearn: .... and then retried? This description could be a bit longer.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        // Return the flow result or wait for the StateRefs to be updated and try again.
+        @Suppress("UNCHECKED_CAST")
+        return when (processedResult) {
+            is FlowResult.Success<*> -> processedResult.value as T  // Will always be T.
+            is FlowResult.Conflict -> {
+                waitForStatesToUpdate(processedResult.stateRefs)
+                // The flow has now woken up because the conflicting
+                // reference states have been updated.
+                subFlow(flowLogic)
```
mikehearn: This probably needs to be in a while(true) loop. You don't know it won't race with another update to the ref state and die again.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,61 @@
...
+class WithReferencedStatesFlow<T : Any>(val flowLogic: FlowLogic<T>) : FlowLogic<T>() {
+
+    private sealed class FlowResult {
+        data class Success<T : Any>(val value: T) : FlowResult()
+        data class Conflict(val stateRefs: Set<StateRef>) : FlowResult()
+    }
+
+    @Suspendable
+    override fun call(): T {
+        // Try starting the flow (as a subFlow) which uses the reference states.
```
mikehearn: This flow could use some logging, maybe progress tracking?

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
```diff
@@ -0,0 +1,28 @@
...
+ * @property services a ServiceHub instance
+ */
+class WaitForStatesToUpdate(val stateRefs: Set<StateRef>, val services: ServiceHub) : FlowAsyncOperation<Unit> {
+    override fun execute(): CordaFuture<Unit> {
+        val futures = stateRefs.map { stateRef -> services.vaultService.whenConsumed(stateRef).toCompletableFuture() }
+        // Hack alert!
+        // CompletableFuture.allOf() returns a CompletableFuture<Void!>! and then returning a CordaFuture<Void>
+        // breaks the state machine manager, so instead, we return a CordaFuture<Unit>, which works.
+        // TODO: Fix the above.
+        return CompletableFuture.allOf(*futures.toTypedArray()).thenApply { Unit }.asCordaFuture()
```
mikehearn: logging

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -146,8 +145,8 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
             for (stx in downloads)
                 check(resultQ.putIfAbsent(stx.id, stx) == null)   // Assert checks the filter at the start.
 
-            // Add all input states to the work queue.
-            val inputHashes = downloads.flatMap { it.inputs }.map { it.txhash }
+            // Add all input states and reference input states to the work queue.
```
mikehearn: We haven't validated the tx by this point, so should we be using toSet() to eliminate duplicates a bit more carefully here?

core/src/main/kotlin/net/corda/core/internal/TopologicalSort.kt
```diff
@@ -17,9 +17,10 @@ class TopologicalSort {
      * Add a transaction to the to-be-sorted set of transactions.
      */
     fun add(stx: SignedTransaction) {
-        for (input in stx.inputs) {
-            // Note that we use a LinkedHashSet here to make the traversal deterministic (as long as the input list is)
-            forwardGraph.getOrPut(input.txhash) { LinkedHashSet() }.add(stx)
+        val stateRefs = stx.references + stx.inputs
```
mikehearn: likewise: should we be assuming non-duplication here or enforcing it with toSet?

core/src/main/kotlin/net/corda/core/internal/notary/TrustedAuthorityNotaryService.kt
```diff
@@ -25,17 +25,20 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
...
             if (e.error is NotaryError.Conflict) {
-                val conflicts = inputs.filterIndexed { _, stateRef ->
+                val allInputs = inputs + references
+                val conflicts = allInputs.filterIndexed { _, stateRef ->
                     val cause = e.error.consumedStates[stateRef]
                     cause != null && cause.hashOfTransactionId != txId.sha256()
                 }
                 if (conflicts.isNotEmpty()) {
                     // TODO: Create a new UniquenessException that only contains the conflicts filtered above.
+                    // TODO: Check if this was a reference input state only conflict.
```
mikehearn: Does this TODO need to be done? I'm not sure what the impact is of not doing it - just a logging issue?

core/src/main/kotlin/net/corda/core/internal/notary/TrustedAuthorityNotaryService.kt
```diff
@@ -25,17 +25,20 @@ abstract class TrustedAuthorityNotaryService : NotaryService() {
...
-    fun commitInputStates(inputs: List<StateRef>, txId: SecureHash, caller: Party, requestSignature: NotarisationRequestSignature, timeWindow: TimeWindow?) {
+    @JvmOverloads
+    fun commitInputStates(inputs: List<StateRef>, txId: SecureHash, caller: Party, requestSignature: NotarisationRequestSignature, timeWindow: TimeWindow?, references: List<StateRef> = emptyList()) {
         try {
-            uniquenessProvider.commit(inputs, txId, caller, requestSignature, timeWindow)
+            uniquenessProvider.commit(inputs, txId, caller, requestSignature, timeWindow, references)
         } catch (e: NotaryInternalException) {
             if (e.error is NotaryError.Conflict) {
-                val conflicts = inputs.filterIndexed { _, stateRef ->
+                val allInputs = inputs + references
```
mikehearn: again, toSet? Or do we need an "assertNoOverlaps" function?

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -105,3 +105,28 @@ data class NetworkParameters(
...
 data class NotaryInfo(val identity: Party, val validating: Boolean)
+
+/**
+ * Used for restricting the use of features which require a platform version greater than the current minimum platform
+ * version for the network.
+ *
+ * @param requiredMinimumVersion the minimum platform version which this feature requires.
+ * @param block the code block which is dependent on a potentially greater minimum platform version than the network's.
+ */
+fun <T : Any> checkVersion(requiredMinimumVersion: Int, block: () -> T): T {
```
mikehearn: This name is a bit vague for a top level function. Maybe move it to the companion object in PlatformVersion? Also why is this taking a lambda? It could just be like check() - i don't see any reason why it has to be scoped?

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -105,3 +105,28 @@ data class NetworkParameters(
...
+                "increased from $minimumPlatformVersion to $requiredMinimumVersion.")
+    } else {
+        block()
+    }
+}
+
+/** For use by the [checkVersion] function. */
+data class PlatformVersionInfo(val minimumPlatformVersion: Int)
+
+/** This variable is set in [AbstractNode] during node startup. */
```
mikehearn: Maybe this stuff should be moved to the companion object in AbstractNode and/or using the thread local variable, because otherwise, it will be hard to do mock networks where two nodes have different platform versions (sorry, I realise I said before we'd probably never need that, but i was forgetting about our own platform unit tests)

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -105,3 +105,28 @@ data class NetworkParameters(
...
+    val minimumPlatformVersion = platformVersionInfo.minimumPlatformVersion
+    return if (requiredMinimumVersion > minimumPlatformVersion) {
+        throw UncheckedVersionException("This feature is disabled until network minimum platform version is " +
+                "increased from $minimumPlatformVersion to $requiredMinimumVersion.")
+    } else {
+        block()
+    }
+}
+
+/** For use by the [checkVersion] function. */
```
mikehearn: Probably can be a better doc, describing what it is actually for e.g. what does "platform version" mean specifically here - could link to the docsite for an explanation.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
mikehearn: This kdoc specifies things only relevant to us, not the user.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -105,3 +105,28 @@ data class NetworkParameters(
...
+data class PlatformVersionInfo(val minimumPlatformVersion: Int)
+
+/** This variable is set in [AbstractNode] during node startup. */
+lateinit var platformVersionInfo: PlatformVersionInfo
+
+class UncheckedVersionException(message: String) : RuntimeException(message)
```
mikehearn: public api needs kdoc.

This stuff might be better kept internal for now. I'm not sure the API / impl is quite right yet, but I don't want to block ref states.

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
```diff
@@ -15,6 +15,8 @@ import java.util.function.Predicate
 @KeepForDJVM
 @DoNotImplement
 abstract class BaseTransaction : NamedByHash {
+    /** A list of reusable reference data states which can be referred to by other contacts in this transaction. */
```
mikehearn: typo: contacts

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -145,42 +149,109 @@ open class TransactionBuilder(
...
+            "A StateRef cannot be both an input and a reference input in the same transaction."
+        }
+    }
+
+    private fun checkReferencesUseSameNotary() = referencesWithTransactionState.map { it.notary }.toSet().size == 1
+
+    /**
+     * Adds a reference input [StateRef] to the transaction.
+     *
+     * This feature was added in version 4 of Corda, so is disabled for any Corda networks with a minimum platform
```
mikehearn: Can you rephrase "is disabled" to say "will throw an exception" and add a throws tag in the kdoc?

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -145,42 +149,109 @@ open class TransactionBuilder(
...
+     * version less than 4.
+     */
+    open fun addReferenceState(referencedStateAndRef: ReferencedStateAndRef<*>): TransactionBuilder {
+        return checkVersion(requiredMinimumVersion = 4) {
+            val stateAndRef = referencedStateAndRef.stateAndRef
+            referencesWithTransactionState.add(stateAndRef.state)
+
+            // It is likely the case that users of reference states do not have permission to change the notary assigned
```
mikehearn: This is a good explanation - it could be copy/pasted into the docsite as well, in the relevant section for reference states.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -145,42 +149,109 @@ open class TransactionBuilder(
...
 
+    /** Adds an output state to the transaction. */
     fun addOutputState(state: TransactionState<*>): TransactionBuilder {
         outputs.add(state)
         return this
     }
 
     @JvmOverloads
-    fun addOutputState(state: ContractState, contract: ContractClassName, notary: Party, encumbrance: Int? = null, constraint: AttachmentConstraint = AutomaticHashConstraint): TransactionBuilder {
+    fun addOutputState(
```
mikehearn: Can you add kdocs for this whilst you're here?

core/src/test/kotlin/net/corda/core/flows/WithReferencedStatesFlowTests.kt
```diff
@@ -0,0 +1,162 @@
...
+        // 3. Update the reference state but don't share the update.
+        val updatedRefTx = nodes[0].services.startFlow(UpdateRefState(newRefState)).resultFuture.getOrThrow()
+        val updatedRefState = updatedRefTx.tx.outRefsOfType<RefState.State>().single()
+
+        // 4. Try to use the old reference state. This will throw a NotaryException.
+        val useRefTx = nodes[1].services.startFlow(WithReferencedStatesFlow(UseRefState(newRefState.state.data.linearId))).resultFuture
+
+        // 5. Give the UseRefState flow time to fail. We can't "GetOrThrow" now because then we won't be able to get
+        // the result later on.
+        Thread.sleep(500)
```
mikehearn: ick .... flaky test alert. I don't fully understand this statement - why is this racing? Perhaps somewhere you want to use the single-threaded mode of MockNetwork? ;)

docs/source/api-transactions.rst
```diff
@@ -90,6 +91,35 @@ The ``StateRef`` links an input state back to the transaction that created it. T
 "chains" linking each input back to an original issuance transaction. This allows nodes verifying the transaction
 to "walk the chain" and verify that each input was generated through a valid sequence of transactions.
 
+Reference input states
+~~~~~~~~~~~~~~~~~~~~~~
+
+A reference input state is a ``ContractState`` which can be referred to in a transaction by the contracts of input and
+output states but whose contract is not executed as part of the transaction verification process and is not consumed
+when the transaction is committed to the ledger but is checked for "current-ness". In other words, the contract logic
```
mikehearn: Run-on sentence

docs/source/api-transactions.rst
```diff
@@ -90,6 +91,35 @@ The ``StateRef`` links an input state back to the transaction that created it. T
...
+isn't run for the referencing transaction only. It's still a normal state when it occurs in an input or output position.
+
+Reference data states enable many parties to "reuse" the same state in their transactions as reference data whilst
+still allowing the reference data state owner the capability to update the state.
+
+A reference input state is added to a transaction as a ``ReferencedStateAndRef``. A ``ReferencedStateAndRef`` can be
+obtained from a ``StateAndRef`` by calling the ``StateAndRef.referenced()`` method which returns a
+``ReferencedStateAndRef``.
```
mikehearn: Can you mention the no-two-different-notaries rule and discuss here?

docs/source/changelog.rst
```diff
@@ -170,6 +170,12 @@ Unreleased
   The log entry starts with `Cross-reference between MappedSchemas.`.
   API: Persistence documentation no longer suggests mapping between different schemas.
 
+* Introduced the concept of "reference input states". A reference input state is a ``ContractState`` which can be referred
```
mikehearn: This is big enough for the release notes.

docs/source/api-transactions.rst
```diff
@@ -90,6 +91,35 @@ The ``StateRef`` links an input state back to the transaction that created it. T
 "chains" linking each input back to an original issuance transaction. This allows nodes verifying the transaction
 to "walk the chain" and verify that each input was generated through a valid sequence of transactions.
 
+Reference input states
+~~~~~~~~~~~~~~~~~~~~~~
+
+A reference input state is a ``ContractState`` which can be referred to in a transaction by the contracts of input and
```
mikehearn: add a `.. warning::` here about the fact that you can't use it unless the network you're on has a minPlatformVersion >= 4

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -126,16 +129,26 @@ class PersistentUniquenessProvider(val clock: Clock) : UniquenessProvider, Singl
...
         }
+
+        checkConflicts(states, ConsumedStateType.INPUT_STATE)
+        checkConflicts(references, ConsumedStateType.REFERENCE_INPUT_STATE)
+
         return conflictingStates
     }
 
     private fun handleConflicts(txId: SecureHash, conflictingStates: LinkedHashMap<StateRef, StateConsumptionDetails>) {
+        // TODO: Deal with old reference states here.
```
mikehearn: ?? important ??

core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
roger3cev: Not sure I understand. This enum needs to be used by the notary to tag whether the conflicting state is a reference state or a normal state. The `StateConsumptionDetails` are then used by the `WithReferencedStateFlow` to determine if a notary conflict was due to out of date reference data. It _may_ be used in other areas we haven't yet thought about? Maybe it is best kept inside the `StateConsumptionDetails` class?

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -169,9 +169,16 @@ data class StateRef(val txhash: SecureHash, val index: Int) {
...
 // DOCSTART 7
-data class StateAndRef<out T : ContractState>(val state: TransactionState<T>, val ref: StateRef)
+data class StateAndRef<out T : ContractState>(val state: TransactionState<T>, val ref: StateRef) {
+    /** For adding [StateAndRef]s as references to a [TransactionBuilder]. */
+    fun referenced() = ReferencedStateAndRef(this)
+}
 // DOCEND 7
 
+/** A wrapper for a [StateAndRef] indicating that it should be added to a transaction as a reference input state. */
+@CordaSerializable
```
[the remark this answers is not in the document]
roger3cev: Actually, you are right, it probably doesn't need to be serializable - it's only used to delineate between normal states and reference states but the class itself is not actually stored in the properties of the builder.

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -397,6 +401,9 @@ abstract class FlowLogic<out T> {
         return stateMachine.suspend(request, maySkipCheckpoint = maySkipCheckpoint)
     }
 
+    @Suspendable
+    fun waitForStatesToUpdate(stateRefs: Set<StateRef>) = executeAsync(WaitForStatesToUpdate(stateRefs, serviceHub))
```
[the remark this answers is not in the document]
roger3cev: Added.

core/src/main/kotlin/net/corda/core/flows/NotaryError.kt
```diff
@@ -65,11 +65,23 @@ sealed class NotaryError {
...
          *
          * Note that this is NOT the transaction id itself – revealing it could lead to privacy leaks.
          */
-        val hashOfTransactionId: SecureHash
-)
\ No newline at end of file
+        val hashOfTransactionId: SecureHash,
+
+        /**
+         * The type of consumed state: Either an reference input state or a regular input state.
```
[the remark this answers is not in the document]
roger3cev: Amended.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
roger3cev: Yes, there is a check in BaseTransaction` to check this. So transaction builders which consume a state and use it as a reference cannot be converted into a wire transaction.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Yes, I was going to put a TODO about this but in the end I forgot to add it. I probably need to re-think the structure of the flow such that it can deal with consecutive reference state conflicts.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Will add some logging + progress tracker.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
roger3cev: I don't understand how there can be dupes? This is checked in `TransactionBuilder` and `BaseTransaction`.

core/src/main/kotlin/net/corda/core/internal/TopologicalSort.kt
roger3cev: See above. I might be missing something though?

core/src/main/kotlin/net/corda/core/internal/notary/TrustedAuthorityNotaryService.kt
roger3cev: Ah this TODO is done now. See `ConsumedStateType` Enum.

core/src/main/kotlin/net/corda/core/internal/notary/TrustedAuthorityNotaryService.kt
roger3cev: See above

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
roger3cev: Removed the `block` parameter, changed the function name to make it more precise and moved it into the companion object.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
roger3cev: Ah yes that makes sense. I'll move it all to `AbstractNode`.

roger3cev: ThreadLocal will only work when the mocknetwork is used in `threadPerNode = true` mode though, right?

roger3cev: Amended.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
roger3cev: I can make it internal for now. Tbh, I haven't put _that_ much thought into it... Just put it together to unblock ref states.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
roger3cev: Done.

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
roger3cev: Done.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
roger3cev: Opted for a ThreadLocal with warning.

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
roger3cev: Ah. I forgot to remove this. Good spot.

docs/source/api-transactions.rst
roger3cev: Added warning

docs/source/api-transactions.rst
roger3cev: Added.

core/src/test/kotlin/net/corda/core/flows/WithReferencedStatesFlowTests.kt
roger3cev: I believe the issue is that the flow to update the state is scheduled to run before the `waitForStatesToUpdate` part of the `WithReferencedStatesFlow` flow, thus the `WithReferencedStatesFlow` suspends forever.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
mikehearn: The method name could also just be checkVersion or something, as the class name gives it context.

core/src/test/kotlin/net/corda/core/flows/WithReferencedStatesFlowTests.kt
mikehearn: We discussed this on slack - the issue is the waitForStatesToUpdate is assuming the state isn't already consumed, so it's got a race condition in it. It needs to check and suspend atomically, and return immediately if the state is already consumed at the time of call.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
mikehearn: OK, fair enough then.

core/src/main/kotlin/net/corda/core/flows/NotaryWireFormat.kt
```diff
@@ -16,6 +16,9 @@ import net.corda.core.transactions.SignedTransaction
  * While the signature must be retained, the notarisation request does not need to be transferred or stored anywhere - it
  * can be built from a [SignedTransaction] or a [CoreTransaction]. The notary can recompute it from the committed states index.
  *
+ * Reference inputs states are not included as a separate property in the [NotarisationRequest] as they are not
```
adagys: is there a situation where we would want proof that someone "queried" certain reference input states? e.g. for audit purposes.
It's definitely correct to exclude them here, just wonering if we need some additional mechanism for that.

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -397,6 +401,18 @@ abstract class FlowLogic<out T> {
...
+     * Suspends the current flow until all the provided [StateRef]s have been updated.
+     *
+     * WARNING! Remember that the flow which uses this async operation will _NOT_ wake-up until all the supplied StateRefs
+     * have been updated. If the node isn't aware of the supplied StateRefs or if the StateRefs are never updated, then
+     * the calling flow will remain suspended.
+     *
+     * @param stateRefs the StateRefs which will be updated in the future.
+     */
+    @Suspendable
+    fun waitForStatesToUpdate(stateRefs: Set<StateRef>) = executeAsync(WaitForStatesToUpdate(stateRefs, serviceHub))
```
adagys: not immediately clear what "update" here exaclty means. Could you elaborate in the comments or change it to "consume" or something, to be more explicit.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,104 @@
...
+import co.paralleluniverse.fibers.Suspendable
+import net.corda.core.contracts.StateRef
+import net.corda.core.internal.uncheckedCast
+import net.corda.core.utilities.ProgressTracker
+import net.corda.core.utilities.contextLogger
+
+/**
+ * Given a flow which uses reference states, the [WithReferencedStatesFlow] will execute the the flow as a subFlow.
+ * If the flow fails due to a [NotaryError.Conflict] for a reference state, then it will be suspended until the
+ * state refs for the reference states are updated, the node will then retry running the flow
```
adagys: same, could you elaborate on the "updated" part

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    // Return a successful flow result or a FlowException.
+    @Suspendable
+    fun runFlow(): Any = try {
+        subFlow(flowLogic)
+    } catch (e: FlowException) {
+        e
+    }
+
+    // Process the flow result. We don't care about anything other than NotaryExceptions. If it is a
```
adagys: minor: could you change method comments to /** **/ so they're visible during method inspection

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,104 @@
...
+    // NotaryException but not a Conflict, then just rethrow. It's it's a Conflict, then extract the reference
```
adagys: typo It's = If

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,104 @@
...
+        // do this because there might be consecutive update races.
+        while (true) {
+            logger.debug("Attempting to run the supplied flow. ${flowLogic.javaClass.canonicalName}")
+            val result = runFlow()
+            val processedResult = processFlowResult(result)
+
+            // Return the flow result or wait for the StateRefs to be updated and try again.
+            // 1. If a success, we can always cast the return type to T.
+            // 2. If there is a conflict, then suspend this flow, only waking it up when the conflicting reference
+            //    states have been updates.
```
adagys: typo: updates -> updated

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,104 @@
...
+            @Suppress("UNCHECKED_CAST")
+            when (processedResult) {
+                is FlowResult.Success<*> -> {
+                    logger.debug("Flow completed successfully.")
```
adagys: could include the flow type in the message

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
```diff
@@ -0,0 +1,50 @@
...
+    override fun execute(): CordaFuture<Unit> {
+        // Determine if any of the states have already been consumed.
+        val query = QueryCriteria.VaultQueryCriteria(stateRefs = stateRefs.toList())
+        val results = services.vaultService.queryBy<ContractState>(query).statesMetadata
+        val notYetConsumed = results.filter { it.consumedTime == null }.map { it.ref }
+
+        // If "notYetConsumed" is empty then all the states have already been consumed. Otherwise,
+        // if the list is non empty, then watch the vault for updates to the "notYetConsumed" StateRefs.
+        return if (notYetConsumed.isEmpty()) {
+            logger.debug("All StateRefs have already been consumed. No need to suspend.")
```
adagys: could include staterefs in the log message

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
```diff
@@ -29,16 +31,25 @@ abstract class BaseTransaction : NamedByHash {
...
 
     private fun checkNoDuplicateInputs() {
         check(inputs.size == inputs.toSet().size) { "Duplicate input states detected" }
+        check(references.size == references.toSet().size) { "Duplicate reference states detected" }
+    }
+
+    private fun checkForInputsAndReferencesOverlap() {
+        val intersection = inputs intersect references
+        require(intersection.isEmpty()) {
+            "A StateRef cannot be both an input and a reference input in the same transaction."
```
adagys: would be helpful to specify which stateref violates this (we haven't been very rigorous with this so far though..)

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,104 @@
...
+        val flowLogic: FlowLogic<T>,
+        override val progressTracker: ProgressTracker = WithReferencedStatesFlow.tracker()
+) : FlowLogic<T>() {
+
+    companion object {
+        val logger = contextLogger()
+
+        object ATTEMPT : ProgressTracker.Step("Attempting to run flow which uses reference states.")
+        object RETRYING : ProgressTracker.Step("Reference states are out of date! Waiting for updated states...")
+        object SUCCESS : ProgressTracker.Step("Flow was ran successfully.")
```
mikehearn: either "ran" or "was run"

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private sealed class FlowResult {
+        data class Success<T : Any>(val value: T) : FlowResult()
+        data class Conflict(val stateRefs: Set<StateRef>) : FlowResult()
+    }
+
+    // Return a successful flow result or a FlowException.
+    @Suspendable
+    fun runFlow(): Any = try {
```
mikehearn: Should this be private? Is a separate function needed at all?

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
```diff
@@ -0,0 +1,50 @@
+package net.corda.core.internal
```
mikehearn: Is the file name and package appropriate for this? I'd have expected it to be in a package with "flows" in the name and the file to be named after the class.

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
```diff
@@ -0,0 +1,50 @@
...
+        val notYetConsumed = results.filter { it.consumedTime == null }.map { it.ref }
+
+        // If "notYetConsumed" is empty then all the states have already been consumed. Otherwise,
+        // if the list is non empty, then watch the vault for updates to the "notYetConsumed" StateRefs.
+        return if (notYetConsumed.isEmpty()) {
+            logger.debug("All StateRefs have already been consumed. No need to suspend.")
+            doneFuture(Unit)
+        } else {
+            logger.debug("Suspending flow. Flow will wake up when the supplied StateRefs are consumed: $stateRefs.")
+            val futures = notYetConsumed.map { services.vaultService.whenConsumed(it).toCompletableFuture() }
```
mikehearn: I suspect this still isn't quite right unfortunately. The check for already being consumed should be inside whenConsumed in the vault. Otherwise you have a race here - the state could be consumed between lines 34 and 44 i.e. after the query but before the call to whenConsumed(it). To get this sort of parallel code right the check has to be atomic with anything that could cause an update.

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
mikehearn: 👍

docs/source/api-transactions.rst
```diff
@@ -90,6 +91,53 @@ The ``StateRef`` links an input state back to the transaction that created it. T
...
+state when it occurs in an input or output position.
+
+Reference data states enable many parties to "reuse" the same state in their transactions as reference data whilst
+still allowing the reference data state owner the capability to update the state.
+
+A reference input state is added to a transaction as a ``ReferencedStateAndRef``. A ``ReferencedStateAndRef`` can be
+obtained from a ``StateAndRef`` by calling the ``StateAndRef.referenced()`` method which returns a
+``ReferencedStateAndRef``.
+
+.. warning:: Reference states are only available on Corda networks with a minimump platform version >= 4.
```
mikehearn: typo: minimum

docs/source/api-transactions.rst
```diff
@@ -90,6 +91,53 @@ The ``StateRef`` links an input state back to the transaction that created it. T
...
+
+.. container:: codeset
+
+    .. literalinclude:: ../../docs/source/example-code/src/main/kotlin/net/corda/docs/FlowCookbook.kt
+:language: kotlin
```
mikehearn: Is the indent right here - do the docs render properly?

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
roger3cev: Doh. You are right. Question is, how can I break whenConsumed without breaking backwards compatibility? Perhaps I can add a new method?

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Updated.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Updated. Removed function.

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
roger3cev: I called it AsyncOperations because we might wish to add more of them here. I guess it's a premature optimisation.. I'll make the suggest changes.

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
roger3cev: OK, this is fixed now but there is still a bug in `VaultService._trackBy` which needs fixing - see Rick's JIRA.

docs/source/api-transactions.rst
roger3cev: Amended.

docs/source/api-transactions.rst
roger3cev: Amended.

core/src/main/kotlin/net/corda/core/internal/AsyncOperations.kt
roger3cev: Updated with better logging.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Amended.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Amended.

core/src/main/kotlin/net/corda/core/transactions/BaseTransaction.kt
roger3cev: Updated.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Amended.

core/src/main/kotlin/net/corda/core/flows/WithReferencedStatesFlow.kt
roger3cev: Amended.

kchalkias:
Great job @roger3cev . We should all agree if we can tolerate referencing ownable states. Well, ownership is a complex topic, i.e., one can simulate it with encumbrance. Also, I don't see how we can distinguish between an ownable and non-ownable state in the case of non validating notaries, except if we change the output state format and we always make visible if a state is ownable (which again leaks some info).

adagys:
@mikehearn my worry was more that the transaction might be unintended, rather than invalid. Say the reference state specifies the price for something and my flow creates a transaction that buys X assets for that price. If the price increases I may no longer want to buy those assets.

PR #3552, opened 2018-07-10
ENT-1968: Add RPC for refreshing network map cache

kasiastreich:

### DIFF
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -368,8 +368,20 @@ interface CordaRPCOps : RPCOps {
      */
     fun nodeInfoFromParty(party: AbstractParty): NodeInfo?
 
-    /** Clear all network map data from local node cache. */
+    /**
+     * Clear all network map data from local node cache. Notice that after invoking this method your node will lose
+     * network map data and effectively won't be able to start any flow with the peers until network map is downloaded
+     * again on next poll - when it happens, depends on polling interval set by network map server. You can also use [refreshNetworkMapCache]
+     * to force next fetch from network map server.
+     */
     fun clearNetworkMapCache()
+
+    /**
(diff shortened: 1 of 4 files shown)


node/src/main/kotlin/net/corda/node/internal/rpc/proxies/ExceptionSerialisingRpcOpsProxy.kt
```diff
@@ -80,7 +78,11 @@ internal class ExceptionSerialisingRpcOpsProxy(private val delegate: CordaRPCOps
 
         private fun ensureSerialisable(error: Throwable): Throwable {
             val serialisable = (superclasses(error::class.java) + error::class.java).any { it.isAnnotationPresent(CordaSerializable::class.java) || it.interfaces.any { it.isAnnotationPresent(CordaSerializable::class.java) } }
-            val result = if (serialisable) error else CordaRuntimeException(error.message, error)
+            val result = if (serialisable) error else {
```
shamsasari: braces for the if part as well

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -368,9 +368,21 @@ interface CordaRPCOps : RPCOps {
...
+    /**
+     * Clear all network map data from local node cache. Notice that after invoking this method your node will lose
+     * network map data and effectively won't be able to start any flow with the peers until network map is downloaded
+     * again on next poll - when it happens, depends on polling interval set by network map server. You can also use [refreshNetworkMapCache]
+     * to force next fetch from network map server.
+     */
     fun clearNetworkMapCache()
 
+    /**
+     * Poll network map server if available for the network map. Notice that you need to have [compatibilityZone]
```
shamsasari: `[compatibilityZone]` and `[networkServices]` are references to the node config? The `[]` notation is only used for references to code so I imagine these are coming as warnings.

PR #3556, opened 2018-07-10
CORDA-1748: Delete unwanted default parameter values from Kotlin metadata.

chrisr3:
Kotlin generates special synthetic constructors and functions to support any default parameter values. So ensure that we also update the Kotlin metadata accordingly, should these synthetic methods happen to disappear (e.g. ProGuard).


chrisr3:
Once upon a time, I believed that tweaking the Kotlin metadata would be a simple "nice to have" feature. Now I realise that it's an _essential_ aspect of running ProGuard over a Kotlin library object :disappointed: .

PR #3557, opened 2018-07-11
ENT-1565 Upgrade Artemis version to latest.

rick-r3:
Upgrading as soon as possible after CE 3.0 GA to maximise regression finding time.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ buildscript {
     ext.capsule_version = '1.0.1'
 
     ext.asm_version = '5.0.4'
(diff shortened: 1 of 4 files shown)


chrisr3:
Can we replace the `ImmutableSet` in `CordaX500Name` with:
```kotlin
private val countryCodes: Set<String> = setOf(*Locale.getISOCountries(), unspecifiedCountry)
```
?

mnesbit:
I am fine with that it if RPC smokeTest works when sending a Party

PR #3559, opened 2018-07-11
CORDA-1664: Blob inspector able to  display SignedTransaction blobs dumped from a node's db

shamsasari:


tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -78,39 +82,44 @@ class Main : Runnable {
...
+                p2pContext = AMQP_P2P_CONTEXT.withLenientCarpenter(),
+                storageContext = AMQP_STORAGE_CONTEXT.withLenientCarpenter()
         ))
     }
 }
 
 private object AMQPInspectorSerializationScheme : AbstractAMQPSerializationScheme(emptyList()) {
     override fun canDeserializeVersion(magic: CordaSerializationMagic, target: SerializationContext.UseCase): Boolean {
-        return magic == amqpMagic && target == SerializationContext.UseCase.P2P
+        return magic == amqpMagic
```
fenryka: technically this only supports STORAGE and P2P

PR #3561, opened 2018-07-11
CORDA-1543 create test for flow checkpoint simple

szymonsztuka:
Integration test for https://r3-cev.atlassian.net/browse/CORDA-1477.
This is minimal test for change JAR file name contains a cordapp, without any alterations to driverDSL.
The check for hash of the file content needs some DriverDSL addition (the next PR).
The node starts a flow which is suspended after initial progress.
The node is restarted, due to driverDSL and CodappLoader for test packages the flow class is loaded from different JAR file (random UUID is added to the file name).
Because the node can't match the restarted flow with the old JAR name, the  node throws the error.
The actual check expects a log entry with `Found checkpoint for flow: <flowClass> that is no longer installed.....`


szymonsztuka:
Moved to other PR

PR #3569, opened 2018-07-12
ENT-1565: Remove Guava code from core.

chrisr3:
Remove references to Guava code from `core` so that modules that use it are free to use whichever version of Guava they like.


chrisr3:
The RPC smoke tests do pass with this change.

PR #3586, opened 2018-07-13
CORDA-1800: Shrink the Simm Valuation Demo's contract states using ProGuard.

chrisr3:
The `contracts-states` CorDapp is so large that it breaks Corda's 10 MB transaction limit. This is because OpenGamma has a lot of transitive dependencies. For now, use ProGuard to discard all of the classes that we are not using.

We cannot be as aggressive with ProGuard as we'd like, partly for licensing reasons but also because the `flows` CorDapp uses the same classpath and so may try to use these OpenGamma classes too.

### DIFF
--- a/samples/simm-valuation-demo/build.gradle
+++ b/samples/simm-valuation-demo/build.gradle
@@ -29,7 +29,7 @@ dependencies {
 
     // The SIMM demo CorDapp depends upon Cash CorDapp features
     cordapp project(':finance')
-    cordapp project(':samples:simm-valuation-demo:contracts-states')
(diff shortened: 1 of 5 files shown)


chrisr3:
Note that the `cordapp` plugin would still include `slf4j-api`, despite it being declared as `cordaCompile`, because it is a transitive dependency of OpenGamma.

samples/simm-valuation-demo/build.gradle
```diff
@@ -1,5 +1,5 @@
-buildscript {
-    ext.strata_version = '1.1.2'
+ext {
```
Clintonio: Could you do subprojects { } around this instead of having them reference it in their own build.gradles?

PR #3595, opened 2018-07-13
[CORDA-1635] Update docs (node naming constraints).

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1635

### DIFF
--- a/core/src/test/kotlin/net/corda/core/identity/CordaX500NameTest.kt
+++ b/core/src/test/kotlin/net/corda/core/identity/CordaX500NameTest.kt
@@ -75,7 +75,7 @@ class CordaX500NameTest {
             CordaX500Name.parse("O=Bཛྷa, L=New York, C=DE, OU=Org Unit, CN=Service Name")
         }
         // doesn't throw
-        checkLocalityAndOrganisationalUnitAndCommonName("Bཛྷa")
+        validateLocalityAndOrganisationalUnitAndCommonName("Bཛྷa")
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


tlil:
Added @mkit to review as well.

As an aside, the ticket says:

>This works fine for other X500 names, even using extended UTF characters - need to isolate which ones cause issues.
>
>Action here is to check the documentation is clear that X500 names only contain latin chars

Did you get clarity on the above? If so, can information on that be included?

docs/source/node-structure.rst
```diff
@@ -71,9 +71,21 @@ The name must also obey the following constraints:
 
 * The ``country`` attribute is a valid ISO 3166-1 two letter code in upper-case
 
-* The ``organisation`` field of the name does not allow double-spacing.
+* The ``organisation`` field of the name obeys the following constraints:
 
-.. note:: The network operator of a Corda network may put additional constraints on node naming in place.
+    * Upper-case first letter
```
mkit: I don't think that Upper-case first latter rule is applied. It is only in the FULL mode that we don't use at the moment.

docs/source/node-structure.rst
```diff
@@ -71,9 +71,21 @@ The name must also obey the following constraints:
...
+    * Has at least two letters
+    * No leading or trailing whitespace
+    * Does not include the following characters: ``,`` , ``=`` , ``$`` , ``"`` , ``'`` , ``\``
+    * Is in NFKC normalization form
+    * Does not contain the null character
+    * Only the latin, common and inherited unicode scripts are supported
```
mkit: Unicode is only in the FULL mode, which is not used.

docs/source/node-structure.rst
```diff
@@ -71,7 +71,7 @@ The name must also obey the following constraints:
 
 * The ``country`` attribute is a valid ISO 3166-1 two letter code in upper-case
 
-* All attributes must obey the following constraints:
+* The ``organisation`` field of the name obeys the following constraints:
 
     * Upper-case first letter
```
mkit: This is in the FULL mode only - not used.

docs/source/node-structure.rst
florian-f: True.

docs/source/node-structure.rst
florian-f: `CordaX500Name.parse("O=Bཛྷa, L=New York, C=DE, OU=Org Unit, CN=Service Name")`
`

florian-f: `CordaX500Name.parse("O=Bཛྷa, L=New York, C=DE, OU=Org Unit, CN=Service Name")` will not work. That's because of the normalisation rule, which is part of MINIMAL mode.

florian-f:
> Could you also test conditions for:
> No leading or trailing whitespace
> No double spacing

Leading or trailing whitespace is OK for all attributes.
Double spacing is allowed except for the organisation attribute.

core/src/test/kotlin/net/corda/core/identity/CordaX500NameTest.kt
```diff
@@ -61,16 +63,127 @@ class CordaX500NameTest {
...
 
     @Test
-    fun `rejects name with unsupported attribute`() {
+    fun `rejects organisation (but not other attributes) with non-latin letters`() {
         assertFailsWith(IllegalArgumentException::class) {
-            CordaX500Name.parse("O=Bank A, L=New York, C=US, SN=blah")
+            CordaX500Name.parse("O=Bཛྷa, L=New York, C=DE, OU=Org Unit, CN=Service Name")
         }
+        // doesn't throw
+        checkLocalityAndOrganisationalUnitAndCommonName("Bཛྷa")
```
mkit: Can you change it to `validateLocalityAndOrganisationalUnitAndCommonName`?

PR #3596, opened 2018-07-13
Fixed out-of-date integration testing tutorial and added Java example

shamsasari:


docs/source/tutorial-integration-testing.rst
```diff
@@ -1,119 +1,128 @@
...
+Next we want Bob to send this cash back to Alice.
 
-.. sourcecode:: bash
+That's it! We saw how to start up several corda nodes locally, how to connect to them, and how to test some simple invariants
+about ``CashIssueAndPaymentFlow`` and ``CashPaymentFlow``.
 
-     # Run example-code integration tests
-     ./gradlew docs/source/example-code:integrationTest -i
\ No newline at end of file
+You can find the complete test at ``example-code/src/integration-test/java/net/corda/docs/JavaIntegrationTestingTutorial.java``
```
joeldudleyr3: Link to the Corda repo here. People reading the docs won't necessarily have the repo.

docs/source/tutorial-integration-testing.rst
```diff
@@ -1,119 +1,128 @@
...
+for the Java version and ``example-code/src/integration-test/kotlin/net/corda/docs/KotlinIntegrationTestingTutorial.kt``
+for the Kotlin one..
```
joeldudleyr3: Ditto.

docs/source/tutorial-integration-testing.rst
shamsasari: I thought about that and it's something that I'd prefer we do but the problem is ensuring that we docs stay in-sync with the code since we'd have to reference the master version while the docs could be for a release version.

joeldudleyr3: I think just say: `You can find the complete test at example-code/src/integration-test/java/net/corda/docs/JavaIntegrationTestingTutorial.java of the Corda repo (link)`.

shamsasari: But the link would be for master right? What if these docs are for a release version and it just so happens that the release and master versions of this class differ?

joeldudleyr3: No, just to https://github.com/corda/corda

shamsasari: Cool

shamsasari:
Comments addressed.

PR #3601, opened 2018-07-13
Rog test tweaks

roger3cev:
Tweaked some of the tests to use static instances of `MockServices` and `MockNetwork`. There's still more to come.

Added a function to purge the DB between tests.

### DIFF
--- a/testing/test-utils/src/main/kotlin/net/corda/testing/DatabaseCleaner.kt
+++ b/testing/test-utils/src/main/kotlin/net/corda/testing/DatabaseCleaner.kt
@@ -1,0 +1,72 @@
+package net.corda.testing
+
+import net.corda.nodeapi.internal.persistence.CordaPersistence
+import java.util.*
+
+/**
+ * Cleans all the vault + transaction tables so we can re-use the same database for multiple tests.
+ */
+fun cleanDatabase(database: CordaPersistence) {
+    database.transaction {
+        val session = database.createSession()
+        val statement = session.createStatement()
+
+        statement.execute("SET REFERENTIAL_INTEGRITY FALSE")
+
+        val availableTables = HashSet<String>()
+
+        val repeats = generateSequence { "?" }.take(tablesToNuke.size).joinToString(",")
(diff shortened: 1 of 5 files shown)


testing/test-utils/src/main/kotlin/net/corda/testing/DatabaseCleaner.kt
```diff
@@ -0,0 +1,72 @@
+package net.corda.testing
+
+import net.corda.nodeapi.internal.persistence.CordaPersistence
+import java.util.*
+
+/**
+ * Cleans all the vault + transaction tables so we can re-use the same database for multiple tests.
+ */
+fun cleanDatabase(database: CordaPersistence) {
```
sollecitom: I appreciate the effort, but also would like to know how much time we're shedding by this. The reason I ask is that, while faster tests are very desirable, this change makes tests quite more complicated.

Ideally, we should have closures passed from the constructor of a `Node` instead of using Hibernate wildly, and then pass stubs from tests.

I'm happy to merge this, just saying that in general I'd prefer the other approach, as discussed face to face.

roger3cev: I don't disagree with you but doing this properly will take a lot of time and effort!

These changes to cash tests saves 25 seconds. For the other tests it saves about 5 seconds each.

testing/test-utils/src/main/kotlin/net/corda/testing/DatabaseCleaner.kt
```diff
@@ -0,0 +1,72 @@
...
+
+        availableTables.forEach {
+            statement.executeUpdate("TRUNCATE TABLE $it;")
+        }
+
+        statement.execute("SET REFERENTIAL_INTEGRITY TRUE")
+    }
+}
+
+val tablesToNuke = setOf(
```
mikehearn: can be private i guess?

mikehearn:
Closing as it needs a rebase.

PR #3615, opened 2018-07-16
Reuse mock network, randomising party names to avoid clash

distributedleetravis:
Modify a test which uses `MockNetwork` so that a) the network is re-used between tests, and b) tests use randomised party names, so that they do not conflict or pollute each other's state.

Added utility functions (grouped into _generators_, _operations_ and _matchers_) to clarify test flow and intention. [Hamkrest](https://github.com/npryce/hamkrest) is used to defined re-usable, composable matchers.

### DIFF
--- a/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
+++ b/core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
@@ -3,6 +3,7 @@ package net.corda.core.flows
 import co.paralleluniverse.fibers.Suspendable
 import net.corda.core.contracts.Attachment
 import net.corda.core.crypto.SecureHash
+import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.internal.FetchAttachmentsFlow
 import net.corda.core.internal.FetchDataFlow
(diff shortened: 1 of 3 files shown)


core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
```diff
@@ -48,10 +45,14 @@ class AttachmentTests {
         return bs.toByteArray()
     }
 
+    private fun createAlice() = mockNet.createPartyNode(randomiseName(ALICE_NAME))
+    private fun createBob() = mockNet.createPartyNode(randomiseName(BOB_NAME))
+    private fun randomiseName(name: CordaX500Name) = name.copy(commonName = "${name.commonName}_${UUID.randomUUID()}")
```
mikehearn: This helper looks generically useful, perhaps we should add it to the test api as an extension?

distributedleetravis: If this approach is accepted, I would look to extract common functionality for re-use across similar tests.

mikehearn: Or possibly MockNetwork should take care of generating names for you, from a randomised stream of names.

core/src/test/kotlin/net/corda/core/flows/AttachmentTests.kt
```diff
@@ -145,4 +111,69 @@ class AttachmentTests {
         @Suspendable
         override fun call() = subFlow(TestDataVendingFlow(otherSideSession))
     }
+
+    //region Generators
+    private fun makeNode(name: CordaX500Name) =
+        mockNet.createPartyNode(randomiseName(name)).apply {
+            registerInitiatedFlow(FetchAttachmentsResponse::class.java)
```
roger3cev: Not sure we still need to register flow initiators - I think it is done automagically!

distributedleetravis: One of the three tests fails if you don't, interestingly enough.

PR #3621, opened 2018-07-16
[CORDA-1572] Create marker file to track initial registration status.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1572

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -230,7 +230,7 @@ open class NodeStartup(val args: Array<String>) {
     private fun deleteNodeRegistrationMarker(baseDir: Path) {
         try {
             val marker = File((baseDir / INITIAL_REGISTRATION_MARKER).toUri())
-            if (!marker.exists()) {
+            if (marker.exists()) {
                 marker.delete()
             }
         } catch (e: Exception) {
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -5,6 +5,8 @@ import com.typesafe.config.Config
 import com.typesafe.config.ConfigException
 import com.typesafe.config.ConfigRenderOptions
 import io.netty.channel.unix.Errors
+import joptsimple.OptionParser
+import joptsimple.util.PathConverter
 import net.corda.core.crypto.Crypto
 import net.corda.core.internal.Emoji
 import net.corda.core.internal.concurrent.thenMatch
@@ -69,21 +71,17 @@ open class NodeStartup(val args: Array<String>) {
             return false
         }
 
+        val registrationMode = checkRegistrationMode()
         val cmdlineOptions: CmdLineOptions
-
-        // Parse the command line args just to get the base directory.
-        if (nodeRegistrationMarkerExists(NodeArgsParser().parseOrExit(*args).baseDirectory)) {
-            // If the marker file exists, that means the node was started before with `--initial-registration`,
-            // but the registration has not completed. So we pretend that the node was started with `--initial-registration`
+        if (registrationMode && !args.contains("--initial-registration")) {
+            // The node was started before with `--initial-registration`, but the registration has not completed.
+            // So we pretend that the node was started with `--initial-registration`
             logger.info("Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.")
             cmdlineOptions = NodeArgsParser().parseOrExit(*args.plus("--initial-registration"))
         } else {
             cmdlineOptions = NodeArgsParser().parseOrExit(*args)
         }
-        // If the node was started with `--initial-registration`, create marker file.
-        if (cmdlineOptions.nodeRegistrationOption != null) {
-            createNodeRegistrationMarker(cmdlineOptions.baseDirectory)
-        }
+
         // We do the single node check before we initialise logging so that in case of a double-node start it
         // doesn't mess with the running node's logs.
         enforceSingleNodeIsRunning(cmdlineOptions.baseDirectory)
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -65,13 +70,23 @@ open class NodeStartup(val args: Array<String>) {
             println("Corda will now exit...")
             return false
         }
-        val cmdlineOptions = NodeArgsParser().parseOrExit(*args)
+
+        val registrationMode = checkRegistrationMode()
+        val cmdlineOptions: CmdLineOptions
+        if (registrationMode && !args.contains("--initial-registration")) {
```
kasiastreich: you can do something like `val mdLineOptions = if (blah) NodeArgsParser... else NodeArgsParser...`

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -65,13 +70,23 @@ open class NodeStartup(val args: Array<String>) {
...
+            // The node was started before with `--initial-registration`, but the registration has not completed.
+            // So we pretend that the node was started with `--initial-registration`
+            logger.info("Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.")
```
kasiastreich: Shouldn't it be printed to console as well?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -184,6 +201,59 @@ open class NodeStartup(val args: Array<String>) {
...
+        // We do this here to ensure the marker is created even if parsing the args with NodeArgsParser fails.
+        if (optionSet.has(isRegistrationArg)) {
+            createNodeRegistrationMarker(baseDirectory)
+        }
+        return nodeRegistrationMarkerExists(baseDirectory)
+    }
+
+    private fun deleteNodeRegistrationMarker(baseDir: Path) {
+        try {
+            val marker = File((baseDir / INITIAL_REGISTRATION_MARKER).toUri())
```
kasiastreich: Here and in `createNodeRegistrationMarker` you use `nodeRegistrationMarkerExist` logic, but you don't use that function.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -184,6 +201,59 @@ open class NodeStartup(val args: Array<String>) {
...
+    private fun checkRegistrationMode(): Boolean {
+        // Parse the command line args just to get the base directory. The base directory is needed to determine
+        // if the node registration marker file exists, _before_ we call NodeArgsParser.parse().
+        // If it does exist, we call NodeArgsParser with `--initial-registration` added to the argument list. This way
+        // we make sure that the initial registration is completed, even if the node was restarted before the first
+        // attempt to register succeeded and the node administrator forgets to specify `--initial-registration` upon
+        // restart.
+        val optionParser = OptionParser()
+        optionParser.allowsUnrecognizedOptions()
+        val baseDirectoryArg = optionParser
```
kasiastreich: I think but you can correct me, that we have two places where baseDirectory comes from. Primarly from config file (!) not command line flag.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -65,13 +70,23 @@ open class NodeStartup(val args: Array<String>) {
             println("Corda will now exit...")
             return false
         }
-        val cmdlineOptions = NodeArgsParser().parseOrExit(*args)
+
+        val registrationMode = checkRegistrationMode()
+        val cmdlineOptions: CmdLineOptions
+        if (registrationMode && !args.contains("--initial-registration")) {
+            // The node was started before with `--initial-registration`, but the registration has not completed.
+            // So we pretend that the node was started with `--initial-registration`
```
kasiastreich: If you forget initial-registration it adds it for you. However with initial registration you need network trustststore path and password, they kind of go together. Saving password is a no. Could you make sure that when node spits out massive error about option parsing failure, you print out instead some information what happened and how to deal with it?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
florian-f: `--baseDirectory` It has a default value (`.`), which is specified in `NodeArgsParser`. The same is used here. The docs don't mention `baseDirectory` as an entry in the config file, only that 

>  `baseDirectory` is available as a substitution value and contains the absolute path to the node’s base directory.

https://docs.corda.net/head/corda-configuration-file.html#fields

However, that does not mean that it's impossible to specify `baseDirectory` in the config file. I will investigate what happens in that case.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
florian-f: Good point, thanks.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
florian-f: Yes, you are right. Updating this.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
florian-f: Currently, it does not spit out a _massive_ error, but it complains that the password option is missing and prints the help message, just like when you would run `java -jar corda.jar --initial-registration`. There's not indication as to _why_ this is required though, but once I address your other comment (about printing the log statement to the console as well) there will be.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
florian-f: True, can be simplified.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
florian-f: OK I checked, putting `baseDirectory="/home/some/directory"` in the `node.conf` does not have any effect on `cmdlineOptions.baseDirectory`.

PR #3622, opened 2018-07-16
[CORDA-1482] Make boolean config variables case insensitive

anthonykeenan:
The current parsing method expects boolean to be either `true` or `false`, I've overriden the function so it doesn't care about the case.


node-api/src/main/kotlin/net/corda/nodeapi/internal/config/ConfigUtilities.kt
```diff
@@ -276,6 +276,15 @@ private fun Iterable<*>.toConfigIterable(field: Field): Iterable<Any?> {
     }
 }
 
+// The typesafe .getBoolean function is case sensitive, this is a case insensitve version
```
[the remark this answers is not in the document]
anthonykeenan: :+1:

node-api/src/test/kotlin/net/corda/nodeapi/internal/config/ConfigParsingTest.kt
```diff
@@ -116,6 +116,22 @@ class ConfigParsingTest {
         assertThat(data.toConfig()).isEqualTo(config)
     }
 
+
+    @Test
+    fun `different boolean casings`() {
```
shamsasari: Better to simply update the existing `Boolean` test or merge the two.

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/ConfigUtilities.kt
```diff
@@ -276,6 +276,15 @@ private fun Iterable<*>.toConfigIterable(field: Field): Iterable<Any?> {
...
+private fun Config.getBooleanCaseInsensitive(path: String): Boolean {
```
shamsasari: I think it would be safer to try `getBoolean` first and then try again with `toLowerCase`

PR #3625, opened 2018-07-16
CORDA-1747 fix issue around RPC return of generic objects

roastario:


serialization/src/main/kotlin/net/corda/serialization/internal/amqp/CollectionSerializer.kt
```diff
@@ -80,7 +80,7 @@ class CollectionSerializer(private val declaredType: ParameterizedType, factory:
         data.withDescribed(typeNotation.descriptor) {
             withList {
                 for (entry in obj as Collection<*>) {
-                    output.writeObjectOrNull(entry, this, declaredType.actualTypeArguments[0], context, debugIndent)
+                    output.writeObjectOrNull(entry, this, resolveTypeVariables(declaredType.actualTypeArguments[0], null), context, debugIndent)
```
fenryka: we shouldn't be doing this every time we serialize an instance of the object

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/CollectionSerializer.kt
```diff
@@ -63,9 +63,13 @@ class CollectionSerializer(private val declaredType: ParameterizedType, factory:
 
     private val typeNotation: TypeNotation = RestrictedType(SerializerFactory.nameForType(declaredType), null, emptyList(), "list", Descriptor(typeDescriptor), emptyList())
 
+    private val outboundType = resolveTypeVariables(declaredType.actualTypeArguments[0], null)
```
fenryka: I feel at this point we should add some comments here because in 6-12 months we're going to come back to this abd be all WTF!!!! :D

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/MapSerializer.kt
```diff
@@ -70,10 +70,15 @@ class MapSerializer(private val declaredType: ParameterizedType, factory: Serial
 
     private val typeNotation: TypeNotation = RestrictedType(SerializerFactory.nameForType(declaredType), null, emptyList(), "map", Descriptor(typeDescriptor), emptyList())
 
+    private val inboundKeyType = declaredType.actualTypeArguments[0]
```
fenryka: same comment as above, lets comment why we need to do this

PR #3630, opened 2018-07-17
Deprecated freeLocalHostAndPort, freePort and getFreeLocalPorts of TestUtils

shamsasari:
They're prone to cause flaky tests due to the "allocated" ports already being taken by the system when eventually needed. Replaced usages with `PortAllocation.Incremental`.


shamsasari:
Not in OS, maybe in ENT.

node/src/test/kotlin/net/corda/node/services/messaging/ArtemisMessagingTest.kt
```diff
@@ -51,7 +52,8 @@ class ArtemisMessagingTest {
     @JvmField
     val temporaryFolder = TemporaryFolder()
 
-    private val serverPort = freePort()
+    private val portAllocation = PortAllocation.Incremental(10000)
```
adagys: would this cause problems when running tests in parallel?

shamsasari: Yes it will cause problems with other unit test classes that use `PortAllocation.Incremental`. I'll assign each one with a different range.

adagys: Hmm, is this a sustainable approach tough, someone could easily forget that and in general it's a bit bothersome looking for a range that's not used yet

shamsasari: That's true. The only alternative I can think of is making these integration tests, which one could argue they are since they're relying on network ports.

shamsasari: WDYT?

mnesbit: I agree that they are really integration tests. In terms of port allocations with parallel runs, the only reliable way would be to have some sort of IPC to negotiate free ports, which is way to complex.

shamsasari: OK, I'll make them integration tests.

With regards to parallel integration test runs, I have a solution using file locks, but as usual works locally but not on TC!

shamsasari:
Those unit tests are now integration tests. @mnesbit I need your re-approval.

PR #3643, opened 2018-07-18
Only build deterministic modules when using JDK8.

chrisr3:
Corda's deterministic Gradle modules assume that the underlying JDK contains these JARs:
- `$JAVA_HOME/lib/rt.jar`
- `$JAVA_HOME/lib/jce.jar`

These modules have never been built on anything other that JDK8, and the JDK layout has changed as of JDK9. So only build them for JDK8 (for now).


joeldudleyr3:
I don't feel qualified to review this 🙂 @exFalso ?

chrisr3:
Heh heh, "it's only Gradle!" :smiling_imp:

PR #3644, opened 2018-07-18
Flow test rationalisation

distributedleetravis:
This PR extends my [previous changes](https://github.com/corda/corda/pull/3615) to one of the flow test classes across the other classes in the same package, extracting common functionality into shared interfaces which are used as mixins to the test class.

Each of the test classes now shares a mock network across all of its test cases in a single class. Sharing a single mock network across all of the test classes is probably possible, given a JUnit `@Rule` or similar.

The mixin approach is taken because many of the extension functions provided depend on state local to the test class (particularly, its running `mockNet`), and so cannot simply be imported.


core/src/test/kotlin/net/corda/core/flows/mixins/WithContracts.kt
```diff
@@ -0,0 +1,97 @@
...
+import net.corda.node.internal.StartedNode
+import net.corda.testing.contracts.DummyContract
+import kotlin.reflect.KClass
+
+/**
+ * Mix this interface into a test class to get useful generator and operation functions for working with dummy contracts
+ */
+interface WithContracts : WithMockNet {
+
+    val magicNumber: Int
```
sollecitom: Not sure this is right. Apart from the name (now, I'm not sure what it's used for, but I bet is has a meaning), why don't you modify `signDummyContract` to take a trailing parameter of type `Int` with an appropriate name and a default value to a constant specified in this interface?

As in, if the test mixing this trait in is interested, they can pass a value, otherwise it's not needed?

core/src/test/kotlin/net/corda/core/flows/mixins/WithContracts.kt
```diff
@@ -0,0 +1,97 @@
...
+    //region
+
+    //region Operations
+    fun StartedNode<*>.createConfidentialIdentity(party: Party) = database.transaction {
+        services.keyManagementService.freshKeyAndCert(
+                services.myInfo.legalIdentitiesAndCerts.single { it.name == party.name },
+                false)
+    }
+
+    fun StartedNode<*>.verifyAndRegister(identity: PartyAndCertificate) = database.transaction {
```
sollecitom: These might be useful, but not sure they have much to do with `contracts`.

core/src/test/kotlin/net/corda/core/flows/mixins/WithFinality.kt
```diff
@@ -0,0 +1,36 @@
...
+import net.corda.node.internal.StartedNode
+import net.corda.testing.core.singleIdentity
+
+interface WithFinality : WithMockNet {
+
+    //region Operations
+    fun StartedNode<*>.finalise(stx: SignedTransaction, vararg additionalParties: Party) =
+        startFlowAndRunNetwork(FinalityFlow(stx, additionalParties.toSet()))
+
+    fun StartedNode<*>.getValidatedTransaction(stx: SignedTransaction) = database.transaction {
```
sollecitom: Do we need transaction blocks here?

core/src/test/kotlin/net/corda/core/flows/mixins/WithFinality.kt
```diff
@@ -0,0 +1,36 @@
...
+        services.validatedTransactions.getTransaction(stx.id)!!
+    }
+
+    fun CordaRPCOps.finalise(stx: SignedTransaction, vararg parties: Party) =
+        startFlow(ContractUpgradeFlowTest::FinalityInvoker, stx, parties.toSet())
+            .andRunNetwork()
+    //endregion
+
+    //region Matchers
+    fun transactionVisibleTo(other: StartedNode<*>) = object : Matcher<SignedTransaction> {
```
sollecitom: Rename to `visibleTo`? `assertThat(transaction, is(visibleTo(otherNode)))` sounds better than `assertThat(transaction, is(transactionVisibleTo(otherNode)))`

sollecitom: If a globally visible function doesn't cut it, perhaps create a matcher class with a static factory method.

core/src/test/kotlin/net/corda/core/flows/mixins/WithMockNet.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Randomise a party name to avoid clashes with other tests
+     */
+    fun randomise(name: CordaX500Name) = name.copy(commonName = "${name.commonName}_${UUID.randomUUID()}")
+
+    /**
+     * Run the mock network before proceeding
+     */
+    fun <T: Any> T.andRunNetwork(): T {
```
sollecitom: Can use `apply` (even if not necessarily better)

core/src/test/kotlin/net/corda/core/flows/mixins/WithMockNet.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    //region Matchers
+    /**
+     * The transaction has the required number of verified signatures
+     */
+    fun requiredSignatures(count: Int = 1) = object : Matcher<SignedTransaction> {
+        override val description: String = "A transaction with valid required signatures"
+
+        override fun invoke(actual: SignedTransaction): MatchResult = try {
+            actual.verifyRequiredSignatures()
```
sollecitom: What's the point of calling `verifyRequiredSignatures()` and following with a check on the size of the signatures collection? I mean, I'd hope `verifyRequiredSignatures()` checks that all required signatures are there right? Is it to check that no additional unrelated signatures appear?

sollecitom:
Given that upgrading to `jUnit 5` is essentially a change in the Gradle files and not much else, why not give it a go? You might find putting `@Before` and `@After` style annotations in the mixin easier.

core/src/test/kotlin/net/corda/core/flows/matchers/FlowMatchers.kt
```diff
@@ -8,11 +8,24 @@ import net.corda.core.utilities.getOrThrow
...
+        MatchResult.Match
+    } catch (e: Exception) {
+        MatchResult.Mismatch("Failed with $e")
+    }
+}
+
+/**
+ * Matches a Flow that succeeds with a result matched by the given matcher
+ */
+fun <T> succeedsWith(successMatcher: Matcher<T>) = object : Matcher<FlowStateMachine<out T>> {
```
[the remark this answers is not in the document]
distributedleetravis: Agreed (changed to `flowSucceedsWithResult`, for consistency with `rpcSucceedsWithResult`).

core/src/test/kotlin/net/corda/core/flows/matchers/FlowMatchers.kt
```diff
@@ -8,11 +8,24 @@ import net.corda.core.utilities.getOrThrow
...
  * Matches a Flow that succeeds with a result matched by the given matcher
  */
-fun <T> succeedsWith(successMatcher: Matcher<T>) = object : Matcher<FlowStateMachine<T>> {
-    override val description: String
-        get() = "A flow that succeeds with ${successMatcher.description}"
+fun <T> succeeds() = object : Matcher<FlowStateMachine<T>> {
+    override val description: String = "is a flow that succeeds"
 
     override fun invoke(actual: FlowStateMachine<T>): MatchResult = try {
+        actual.resultFuture.getOrThrow()
```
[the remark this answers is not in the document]
distributedleetravis: Done - it means the flow/rpc matchers can both re-use the same future matchers.

core/src/test/kotlin/net/corda/core/flows/matchers/RpcMatchers.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import com.natpryce.hamkrest.MatchResult
+import com.natpryce.hamkrest.Matcher
+import net.corda.core.messaging.FlowHandle
+import net.corda.core.utilities.getOrThrow
+
+/**
+ * Matches an RPC flow handle that succeeds with a result matched by the given matcher
+ */
+fun <T> rpcSucceeds() = object : Matcher<FlowHandle<T>> {
```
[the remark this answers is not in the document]
distributedleetravis: The problem with this is that if we have

```kotlin
fun <T> succeeds(): Matcher<FlowHandle<T>>...
```

and

```kotlin
fun <T> succeeds(): Matcher<FlowStateMachine<T>>...
```

they look like duplicates to the compiler - we can't overload return types solely.

core/src/test/kotlin/net/corda/core/flows/mixins/WithContracts.kt
distributedleetravis: Moved into the more generic `WithMockNet` for now.

core/src/test/kotlin/net/corda/core/flows/mixins/WithMockNet.kt
distributedleetravis: Much nicer :)

core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
```diff
@@ -49,10 +58,9 @@ class ReceiveMultipleFlowTests {
             } as FlowLogic<Unit>
         }
 
-        val flow = nodes[0].services.startFlow(initiatingFlow)
-        mockNet.runNetwork()
-        val receivedAnswer = flow.resultFuture.getOrThrow()
-        assertThat(receivedAnswer).isEqualTo(answer)
+        assert.that(
+                nodes[0].startFlowAndRunNetwork(initiatingFlow),
+                willReturn(isA(equalTo(answer))))
```
sollecitom: No way of making it read like `willReturn(answer)`?

core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
```diff
@@ -61,10 +69,10 @@ class ReceiveMultipleFlowTests {
...
         val stringValue = "Thriller"
         nodes[2].registerAnswer(AlgorithmDefinition::class, stringValue)
-        val flow = nodes[0].services.startFlow(ParallelAlgorithmMap(nodes[1].info.singleIdentity(), nodes[2].info.singleIdentity()))
-        mockNet.runNetwork()
-        val result = flow.resultFuture.getOrThrow()
-        assertThat(result).isEqualTo(doubleValue * stringValue.length)
+
+        assert.that(
+                nodes[0].startFlowAndRunNetwork(ParallelAlgorithmMap(nodes[1].info.singleIdentity(), nodes[2].info.singleIdentity())),
+                willReturn(equalTo(doubleValue * stringValue.length)))
```
sollecitom: Same here, `willReturn(doubleValue * stringValue.length)`?

core/src/test/kotlin/net/corda/core/flows/ReceiveAllFlowTests.kt
```diff
@@ -73,12 +81,10 @@ class ReceiveMultipleFlowTests {
...
-        val flow = nodes[0].services.startFlow(ParallelAlgorithmList(nodes[1].info.singleIdentity(), nodes[2].info.singleIdentity()))
-        mockNet.runNetwork()
-        val data = flow.resultFuture.getOrThrow()
-        assertThat(data[0]).isEqualTo(value1)
-        assertThat(data[1]).isEqualTo(value2)
-        assertThat(data.fold(1.0) { a, b -> a * b }).isEqualTo(value1 * value2)
+
+        assert.that(
+                nodes[0].startFlowAndRunNetwork(ParallelAlgorithmList(nodes[1].info.singleIdentity(), nodes[2].info.singleIdentity())),
+                willReturn(equalTo(listOf(value1, value2))))
```
sollecitom: As above

vkolomeyko:
It is best to squash and rebase changes when integrating into OS `master`

PR #3657, opened 2018-07-19
CORDA-1686: Make "rpc-client-sender" daemon (#3455)

vkolomeyko:
Such that they do not prevent application that started them from exiting.

(cherry picked from commit 3492350)


fenryka:
What does this enable? does this only impact DemoBench?

vkolomeyko:
Jira: https://r3-cev.atlassian.net/browse/CORDA-1686 has details.

So yes it impacts DemoBench + when people close Explorer GUI there is a dangling Java process remains on Windows at least that has to be killed manually.

vkolomeyko:
This is now dependant on #3689 being merged.

vkolomeyko:
There is no clarity which additional OS branch either than `master` should receive those changes and no-one aggressively asking for this fix.
Closing PR without merging.

PR #3658, opened 2018-07-19
Faster lookup of node version info which improves startup time

shamsasari:
It turns out that the library we use is slow. This shaves off at least 0.5 secs.


shamsasari:
This is part of more work to speed up the startup of the node. I can include this change in that if this PR is too insignificant.

PR #3411, opened 2018-06-21
H2 cash selection needs to handle multiple onlyFromIssuerParties and/…

cxyzhang0:
…or withIssuerRefs

Current CashSelectionH2Impl works only if set onlyFromIsuerParties contains 0 or one member.  Looks like with two members, the IN clause becomes ('A, B') instead of ('A', 'B').  The latter is our desired outcome.  
This PR fixes that, and the potential issue with withIssuerRefs as well by explicitly declaring and setting a query parameter for each member in the sets onlyFromIssuerParties and withIssuerRefs. That should be ok since I don't anticipate either set being large - as a matter of fact, Cash masks withIssuerRefs as an empty set as this moment.
I manually tested the onlyFromIssuerParties and withIssuerRefs.

# PR Checklist:

- [ ] Have you run the unit, integration and smoke tests as described here? https://docs.corda.net/head/testing.html
The current suit of tests should cover that.  I added one unit test to cover the multiple onlyFromParties. Manual testing of multiple members in either set has been done.
- [ ] If you added/changed public APIs, did you write/update the JavaDocs?
N/A
- [ ] If the changes are of interest to application developers, have you added them to the changelog, and potentially release notes?
N/A
- [ ] If you are contributing for the first time, please read the agreement in CONTRIBUTING.md now and add to this Pull Request that you agree to it.
Done.
Thanks for your code, it's appreciated! :)

PR #3659, opened 2018-07-20
h2 cash selection query to support multiple onlyFromIssuerParties and…

cxyzhang0:
… withIssuerRefs

This is the same PR as #3411 which was closed due to merge complications.

Current CashSelectionH2Impl works only if set onlyFromIsuerParties contains 0 or one member.  Looks like with two members, the IN clause becomes ('A, B') instead of ('A', 'B').  The latter is our desired outcome.  
This PR fixes that, and the potential issue with withIssuerRefs as well by explicitly declaring and setting a query parameter for each member in the sets onlyFromIssuerParties and withIssuerRefs. That should be ok since I don't anticipate either set being large - as a matter of fact, Cash masks withIssuerRefs as an empty set as this moment.
I manually tested the onlyFromIssuerParties and withIssuerRefs.


cxyzhang0:
@szymonsztuka I closed #3411 due to git push having brought in many unrelated commits.  So open this one as the replacement.  Please review.

cxyzhang0:
@szymonsztuka Szymon, thank you for the review.  I have made the two changes in the test.  Please let me know if additional changes are needed.

PR #3661, opened 2018-07-20
Removed PersistentNetworkMapCache.start()

shamsasari:
The loading of all node infos in the start method was unnecessary, both for the changePublisher and _loadDBSuccess, and the setting of _registrationFuture was incorrect.


shamsasari:
Good spot, addressed.

PR #3665, opened 2018-07-23
ENT-2296 - Fix typo in docs

tlil:


joeldudleyr3:
@tlil Hey Tommy, could you raise changes like these on OS instead? Then they'll get pulled over by the OS -> ENT merge and both docsites will be updated.

tlil:
This is open source @joeldudleyr3  - the ticket was moved from ENT to OS already. :-)

joeldudleyr3:
Ah, sorry :)

PR #3674, opened 2018-07-23
CORDA-1585 - Tidy up participate docs + pdf toctree backport

anthonykeenan:
I've re-arranged the contribution section to put it at the top level of the tree, removed the 'release process' section, moved the release notes/change log to also put them at the top of the toctree, and everything else I've put in order as subheadings under contribution. Which I think makes more logical sense.

image

I've also backported @tlil's commit from enterprise that fixes the pdf toctree issue.


docs/source/index.rst
```diff
@@ -55,38 +55,39 @@ We look forward to seeing what you can do with Corda!
...
+
+.. conditional-toctree::
+   :caption: Participate
+   :maxdepth: 2
+   :if_tag: htmlmode
+   
+   contributing-index.rst
+   corda-repo-layout.rst
+   deterministic-modules.rst
+   building-container-images.rst
```
joeldudleyr3: I know it was already like this, but this doesn't seem like part of `Participate`? Would it fit better under `Nodes`?

docs/source/contributing-index.rst
```diff
@@ -0,0 +1,17 @@
...
+.. toctree::
+   :maxdepth: 1
+
+   contributing
+   contributing-flow-state-machines
+   building-corda
+   codestyle
+   testing
+   building-the-docs
+   api-scanner
```
joeldudleyr3: I'd reorder this. `contributing` -> `building-corda` -> `testing` -> `codestyle` -> `building the docs` -> `api-scanner` -> `contributing-flow-state-machines`.

`contributing-flow-state-machines`/`api-scanner` are advanced topics. Makes sense to me to put building and testing next to each other.

docs/source/index.rst
anthonykeenan: I totally agree, looking at it.

tlil:
Ah, thanks, I'd forgotten to port the conditional TOC back to OS. Whoops. :)

docs/source/contributing-index.rst
tlil: +1

PR #3677, opened 2018-07-24
CORDA-1840 Smarter checkpoint writing

blsemo:
Explicitly add/update checkpoints rather than calling `addOrUpdate` which will run a `select` statement and then a add or update statement.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/persistence/DBCheckpointStorage.kt
+++ b/node/src/main/kotlin/net/corda/node/services/persistence/DBCheckpointStorage.kt
@@ -36,12 +36,21 @@ class DBCheckpointStorage : CheckpointStorage {
     )
 
     override fun addCheckpoint(id: StateMachineRunId, checkpoint: SerializedBytes<Checkpoint>) {
-        currentDBSession().saveOrUpdate(DBCheckpoint().apply {
+        currentDBSession().save(DBCheckpoint().apply {
             checkpointId = id.uuid.toString()
             this.checkpoint = checkpoint.bytes
             log.debug { "Checkpoint $checkpointId, size=${this.checkpoint.size}" }
         })
     }
+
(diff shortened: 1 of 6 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/transitions/TopLevelTransition.kt
```diff
@@ -162,7 +162,7 @@ class TopLevelTransition(
                 )
             } else {
                 actions.addAll(arrayOf(
-                        Action.PersistCheckpoint(context.id, newCheckpoint),
+                        Action.PersistCheckpoint(context.id, newCheckpoint, initialCheckpoint = false),
```
rick-r3: I was thinking we would look at `isAnyCheckpointPersisted` in `StateMachineState` (i.e. `currentState`).

I think there's potentially for unstarted flows to have a checkpoint.... maybe 

See `SingleThreadedStateMachine.createFlowFromCheckpoint`.  It can create unstarted...

node/src/main/kotlin/net/corda/node/services/persistence/DBCheckpointStorage.kt
```diff
@@ -36,13 +36,22 @@ class DBCheckpointStorage : CheckpointStorage {
...
-        currentDBSession().saveOrUpdate(DBCheckpoint().apply {
+        currentDBSession().save(DBCheckpoint().apply {
             checkpointId = id.uuid.toString()
             this.checkpoint = checkpoint.bytes
             log.debug { "Checkpoint $checkpointId, size=${this.checkpoint.size}" }
         })
     }
 
+    override fun updateCheckpoint(id: StateMachineRunId, checkpoint: SerializedBytes<Checkpoint>) {
+        currentDBSession().update(DBCheckpoint().apply {
```
rick-r3: I'm assuming this does a blind update (and no select) ?

PR #3682, opened 2018-07-24
CORDA-1585 - Tidy up participate docs

anthonykeenan:
Backported the changes to the participate docs from master
Original PR here: https://github.com/corda/corda/pull/3674


tlil:
Duh, GitHub... approving on a stale page doesn't take the latest into account? :P 👍 with that latest commit.

anthonykeenan:
@tlil - I've removed your PDF toctree thing from v3 as it looks like it still uses the old pdf generation method which I'm not entirely sure it would work with, also the v3 branch has the old index ordering so I've just put it into there.

anthonykeenan:
Oh I hadn't finished writing, you're too efficient.

tlil:
>@tlil - I've removed your PDF toctree thing from v3 as it looks like it still uses the old pdf generation method which I'm not entirely sure it would work with, also the v3 branch has the old index ordering so I've just put it into there.

Yes, that's fine. Should work, but with regards to the TOC layout, it doesn't make sense to have it anyway, so I'm fine with that.

PR #3686, opened 2018-07-24
CORDA-1760 CLI UX guidelines

anthonykeenan:


docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
+CLI UX Guide
+============
+
+Command line options
+--------------------
+
+Command line utilities should use picocli (http://picocli.info) to provide a unified interface and follow the conventions in the picocli documentation, some of the more important of which are repeated below.
```
tlil: `s/picocli/Picocli/g`

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Option names
+~~~~~~~~~~~~
+
+* Options should be specified command line using a double dash, e.g. ``--parameter``.
```
tlil: ... on the ...

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+* Options that consist of multiple words should be separated via hyphens e.g. ``--my-multiple-word-paramter-name``
```
tlil: `s/paramter/parameter/`

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Short names
+~~~~~~~~~~~
+
+* Where possible a POSIX style short option should be provided for ease of use (see http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02)
+
+  * These should be prefixed with a single hyphen
+  * For example ``-v`` for ``--verbose``, ``-d`` for ``--dev-mode``
+
+* The picocli interface allows combinations of options without parameters, for example, -v and -d can be combined as -vd
```
tlil: ... allows the user to combine parameterless options, for example, ...

```
``-v``
``-d``
``-vd``
```

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Positional parameters
+~~~~~~~~~~~~~~~~~~~~~
+
+* Parameters specified without an option should ideally all be part of a list
+
+  * For example, use ``java -jar test.jar file1 file2 file3``, the parameters file1, file2 and file3 should be a list of files that all are all acted on together (e.g. a list of cordapps)
```
tlil: `s/use/in/`

```
``file1``
``file2``
``file3``
```

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+* Avoid using positional parameters to mean different thing, which involves someone remembering which order things need to be specified
```
tlil: `s/thing/things/`

tlil: ... remembering in which order ...

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+  * For example, avoid ``java -jar test.jar configfile1 cordapp1 cordapp2`` where parameter 1 is the config file and any subsequent parameters are the cordapps
```
tlil: `s/cordapps/CorDapps/`

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+  * Use ``java -jar test.jar cordapp1 cordapp2 --config-file configfile1`` instead
```
tlil: Arguably, here it'd actually be cleaner to use something like:

```
java -jar test.jar --cordapps cordapp1 cordapp2 --config-file configfile1
```

Wdyt?

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Defaults
+~~~~~~~~
+
+* Flags should have sensible defaults
+* Boolean flags should always default to false. Specifying the flag without a parameter should set it to true. For example --my-flag should be equal to --my-flag=true and no option should be equal to --my-flag=false
```
tlil: Double backticks:

```
``--my-flag``
```

etc.

tlil: In this case, perhaps use a flag that's actually a verb to make that clearer and neater:

```
``--use-blah``
```

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+* Do a bit of work to figure out reasonable defaults. Nobody likes having to set a dozen flags before the tool will cooperate
+
+Adding a new option
+~~~~~~~~~~~~~~~~~~~
+
+* Boolean options should start with is or has?
```
tlil:
```
``--is-X``
``--has-Y``
``--use-Z``
```

Do you have any examples of `--is-X` and `--has-Y` ?

Let's try to spell out some realistic examples and see what consistent style makes the most sense. :)

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+* Any new options must be  documented in the docsite.
```
tlil: and in the ``--help`` screen

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+* Avoid using acronyms in option names and try and make them as descriptive as possible.
```
tlil: Avoid using -> Never use ? :)

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Parameter stability
+~~~~~~~~~~~~~
```
tlil: Some more tildes here :)

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+
+* Avoid removing parameters. If, for some reason, a parameter needs to be renamed, add a new parameter with the new name and deprecate the old parameter, or alternatively keep both versions of the parameter.
+
+
+Application behavior
+--------------------
+
+* Set exit codes using exitProcess. Zero means success. Other numbers mean errors. Setting a unique error code (starting from 1) for each thing that can conceivably break makes your tool shell-scripting friendly
```
tlil:
```
``exitProcess``
```

tlil: Set exit codes using exitProcess. Zero means success. Other numbers mean errors. Setting a unique error code (starting from 1) for each thing that can conceivably break makes your tool shell-scripting friendly

=>

Set exit codes using exitProcess.
* Zero means success.
* Other numbers mean errors.

Setting a unique error code (starting from 1) for each thing that can conceivably break makes your tool shell-scripting friendly.

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+* Your ``--help`` text or other docs should ideally include examples. Writing examples is also a good way to find out that your program requires a dozen flags to do anything
```
tlil: `s/that/if/` ?

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,62 @@
...
+* Don’t print logging output to the console unless the user requested it via a ``-–verbose`` flag (conventionally shortened to -v) or a ``-–log-to-console`` flag. Logs should be either suppressed or saved to a text file during normal usage, except for errors, which are always OK to print
```
tlil:
```
``-v``
```

tlil: Mention that stacktraces should never be seen in the console unless `--log-to-console`` is specified.

tlil: We can also standardise the ``--logging-level`` option used by the node.

docs/source/contributing.rst
```diff
@@ -65,20 +65,7 @@ make merging more complicated.
...
-
-* Your ``--help`` text or other docs should ideally include examples. Writing examples is also a good way to find out
-  that your program requires a dozen flags to do anything
-
-* Flags should have sensible defaults
-
-* Don’t print logging output to the console unless the user requested it via a ``–verbose`` flag (conventionally
-  shortened to ``-v``) or a ``–log-to-console`` flag. Logs should be either suppressed or saved to a text file during
-  normal usage, except for errors, which are always OK to print
+Make sure any changes to CLI applications conform to the :doc:`cli-ux-guidelines`.
```
tlil: Once we have the base class(es) for CLI, we should make sure we reference to these as well in `cli-ux-guidelines` and how to achieve conformity

docs/source/cli-ux-guidelines.rst
anthonykeenan: Ahah... I anticipated this... and after looking at their website they refer to it as picocli, hence why I've referred to it as such!

tlil: Ah, fair enough.. I just read the first sentence where it was capitalised but that was obvs since it was at the beginning of a sentence :)

docs/source/cli-ux-guidelines.rst
anthonykeenan: Yeah, I was 50/50 over that. I think for brevity, having one parameter, as long as it is obvious, that doesn't require an option flag is pretty standard. Am happy to change it though and add a 'no parameters with an option flag' directive. Do you have an opinion @roastario ?

tlil: You could show both alternatives?

docs/source/cli-ux-guidelines.rst
anthonykeenan: Well... sometimes they're so obvoius --use-tcp? --vm-name? Would we really want to spell those out?

tlil: `--use-tcp` is fine.  The latter I'd argue for using `--machine-name` instead perhaps? :)

So yeah, okay, you can keep it to not confuse the former case I suppose.

docs/source/contributing.rst
kasiastreich: I will raise a PR with general class for help, version, logging level + config file handling with some validation. It can be reused in Corda.

docs/source/cli-ux-guidelines.rst
```diff
@@ -0,0 +1,72 @@
...
+--------------------
+
+Command line utilities should use picocli (http://picocli.info) to provide a unified interface and follow the conventions in the picocli documentation, some of the more important of which are repeated below.
+
+Option names
+~~~~~~~~~~~~
+
+* Options should be specified on the command line using a double dash, e.g. ``--parameter``.
+* Options that consist of multiple words should be separated via hyphens e.g. ``--my-multiple-word-parameter-name``.
+* A ``--help`` option should be provided which details all possible options with a brief description and any short name equivalents.
```
kasiastreich: We should also provide --version option (devops very often asked for that)

tlil: Yep, good point :+1:

docs/source/contributing.rst
tlil: Please coordinate with @anthonykeenan and @roastario , @kasiastreich - they are looking at this as well.

docs/source/cli-ux-guidelines.rst
anthonykeenan: Oh yes good idea

anthonykeenan:
Yes, I was just looking at that. It makes sense.

anthonykeenan:
Could we create a story to create the script as part of the build? Would the information in that script be enough to use for https://r3-cev.atlassian.net/browse/CORDA-1759?

tlil:
As part of the build can be done, ... but it would be nice to make the application able to produce it when asked instead of having to distribute a second artefact and keep that in sync etc. I'll create a story for generating and "installing" completion scripts, but I'd be inclined to do so through the app/tool itself rather than at build time. Will elaborate on the ticket.

As for your second question, yes, the output of that should give sufficient info. We can even hook into that process and spit out the info in our own format.

tlil:
I can pick up that ticket btw. I've already done something similar for DJVM

tlil:
Also, something along the lines of "once installed, will automatically be kept up-to-date with the latest version".

We can add  that after CORDA-1848 is complete and integrated with @kasiastreich's work though.

anthonykeenan:
Actually, do we need an option at all?? Should we just generate and install it when it's run? Since we're going to be checking whether it exists anyway to see whether it exists as part of the upgrade functionality?

tlil:
@anthonykeenan let's move that discussion to the JIRA ticket :)

anthonykeenan:
Pushed some more information based on adding the CLI base class and implementing it in the node.

docs/source/cli-ux-guidelines.rst
```diff
@@ -1,3 +1,9 @@
+.. highlight:: kotlin
+.. raw:: html
+
+   <script type="text/javascript" src="_static/jquery.js"></script>
```
tlil: Fanciness? :P

docs/source/cli-ux-guidelines.rst
```diff
@@ -39,6 +42,17 @@ Positional parameters
...
 
+Standard options
+~~~~~~~~~~~~~~~~
+
+* A ``--help`` option should be provided which details all possible options with a brief description and any short name equivalents. A ``-h`` short option should also be provided.
+* A ``--version`` option that should output the version number of the software. A ``-V`` short option should also be provided.
+* A ``--logging-level`` option should be provided which specifies the logging level to be used in any logging files. Acceptable values should be ``DEBUG``, ``TRACE``, ``INFO``, ``WARN`` and ``ERROR``.
+* ``--verbose`` and ``--log-to-console`` options should be provided (both equivalent) which specifies that logging output should be displayed in the console.
+  A ``-v`` short option should also be provided.
+* A ``--install-shell-extensions`` option should be provided that creates and installs a bash completion file.
```
tlil: Re my comment on the other PR, perhaps add a ``--color <ALWAYS,NEVER,AUTO>`` option as well? Where `AUTO` is the default.

docs/source/cli-ux-guidelines.rst
```diff
@@ -59,6 +73,53 @@ Parameter stability
...
+            private var extraUsefulness: Boolean = false // This default option will be shown in the help output
+
+            override fun runProgram(): Int { // override this function to run the actual program
+                try {
+                    // do some stuff
+                } catch (KnownException: ex) {
+                    return 100 // return a special exit code for known exceptions
+                }
+
+                return 0 // this is the exit code to be returned to the system
```
tlil: Update with `SUCCESS` after exposing `SUCCESS` and `FAILURE` in `cliutils`?

docs/source/cli-ux-guidelines.rst
anthonykeenan: Agreed not to implement this after discussions

PR #3687, opened 2018-07-25
Removes Identity page. Reworks it into the Network page.

joeldudleyr3:


docs/source/key-concepts-ecosystem.rst
```diff
@@ -4,41 +4,44 @@ The network
 .. topic:: Summary
 
    * *A Corda network is made up of nodes running Corda and CorDapps*
-   * *The network is permissioned, with access controlled by a doorman*
    * *Communication between nodes is point-to-point, instead of relying on global broadcasts*
+   * *Each node has a certificate mapping their network identity to a real-world legal identity*
+   * *The network is permissioned, with access controlled by a doorman*
```
shamsasari: Perhaps just add why we have a doorman - to verify those joining the network are who they say they are.

docs/source/key-concepts-ecosystem.rst
```diff
@@ -4,41 +4,44 @@ The network
 .. topic:: Summary
 
    * *A Corda network is made up of nodes running Corda and CorDapps*
-   * *The network is permissioned, with access controlled by a doorman*
    * *Communication between nodes is point-to-point, instead of relying on global broadcasts*
+   * *Each node has a certificate mapping their network identity to a real-world legal identity*
+   * *The network is permissioned, with access controlled by the network operator*
```
shamsasari: Hmm, still gives the impression of a single overlord. How about "The network is permissioned, with real-world identities verified by the network operator"?

shamsasari: Yep.

PR #3688, opened 2018-07-25
Removed StateMachineManagerInternal as StateMachineManager is already internal

shamsasari:


exFalso:
Fair enough, but it is exposed through StartedNode to internal tests.

Is there any reason why these should be exposed? Currently the functions on `StateMachineManager` are "safe" to call and their contract is quite clear. The Internal functions however are called in very specific contexts and are unsafe to call from other places.

PR #3690, opened 2018-07-25
Eliminate StartedNode

distributedleetravis:
It turns out that when you remove from `StartedNode` all of the properties that are only used in testing (leaving these in `TestStartedNode` and `StartedNodeWithInternals`), all you're left with is `nodeInfo`. So we make `Node.start()` just return a `NodeInfo`, and `InternalMockNetwork.MockNode.start()` return a `TestStartedNode`, eliminating `StartedNode` entirely.

It's not ideal that `AbstractNode` should have a type parameter indicating whether it returns a `NodeInfo` or a `TestStartedNode`. We should have `start` return `NodeInfo` in all cases, and provide a separate test method on `MockNode` for running `start` then constructing a `TestStartedNode` using the resulting `NodeInfo` and whatever internals of `MockNode` need to be made available.

### DIFF
--- a/client/rpc/src/integration-test/java/net/corda/client/rpc/CordaRPCJavaClientTest.java
+++ b/client/rpc/src/integration-test/java/net/corda/client/rpc/CordaRPCJavaClientTest.java
@@ -8,8 +8,7 @@ import net.corda.finance.flows.AbstractCashFlow;
 import net.corda.finance.flows.CashIssueFlow;
 import net.corda.finance.flows.CashPaymentFlow;
 import net.corda.finance.schemas.CashSchemaV1;
-import net.corda.node.internal.StartedNode;
-import net.corda.node.internal.StartedNodeWithInternals;
+import net.corda.node.internal.NodeWithInfo;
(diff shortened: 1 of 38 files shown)


shamsasari:
This is looking good.

You could get rid of the AbstractNode type parameter by doing the `TestStartedNode` stuff in `net.corda.testing.node.internal.InternalMockNetwork#createNode`.

shamsasari:
Also you have access to the NodeInfo via `services.myInfo` so maybe returning it is not necessary.

distributedleetravis:
In order to create a `TestStartedNode`, we need the `rpcOps` and `notaryService` values created inside `start` - they do get registered elsewhere in the internals of the node, but aren't readily accessible (which is why they get pulled out and exposed via `TestStartedNode`).

It would be good to have `start` just return `NodeInfo` in all cases, but also a lot of work to change all the test code that's currently expected to call it and get back a `TestStartedNode`.

I'm sure further improvement is possible, but I think this PR is about as big as it's safe for it to be.

PR #3698, opened 2018-07-26
CORDA-1844: Support for high throughput Observables shipped via RPC

vkolomeyko:
Not a proper fix for production code yet, but a testcase exposing the problem.
Also minor extension to RPCServer which will allow to shut it down quicker.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/messaging/RPCServer.kt
+++ b/node/src/main/kotlin/net/corda/node/services/messaging/RPCServer.kt
@@ -18,11 +18,7 @@ import net.corda.core.serialization.SerializationContext
 import net.corda.core.serialization.SerializationDefaults
 import net.corda.core.serialization.SerializationDefaults.RPC_SERVER_CONTEXT
 import net.corda.core.serialization.deserialize
-import net.corda.core.utilities.Try
-import net.corda.core.utilities.contextLogger
-import net.corda.core.utilities.days
-import net.corda.core.utilities.debug
-import net.corda.core.utilities.seconds
+import net.corda.core.utilities.*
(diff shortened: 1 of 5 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/RPCDriver.kt
```diff
@@ -485,7 +486,7 @@ data class RPCDriverDSL(
                 configuration
         )
         driverDSL.shutdownManager.registerShutdown {
-            rpcServer.close()
+            rpcServer.close(1.millis)
```
mnesbit: This does look very fragile

vkolomeyko: I can add an optional parameter to `RPCDriver`, but if we are shutting down the whole driver we might just shut down RPCServer as quickly as possible. That was my reason for 1ms.

node/src/main/kotlin/net/corda/node/services/messaging/RPCServer.kt
```diff
@@ -247,9 +243,10 @@ class RPCServer(
         }
     }
 
-    fun close() {
+    fun close(queueDrainTimeout: Duration = 1.hours) {
```
mnesbit: 5 seconds?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/RPCDriver.kt
mnesbit: OK make it configurable

PR #3699, opened 2018-07-26
CORDA-1839: VaultService.trackBy races with the vault, as does whenConsumed

josecoll:
See details in JIRA.

Effectively there are 3 separate fixes in this PR:
1. Re-written `whenConsumed` in terms of `trackBy` to take into account anything already in the vault.
2. Remove redundant new transaction block.
3. Fix race condition in trackBy() when used concurrently with Vault updates.

Items 1 and 2 are already included in the [Reference States PR](https://github.com/corda/corda/pull/3525).
Item 3 is resolved by alligning the mutex lock closer to transaction commit (in the recordTransactions calling stack) and thus preventing the trackBy() method from acquiring the lock (and associated observable stream) before state is committed to the database.

### DIFF
--- a/client/rpc/src/smoke-test/kotlin/net/corda/kotlin/rpc/StandaloneCordaRPClientTest.kt
+++ b/client/rpc/src/smoke-test/kotlin/net/corda/kotlin/rpc/StandaloneCordaRPClientTest.kt
@@ -199,31 +199,6 @@ class StandaloneCordaRPClientTest {
     }
 
     @Test
-    fun `test vault track by after triggering flow`() {
-
-        // Issue some cash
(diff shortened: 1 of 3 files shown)


josecoll:
@rick-r3 any suggestions on how to put this change through its paces? 
(eg. simulate the race scenario defined in the JIRA)

rick-r3:
I don't think this solves the problem.  Referring back to the JIRA ticket, it's still possible for F to do all steps except commit, prior to all of T's steps.  I'm thinking about it some more.  Will put some more thoughts on the JIRA.

mikehearn:
ping!

josecoll:
Overall design under review.

PR #3700, opened 2018-07-26
CORDA-1848 Add example alias and autocomplete for CLI tools

anthonykeenan:
If you run bootstrapper with the `--install-shell-extensions` parameter it will add a line to the `~/.bashrc` settings file creating an alias so that the bootstrapper can be run using `bootstrapper <options>` from the commandline.

It will also create an auto-completion file in `~/.completion/bootstrapper` and add another line to the `~/.bashrc` file which loads all of the files in the `~/.completion` directory.

To test for bash run
```bash
./gradlew clean :tools:bootstrapper:jar
cd ./tools/bootstrapper/build/libs
java -jar network-bootstrapper-4.0-SNAPSHOT.jar --install-shell-extensions
```

Then for bash run:
```bash
. ~/.bashrc
```

For zsh run:
```bash
. ~/.zshrc
```

### TODO
- [x] Test on a unix & mac - I tested on Unix - Tommy tested on mac
- [x] Add support for zsh - done but need to update to picocli 3.3.1 when released to complete
- ~~Merge with @kasiastreich's base class~~ - will do as part of https://r3-cev.atlassian.net/browse/CORDA-1833

### DIFF
--- a/tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
+++ b/tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
@@ -47,7 +47,7 @@ class Main : Runnable {
     @Option(names = ["--verbose"], description = ["Enable verbose output."])
     var verbose: Boolean = false
 
-    @Option(names = ["--install-shell-extensions"], description = ["Install bootstrapper alias and autocompletion in bash"])
+    @Option(names = ["--install-shell-extensions"], description = ["Install bootstrapper alias and autocompletion for bash and zsh"])
     var install: Boolean = false
 
     // Return the lines in the file if it exists, else return an empty mutable list
@@ -123,7 +123,7 @@ class Main : Runnable {
 
         println("Installation complete, $alias is available in bash with autocompletion. ")
         println("Type `$alias <options>` from the commandline.")
-        println("Restart bash for this to take effect, or run `. ~/.bashrc` to re-initialise bash now")
+        println("Restart bash for this to take effect, or run `. ~/.bashrc` in bash or `. ~/.zshrc` in zsh to re-initialise your shell now")
     }
 
     private fun checkForAutoCompleteUpdate(alias: String) {
--- a/tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
+++ b/tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
@@ -1,7 +1,6 @@
 package net.corda.bootstrapper
 
 import com.jcabi.manifests.Manifests
-import net.corda.core.crypto.SecureHash
 import net.corda.core.internal.*
 import net.corda.nodeapi.internal.network.NetworkBootstrapper
 import picocli.CommandLine
@@ -77,11 +76,10 @@ class Main : Runnable {
 
     private val userHome: Path by lazy { Paths.get(System.getProperty("user.home")) }
     private val jarLocation: Path by lazy { this.javaClass.location.toPath() }
-    private val jarHash: SecureHash by lazy { jarLocation.toFile().inputStream().hash() }
 
     // If on Windows, Path.toString() returns a path with \ instead of /, but for bash Windows users we want to convert those back to /'s
     private fun Path.toStringWithDeWindowsfication(): String = this.toAbsolutePath().toString().replace("\\", "/")
-    private fun jarSignature(alias: String, jarHash: SecureHash) = "# $alias: $jarHash"
+    private fun jarVersion(alias: String) = "# $alias - Version: ${CordaVersionProvider.releaseVersion}, Revision: ${CordaVersionProvider.revision}"
     private fun getAutoCompleteFileLocation(alias: String) = userHome / ".completion" / alias
 
     private fun generateAutoCompleteFile(alias: String) {
@@ -91,7 +89,7 @@ class Main : Runnable {
         picocli.AutoComplete.main("-f", "-n", alias, this.javaClass.name, "-o", autoCompleteFile.toStringWithDeWindowsfication())
 
         // Append hash of file to autocomplete file
-        autoCompleteFile.toFile().appendText(jarSignature(alias, jarHash))
+        autoCompleteFile.toFile().appendText(jarVersion(alias))
     }
 
     private fun installShellExtensions(alias: String) {
@@ -127,7 +125,7 @@ class Main : Runnable {
         var lastLine = ""
         autoCompleteFile.toFile().forEachLine { lastLine = it }
 
-        if (lastLine != jarSignature(alias, jarHash)) {
+        if (lastLine != jarVersion(alias)) {
             println("Old auto completion file detected... regenerating")
             generateAutoCompleteFile(alias)
             println("Restart bash for this to take effect, or run `. ~/.bashrc` to re-initialise bash now")
@@ -153,10 +151,15 @@ class Main : Runnable {
 }
 
 private class CordaVersionProvider : IVersionProvider {
+    companion object {
+        val releaseVersion: String by lazy { Manifests.read("Corda-Release-Version") }
+        val revision: String by lazy { Manifests.read("Corda-Revision") }
+    }
+
     override fun getVersion(): Array<String> {
         return arrayOf(
-                "Version: ${Manifests.read("Corda-Release-Version")}",
-                "Revision: ${Manifests.read("Corda-Revision")}"
+                "Version: $releaseVersion",
+                "Revision: $revision"
         )
     }
 }
--- a/tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
+++ b/tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
@@ -29,13 +29,13 @@ fun main(args: Array<String>) {
         versionProvider = CordaVersionProvider::class,
         mixinStandardHelpOptions = true,
         showDefaultValues = true,
-        description = ["Bootstrap a local test Corda network using a set of node conf files and CorDapp JARs"]
+        description = ["Bootstrap a local test Corda network using a set of node configuration files and CorDapp JARs"]
 )
 class Main : Runnable {
     @Option(
             names = ["--dir"],
             description = [
-                "Root directory containing the node conf files and CorDapp JARs that will form the test network.",
+                "Root directory containing the node configuration files and CorDapp JARs that will form the test network.",
                 "It may also contain existing node directories."
             ]
     )
@@ -47,7 +47,7 @@ class Main : Runnable {
     @Option(names = ["--verbose"], description = ["Enable verbose output."])
     var verbose: Boolean = false
 
-    @Option(names = ["--install", "-i"], description = ["Install bootstrapper alias and autocompletion in bash"])
+    @Option(names = ["--install-shell-extensions", "-i"], description = ["Install bootstrapper alias and autocompletion in bash"])
     var install: Boolean = false
 
     // Return the lines in the file if it exists, else return an empty mutable list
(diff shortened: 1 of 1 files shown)


tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,74 @@ class Main : Runnable {
...
+        }
+    }
+
+    // If on windows, Path.toString() returns a path with \ instead of /, but for bash windows users we want to convert those back to /'s
+    private fun Path.toStringWithDeWindowsfication(): String = this.toAbsolutePath().toString().replace("\\", "/")
+
+    private fun install(alias: String) {
+        // Get jar location and generate alias command
+        val jarLocation = this.javaClass.location.toPath()
+        val command = "alias $alias='java -jar ${jarLocation.toStringWithDeWindowsfication()}'"
```
tlil: Is this sufficiently escaped? Say, for instance, the JAR location contains a space :)

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,74 @@ class Main : Runnable {
     @Option(names = ["--verbose"], description = ["Enable verbose output."])
     var verbose: Boolean = false
 
+    @Option(names = ["--install", "-i"], description = ["Install bootstrapper alias and autocompletion in bash"])
+    var install: Boolean = false
+
+    // Return the lines in the file if it exists, else return an empty mutable list
+    private fun getFileLines(filePath: Path): MutableList<String> {
```
tlil: Pull helpers out to a separate class/object

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,74 @@ class Main : Runnable {
...
+    private fun install(alias: String) {
+        // Get jar location and generate alias command
+        val jarLocation = this.javaClass.location.toPath()
+        val command = "alias $alias='java -jar ${jarLocation.toStringWithDeWindowsfication()}'"
+
+        val userHome = Paths.get(System.getProperty("user.home"))
+
+        System.out.println("Generating $alias auto completion file")
+        val completionDir = userHome / ".completion"
+        if (!completionDir.exists()) {
```
tlil: Use `Files.createDirectories()` and thus loose the `exists()` check up front.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,74 @@ class Main : Runnable {
...
+        val autoCompletePath = (completionDir / "$alias").toStringWithDeWindowsfication()
+        picocli.AutoComplete.main("-f", "-n", alias, this.javaClass.name, "-o", autoCompletePath)
+
+        // get bash settings file
+        val bashSettingsFile = userHome / ".bashrc"
+        val bashSettingsFileLines = getFileLines(bashSettingsFile).toMutableList()
+
+        System.out.println("Updating bash settings files")
+        // Replace any existing bootstrapper alias. There can be only one.
+        bashSettingsFileLines.addOrReplaceIfStartsWith("alias $alias", command)
```
tlil: I wonder if we should have a comment with some meta-data so that we can further check whether to regenerate the completion files (if the JAR has been updated), or not. That way, we can provide a facility of keeping the completion files up-to-date automatically if that's what the user wants.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -29,7 +29,7 @@ fun main(args: Array<String>) {
         versionProvider = CordaVersionProvider::class,
         mixinStandardHelpOptions = true,
         showDefaultValues = true,
-        description = [ "Bootstrap a local test Corda network using a set of node conf files and CorDapp JARs" ]
+        description = ["Bootstrap a local test Corda network using a set of node conf files and CorDapp JARs"]
```
tlil: `s/conf/configuration/` (I know it's not your initial code, but would be nice to fix up :))

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,74 @@ class Main : Runnable {
     @Option(names = ["--verbose"], description = ["Enable verbose output."])
     var verbose: Boolean = false
 
+    @Option(names = ["--install", "-i"], description = ["Install bootstrapper alias and autocompletion in bash"])
```
tlil: Perhaps we should make this a bit more descriptive, e.g. `--install-shell-extensions` ? In case we have tools in which `--install` has a different contextual meaning / the end-user expectations around `install` is different. _Install_ is quite vague.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,74 @@ class Main : Runnable {
...
+        }
+    }
+
+    private fun MutableList<String>.addIfNotExists(line: String) {
+        if (!this.contains(line)) {
+            this.add(line)
+        }
+    }
+
+    // If on windows, Path.toString() returns a path with \ instead of /, but for bash windows users we want to convert those back to /'s
```
tlil: Windows :)

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,74 @@ class Main : Runnable {
...
+            completionDir.createDirectory()
+        }
+        val autoCompletePath = (completionDir / "$alias").toStringWithDeWindowsfication()
+        picocli.AutoComplete.main("-f", "-n", alias, this.javaClass.name, "-o", autoCompletePath)
+
+        // get bash settings file
+        val bashSettingsFile = userHome / ".bashrc"
+        val bashSettingsFileLines = getFileLines(bashSettingsFile).toMutableList()
+
+        System.out.println("Updating bash settings files")
```
tlil: Just `println` ? :)

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
anthonykeenan: I was just going to have a seperate section of the start up method that just checked whether the completion file existed, and if so, overwrote it with a new one - to keep it simple.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
anthonykeenan: This is all getting moved around later anyways, which is why I've left it all together for now. It will need more polishing.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
tlil: I was just thinking that we don't need to generate anything if `# XXX some hash / signature` exists... So makes it a quicker check. That way, after `--install-shell-extensions` has been run once, we could even look at running that check upon startup unless opted out.... ?

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
tlil: Fair 'nuf

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
tlil: I was thinking something like, we can append to `.bashrc` / `.zshrc`:

```bash
# completion: $alias ($versionHash)
alias $alias="java -jar ..."
```

tlil: Then, you can match against `# completion: $alias`, if exist, check `$versionHash` with current hash.... Et voila

tlil: So if you get multiple commands, it'd look something like:

```bash
# completion: network-bootstrapper (2344afd0)
alias network-bootstrapper="java -jar ..."

# completion: blob-inspector (ff0cb012)
alias blob-inspector="java -jar ..."
```

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
anthonykeenan: Thinking about it, it probably doesn't warrant a short option as it's a one off command rather than a regular one? Wdyt?

tlil: Yes, agree. The short option can go IMHO.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +48,106 @@ class Main : Runnable {
     @Option(names = ["--verbose"], description = ["Enable verbose output."])
     var verbose: Boolean = false
 
+    @Option(names = ["--install-shell-extensions", "-i"], description = ["Install bootstrapper alias and autocompletion in bash"])
```
tlil: Remove `-i` (short option)?

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +48,106 @@ class Main : Runnable {
...
+    var install: Boolean = false
+
+    @Option(names = ["--new-option"], description = ["This is a new option"])
```
tlil: ?

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +48,106 @@ class Main : Runnable {
...
+
+        generateAutoCompleteFile(alias)
+
+        // Get bash settings file
+        val bashSettingsFile = userHome / ".bashrc"
+        val bashSettingsFileLines = getFileLines(bashSettingsFile).toMutableList()
+
+        println("Updating bash settings files")
+        // Replace any existing bootstrapper alias. There can be only one.
+        bashSettingsFileLines.addOrReplaceIfStartsWith("alias $alias", command)
```
tlil: Just to deal with the unlikely scenario that the user has a pre-existing alias with the same name:

Let's either:
 (1) Make it `alias $alias="..." # some unique comment` and match including the comment
 (2) Create a `.bashrc.bak` just in case you were to overwrite something (which could piss off the user)

:-)

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -56,10 +151,15 @@ class Main : Runnable {
...
+        val releaseVersion: String by lazy { Manifests.read("Corda-Release-Version") }
+        val revision: String by lazy { Manifests.read("Corda-Revision") }
+    }
+
     override fun getVersion(): Array<String> {
         return arrayOf(
-                "Version: ${Manifests.read("Corda-Release-Version")}",
-                "Revision: ${Manifests.read("Corda-Revision")}"
+                "Version: $releaseVersion",
+                "Revision: $revision"
```
tlil: I wonder if it would be helpful to have a build date here as well. Wdyt?

tlil:
How are you getting on with the ZSH part, btw?

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
anthonykeenan: Yes good idea, I'll look at adding that to the manifest

tlil: Cool. Yeah, you can introduce it in the `jar` Gradle task, and then extract it here. :)

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,112 @@ class Main : Runnable {
     @Option(names = ["--verbose"], description = ["Enable verbose output."])
     var verbose: Boolean = false
 
+    @Option(names = ["--install-shell-extensions"], description = ["Install bootstrapper alias and autocompletion for bash and zsh"])
+    var install: Boolean = false
```
tlil: `s/install/installShellExtensions/` ?

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -47,7 +47,112 @@ class Main : Runnable {
...
+        if (install) {
+            installShellExtensions(alias)
+            return
+        } else {
+            checkForAutoCompleteUpdate(alias)
+        }
+    }
+
     override fun run() {
+        installOrUpdateShellExtensions("bootstrapper")
```
tlil: Should end the program after having processed `--install-shell-extensions`, rather than continue running I reckon.

tlil:
Comment out `shopt -s progcomp` to load for ZSH.

Also, for use with aliases in ZSH, we need to set `setopt COMPLETE_ALIASES`

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
anthonykeenan: Oh yes - missed that when I refactored

tlil:
Also, for ZSH, `compopt -o filenames` needs replacing. Use `complete -o filenames` instead. That's only for ZSH though, mind. So for Bash, you'd still need to use `compopt`. So we can switch based on `$BASH` and `$ZSH_NAME`

tlil:
`complete` is also present in Bash but got a different usage. So that's another avenue to explore, potentially.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
anthonykeenan: I've had a go at this but adding build date to the jar task causes it always to be seen as changed, which causes it to get built every time. I think the disadvantages of adding it in this case outweight the advantages of it.

see: https://discuss.gradle.org/t/dynamic-manifest-mf-attribute-new-date-causes-gradle-to-build-the-jar-every-time-even-though-nothing-else-changed/7347/12

tlil: Ah okay. I see

PR #3705, opened 2018-07-27
CORDA-1845: Check for min plaform version of 4 when building transactions with reference states

shamsasari:
Also includes some minor cleanup brought up in a previous PR.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
@@ -1,6 +1,7 @@
 package net.corda.core.transactions
 
 import co.paralleluniverse.strands.Strand
+import net.corda.core.CordaInternal
 import net.corda.core.DeleteForDJVM
 import net.corda.core.contracts.*
 import net.corda.core.cordapp.CordappProvider
@@ -9,9 +10,11 @@ import net.corda.core.crypto.SignableData
 import net.corda.core.crypto.SignatureMetadata
 import net.corda.core.identity.Party
 import net.corda.core.internal.FlowStateMachine
+import net.corda.core.internal.ensureMinimumPlatformVersion
 import net.corda.core.node.NetworkParameters
 import net.corda.core.node.ServiceHub
 import net.corda.core.node.ServicesForResolution
+import net.corda.core.node.ZoneVersionTooLowException
 import net.corda.core.node.services.AttachmentId
 import net.corda.core.node.services.KeyManagementService
 import net.corda.core.serialization.SerializationContext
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -95,11 +96,23 @@ open class TransactionBuilder @JvmOverloads constructor(
      * [HashAttachmentConstraint].
      *
      * @returns A new [WireTransaction] that will be unaffected by further changes to this [TransactionBuilder].
+     *
+     * @throws IllegalStateException if there are reference states and the minimum platform version is less than 4.
```
mikehearn: Can we define an unchecked custom exception for zone version being too low?

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -95,11 +96,23 @@ open class TransactionBuilder @JvmOverloads constructor(
...
      */
     @Throws(MissingContractAttachments::class)
     fun toWireTransaction(services: ServicesForResolution): WireTransaction = toWireTransactionWithContext(services)
 
+    @CordaInternal
     internal fun toWireTransactionWithContext(services: ServicesForResolution, serializationContext: SerializationContext? = null): WireTransaction {
+        val referenceStates = referenceStates()
+        if (referenceStates.isNotEmpty()) {
+            val minimumPlatformVersion = services.networkParameters.minimumPlatformVersion
+            check(minimumPlatformVersion >= 4) {
```
mikehearn: Can we extract this out to a function that throws a custom exception, so we always get the useful error message for every case where we do this type of check in future?

core/src/test/kotlin/net/corda/core/transactions/TransactionBuilderTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Rule
+import org.junit.Test
+
+class TransactionBuilderTest {
+    @Rule
+    @JvmField
+    val testSerialization = SerializationEnvironmentRule()
+
+    private val notary = TestIdentity(DUMMY_NOTARY_NAME).party
+    private val services = rigorousMock<ServicesForResolution>()
```
mikehearn: Mockito can be quite slow, maybe we can introduce a custom mock for ServicesForResolution for this sort of case of testing transaction building?

shamsasari: I think something for a separate PR.

shamsasari:
Comments addressed.

PR #3706, opened 2018-07-27
 Some base picocli classes used in network services 

kasiastreich:
Enjoy! Maybe it is of some use for Corda.
I also added hiding of stack traces from console appenders for node, no worries they are still written to file!

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -68,10 +68,7 @@ open class NodeStartup(val args: Array<String>) {
 
         val registrationMode = checkRegistrationMode()
         val cmdlineOptions: CmdLineOptions = if (registrationMode && !args.contains("--initial-registration")) {
-            "Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.".let {
-                println(it)
-                logger.info(it)
-            }
+            println("Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.")
             // Pretend that the node was started with `--initial-registration` to help prevent user error.
             NodeArgsParser().parseOrExit(*args.plus("--initial-registration"))
         } else {
--- a/node/src/main/kotlin/net/corda/node/utilities/CLIUtils.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/CLIUtils.kt
@@ -1,0 +1,119 @@
+package net.corda.node.utilities
+
+import com.jcabi.manifests.Manifests
+import net.corda.core.internal.exists
+import net.corda.core.internal.isReadable
+import net.corda.core.utilities.contextLogger
+import org.apache.logging.log4j.Level
+import picocli.CommandLine
+import picocli.CommandLine.*
+import java.nio.file.Path
+import kotlin.system.exitProcess
+import java.util.*
+
+/**
+ * Something heavily used in network services, I am not sure it's of much use in corda, but who knows. Definitely it was the key to making DevOps happy.
+ * Add it as
+ * `@CommandLine.Mixin
+ * lateinit var configParser: ConfigFilePathArgsParser`
+ *
+ * in your command class and then validate()
+ */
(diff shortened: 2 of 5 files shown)


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -68,10 +68,7 @@ open class NodeStartup(val args: Array<String>) {
 
         val registrationMode = checkRegistrationMode()
         val cmdlineOptions: CmdLineOptions = if (registrationMode && !args.contains("--initial-registration")) {
-            "Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.".let {
-                println(it)
-                logger.info(it)
-            }
+            println("Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.")
```
tlil: Why is this no longer logged, @kasiastreich ? Any particular reason?

node/src/main/kotlin/net/corda/node/utilities/CLIUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+     */
+    fun validator(): List<String>
+
+    /**
+     * Function that provides nice error handing of command line validation.
+     */
+    fun validate() {
+        val errors = validator()
+        if (errors.isNotEmpty()) {
+            logger.error(RED + "Exceptions when parsing command line arguments:")
```
tlil: You can use `picocli`'s built in console printer here to do conditional formatting (which nicely detects whether piped on printing to the console, etc). cc @anthonykeenan - for your follow-up PR.

node/src/main/kotlin/net/corda/node/utilities/CLIUtils.kt
```diff
@@ -0,0 +1,119 @@
...
+        descriptionHeading = "%n@|bold,underline Description|@:%n%n",
+        parameterListHeading = "%n@|bold,underline Parameters|@:%n%n",
+        optionListHeading = "%n@|bold,underline Options|@:%n%n",
+        commandListHeading = "%n@|bold,underline Commands|@:%n%n")
+abstract class ArgsParser {
+    @Option(names = [ "-v", "--verbose" ], description = ["If set, prints logging to the console as well as to a file."])
+    var verbose: Boolean = false
+
+    @Option(names = ["--logging-level"],
+            // TODO For some reason I couldn't make picocli COMPLETION-CANDIDATES work
```
tlil: @anthonykeenan if you struggle to get this working, let me know. I used that for DJVM and got it working, so if you run into any issues, I should be able to help. :)

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
kasiastreich: Because it was added by accident by Florian. This line initializes `logger` (by lazy) before function `initLogging` in `NodeStartup.kt`, there is meticulous initialization ordering ;)

node/src/main/kotlin/net/corda/node/utilities/CLIUtils.kt
kasiastreich: Good to know!

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
tlil: Ah :-) I see. Thanks

node/src/main/kotlin/net/corda/node/utilities/CLIUtils.kt
anthonykeenan: @kasiastreich do you use `bash` or `zsh`?
I wonder if it's another `zsh` issue.

kasiastreich: bash, but also I didn't spend much time on that. Where is DJVM code?

anthonykeenan: I had a play around and got it working so I'll update it and let you know.

PR #3707, opened 2018-07-27
CORDA-1568: Do not use "£" (GBP) symbol on Java source files as it causes problems on Windows.

vkolomeyko:
See: https://ci-master.corda.r3cev.com/viewLog.html?buildId=103073&tab=buildResultsDiv&buildTypeId=Corda_BuildWindowsTest

```
InteractiveShellJavaTest.flowStartWithComplexTypes
show details
No applicable constructor for flow. Problems were:
[party: Party]: missing parameter party
[pair: Pair]: missing parameter pair
[Amount]: Could not parse as a command: Did not recognise the currency in Â£10 or could not parse
```

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -162,7 +162,3 @@ scanApi {
 publish {
     name jar.baseName
 }
-
-compileTestJava {
-    options.encoding = 'UTF-8'
-}
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -162,3 +162,7 @@ scanApi {
 publish {
     name jar.baseName
 }
+
+compileTestJava {
+    options.encoding = 'UTF-8'
+}
--- a/tools/shell/src/test/java/net/corda/tools/shell/InteractiveShellJavaTest.java
+++ b/tools/shell/src/test/java/net/corda/tools/shell/InteractiveShellJavaTest.java
@@ -152,7 +152,7 @@ public class InteractiveShellJavaTest {
 
     @Test
     public void flowStartWithComplexTypes() throws InteractiveShell.NoApplicableConstructor {
-        check("amount: \u00A310", "10.00 GBP", FlowA.class);
+        check("amount: £10", "10.00 GBP", FlowA.class);
     }
 
     @Test
--- a/tools/shell/src/test/java/net/corda/tools/shell/InteractiveShellJavaTest.java
+++ b/tools/shell/src/test/java/net/corda/tools/shell/InteractiveShellJavaTest.java
@@ -152,7 +152,7 @@ public class InteractiveShellJavaTest {
 
     @Test
     public void flowStartWithComplexTypes() throws InteractiveShell.NoApplicableConstructor {
-        check("amount: £10", "10.00 GBP", FlowA.class);
+        check("amount: \u00A310", "10.00 GBP", FlowA.class);
     }
 
     @Test
--- a/build.gradle
+++ b/build.gradle
@@ -160,6 +160,7 @@ allprojects {
 
     tasks.withType(JavaCompile) {
         options.compilerArgs << "-Xlint:unchecked" << "-Xlint:deprecation" << "-Xlint:-options" << "-parameters"
+        options.encoding = 'UTF-8'
     }
 
     tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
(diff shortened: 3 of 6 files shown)


tools/shell/src/test/java/net/corda/tools/shell/InteractiveShellJavaTest.java
```diff
@@ -152,7 +152,7 @@ public void flowStartSimple() throws InteractiveShell.NoApplicableConstructor {
 
     @Test
     public void flowStartWithComplexTypes() throws InteractiveShell.NoApplicableConstructor {
-        check("amount: £10", "10.00 GBP", FlowA.class);
+        check("amount: \u00A310", "10.00 GBP", FlowA.class);
```
mnesbit: This looks wrong unicode symbol

mnesbit: OK I understand this is a mixture of a literal 10 and the pound sign, but I don't think this is the actual problem here. I think somebody has messed up the Jackson serializer again.

vkolomeyko: As part of this PR I wrote a new (simple) unit test `AmountParsingTest.java` which exposes the problem.
No Jackson or any other funky business... looks like Java not interpreting "£" symbol in the source code the way we would have hoped.

mnesbit: Pound symbol is in the normal ascii 7-bit range, so this is NOT the problem.

mnesbit: OK it is in the 8-bit range and the raw bytes of the original test contain C2 A3 bytes which is the UTF-8 of the pound sign. So the question what encoding is being passed to the compiler?

mnesbit: http://blog.ehrnhoefer.com/2015-02-28-gradle-sourcefile-encoding/ perhaps?

vkolomeyko: But I wonder if we should go down this Gradle source encoding root.
At the moment it is only a problem for some Java test classes (Kotlin seems to be fine).

mnesbit: According to Chris Rankin Kotlin is always UTF-8 source files and has no flag. Javac defaults to platform encoding which is UTF-8 on linux and MacOS, but CP-1252 on Windows. Plus IntelliJ seems to always be UTF-8 even on windows.

chrisr3: [This 'ere](https://stackoverflow.com/questions/45564450/how-to-set-kotlin-source-encoding-in-gradle) is the source of my wisdom.

mikehearn:
Can we please just make Windows use UTF-8? I'd rather not downgrade the code when the correct solution is to upgrade Windows to the modern Unicode era. It looks to be 3 lines of Gradle.

core/build.gradle
```diff
@@ -162,3 +162,7 @@ scanApi {
 publish {
     name jar.baseName
 }
+
+compileTestJava {
+    options.encoding = 'UTF-8'
```
mnesbit: I think Mike was asking for us to apply this globally to normal+test profiles of Java.

vkolomeyko:
I have pushed changes this morning which changes encoding for Java files to "UTF-8".
It had to be be per project as changes on the root project do not seem to propagate correctly.

chrisr3:
So are you saying that adding this line to `allprojects` doesn't work?
```diff
diff --git a/build.gradle b/build.gradle
index dfb4c7f52b..76fef2e630 100644
--- a/build.gradle
+++ b/build.gradle
@@ -160,6 +160,7 @@ allprojects {
 
     tasks.withType(JavaCompile) {
         options.compilerArgs << "-Xlint:unchecked" << "-Xlint:deprecation" << "-Xlint:-options" << "-parameters"
+        options.encoding = 'UTF-8'
     }
 
     tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
```

vkolomeyko:
Sorry @chrisr3 - my bad, I tried the same option outside of `allprojects`. Doh!
Your proposed change works as intended.

PR #3716, opened 2018-07-30
Matchers

distributedleetravis:
Moving some common [hamkrest](https://github.com/npryce/hamkrest) matchers into the shared testing module, using hamkrest to express a test case in `SwapIdentitiesFlowTests` differently.

Why use hamkrest? The basic idea is that test expectations, and failure reasons, should be reasonably self-describing based on the criteria defined in code. For example, the expectation defined in the test `issue key` looks like this in code:

```kotlin
assert.that(
            aliceNode.services.startFlow(SwapIdentitiesFlow(bob)),
            willReturn(
                hasOnlyEntries(
                    alice to allOf(
                        !equalTo<AbstractParty>(alice),
                        aliceNode.resolvesToWellKnownParty(alice),
                        aliceNode.holdsOwningKey(),
                        !bobNode.holdsOwningKey()
                    ),
                    bob to allOf(
                        !equalTo<AbstractParty>(bob),
                        bobNode.resolvesToWellKnownParty(bob),
                        bobNode.holdsOwningKey(),
                        !aliceNode.holdsOwningKey()
                    )
                )
            )
        )
```

And describes itself like this:

```
a value that is a flow that will return with a value that meets all of the criteria:
	is a map of size 2
	is a map containing the entries:
		CN=1, O=Alice Corp, L=Madrid, C=ES: meets all of the criteria:
			is not equal to CN=1, O=Alice Corp, L=Madrid, C=ES
			is resolved by "CN=1, O=Alice Corp, L=Madrid, C=ES" to well-known party "CN=1, O=Alice Corp, L=Madrid, C=ES"
			has an owning key which is held by CN=1, O=Alice Corp, L=Madrid, C=ES
			has an owning key which is not held by CN=2, O=Bob Plc, L=Rome, C=IT
		CN=2, O=Bob Plc, L=Rome, C=IT: meets all of the criteria:
			is not equal to CN=2, O=Bob Plc, L=Rome, C=IT
			is resolved by "CN=2, O=Bob Plc, L=Rome, C=IT" to well-known party "CN=2, O=Bob Plc, L=Rome, C=ITS"
			has an owning key which is held by CN=2, O=Bob Plc, L=Rome, C=IT
			has an owning key which is not held by CN=1, O=Alice Corp, L=Madrid, C=ES
```

A mismatch is described like this:

```
but succeeded with value that did not meet criteria:
	had entries which did not meet criteria:
		CN=2, O=Bob Plc, L=Rome, C=IT: did not meet criteria:
			was resolved to CN=1, O=Alice Corp, L=Madrid, C=ES
```

This makes it easier to understand quickly the expected outcome of a test, and understand the reasons for test failures.


confidential-identities/src/test/kotlin/net/corda/confidential/SwapIdentitiesFlowTests.kt
```diff
@@ -1,113 +1,161 @@
...
-        // Verify that the nodes have the right anonymous identities
-        assertTrue { aliceAnonymousIdentity.owningKey in aliceNode.services.keyManagementService.keys }
-        assertTrue { bobAnonymousIdentity.owningKey in bobNode.services.keyManagementService.keys }
-        assertFalse { aliceAnonymousIdentity.owningKey in bobNode.services.keyManagementService.keys }
-        assertFalse { bobAnonymousIdentity.owningKey in aliceNode.services.keyManagementService.keys }
-
-        mockNet.stopNodes()
+        assert.that(
+            aliceNode.services.startFlow(SwapIdentitiesFlow(bob)),
+            willReturn(
```
sollecitom: Not sure I like how this reads: "will return has only entries".

distributedleetravis:
The immediate purpose is to make our own tests easier to express. If after dogfooding this for a while we think we can stabilise a good set of generically useful Corda Matchers, we should consider doing so and making them public.

In that case, it might make sense to wrap the hamkrest (Kotlin) matchers as hamcrest (Java) matchers, so that they can be comfortably used from Java too.

The reason for not writing them as hamcrest matchers in the first place is that hamcrest matchers can be cumbersome to implement. Hamkrest makes fairly intelligent use of some Kotlin-only language features, and in general there seem to be far fewer headaches with generics than hamcrest tends to throw up.

PR #3717, opened 2018-07-30
ENT-2334 PoC for the CRL custom provisioning

mkit:
This PR is PoC for the new way we provide the CRLs. The idea is based on implementing our own CertStore that then is injected into the revocation checking process of the SSL stack.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/revocation/RevocationConfig.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/revocation/RevocationConfig.kt
@@ -1,0 +1,8 @@
+package net.corda.nodeapi.internal.revocation
+
+import java.security.cert.CertStore
+
+/**
+ * Holds the configuration related to the certificate revocation.
+ */
+data class RevocationConfig(val crlCheckSoftFail: Boolean = true, val certStores: Set<CertStore> = emptySet())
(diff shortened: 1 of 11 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/revocation/RevocationConfig.kt
```diff
@@ -0,0 +1,8 @@
+package net.corda.nodeapi.internal.revocation
+
+import java.security.cert.CertStore
+
+/**
+ * Holds the configuration related to the certificate revocation.
+ */
+data class RevocationConfig(val crlCheckSoftFail: Boolean = true, val certStores: Set<CertStore> = emptySet())
```
mnesbit: An interface would be my preference, since this is very likely to change.

patrickkuo:
just an idea.

Instead of changing the trust manager and having our own CRL store implementation, will it be simpler to set the CRL url in the certificate to `http://localhost:<<Port>>` and have the node start up a CRL endpoint. The node can serve a CRL endpoint cached from the network map.

one benefit of this is the SSL engine won't need to reach out to the CRL address, which can be a problem if the node is behind firewall, also we can allow user to choose which CRL to serve in their node.

mnesbit:
localhost is not well defined and I think this will probably cause pain to configure as it requires the port to be shareable across multiple nodes on the same box. However, I also would prefer we remove the CRL URL from the certificates completely and provide the CRL in the network map document. Passing in a custom PKIXValidator would probably let is build up arbitrarily complicate management here/

node-api/src/main/kotlin/net/corda/nodeapi/internal/revocation/RevocationConfig.kt
mkit: 👍

shamsasari:
@mkit what's the status of this PR? Each close or work to get it merged in.

shamsasari:
PoC is marked as done, so closing.

PR #3720, opened 2018-07-30
CORDA-916: Add registerFlowFactory method to public test API

distributedleetravis:
An addition (non-breaking change) to the public test API (`MockNet` and `StartedMockNode`) to support a customer use case (see [CORDA-916](https://r3-cev.atlassian.net/browse/CORDA-916) and the associated [StackOverflow query](https://stackoverflow.com/questions/48166626/how-can-an-acceptor-flow-in-corda-be-isolated-for-unit-testing/)).

This PR adds a method, `StartedMockNode.registerFlowFactory` to the `StartedMockNode` API, and includes tests in Java and Kotlin, based on the original use case, which show this method in use.

It also adds a Kotlin-only extension method to `StartedMockNode` which makes use of Kotlin language features (reified type-parameters and lambda parameters) to provide a simplified form of the same functionality for Kotlin users.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
@@ -186,9 +186,9 @@ class StartedMockNode private constructor(private val node: TestStartedNode) {
      * Register an [InitiatedFlowFactory], to control relationship between initiating and receiving flow classes
      * explicitly on a node-by-node basis.
      */
-    fun <F : FlowLogic<*>> registerFlowFactory(initiatingFlowClass: Class<out FlowLogic<*>>,
-                                               flowFactory: InitiatedFlowFactory<F>,
-                                               initiatedFlowClass: Class<F>): CordaFuture<F> =
+    fun <F : FlowLogic<*>> registerResponderFlow(initiatingFlowClass: Class<out FlowLogic<*>>,
+                                                 flowFactory: InitiatedFlowFactory<F>,
+                                                 initiatedFlowClass: Class<F>): CordaFuture<F> =
             node.registerFlowFactory(initiatingFlowClass, flowFactory, initiatedFlowClass, true)
                     .toFuture()
 }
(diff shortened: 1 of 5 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -177,8 +181,30 @@ class StartedMockNode private constructor(private val node: TestStartedNode) {
             statement()
         }
     }
+
+    /**
+     * Register an [InitiatedFlowFactory], to control relationship between initiating and receiving flow classes
+     * explicitly on a node-by-node basis.
+     */
+    fun <F : FlowLogic<*>> registerFlowFactory(initiatingFlowClass: Class<out FlowLogic<*>>,
```
joeldudleyr3: I think this method name is confusing. Would it be better as `registerResponderFlow`?

testing/node-driver/src/test/java/net/corda/testing/node/TestResponseFlowInIsolationInJava.java
```diff
@@ -0,0 +1,122 @@
...
+    @Rule
+    public final ExpectedException exception = ExpectedException.none();
+
+    @Test
+    public void test() throws Exception {
+        // This method returns the Responder flow object used by node B.
+        Future<Responder> initiatedResponderFlowFuture = b.registerFlowFactory(
+                // We tell node B to respond to BadInitiator with Responder.
+                // We want to observe the Responder flow object to check for errors.
+                BadInitiator.class, flowFactory, Responder.class);
```
joeldudleyr3: Where does `flowFactory` come from here?

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
```diff
@@ -177,8 +181,30 @@ class StartedMockNode private constructor(private val node: TestStartedNode) {
             statement()
         }
     }
+
+    /**
+     * Register an [InitiatedFlowFactory], to control relationship between initiating and receiving flow classes
+     * explicitly on a node-by-node basis.
```
joeldudleyr3: Can you add Kdocs for the parameters?

joeldudleyr3: Can you document the behaviour if you register a flow that already has a registered responder?

testing/node-driver/src/test/java/net/corda/testing/node/TestResponseFlowInIsolationInJava.java
distributedleetravis: It's declared further down in the class. I'll move it to the top of the class and uppercase it to make it clear that it's a constant.

PR #3729, opened 2018-07-31
CORDA-1861: Remove default value to make createUnstartedNode() unambiguous.

chrisr3:
There are two variants of `MockNetwork.createUnstartedNode(...)`, and their default parameters make
the following call ambiguous:
```kotlin
val node = mockNetwork.createUnstartedNode(DUMMY_BANK_A_NAME)
```
Remove the default value for `additionalCordapps` to align this API with `MockNetwork.createNode(...)`.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/MockNetwork.kt
@@ -411,7 +411,7 @@ open class MockNetwork(
                             forcedID: Int? = null,
                             entropyRoot: BigInteger = BigInteger.valueOf(random63BitValue()),
                             configOverrides: (NodeConfiguration) -> Any? = {},
-                            additionalCordapps: Set<TestCorDapp> = emptySet()): UnstartedMockNode {
+                            additionalCordapps: Set<TestCorDapp>): UnstartedMockNode {
         val parameters = MockNodeParameters(forcedID, legalName, entropyRoot, configOverrides, additionalCordapps)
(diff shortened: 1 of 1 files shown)


shamsasari:
I presume the API checker isn't checking default values. I believe we're aiming for source compatibility for the test APIs. This change breaks that as developers will now be required to fill in the non-default value.

shamsasari:
Of course this doesn't affect Java developers, but it will those using Kotlin.

chrisr3:
So "broken by design"?

shamsasari:
How about a no-params overload?

chrisr3:
Come to think of it:
```kotlin
fun createUnstartedNode(legalName: CordaX500Name? = null,
                        forcedID: Int? = null,
                        entropyRoot: BigInteger = BigInteger.valueOf(random63BitValue()),
                        configOverrides: (NodeConfiguration) -> Any? = {},
                        extraCordappPackages: List<String> = emptyList()): UnstartedMockNode

fun createUnstartedNode(legalName: CordaX500Name? = null,
                        forcedID: Int? = null,
                        entropyRoot: BigInteger = BigInteger.valueOf(random63BitValue()),
                        configOverrides: (NodeConfiguration) -> Any? = {},
                        additionalCordapps: Set<TestCorDapp>): UnstartedMockNode
```
Unless I'm reading this wrong, the only way to disambiguate these two prototypes _is_ to supply an explicit value for either `additionalCordapps` or `extraCordappPackages`. In other words, it's already impossible to write the case that would otherwise be broken by this change.

PR #3745, opened 2018-08-06
ENT-2359: Complete the nodeReady future if there are already node-infos in the db on startup

shamsasari:
Otherwise PersistentNetworkMapCache only completed the future if a node-info was added, which may not occur after a node restart and thus prevented the SMM from registering its listener with the messaging layer.

Further, the future is only completed if the node-info is other than the node's own one.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
+++ b/node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
@@ -3,7 +3,6 @@ package net.corda.node.services.network
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.crypto.SecureHash
 import net.corda.core.crypto.toStringShort
-import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.identity.PartyAndCertificate
@@ -12,7 +11,6 @@ import net.corda.core.internal.concurrent.openFuture
 import net.corda.core.messaging.DataFeed
 import net.corda.core.node.NodeInfo
 import net.corda.core.node.NotaryInfo
-import net.corda.core.node.services.IdentityService
 import net.corda.core.node.services.NetworkMapCache.MapChange
 import net.corda.core.node.services.PartyInfo
 import net.corda.core.serialization.SingletonSerializeAsToken
@@ -22,7 +20,6 @@ import net.corda.core.utilities.contextLogger
 import net.corda.core.utilities.debug
 import net.corda.node.internal.schemas.NodeInfoSchemaV1
 import net.corda.node.services.api.NetworkMapCacheBaseInternal
-import net.corda.node.services.api.NetworkMapCacheInternal
 import net.corda.node.utilities.NonInvalidatingCache
 import net.corda.nodeapi.internal.persistence.CordaPersistence
 import net.corda.nodeapi.internal.persistence.bufferUntilDatabaseCommit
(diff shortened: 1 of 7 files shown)


node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -202,7 +173,9 @@ open class PersistentNetworkMapCache(private val database: CordaPersistence) : S
                 logger.info("Previous node was identical to incoming one - doing nothing")
             }
         }
-        _nodeReady.set(null)
+        if (node.legalIdentities[0].name != myLegalName) {
```
rick-r3: Just a thought: Do you think there would be an issue with excluding “self”?  It seems like a good idea, but then if there’s a "network" of one node…. it won’t start.  I'm surprised we didn't hit a test like that.  I think we go ahead.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -105,6 +67,15 @@ open class PersistentNetworkMapCache(private val database: CordaPersistence) : S
 
     fun start(notaries: List<NotaryInfo>) {
         this.notaries = notaries
+        val otherNodeInfoCount = database.transaction {
+            session.createQuery(
+                    "select count(*) from ${NodeInfoSchemaV1.PersistentNodeInfo::class.java.name} n join n.legalIdentitiesAndCerts i where i.name != :myLegalName")
```
rick-r3: Ordinarily I'd use the CriteriaAPI to a void parsing the text.  Given this is only executed once, I don't think it matters.  

What we should do though is use `n.${NodeInfoSchemaV1.PersistentNodeInfo::legalIdentitiesAndCerts.name}` in place of `n.legalIdentitiesAndCerts` and `i.${DBPartyAndCertificate::name.name}` in place of `i.name`.  Think I got that right....!

It makes it more refactor (well, rename) proof.

node/src/test/kotlin/net/corda/node/services/network/PersistentNetworkMapCacheTest.kt
```diff
@@ -0,0 +1,153 @@
...
+import net.corda.nodeapi.internal.persistence.DatabaseConfig
+import net.corda.testing.core.*
+import net.corda.testing.node.MockServices.Companion.makeTestDataSourceProperties
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions.assertThatIllegalArgumentException
+import org.junit.After
+import org.junit.Rule
+import org.junit.Test
+
+class PersistentNetworkMapCacheTest {
```
rick-r3: Why move this out of the integration test suite where it was testing against different database vendors?

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
shamsasari: A network of one is a bit unless so I don't think we should worry about that. Plus, the self node-info is *always* added, which really brings up the question of whether we even need this `nodeReady` future at all if we accept self as well.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
shamsasari: That doesn't make the query very readable, and we have the tests so any breakage due to refactoring will be caught.

I personally find the CriteriaAPI very hard to use and not a fan, and JPQL is easier to understand due to its similarity to SQL.

node/src/test/kotlin/net/corda/node/services/network/PersistentNetworkMapCacheTest.kt
shamsasari: Ah!

rick-r3:
Apparently my brain is not a compiler and it didn't spot the compilation error!

PR #3749, opened 2018-08-07
Add jitter to retries

thschroeter:


node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -590,10 +590,11 @@ class SingleThreadedStateMachineManager(
     private fun scheduleTimeoutException(flow: Flow, retryCount: Int): ScheduledFuture<*> {
         return with(serviceHub.configuration.flowTimeout) {
             val timeoutDelaySeconds = timeout.seconds * Math.pow(backoffBase, retryCount.toDouble()).toLong()
+            val jitteredDelaySeconds = maxOf(1L ,timeoutDelaySeconds/2 + (Random().nextDouble() * timeoutDelaySeconds/2).toLong())
```
adagys: could use Math.random() instead of instantiating the object every time. Also missing a space after the comma.

PR #3762, opened 2018-08-09
ENT-2168: Add a shell command to check for an existing transaction

adagys:
When a double-spend occurs the notary returns the hash of the consuming
transaction id. I've added a 'hash-lookup' shell command that matches
any recorded transactions on the node against this id hash to determine
whether the state has been consumed by this node (that could happen in certain race conditions).


tools/shell/src/main/java/net/corda/tools/shell/HashLookupShellCommand.java
```diff
@@ -0,0 +1,59 @@
...
+public class HashLookupShellCommand extends InteractiveShellCommand {
+    private static Logger logger = LoggerFactory.getLogger(HashLookupShellCommand.class);
+
+    @Command
+    @Man("Checks if a transaction matching a specified Id hash value is recorded on this node.\n\n" +
+            "This is mainly intended to be used for troubleshooting notarisation issues when a\n" +
+            "state is claimed to be already consumed by another transaction.\n\n" +
+            "Example usage: hash-lookup E470FD8A6350A74217B0A99EA5FB71F091C84C64AD0DE0E72ECC10421D03AAC9"
+    )
+    public void main(@Usage("A hexadecimal SHA-256 hash value representing the hash of a transaction Id") @Argument(unquote = false) String txIdHash) {
```
tlil: *hash* value representing the *hash* #nitpick

adagys: ha, not sure how to paraphrase this, it's a hash of a hash.  "..a hashed transaction id" ?

tlil: Yeah, that's better I guess :)

PR #3763, opened 2018-08-09
Rog data distribution groups

roger-that-dev:
Added a new module for data distribution groups and worked on a PoC. This PR is just to review the PoC. Do not merge!

ReferenceDataUpdateRequest.kt is not part of the data distribution group flows. It's a separate flow to request an update to a reference state from the maintainer party.

### DIFF
--- a/data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
+++ b/data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
@@ -2,13 +2,37 @@
 
 ## Stage one.
 
-Use the confidential identities infrastructure to create a key and
-certificate for the group. Store the group as a ContractState in the
-vault.
+* Groups currently have a name and key. The name is just a string.
 
-The assumption at stage one is that only the creator of the group can
-write to it.
+* the confidential identities infrastructure added to the key management
+  service is used to create a key and certificate for the group.
 
-Multiple states for the same data distribution group can be stored in
-the vault. The issue here, though, is that the PartyAndCertificate is
-duplicated for each one.
+* The certificate is stored by the creator and sent to any invitees and
+  so on.
+
+* Groups are stored as ContractStates in the vault.
+
+* The assumption at stage one is that only the creator of the group can
+  write to it. Only the founder can add transactions to the group.
+
+* Multiple states for the same data distribution group can be stored in
+  the vault. Each state stores a bi-lateral group relationship apart
+  from the initial group state which just contains the group founder.
+
+* Inviting the same node twice can happen but doesn't have any impact
+  as the flows deal in Sets of parties.
+
+* Transactions are signed with the group key by the founder.
+
+* Invitees check the signature on new transactions added to the group.
+  They store all transactions anyway as currently the pre-packaged
+  SendTransactionFlow and ReceiveTransactionFlows are used. This needs
+  to change going forward.
+
+* Transactions are propagated on to neighbours until there are no new
+  neighbours to send the transactions to.
+
+* There is currently no persistence of transaction which have been sent
+  to a group.
+
+* Only signed transactions can be added to a group.
--- a/data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
+++ b/data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
@@ -1,0 +1,14 @@
+# Staged approach for introducing DDGs
+
+## Stage one.
+
+Use the confidential identities infrastructure to create a key and
+certificate for the group. Store the group as a ContractState in the
+vault.
+
+The assumption at stage one is that only the creator of the group can
+write to it.
+
+Multiple states for the same data distribution group can be stored in
+the vault. The issue here, though, is that the PartyAndCertificate is
+duplicated for each one.
--- a/data-distribution-groups/src/main/kotlin/net/corda/groups/contracts/Group.kt
+++ b/data-distribution-groups/src/main/kotlin/net/corda/groups/contracts/Group.kt
@@ -1,0 +1,30 @@
+package net.corda.groups.contracts
+
+import net.corda.core.contracts.CommandData
+import net.corda.core.contracts.Contract
+import net.corda.core.contracts.ContractState
+import net.corda.core.identity.Party
+import net.corda.core.identity.PartyAndCertificate
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.transactions.LedgerTransaction
+import java.security.PublicKey
+
+class Group : Contract {
+    companion object {
+        @JvmStatic
+        val contractId = "net.corda.groups.contracts.Group"
+    }
+
+    class Create : CommandData
+    class Invite : CommandData
+
+    override fun verify(tx: LedgerTransaction) = Unit
+
+    @CordaSerializable
+    data class Details(val name: String, val key: PublicKey)
+
+    @CordaSerializable
+    data class DetailsWithCert(val details: Details, val cert: PartyAndCertificate)
+
+    data class State(val details: Details, override val participants: List<Party>) : ContractState
+}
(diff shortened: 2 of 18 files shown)


dazraf:
will there be permissioning / voting for adding a member to the d-group?

roger3cev:
Currently there is only support for existing members of a group to invite new members. However, we are open to other approaches as invite only doesn't cater for all scenarios.

For the time being only the founder of the group can write to it. The founder can add readonly members to the group and those readonly members can add other readonly members.

For the MVP, nodes will be able to invite others as read/write members of the group.

roger3cev:
@dazraf currently there's no support for leaving :) However, if I were to add support and the founder left, then the group would be left to be useless as no-one else can add transactions to it! This might be desired behaviour depending on the rationale for the group. Probably more correctly though, if the only read/write member of the group leaves, then an operation should propagate through the group to exit all of the states representing that group.

Re: proposal flows. Yes, using an abstraction here makes sense. However I'm not sure how voting would work with the current construct. group members are only aware of those who invited them and those they invited. There's no omniscient member who can see all other members. Collecting votes would be problematic. We'd need to build out flows similar to the ones we used in project Ubin. It _can_ be done but will take a fair bit of effort. 

What do you mean by:
> Also, we're wondering if it would be possible for the datagroup to be a synthetic AbstractParty ...

Cheers

data-distribution-groups/src/main/kotlin/net/corda/groups/contracts/Group.kt
```diff
@@ -0,0 +1,50 @@
...
+import net.corda.core.schemas.PersistentState
+import net.corda.core.schemas.QueryableState
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.transactions.LedgerTransaction
+import net.corda.groups.schemas.GroupSchemaV1
+import java.security.PublicKey
+
+/** Just a stub for now. */
+// TODO: Update and add KDocs.
+class Group : Contract {
```
rick-r3: Some thoughts:

1. Make this abstract / open
2. Have a single subclass for now which is `SingleOwnerGroupWithNoLeave`
3. Add a `Leave` command, and leave flows.  Probably also want a `ChangeRole` or some such for making people owners or demoting owners.  All should be governed by the `verify`  of the contracts and in the one (current) implementation subclass, could throw unsupported.
4. This means custom "group" contracts can govern membership as they see fit, without needing any new flows, I think.  

Thus the concept and flows are defined, but new membership / governance models can sprout from new contracts.

WDYT?

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+* Groups currently have a name and key. The name is just a string.
+
+* the confidential identities infrastructure added to the key management
+  service is used to create a key and certificate for the group.
+
+* The certificate is stored by the creator and sent to any invitees and
+  so on.
+
+* Groups are stored as ContractStates in the vault.
```
mikehearn: should they be? which features of states do they need?

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+* The assumption at stage one is that only the creator of the group can
+  write to it. Only the founder can add transactions to the group.
```
mikehearn: Can we extend this to "anyone with the key can write"? To allow explicitly for the possibility of key sharing?

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+* Multiple states for the same data distribution group can be stored in
+  the vault. Each state stores a bi-lateral group relationship apart
+  from the initial group state which just contains the group founder.
+
+* Inviting the same node twice can happen but doesn't have any impact
```
mikehearn: Should such an invite flow throw an exception?

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+* Invitees check the signature on new transactions added to the group.
+  They store all transactions anyway as currently the pre-packaged
+  SendTransactionFlow and ReceiveTransactionFlows are used. This needs
+  to change going forward.
+
+* Transactions are propagated on to neighbours until there are no new
+  neighbours to send the transactions to.
+
+* There is currently no persistence of transaction which have been sent
```
mikehearn: ?

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
```diff
@@ -0,0 +1,38 @@
...
+  to a group.
+
+* Only signed transactions can be added to a group.
```
mikehearn: Isn't this part implied by the above statement that the founder signs the tx?

mikehearn (changes requested):
I haven't read the code, just commented on the readme file.

I wonder if before we proceed with this we should tackle storing non-participatory states in the vault properly? That would fix issues with observer states and this too. Feels like a prereq.

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
```diff
@@ -0,0 +1,38 @@
+# Staged approach for introducing DDGs
+
+## Stage one.
+
+* Groups currently have a name and key. The name is just a string.
```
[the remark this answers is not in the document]
roger3cev: The idea behind name was so that users can choose a group by a human readable identifier. The public key is the unique identifier.

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
roger3cev: This is how we designed the obligation graph for Ubin. It worked quite well. It also allows us to re-use all the infrastructure around vault querying. The alternative is to store the data in manually generated SQL tables - requires more effort.

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
roger3cev: In the design I'm writing up, key sharing is not allowed. Instead, those that are invited with read/write access will be asked to generate a key pair. The inviter will sign the newly generated public key of the invitee, thus forming a key chain back to the key of the founder of the group (which may or may not be signed by the founder's legal identity key).

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
roger3cev: It can do!

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
roger3cev: Whoops not sure what happened here. Persistence will be added later, as it will take a bit of work.

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
roger3cev: Yes. These notes are just some initial thoughts... Probably tomorrow or Wednesday I'll follow up with a design document.

data-distribution-groups/src/main/kotlin/net/corda/groups/contracts/Group.kt
roger3cev: This is a great idea! I'll incorporate it into the design document - thanks!

data-distribution-groups/src/main/kotlin/net/corda/groups/README.md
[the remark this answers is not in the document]
mikehearn: Sure, maybe "Title" then? The word "name" has quite a few connotations in programming that don't apply for the way this field is actually used.

vkolomeyko:
Last comment on this PR is from 13-Aug-2018 and the branch is in conflicted state.
Is there still hope to have this PR merged?

roger3cev:
@vkolomeyko It's ok we can close this PR - it was raised just so someone can review it.

PR #3773, opened 2018-08-13
CORDA-1894 Remove if condition from service hub that prevents vault logic executing.

rick-r3:
Vault logic should be in the vault.  And it stops some code in CE from executing...


josecoll:
I thought this was by Design, as per the definition of `NONE` enum status:

```
    /** The received transaction is not sent to the vault at all. This is used within transaction resolution. */
    NONE,
```

PR #3777, opened 2018-08-13
Azure OS testnet instructions - minor changes

joeldudleyr3:


Clintonio:
Does this mention cloud templates anywhere? If not it should so that we have more people using those if they somehow end up on this doc.

PR #3779, opened 2018-08-13
Fix duplicate index declaration in Cash schema.

josecoll:
This PR fixes the annoying Exception stacks generated in tests:

```
Caused by: org.h2.jdbc.JdbcSQLException: Index "CCY_CODE_IDX" already exists; SQL statement:
create index ccy_code_idx on contract_cash_states_v1 (ccy_code) [42111-197]
Caused by: org.h2.jdbc.JdbcSQLException: Index "PENNIES_IDX" already exists; SQL statement:
create index pennies_idx on contract_cash_states_v1 (pennies) [42111-197]
```


roger3cev:
Boom!

mikehearn:
Couple of questions about this:

1. Why does this work? Is there an implicit index defined that we're not using?
2. Do we need to prepare database migration scripts for it as part of the upgrade?

PR #3780, opened 2018-08-13
Corda-1869 serialisation refactor

distributedleetravis:
A few small sets of changes related to https://r3-cev.atlassian.net/browse/CORDA-1869, mainly the result of tidying up code as I was reading it in order to understand it better.

The first set of changes is a fairly trivial refactoring of the `inferTypeVariables` method, aimed at improving readability.

The second set of changes is the result of changing the extension method `Type.asClass()` so that its return type is `Class<*>` rather than `Class<*>?`, because it can never return `null` and was only suggesting that it might because Kotlin can't ensure that the list of possible cases given to its `when` clause is exhaustive. Instead of returning `null` on the `else` case, we offer to throw an `UnsupportedOperationException`. Having done this, it was then possible to eliminate a lot of unnecessary null checks and not-null assertions.

The third set of changes is a refactor of the `FingerPrinter` interface and `SerializerFingerPrinter` implementation. Three changes are involved:

1) We remove the `setFactory` method from `FingerPrinter`, and insist that `SerializerFingerPrinter` is initialised with the factory it is going to need. We can now also insist that this factory is never `null`, without resorting to `lateinit`.
2) We break up some deeply-nested logic to make the execution flow a bit easier to follow.
3) We pull out the working state of the fingerprinter into a set of fields in their own class, `FingerPrinterState`, instead of passing it down in pieces through a chain of method calls. Because the recursion never needs to backtrack, we don't need to be able to pull previous state values out of the stack - it's safe just to mutate the current state variables freely as we go along. The `debugIndent` parameter, which is never used anywhere, disappears altogether.

Finally, there is some general tidying and refactoring of code around `PropertyDescriptor`s, including a minor bugfix (we should be checking the first parameter type, not the return type, of setter methods when deciding which one is the least generic).

### DIFF
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
@@ -7,6 +7,7 @@ import net.corda.serialization.internal.amqp.MethodClassifier.*
 import java.lang.reflect.Field
 import java.lang.reflect.Method
 import java.lang.reflect.Type
+import java.util.*
 
 /**
  * Encapsulates the property of a class and its potential getter and setter methods.
@@ -126,11 +127,11 @@ private fun Sequence<PropertyNamedMethod>.withValidSignature() = filter { it.has
 // Group methods by name and classifier, picking the method with the least generic signature if there is more than one
 // of a given name and type.
 private fun Sequence<PropertyNamedMethod>.byNameAndClassifier(fieldNames: Set<String>): Map<String, Map<MethodClassifier, Method>> {
-    val result = mutableMapOf<String, MutableMap<MethodClassifier, Method>>()
+    val result = mutableMapOf<String, EnumMap<MethodClassifier, Method>>()
 
     forEach { (fieldName, classifier, method) ->
         result.compute(getPropertyName(fieldName, fieldNames)) { _, byClassifier ->
-            (byClassifier ?: mutableMapOf()).merge(classifier, method)
+            (byClassifier ?: EnumMap(MethodClassifier::class.java)).merge(classifier, method)
         }
     }
 
@@ -138,7 +139,7 @@ private fun Sequence<PropertyNamedMethod>.byNameAndClassifier(fieldNames: Set<St
 }
 
 // Merge the given method into a map of methods by method classifier, picking the least generic method for each classifier.
-private fun MutableMap<MethodClassifier, Method>.merge(classifier: MethodClassifier, method: Method): MutableMap<MethodClassifier, Method> {
+private fun EnumMap<MethodClassifier, Method>.merge(classifier: MethodClassifier, method: Method): EnumMap<MethodClassifier, Method> {
     compute(classifier) { _, existingMethod ->
         if (existingMethod == null) method
         else when (classifier) {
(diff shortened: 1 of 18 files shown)


serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializationHelper.kt
```diff
@@ -75,145 +67,13 @@ fun constructorForDeserialization(type: Type): KFunction<Any>? {
...
-                        // take into account those constructor properties that don't directly map to a named
-                        // property which are, by default, already added to the map
-                        classProperties.computeIfAbsent(groups[2]!!.value) { PropertyDescriptor() }
-
-                properties.apply {
-                    when (groups[1]!!.value) {
-                        "set" -> {
-                            if (func.parameterCount == 1) {
-                                if (setter == null) setter = func
-                                else if (TypeToken.of(setter!!.genericReturnType).isSupertypeOf(func.genericReturnType)) {
```
distributedleetravis: Tiny bug: we ought to be comparing the first generic parameter type, not the return type.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
```diff
@@ -0,0 +1,190 @@
...
+
+// Map a sequence of fields by field name.
+private fun Sequence<Field>.byFieldName() = map { it.name to it }.toMap()
+
+// Select only those methods that are public (and are not the "getClass" method)
+private fun Sequence<Method>.thatArePublic() = filter { it.isPublic && it.name != "getClass" }
+
+// Select only those methods that are isX/getX/setX methods
+private fun Sequence<Method>.thatArePropertyMethods() = map { method ->
+    propertyMethodRegex.find(method.name)?.let {
```
fenryka: name implicit param

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializationHelper.kt
fenryka: good catch

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializationHelper.kt
```diff
@@ -235,174 +95,132 @@ internal fun <T : Any> propertiesForSerializationFromConstructor(
...
+
+    // We will already have disambiguated getA for property A or a but we still need to cope
+    // with the case we don't know the case of A when the parameter doesn't match a property
+    // but has a getter
+    val matchingProperty = classProperties[name] ?: classProperties[name.capitalize()]
+    ?: throw AMQPNotSerializableException(type,
+            "Constructor parameter - \"$name\" -  doesn't refer to a property of \"$clazz\"")
+
+    // If the property has a getter we'll use that to retrieve it's value from the instance, if it doesn't
+    // *for *know* we switch to a reflection based method
```
fenryka: *now* not *know* (which to be fair is probably my typo but... you're moving the code :)

mikehearn:
This PR should probably be re-reviewed now you have added more commits, if they aren't addressing review comments. Better to get this PR merged and then start a new one, I think.

fenryka:
Remember to squash :)

PR #3790, opened 2018-08-15
CORDA-1905: Extend JSON deserialisation to handle Amount<T> for any T.

chrisr3:
Stop assuming that every `Amount` is actually `Amount<Currency>` by allowing Jackson to inspect the bean types at runtime. Falls back to using `Currency` if the type cannot be determined.

### DIFF
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
@@ -413,10 +413,11 @@ object JacksonSupport {
         }
     }
 
-    @Deprecated("This is an internal class, do not use")
+    @Deprecated("No longer used",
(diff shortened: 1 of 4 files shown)


client/jackson/src/main/kotlin/net/corda/client/jackson/JacksonSupport.kt
```diff
@@ -413,10 +413,11 @@ object JacksonSupport {
         }
     }
 
-    @Deprecated("This is an internal class, do not use")
+    @Deprecated("No longer used",
+            replaceWith = ReplaceWith("Corda's internal AmountDeserializer and TokenDeserializer classes"))
```
mikehearn: Please don't specify replaceWith as natural language - the IDE parses the contents to auto-fix code. Put this advice in the deprecation message.

PR #3794, opened 2018-08-16
CORDA-1906: Add Servlet 3.x implementation into the WebServer.

chrisr3:
Corda includes the Servlet 3.x APIs, but we only include the Servlet 2.0 implementation inside the WebServer. This prevents anyone from writing (say) an asynchronous servlet inside a CorDapp.

Include Jersey's Servlet 3.x implementation, which also includes the 2.0 implementation.


josecoll:
@chrisr3 is merely adding an implementation of an API we already export.
Until we really do deprecate (eg. remove the webserver module) this seems a sensible, non-code,  change to me.

chrisr3:
This isn't new functionality - it's a bug fix for existing functionality that we already claim to have.

PR #3798, opened 2018-08-17
CorDapp minimum and target version design doc.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-470

Was previously at https://github.com/corda/corda/pull/3795
New branch with the design doc only.


docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
+# CorDapp Minimum and Target Platform Version
+
+## Overview
+
+We want to give CorDapps the ability to specify which versions of the platform they support.
+
+This will make it easier for CorDapp developers to support multiple platform versions, and enable CorDapp developers to ["tweak behaviour and [...] opt in to changes that might be breaking (e.g. sandboxing)"](https://cordaledger.slack.com/archives/C3J04VC3V/p1534170356000500).
+
+This document proposes that CorDapps will have metadata associated with them specifying a minimum mlatform version and a target platform Version. The minimum platform version of a CorDapp would indicate that a Corda Node would have to be running at least this version of the Corda platform in order to be able to run this CorDapp. The target platform version of a CorDapp would indicate that it was tested for this version of the Corda Platform.
```
florian-f: @gendal commented on the previous PR:

> strictly speaking, is the target version a statement of what has been tested or actually a statement of the highest set of features/behaviours the platform is entitled to adopt? ie if my min is 3 and target is 4 then I may also have been tested on 3 and 5 (who cares?) but what I'm saying is that the platform (whether it is 4 or 5 or whatever) is entitled to operate to the version 4 design if it wants to? ie min=3, target=4 means "I will run on 3 but will happily take advantage of v4 features if they're there"

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+This is intended as a long-term solution. 
+
+## Requirements
+  
+* The CorDapp's minimum and target platform version must be accessible to nodes at CorDapp load time, so they decide whether to load this CorDapp or not and to make the minimum and target version information available ([see Jira](https://r3-cev.atlassian.net/browse/CORDA-470))  .
+
+* The node's platform version must be accessible to CorDapps
+
+* The CorDapp's target platform version must be accessible to the node when running CorDapps.
```
florian-f: @gendal commented on the previous PR:
> This may sound weird but should the CorDapp's target version also be accessible to the CorDapp (and its dependencies) too? ie imagine the cordapp uses a third-party library that makes use of Corda features... might that library want to adapt its behaviour based on the cordapp's target version and not just the actual platform version? (dunno - just thinking out loud)

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Design Decisions
+
+ Corda nodes running a platform version that is lower than the version which will introduce the changes described in this document will not check the minimum version of CorDapps. Thus, they will attempt to run CorDapps with a `minPlatformVersion` higher than their platform version. 
+To prevent this, the check for `minPlatformVersion` could be performed in the CorDapp as well as in the node. When a CorDapp detects that a node which is not fulfilling its `minPlatformVersion` requirement is attempting to run it, it could stop itself from being loaded (or react in a less drastic fashion). However, it may be desirable that the same CorDapp will keep working on some other, older nodes that are located in a different, private, network with fewer restrictions in place. 
+To allow for flexibility it seems reasonable to let CorDapp developers decide how to handle this situation.
```
florian-f: @gendal commented on the previous PR:
> Is this not an abrogation of responsibility? Should we at least commit to documenting advice here or providing a sample?

docs/source/design/targetversion/design.md
mikehearn: It can also mean that, but what it really does is change the node's behaviour so that if there are changes that might have broken apps that worked on v4, they don't apply to the app. So in a sense it really is "what did you test with". Of course we should document such changes clearly too, so they don't _have_ to be discovered via testing.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Goals
+
+Define the semantics of target platform version and minimum platform version attributes for CorDapps. Describe how target and platform versions would be specified by CorDapp developers. Define how these values can be accessed by the node and the CorDapp itself.
+
+## Non-goals
+
+In the future it might make sense to integrate the minimum and target versions into a Corda gradle plugin. Such a plugin is out of scope of this document.
+
+This document does not concern itself with how the _minimum target version_ should work.
```
mikehearn: We should probably remove the min target version if we aren't going to do it fully (and that may well be fine, we can always add it later).

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Timeline
```
mikehearn: We can erase this section, it doesn't add much (i know, the template has too many sections...)

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+This is intended as a long-term solution. 
+
+## Requirements
+  
+* The CorDapp's minimum and target platform version must be accessible to nodes at CorDapp load time, so they decide whether to load this CorDapp or not and to make the minimum and target version information available ([see Jira](https://r3-cev.atlassian.net/browse/CORDA-470))  .
```
mikehearn: Trailing space before full stop. We can just use "app" btw - the term CorDapp is a bit ugly tbh. James likes it though :)

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+* The node's platform version must be accessible to CorDapps
```
mikehearn: full stop at the end.

docs/source/design/targetversion/design.md
mikehearn: Yeah we should probably expose the "what targetVersion/minVersion am I?" API to make it usable by app/library developers too, not just for internal usage.

docs/source/design/targetversion/design.md
mikehearn: I think it's OK, we can get away with it. I'd rather not complicate app programming any further in this area.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+ Corda nodes running a platform version that is lower than the version which will introduce the changes described in this document will not check the minimum version of CorDapps. Thus, they will attempt to run CorDapps with a `minPlatformVersion` higher than their platform version. 
+To prevent this, the check for `minPlatformVersion` could be performed in the CorDapp as well as in the node. When a CorDapp detects that a node which is not fulfilling its `minPlatformVersion` requirement is attempting to run it, it could stop itself from being loaded (or react in a less drastic fashion). However, it may be desirable that the same CorDapp will keep working on some other, older nodes that are located in a different, private, network with fewer restrictions in place. 
+To allow for flexibility it seems reasonable to let CorDapp developers decide how to handle this situation.
+
+
+## Design
+
+### Testing 
+
+Do CorDapp developers have to test their CorDapps on all platform versions v, `minPlatformVersion` <= v <= `targetPlatformVersion`?
```
mikehearn: No, just the latest version and then set targetPlatformVersion to it.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+Developers will need to make sure that their CorDapp runs on a newer platform version. A procedure that can be followed when a new platform version is released should be included in the documentation as a best practise.
```
mikehearn: We already provide upgrade notes, so I suspect this paragraph is superfluous.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+* Suppose our target platform version is 5, and a newer platform version 6 is released. We want to find out if our CorDapp works as expected (without raising the target version)
```
mikehearn: Sure. But why not raise the target version if you're about to test on that?

Perhaps we should specify explicitly that targetVersion is intended only for changes that ratchet in one direction - in the direction of more correctness. Therefore, an app that has been tested and targets v6 should always work on v5. Now if the app advertises minPlatformVersion=6 & targetVersion=6 *then* we can make backwards incompatible changes because we know the app won't work on v5 anyway, it has opted out of backwards compatibility past that version. So we can e.g. reverse the order of a list because we know the app won't be reversing it back.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+* A procedure could look like this:
+  - A new platform version is released/ anounced (network/ nodes not upgraded yet)
+  - CorDapp developers build their CorDapp against the new version, run/ test against nodes running the old/current version and nodes running the new versions. If this works, they can raise the target version of their CorDapp and test again.
+  - CorDapp developers test their old build against a node running the new versions
+    
+### Implications for platform developers
+
+* Should we/ can we make sure CorDapps do not call APIs newer than their target Version
```
mikehearn: Can we write the doc as prose, rather than a set of bullet points? I don't think the bullets are adding much really. Also the design doc describes what we are going to do rather than ask questions about what we should do. Can you rewrite this section to specify the actual proposed plan?

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+    ```
+    // in node
+    if( targetVersion < SOME_SPECIFIC_VERSION ) {
+        oldBehaviour()
+    } else {
+        newBehaviour()
+    }
+    ```
+
+* Will all changes to existing APIs neet to receive version guards, or only some (breaking) changes?
```
mikehearn: We should only be guarding APIs in two ways:

* New features that require other nodes to understand them e.g. changes in the wire tx format, must be gated on the network parameters saying minPlatformVersion=X, where X is the version it was introduced in.
* Changes that risk breaking apps must be gated on targetVersion>=X where X is the version where the change was made, and the old behaviour must be preserved if that condition isn't met.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+* Changes to the Corda Platform that are not directly tied to the node calling the API may also need version guards.
+
+## Technical Design
+
+* How are the minimum- and target version stored and encoded?
```
mikehearn: Again let's rephrase this as prose rather than PowerPoint style Q&A

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+    * For now it would be enough to specify them as properties and add them to the JAR's manifest in gradle:
+    ```
+    ext {
+     targetPlatformVersion=4
+     minPlatformVersion=2
+    }
+    ```
+    
+    ```
+    jar {
```
mikehearn: Can we formalise where these variables are set, so a future Gradle plugin or app build process can inspect the target version and e.g. build the app JAR differently?

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+                    'Min-Platform-Version': minPlatformVersion,
+                    'Target-Platform-Version': targetPlatformVersion
+            )
+        }
+    }
+    ```
+    
+* How can code inside the node find out the minimum and target version of the app that is calling it?
+    * The node's CorDapp loader reads these values from the Manifest when loading the CorDapp
+    * It should be up to the node operator how this error is handled (what should be the default?)
```
mikehearn: Which error? If the fields are missing? Then they default to 1, right?

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+        * Log an error and shut down the node
+        * Log a warning and continue without loading the CorDapp
+    * Would it be enough to add them to `net.corda.core.cordapp.Cordapp`?
+* How can code inside the CorDapp find out its target version?
```
mikehearn: If we do a stack walk we can find the class and thus classloader of the last app on the stack, and make caller-sensitive APIs that way too.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+    * If this information is in `Cordapp`, it can be obtained via the `CorDappContext` from the serviceHub.
+* The node's platform version is already accessible to CorDapps `serviceHub.nodeInfo`
+
+## Observations/ Open Questions
+
+Changing/ Upgrading minimum and target version of a CorDapp: Should there be any constraints? (e.g. the target version of version 11 of a CorDapp must be greater than the target version of version 10)
```
mikehearn: Probably not - what if you don't want to bump targetVersion because you aren't ready to opt in to a breaking change yet?

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+Interactions between minimum and target version and CorDapp/ Flow versioning? Anything to consider/ be aware of? https://docs.corda.net/head/upgrading-cordapps.html#
```
mikehearn: I don't think so, not yet at least. But it might be interesting to expose these to peers in the same way we do for app versions.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+```
+// in CorDapp
+if(NODE_PLATFORM_VERSION < CORDAPP_TARGET_VERSION)
+    oldApi()
+else
+    newApi() // Problem: `newApi()` does not exist on the node, can't load CorDapp   
+    
+```
+
+There is no guarantee with which API version a node is going to run a CorDapp, the Target Version merely indicates a preference (?).
```
mikehearn: Well, it indicates a set of behaviours. If the node doesn't provide those behaviours then it's buggy.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+  * A node running platform version 5 will load and run this CorDapp (at API level 5)
+
+  * A node running platform version 6 Will load and run this CorDapp (at API level 6)
+
+  * A node running platform version 7 Will load and run this CorDapp (at API level 7)
+
+  * A node running platform version 8 Will load and run this CorDapp (at API level 7)
+
+Example: Let's assume that we have a CorDapp with minimum platform version 5 and target platform version 6. A node advertising that it is running platform version 6 on a network with minimum platform version 4 would be free to choose between running our CorDapp on platform version 5 or platform version 6 (TODO should we define a convention for this?).
```
mikehearn: Well, almost by definition it'll be running on platform version 6 because that's what the node is + the app didn't opt into any backwards compatibility goo.

docs/source/design/targetversion/design.md
sollecitom: The node should **definitely** refuse to start in presence of non-compliant Cordapps. There are 2 ways of reacting to such an occasion for a node, either warn and keep going or warn and exit.

The difference is that a warning will easily get ignored with potentially serious consequences, while existing will make sure the issue is promptly faced by the node administrator before any damage.

Example:

Cordapp Blah is incompatible, the node warns about the issue, but keeps going. There are 3 possibility, imaging that Cordapp Blah is used to provide functionality and has been upgraded:

1. The customer has canary deployments with gradual rolling upgrades. In this case some user requests / workflows will fail because of the ignored Cordapp Blah, causing the deployment to fail. Even if production is safe, some users will notice failures, and potentially there might be some necessary rolling back things.

2. The customer has blue / green deployments. Pretty much as above, but with a larger impacted users base.

3. The customer does not have a blue / green or canary deployment strategy. In this case, things will go south at runtime, involving downtime until the previous version in somewhat put back in place.

For all the above cases, if the node refuses to start in presence of non-compliant Cordapps, an operator can easily fix that early on, without any risks. Also, it's much easier to spot in non-production environments.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
+# CorDapp Minimum and Target Platform Version
+
+## Overview
+
+We want to give CorDapps the ability to specify which versions of the platform they support.
+
+This will make it easier for CorDapp developers to support multiple platform versions, and enable CorDapp developers to ["tweak behaviour and [...] opt in to changes that might be breaking (e.g. sandboxing)"](https://cordaledger.slack.com/archives/C3J04VC3V/p1534170356000500).
```
josecoll: Please summarise the Slack conversation here in one paragraph.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+This document proposes that CorDapps will have metadata associated with them specifying a minimum mlatform version and a target platform Version. The minimum platform version of a CorDapp would indicate that a Corda Node would have to be running at least this version of the Corda platform in order to be able to run this CorDapp. The target platform version of a CorDapp would indicate that it was tested for this version of the Corda Platform.
+
+For reference, see [this ticket in Jira](https://r3-cev.atlassian.net/browse/CORDA-470).
```
josecoll: Again, please summarise any important relevant content here (and reference the JIRA later on in the references section of the template)

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Background
+
+* *Platform version (Corda)* An integer representing the API version of the Corda platform ([see docs](https://docs.corda.net/head/versioning.html#versioning)).
```
josecoll: Please add definitions in this document as web docs may change / not be accessible if reading offline.

docs/source/design/targetversion/design.md
josecoll: I think it is fair to state this in this section of the design document (the technical section may add a disclaimer stating this is not going to be implemented in the 1st iteration).

docs/source/design/targetversion/design.md
josecoll: This also relates to the implementation timeline.
eg. 
iteration 1 (release 4) will implement all features except _minimum target version_
iteration 2 (release vNext) will implement _minimum target version_

docs/source/design/targetversion/design.md
josecoll: Paragraph kind of repeats itself (or is not entirely clear)
"The CorDapp's minimum and target platform version must be accessible to nodes"
"to make the minimum and target version information available"

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+  
+* The CorDapp's minimum and target platform version must be accessible to nodes at CorDapp load time, so they decide whether to load this CorDapp or not and to make the minimum and target version information available ([see Jira](https://r3-cev.atlassian.net/browse/CORDA-470))  .
+
+* The node's platform version must be accessible to CorDapps
+
+* The CorDapp's target platform version must be accessible to the node when running CorDapps.
+
+## Design Decisions
+
+ Corda nodes running a platform version that is lower than the version which will introduce the changes described in this document will not check the minimum version of CorDapps. Thus, they will attempt to run CorDapps with a `minPlatformVersion` higher than their platform version.
```
josecoll: This sounds confusing and dismissive of existing CorDapps. 
Surely they need to be checked in the same way (whether they have the new metadata or not).

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+* The node's platform version must be accessible to CorDapps
+
+* The CorDapp's target platform version must be accessible to the node when running CorDapps.
+
+## Design Decisions
+
+ Corda nodes running a platform version that is lower than the version which will introduce the changes described in this document will not check the minimum version of CorDapps. Thus, they will attempt to run CorDapps with a `minPlatformVersion` higher than their platform version. 
+To prevent this, the check for `minPlatformVersion` could be performed in the CorDapp as well as in the node. When a CorDapp detects that a node which is not fulfilling its `minPlatformVersion` requirement is attempting to run it, it could stop itself from being loaded (or react in a less drastic fashion). However, it may be desirable that the same CorDapp will keep working on some other, older nodes that are located in a different, private, network with fewer restrictions in place.
```
josecoll: Perhaps we can relax certain checks for development/test environments, but strictly enforce them for production environments (eg. we don't want to allow malicious CorDapps running in the Corda Network by masquerading as pre-version checking instances)

docs/source/design/targetversion/design.md
josecoll: CorDapps should run on newer platform versions without change. Any failures would be picked up in a QA release cycle.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Testing 
+
+Do CorDapp developers have to test their CorDapps on all platform versions v, `minPlatformVersion` <= v <= `targetPlatformVersion`?
+
+Developers will need to make sure that their CorDapp runs on a newer platform version. A procedure that can be followed when a new platform version is released should be included in the documentation as a best practise.
+
+* Suppose our target platform version is 5, and a newer platform version 6 is released. We want to find out if our CorDapp works as expected (without raising the target version)
+* A procedure could look like this:
+  - A new platform version is released/ anounced (network/ nodes not upgraded yet)
```
josecoll: typo: "announced"

docs/source/design/targetversion/design.md
florian-f: I'm not sure what you mean. What I want to point out here is that nodes running Corda platform version 3 will not check `minPlatformVersion`, because they don't have that functionality built in.

docs/source/design/targetversion/design.md
florian-f: Not sure what you mean by formalising this.

docs/source/design/targetversion/design.md
florian-f: Yes this is unclear. I meant, what should happen when e.g. the minimum platform version is greater than the node's platform version. It's what @sollecitom discusses here https://github.com/corda/corda/pull/3798#discussion_r210933624

docs/source/design/targetversion/design.md
florian-f: I'm turning this into a "design decision" section.

docs/source/design/targetversion/decisions/decision.md
```diff
@@ -0,0 +1,39 @@
...
+2. ​
+
+#### Disadvantages
+
+1. ​
+2. ​
+
+## Recommendation and justification
+
+Use either A or B as default, make the behaviour configurable.
```
josecoll: I think offering configurability for this operation will lead to confusion.
IMHO the default behaviour of a node should be to log a WARNING or custom-defined severity (that can be picked up by a monitoring agent for operational control) and continue loading.

docs/source/design/targetversion/decisions/decision.md
```diff
@@ -0,0 +1,39 @@
+![Corda](https://www.corda.net/wp-content/uploads/2016/11/fg005_corda_b.png)
+
+--------------------------------------------
+Design Decision: How should the node react when the check for minimum or target version fails?
+============================================
+
+## Background / Context
+
+What should happen when the node loads a CorDapp and detects that the CorDapp's minimum platform version is greater than the node's platform versoin, or the CorDapps target platform version is less than the network's minimum target version?
```
josecoll: typo: versoin => version

docs/source/design/targetversion/decisions/decision.md
```diff
@@ -0,0 +1,39 @@
...
+
+--------------------------------------------
+Design Decision: How should the node react when the check for minimum or target version fails?
+============================================
+
+## Background / Context
+
+What should happen when the node loads a CorDapp and detects that the CorDapp's minimum platform version is greater than the node's platform versoin, or the CorDapps target platform version is less than the network's minimum target version? 
+
+## Options Analysis
```
josecoll: The arguments for and against are so minimal that I don't think this even warrants a decision.

docs/source/design/targetversion/design.md
josecoll: Please summarise the Slack conversation here in one paragraph.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,91 @@
+# CorDapp Minimum and Target Platform Version
+
+## Overview
+
+We want to give CorDapps the ability to specify which versions of the platform they support.
+
+This will make it easier for CorDapp developers to support multiple platform versions, and enable CorDapp developers to ["tweak behaviour and [...] opt in to changes that might be breaking (e.g. sandboxing)"](https://cordaledger.slack.com/archives/C3J04VC3V/p1534170356000500).
+
+This document proposes that CorDapps will have metadata associated with them specifying a minimum mlatform version and a target platform Version. The minimum platform version of a CorDapp would indicate that a Corda Node would have to be running at least this version of the Corda platform in order to be able to run this CorDapp. The target platform version of a CorDapp would indicate that it was tested for this version of the Corda Platform.
```
josecoll: Add a reference to https://docs.corda.net/head/versioning.html so the reader understands what is currently in place.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Testing 
+
+When a new platform version is released, CorDapp developers can test their CorDapp against that version and increase its target version. This way they would opt-in to potentially breaking changes that were introduced in that version. If they choose to not increase the target version, their CorDapp will continue to work. The only scenario in which the target version has to be increased is when the minimum target version of the network is set to a higher value.
+    
+### Implications for platform developers
+
+When new features or changes are introduced that require all nodes on the network to understand them (e.g. changes in the wire transaction format), they must be version-gated on the network level. The minimum platform version in the network parameters would have to be changed the version these features or changes are introduced in.
+
+Changes that risk breaking apps must be gated on targetVersion>=X where X is the version where the change was made, and the old behaviour must be preserved if that condition isn't met.
```
josecoll: I don't understand this sentence. What do you mean by "changes must be gated" ?

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,91 @@
...
+   
+## Design
+
+### Testing 
+
+When a new platform version is released, CorDapp developers can test their CorDapp against that version and increase its target version. This way they would opt-in to potentially breaking changes that were introduced in that version. If they choose to not increase the target version, their CorDapp will continue to work. The only scenario in which the target version has to be increased is when the minimum target version of the network is set to a higher value.
+    
+### Implications for platform developers
+
+When new features or changes are introduced that require all nodes on the network to understand them (e.g. changes in the wire transaction format), they must be version-gated on the network level. The minimum platform version in the network parameters would have to be changed the version these features or changes are introduced in.
```
josecoll: What does "version-gated on the network level" mean?
The following sentence is also grammatically incorrect.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,91 @@
...
+
+Changes that risk breaking apps must be gated on targetVersion>=X where X is the version where the change was made, and the old behaviour must be preserved if that condition isn't met.
+
+## Technical Design
+
+The minimum- and target platform version will be written to the Manifest of the CorDapp's JAR, in fields called `Min-Platform-Version` and `Target-Platform-Version`.
+The node's CorDapp loader reads these values from the Manifest when loading the CorDapp and adds them to `net.corda.core.cordapp.Cordapp`, which can be obtained via the `CorDappContext` from the service hub. For cases where the service hub is not available, it is possible to do a stack walk to find the class and thus classloader of the last app on the stack. This way it is also possible to make caller-sensitive APIs.
```
josecoll: Sorry ... I'm not sure I understand what you mean by certain items in this paragraph:
- "For cases where the service hub is not available" (which ones? the service hub is always available when bootstrapping the node), 
- "stack walk": why and when would you need to do this?
-  "caller-sensitive APIs": what does this mean?

docs/source/design/targetversion/design.md
florian-f: That information is in the background section. Do you want to have this information both in this section and the background section?

docs/source/design/targetversion/design.md
florian-f: I will double-check if there's anything in the discussion that is not mentioned in this document yet.

docs/source/design/targetversion/design.md
florian-f: I'll rewrite this section to make it more clear.

Basically, on the CorDapp level it means that if a change potentially breaks CorDapps that are older / do not opt in to this change, we need to make sure to version-gate the change like this:

```
    if( corDapp.targetVersion < SOME_SPECIFIC_VERSION ) {
        oldBehaviour()
    } else {
        newBehaviour()
    }
```

on the network level it's similar, it would look something like
```
    if( NetworkParams.minTargetVersion < SOME_SPECIFIC_VERSION ) {
        oldBehaviour()
    } else {
        newBehaviour()
    }
```
Does that make sense?

docs/source/design/targetversion/design.md
florian-f: > "For cases where the service hub is not available" (which ones? the service hub is always available when bootstrapping the node),
This does not just affect bootstrapping.

> "stack walk": why and when would you need to do this?
> "caller-sensitive APIs": what does this mean?

Let's say we want to make FlowStateMachineImpl.checkFlowPermission more "strict" (whatever that could mean), but doing so would potentially break existing CorDapps. Now, checkFlowPermission is called from some flow, which is part of some CorDapp. But inside checkFlowPermission we don't have any information available what the version of that CorDapp might be. This is where a stack-walk can help. This technique is also used by `CordappProvider.getAppContext()`.

docs/source/design/targetversion/design.md
florian-f: > What does "version-gated on the network level" mean?

Answered here https://github.com/corda/corda/pull/3798#discussion_r211892795

> The following sentence is also grammatically incorrect.

Thanks, will fix that.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,91 @@
...
+  
+([from the docs](https://docs.corda.net/head/versioning.html#versioning)). 
+
+* *Platform version (Node)* The value of the Corda platform version that a node is running and advertising to the network.
+
+* *Minimum platform version (Network)* The minimum platform version that the nodes must run in order to be able to join the network. Set by the network zone operator. The minimum platform version is distributed with the network parameters as `minimumPlatformVersion`.
+ ([see docs:](https://docs.corda.net/network-map.html#network-parameters))
+ 
+* *Minimum target version (Network)* Does not exist yet. We are planning to introduce the minimum target Version as part of the network parameters. This document assumes that it indicates to nodes in the network that they should not run CorDapps with a target version lower than this.
```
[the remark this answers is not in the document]
florian-f: Will do. When I wrote the first draft I assumed this would be defined somewhere else, that's why I wrote it like this.

docs/source/design/targetversion/design.md
josecoll: Thanks for clarifying. I recommend you add the above example to the mainline text for the benefit of other readers.

docs/source/design/targetversion/design.md
josecoll: Ok - well remove the Slack reference then please.

docs/source/design/targetversion/design.md
josecoll: The Background section should summarise what is currently in place, and the motivation for extending what we have. No need to duplicate anything (just cross-reference within the doc if necessary).

docs/source/design/targetversion/design.md
josecoll: OK - yes, please include the definition somewhere in the doc.

florian-f:
> The only major point I would address is the "Decision point" - which I think is probably unnecessary.
eg. we prescribe what we are going to do and agree on it in this PR (as this design is lightweight and doesn't need to go to the DRB).

@josecoll Do we agree on this though? @sollecitom seems to have a different opinion https://github.com/corda/corda/pull/3798#discussion_r210933624

We could just leave the behaviour as an implementation detail and not specify it here.

josecoll:
The behaviour should be specified and agreed in this design before implementation.
As I mentioned in a previous comment:

```
"I think offering configurability for this operation will lead to confusion.
IMHO the default behaviour of a node should be to log a WARNING or custom-defined severity (that can be picked up by a monitoring agent for operational control) and continue loading."
```
Perhaps @mikehearn can jump in and mediate.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,85 @@
+# CorDapp Minimum and Target Platform Version
+
+## Overview
+
+We want to give CorDapps the ability to specify which versions of the platform they support. This will make it easier for CorDapp developers to support multiple platform versions, and enable CorDapp developers to tweak behaviour and opt in to changes that might be breaking (e.g. sandboxing). Corda developers gain the ability to introduce changes to the implementation of the API that would otherwise break existing CorDapps.
+
+This document proposes that CorDapps will have metadata associated with them specifying a minimum mlatform version and a target platform Version. The minimum platform version of a CorDapp would indicate that a Corda Node would have to be running at least this version of the Corda platform in order to be able to run this CorDapp. The target platform version of a CorDapp would indicate that it was tested for this version of the Corda Platform. In addition to this, a minimum target version will be introduced on the level of the network.
```
mikehearn: typo: mlatform ... version, node and platform decapitalised please.

Let's skip the min target version for now. I think you convinced me this concept wasn't well thought out enough and I'd like this work to proceed smoothly.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,85 @@
...
+.. toctree::
+   :maxdepth: 2
+
+   decisions/decision.md
+   
+## Design
+
+### Testing 
+
+When a new platform version is released, CorDapp developers can test their CorDapp against that version and increase its target version. This way they would opt-in to potentially breaking changes that were introduced in that version. If they choose to not increase the target version, their CorDapp will continue to work. The only scenario in which the target version has to be increased is when the minimum target version of the network is set to a higher value.
```
mikehearn: It has to be the other way around, right - increase the target version and then re-test their app (after reading what the new target version changes in the docs).

Let's boot out the concept of a min target version completely for now. We can just erase the references to it and tackle it later.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,85 @@
...
+    
+### Implications for platform developers
+
+When new features or changes are introduced that require all nodes on the network to understand them (e.g. changes in the wire transaction format), they must be version-gated on the network level. This means that the new behaviour should only take effect if the minimum target verison of the network is equal to or greater than the version in which these changes were introduced. Failing that, the old behaviour must be used instead.
```
mikehearn: min *platform* version, right?

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Technical Design
+
+The minimum- and target platform version will be written to the Manifest of the CorDapp's JAR, in fields called `Min-Platform-Version` and `Target-Platform-Version`.
+The node's CorDapp loader reads these values from the Manifest when loading the CorDapp and adds them to `net.corda.core.cordapp.Cordapp`, which can be obtained via the `CorDappContext` from the service hub. For cases where the service hub is not available, it is possible to do a stack walk to find the class and thus classloader of the last app on the stack. This way it is also possible to make caller-sensitive APIs. As an example, let's say we want to make `FlowStateMachineImpl.checkFlowPermission` more "strict" (whatever that could mean), but doing so would potentially break existing CorDapps. Now, `checkFlowPermission` is called from some flow, which is part of some CorDapp. But inside `checkFlowPermission` we don't have any information available what the version of that CorDapp might be, since the service hub is not available there. This is where a stack-walk can help.
```
mikehearn: Can you describe the API you plan to implement for this here?

docs/source/design/targetversion/design.md
florian-f: OK, removing min target version

docs/source/design/targetversion/design.md
florian-f: true

docs/source/design/targetversion/design.md
florian-f: I was planning to have a static method similar to `CordappProviderImpl.getAppContext` in the companion object of either Cordapp or CordappContext.

florian-f: Also, `FlowStateMachineImpl.checkFlowPermission` is a bad example because `servicehub` is actually available there

docs/source/design/targetversion/design.md
florian-f: Right, this is poorly worded. What I meant was increase the target version, test, and then _release with the new target version_.

docs/source/design/targetversion/design.md
```diff
@@ -0,0 +1,90 @@
...
+
+To make APIs caller-sensitive in cases where the service hub is not available a different approach has to be used. It would possible to do a stack walk, and parse the manifest of each class on the stack to determine if it belongs to a CorDapp, and if yes, what its target version is. Alternatively, the mapping of classes to `Cordapp`s obtained by the CorDapp loader could be stored in a global singleton. This singleton would expose a lambda returning the current CorDapp's version information (e.g. `() -> Cordapp.Info`).  
+  
+Let's assume that we want to change `TimeWindow.Between` to make it inclusive, i.e. change `contains(instant: Instant) = instant >= fromTime && instant < untilTime` to `contains(instant: Instant) = instant >= fromTime && instant <= untilTime`. However, doing so will break existing CorDapps. We could then version-guard the change such that the new behaviour is only used if the target version of the CorDapp calling `contains` is equal to or greater than the platform  version that contains this change. It would look similar to this:
```
shamsasari: Minor: This is not a good example of using target version - a `containsInclusive` would suffice.

florian-f: Right, I will update this with a better example once I come up with one.

PR #3809, opened 2018-08-17
Add a design doc for signature constraints.

mikehearn:
This PR adds a design doc for signature constraints.


docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
+# Signature constraints
+
+This design document outlines an additional kind of *contract constraint*, used for specifying inside a transaction what the set of allowable attached contract JARs can be for each state.
+
+## Background
+
+Contract constraints are a part of how Corda manages application upgrades. There are two kinds of upgrade that can be applied to the ledger:
```
josecoll: Isn't the primary purpose of Contract Constraints to ensure the correct verification logic is executed (from the universe of attachment jars)? 
As per the definition in https://docs.corda.net/head/api-contract-constraints.html#contract-constraints

docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
...
+
+* Explicit
+* Implicit
+
+An *explicit* upgrade is when a special kind of transaction is used, the *contract upgrade transaction*, which has the power to suspend normal contract execution and validity checking. The new contract being upgraded-to must be willing to accept the old state and can replace it with a new one. Because this can cause arbitrary edits to the ledger, every participant in a state must sign the contract upgrade transaction for it to be considered valid. 
+
+Note that in the case of single-participant states whilst you could unilaterally replace a token state with a different state, this would be a state controlled by an application that other users wouldn't recognise, so you cannot transmute a token into a private contract with yourself then transmute it back, because contracts will only upgrade states they created themselves.
```
josecoll: What do you mean by "transmute a token into a private contract with yourself then transmute it back"?
Is I upgraded a fungible Cash state with a new Issuer reference (I am the owner and issuer), why would that no longer be recognised by other users?

docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
...
+* `WhitelistedByZoneContractConstraint` - the network parameters contains a map of state class name to allowable hashes for the attachments. 
+
+The last constraint allows upgrades 'from the future' to be applied, without disabling ledger security. However it is awkward to use, because any new version of any app requires a new set of network parameters to be signed by the zone operator and accepted by all participants, which in turn requires a node restart.
+
+The problems of `WhitelistedByZone` were known at the time it was developed, however, the feature was implemented anyway to reduce schedule slip for the Corda 3.0 release, whilst still allowing some form of application upgrade.
+
+We would like a new kind of constraint that is more convenient and decentralised whilst still being secure.
+
+
+## Goals
```
josecoll: Should this design not specify how existing Constraint Types (hash, CZ whitelist) on existing states are handling when processing a a new transaction with signature constraints? 
(I believe @tudor-malene has specified some clear rules and scenarios)
eg. "In a transaction, the constraints of Output states should have at least the same level of restriction as the constraints of the corresponding Input States."

docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Non-goals
+
+* Preventing downgrade attacks. Downgrade attack prevention will be tackled in a different design effort.
+* Phase out of hash constraints. If malicious app creators are in the users threat model then hash constraints are the way to go.
+* Handling the case where third parties re-sign app jars.
+
+## Design details
+
+We propose being able to constrain to any attachments signed by a specified set of keys.
```
josecoll: I interpreted this statement wrong and assumed it to mean that "an attachment jar  is signed by a specified set of keys" (not the individual files within the attachment jar).

docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
...
+* Improve decentralisation by allowing apps to be developed and upgraded without the zone operator knowing or being able to influence it.
+
+* Eventually, phase out zone whitelisting constraints.
+
+
+## Non-goals
+
+* Preventing downgrade attacks. Downgrade attack prevention will be tackled in a different design effort.
+* Phase out of hash constraints. If malicious app creators are in the users threat model then hash constraints are the way to go.
+* Handling the case where third parties re-sign app jars.
```
josecoll: This design later goes on to discuss how to deal withidentical jars signed by multiple parties (eg. Third Party signing of JARs)

docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
...
+## Goals
+
+* Improve usability by eliminating the need to change the network parameters.
+
+* Improve decentralisation by allowing apps to be developed and upgraded without the zone operator knowing or being able to influence it.
+
+* Eventually, phase out zone whitelisting constraints.
+
+
+## Non-goals
```
josecoll: Package hierarchy ownership (discussed in the alternative "Package name constraints" section of the document as further work).

docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
...
+
+```kotlin
+data class SignatureAttachmentConstraint(
+    val key: CompositeKey
+) : AttachmentConstraint
+```
+
+Therefore if a state advertises this constraint, along with a class name of `com.foo.Bar` then the definition of Bar must reside in an attachment with signatures sufficient to meet the given composite key. Multiple signers of a JAR is useful for decentralised administration of an app that wishes to have a threat model in which one of the app developers may go bad, but not a majority of them. For example there could be a 2-of-3 threshold of {app developer, auditor, R3} in which R3 is legally bound to only sign an upgrade if the auditor is unavailable e.g. has gone bankrupt. However, we anticipate that most constraints will be one-of-one for now.
+
+We will add a `signers` field to the `ContractAttachment` class that will be filled out at load time if the JAR is signed. The signers will be computed by checking the certificate chain for every file in the JAR, and any unsigned files will cause an exception to be thrown.
```
josecoll: If we are enforcing all files to be checked/hashed, why not just perform this once at the JAR level?

docs/source/design/data-model-upgrades/signature-constraints.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Transaction building
+
+The `TransactionBuilder` class can select the right constraint given what it already knows. If it locates the attachment JAR and discovers it has signatures in it, it can automatically set an N-of-N constraint that requires all of them on any states that don't already have a constraint specified. If the developer wants a more sophisticated constraint, it is up to them to set that explicitly in the usual manner.
+
+### Tooling and workflow
+
+The primary tool required is of course `jarsigner`. In dev and integration test modes, the node will ignore missing signatures in attachment JARs and will simply log a warning if no signature is present.
```
josecoll: do you mean "dev" and "unit test" modes? Otherwise, where do you actually write tests to verify this functionality works?

docs/source/design/data-model-upgrades/signature-constraints.md
josecoll: Using network parameters to provide a selective override mechanism (discussed in "Upgrading from other constraints") section.

josecoll:
Damn pokeybot!!!

docs/source/design/data-model-upgrades/signature-constraints.md
mikehearn: Sure but if apps never changed we wouldn't need any notion of constraints or attachments. We could remove the concept and always hard-code the hash of the app.

docs/source/design/data-model-upgrades/signature-constraints.md
mikehearn: What I mean is, grab yourself a token of $50 from a bank, do an explicit upgrade to com.colljos.MyToken, edit the amount to $500 then do an "upgrade" back to cash. The last step is blocked because the real cash app will refuse to "upgrade" from a state it doesn't recognise (or that's how it's meant to work!).

docs/source/design/data-model-upgrades/signature-constraints.md
mikehearn: That's a separate initiative - it isn't affected by the presence or absence of signature constraints.

docs/source/design/data-model-upgrades/signature-constraints.md
josecoll: Right, yes I understand it facilitates upgrades. 
However it ensures that the correct code is executed in the presence of potentially accidental incorrect builds and/or rogue/incorrectly distributed cordapps.

docs/source/design/data-model-upgrades/signature-constraints.md
mikehearn: right - it describes the issue, but says it's out of scope to solve.

docs/source/design/data-model-upgrades/signature-constraints.md
mikehearn: I think this is discussed in the "out of line signatures" section.

PR #3811, opened 2018-08-20
Fix Jib build

thschroeter:


node/build.gradle
```diff
@@ -48,7 +48,7 @@ sourceSets {
 jib.container {
```
adagys: what's jib?

thschroeter: a tool to build container images fast (no docker daemon required). https://github.com/GoogleContainerTools/jib

PR #3817, opened 2018-08-20
CORDA-1787 Bring threat model into the docsite 

jamesbr3:
The threat-model has been updated, refreshed and formatted to look more like a threat model. 

The content has been copied across (auto-formatted) from:
https://r3-cev.atlassian.net/wiki/spaces/SEC/pages/561316232/Corda+Threat+model+interim

and will appear under:

docs/source/threat-model/


docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+as ARP & IP spoofing or DNS compromise.
+
+Spoofing can also occur at an application or user-level. Attacks at this level
+typically target authentication logic, using compromised passwords and
+cryptographic keys, or by subverting cryptography systems.
+
+Corda employs a Public Key Infrastructure (PKI) to validate the identity of
+nodes, both at the point of registration with the network map service and
+subsequently through the cryptographic signing of transactions. An imposter
+would need to acquire an organisation's private keys in order to meaningfully
```
gendal: They could also trick the doorman into issuing an incorrect certificate, right? Which leads to a discussion of a plausible end-state vision where, to mitigate the risk of incorrectly issued certificates, we highlight that it is the responsibility of a business network operator only to whitelist (for whatever definition is meaningful in the context of their app(s)) certificates that they have independently confirmed have been issued correctly.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+impersonate that organisation. R3 provides guidance to all ledger network
+participants to ensure adequate security is maintained around cryptographic
+keys.
```
gendal: ... and we are working on HSM support for nodes. Plus the longer-term direction for client-side signing devices in the whitepaper.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+Connections to the Network Map service are secured using the HTTPS protocol.
+The connecting node authenticates the NetworkMap servers using their public
+certificates, to ensure the identity of these servers is correct.
+
+All data received from the NetworkMap is digitally signed - an attacker
+attempting to spoof either service would need access to the corresponding
+private keys.
+
+The Doorman and NetworkMap signing keys are stored inside a (Hardware Security
```
gendal: clarify that the key used to sign the netmap is different to the key used to secure the TLS session?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+deletion of data to malicious tampering of financial detail.
+
+</td>  
+<td>
+
+There are not currently any direct controls to mitigate this kind of attack.
+The node's vault is assumed to be within the same trust boundary of the node.
+Access to the vault must be restricted such that only the node can access it.
+Both network-level controls (fire-walling) and database permissions must be
+employed.
```
gendal: elaborate on the point that the only party this harms is the party whose vault it is, right?  ie nobody else can be fooled.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+Individual Node entries in the NetworkMap must be signed by the associated
+Node. The signatures are validated by the NetworkMap service, and all other
+Nodes in the network, to ensure they have not been tampered with. An attacker
+would need to acquire a node's private identity signing key to be able to make
+modifications to a NodeInfo. It is not possible for the NetworkMap service (or
+R3) to modify the NodeInfo entries, therefore an attacker would also be
+limited in this regard.
+
+The only impact the attacker could have would be to remove individual entries
+in the map, or insert their own NodeInfo.
```
gendal: perhaps elaborate further on the role of the additional-node-infos directory?  ie even in the event of a total network map server outage (perhaps this should be under denial of service below), it  _would_ be possible, if annoying to distribute NodeInfos out of band?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+Repudiation refers to the ability to claim a malicious action did not take
+place. Repudiation becomes relevant when it is not possible to verify the
+identity of an attacker, or there is a lack of evidence to link their
+malicious actions with events in a system.
+
+Preventing repudiation does not prevent other forms of attack. Rather, the
+goal is to ensure that the attacker is identifiable, their actions can be
+traced, and there is no way for the attacker to deny having committed those
+actions.
```
gendal: these two paragraphs are a little confusing.  Perhaps it's my lack of understanding but my mental model (until now!) of repudiation was that it 1) ENABLES me to prove/deny something that didn't (correctly) happen actually didn't happen and 2) PREVENTS me from denying something that DID happen actually happened.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+</th></tr>  
+<tr>  
+<td>
+
+RPC Client
+
+</td>  
+<td>
+
+Attacker attempts to initiate a transcation they are not entitled to perform
```
gendal: typo

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+</td>  
+<td>
+
+A malicious CordApp attempts to spend a transaction that does not belong to
+them
+
+</td>  
+<td>
+
+All P2P transactions are logged by the node.
```
gendal: this section seems a bit light.     And isn't the thing that prevents me from spending something that doesn't belong to me the verification logic of the associated transaction?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+</td></tr>  
+<tr>  
+<td>
+
+Node
+
+</td>  
+<td>
+
+A malicious CordApp attempts to spend a transaction that does not belong to
```
gendal: CorDapp (capital D).  Also - what does it mean to spend a transaction. Do you mean a state?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+If successful, an exploit of the form above could result in information
+private to specific participants being shared with one or more non-privileged
+parties. This may include market-sensitive information used to derive
+competitive advantage.
+
+</td>  
+<td>
+
+Corda will not divulge transaction history to a peer node unless that node is
```
gendal: Is that correct?  I thought the current implementation (design?) is such that a node will only divulge a tx history to a peer if the tx being asked for is a descendent of a tx that the node itself had previously shared as part of the current session - or something like that.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+an explicit party to the transaction.
+
+By design, Corda nodes do not globally broadcast transaction information to
+all participants in the network. Hash functions used by Corda (primarily
+SHA256) are effectively immune to brute-force attacks, and integration of
+blinding nonces into the transaction data will be used to prevent an attacker
+from successfully obtaining a transaction hash even if the transaction content
+is known/guessed. Risk from a guessing-based approach as described above is
+therefore believed to be negligible.
```
gendal: I think the risk of guessing has been mitigated through the "you can only request txs or their histories if they were sent to you by the node you're asking" feature (I think!).  Andrius implemented it IIRC

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+</td>  
+<td>
+
+Corda Nodes are designed to be executed using a designated 'corda' system
+process, which other users and processes on the system do not have permission
+to access.
+
+The node's Java Key Store is encrypted using PKCS#12 encryption. In the future
+Corda will eventually store it's keys in a HSM (Hardware Security Module).
```
gendal: its

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+</td>  
+<td>
+
+RPC communications are protected by the TLS protocol.
+
+Permission to query a node's vault must be explicitly granted on a per-user
+basis.
+
+</td></tr></table>
```
gendal: do we need to talk about how the credentials used by the node to connect to the database or HSM are protected? ie by default in a plain text file but we have mechanisms to inject credentials via env vars, etc?

also - how is the float/bridge protected? especially the bridge key material.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+  
+
+## Denial of Service
+
+Denial-of-service (DoS) attacks target the availability of a resource from its
+intended users. There are two anticipated targets of a DoS attack - network
+participants (Corda Nodes) and network services (Doorman and the Network Map).
```
gendal: and notaries?  or do we regard them as special cases of nodes?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+Node
+
+</td>  
+<td>
+
+An attacker control sends high volume of malformed transactions to a node.
+
+ **Impact**
+
+Nodes targeted by this attack could exhaust theor processing & memory
```
gendal: their?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+Depending on the nature of the flow, a party could be financially impacted by
+failure to resolve a flow on an indefinite basis. For example, a party may be
+left in possession of a digital asset without the means to transfer it to
+another party.
+
+</td>  
+<td>
+
+The network agreement will stipulate a default maximum allowable period of 30
```
gendal: rather than specify the time maybe just refer to the event horizon concept?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+
+</td>  
+<td>
+
+The AMQP serialiser uses a combination of white and black-lists to mitigate
+against de-serialisation vulnerabilities.
+
+Corda does not currently provide specific security controls to mitigate all
+classes of privilege escalation vulnerabilities. The design of Corda requires
+that CordApps are inherently trusted by the node administrator. Likewise,
```
gendal: CorDapps

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+node-on-node attacks are not part of the formal threat model, and P2P traffic
```
gendal: this is a critical point... perhaps important enough to elevate to the introduction?  ie it's a departure from the threat model of bitcoin/ethereum (whether they achieve resilience to the threat is another question) but it also directly leads to business/commercial decisions about how widely available to make the testnet, etc...     should we want to be more resilient against nOn attacks, what would we do?

gendal: ah - I see... you answer that last question on the next line :)

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,883 @@
...
+Built on the [Corda](http://www.corda.net/) distributed ledger platform
+designed by R3, the ledger network enables the origination and management of
+(primarily financial) agreements between business partners. Participants to
+the network create and maintain Corda _nodes,_ each hosting one or more
+pluggable applications ( _CorDapps_ ) which define the data to be exchanged
+and its workflow. See the [Corda Technical White
+Paper](https://docs.corda.net/_static/corda-technical-whitepaper.pdf) for a
+detailed description of Corda's design and functionality.
+
+R3 will provide and maintain a number of essential services underpinning the
```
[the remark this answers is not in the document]
jamesbr3: ok updated

docs/source/threat-model/corda-threat-model.md
jamesbr3: good point, i've updated to mention hsm here

docs/source/threat-model/corda-threat-model.md
jamesbr3: done

docs/source/threat-model/corda-threat-model.md
jamesbr3: very good point; updated

docs/source/threat-model/corda-threat-model.md
jamesbr3: this is covered in the denial-of-service section so i'll leave this as-is

docs/source/threat-model/corda-threat-model.md
jamesbr3: i'll try and clarify this, obviously this part needs work. 

Repudation is an attacker's ability to deny their actions. Preventing repudation is therefore providing some kind of a defence. i'll fix

docs/source/threat-model/corda-threat-model.md
jamesbr3: noted. i'll try and include some better descriptions here. Repudiation is a difficult one to describe :D

docs/source/threat-model/corda-threat-model.md
jamesbr3: yes, i meant state. fixed

docs/source/threat-model/corda-threat-model.md
jamesbr3: i need to think about this!

docs/source/threat-model/corda-threat-model.md
jamesbr3: I haven't drawn any distinction actually. Arguably I should but I was thinking to leave that to a separate higher-level 'ledger' threat model rather than give too much detail here. But I think it is worth mentioning, I will make this change, but not go into too many  specifics

docs/source/threat-model/corda-threat-model.md
jamesbr3: fixed

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+It is incumbent on all ledger network participants to review and assess the
+security measures described in this document against their specific
+organisational requirements and policies, and to implement any additional
+measures needed.
+
+## Scope
+
+Built on the [Corda](http://www.corda.net/) distributed ledger platform
+designed by R3, the ledger network enables the origination and management of
+(primarily financial) agreements between business partners. Participants to
```
mikehearn: Can we take out the (primarily financial) part? We're getting quite a lot of non financial firms these days.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+and its workflow. See the [Corda Technical White
+Paper](https://docs.corda.net/_static/corda-technical-whitepaper.pdf) for a
+detailed description of Corda's design and functionality.
+
+R3 provide and maintain a number of essential services underpinning the ledger
+network. In the future these services are intended to be operated by a
+separate Corda Foundation. The network services currently include:
+
+  * Network map service: Service for distributing and providing updates for a network map document enabling nodes to identify other nodes on the network, their network address and advertised services. 
+  * Network permissioning service ('Doorman'): Issues signed digital certificates credentialising the identity of parties on the network to conduct peer-to-peer communication.
```
mikehearn: credentialising? :)

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+The diagram below illustrates the network architecture, protocols and high
+level data flows that comprise the Corda Network. The threat model has been
+developed based upon this architecture.
+
+![Threat Model](./images/threat-model.png)
+
+# The Threat Model
+
+Threat Modelling is an iterative process that works to identity, describe and
+mitigate threats to a system. One of the most common models for identify
```
mikehearn: typo: identify -> identifying

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+Node
+
+</td>  
+<td>
+
+An attacker attempts to impersonate a node and issue a transaction using their
+identity.
+
+An attacker attempts to impersonate another node on the network by submitting
+NodeInfo updates with - falsified address and/or identity information.
```
mikehearn: do we need the hyphen here?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+NodeInfo updates contain the Node's public identity certificate and must be
+signed by the corresponding private key. Only the node in possession of this
+private key can sign the NodeInfo.
+
+Corda employs a Public Key Infrastructure (PKI) to validate the identity of
+nodes. An imposter would need to acquire an organisation's private keys in
+order to meaningfully impersonate that organisation. Corda will soon support a
+range of HSMs (Hardware Security Modules) for storing a node's private keys,
+which mitigates this risk.
+
```
mikehearn: Do we need to discuss the case where the zone operator is the attacker here? The operator can issue an identity cert for anyone, publish a valid NodeInfo and redirect all traffic to themselves in the clear.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+ **Impact**
+
+An attacker would be able to modify transactions between participating nodes.
+
+</td>  
+<td>
+
+Mutually authenticated TLS connections between nodes ensures that Man-In-The-
+Middle (MITM) attacks cannot take place. Corda Nodes restrict their
+connections TLS v1.2 and also restrict which cipher suites are accepted.
```
mikehearn: connections _to_ TLS 1.2

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+Individual Node entries in the NetworkMap must be signed by the associated
+Node. The signatures are validated by the NetworkMap service, and all other
+Nodes in the network, to ensure they have not been tampered with. An attacker
+would need to acquire a node's private identity signing key to be able to make
+modifications to a NodeInfo. It is not possible for the NetworkMap service (or
+R3) to modify the NodeInfo entries, therefore an attacker would also be
+limited in this regard.
+
+The only impact the attacker could have would be to remove individual entries
+in the map, or insert their own NodeInfo.
```
mikehearn: The operator could change a NodeInfo by also issuing a new identity cert, right? There's nothing in the node today that flags unexpected key changes.

docs/source/threat-model/corda-threat-model.md
mikehearn: Yeah, isn't the thing that prevents repudiation here the signatures?

We could also mention that non-validating notaries require a signature over the request so nobody can deny doing a denial-of-state attack.

docs/source/threat-model/corda-threat-model.md
mikehearn: yes richard is right - you can request any tx in the transitive dependency graph of a transaction. It must be so otherwise you couldn't check tx validity.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+Denial of Service, DDoS). A successful DoS attack may result in non-
+availability of targeted ledger network node(s)/service(s), both during the
+attack and thereafter until normal service can be resumed.
+
+Communication over the ledger network is primarily peer-to-peer. Therefore the
+network as a whole is relatively resilient to DoS attacks. The primary threat
+is anticipated as being to the Corda network services which are depended on by
+network participants. These services will be protected by automated
+enterprise-grade DDoS detection and mitigation services, which will conduct
+intelligent traffic profiling in order to detect and block malicious traffic.
```
mikehearn: This sounds a bit vague. Perhaps we can say, network map is secured by CDNs and large-scale anti-DDoS networks. Notaries and oracles will only communicate with peers in the network, so are protected from non-member-on-member app level attacks.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+</td>  
+<td>
+
+The network agreement will stipulate a default maximum allowable period time -
+the 'event horizon' - within which a party is required to provide a valid
+response to any message sent to it in the course of a Flow. If that period is
+exceeded, the Flow will be considered to be cancelled and may be discontinued
+without prejudice by all parties. The event horizon may be superseded by
+agreements between parties specifying other timeout periods, which may be
+encoded into Flows under the Corda Flow Framework.
```
mikehearn: should flow be all lower case here?

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+
+Additional measures may be taken under the agreement against parties who
+repeatedly fail to meet their response obligations under the network
+agreement.
+
+</td></tr>  
+<tr>  
+<td>
+
+Doormap
```
mikehearn: typo: Doorman

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+
+</td>  
+<td>
+
+The AMQP serialiser uses a combination of white and black-lists to mitigate
+against de-serialisation vulnerabilities.
+
+Corda does not currently provide specific security controls to mitigate all
+classes of privilege escalation vulnerabilities. The design of Corda requires
+that CorDapps are inherently trusted by the node administrator. Likewise,
```
mikehearn: that *installed* CorDapps are trusted. We must be careful to distinguish here between contracts/states and flows.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+node-on-node attacks are not part of the formal threat model, and P2P traffic
+is assumed to originate from trusted network participants.
```
mikehearn: We *do* assume peers may be sending malformed data. We *don't* trust other network participants.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+
+Future security research will introduce stronger controls that can mitigate
+this class of threat. The Deterministic JVM will provide a sandbox that
+prevents execution of code & classes outside of the security boundary that
+contract code is restricted to.
+
+In addition, the SGX integration will provide an additional layer of security
```
mikehearn: SGX provides privacy but not security - it stops code getting in not out.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -0,0 +1,889 @@
...
+necessary to restrict this model to a high-level perspective of the Corda
+Network. It is hoped that enough information is provided to allow network
+participants to understand the security model of Corda.
+
+Threat modelling is an on-going process. There is active research at R3 to
+continue evolving the Corda Threat Model. In particular, models are being
+developed that focus more closely on individual components - such as the Node,
+Network Map and Doorman.
+
+
```
mikehearn: This is a good document. I'd like to see some aspects of the original document in perhaps a separate section, in particular enumerating what attacks each class of participant are assumed to be capable of. For instance what can and cannot the zone operator do? What about peers? What about notaries? etc. People need to be able to do reasoning of the form "if company X goes bad what's the damage"? The STRIDE framework seems like a good way to force thinking about different classes of attack, but I suspect from a business level security reasoning will be people oriented rather than attack class oriented.

mikehearn:
Also can you edit the toc/index files so it will be rendered as part of the docsite? Not sure which section it should go under ....

docs/source/threat-model/corda-threat-model.md
jamesbr3: updated & clarified

docs/source/threat-model/corda-threat-model.md
jamesbr3: i've included TLS authentication & transaction signing as mitigations

docs/source/threat-model/corda-threat-model.md
jamesbr3: yes ok. i copy-pasted as much as I could from the previous threat-model document - this bit was clearly out-of-date. fixed

docs/source/threat-model/corda-threat-model.md
jamesbr3: i copy-pasted from the old document :D

docs/source/threat-model/corda-threat-model.md
jamesbr3: fixed

docs/source/threat-model/corda-threat-model.md
jamesbr3: fixed

docs/source/threat-model/corda-threat-model.md
jamesbr3: good point. i've clarified this in the 'doorman' element further down

docs/source/threat-model/corda-threat-model.md
jamesbr3: Hmm OK I was under the impression that the current design was never intended to mitigate this. But I agree, it should be part of the threat model - i'll remvoe this part

docs/source/threat-model/corda-threat-model.md
jamesbr3: good point, i'll remove this from the EoP section - not relevant

docs/source/threat-model/corda-threat-model.md
```diff
@@ -228,14 +228,15 @@ Doorman
 </td>  
 <td>
 
-An attacker attempts to join the Corda Network by impersonating an existing
-organisation and issues a fraudulent registration request
+An malicious zone operator attempts to join the Corda Network by impersonating
```
mikehearn: I think "attacker" was right before. Zone operators can't join zones, they run them.

docs/source/threat-model/corda-threat-model.md
```diff
@@ -379,15 +380,16 @@ or removed from the map.
 <td>
 
 Individual Node entries in the NetworkMap must be signed by the associated
-Node. The signatures are validated by the NetworkMap service, and all other
-Nodes in the network, to ensure they have not been tampered with. An attacker
-would need to acquire a node's private identity signing key to be able to make
-modifications to a NodeInfo. It is not possible for the NetworkMap service (or
-R3) to modify the NodeInfo entries, therefore an attacker would also be
-limited in this regard.
+Node's private key. The signatures are validated by the NetworkMap service,
```
mikehearn: node to lowercase

docs/source/threat-model/corda-threat-model.md
```diff
@@ -632,11 +695,12 @@ availability of targeted ledger network node(s)/service(s), both during the
 attack and thereafter until normal service can be resumed.
 
 Communication over the ledger network is primarily peer-to-peer. Therefore the
-network as a whole is relatively resilient to DoS attacks. The primary threat
-is anticipated as being to the Corda network services which are depended on by
-network participants. These services will be protected by automated
-enterprise-grade DDoS detection and mitigation services, which will conduct
-intelligent traffic profiling in order to detect and block malicious traffic.  
+network as a whole is relatively resilient to DoS attacks. Notaries and
+Oracles will only communicate with peers in the network, so are protected from
```
mikehearn: oracles to lower case

docs/source/threat-model/corda-threat-model.md
```diff
@@ -763,9 +826,11 @@ Updates to the network map must be signed by participant nodes and are
 authenticated before being processed.
 
 The network map is designed to be distributed by a CDN (Content Delivery
-Network).
+Network). It is secured large-scale anti-DDoS networks.
```
mikehearn: This sentence isn't grammatical

mikehearn:
Also please remember to link it into the docsite toctree, and do a local re-render to ensure it looks good.

docs/source/threat-model/corda-threat-model.md
jamesbr3: fixed

docs/source/threat-model/corda-threat-model.md
jamesbr3: fixed

docs/source/threat-model/corda-threat-model.md
jamesbr3: fixed

docs/source/threat-model/corda-threat-model.md
jamesbr3: fixed

jamesbr3:
I've updated typos and have moved the threat model under the Design section of the docsite (and updated the toc). Just need to get it rendered locally

PR #3818, opened 2018-08-21
[CORDA-1917] Fix smoketests

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1917


mikehearn:
Why does this PR delete the BFT-SMaRT tests? That should be a separate PR with justification provided?

florian-f:
> Why does this PR delete the BFT-SMaRT tests? That should be a separate PR with justification provided?
@mikehearn I can put this in a separate PR, but it's basically just a test that was hanging and can be deleted according to the notaries team.

mikehearn:
Ah OK - in general please put justifications for things you heard in person into writing. Otherwise it's impossible to look back and understand why changes were made.

PR #3828, opened 2018-08-21
Stop bootstrapper errors caused by maxTransactionSize > maxMessageSize

anthonykeenan:
Since [this](https://github.com/corda/corda/commit/c0207b2219f3dd647f84da72452492d0c3fc15fb) commit, if maxTransactionSize > maxMessageSize the node will fail to start up.


mikehearn:
This looks right to me, but assigning @mkit because we discussed this recently.

anthonykeenan:
There was a similar issue in demobench which I fixed here: https://github.com/corda/corda/pull/3650

As far as I can see this only affects the 'new' cloud bootstrapper, not the classic bootstrapper.

mikehearn:
I'm not sure this is the right fix because transaction sizes *can* be larger than message sizes (max tx size includes the attachment size) but it unbreaks the bootstrapper tool so I'll merge.

PR #3832, opened 2018-08-22
Improve the error messages printed by the shell when a flow c'tor doesn't match

mikehearn:
This makes the constructor prototypes more concise when a flow doesn't match, by using just the class names of types that are in "well known" package namespaces:

shell-errors

### DIFF
--- a/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
+++ b/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
@@ -47,8 +47,7 @@ import java.io.FileDescriptor
 import java.io.FileInputStream
 import java.io.InputStream
 import java.io.PrintWriter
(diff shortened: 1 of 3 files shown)


mikehearn:
Is it necessary? That'd just duplicate the unit test, I'm not sure why we would want that - the type signatures aren't going to be different.

chrisr3:
Isn't the point of `InteractiveShellJavaTest` to test the behaviour of flows written in Java? And Java can use raw types, whereas Kotlin can't.

mikehearn:
I added a quick test to the Java version too, but it doesn't demonstrate much. Raw types for generic types like Amount don't work because they won't be recognised as the right type (which is correct).

How does it look now?

mikehearn:
I forgot that reflection isn't deterministic so the test was flaky. I changed the test and am requesting a re-review by @roger3cev

PR #3834, opened 2018-08-22
Propagate RPC deserialisation faults back to the caller

mikehearn:


mikehearn:
@distributedleetravis Sorry dude, can I ask for a re-review, I changed how this works to be less confusing and rubbish.

PR #3836, opened 2018-08-22
ENT-2202 Fold transaction mapping column into transactions table

rick-r3:


shamsasari:
What happens to nodes with existing data? Are we expecting O/S users to manage the db migration themselves? Slightly concerned there's no documentation for this change.

shamsasari:
OK ignore me, I see that we've introduced Liquibase!

PR #3839, opened 2018-08-22
Corda 1916: signature attachment constraints

distributedleetravis:
See [design document](https://github.com/corda/corda/blob/e8a15644297335544979fa281d4fa772d4cd782e/docs/source/design/data-model-upgrades/signature-constraints.md) for details of what this is trying to achieve.

We introduce a new type of `AttachmentConstraint`, `SignatureAttachmentConstraint`, which verifies that the attachment has signers that fulfil the provided `PublicKey` (which will typically be a `CompositeKey` stipulating multiple signers).

The function which obtains a list of signing parties for a given jar file is extracted to the object class `JarSignatureCollector` and independently tested in `JarSignatureCollectorTest`.

The `TransactionBuilder`'s automatic constraint selection is extended to load each output state's contract attachment and, if it has any signers, to associate the state with a matching `SignatureAttachmentConstraint`. A new test is added to `TransactionBuilderTest` to verify this behaviour.

Some otherwise unrelated tests which use mock services have to have some additional mocking behaviour defined, since all uses of `TransactionBuilder` now potentially involve trying to load each state's contract attachment's JAR data to find out what signers it has.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
@@ -3,6 +3,7 @@ package net.corda.core.contracts
 import net.corda.core.DoNotImplement
 import net.corda.core.KeepForDJVM
 import net.corda.core.contracts.AlwaysAcceptAttachmentConstraint.isSatisfiedBy
+import net.corda.core.crypto.CompositeKey
 import net.corda.core.crypto.SecureHash
 import net.corda.core.internal.AttachmentWithContext
 import net.corda.core.internal.isUploaderTrusted
@@ -66,4 +67,18 @@ object AutomaticHashConstraint : AttachmentConstraint {
     override fun isSatisfiedBy(attachment: Attachment): Boolean {
         throw UnsupportedOperationException("Contracts cannot be satisfied by an AutomaticHashConstraint placeholder")
     }
+}
+
+/**
+ * An [AttachmentConstraint] that verifies that the attachment has signers that satisfy the provided [CompositeKey].
+ * See: corda/docs/source/design/data-model-upgrades/signature-constraints.md
+ *
+ * @param key A [CompositeKey] that must be satisfied by the owning keys of the attachment's signing parties.
+ */
+@KeepForDJVM
+data class SignatureAttachmentConstraint(
+        val key: CompositeKey
+) : AttachmentConstraint {
+    override fun isSatisfiedBy(attachment: Attachment): Boolean =
+        key.isFulfilledBy(attachment.signers.map { it.owningKey })
 }
--- a/core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
@@ -5,9 +5,11 @@ import net.corda.core.KeepForDJVM
 import net.corda.core.contracts.AlwaysAcceptAttachmentConstraint.isSatisfiedBy
 import net.corda.core.crypto.CompositeKey
 import net.corda.core.crypto.SecureHash
+import net.corda.core.crypto.isFulfilledBy
 import net.corda.core.internal.AttachmentWithContext
 import net.corda.core.internal.isUploaderTrusted
 import net.corda.core.serialization.CordaSerializable
+import java.security.PublicKey
 
 /** Constrain which contract-code-containing attachment can be used with a [ContractState]. */
 @CordaSerializable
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
```diff
@@ -66,4 +69,18 @@ object AutomaticHashConstraint : AttachmentConstraint {
...
+
+/**
+ * An [AttachmentConstraint] that verifies that the attachment has signers that fulfil the provided [PublicKey].
+ * See: corda/docs/source/design/data-model-upgrades/signature-constraints.md
+ *
+ * @param key A [PublicKey] that must be fulfilled by the owning keys of the attachment's signing parties.
+ */
+@KeepForDJVM
+data class SignatureAttachmentConstraint(
+        val key: PublicKey
```
distributedleetravis: Why not `CompositeKey`, as advertised in the design doc? Because `CompositeKey` inherits from `PublicKey`, and `CompositeKey.Builder()...build()` returns a `PublicKey`.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -111,20 +109,21 @@ open class TransactionBuilder @JvmOverloads constructor(
...
-                    services.cordappProvider.getContractAttachmentID(state.contract)?.let {
-                        state.copy(constraint = HashAttachmentConstraint(it))
-                    } ?: throw MissingContractAttachments(listOf(state))
-                }
-            }
+            state.withConstraint(when {
+                state.constraint !== AutomaticHashConstraint -> state.constraint
+                useWhitelistedByZoneAttachmentConstraint(state.contract, services.networkParameters) ->
+                    WhitelistedByZoneAttachmentConstraint
+                else -> makeAttachmentConstraint(services, state)
```
distributedleetravis: Note that in this logic, `WhitelistedByZoneAttachmentConstraint` still has priority over `SignatureAttachmentConstraint`.

core/src/main/kotlin/net/corda/core/internal/JarSignatureCollector.kt
```diff
@@ -0,0 +1,67 @@
...
+    private val shredder = ByteArray(1024)
+
+    /**
+     * Returns an ordered list of every [Party] which has signed every signable item in the given [JarInputStream].
+     *
+     * @param jar The open [JarInputStream] to collect signing parties from.
+     * @throws InvalidJarSignersException If the signer sets for any two signable items are different from each other.
+     */
+    fun collectSigningParties(jar: JarInputStream): List<Party> {
+        val signerSets = jar.fileSignerSets
```
mikehearn: Nice style - using lots of private extension functions like this seems like a good way to write such logic. I could use a few more explicit type annotations here though (and on the return types of some of the functions below).

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -143,10 +142,28 @@ open class TransactionBuilder @JvmOverloads constructor(
...
+
+    private fun makeAttachmentConstraint(services: ServicesForResolution, state: TransactionState<ContractState>): AttachmentConstraint {
+        val attachmentId = services.cordappProvider.getContractAttachmentID(state.contract)
+            ?: throw MissingContractAttachments(listOf(state))
+        
+        val attachmentSigners = services.attachments.openAttachment(attachmentId)?.signers
+            ?: throw MissingContractAttachments(listOf(state))
+
+        return when {
+            attachmentSigners.isEmpty() -> HashAttachmentConstraint(attachmentId)
```
mikehearn: I think Tudor is working on code that affects this decision too.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -143,10 +142,28 @@ open class TransactionBuilder @JvmOverloads constructor(
...
+            else -> makeSignatureAttachmentConstraint(attachmentSigners)
+        }
     }
 
+    private fun makeSignatureAttachmentConstraint(attachmentSigners: List<Party>) =
+            SignatureAttachmentConstraint(CompositeKey.Builder().addKeys(attachmentSigners.map { it.owningKey }).build())
```
mikehearn: Can we simplify in the common case of a single signer to a non-CompositeKey?

core/src/test/kotlin/net/corda/core/internal/JarSignatureCollectorTest.kt
```diff
@@ -0,0 +1,159 @@
...
+    fun `all files must be signed by the same set of signers`() {
+        createJar("_signable1")
+        signAsAlice()
+        assertEquals(listOf(ALICE_NAME), getJarSigners().names)
+
+        updateJar("_signable2")
+        signAsBob()
+        assertFailsWith<InvalidJarSignersException>(
+            """
+            Mismatch between signers [O=Alice Corp, L=Madrid, C=ES, O=Bob Plc, L=Rome, C=IT] for file _signable1
```
mikehearn: Can we add a link from the error message to the docsite page that explains signature constraints? We've noticed a lot of people have trouble going from error messages to "but how do i solve this / what does it mean in general"

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
distributedleetravis: The `build` method already contains this logic:

```kotlin
fun build(threshold: Int? = null): PublicKey {
            require(threshold == null || threshold > 0)
            val n = children.size
            return when {
                n > 1 -> CompositeKey(threshold ?: children.map { (_, weight) -> weight }.sum(), children)
                n == 1 -> {
                    require(threshold == null || threshold == children.first().weight)
                    { "Trying to build invalid CompositeKey, threshold value different than weight of single child node." }
                    // Returning the only child node which is [PublicKey] itself. We need to avoid single-key [CompositeKey] instances,
                    // as there are scenarios where developers expected the underlying key and its composite versions to be equivalent.
                    children.first().node
                }
                else -> throw IllegalStateException("Trying to build CompositeKey without child nodes.")
            }
        }
```

distributedleetravis:
Have made requested changes to type signatures and comments / log messages. It seems that `CompositeKey.Builder()...build()` already does the right thing when only one key was supplied in the first place (i.e. just returns that key).

mikehearn:
Seems there's a build failure in the unit tests.

PR #3840, opened 2018-08-22
CORDA-1878: Improve exception handling when base directory is not readable or writable by the application user.

vkolomeyko:


vkolomeyko:
I was thinking of using `Try` (or `attempt`) construct, but given that in case of `IOException` we terminate JVM there is little point in doing that.

PR #3843, opened 2018-08-23
CORDA-1932 Fixing network map certificate path verification

mkit:
This PR fixes the certificate path verification process in case of the incoming network map data. It no longer assumes any specific certificate path length.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
@@ -7,15 +7,19 @@ import net.corda.core.serialization.CordaSerializable
 import net.corda.core.serialization.SerializedBytes
 import net.corda.core.serialization.deserialize
 import net.corda.core.utilities.OpaqueBytes
+import java.security.cert.CertPath
 import java.security.cert.X509Certificate
 
 // TODO: Rename this to DigitalSignature.WithCert once we're happy for it to be public API. The methods will need documentation
 // and the correct exceptions will be need to be annotated
 /** A digital signature with attached certificate of the public key. */
-class DigitalSignatureWithCert(val by: X509Certificate, bytes: ByteArray) : DigitalSignature(bytes) {
+open class DigitalSignatureWithCert(val by: X509Certificate, bytes: ByteArray) : DigitalSignature(bytes) {
     fun verify(content: ByteArray): Boolean = by.publicKey.verify(content, this)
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
```diff
@@ -7,16 +7,20 @@ import net.corda.core.serialization.CordaSerializable
...
 // TODO: Rename this to DigitalSignature.WithCert once we're happy for it to be public API. The methods will need documentation
 // and the correct exceptions will be need to be annotated
 /** A digital signature with attached certificate of the public key. */
-class DigitalSignatureWithCert(val by: X509Certificate, bytes: ByteArray) : DigitalSignature(bytes) {
+open class DigitalSignatureWithCert(val by: X509Certificate, bytes: ByteArray) : DigitalSignature(bytes) {
     fun verify(content: ByteArray): Boolean = by.publicKey.verify(content, this)
     fun verify(content: OpaqueBytes): Boolean = verify(content.bytes)
 }
 
+/** A digital signature with attached certificate path. The first certificate in the path corresponds to the data signer key. */
```
fenryka: If we're doing Java Docs lets list the properties as well

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkMap.kt
```diff
@@ -58,19 +56,12 @@ data class ParametersUpdate(
 )
 
 /** Verify that a Network Map certificate is issued by Root CA and its [CertRole] is correct. */
-// TODO: Current implementation works under the assumption that there are no intermediate CAs between Root and
-//      Network Map. Consider a more flexible implementation without the above assumption.
-
 fun <T : Any> SignedDataWithCert<T>.verifiedNetworkMapCert(rootCert: X509Certificate): T {
     require(CertRole.extract(sig.by) == CertRole.NETWORK_MAP) { "Incorrect cert role: ${CertRole.extract(sig.by)}" }
-    X509Utilities.validateCertificateChain(rootCert, sig.by, rootCert)
+    if (this.sig is DigitalSignatureWithCertPath) {
```
fenryka: do we need this?

fenryka:
This is going into Corda so can we move the ticket into the Corda Namespace in Jira

core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
mkit: Done.

PR #3844, opened 2018-08-23
CORDA-1934 Renaming INTERMEDIATE_CA certificate role to DOORMAN_CA cert…

mkit:
This PR renames the INTERMEDIATE_CA certificate role to DOORMAN_CA certificate role. It is required since our certificate hierarchy changes and the `intermediate` is too vague in that context.


fenryka:
Additionally.... isn't this documented as INTERMEDIATE_CA all over the code, so the docs need updating as well I think

mkit:
@fenryka regarding the compatibility issue. In the certificate we encode the ordinal number and not the name token so it should be fine. With respect to the INTERMEDIATE_CA name, there is another JIRA ticket to address this issue. Docs do not need to be updated with this change.

PR #3847, opened 2018-08-24
CORDA-1913 Revert back to using notion of "relevancy"

josecoll:
instead of "modifiable" (states are not modifiable by definition on an immutable ledger), 
plus we already expose the notion of **relevancy** in our public API via the `StatesToRecord`  `ONLY_RELEVANT` enum type.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/VaultService.kt
@@ -11,7 +11,7 @@ import net.corda.core.flows.FlowLogic
 import net.corda.core.identity.AbstractParty
 import net.corda.core.internal.concurrent.doneFuture
 import net.corda.core.messaging.DataFeed
-import net.corda.core.node.services.Vault.StateModificationStatus.*
+import net.corda.core.node.services.Vault.RelevancyStatus.*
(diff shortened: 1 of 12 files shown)


josecoll:
I think it is because of the interpretation attached to the new _relevancy_ flag in the database.
It means that those states flagged as relevant in the database (which may be spendable or just reference data) can be "**evolved**" (spent or copied & changed) by the Vault owner, whereas non-relevant states (e.g. reference data mastered by another participant) cannot.

The requirement that led to this overloaded usage of the term relevant is:
"ability to attach existing fungible states as reference data"
(used by @roger3cev in the [cash-on-ledger design and reference application](https://github.com/corda/cash))

mikehearn:
I'll remove myself as a reviewer on this one and let you figure it out with Roger.

rick-r3:
Happy to see the naming unified with the terminology in `StatesToRecord`.  Roger had missed a rename in his original PR and it was still `isRelevant` there rather than `isModifiable` on `LinearStateQueryCriteria`.  Looks like you spotted that too.

roger3cev:
Good job chaps! Can’t approve as on train. Will look when I arrive...

Get Outlook for iOS<https://aka.ms/o0ukef>

Issue #3804, opened 2018-08-17
FR: Support for case-insensitive vault queries


joeldudleyr3:
The vault query API does not support comparing strings in a case-insensitive way. For example, the following query will be case-sensitive:

```
searchCriteria = builder { 
    BondContract.BondEntity::customerName.`in`(listOf(name1, name2))
}
```

The vault query API should be extended to allow case-insensitive vault queries.

lankydan:
I got this 👍

joeldudleyr3:
Fixed :)

PR #3853, opened 2018-08-28
Support for case insensitive vault queries

lankydan:
I have allowed the SQL operation criteria builders of EQUAL, NOT EQUAL, IN, NOT IN, LIKE and NOT LIKE to be able to specify whether they are case sensitive or not. By default they will be case sensitive and by specifying the `exactMatch` argument as `false` it can be made case insensitive.

Below is a summary of the changes.

Add IGNORE_CASE versions of the comparison operator enums
Add exactMatch argument to criteria builder functions where strings can be passed in and set its default value to true
Use JvmOverrides to provide the default true version to java without needing to specify a value manually
If exactMatch is true then the original enums will be used, if false the IGNORE_CASE enums will be used instead
HibernateQueryCriteriaParser.columnPredicateToPredicate now takes into account the IGNORE_CASE versions of the enums

Additional note, I have used enums to specify whether the commands are case sensitive or not. Maybe this should be done with the addition of boolean values to the criteria parser but I think that might require larger changes. But, this might be the more correct solution.


lankydan:
Created to resolve issue [https://github.com/corda/corda/issues/3804](https://github.com/corda/corda/issues/3804)

josecoll:
Please can you add a one liner description to the Changelog, and update the vault query documentation accordingly (eg. new operators and usage example with case insensitive operator).
1. `changelog.rst` (under section "unreleased")
2. `api-vault-query.rst` (scroll to section that describes `QueryCriteria` operators)

lankydan:
just looking at the `api-vault-query` docs. Should the custom vault criteria section be split off into its own section? Its a got content in it and works slightly different to the other criterias.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -24,7 +28,9 @@ enum class BinaryLogicalOperator : Operator {
 
 enum class EqualityComparisonOperator : Operator {
     EQUAL,
-    NOT_EQUAL
+    EQUAL_IGNORE_CASE,
```
mikehearn: I think the new ones have to go at the end of the enums:

https://docs.corda.net/head/serialization-enum-evolution.html?highlight=enum#adding-constants

We don't need full evolution support here because the RPC library will check that the node is at least platform version 4 or higher, so the client should always understand the server's enums, but it's still a good idea not to re-order them in case someone is using the ordinals.

lankydan: Looking at it now though. I understand it could possible someone could be depending on the ordinals but what could they really be doing with them? I think maybe these enums should have used the `internal` modifier so people just rely on the functions provided to them (`equals`, `greaterThan` etc...).  Is it too late to do this? If so I can make the enum order change.

mikehearn: Unfortunately we've exposed it in the API now, so it's possible someone is using them. We could try to phase them out of the API but I'd rather bundle that up with a more general project to shrink the API surface area.

It looks like we have a new conflict on the changelog, so if you could rebase and reorder, that'd be great and I'll merge ASAP.

lankydan: Got it, can sort this out later.

lankydan:
can we merge this so i dont have to keep resolving conflicts in the changelog file :)

PR #3854, opened 2018-08-28
CORDA-1938 Adding version info to the node info submission request

mkit:
This PR adds version info to the node info submission request.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
+++ b/node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
@@ -11,6 +11,7 @@ import net.corda.core.serialization.serialize
 import net.corda.core.utilities.contextLogger
 import net.corda.core.utilities.seconds
 import net.corda.core.utilities.trace
+import net.corda.node.VersionInfo
 import net.corda.node.utilities.registration.cacheControl
 import net.corda.nodeapi.internal.SignedNodeInfo
 import net.corda.nodeapi.internal.network.NetworkMap
@@ -23,7 +24,7 @@ import java.security.cert.X509Certificate
 import java.time.Duration
 import java.util.*
 
-class NetworkMapClient(compatibilityZoneURL: URL) {
+class NetworkMapClient(compatibilityZoneURL: URL, private val versionInfo: VersionInfo) {
     companion object {
         private val logger = contextLogger()
     }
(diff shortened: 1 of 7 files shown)


node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -45,7 +46,9 @@ class NetworkMapClient(compatibilityZoneURL: URL) {
     fun ackNetworkParametersUpdate(signedParametersHash: SignedData<SecureHash>) {
         val ackURL = URL("$networkMapUrl/ack-parameters")
         logger.trace { "Sending network parameters with hash ${signedParametersHash.raw.deserialize()} approval to $ackURL." }
-        ackURL.post(signedParametersHash.serialize())
+        ackURL.post(signedParametersHash.serialize(),
+                "Platform-Version" to "${versionInfo.platformVersion}",
```
kasiastreich: Platform version is stored in `NodeInfo` already

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
```diff
@@ -45,7 +46,9 @@ class NetworkMapClient(compatibilityZoneURL: URL) {
...
+                "Client-Version" to versionInfo.releaseVersion)
```
kasiastreich: Also, do we need that much granularity?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
mkit: Yup but for the consistency with the CSR submission we unify the headers.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
mkit: We might in the future.

fenryka: We do for the cert version change, plus it's potentially useful

node/src/main/kotlin/net/corda/node/services/network/NetworkMapClient.kt
fenryka: useful to have in the header though so we don't have to pull apart the blob before deciding what to do

PR #3855, opened 2018-08-28
CORDA-1939 Regression: Liquibase fails to migrate existing pre-liquibase database

szymonsztuka:
Recently, Liquibase was added to support schema modification to internal Corda tables.
The issue occurred when for initial Liquibase setup for the existing H2 database, the coda-finance module was not up to date (were missing Liquibase migration scripts) and the migration failed leaving database in inconsistent state.
The fix:
* corda.jar can be manually deployed without coping the matching corda-finance-VERSION.jar.
If  the older finance cordapp version is detected (which doesn't have Liquibase migration scripts) fail node at startup and print message:
"Could not create the DataSource: Detected incompatible corda-finance cordapp without database migration scripts, replace the existing corda-finance-VERSION.jar with the latest one."
Since coda-finance is an optional cordapp,the  presence of Liquibase scripts is only checked if corda-finance-VERSION.jar is present in cordapps folder.

Unrelated change to the bug fix:
* Allow to start using Liquibase from any point of 4.0-SNAPSHOT before Liquibase was introduced (not only from 3.0/3.X versions) - 2 database changes introduced after 3.2 but before Liquibase are now conditional database changes

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
@@ -106,10 +106,16 @@ class SchemaMigration(
 
     /** For existing database created before verions 4.0 add Liquibase support - creates DATABASECHANGELOG and DATABASECHANGELOGLOCK tables and mark changesets are executed. */
     private fun migrateOlderDatabaseToUseLiquibase(existingCheckpoints: Boolean): Boolean {
+        //workaround to detect that if Corda finance module is in use then the most recent version is required, which contains Liquibase migration scripts
+        if (schemas.any { schema ->
+                    (schema::class.qualifiedName == "net.corda.finance.schemas.CashSchemaV1" || schema::class.qualifiedName == "net.corda.finance.schemas.CommercialPaperSchemaV1")
+                            && schema.migrationResource == null
+                })
+            throw DatabaseMigrationException("Could not migrate database, detected incompatible corda-finance cordapp, ensure to replace the existing corda-finance-XX.jar with the latest one matching Corda version.")
+
         val isExistingDBWithoutLiquibase = dataSource.connection.use {
-            it.metaData.getTables(null, null, "NODE%", null).next() &&
-                    !it.metaData.getTables(null, null, "DATABASECHANGELOG", null).next() &&
-                    !it.metaData.getTables(null, null, "DATABASECHANGELOGLOCK", null).next()
+            (it.metaData.getTables(null, null, "NODE%", null).next() &&
+                    !it.metaData.getTables(null, null, "DATABASECHANGELOG%", null).next())
         }
         when {
             isExistingDBWithoutLiquibase && existingCheckpoints -> throw CheckpointsException()
@@ -119,7 +125,7 @@ class SchemaMigration(
 
                 dataSource.connection.use { connection ->
                     // Schema migrations pre release 4.0
-                    val node = listOf("migration/common.changelog-init.xml",
+                    val preV4Baseline = mutableListOf("migration/common.changelog-init.xml",
                             "migration/node-info.changelog-init.xml",
                             "migration/node-info.changelog-v1.xml",
                             "migration/node-info.changelog-v2.xml",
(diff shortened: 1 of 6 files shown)


node/src/main/kotlin/net/corda/node/services/schema/NodeSchemaService.kt
```diff
@@ -69,7 +69,7 @@ class NodeSchemaService(private val extraSchemas: Set<MappedSchema> = emptySet()
      if (includeNotarySchemas) mapOf(Pair(NodeNotaryV1, SchemaOptions())) else emptyMap()
 
     fun internalSchemas() = requiredSchemas.keys + extraSchemas.filter { schema -> // when mapped schemas from the finance module are present, they are considered as internal ones
-        schema::class.simpleName == "net.corda.finance.schemas.CashSchemaV1" || schema::class.simpleName == "net.corda.finance.schemas.CommercialPaperSchemaV1" }
+        schema::class.qualifiedName == "net.corda.finance.schemas.CashSchemaV1" || schema::class.qualifiedName == "net.corda.finance.schemas.CommercialPaperSchemaV1" }
```
josecoll: How did the code ever work before if this conditional was incorrect?

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
```diff
@@ -119,29 +125,32 @@ class SchemaMigration(
...
+                                "migration/cash.changelog-v1.xml"))
+
+                    if (schemas.any { schema -> schema.migrationResource == "commercial-paper.changelog-master" })
+                        preV4Baseline.addAll(listOf("migration/commercial-paper.changelog-init.xml",
+                                "migration/commercial-paper.changelog-v1.xml"))
+
+                    if (schemas.any { schema -> schema.migrationResource == "node-notary.changelog-master" })
+                        preV4Baseline.addAll(listOf("migration/node-notary.changelog-init.xml",
+                                "migration/node-notary.changelog-v1.xml",
+                                "migration/vault-schema.changelog-pkey.xml"))
```
josecoll: Seems rather odd that the Vault Schema is updated differently for a Notary node.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
```diff
@@ -106,10 +106,16 @@ class SchemaMigration(
 
     /** For existing database created before verions 4.0 add Liquibase support - creates DATABASECHANGELOG and DATABASECHANGELOGLOCK tables and mark changesets are executed. */
     private fun migrateOlderDatabaseToUseLiquibase(existingCheckpoints: Boolean): Boolean {
+        //workaround to detect that if Corda finance module is in use then the most recent version with Liquibase migration scripts was deployed
+        if (schemas.any { schema ->
+                    (schema::class.qualifiedName == "net.corda.finance.schemas.CashSchemaV1" || schema::class.qualifiedName == "net.corda.finance.schemas.CommercialPaperSchemaV1")
```
josecoll: I'm not a fan of using FQCN in strings.
Perhaps use `CashSchemaV1::class.qualifiedName`

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
szymonsztuka: Indeed this is wrong, vault-schema is already included before, so removing it.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
szymonsztuka: The problem here (and in the other place) is that node-api doesn't know about finance module, but SchemaMigration needs to know that MappedSchemas from finance are tracked by Liquibase.

szymonsztuka:
yes, it introduces  tighter coupling  between node and finance but not compile one, it comes down to marking MappedSchemas from optional Finance Cordapp to migrate via Liquibase, there could be other ways to do it  though (e.g. a subclass `InternalMappedSchema` )

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
josecoll: But why is `SchemaMigration` in the `node-api` module if it is solely used internally by the node?

szymonsztuka: This is to be consistent with Enterprise repo.

PR #3856, opened 2018-08-28
[CORDA-1937]: Refactor NodeConfiguration hierarchy.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1937

- Removed `NodeSSLConfiguration`
- Changed the relationship between `NodeConfiguration` and `SSLConfiguration` from inheritance to delegation
- Introduced a more abstract `SslConfiguration` to replace `SSLConfiguration`
- Introduced `CertificateStore` as a more pleasant way of handling certificates
- Introduced test stubs for certificate stores
- Fixed bugs around certificates with Shell, WebServer and DriverDSLImpl

Some findings that came up while working:

1. It's currently not possible to specify individual paths for cert stores.
2. It's currently not possible to have a different password for the P2P and signing cert stores.
3. Shell and Driver are using the P2P certificates over adminAddress with an internal RPC client to talk to the node, instead of providing a trust-store that includes the cert in the RPC key-store.
4. RPC client SSL has been maimed to only support keyStore for node and trustStore for client (no other way around, no two-way SSL).
5. Webserver is using the signing certificate for HTTPS.

Not much I can currently do about point 4, for public API is involved, but all other points are either fixed or made easy to fix by this PR.


josecoll:
A lot of this refactoring is unnecessary and obfuscates readability (eg. unnecessary lambdas/renaming, companion object imports, re-defined basic data structures such as Iterator). 

Reviewers of this PR should include at least experienced/major contributors to refactored modules (messaging, security) - @mnesbit @kchalkias 

It is high time we introduced some developer practice guidelines around when/what/how major refactorings take place, and assign some controls around proposal/review/sign-off of these types of PR (eg. balance value/effort vs ongoing maintenance, readability, re-learning curve, etc).

sollecitom:
Removed `MappingIterator`, by replacing it with appropriate usage of `Sequence`. Thanks @distributedleetravis for the hint.

Clintonio:
Just to weigh in but I think most of what I saw looked fine on the readability front and in line with the sort of code we already have in the codebase. It also seems to improve various things as well.

node-api/src/main/kotlin/net/corda/nodeapi/internal/DevIdentityGenerator.kt
```diff
@@ -31,15 +32,15 @@ object DevIdentityGenerator {
...
-        val nodeSslConfig = object : NodeSSLConfiguration {
-            override val baseDirectory = nodeDir
-            override val keyStorePassword: String = "cordacadevpass"
-            override val trustStorePassword get() = throw NotImplementedError("Not expected to be called")
-            override val crlCheckSoftFail: Boolean = true
-        }
+        val certificatesDirectory = nodeDir / "certificates"
+        val signingCertStore = FileBasedCertificateStoreSupplier(certificatesDirectory / "nodekeystore.jks", "cordacadevpass")
+        val p2pKeyStore = FileBasedCertificateStoreSupplier(certificatesDirectory / "sslkeystore.jks", "cordacadevpass")
+        val p2pTrustStore = FileBasedCertificateStoreSupplier(certificatesDirectory / "truststore.jks", "trustpass")
```
kchalkias: Referring to the naming: We used to share this `truststore` between signing and TLS (p2p) services, so technically it belongs to both. Also, I would prefer to have to distinct truststores per service, but this is another PR.

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStore.kt
```diff
@@ -0,0 +1,82 @@
...
+import net.corda.nodeapi.internal.crypto.X509KeyStore
+import net.corda.nodeapi.internal.crypto.X509Utilities
+import net.corda.nodeapi.internal.crypto.addOrReplaceCertificate
+import java.io.InputStream
+import java.io.OutputStream
+import java.nio.file.OpenOption
+import java.nio.file.Path
+import java.security.cert.X509Certificate
+
+interface CertificateStore : Iterable<Pair<String, X509Certificate>> {
```
kchalkias: I would recommend 1-2 docs-lines, to explain why this is required.

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/SslConfiguration.kt
```diff
@@ -4,12 +4,34 @@ import net.corda.core.internal.div
...
+    companion object {
+
+        fun twoWay(keyStore: FileBasedCertificateStoreSupplier, trustStore: FileBasedCertificateStoreSupplier): TwoWaySslConfiguration {
+
+            return TwoWaySslOptions(keyStore, trustStore)
+        }
+    }
+}
+
+interface TwoWaySslConfiguration : SslConfiguration {
```
mnesbit: This is a misnomer. All TLS is two way. You mean client authenticating TLS, or mutally authenticating.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPServer.kt
```diff
@@ -60,7 +60,7 @@ class AMQPServer(val hostName: String,
         private val conf = parent.configuration
 
         init {
-            keyManagerFactory.init(conf.keyStore, conf.keyStorePrivateKeyPassword)
+            keyManagerFactory.init(conf.keyStore)
```
mnesbit: No this is needed in ENT

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStore.kt
sollecitom: Sure, I can add. In general we don't KDoc internal types, but will do.

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/SslConfiguration.kt
mnesbit: This should be CertificateStoreProvider, not FileBasedCertificateStoreProvider as I do have other types

kchalkias: Note that both 2way and mutual TLS are acceptable, but yeah we usually refer to it as mutual (mTLS).

sollecitom:
@mnesbit , as requested: https://github.com/corda/enterprise/compare/ports/CORDA-1937-ENT

node/src/main/kotlin/net/corda/node/services/config/shell/ShellConfig.kt
```diff
@@ -14,8 +15,8 @@ fun NodeConfiguration.toShellConfig() = ShellConfiguration(
         cordappsDirectory = this.baseDirectory.toString() / CORDAPPS_DIR,
         user = INTERNAL_SHELL_USER,
         password = INTERNAL_SHELL_USER,
-        hostAndPort = this.rpcOptions.adminAddress,
-        nodeSslConfig = this,
+        hostAndPort = this.rpcOptions.address,
+        ssl = clientSslOptionsCompatibleWith(this.rpcOptions),
```
tlil: Double check that this doesn't interfere with SSL vs non-SSL configs? Have you manually tested this as well? I.e. does `clientSslOptionsCompatibleWith()` ever return `null` when it shouldn't?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -69,9 +71,16 @@ interface NodeConfiguration : NodeSSLConfiguration {
     val effectiveH2Settings: NodeH2Settings?
     val flowMonitorPeriodMillis: Duration get() = DEFAULT_FLOW_MONITOR_PERIOD_MILLIS
     val flowMonitorSuspensionLoggingThresholdMillis: Duration get() = DEFAULT_FLOW_MONITOR_SUSPENSION_LOGGING_THRESHOLD_MILLIS
-    val cordappDirectories: List<Path> get() = listOf(baseDirectory / CORDAPPS_DIR_NAME_DEFAULT)
+    val crlCheckSoftFail: Boolean
```
tlil: Just to make sure this doesn't have any adverse effects on CRL checking, is the default value of this config option the same as prior to the refactor?

PR #3859, opened 2018-08-29
CORDA-1945: properly support double-width interp stack slots in superclasses when synthesising

distributedleetravis:
The following test fails, because the `ClassCarpenter` incorrectly calculates the stack size of fields if there are double-size primitive fields (e.g. long or double) in a superclass constructor:

```kotlin
    @Test
    fun `superclasses with double size primitive constructor parameters`() {
        val schema1 = ClassSchema(
                "gen.A",
                mapOf("a" to NonNullableField(Long::class.javaPrimitiveType!!)))

        val schema2 = ClassSchema(
                "gen.B",
                mapOf("b" to NonNullableField(String::class.java)),
                schema1)

        val clazz = cc.build(schema2)
        val i = clazz.constructors[0].newInstance(1L, "xb") as SimpleFieldAccess
        assertEquals(1L, i["a"])
        assertEquals("xb", i["b"])
        assertEquals("B{a=1, b=xb}", i.toString())
    }
```

The fault is in `ClassWriter.generateClassConstructor`, and the solution is to sum the sizes of the fields in the superclass constructor and make sure this value is used to set the initial stack slot correctly for the parameters.

### DIFF
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/carpenter/ClassCarpenter.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/carpenter/ClassCarpenter.kt
@@ -139,7 +139,7 @@ class ClassCarpenterImpl @JvmOverloads constructor (override val whitelist: Clas
             hierarchy += cursor
             cursor = cursor.superclass
         }
-         
+
         hierarchy.reversed().forEach {
             when (it) {
                 is InterfaceSchema -> generateInterface(it)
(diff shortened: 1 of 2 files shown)


serialization/src/main/kotlin/net/corda/serialization/internal/carpenter/ClassCarpenter.kt
```diff
@@ -130,16 +130,16 @@ class ClassCarpenterImpl @JvmOverloads constructor (override val whitelist: Clas
      */
     override fun build(schema: Schema): Class<*> {
         validateSchema(schema)
-        // Walk up the inheritance hierarchy and then start walking back down once we either hit the top, or
-        // find a class we haven't generated yet.
+        // Walk up the inheritance hierarchy until we hit either the top or a class we've already generated,
```
distributedleetravis: The traversal of the class hierarchy actually stops when we find a class that's _already_ been generated.

PR #3862, opened 2018-08-29
Small RPC improvements

mikehearn:
* Enable trackCallSites to be enabled from the command line using a Java property.
* Mark as deprecated a confusing configuration option that wasn't used anywhere anyway.
* Improve the docs by specifying the defaults.
* RPC will throw at startup if you're using the v4 library against a v3 server, rather than waiting until you invoke a bad RPC (better to fail early than half way through an important operation).

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
@@ -49,6 +49,12 @@ internal class CordaRPCOpsImpl(
         private val flowStarter: FlowStarter,
         private val shutdownNode: () -> Unit
 ) : CordaRPCOps {
+    /**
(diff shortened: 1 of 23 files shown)


mikehearn:
The API checker is complaining because I removed an apparently needless interface override, I think it's a false positive and will investigate.

node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
```diff
@@ -38,6 +38,7 @@ import java.io.InputStream
 import java.net.ConnectException
 import java.security.PublicKey
 import java.time.Instant
+import java.util.concurrent.TimeUnit
```
the-root: Is this import used?

mikehearn: Good spot! It's not - left over from a commit split. I'll remove it.

Issue #2913, opened 2018-04-03
Flows cannot explicitly request a transaction be notarized


siddtwm:
In corda-core (**release-V3.0** on commit  **d9d6684**), the FinalityFlow determines whether the signed transaction should be notarized based on presence of inputs, time window and by the presence of the notary's key in the list of signers. This approach does not allow the flow creation process to explicitly request notarization of the transaction. 

The presence of the notary's signature will be helpful if any of the participants on the newly generated contract state demand that every newly issued contract be signed off by the validator.

Running any flow that only creates a single contract state while the validator node is not running demonstrates this issue.

mikehearn:
So to clarify you want to notarise a transaction that is purely an issuance, even though it doesn't have a timestamp?

The reason we don't allow this is that the notary wouldn't check or do anything. It'd just sign, but the signature wouldn't mean anything (unless it was a validating notary, in which case it'd have validated, but the recipient is supposed to do that anyway).

Can you explain a bit more about why you'd like to do this? Does it meet some business requirement we overlooked?

joeldudleyr3:
@siddtwm See Mike's questions above :)

siddtwm:
Yes, indeed. We only use the validating notary in our use-case. From our perspective issuance notarization has to be one of the functions validator provides - ensure that all new contract states created by participants meet our definition of state instance. This is, of course over and above the functions Corda's validating notary already has. Further, if we introduce special audit notaries, they too may have custom logic around what constitutes a valid contract state instance.

mikehearn:
@siddtwm If a party issues a state, then it resides only in their own vault and thus nobody else cares. If they then send it to a counterparty then that tx will be validated by the notary and only signed if it's valid, so the validity of the move tx captures the validity of the issuance.

So I'm still not quite sure what notarisation of issuances without timestamps will accomplish.

lankydan:
Since this has received no further feedback, would it be best to pose this as a question to the dev mailing list and see what others think? @mikehearn

roger3cev:
Also don't understand the need for a signature over a transaction with no input states or timestamp. Given that the notary only checks for input state uniqueness and/or acts as a time stamping authority, if it signs a transaction with a absence of both, then it doesn't mean anything. 

In terms of validation, all participants do this anyway. Maybe I'm missing something?

PR #3863, opened 2018-08-29
Add forceNotarySignature to FinalityFlow to provide a way for a developer to force the notary to sign a transaction

lankydan:
This change is in response to issue [https://github.com/corda/corda/issues/2913](https://github.com/corda/corda/issues/2913).

If it is decided that this is not something that should be done, then just reject this PR.

`forceNotarySignature` has been added to `FinalityFlow` so that a developer can decide whether to notarise a transaction that doesn't require notarisation (base on transaction's properties).

Extra constructors were added and the existing ones simply set `forceNotarySignature` to `false`.

The downside to this code is the ugliness of all the constructors that `FinalityFlow` now has.

If this solution of allowing devs to force notarisation, then let me know and I can add some extra tests if required as I have only put a very basic test in place at the moment.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
@@ -28,10 +28,15 @@ import net.corda.core.utilities.ProgressTracker
 @InitiatingFlow
 class FinalityFlow(val transaction: SignedTransaction,
                    private val extraRecipients: Set<Party>,
+                   private val forceNotarySignature: Boolean,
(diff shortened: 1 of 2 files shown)


mikehearn:
Hmm, thanks, but I've gone back and asked some more questions now on the original issue as I'm not entirely sure still why it's useful to do this.

lankydan:
ye, thats fair enough. Didn't want to spend to much time on this in case we decided not to both anyway

lankydan:
based on the comments on the issue, should we close this PR?

Issue #3803, opened 2018-08-17
FR: Check Java version on node start-up and abort if unsupported


joeldudleyr3:
Corda only supports Oracle JDK 8 JVM, with a minimum supported version of 8u171 (see https://docs.corda.net/getting-set-up.html#software-requirements).

However, no check is currently performed at start-up, and using the wrong Java version often leads to opaque error messages.

The Java version should be checked when the node starts up.

opticyclic:
The documentation says "JVM implementations other than Oracle JDK 8 are not actively supported. However, if you do choose to use OpenJDK, you will also need to install OpenJFX".

That means that the code should work with OpenJDK if you are happy to not have support.
The Azure templates have Azul Zulu as the runtime and not Oracle.

Checking for the Java version on startup or vendor isn't the right fix here.

lankydan:
fair point, ill revert that commit

PR #3870, opened 2018-08-30
Check java version is above 1.8.0_171 on node startup

lankydan:
In response to issue [https://github.com/corda/corda/issues/3803](https://github.com/corda/corda/issues/3803).

Checks the minimum version is 1.8.0_171 and that it that its vendor is Oracle.

I have updated also update the error message to state it must use Java 8 (this can be reverted if required).

I have not tested this above a Java 8 version since I believe Corda is relying on internal sun packages that have changed in higher java versions. For example, I could not even build the Corda project when I moved up to Java 10 and could not run `deployNodes` on an external project when using Java 10.


lankydan:
reset to only checking the java version rather than checking that it uses an oracle jdk

lankydan:
@joeldudleyr3 Can this get reviewed sometime soon? I'll fix the conflict later today.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -491,14 +492,10 @@ open class NodeStartup(val args: Array<String>) {
...
-        return try {
-            Paths.get("").normalize()
-            true
-        } catch (e: ArrayIndexOutOfBoundsException) {
-            false
-        }
+    private fun hasMinimumJavaVersion(): Boolean {
+        val major = SystemUtils.JAVA_VERSION_FLOAT
+        val update = SystemUtils.JAVA_VERSION.substringAfter("_").toLong()
+        return major > 1.8 || (major == 1.8F && update >= 171)
```
shamsasari: Can you add a comment here that if this changes then the `ext.java8_minUpdateVersion` line in build.gradle also has to change, and vice-versa.

lankydan:
another comment, me questioning how i've done it myself. I have checked that it is above the minimum java 8 version, but should I also check that the version is java 8 and not 9+. This is because I cannot build my node with java 9+. It just doesnt work. Thoughts?

shamsasari:
Building or running? But yes I think it doesn't run on Java 9+ either, so perhaps best to stick to 8.

lankydan:
i think it failed in just building the nodes (using `deployNodes`)

lankydan:
ill change the check to be must be java 8 then

lankydan:
haha that rebase does not look good.... Well, I fixed the conflict atleast...

build.gradle
```diff
@@ -78,6 +77,7 @@ buildscript {
 
     // Update 121 is required for ObjectInputFilter.
     // Updates [131, 161] also have zip compression bugs on MacOS (High Sierra).
+    // when the java version in NodeStartup.hasMinimumJavaVersion() changes, so much this check
```
shamsasari: typo: must

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -49,80 +52,108 @@ import java.util.*
...
+            println("You are using a version of Java that is not supported (${SystemUtils.JAVA_VERSION}). Please upgrade to the latest version of Java 8.")
+            println("Corda will now exit...")
+            return false
+        }
+        return true
+    }
 
-        logStartupInfo(versionInfo, cmdlineOptions, configuration)
+    private fun hasMinimumJavaVersion(): Boolean {
+        // when the ext.java8_minUpdateVersion gradle constant changes, so much this check
```
shamsasari: ditto

PR #3872, opened 2018-08-30
CORDA-1755: Modify the node to run using picocli

anthonykeenan:
* Uses the new picocli base class shared with network bootstrapper
* Fixed the `--sshd` option which never worked and added a `--sshd-port` option
* Added documentation on how to install and upgrade shell extensions
* Made sure the documentation of the node command line options was up to date with the help output
* Added some POSIX style short options for some commands (my idea here was to add them for commands that might be ran on the node regularly, as opposed to one time 'start-up' type options - was going to add that to the UX guidelines)
* Made some slight modifications to the base class
  * Made `initLogging` overridable as the Node logging is slightly more complex than the bootstrapper
  * Moved the `initLogging` function to be the first thing that runs, as it was causing the logging options not to work
  * Combined the --log-to-console and --verbose options in the base class as they were doing the same thing
  * Made the run procedure return an Int which is used as the exit code for the app
  * Added the `AnsiConsole.systemInstall()` so all apps have the same cmdline settings by default
  * Check for a `caplet.jar` property and if so use that instead of the running jar location


docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
 
-* ``--base-directory``: The node working directory where all the files are kept (default: ``.``)
+* ``--base-directory``, ``-b``: The node working directory where all the files are kept (default: ``.``)
 * ``--bootstrap-raft-cluster``: Bootstraps Raft cluster. The node forms a single node cluster (ignoring otherwise configured peer 
   addresses), acting as a seed for other nodes to join the cluster
-* ``--config-file``: The path to the config file (default: ``node.conf``)                          
-* ``--help``
+* ``--clear-network-map-cache``, ``-c``: Clears local copy of network map, on node startup it will be restored from server or file system.
+* ``--config-file``, ``-f``: The path to the config file. Defaults to ``node.conf``.
+* ``--dev-mode``, ``-d``: Runs the node in developer mode. Unsafe in production. Defaults to true on MacOS and desktop versions of Windows. False otherwise.
```
sollecitom: "Defaults to true on MacOS and desktop versions of Windows. False otherwise.", it's not true. It defaults to false, period. There's a third option - other than true or false - called "auto", which will then behave like you described.

node/src/main/java/CordaCaplet.java
```diff
@@ -21,7 +21,7 @@ protected CordaCaplet(Capsule pred) {
 
     private Config parseConfigFile(List<String> args) {
         String baseDirOption = getOption(args, "--base-directory");
-        // Ensure consistent behaviour with NodeArgsParser.kt, see CORDA-1598.
+        // Ensure consistent behaviour with NodeCmdLineOptions.kt, see CORDA-1598.
```
sollecitom: Is this still relevant? https://r3-cev.atlassian.net/browse/CORDA-1598 is `done`.

node/src/main/kotlin/net/corda/node/Corda.kt
```diff
@@ -3,12 +3,13 @@
...
+import net.corda.cliutils.start
 import net.corda.node.internal.NodeStartup
-import kotlin.system.exitProcess
 
 fun main(args: Array<String>) {
     // Pass the arguments to the Node factory. In the Enterprise edition, this line is modified to point to a subclass.
     // It will exit the process in case of startup failure and is not intended to be used by embedders. If you want
     // to embed Node in your own container, instantiate it directly and set up the configuration objects yourself.
-    exitProcess(if (NodeStartup(args).run()) 0 else 1)
+    //exitProcess(if (NodeStartup(args).run()) 0 else 1)
```
sollecitom: Remove the commented line.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+    )
+    var baseDirectory: Path = Paths.get(".")
+
+    @Option(
+            names = ["-f", "--config-file"],
+            description = ["The path to the config file."]
+    )
+    var configFileArgument: Path? = null
+
+    val configFile : Path
```
sollecitom: Can we try and make this optional? At the moment it's impossible to start a node without a config file - as in, by using VM options or ENV variables only - and it's quite sad.

sollecitom: Perhaps a different story.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+        get() {
+            return if (configFileArgument == null) {
+                baseDirectory / "node.conf"
```
sollecitom: Do we need the default value to be part of the arguments parser? Can we not return an optional type and let calling code decide what to do?

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+            } else {
+                configFileArgument!!
+            }
+        }
+
+    @Option(
+            names = ["--sshd"],
+            description = ["If set, enables sshd server for node administration."]
+    )
+    var sshdServer: Boolean = false
```
sollecitom: Well, I see that having default values in here is a pattern. I'd rather have them outside, but not sure if that would be too problematic.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -370,12 +352,13 @@ open class NodeStartup(val args: Array<String>) {
         logger.info("PID: ${info.name.split("@").firstOrNull()}")  // TODO Java 9 has better support for this
         logger.info("Main class: ${NodeConfiguration::class.java.location.toURI().path}")
         logger.info("CommandLine Args: ${info.inputArguments.joinToString(" ")}")
-        logger.info("Application Args: ${args.joinToString(" ")}")
+        //TODO: Make the opts parser print this or something
```
sollecitom: Is this to be addressed in the future?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -454,14 +437,14 @@ open class NodeStartup(val args: Array<String>) {
...
+    override fun initLogging() {
+        val loggingLevel = loggingLevel.name().toLowerCase(Locale.ENGLISH)
         System.setProperty("defaultLogLevel", loggingLevel) // These properties are referenced from the XML config file.
-        if (cmdlineOptions.logToConsole) {
+        if (verbose) {
             System.setProperty("consoleLogLevel", loggingLevel)
             Node.renderBasicInfoToConsole = false
         }
-        System.setProperty("log-path", (cmdlineOptions.baseDirectory / LOGS_DIRECTORY_NAME).toString())
+        System.setProperty("log-path", (cmdLineOptions.baseDirectory / LOGS_DIRECTORY_NAME).toString())
```
sollecitom: Can we try to have a `logsDirectory` command line option, rather than `baseDirectory`?

node/src/test/kotlin/net/corda/node/NodeCmdLineOptionsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+        @BeforeClass
+        @JvmStatic
+        fun initDirectories() {
+            workingDirectory = Paths.get(".").normalize().toAbsolutePath()
+            buildDirectory = workingDirectory.resolve("build")
+        }
+    }
+
+    @Test
+    fun `no command line arguments`() {
```
sollecitom: Not necessarily for now, but can we check the parser against specified options?

tools/cliutils/src/main/kotlin/net/corda/cliutils/ConfigFilePathArgsParser.kt
```diff
@@ -13,8 +14,8 @@ import java.nio.file.Path
  */
 @CommandLine.Command(description = ["Parse configuration file. Checks if given configuration file exists"])
 class ConfigFilePathArgsParser : Validated {
-    @CommandLine.Option(names = ["--config-file", "-f"], required = true, paramLabel = "FILE", description = ["The path to the config file"])
-    lateinit var configFile: Path
+    @CommandLine.Option(names = ["--config-file", "-f"], paramLabel = "FILE", description = ["The path to the config file"])
+    var configFile: Path = Paths.get("./node.conf")
```
sollecitom: Same here, do we have to hardcode the default here?

tools/cliutils/src/main/kotlin/net/corda/cliutils/InstallShellExtensionsParser.kt
```diff
@@ -53,7 +53,15 @@ private class ShellExtensionsGenerator(val alias: String, val className: String)
     }
 
     private val userHome: Path by lazy { Paths.get(System.getProperty("user.home")) }
-    private val jarLocation: Path by lazy { this.javaClass.location.toPath() }
+    private val jarLocation: Path by lazy {
+        val capsuleJarProperty = System.getProperty("capsule.jar")
+        if (capsuleJarProperty != null) {
+            //Use capsule jar property
```
sollecitom: Ah, thanks :)

docs/source/running-a-node.rst
anthonykeenan: No I don't think that's true.
It's just a boolean. If it's specified via --devMode it uses that value first as that is in the `configOverrides`, then if it's specified in the `node.conf` it uses that from `appConfig`, and if it's not specified in either of those places, it gets the `smartDevMode` value based on the operating system in this [function](https://github.com/corda/corda/blob/7853cfe0034cc769358193f26afa3021c38a0554/node/src/main/kotlin/net/corda/node/services/config/ConfigUtilities.kt#L48).

node/src/main/java/CordaCaplet.java
anthonykeenan: Sort of... It doesn't need the single hyphen options, they aren't accepted anymore, but it needs to know about the -b and -f options which are the short forms of them, picocli will also accept -bf so I think I need to do some more work here.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
anthonykeenan: Yes I think that's one for another day - I think a bigger issue personally is that you can't start up a node for testing without generating a network-parameters file.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: Ahh the application args never get this far down the stack now so that has to be logged further up. I'll add it to the base class.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
anthonykeenan: It needs the default values because it reads them and returns them as part of the help, which is why it wouldn't understand optional types. Not sure about the `configFile` one above though, I can't remember why I had to do that, I'll have a play.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: That seems eminently sensible

node/src/test/kotlin/net/corda/node/NodeCmdLineOptionsTest.kt
anthonykeenan: Yeah, I know what you mean, but that just seems like testing that picocli works really though, and it already has it's own tests.

tools/cliutils/src/main/kotlin/net/corda/cliutils/InstallShellExtensionsParser.kt
anthonykeenan: Possibly the comment is a little overkill ;-)

tools/cliutils/src/main/kotlin/net/corda/cliutils/ConfigFilePathArgsParser.kt
anthonykeenan: Yes, to display the defaults as part of the help, see above

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: I started doing this and it needs to be changed in quite a few places including changing some of the API so I've created a story for it: https://r3-cev.atlassian.net/browse/CORDA-1956

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
anthonykeenan: So as I mentioned below, all the defaults are in this file because picocli uses it as part of the help display. This is the only one that can't default directly because the default changes dependent on whether the `baseDirectory` parameter is set, so since the other command line defaults are in this file it seemed to made sense to put this here too.
Although I've also updated the option description to make the default clearer since picocli won't add it automatically to the help message.

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
-* ``--version``: Print the version and exit
+  quit.
+* ``--just-generate-rpc-ssl-settings``: Generate the ssl keystore and truststore for a secure RPC connection.
+* ``--log-to-console``, ``--verbose``, ``-v``: If set, prints logging to the console as well as to a file.
+* ``--logging-level <[ERROR,WARN,INFO, DEBUG,TRACE]>``: Enable logging at this level and higher. Defaults to INFO.
+* ``--network-root-truststore``, ``-t``: Network root trust store obtained from network operator.
+* ``--network-root-truststore-password``, ``-p``: Network root trust store password obtained from network operator.
+* ``--no-local-shell``, ``-n``: Do not start the embedded shell locally.
+* ``--on-unknown-config-keys``: How to behave on unknown node configuration. Valid options are FAIL, WARN or INFO. Defaults to FAIL.
+* ``--sshd``: Enables SSHD server for node administration.
```
tlil: SSH server

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
+* ``--sshd-port``: Sets the port for the SSHD server. If not supplied and SSHD server is enabled, the port defaults to 2222.
```
tlil: SSH server x 2

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
-* ``--sshd``: Enables SSHD server for node administration
-* ``--version``: Print the version and exit
+  quit.
+* ``--just-generate-rpc-ssl-settings``: Generate the ssl keystore and truststore for a secure RPC connection.
+* ``--log-to-console``, ``--verbose``, ``-v``: If set, prints logging to the console as well as to a file.
+* ``--logging-level <[ERROR,WARN,INFO, DEBUG,TRACE]>``: Enable logging at this level and higher. Defaults to INFO.
+* ``--network-root-truststore``, ``-t``: Network root trust store obtained from network operator.
+* ``--network-root-truststore-password``, ``-p``: Network root trust store password obtained from network operator.
+* ``--no-local-shell``, ``-n``: Do not start the embedded shell locally.
+* ``--on-unknown-config-keys``: How to behave on unknown node configuration. Valid options are FAIL, WARN or INFO. Defaults to FAIL.
```
tlil: Add `<[FAIL,WARN,INFO]>`?

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
-* ``--logging-level <[ERROR,WARN,INFO, DEBUG,TRACE]>``: Enable logging at this level and higher (default: INFO)
-* ``--network-root-truststore``: Network root trust store obtained from network operator
-* ``--network-root-truststore-password``: Network root trust store password obtained from network operator
-* ``--no-local-shell``: Do not start the embedded shell locally
-* ``--sshd``: Enables SSHD server for node administration
-* ``--version``: Print the version and exit
+  quit.
+* ``--just-generate-rpc-ssl-settings``: Generate the ssl keystore and truststore for a secure RPC connection.
+* ``--log-to-console``, ``--verbose``, ``-v``: If set, prints logging to the console as well as to a file.
+* ``--logging-level <[ERROR,WARN,INFO, DEBUG,TRACE]>``: Enable logging at this level and higher. Defaults to INFO.
```
tlil: Extra space between `INFO,` and `DEBUG`

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
 ~~~~~~~~~~~~~~~~~~~~
 The node can optionally be started with the following command-line options:
 
-* ``--base-directory``: The node working directory where all the files are kept (default: ``.``)
+* ``--base-directory``, ``-b``: The node working directory where all the files are kept (default: ``.``)
```
tlil: Trailing period #nit

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
 * ``--bootstrap-raft-cluster``: Bootstraps Raft cluster. The node forms a single node cluster (ignoring otherwise configured peer
```
tlil: Trailing period #nit

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
+* ``--sshd``: Enables SSHD server for node administration.
+* ``--sshd-port``: Sets the port for the SSHD server. If not supplied and SSHD server is enabled, the port defaults to 2222.
+* ``--version``, ``-V``: Prints the version and exits.
+
+.. _installing-shell-extensions:
+
+Installing shell extensions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Users of ``bash`` or ``zsh`` can install an alias and command line completion for corda. Run:
```
tlil: Corda

docs/source/running-a-node.rst
```diff
@@ -47,22 +47,75 @@ Command-line options
...
-* ``--config-file``: The path to the config file (default: ``node.conf``)                          
-* ``--help``
+* ``--clear-network-map-cache``, ``-c``: Clears local copy of network map, on node startup it will be restored from server or file system.
+* ``--config-file``, ``-f``: The path to the config file. Defaults to ``node.conf``.
+* ``--dev-mode``, ``-d``: Runs the node in developer mode. Unsafe in production. Defaults to true on MacOS and desktop versions of Windows. False otherwise.
+* ``--help``, ``-h``: Displays the help message and exits.
 * ``--initial-registration``: Start initial node registration with Corda network to obtain certificate from the permissioning 
-  server      
+  server.
+* ``--install-shell-extensions``: Installs an alias and autocomplete for users of ``bash`` or ``zsh``. See below for more information.
```
tlil: auto-completion

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    @Option(
+            names = ["-f", "--config-file"],
+            description = ["The path to the config file. By default this is node.conf in the base directory."]
+    )
+    var configFileArgument: Path? = null
+
+    val configFile : Path
+        get() {
+            return if (configFileArgument == null) {
```
tlil: Can rephrase this to:

```kotlin
get() = configFileArgument ?: (baseDirectory / "node.conf")
```

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+                baseDirectory / "node.conf"
+            } else {
+                configFileArgument!!
+            }
+        }
+
+    @Option(
+            names = ["--sshd"],
+            description = ["If set, enables sshd server for node administration."]
```
tlil: enables SSH server

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+    )
+    var sshdServer: Boolean = false
+
+    @Option(
+            names = ["--sshd-port"],
+            description = ["The port to start the sshd server on, if enabled."]
```
tlil: SSH server

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    @Option(
+            names = ["-d", "--dev-mode"],
+            description = ["Run the node in developer mode. Unsafe for production."]
+    )
+    var devMode: Boolean? = null
+
+    @Option(
+            names = ["--just-generate-node-info"],
+            description = ["Perform the node start-up task necessary to generate its nodeInfo, save it to disk, then quit"]
```
tlil: node info

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+    )
+    var justGenerateNodeInfo: Boolean = false
```
tlil: Just a general note that in the future we should really be providing these commands as ehrm... commands :) Obviously keeping these for backwards compatibility, but the for new commands we be using sub-commands. Could consider adding duplicate commands for these `@Option`s as well. What do you think? Doesn't need to be addressed in this PR.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    @Option(
+            names = ["--just-generate-rpc-ssl-settings"],
+            description = ["Generate the ssl keystore and truststore for a secure RPC connection."]
```
tlil: SSL key and trust stores

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+            names = ["--bootstrap-raft-cluster"],
+            description = ["Bootstraps Raft cluster. The node forms a single node cluster (ignoring otherwise configured peer addresses), acting as a seed for other nodes to join the cluster."]
+    )
+    var bootstrapRaftCluster: Boolean = false
+
+    @Option(
+            names = ["-c", "--clear-network-map-cache"],
+            description = ["Clears local copy of network map, on node startup it will be restored from server or file system."]
+    )
+    var clearNetworkMapCache: Boolean = false
```
tlil: See comment further up about commands.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+                baseDirectory,
+                configFile,
+                configOverrides = ConfigFactory.parseMap(mapOf("noLocalShell" to this.noLocalShell) +
+                        if (sshdServer) mapOf("sshd" to mapOf("port" to sshdServerPort.toString())) else emptyMap<String, Any>() +
+                        if (devMode != null) mapOf("devMode" to this.devMode) else emptyMap())
+        )
+        return rawConfig to Try.on {
+            rawConfig.parseAsNodeConfiguration(unknownConfigKeysPolicy::handle).also { config ->
+                if (nodeRegistrationOption != null) {
+                    require(!config.devMode) { "registration cannot occur in devMode" }
```
tlil: Registration (for consistency with other `require` statements)

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -0,0 +1,144 @@
...
+                    require(config.compatibilityZoneURL != null || config.networkServices != null) {
+                        "compatibilityZoneURL or networkServices must be present in the node configuration file in registration mode."
+                    }
+                }
+            }
+        }
+    }
+}
+
+data class NodeRegistrationOption(val networkRootTrustStorePath: Path, val networkRootTrustStorePassword: String)
```
tlil: In general, re node registration:

Have you performed any testing against the Network Services test environments? Just to go through the same steps as previously used to get registered on a network, and to make sure that this doesn't break any of the steps?

Could use #operational-qa-env for testing if you want.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -49,80 +48,83 @@ import java.util.*
...
         private val logger by lazy { loggerFor<Node>() } // I guess this is lazy to allow for logging init, but why Node?
         const val LOGS_DIRECTORY_NAME = "logs"
         const val LOGS_CAN_BE_FOUND_IN_STRING = "Logs can be found in"
         private const val INITIAL_REGISTRATION_MARKER = ".initialregistration"
     }
 
+    @Mixin
+    var cmdLineOptions = NodeCmdLineOptions()
+
+    private class ExitCodes {
```
tlil: Could these two basic exit codes live in the `cliutils` module?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -49,80 +48,83 @@ import java.util.*
...
         // Required to install our [SecureRandom] before e.g., UUID asks for one.
         // This needs to go after initLogging(netty clashes with our logging).
         Crypto.registerProviders()
 
         val versionInfo = getVersionInfo()
 
-        if (cmdlineOptions.isVersion) {
-            println("${versionInfo.vendor} ${versionInfo.releaseVersion}")
-            println("Revision ${versionInfo.revision}")
-            println("Platform Version ${versionInfo.platformVersion}")
```
tlil: Is revision and platform version included in the `--version` dump?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -199,14 +199,13 @@ open class NodeStartup(val args: Array<String>) {
             """.trimIndent()
     }
 
-    private fun loadConfiguration(cmdlineOptions: CmdLineOptions): NodeConfiguration {
-
-        val (rawConfig, configurationResult) = loadConfigFile(cmdlineOptions)
-        if (cmdlineOptions.devMode) {
+    private fun loadConfiguration(): NodeConfiguration {
+        val (rawConfig, configurationResult) = loadConfigFile()
+        if (cmdLineOptions.devMode != null) {
```
tlil: What if `cmdLineOptions.devMode != null`, but `cmdLineOptions.devMode == false` ?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
tlil: Yep, thanks for raising CORDA-1956. Would be good to get that addressed soonish.

tools/cliutils/src/main/kotlin/net/corda/cliutils/ConfigFilePathArgsParser.kt
```diff
@@ -13,8 +14,8 @@ import java.nio.file.Path
  */
 @CommandLine.Command(description = ["Parse configuration file. Checks if given configuration file exists"])
 class ConfigFilePathArgsParser : Validated {
-    @CommandLine.Option(names = ["--config-file", "-f"], required = true, paramLabel = "FILE", description = ["The path to the config file"])
-    lateinit var configFile: Path
+    @CommandLine.Option(names = ["--config-file", "-f"], paramLabel = "FILE", description = ["The path to the config file"])
```
tlil: Is this mix-in actually used any more?

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
```diff
@@ -39,14 +42,26 @@ interface Validated {
...
+    AnsiConsole.systemInstall()
+
     val cmd = CommandLine(this)
+    this.args = args
     cmd.commandSpec.name(alias)
     cmd.commandSpec.usageMessage().description(description)
     try {
-        cmd.parseWithHandlers(RunLast().useOut(System.out).useAnsi(Help.Ansi.AUTO),
-                DefaultExceptionHandler<List<Any>>().useErr(System.err).useAnsi(Help.Ansi.AUTO),
+        val results = cmd.parseWithHandlers(RunLast().useOut(System.out).useAnsi(Help.Ansi.ON),
```
tlil: Could we keep this as `AUTO` and have a command-line option to set `--color <ALWAYS,NEVER,AUTO>`?

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
```diff
@@ -39,14 +42,26 @@ interface Validated {
...
+                DefaultExceptionHandler<List<Any>>().useErr(System.err).useAnsi(Help.Ansi.ON),
                 *args)
+        // If an error code has been returned, use this and exit
+        results?.firstOrNull()?.let {
+            if (it is Int) {
+                exitProcess(it)
+            }
```
tlil: If it's non-int for some bizarre reason, could we return `1`?

node/src/test/kotlin/net/corda/node/NodeCmdLineOptionsTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import net.corda.node.internal.NodeStartup
+import net.corda.nodeapi.internal.config.UnknownConfigKeysPolicy
+import org.apache.logging.log4j.Level
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.BeforeClass
+import org.junit.Test
+import java.nio.file.Path
+import java.nio.file.Paths
+
+class NodeCmdLineOptionsTest {
```
tlil: We should integrate this with @roastario's CLI checker as well so that we can have tests that check for breaking changes.

tlil:
For traceability, could you document (on this PR) what testing you've done of this as well?

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
anthonykeenan: This is the only thing I haven't done, I used bootstrapped networks for all of my testing, I'll have a go at that this afternoon.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: Yes, they could, the reason I put them there was that I was expecting to add more to them for the Node as part of the exit code handling stuff. I suppose they could be subclassed.

tlil: Yep :+1:

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: It is now!

tools/cliutils/src/main/kotlin/net/corda/cliutils/ConfigFilePathArgsParser.kt
anthonykeenan: Yes, so this is part of the utilities that @kasiastreich added. I believe the network services app requires a `node.conf` file, however Corda doesn't require one for all scenarios and the validation is different. I've kept it for now in case any of the other CLI utilities use it.

tlil: Can it then be reused in the node? Seems like we now have 2 ways of doing the same, no?

anthonykeenan: Yes, it didn't work with the node, I spent a while trying to make it work and then decided it wasn't the right fit and added it directly to the class.

tlil: Doesn't look like it is used though? At least not in the OS and ENT repositories. Is the thinking that the `cliutils` will be published and used in Network Services? Or will it remain copy-pasta, @kasiastreich ? At the moment, the mixin is defined in the network-services repo as well. So should be fine to remove for now I would've thought.

kasiastreich: So network services use it. I finished picocli porting a month ago. Probably someone will have to revisit it later to unify it between repos?

tlil: Yeah, let's do that as a follow-up ticket, @anthonykeenan. Mind removing for now?

Out of curiosity, what problems did you run into?

anthonykeenan: I _think_ it was the fact that it had validation that failed when no config file was supplied. I tried to fix it by making the option not required, but then it still failed the validation. I think if we do have any apps that require a config file, it would be the right fit.

kasiastreich: In network services we have lots of utilities that require config file, that's why it is needed. Probably it isn't useful for corda itself that much.

tlil: Thanks @kasiastreich

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
anthonykeenan: There is already a story for that: https://r3-cev.atlassian.net/browse/CORDA-1838 so I'm going to come back to that if that's ok

tlil: Yep, duh, forgot that story. Let's treat that one separately, yes.

node/src/test/kotlin/net/corda/node/NodeCmdLineOptionsTest.kt
tlil: As discussed F2F, let's address that in a separate story.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
anthonykeenan: Because we're forcing Ansi on [here](https://github.com/corda/corda/pull/3872/files/6387407d4c37944c3463b28e113469c21b02c8b0#diff-11319e90404ef45fdef07f94706e5e65R47), I think `Help.Ansi.ON` is the right option - for example, in Windows I don't get any colours unless it's set to ON.

anthonykeenan:
Testing wise I just started the node with every option and checked that it parsed the option correctly.
I also tested a couple of combinations.
I tested it both on Windows and then on a Linux (Ubuntu) machine using the `corda` alias and auto-complete as well.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
tlil: As discussed in person, forcing it to `ON` will most likely cause the control characters to be included if the command output get piped to another command or a file. So on macOS and Linux, that's sub-optimal. If Windows is behaving differently, we can look at an alternative solution/default for that platform in particular. :-)

PR #3874, opened 2018-08-30
CORDA-1892 CRaSH shell flow start fix for similar flow names.

josecoll:
If a node has two flows, where one's name is a longer version of the other's, they cannot be started

### DIFF
--- a/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
+++ b/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
@@ -228,14 +228,15 @@ object InteractiveShell {
     // TODO: This should become the default renderer rather than something used specifically by commands.
     private val outputMapper by lazy { createOutputMapper() }
 
+    @VisibleForTesting
+    lateinit var latch: CountDownLatch
+        private set
+
     /**
      * Called from the 'flow' shell command. Takes a name fragment and finds a matching flow, or prints out
      * the list of options if the request is ambiguous. Then parses [inputData] as constructor arguments using
      * the [runFlowFromString] method and starts the requested flow. Ctrl-C can be used to cancel.
      */
-    @VisibleForTesting
-    lateinit var latch: CountDownLatch
-
     @JvmStatic
     fun runFlowByNameFragment(nameFragment: String,
                               inputData: String,
--- a/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
+++ b/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
@@ -259,7 +259,6 @@ object InteractiveShell {
         }
 
         val flowName = matches.find { it == nameFragment }
-        output.println("Executing flow: $flowName ")
         val flowClazz: Class<FlowLogic<*>> = if (classLoader != null) {
             uncheckedCast(Class.forName(flowName, true, classLoader))
         } else {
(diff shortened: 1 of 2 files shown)


tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -249,23 +252,27 @@ object InteractiveShell {
...
             matches.forEachIndexed { i, s -> output.println("${i + 1}. $s", Color.yellow) }
             return
         }
 
         val flowClazz: Class<FlowLogic<*>> = if (classLoader != null) {
-            uncheckedCast(Class.forName(matches.single(), true, classLoader))
+            uncheckedCast(Class.forName(flowName, true, classLoader))
         } else {
-            uncheckedCast(Class.forName(matches.single()))
+            uncheckedCast(Class.forName(flowName))
```
chrisr3: My poor little brain says that `flowName` _might_ be `null` here if `matches.size == 1`.
I.e. suppose `nameFragment = 'Burble'`, and the only flow were called `BurbleFlow`. Wouldn't this be a NPE?

chrisr3:
```kotlin
val flowClazz: Class<FlowLogic<*>> = (flowName ?: matches[0]).let { className ->
    if (classLoader != null) {
        uncheckedCast(Class.forName(className, true, classLoader))
    } else {
        uncheckedCast(Class.forName(className))
    }
}
```

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -255,11 +255,13 @@ object InteractiveShell {
...
-        val flowName = matches.find { it.endsWith(nameFragment) }
-        if (matches.size > 1 && flowName == null) {
+        val flowName = matches.find { it.endsWith(nameFragment) } ?: if (matches.size > 1) {
             output.println("Ambiguous name provided, please be more specific. Your options are:")
             matches.forEachIndexed { i, s -> output.println("${i + 1}. $s", Color.yellow) }
             return
+        } else {
+            output.println("No matching flow found, run 'flow list' to see your options.", Color.red)
+            return
         }
```
chrisr3: But in your test case, there _is_ a flow called `BurbleFlow` so why shouldn't `nameFragment = 'Burble'` match?
Wouldn't this make more sense?
```kotlin
val flowName = matches.find { it.endsWith(nameFragment) } ?: if (matches.size > 1) {
   // ambiguous
} else if (matches.isEmpty()) {
   // no flow found
} else {
    // Unique match - use this one!
    matches[0]
}
```

josecoll: The code will only execute a flow whereby the flowName is completely specified (either as a single classname or FQ package name) and is not ambiguous.
So in your example, "Burble" is not a valid flow, but "BurbleFlow" and "net.corda.tools.shell.BurbleFlow" are.
Your interpretation is a different requirement and not related to this fix.

josecoll: Also note that `matches.find { it.endsWith(nameFragment) }` means that `Burble` will never match as no actual flows end in `Burble` ;-)

chrisr3: So is this a change in behaviour? Would "Burble" previously have matched `BurbleFlow`?

josecoll:
@chrisr3 reverted back to original behaviour as per your latest comment.

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
josecoll: Good point - let me ensure the original behaviour is unchanged.

PR #959, opened 2017-07-03
Changes the name of the addTimeWindow method to setTimeWindow.

joeldudleyr3:

PR #3879, opened 2018-08-31
Cherry-pick: CORDA-1604 Node Info file watcher should block and load node info when node startup (#959) (#3333) 

patrickkuo:
https://github.com/corda/corda/pull/3333


josecoll:
Please re-run integration tests.

PR #3889, opened 2018-09-04
ENT-1755: Support HA P2P Addresses without load balancer

mnesbit:
It seems that previous problems I had with supporting this and not breaking service address behaviour have disappeared.

The new additionalP2PAddresses property allows these extra HA/alias names, but note older code will slice off the extra addresses, so we require all nodes to be on consistent versions of software for this to be used fully.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
+++ b/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
@@ -120,7 +120,7 @@ class P2PMessagingClient(val config: NodeConfiguration,
     private lateinit var advertisedAddress: NetworkHostAndPort
     private var maxMessageSize: Int = -1
 
-    override val myAddress: SingleMessageRecipient get() = NodeAddress(myIdentity, advertisedAddress)
+    override val myAddress: SingleMessageRecipient get() = NodeAddress(myIdentity)
     override val ourSenderUUID = UUID.randomUUID().toString()
 
     private val state = ThreadBox(InnerState())
@@ -233,7 +233,7 @@ class P2PMessagingClient(val config: NodeConfiguration,
         fun gatherAddresses(node: NodeInfo): Sequence<BridgeEntry> {
             return state.locked {
                 node.legalIdentitiesAndCerts.map {
-                    val messagingAddress = NodeAddress(it.party.owningKey, node.addresses.first())
+                    val messagingAddress = NodeAddress(it.party.owningKey)
                     BridgeEntry(messagingAddress.queueName, node.addresses, node.legalIdentities.map { it.name })
                 }.filter { producerSession!!.queueQuery(SimpleString(it.queueName)).isExists }.asSequence()
             }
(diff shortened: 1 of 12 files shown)


node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -540,7 +540,7 @@ class P2PMessagingClient(val config: NodeConfiguration,
 
     override fun getAddressOfParty(partyInfo: PartyInfo): MessageRecipients {
         return when (partyInfo) {
-            is PartyInfo.SingleNode -> NodeAddress(partyInfo.party.owningKey, partyInfo.addresses.single())
+            is PartyInfo.SingleNode -> NodeAddress(partyInfo.party.owningKey, partyInfo.addresses.first())
```
exFalso: Doesn't this mean we'll only use the first address? Is there a test connecting to other ones?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -604,7 +605,8 @@ class DriverDSLImpl(
                     }
                 }
             }
-            val p2pReadyFuture = addressMustBeBoundFuture(executorService, config.corda.p2pAddress, process)
+            val p2pAddress = if (config.corda.messagingServerAddress != null) config.corda.messagingServerAddress!! else config.corda.p2pAddress
```
exFalso:
```kotlin
config.corda.messagingServerAddress ?: config.corda.p2pAddress
```

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -689,6 +691,7 @@ class DriverDSLImpl(
                 executorService: ScheduledExecutorService,
                 config: NodeConfig
         ): CordaFuture<Pair<NodeWithInfo, Thread>> {
+            val effectiveP2PAddress = if (config.corda.messagingServerAddress != null) config.corda.messagingServerAddress!! else config.corda.p2pAddress
```
exFalso: same

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
mnesbit: Oh yes that will work!

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
mnesbit: No in fact it doesn't. The NodeAddress is created just to get the communication topic which is based upon the key. The topic doesn't in fact contain the IP address. I would like to remove the constructor parameter, but I am not sure if it is 'api'.

PR #3891, opened 2018-09-04
Updated repository lists to reduce dependency on Jitpack and removed …

adagys:
…unused repositories.

This fixes the problem with the broken bft-smart dependency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,7 @@ buildscript {
     ext.jsr305_version = constants.getProperty("jsr305Version")
     ext.shiro_version = '1.4.0'
     ext.artifactory_plugin_version = constants.getProperty('artifactoryPluginVersion')
+    ext.artifactory_contextUrl = 'https://ci-artifactory.corda.r3cev.com/artifactory'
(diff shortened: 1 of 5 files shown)


opticyclic:
How does this reduce the dependency on jitpack? You haven't removed jitpack from the repository list.

Clintonio:
(Or whoever is currently release managing V3). 

Also the branch needs some changes. The Artifactory context URL variable isn't in the V3 build.gradles it seems.

opticyclic:
So if there is a corrupt dependency in jitpack wont this propagate down to when people use corda as a dependency?

fenryka:
Seems acceptable to me, this won't impact users I assume

opticyclic:
All this has done is hide the issue internally at R3 but it still exists for all the clients.

mikehearn:
Yes, we are sending some instructions to corda-dev and I'm asking JitPack support if they can restore the package. We've added the package to our own artifactory but beyond distributing instructions and trying to get jitpack fixed, I'm not sure there's much more we can do - Maven/Gradle type build systems assume repositories are immutable and it seems to be a bug in jitpack that it's not.

PR #3892, opened 2018-09-04
CORDA-1864: Added getter for network parameters to RPC

shamsasari:


core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -227,6 +228,9 @@ interface CordaRPCOps : RPCOps {
     @RPCReturnsObservables
     fun networkMapFeed(): DataFeed<List<NodeInfo>, NetworkMapCache.MapChange>
 
+    /** Returns the network parameters the node is operating under. */
+    val networkParameters: NetworkParameters
```
mkit: perhaps `nodeNetworkParameters` will be better.

PR #3895, opened 2018-09-05
ENT-1906: Remove SLF4J's JUL logging bridge from DJVM.

chrisr3:
The SLF4J logging bridge for JUL requires some extra configuration steps (which haven't been implemented) before it will do anything. Remove this unused dependency.

Also move SLF4J's Log4J back-end into the CLI project, along with `jcabi-manifests`.


tlil:
Ah, and jcabi-manifests was left lingering in `djvm` when moved to `djvm-cli`. Ta

PR #3899, opened 2018-09-05
[CORDA-1926] Implement target version and min platform version 

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1926


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -134,8 +134,8 @@ open class Node(configuration: NodeConfiguration,
         }
 
         private val sameVmNodeCounter = AtomicInteger()
-        private fun makeCordappLoader(configuration: NodeConfiguration): CordappLoader {
-            return JarScanningCordappLoader.fromDirectories(configuration.cordappDirectories)
+        private fun makeCordappLoader(configuration: NodeConfiguration, platformVerison: Int): CordappLoader {
```
distributedleetravis: typo: `platformVersion`

node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
```diff
@@ -32,7 +33,7 @@ import kotlin.streams.toList
  *
  * @property cordappJarPaths The classpath of cordapp JARs
  */
-class JarScanningCordappLoader private constructor(private val cordappJarPaths: List<RestrictedURL>) : CordappLoaderTemplate() {
+class JarScanningCordappLoader private constructor(private val cordappJarPaths: List<RestrictedURL>, private val platformVerison: Int) : CordappLoaderTemplate() {
```
distributedleetravis: type: `platformVersion`

node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
```diff
@@ -54,18 +55,18 @@ class JarScanningCordappLoader private constructor(private val cordappJarPaths:
          *
          * @param corDappDirectories Directories used to scan for CorDapp JARs.
          */
-        fun fromDirectories(corDappDirectories: Iterable<Path>): CordappLoader {
+        fun fromDirectories(corDappDirectories: Iterable<Path>, platformVerison: Int): CordappLoader {
```
distributedleetravis: ditto

node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
```diff
@@ -54,18 +55,18 @@ class JarScanningCordappLoader private constructor(private val cordappJarPaths:
...
+            return JarScanningCordappLoader(corDappDirectories.distinct().flatMap(this::jarUrlsInDirectory).map { it.restricted() }, platformVerison)
         }
 
         /**
          * Creates a CordappLoader loader out of a list of JAR URLs.
          *
          * @param scanJars Uses the JAR URLs provided for classpath scanning and Cordapp detection.
          */
-        fun fromJarUrls(scanJars: List<URL>) = JarScanningCordappLoader(scanJars.map { it.restricted() })
+        fun fromJarUrls(scanJars: List<URL>, platformVerison: Int) = JarScanningCordappLoader(scanJars.map { it.restricted() }, platformVerison)
```
distributedleetravis: ditto

node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
```diff
@@ -106,11 +107,21 @@ class JarScanningCordappLoader private constructor(private val cordappJarPaths:
     }
 
     private fun loadCordapps(): List<Cordapp> {
-        return cordappJarPaths.map { scanCordapp(it).toCordapp(it) }
+        val unfilteredCordapps = cordappJarPaths.map { scanCordapp(it).toCordapp(it) }
+        var cordapps: MutableList<Cordapp> = mutableListOf()
+        for (cordapp in unfilteredCordapps) {
+            if (cordapp.info.minPlatformVersion > platformVerison) {
```
distributedleetravis: ditto

node-api/src/test/kotlin/net/corda/nodeapi/internal/AttachmentsClassLoaderStaticContractTests.kt
```diff
@@ -67,7 +67,7 @@ class AttachmentsClassLoaderStaticContractTests {
     }
 
     private val serviceHub = rigorousMock<ServicesForResolution>().also {
-        val cordappProviderImpl = CordappProviderImpl(cordappLoaderForPackages(listOf("net.corda.nodeapi.internal")), MockCordappConfigProvider(), MockAttachmentStorage())
+        val cordappProviderImpl = CordappProviderImpl(cordappLoaderForPackages(listOf("net.corda.nodeapi.internal"), 4), MockCordappConfigProvider(), MockAttachmentStorage())
```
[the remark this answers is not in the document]
florian-f: Sure.
Alternatively, we could read the platform version from `constants.properties`, what do you think?

core/src/main/kotlin/net/corda/core/cordapp/Cordapp.kt
```diff
@@ -55,13 +57,16 @@ interface Cordapp {
      * @property shortName Cordapp's shortName
      * @property vendor Cordapp's vendor
      * @property version Cordapp's version
+     * @property minPlatformVersion the minimum platform version required to run this Cordapp
+     * @property targetPlatformVersion the platform version this Cordapp was tested on
      */
     @DoNotImplement
     interface Info {
```
shamsasari: As discussed offline you're having merge conflicts here as I've just removed this interface (https://github.com/corda/corda/pull/3914). I understood (and maybe I was wrong) that we weren't going to expose the `Info` hence why there wasn't a `info` property on `Cordapp`. If that's no longer the case then I think it's fine to have `minPlatformVersion` and `targetPlatformVersion` here and resurrect this interface.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -108,7 +108,7 @@ open class MockServices private constructor(
                                             networkParameters: NetworkParameters = testNetworkParameters(),
                                             vararg moreKeys: KeyPair): Pair<CordaPersistence, MockServices> {
 
-            val cordappLoader = cordappLoaderForPackages(cordappPackages)
+            val cordappLoader = cordappLoaderForPackages(cordappPackages, 4 /* TODO get current platform version */)
```
shamsasari: There's `PLATFORM_VERSION` you can use

core/src/main/kotlin/net/corda/core/cordapp/Cordapp.kt
```diff
@@ -55,13 +57,16 @@ interface Cordapp {
...
         val shortName: String
         val vendor: String
         val version: String
-
+        val minPlatformVersion: Int
```
shamsasari: Consider `minimumPlatformVersion` to be consistent with the one in `NetworkParameters`

node/src/main/kotlin/net/corda/node/internal/cordapp/ManifestUtils.kt
```diff
@@ -43,5 +43,10 @@ internal fun Manifest?.toCordappInfo(defaultShortName: String): Cordapp.Info {
     this?.mainAttributes?.getValue("Implementation-Version")?.let { version ->
         unknown = unknown.copy(version = version)
     }
+    val minPlatformVersion = this?.mainAttributes?.getValue("Min-Platform-Version")?.toInt() ?: 1
```
shamsasari: We need to also update the gradle plugins to so that it allows these to be input in a nice way:
```
cordapp {
    info {
...
        minPlatformVersion 2
        targetPlatformVersion 4
    }
}
```

Until that's implemented can you update cordapp-build-systems.rst on how a CorDapp dev is to specify these according to these changes.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
florian-f: Right, that didn't exist yet when I made the PR :)

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -69,10 +71,10 @@ open class MockServices private constructor(
 
     companion object {
 
-        private fun cordappLoaderForPackages(packages: Iterable<String>): CordappLoader {
+        private fun cordappLoaderForPackages(packages: Iterable<String>, versionInfo: VersionInfo = VersionInfo.UNKNOWN.copy(PLATFORM_VERSION)): CordappLoader {
```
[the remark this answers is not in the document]
florian-f: Good idea, I changed it accordingly.

florian-f:
@shamsasari We can't use the `location` extension method in InternalUtils.kt because `StackTraceElement`s do not refer to _Class_ instances, they merely hold
information like the class name as strings.

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
```diff
@@ -0,0 +1,42 @@
...
+            cordappInfoResolver = tempResolver as CordappInfoResolver
+            block()
+            cordappInfoResolver = resolver
+        }
+    }
+}
+
+class CordappInfoResolverImpl(private val cordapps: List<CordappImpl>) : CordappInfoResolver() {
+    override fun invoke(): CordappImpl.Info? {
+        Exception().stackTrace.forEach { stackFrame: StackTraceElement ->
```
mikehearn: Can we add a // TODO here to switch to the new stack walking API once we upgrade past java 8?

core/src/test/kotlin/net/corda/core/internal/cordapp/CordappInfoResolverTest.kt
```diff
@@ -0,0 +1,34 @@
...
+        val default = object : CordappInfoResolver() {
+            override fun invoke(): CordappImpl.Info? {
+                return CordappImpl.Info("default", "default", "1", 2, defaultTargetVersion)
+            }
+        }
+        CordappInfoResolver.init(default)
+        assertEquals(defaultTargetVersion, returnCallingTargetVersion())
+
+        val expectedTargetVersion = 555
+        CordappInfoResolver.withCordappInfoResolution(object : CordappInfoResolver() {
```
mikehearn: Hmm, why not just use a typealias instead of an abstract class inheriting from a function type (which is kind of unusual) - that way you could just pass a function body in here directly?

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
florian-f: Yes, added.

core/src/test/kotlin/net/corda/core/internal/cordapp/CordappInfoResolverTest.kt
florian-f: Good point. I changed it (not using a typealias though), wdyt?

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Temporarily switch out the internal resolver for another one. For use in testing.
+     */
+    @Synchronized
+    fun withCordappInfoResolution(tempResolver: () -> CordappImpl.Info?, block: () -> Unit) {
+        val resolver = cordappInfoResolver
+        cordappInfoResolver = tempResolver
+        block()
+        cordappInfoResolver = resolver
```
shamsasari: This needs to be done in a `finally` block in case `block()` throws an exception

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @return Information about the CorDapp from which the invoker is called, null if called outside a CorDapp.
+     */
+    fun getCorDappInfo(): CordappImpl.Info? = cordappInfoResolver()
+
+    /**
+     * Temporarily switch out the internal resolver for another one. For use in testing.
+     */
+    @Synchronized
+    fun withCordappInfoResolution(tempResolver: () -> CordappImpl.Info?, block: () -> Unit) {
```
shamsasari: If instead of `tempResolver` you pass in a `Map<CordappImpl.Info, List<String>>` parameter as the temporary contents for `cordappClasses` then you will also end up testing `register`.

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
florian-f: Good point

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    /*
+     * Associates class names with CorDapps or logs a warning when a CorDapp is already registered for a given class.
+     * This could happen when trying to run different versions of the same CorDapp on the same node.
+     */
+    @Synchronized
+    fun register(classes: List<String>, cordapp: CordappImpl.Info) {
+        classes.forEach {
+            if (cordappClasses.contains(it)) {
+               logger.warn("Not registering ${cordapp.shortName} for $it, a different CorDapp (${cordappClasses[it]?.shortName}) has already been registered for this class.")
```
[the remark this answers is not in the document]
florian-f: If we do that it would slightly change the semantics of `getCorDappInfo()`. When `getCorDappInfo()` returns null, one might conclude that code is not being called from a cordapp, which would not be correct any more. We need to make it clear that it behaves that way and developers don't use it to find out if code is being called from a cordapp. 
By the way, `CordappProviderImpl` currently does the same, it returns the first cordapp it finds even if there might be others. I'm not saying we should do the same here, just thought I should mention it.

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
florian-f: I thought about this, but I'm not sure if we really want to do this. Other threads might be registering CorDapps in the background (e.g. mock network with multiple mock nodes), and you don't want them to end up in your temporary resolver.

core/src/test/kotlin/net/corda/core/internal/cordapp/CordappInfoResolverTest.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.Test
+import kotlin.test.assertEquals
+
+class CordappInfoResolverTest {
+
+    @Test()
+    fun `The correct cordapp resolver is used after calling withCordappResolution`() {
+        val defaultTargetVersion = 222
+
+        CordappInfoResolver.register(listOf(javaClass.name), CordappImpl.Info("test", "test", "2", 3, defaultTargetVersion))
```
shamsasari: You're not cleaning this up. Subsequent tests will be contaminated.

core/src/test/kotlin/net/corda/core/internal/cordapp/CordappInfoResolverTest.kt
```diff
@@ -19,14 +19,15 @@ class CordappInfoResolverTest {
...
+        CordappInfoResolver.clear()
     }
 
     @Test()
     fun `When more than one cordapp is registered for the same class, the resolver returns null`() {
         CordappInfoResolver.register(listOf(javaClass.name), CordappImpl.Info("test", "test", "2", 3, 222))
         CordappInfoResolver.register(listOf(javaClass.name), CordappImpl.Info("test1", "test1", "1", 2, 456))
         assertEquals(0, returnCallingTargetVersion())
-
+        CordappInfoResolver.clear()
```
shamsasari: Put inside an `@After` method to ensure it's run on test failures.

node/src/test/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoaderTest.kt
```diff
@@ -101,13 +103,70 @@ class JarScanningCordappLoaderTest {
...
+        // exclude the core cordapp
+        assertThat(loader.cordapps.size).isEqualTo(2)
+    }
+
+    @Test
+    fun `cordapp classloader does load apps when their min platform version is equal to the platform version`() {
+        val jar = JarScanningCordappLoaderTest::class.java.getResource("versions/min-2-target-3.jar")!!
+        val loader = JarScanningCordappLoader.fromJarUrls(listOf(jar), VersionInfo.UNKNOWN.copy(platformVersion = 2))
+        // exclude the core cordapp
+        assertThat(loader.cordapps.size).isEqualTo(2)
```
shamsasari: The previous test is also asserting 2 CorDapps, or am I missing something?

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
```diff
@@ -0,0 +1,65 @@
...
+        val resolver = cordappInfoResolver
+        cordappInfoResolver = tempResolver
+        try {
+            block()
+        } finally {
+            cordappInfoResolver = resolver
+        }
+    }
+
+    internal fun clear() {
```
shamsasari: `@VisibleForTesting`

node/src/test/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoaderTest.kt
florian-f: Yes, the only difference is the (node) platform version that is passed to the class loader. There are two because it's the one from the jar and the core cordapp.

florian-f: ... with different configurations for the loader

shamsasari: Yep, I've been mis-reading the above test as "does not load". Doh!

PR #3900, opened 2018-09-05
Store default platform version in a constant

bpaunescu:
Running tests from Intellij will trigger the use of a default platform version when out of process nodes start. This value is hardcoded in 2 places. This PR introduces a constant to help avoid having version mismatch problems in the future.

@mikehearn I believe you had some plans to deal with versioning, so your thought on this quick fix are welcome.


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -427,7 +428,7 @@ open class NodeStartup(val args: Array<String>) {
         fun manifestValue(name: String): String? = if (Manifests.exists(name)) Manifests.read(name) else null
 
         return VersionInfo(
-                manifestValue("Corda-Platform-Version")?.toInt() ?: 1,
+                manifestValue("Corda-Platform-Version")?.toInt() ?: DEFAULT_PLATFORM_VERSION,
```
sollecitom: One thing here -> this should not be read from the JAR at all!

PR #3901, opened 2018-09-06
Fixed some problem I discovered while working on another story

sollecitom:
- Fixed some problems with error handling for Observables.
- Eliminated incorrect double `stop()` call for RpcBroker.
- Added `Schedulers.shutdown()` call in `stop()` implementation for Node and Driver, to avoid stuck processes when observable pipelines go wrong.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
+++ b/core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
@@ -432,7 +432,7 @@ fun CordaRPCOps.pendingFlowsCount(): DataFeed<Int, Pair<Int, Int>> {
                     }
                 }
             }.subscribe()
-    if (completedFlowsCount == 0) {
+    if (pendingFlowsCount == 0) {
         updates.onCompleted()
     }
     return DataFeed(pendingFlowsCount, updates)
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
@@ -875,7 +875,7 @@ private class NetworkVisibilityController {
             val (snapshot, updates) = rpc.networkMapFeed()
             visibleNodeCount = snapshot.size
             checkIfAllVisible()
-            subscription = updates.subscribe {
+            subscription = updates.subscribe({
                 when (it) {
                     is NetworkMapCache.MapChange.Added -> {
                         visibleNodeCount++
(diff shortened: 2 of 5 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -889,7 +889,9 @@ private class NetworkVisibilityController {
                         // Nothing to do here but better being exhaustive.
                     }
                 }
-            }
+            }, { _ ->
+                // Nothing to do on errors here.
```
shamsasari: What errors are we getting that we can ignore?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -955,9 +957,10 @@ fun <DI : DriverDSL, D : InternalDriverDSL, A> genericDriver(
     val shutdownHook = addShutdownHook(driverDsl::shutdown)
     try {
         driverDsl.start()
-        return dsl(coerce(driverDsl))
+        return dsl(coerce(driverDsl)).also { Schedulers.shutdown() }
```
shamsasari: This should in the driver dsl `shutdown` method, but even then I think this is problematic for the reasons given above.

shamsasari:
>Added Schedulers.shutdown() call in stop() implementation for Node and Driver, to avoid stuck processes when observable pipelines go wrong.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
sollecitom: Removed the `Schedulers.shutdown()` everywhere, until I come up with a better way of doing it.

sollecitom:
@shamsasari about the `Schedulers.,shutdown()`, the idea is to run `Observable.observeOn(Schedulers)`, which allows to prevent deadlocks.
When using it that way, though, it takes explicitly shutdown of `Schedulers`, which is sadly a singleton.

core/src/main/kotlin/net/corda/core/messaging/CordaRPCOps.kt
```diff
@@ -432,7 +432,7 @@ fun CordaRPCOps.pendingFlowsCount(): DataFeed<Int, Pair<Int, Int>> {
                     }
                 }
             }.subscribe()
-    if (completedFlowsCount == 0) {
+    if (pendingFlowsCount == 0) {
```
shamsasari: I just noticed this method. Is it intended to be public API? It seems more like a support function than an end-user method.

PR #3908, opened 2018-09-06
CORDA-1509  Configure and enable Gradle Build Cache

josecoll:
Gradle configuration changes to support using Gradle Build Cache in a CI environment.

Usage of Gradle Build cache in Azure requires a patched version of Gradle to include TCP Keep-Alive, as per description and resolution described in https://github.com/gradle/gradle/issues/6018.
This custom distribution has been uploaded to:
https://ci-artifactory.corda.r3cev.com/artifactory/downloads/org/gradle/gradle-4.9-keep-alive-bin.zip
This distribution is used by setting an additional Build Configuration step to set the gradle wrapper to use this version. See step 2 in `https://ci-master.corda.r3cev.com/admin/editBuildRunners.html?id=buildType:Corda_Build_BuildGradleCache`

These changes only take effect when the CI environment is correctly configured. Specifically the Build Configuration must define the following environment variable:
`CI` => must be set (to any value)
`GRADLE_BUILD_CACHE_URL` => must point to a Java VM running the Gradle Cache process.

and the build steps must include the following additional gradle command line parameters:
`--build-cache` (and optionally:  `--scan --stacktrace -Dorg.gradle.caching.debug=true`)


build.gradle
```diff
@@ -91,6 +91,9 @@ buildscript {
         maven {
             url "$artifactory_contextUrl/corda-releases"
         }
+        maven {
+            url 'https://plugins.gradle.org/m2/'
```
chrisr3: I think this `maven {}` repository is better replaced by
```
gradlePluginPortal()
```

chrisr3: And this (empty) file can be deleted too:
```
buildSrc/jarfilter/build.gradle
```

chrisr3:
And this (empty) file can be deleted too:
```
buildSrc/jarfilter/build.gradle
```

build.gradle
chrisr3: Actually, you could just do this instead:
```diff
diff --git a/build.gradle b/build.gradle
diff --git a/build.gradle b/build.gradle
index b24135b5d4..6c7ba15642 100644
--- a/build.gradle
+++ b/build.gradle

@@ -121,6 +117,8 @@ plugins {
     // but the DSL has some restrictions e.g can't be used on the allprojects section. So we should revisit this if there are improvements in Gradle.
     // Version 1.0.2 of this plugin uses capsule:1.0.1
     id "us.kirchmeier.capsule" version "1.0.2"
+
+    id 'com.gradle.build-scan' version '1.16'
 }
 
 ext {
```
This would allow you to remove the extra repository entirely, along with the line adding the plugin to the classpath and the `apply plugin: 'com.gradle.build-scan' line.

build.gradle
```diff
@@ -201,6 +204,12 @@ allprojects {
         if (project.path.startsWith(':experimental') && System.getProperty("experimental.test.enable") == null) {
             enabled = false
         }
+
+        // Required to use Gradle build cache (until Gradle 5.0 is released with default value of "append" set to false)
+        // See https://github.com/gradle/gradle/issues/5269 and https://github.com/gradle/gradle/pull/6419
+        extensions.configure(TypeOf.typeOf(JacocoTaskExtension), { ex ->
```
chrisr3: Does this work?
```gradle
extensions.configure(TypeOf.typeOf(JacocoTaskExtension)) { ex ->
    ex.append = false
}
```

josecoll:
@chrisr3 response from Gradle forum:
```
@josecoll We probably won't be back-porting this feature to the 4.x series. We only do bugfix releases.
Is there something in particular which would block you from upgrading to 5.0 (i.e. do you have some deprecation messages in your build)?
```

chrisr3:
The warning we're going to have trouble working around is:
```
Gradle does not allow passing null for the configuration action for CopySpec.from(). This behaviour has been deprecated and is scheduled to be removed in Gradle 5.0.
        at org.gradle.api.internal.file.copy.DefaultCopySpec.from(DefaultCopySpec.java:136)
        at org.gradle.api.tasks.AbstractCopyTask.from(AbstractCopyTask.java:324)
        at org.gradle.api.tasks.AbstractCopyTask$from.callCurrent(Unknown Source)
        at us.kirchmeier.capsule.task.Capsule.applyDefaultCapsuleSet(Capsule.groovy:150)
        at us.kirchmeier.capsule.task.Capsule.finalizeSettings(Capsule.groovy:125)
```
Because it's coming from the Capsule plugin, which seems to be unmaintained.

josecoll:
@chrisr3 anything outstanding preventing you from approving this PR?

buildCacheSettings.gradle
```diff
@@ -0,0 +1,14 @@
+// Gradle Build Cache configuration recommendation: https://docs.gradle.org/current/userguide/build_cache.html
+
```
chrisr3: Better:
```gradle
ext {
    isCiServerBuildCacheURL = ...
    grdleBuildCacheURL = ...
}
```

buildSrc/settings.gradle
```diff
@@ -1,2 +1,4 @@
 rootProject.name = 'buildSrc'
 include 'canonicalizer'
+
+apply from: new File(settingsDir, '../buildCacheSettings.gradle')
```
chrisr3: Why are you using `new File(settingsDir, ...)` here? I would have thought that
```gradle
apply from: 'buildCacheSettings.gradle
```
would be enough. According to the documentation, the file is assumed to be relative to the project directory.

settings.gradle
```diff
@@ -60,6 +60,8 @@ include 'samples:cordapp-configuration'
 include 'samples:network-verifier'
 include 'serialization'
 
+apply from: new File(settingsDir, 'buildCacheSettings.gradle')
```
chrisr3: Same comment as above.

buildCacheSettings.gradle
josecoll: picky ;-)

chrisr3: :nerd_face: But the `ext.blah` pattern is a nonsense that needs to die...!

buildCacheSettings.gradle
```diff
@@ -0,0 +1,15 @@
+// Gradle Build Cache configuration recommendation: https://docs.gradle.org/current/userguide/build_cache.html
+ext {
+    isCiServer = System.getenv().containsKey("CI")
```
chrisr3: BTW, Java _does_ have a [`System.getenv(String)`](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getenv(java.lang.String)) API.

josecoll: in this case we only case if it exists  (eg. want a boolean value, not its actual value).

chrisr3: `System.getenv("CI")` will return `null` if there is no such environment variable.

josecoll: It really makes no difference whether we check for NULL or a boolean.
Stop being so pedantic ... ;-)

chrisr3: Well, I'll let @Clintonio be the final arbiter :wink:

.ci/ci-gradle-build-cache-init.sh
```diff
@@ -0,0 +1,23 @@
...
+export GRADLE_BUILD_CACHE_URL="${GRADLE_BUILD_CACHE_URL:-http://localhost:5071/cache/}"
+export USE_GRADLE_DAEMON="${USE_GRADLE_DAEMON:-false}"
+export GRADLE_CACHE_DEBUG="${GRADLE_CACHE_DEBUG:-false}"
+export PERFORM_GRADLE_SCAN="${PERFORM_GRADLE_SCAN:---scan}"
+
+# GRADLE HOME
+# Required to use custom patched Gradle 4.9 with KEEP-ALIVE timeout fix
+# Currently built on Azure "gradle-build-client" machine at:
+# export GRADLE_HOME=~/gradle/install
+if [[ ! ${GRADLE_HOME} ]]; then
```
Clintonio: Could we not make the educated guess of ~/.gradle here to preserve sensible default behaviour?

buildCacheSettings.gradle
Clintonio: I don't mind either way here. The string constant "CI" here is a little bit too generic, I usually prefix any Corda specific envvar with CORDA_ to avoid collisions. Like CORDA_BINTRAY_USERNAME/CORDA_BINTRAY_PASSWORD, etc. 

I won't block the review on it but I'd strongly recommend a prefix.

.ci/ci-gradle-build-cache-init.sh
josecoll: Removed GRADLE_HOME as no longer needed.
Everything is driven via gradle wrapper.

josecoll:
Thanks @Clintonio 
Adjusted as per you final comments.

PR #3912, opened 2018-09-07
Record Hikari metrics

thschroeter:


thschroeter:
The metrics show up after the latest change.

screen shot 2018-09-10 at 11 44 12

PR #3915, opened 2018-09-07
ENT-2489: Additional optional parameter to specify target for RPC invocation.

vkolomeyko:
This is a split-out change from an enterprise feature, but since we do have: `RPCApi.RPC_TARGET_LEGAL_IDENTITY` header constant even in OS, it would make sense to extend RPC client API to populate this.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/SessionRejectException.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/SessionRejectException.kt
@@ -5,4 +5,4 @@ import net.corda.core.CordaException
 /**
  * An exception propagated and thrown in case a session initiation fails.
  */
-class SessionRejectException(reason: String) : CordaException(reason)
+class SessionRejectException(val reason: String) : CordaException(reason)
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/SessionRejectException.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/SessionRejectException.kt
@@ -5,4 +5,4 @@ import net.corda.core.CordaException
 /**
  * An exception propagated and thrown in case a session initiation fails.
  */
-class SessionRejectException(val reason: String) : CordaException(reason)
+class SessionRejectException(reason: String) : CordaException(reason)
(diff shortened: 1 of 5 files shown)


vkolomeyko:
Gentle reminder to review and approve this PR.

node/src/main/kotlin/net/corda/node/services/statemachine/SessionRejectException.kt
```diff
@@ -5,4 +5,4 @@ import net.corda.core.CordaException
 /**
  * An exception propagated and thrown in case a session initiation fails.
  */
-class SessionRejectException(reason: String) : CordaException(reason)
+class SessionRejectException(val reason: String) : CordaException(reason)
```
mnesbit: This looks like it hides the base class reason?

mnesbit: OK, doesn't hide, but the message property of the base is hidden for a reason.

vkolomeyko: I can rename `reason` in `SessionRejectException` to eliminate potential clash with base class.
But in order to be AMQP serializable it needs to have this sort of constructor, or else it will not marshal properly.

vkolomeyko:
Yes, indeed public API is affected in a sense that we are adding additional methods.
But in my view there is no harm in doing so. Even in current codebase there is already a header property `RPCApi.RPC_TARGET_LEGAL_IDENTITY`, so it will not harm to have an API to populate the value for this header property.

node/src/main/kotlin/net/corda/node/services/statemachine/SessionRejectException.kt
mnesbit: But you don't throw this from anywhere in your changes? Why is this happening now?
Certainly if it is a bug needs to be in a separate PR

vkolomeyko:
As discussed with @mnesbit - I have removed changes to exceptions into a separate PR and this PR only adds additional methods for populating `RPCApi.RPC_TARGET_LEGAL_IDENTITY` header.

PR #3917, opened 2018-09-07
CORDA-1968 Ensure we check for the trustroot existence and alias validity.

kchalkias:
There was no check for TrustStore validity during node startup. Apart from adding extra checks, similarly to the other two `KeyStore`s, I slightly refactored `validateKeyStores()` so it gets more readable.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -71,6 +71,9 @@ import net.corda.nodeapi.internal.NodeInfoAndSigned
 import net.corda.nodeapi.internal.SignedNodeInfo
 import net.corda.nodeapi.internal.config.CertificateStore
 import net.corda.nodeapi.internal.crypto.X509Utilities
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -694,30 +698,51 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
...
-        require(containCorrectKeys) {
-            "Identity certificate not found. " +
-                    "Please either copy your existing identity key and certificate from another node, " +
+    }
+
+    private data class AllCertificateStores(val trustStore: CertificateStore, val sslKeyStore: CertificateStore, val identitiesKeyStore: CertificateStore)
+
+    private fun validateKeyStores(): X509Certificate {
+        // Step 1. Check trustStore, sslKeyStore and identitiesKeyStore exist.
+        val certStores = getCertificateStores()
```
patrickkuo: you can use `val certStores = requireNotNull(getCertificateStores()) {...}` instead to avoid using `!!` later on.

PR #3919, opened 2018-09-10
CORDA-1949, CORDA-1950 - Miscellaneous doc fixes (#3760)

fenryka:
Backport of https://github.com/corda/corda/pull/3760

### DIFF
--- a/docs/source/hello-world-running.rst
+++ b/docs/source/hello-world-running.rst
@@ -109,10 +109,6 @@ We want to create an IOU of 99 with PartyB. We start the ``IOUFlow`` by typing:
 
 .. container:: codeset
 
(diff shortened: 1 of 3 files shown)


docs/source/hello-world-running.rst
```diff
@@ -109,10 +109,6 @@ We want to create an IOU of 99 with PartyB. We start the ``IOUFlow`` by typing:
 
 .. container:: codeset
 
-    .. code-block:: java
```
gendal: I know this is a total drive-by review but why are we deleting Java code samples? Our rule is "always Java, and Kotlin too if possible", no?  (i call this a drive-by because I've not looked at the broader context... it just jumped out at me.

PR #3922, opened 2018-09-10
CORDA-1391: Separate out Checkpoint serialization

distributedleetravis:
As per [Corda-1391](https://r3-cev.atlassian.net/browse/CORDA-1391), separate out checkpoint serialization using Kryo from non-checkpoint serialization using the AMQP serializers.

This PR contains the minimal changes required to introduce new classes for the checkpoint serialization use case, and eliminate all references to `UseCase.CHECKPOINT` when using AMQP serialization. Further simplification of the serialization framework (e.g. removing now-redundant references to `amqpMagic` in the AMQP serialization classes) will be attempted in a follow-up PR.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/serialization/CheckpointSerializationAPI.kt
+++ b/core/src/main/kotlin/net/corda/core/serialization/CheckpointSerializationAPI.kt
@@ -1,0 +1,190 @@
+package net.corda.core.serialization
+
+import net.corda.core.DoNotImplement
+import net.corda.core.KeepForDJVM
+import net.corda.core.crypto.SecureHash
+import net.corda.core.serialization.internal.effectiveSerializationEnv
+import net.corda.core.utilities.ByteSequence
+import net.corda.core.utilities.sequence
+
+/**
+ * An abstraction for serializing and deserializing objects, with support for versioning of the wire format via
+ * a header / prefix in the bytes.
+ */
+abstract class CheckpointSerializationFactory {
+    /**
+     * Deserialize the bytes in to an object, using the prefixed bytes to determine the format.
+     *
+     * @param byteSequence The bytes to deserialize, including a format header prefix.
+     * @param clazz The class or superclass or the object to be deserialized, or [Any] or [Object] if unknown.
+     * @param context A context that configures various parameters to deserialization.
+     */
+    abstract fun <T : Any> deserialize(byteSequence: ByteSequence, clazz: Class<T>, context: CheckpointSerializationContext): T
+
+    /**
+     * Serialize an object to bytes using the preferred serialization format version from the context.
+     *
+     * @param obj The object to be serialized.
+     * @param context A context that configures various parameters to serialization, including the serialization format version.
+     */
+    abstract fun <T : Any> serialize(obj: T, context: CheckpointSerializationContext): SerializedBytes<T>
+
+    /**
+     * If there is a need to nest serialization/deserialization with a modified context during serialization or deserialization,
+     * this will return the current context used to start serialization/deserialization.
+     */
+    val currentContext: CheckpointSerializationContext? get() = _currentContext.get()
+
+    /**
+     * A context to use as a default if you do not require a specially configured context.  It will be the current context
+     * if the use is somehow nested (see [currentContext]).
+     */
+    val defaultContext: CheckpointSerializationContext get() = currentContext ?: effectiveSerializationEnv.checkpointContext
+
+    private val _currentContext = ThreadLocal<CheckpointSerializationContext?>()
+
+    /**
+     * Change the current context inside the block to that supplied.
+     */
+    fun <T> withCurrentContext(context: CheckpointSerializationContext?, block: () -> T): T {
+        val priorContext = _currentContext.get()
+        if (context != null) _currentContext.set(context)
+        try {
+            return block()
+        } finally {
+            if (context != null) _currentContext.set(priorContext)
+        }
+    }
+
+    /**
+     * Allow subclasses to temporarily mark themselves as the current factory for the current thread during serialization/deserialization.
+     * Will restore the prior context on exiting the block.
+     */
+    fun <T> asCurrent(block: CheckpointSerializationFactory.() -> T): T {
+        val priorContext = _currentFactory.get()
+        _currentFactory.set(this)
+        try {
+            return this.block()
+        } finally {
+            _currentFactory.set(priorContext)
+        }
+    }
+
+    companion object {
+        private val _currentFactory = ThreadLocal<CheckpointSerializationFactory?>()
+
+        /**
+         * A default factory for serialization/deserialization, taking into account the [currentFactory] if set.
+         */
+        val defaultFactory: CheckpointSerializationFactory get() = currentFactory ?: effectiveSerializationEnv.checkpointSerializationFactory
+
+        /**
+         * If there is a need to nest serialization/deserialization with a modified context during serialization or deserialization,
+         * this will return the current factory used to start serialization/deserialization.
+         */
+        val currentFactory: CheckpointSerializationFactory? get() = _currentFactory.get()
+    }
+}
+
+/**
+ * Parameters to serialization and deserialization.
+ */
+@KeepForDJVM
+@DoNotImplement
+interface CheckpointSerializationContext {
+    /**
+     * If non-null, apply this encoding (typically compression) when serializing.
+     */
+    val encoding: SerializationEncoding?
+    /**
+     * The class loader to use for deserialization.
+     */
+    val deserializationClassLoader: ClassLoader
+    /**
+     * A whitelist that contains (mostly for security purposes) which classes can be serialized and deserialized.
+     */
+    val whitelist: ClassWhitelist
+    /**
+     * A whitelist that determines (mostly for security purposes) whether a particular encoding may be used when deserializing.
+     */
+    val encodingWhitelist: EncodingWhitelist
+    /**
+     * A map of any addition properties specific to the particular use case.
+     */
+    val properties: Map<Any, Any>
+    /**
+     * Duplicate references to the same object preserved in the wire format and when deserialized when this is true,
+     * otherwise they appear as new copies of the object.
+     */
+    val objectReferencesEnabled: Boolean
+
+    /**
+     * Helper method to return a new context based on this context with the property added.
+     */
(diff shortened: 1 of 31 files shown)


node/src/main/kotlin/net/corda/node/serialization/kryo/Kryo.kt
```diff
@@ -275,16 +273,16 @@ object SignedTransactionSerializer : Serializer<SignedTransaction>() {
     }
 }
 
+// Retained because it's part of a public API.
 sealed class UseCaseSerializer<T>(private val allowedUseCases: EnumSet<SerializationContext.UseCase>) : Serializer<T>() {
```
fenryka: Do we need this at all still?

node/src/test/kotlin/net/corda/node/serialization/kryo/KryoTests.kt
```diff
@@ -36,15 +35,8 @@ import java.util.*
...
-class TestScheme : AbstractKryoSerializationScheme() {
-    override fun canDeserializeVersion(magic: CordaSerializationMagic, target: SerializationContext.UseCase): Boolean {
-        return magic == kryoMagic && target != SerializationContext.UseCase.RPCClient
-    }
-
-    override fun rpcClientKryoPool(context: SerializationContext): KryoPool = throw UnsupportedOperationException()
-
-    override fun rpcServerKryoPool(context: SerializationContext): KryoPool = throw UnsupportedOperationException()
-}
+//TODO: get rid of this.
```
fenryka: should probably log a Jira for this

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/custom/PrivateKeySerializer.kt
```diff
@@ -13,7 +12,7 @@ import java.util.*
 
 object PrivateKeySerializer : CustomSerializer.Implements<PrivateKey>(PrivateKey::class.java) {
 
-    private val allowedUseCases = EnumSet.of(Storage, Checkpoint)
+    private val allowedUseCases = EnumSet.of(Storage)
```
fenryka: Seems overkill for a single value

testing/test-utils/src/main/kotlin/net/corda/testing/internal/InternalSerializationTestHelpers.kt
```diff
@@ -33,16 +33,15 @@ internal fun createTestSerializationEnv(label: String): SerializationEnvironment
...
     }
     return object : SerializationEnvironmentImpl(
             factory,
             AMQP_P2P_CONTEXT,
             AMQP_RPC_SERVER_CONTEXT,
             AMQP_RPC_CLIENT_CONTEXT,
             AMQP_STORAGE_CONTEXT,
-            KRYO_CHECKPOINT_CONTEXT
+            KRYO_CHECKPOINT_CONTEXT,
+            CheckpointSerializationFactoryImpl(KryoServerSerializationScheme())
```
fenryka: I kinda don't like this structure now

When it was a single factory that applied to all other elements it kinda made sense, now it's very ad-hoc and as an interface is very messy in terms of which factory applies to which context. Can we clean this up to store them as pairs or something that makes the relationships more obvious

core/src/main/kotlin/net/corda/core/serialization/CheckpointSerializationAPI.kt
```diff
@@ -0,0 +1,194 @@
...
+    fun withEncoding(encoding: SerializationEncoding?): CheckpointSerializationContext
+
+    /**
+     * A shallow copy of this context but with the given encoding whitelist.
+     */
+    fun withEncodingWhitelist(encodingWhitelist: EncodingWhitelist): CheckpointSerializationContext
+}
+
+/*
+ * The following extension methods are disambiguated from the AMQP-serialization methods by requiring that an
```
fenryka: this.... I kinda hate :D it smells a bit magic and is goign to be non-obvious reading the code this is going on.

node/src/main/kotlin/net/corda/node/serialization/kryo/Kryo.kt
distributedleetravis: No, but if we delete it the API checker will complain.

core/src/main/kotlin/net/corda/core/serialization/CheckpointSerializationAPI.kt
distributedleetravis: Renamed to `checkpointSerialize` and `checkpointDeserialize` respectively.

node/src/test/kotlin/net/corda/node/serialization/kryo/KryoTests.kt
distributedleetravis: Will add to the follow-up JIRA (see below)

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/custom/PrivateKeySerializer.kt
distributedleetravis: Agreed, will tidy as part of follow-up.

testing/test-utils/src/main/kotlin/net/corda/testing/internal/InternalSerializationTestHelpers.kt
distributedleetravis: I agree that it's nasty, but I'd like to punt improving it to a follow-up JIRA/PR, just to limit the nuisance caused by this one.

distributedleetravis:
Requesting review from @mikehearn as there's an `api-current` change here (introducing an entirely new abstract interface), and there are some changes to methods in the public API where the existing method is now deprecated and throws `UnsupportedOperationException` (it would be better to remove these altogether, but this would break API compatibility).

node/src/main/kotlin/net/corda/node/serialization/kryo/Kryo.kt
mikehearn: Hm, but this is in the `node` module - which isn't at this time public API?

.ci/api-current.txt
```diff
@@ -4360,6 +4360,12 @@ public interface net.corda.core.schemas.QueryableState extends net.corda.core.co
 ##
 public interface net.corda.core.schemas.StatePersistable
 ##
+public interface net.corda.core.serialization.CheckpointSerializationScheme
```
mikehearn: Do we really need this in the public API or can we mark it `@CordaInternal`?

core/src/main/kotlin/net/corda/core/serialization/CheckpointSerializationAPI.kt
```diff
@@ -0,0 +1,190 @@
...
+import net.corda.core.serialization.internal.effectiveSerializationEnv
+import net.corda.core.utilities.ByteSequence
+import net.corda.core.utilities.sequence
+import java.io.NotSerializableException
+
+/**
+ * A class for serializing and deserializing objects at checkpoints, using Kryo serialization.
+ */
+@KeepForDJVM
+class CheckpointSerializationFactory(
```
mikehearn: Is this new code? Can it be *not* in the public API?

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -226,10 +233,18 @@ enum class ContextPropertyKeys {
 @KeepForDJVM
 object SerializationDefaults {
     val SERIALIZATION_FACTORY get() = effectiveSerializationEnv.serializationFactory
+    val CHECKPOINT_SERIALIZATION_FACTORY get() = effectiveSerializationEnv.checkpointSerializationFactory
     val P2P_CONTEXT get() = effectiveSerializationEnv.p2pContext
     @DeleteForDJVM val RPC_SERVER_CONTEXT get() = effectiveSerializationEnv.rpcServerContext
     @DeleteForDJVM val RPC_CLIENT_CONTEXT get() = effectiveSerializationEnv.rpcClientContext
     @DeleteForDJVM val STORAGE_CONTEXT get() = effectiveSerializationEnv.storageContext
+    @Deprecated("Checkpoint context is no longer available from SerializationDefaults, but must be obtained from CheckpointSerializationDefaults instead.")
+    @DeleteForDJVM val CHECKPOINT_CONTEXT: SerializationContext get() =
```
mikehearn: It makes no difference if the behaviour is just to throw an exception, we may as well just remove the method entirely - all you're doing here is changing the exception message in use.

I strongly suspect nobody has ever used this - I can't think of a single use case for it - so I'm willing to approve an exception to the usual rules here to get rid of checkpoint serialisation stuff from the public API.

distributedleetravis:
I've moved everything checkpoint-related into `internal` packages and cut the umbilical chord connecting it to the public API, which has required some changes to `api-current`.

core-deterministic/src/main/kotlin/net/corda/core/serialization/internal/CheckpointSerializationFactory.kt
```diff
@@ -0,0 +1,74 @@
+package net.corda.core.serialization.internal
+
+import net.corda.core.KeepForDJVM
+import net.corda.core.serialization.SerializedBytes
+import net.corda.core.utilities.ByteSequence
+import java.io.NotSerializableException
+
+/**
+ * A deterministic version of [CheckpointSerializationFactory] that does not use thread-locals to manage serialization
```
mikehearn: Thanks - this doc comment reads weird now. Is it a copy/paste error?

testing/test-utils/src/main/kotlin/net/corda/testing/internal/InternalSerializationTestHelpers.kt
fenryka: Can we see that Jira logged (Jiras?) just so we make sure none of the points are forgotten along the way :)

core/src/main/kotlin/net/corda/core/serialization/CheckpointSerializationAPI.kt
fenryka: 👍

fenryka:
As long as all the items mentioned are captured inthe follow up Jira it looks good to me

PR #3923, opened 2018-09-10
CORDA-1764: Make shell use picocli for parsing command line options

anthonykeenan:
Also: 
* fixed some issues with the logging dependencies
* cleaned up some gradle


tools/shell-cli/build.gradle
```diff
@@ -19,7 +19,9 @@ apply plugin: 'com.jfrog.artifactory'
 
 dependencies {
     compile project(':tools:shell')
-    compile "org.slf4j:slf4j-simple:$slf4j_version"
+    compile project(':tools:cliutils')
+    compile "org.apache.logging.log4j:log4j-slf4j-impl:$log4j_version"
+    compile "org.slf4j:jul-to-slf4j:$slf4j_version"
```
chrisr3: Do you _really_ need `jul-to-slf4j`? This logging bridge between `java.util.logging` and SLF4J is not "drop-in configurable", so how is it being used?

anthonykeenan: Yes, I think I do. I think CRaSH uses it, and I think I'm initialising it here: https://github.com/corda/corda/pull/3923/files#diff-e4e2b3306a81fb0b372f5fe7f5fe8175R61
If I remove that, the logging configuration doesn't take.

tools/shell-cli/src/main/kotlin/net/corda/tools/shell/ShellCmdLineOptions.kt
```diff
@@ -0,0 +1,195 @@
...
+
+    @Option(
+            names = ["-c", "--cordapp-directory"],
+            description = ["The path to the directory containing CorDapp jars, CorDapps are required when starting flows."]
+    )
+    var cordappDirectory: Path? = null
+
+    @Option(
+            names = ["-o", "--commands-directory"],
+            description = ["The path to the directory containing additional CrAsH shell commands."]
```
tlil: CRaSH

tools/shell-cli/src/main/kotlin/net/corda/tools/shell/ShellCmdLineOptions.kt
```diff
@@ -0,0 +1,195 @@
...
+class ShellCmdLineOptions {
+    @Option(
+            names = ["-f", "--config-file"],
+            description = ["The path to the shell configuration file, used instead of providing the rest of command line options."]
+    )
+    var configFile: Path? = null
+
+    @Option(
+            names = ["-c", "--cordapp-directory"],
+            description = ["The path to the directory containing CorDapp jars, CorDapps are required when starting flows."]
```
tlil: JARs

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
```diff
@@ -147,11 +149,11 @@ abstract class CordaCliWrapper(val alias: String, val description: String) : Cal
 }
 
 /**
- * Converter from String to log4j logging Level.
+ * Converter from String to slf4j logging Level.
  */
 class LoggingLevelConverter : ITypeConverter<Level> {
     override fun convert(value: String?): Level {
-        return value?.let { Level.getLevel(it) }
+        return value?.let { Level.valueOf(it) }
```
tlil: Could potentially make this a bit better at guessing.. E.g., for provided lower-case values and partial matches (if only 1 level matches the input, go and assume that's what the user meant)

tools/shell-cli/build.gradle
anthonykeenan: Here I think: https://github.com/crashub/crash/blob/master/shell/src/main/java/org/crsh/plugin/ServiceLoaderDiscovery.java#L25

PR #1009, opened 2017-07-11
Add all properties of a class into the amqp envelope even if not set by the constructor

fenryka:
As they can still be used by the carpenter

PR #3938, opened 2018-09-13
Corda 1922 serialize states with calculated values

distributedleetravis:
A new type of `PropertyAccessor` is introduced to the AMQP serialization code, `CalculatedPropertyAccessor`, which represents a calculated property that can be read (via a getter method) but not written as there is no corresponding setter method or constructor parameter slot. Values read by this accessor are serialized, and included in the AMQP schema of serialized types, but are normally thrown away on deserialization.

However, when the `ClassCarpenter` comes to synthesize a new class to match the AMQP schema, it _will_ create a constructor slot for the calculated value, which will then be injected into new instances of the synthesized class. This enables classes implementing interfaces such as `FungibleState` to be correctly synthesized by the class carpenter even when those classes (such as `Cash`) expose calculated properties (such as `exitKeys`).

The end-user indicates that a read-only method (which must be a getter of some kind) is to be treated in this way by annotating it with a new annotation, `SerializableCalculatedProperty`. If the property is represented by a Kotlin `val`, it must be annotated with `get:SerializableCalculatedProperty`, indicating that it is the generated getter method that is to be marked with the annotation.

These changes implement one possible way of resolving [Corda-1922](https://r3-cev.atlassian.net/browse/CORDA-1922).


finance/src/main/kotlin/net/corda/finance/contracts/asset/Cash.kt
```diff
@@ -59,6 +60,7 @@ class Cash : OnLedgerAsset<Currency, Cash.Commands, Cash.State>() {
         constructor(deposit: PartyAndReference, amount: Amount<Currency>, owner: AbstractParty)
                 : this(Amount(amount.quantity, Issued(deposit, amount.token)), owner)
 
+        @get:SerializableCalculatedProperty
```
mikehearn: Couldn't it be on the super-interface rather than cash?

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
```diff
@@ -13,6 +13,13 @@ import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.utilities.sequence
 import java.sql.Blob
 
+/**
+ * Used to annotate methods which expose calculated values that we want to be serialized for use by the class carpenter.
```
mikehearn: I wonder if per-property is right, or if annotating on the class/interface level makes more sense (or both?)

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ObjectSerializer.kt
```diff
@@ -109,8 +110,19 @@ open class ObjectSerializer(val clazz: Type, factory: SerializerFactory) : AMQPS
...
+
+                    when(accessor) {
+                        is PropertyAccessorConstructor -> accessor.initialPosition to value
+                        is CalculatedPropertyAccessor -> null
+                        else -> throw UnsupportedOperationException(
+                                "${accessor::class.simpleName} accessor not supported " +
+                                        "for constructor-based object building")
+                    }
+                }
+                .sortedWith(compareBy { it.first })
```
mikehearn: What happens here if there are duplicate keys in the stream?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
```diff
@@ -96,9 +97,24 @@ fun Class<out Any?>.propertyDescriptors(): Map<String, PropertyDescriptor> {
             .validated()
 }
 
+/**
+ * Obtain [PropertyDescriptor]s for those calculated properties of a class which are annotated with
+ * [SerializableCalculatedProperty]
+ */
+fun Class<out Any?>.calculatedPropertyDescriptors(): Map<String, PropertyDescriptor> =
```
mikehearn: Mark as private or internal.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
```diff
@@ -96,9 +97,24 @@ fun Class<out Any?>.propertyDescriptors(): Map<String, PropertyDescriptor> {
...
+        superclassChain().withInterfaces().declaredMethods()
+                .thatArePublic()
+                .thatAreCalculated()
+                .toCalculatedProperties()
+
 // Generate the sequence of classes starting with this class and ascending through it superclasses.
 private fun Class<*>.superclassChain() = generateSequence(this, Class<*>::getSuperclass)
 
+private fun Sequence<Class<*>>.withInterfaces() = flatMap {
+    sequenceOf(it) +
```
mikehearn: odd line break?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
```diff
@@ -108,18 +124,45 @@ private fun Sequence<Class<*>>.declaredMethods() = flatMap { it.declaredMethods.
...
+    return methodsByName.mapValues { (_, method) -> PropertyDescriptor(null, null, method) }
+}
+
 // Select only those methods that are isX/getX/setX methods
-private fun Sequence<Method>.thatArePropertyMethods() = map { method ->
-    propertyMethodRegex.find(method.name)?.let { result ->
+private fun Sequence<Method>.thatArePropertyMethods() = mapNotNull(::getPropertyNamedMethod)
+
+private fun getPropertyNamedMethod(method: Method): PropertyNamedMethod? {
+    return propertyMethodRegex.find(method.name)?.let { result ->
```
mikehearn: regex is probably quite expensive to have on the deserialisation hot-path!

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertySerializers.kt
```diff
@@ -159,16 +159,34 @@ class PropertyAccessorGetterSetter(
...
+            "${serializer.name}($initialPosition)"
+}
+
+/**
+ * Implementation of [PropertyAccessor] representing a calculated property of an object that is serialized
+ * so that it can be used by the class carpenter, but ignored on deserialisation as there is no setter or
+ * constructor parameter to receive its value.
+ *
+ * This will only be created for calculated properties that are accessible via no-argument methods annotated
+ * with [SerializeForCarpenter].
```
mikehearn: The new name is definitely better than the old name :)

serialization/src/test/kotlin/net/corda/serialization/internal/carpenter/CalculatedValuesToClassCarpenterTests.kt
```diff
@@ -0,0 +1,68 @@
...
+
+        val concrete = serSchema.types[0] as CompositeType
+        assertEquals(3, concrete.fields.size)
+        assertEquals("doubled", concrete.fields[0].name)
+        assertEquals("int", concrete.fields[0].type)
+        assertEquals("i", concrete.fields[1].name)
+        assertEquals("int", concrete.fields[1].type)
+        assertEquals("squared", concrete.fields[2].name)
+        assertEquals("string", concrete.fields[2].type)
+
```
fenryka: additional line

fenryka:
It's worth noting, from a historical perspective, we thought about this a while ago (https://github.com/corda/corda/pull/1009). Looks like we have a use-case now to justify it :)

finance/src/main/kotlin/net/corda/finance/contracts/asset/Cash.kt
distributedleetravis: It could be, but we would then be deciding for all possible implementations that this value would be calculated (rather than passed in through a constructor parameter, which is a perfectly valid option).

core/src/main/kotlin/net/corda/core/serialization/SerializationAPI.kt
distributedleetravis: We could support annotating classes/interfaces to indicate that every method matching the "getter" contract is calculated unless we know otherwise (i.e. there's actually a corresponding setter or constructor parameter), but it's a bit magic-y and I would prefer making this an explicit, and exceptional, option rather than something you could accidentally enable for something you didn't intend to.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ObjectSerializer.kt
distributedleetravis: The short answer is: what's always happened. The long answer is: I don't know, let's raise a separate JIRA to investigate whether this might be a problem, and what to do about it if it is.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
distributedleetravis: I forget what's cached here - we shouldn't be running through _any_ of this code more than once per class...

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertySerializers.kt
```diff
@@ -198,8 +216,12 @@ abstract class PropertySerializers(
 
     val size get() = serializationOrder.size
     abstract val byConstructor: Boolean
+    val deserializableSize get() = serializationOrder.count { !it.isCalculated }
```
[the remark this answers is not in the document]
distributedleetravis: I have a feeling it's read precisely once, so not worth paying the _lazy_ tax, but will check just in case.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertySerializers.kt
```diff
@@ -198,8 +216,12 @@ abstract class PropertySerializers(
...
 }
 
+inline val List<PropertyAccessor>.ignoringCalculatedProperties: List<PropertyAccessor> get() =
+        filter { !it.isCalculated }
```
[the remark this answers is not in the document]
distributedleetravis: Same as above. It's definitely not cached, but I also don't think it ever gets called more than once.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/PropertyDescriptor.kt
distributedleetravis: Considered adding caches for property descriptor lookups, but it looks like they're only ever performed during serializer construction, and serializers are cached by type, so the benefits are negligible (and the memory cost probably quite high).

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ObjectSerializer.kt
distributedleetravis: What's being sorted here is the property accessors for the constructor, zipped together with the objects recovered from the stream. We've already checked that the number of objects and the number of accessors match up, and will throw an exception if they don't:

```
if (obj.size != propertySerializers.size) {
                throw AMQPNotSerializableException(type, "${obj.size} objects to deserialize, but " +
                        "${propertySerializers.size} properties in described type $typeName")
            }
```

distributedleetravis:
@fenryka I've added a (deep breath) `JavaCalculatedValuesToClassCarpenterTest` which verbosely replicates the Kotlin one, and updated the serialization documentation.

mikehearn:
heh, good point. Looks like I changed my mind when I saw our own interfaces require it for anything to work, doh.

PR #3939, opened 2018-09-13
CORDA-1959 - Enforce backwards compatibility of new CLI's

roastario:
* https://r3-cev.atlassian.net/browse/CORDA-1959


testing/test-cli/src/main/resources/net.corda.tools.shell.StandaloneShell.yml
```diff
@@ -0,0 +1,119 @@
+- commandName: "<main class>"
```
anthonykeenan: I think this file should be in the shell-cli test project rather than the test-cli project?

roastario:
Whilst is not documented as such. When you create a test without a corresponding yml, it prints out a warning along with the YAML ready to be copied and pasted into a file.

PR #3940, opened 2018-09-13
CORDA-1764: Make blobinspector use picocli base class

anthonykeenan:
Also harmonise the documentation for all of the CLI apps.


docs/source/blob-inspector.rst
```diff
@@ -90,3 +90,25 @@ Here's what a node-info file from the node's data directory may look like:
...
+    blob-inspector [-hvV] [--full-parties] [--install-shell-extensions] [--schema]
+                   [--format=type] [--input-format=type]
+                   [--logging-level=<loggingLevel>] [SOURCE]
+
+* ``--format=type``: Output format. Possible values: [YAML, JSON]. Default: YAML.
+* ``--input-format=type``: Input format. If the file can't be decoded with the given value it's auto-detected, so you should
+  never normally need to specify this. Possible values [BINARY, HEX, BASE64]. Default: BINARY.
+* ``--full-parties``: Display the owningKey and certPath properties of Party and PartyAndReference objects respectively.
+* ``--schema``: Print the blob's schema first.
+* ``--verbose``, ``--log-to-console``, ``-V``: If set, prints logging to the console as well as to a file.
```
tlil: Are both this and version abbreviated to `-V` (capital V)?

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
```diff
@@ -68,17 +55,19 @@ class BlobInspector : Runnable {
...
-    var verbose: Boolean = false
+    override fun runProgram() = run(System.out)
 
-    override fun run() = run(System.out)
-
-    fun run(out: PrintStream) {
+    override fun initLogging() {
         if (verbose) {
-            System.setProperty("logLevel", "trace")
+            loggingLevel = Level.TRACE
```
tlil: In the other utilities, this is setting log-to-console as well, isn't it? Does this make the utility behave differently to the others?

anthonykeenan: Yes it does, it has a custom `log4j2.xml` file which only logs to console. I can change it to use the shared one like I did with the bootstrapper. Wasn't sure about making too many changes.

anthonykeenan: Didn't want to incur the Wrath of Kat, my favourite star trek fim. @fenryka

docs/source/blob-inspector.rst
anthonykeenan: I've fixed it, it was wrong for all of the docs.

tools/blobinspector/src/main/kotlin/net/corda/blobinspector/BlobInspector.kt
tlil: Cool, will leave it to Kat :-)

fenryka: My argument here is that the blob inspector is a command line utility, not a server / service. I would expect to run it and have it display expected things... logging is going to get in the way of that 99% of the time

anthonykeenan: I agree, I'll leave it the way it is then.

PR #3945, opened 2018-09-14
RELEASE - Release script to extract which issues are in a release

fenryka:
In addition, format the output for inclusion in the release notes. Not 100% sure
if this should be in the tree or out in a separate repo but we have
*a lot* of small repo#s now and this is fine to live here unless
anyone vehemently disagrees

I've created a top level dir called release-tools to give it somewhere to live, of course, can really go anywhere

If people think this is fine/useful I'll forward port it to master for use in V4.X


release-tools/fixedIn.py
```diff
@@ -0,0 +1,151 @@
...
+#-------------------------------------------------------------------------------
+#
+# The purpose of this script is to operate on a release branch for a release
+# and automagically infer the list of fixed issues contained within that
+# release for inclusion within the release notes
+#
+#  Usage
+# -------
+#
+# ./fixedIn.py <previous release tag> <Jira release version> <Jira user>
```
josecoll: I'm trying to think of a better name for this tool:
fixDiffer, releaseDiffer, releaseFixDiffer

release-tools/fixedIn.py
```diff
@@ -0,0 +1,151 @@
...
+#
+# The previous release tag refers to the previous release on the branch or, if
+# a new release branch from master, the branch point. For example, for the 3.3
```
josecoll: The example sounds a little confusing .... but I think I get the drift of the explanation.
eg. the tool must be run on the *same* branch as the commit you are diff'ing against.

release-tools/fixedIn.py
fenryka: issuesIn? essentially it takes a release branch and spits out all the issues fixed since a point in the past... 

releaseContains
jirasIn

I'm open to suggestions  :)

release-tools/fixedIn.py
fenryka: I'll re-word

release-tools/fixedIn.py
```diff
@@ -0,0 +1,151 @@
...
+# the native OS keyring implementation.
+#
+# The script should be run on the relevant release branch within the git
+# repository.
+#
+#  For Example
+# -------------
+# ./fixedIn.py release-V3.1 "Corda 3.3" some.user@r3.com
+#
+#  Pre Requisites
```
josecoll: For total noobs can you list the commands one would need to run to install these pre-reqs.

release-tools/fixedIn.py
Clintonio: `findJiraRefsSince.py <commit>` perhaps?

release-tools/fixedIn.py
Clintonio: Maybe make an analogy to git bisect, which also has a similar set of rules about how it can be run.

release-tools/fixedIn.py
```diff
@@ -0,0 +1,151 @@
...
+JIRA_MAX_RESULTS = 50
+
+#-------------------------------------------------------------------------------
+
+#
+# For a given user (provide via the command line) authenticate with Jira and
+# return an interface object instance
+#
+def jiraLogin(user) :
+    password = keyring.get_password ('jira', user)
```
Clintonio: Apparently this `keyring` is Windows compatible. I'm legitimately surprised enough to make a comment on it.

release-tools/fixedIn.py
```diff
@@ -0,0 +1,151 @@
...
+    return "* %s [`%s <%s/browse/%s>`_]" % (
+            issue.fields.summary,
+            issue.key,
+            R3_JIRA_ADDR,
+            issue.key)
+
+#-------------------------------------------------------------------------------
+
+def main (args_) :
+    jiraMatch = re.compile("(CORDA-\d+|ENT-\d+)")
```
Clintonio: I'd also recommend "DEVOPS-" in the case that a devops ticket has been referenced. Might also be worth double checking if any "CN-" tickets have made it into Corda somehow.

PR #3950, opened 2018-09-17
CORDA-1986 Ensure key alias format is supported by the major HSM vendors

kchalkias:
According to the most restrictive HSM vendor until now (Azure Key Vault), key aliases should be 1-127 characters in length containing only 0-9, a-z, A-Z, and - (nothing else).
_Note that the rest of tested vendors have more relaxed restrictions (i.e., Utimaco supports up to a maximum of 256 characters, including symbols and space)._

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
@@ -41,15 +41,21 @@ object X509Utilities {
     val DEFAULT_IDENTITY_SIGNATURE_SCHEME = Crypto.EDDSA_ED25519_SHA512
     val DEFAULT_TLS_SIGNATURE_SCHEME = Crypto.ECDSA_SECP256R1_SHA256
 
-    // TODO This class is more of a general purpose utility class and as such these constants belong elsewhere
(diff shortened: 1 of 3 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/X509Utilities.kt
```diff
@@ -41,15 +41,21 @@ object X509Utilities {
...
-    // TODO This class is more of a general purpose utility class and as such these constants belong elsewhere
+    // TODO This class is more of a general purpose utility class and as such these constants belong elsewhere.
     // Aliases for private keys and certificates.
     const val CORDA_ROOT_CA = "cordarootca"
     const val CORDA_INTERMEDIATE_CA = "cordaintermediateca"
     const val CORDA_CLIENT_TLS = "cordaclienttls"
     const val CORDA_CLIENT_CA = "cordaclientca"
 
-    // TODO These don't need to be prefixes, but can be the full aliases.
+    // TODO These don't need to be prefixes, but can be the full aliases. However, because they are used as key aliases
```
patrickkuo: the prefix are used for services identity, since we have removed support for service identity, we could change these (`identity-private-key`, `distributed-notary-private-key`)to constants, perhaps in a separated PR.

PR #3955, opened 2018-09-17
ENT-2431 Lay foundations for caching metrics

rick-r3:
Ground work for further changes.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -63,10 +63,7 @@ import net.corda.node.services.statemachine.*
 import net.corda.node.services.transactions.*
 import net.corda.node.services.upgrade.ContractUpgradeServiceImpl
 import net.corda.node.services.vault.NodeVaultService
-import net.corda.node.utilities.AffinityExecutor
-import net.corda.node.utilities.JVMAgentRegistry
-import net.corda.node.utilities.NamedThreadFactory
-import net.corda.node.utilities.NodeBuildProperties
+import net.corda.node.utilities.*
 import net.corda.nodeapi.internal.NodeInfoAndSigned
 import net.corda.nodeapi.internal.SignedNodeInfo
 import net.corda.nodeapi.internal.config.CertificateStore
@@ -115,6 +112,7 @@ import net.corda.core.crypto.generateKeyPair as cryptoGenerateKeyPair
 // TODO Log warning if this node is a notary but not one of the ones specified in the network parameters, both for core and custom
 abstract class AbstractNode<S>(val configuration: NodeConfiguration,
                                val platformClock: CordaClock,
+                               cacheFactoryPrototype: NamedCacheFactory,
                                protected val versionInfo: VersionInfo,
                                protected val cordappLoader: CordappLoader,
                                protected val serverThread: AffinityExecutor.ServiceAffinityExecutor,
@@ -125,7 +123,8 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
     @Suppress("LeakingThis")
     private var tokenizableServices: MutableList<Any>? = mutableListOf(platformClock, this)
 
-    val metricRegistry = MetricRegistry()
+    protected val metricRegistry = MetricRegistry()
+    protected val cacheFactory = cacheFactoryPrototype.bindWithConfig(configuration).bindWithMetrics(metricRegistry).tokenize()
     val monitoringService = MonitoringService(metricRegistry).tokenize()
 
     protected val runOnStop = ArrayList<() -> Any?>()
@@ -142,7 +141,7 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
     }
 
     val schemaService = NodeSchemaService(cordappLoader.cordappSchemas, configuration.notary != null).tokenize()
-    val identityService = PersistentIdentityService(metricRegistry).tokenize()
+    val identityService = PersistentIdentityService(cacheFactory).tokenize()
     val database: CordaPersistence = createCordaPersistence(
             configuration.database,
             identityService::wellKnownPartyFromX500Name,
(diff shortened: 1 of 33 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/performance/Reporter.kt
```diff
@@ -12,12 +12,14 @@ fun startReporter(shutdownManager: ShutdownManager, metricRegistry: MetricRegist
     val jmxReporter = thread {
         JmxReporter.forRegistry(metricRegistry).inDomain("net.corda").createsObjectNamesWith { _, domain, name ->
             // Make the JMX hierarchy a bit better organised.
-            val category = name.substringBefore('.')
+            val category = name.substringBefore('.').substringBeforeLast('/')
+            val component = name.substringBefore('.').substringAfterLast('/', "")
             val subName = name.substringAfter('.', "")
             if (subName == "")
                 ObjectName("$domain:name=$category")
             else
```
blsemo: What happens with component if subname is empty?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -124,6 +124,10 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
 
     @Suppress("LeakingThis")
     private var tokenizableServices: MutableList<Any>? = mutableListOf(platformClock, this)
+
+    val metricRegistry = MetricRegistry()
```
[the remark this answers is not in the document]
rick-r3: It was already a `val` a few lines below.  But it was `private`, so I should fix that.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/performance/Reporter.kt
rick-r3: Whoops.  It should be included in the `ObjectName` on that branch of the `if`.

rick-r3:
I implemented the factory idea.

node/src/main/kotlin/net/corda/node/utilities/NodeNamedCache.kt
```diff
@@ -0,0 +1,54 @@
...
+class DefaultNamedCacheFactory private constructor(private val metricRegistry: MetricRegistry?, private val nodeConfiguration: NodeConfiguration?) : NamedCacheFactory, SingletonSerializeAsToken() {
+    constructor() : this(null, null)
+
+    override fun bindWithMetrics(metricRegistry: MetricRegistry): NamedCacheFactory = DefaultNamedCacheFactory(metricRegistry, this.nodeConfiguration)
+    override fun bindWithConfig(nodeConfiguration: NodeConfiguration): NamedCacheFactory = DefaultNamedCacheFactory(this.metricRegistry, nodeConfiguration)
+
+    override fun <K, V> buildNamed(caffeine: Caffeine<in K, in V>, name: String): Cache<K, V> {
+        checkNotNull(metricRegistry)
+        checkNotNull(nodeConfiguration)
+        return caffeine.buildNamed<K, V>(name)
```
blsemo: Shouldn't that have a maximum size as well?

node/src/main/kotlin/net/corda/node/utilities/NodeNamedCache.kt
```diff
@@ -0,0 +1,54 @@
...
+    fun <K, V> buildNamed(caffeine: Caffeine<in K, in V>, name: String, loader: CacheLoader<K, V>): LoadingCache<K, V>
+}
+
+class DefaultNamedCacheFactory private constructor(private val metricRegistry: MetricRegistry?, private val nodeConfiguration: NodeConfiguration?) : NamedCacheFactory, SingletonSerializeAsToken() {
+    constructor() : this(null, null)
+
+    override fun bindWithMetrics(metricRegistry: MetricRegistry): NamedCacheFactory = DefaultNamedCacheFactory(metricRegistry, this.nodeConfiguration)
+    override fun bindWithConfig(nodeConfiguration: NodeConfiguration): NamedCacheFactory = DefaultNamedCacheFactory(this.metricRegistry, nodeConfiguration)
+
+    override fun <K, V> buildNamed(caffeine: Caffeine<in K, in V>, name: String): Cache<K, V> {
```
blsemo: I think it's a bit dodgy design wise that a `CacheFactory` actually takes a Caffeine builder as argument - if it was a real factory, it should make those things, and either take extra required args like a weigher as argument, or know from the name what to do there.
Admittedly, getting the abstraction right here is quite tricky, because we'd like to hide most of the nitty gritty cache stuff in the factory, but taking very specific things away from the calling code and putting it all in a magic factory makes understanding the code harder.

node/src/main/kotlin/net/corda/node/utilities/NodeNamedCache.kt
```diff
@@ -0,0 +1,54 @@
...
+        checkNotNull(metricRegistry)
+        checkNotNull(nodeConfiguration)
+        return caffeine.buildNamed<K, V>(name)
+    }
+
+    override fun <K, V> buildNamed(caffeine: Caffeine<in K, in V>, name: String, loader: CacheLoader<K, V>): LoadingCache<K, V> {
+        checkNotNull(metricRegistry)
+        checkNotNull(nodeConfiguration)
+        val configuredCaffeine = when (name) {
```
blsemo: Tbh, I think the factory should have a `buildNamedWeighted` or some such taking a weigher, than we could create the Caffeine in the factory, and it would be impossible that a misplaced call to the factory creates a cache without a weigher but a maximum weight...

node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -81,6 +83,9 @@ class P2PMessagingClient(val config: NodeConfiguration,
                          private val nodeExecutor: AffinityExecutor.ServiceAffinityExecutor,
                          private val database: CordaPersistence,
                          private val networkMap: NetworkMapCacheInternal,
+                         @Suppress("UNUSED")
+                         private val metricRegistry: MetricRegistry,
```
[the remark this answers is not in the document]
rick-r3: Yes.  My other branch is using it so I'd have to add it back straight away.

node/src/main/kotlin/net/corda/node/utilities/NodeNamedCache.kt
rick-r3: I gave up on a full abstraction (seemed like overkill) because the `Cache` interface is Caffeine specific, as are the weighers etc.  I just want a different factory in Enterprise, which this enables.

node/src/main/kotlin/net/corda/node/utilities/NodeNamedCache.kt
rick-r3: Caffeine does validation.  I don't want to duplicate the work of Caffeine.  I just want a point at which I can enhance with metrics and external configuration.

PR #3959, opened 2018-09-18
Config util write incorrect duration value when converting node config to typesafe config

patrickkuo:


node-api/src/test/kotlin/net/corda/nodeapi/internal/config/ConfigParsingTest.kt
```diff
@@ -254,6 +259,27 @@ class ConfigParsingTest {
         assertThat(objects.values).containsExactly(TestObject.Type1("type 1 value"), TestObject.Type2("type 2 value"))
     }
 
+    @Test
+    fun `parsing is reversible`() {
```
shamsasari: The `testPropertyType` should already be testing this

patrickkuo:
now this PR only change how we write duration config when using toConfig function, all reading are using typesafe's duration handling logic, so shouldn't have any compatibility issues.

shamsasari:
Can you confirm with tests

vkolomeyko:
It has been a while since this PR been raised.
Is it still relevant and needs to be merged?
Also, would it be possible to have a Jira item with description what the problem is?

patrickkuo:
We never write config to file and read it back, so this changes is probably not really needed, to avoid breaking anything I will close this PR for now

Issue #3961, opened 2018-09-18
Corda Network Builder fails if run on project with windows line endings


roastario:
When a project is checked out with windows line endings the network-builder fails to start.

PR #3962, opened 2018-09-18
make sure that the run-corda script is in unix line-ending format

roastario:
Fixes: https://github.com/corda/corda/issues/3961
Raised via: https://stackoverflow.com/questions/52209865/error-in-dockers-when-using-network-builder-to-start-up/52390080#52390080


tlil:
Is the SH file not checked in with UNIX line endings? You could add a `.gitattributes` entry like this?

```
*.sh eol=lf
```

tlil:
@roastario, how did you get on with the `.gitattributes` approach?

tlil (commented):
Make any progress on this? :)

PR #3968, opened 2018-09-18
CORDA-2001: added SNI header to TLS connections based on hashed CordaX500

bpaunescu:
Suffix used not awaiting feedback from other parties


node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/SSLHelper.kt
```diff
@@ -16,6 +18,8 @@ import java.security.cert.*
 import java.util.*
 import javax.net.ssl.*
 
+private const val HOSTNAME_FORMAT = "%s.corda.net"
```
gendal: I know it sounds like I'm constantly banging on but the use of something that *looks* like an FQDN and, in other similar contexts is intended to BE precisely that, makes me jittery.  Can we at least document prominently in the code the assumption we're making about why it's safe to use corda.net in this context? And why we don't consider it necessary to use a domain that exists for the sole purpose of avoiding ambiguity, etc?

vkolomeyko: I do agree with @gendal - explanation in code would be good.
There is some remark about RFC1035 below, I think it would be great to expand on that.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/SSLHelper.kt
```diff
@@ -152,3 +162,10 @@ internal fun initialiseTrustStoreAndEnableCrlChecking(trustStore: CertificateSto
 fun KeyManagerFactory.init(keyStore: CertificateStore) = init(keyStore.value.internal, keyStore.password.toCharArray())
 
 fun TrustManagerFactory.init(trustStore: CertificateStore) = init(trustStore.value.internal)
+
+internal fun x500toHostName(x500Name: CordaX500Name): String {
```
vkolomeyko: I know this code looks trivial and doesn't do a lot. But this will be a critical piece of code when it comes to translating X500 names to certificates.
I think it would be good to add a unit test or two to pin-point what it currently returns.

Also, the function is `internal` to NodeAPI, surely in addition to client code the same function will be used on server side to prepare a cache of certificates it is able to serve. That said it may be accessing this function from a different module. Perhaps it is a good idea to make function `public`? Its location within `internal` package should tell the story that it is not part of public API and reserved for internal use.

PR #3971, opened 2018-09-19
Corda-2006 serialization refactor

distributedleetravis:
A follow-up to [this PR](https://github.com/corda/corda/pull/3922) implementing [Corda-1391](https://r3-cev.atlassian.net/browse/CORDA-1391), this PR implements [Corda-2006](https://r3-cev.atlassian.net/browse/CORDA-2006) and is mainly focused on simplifying the checkpoint serialization classes and interfaces to eliminate features that they only had (or pretended to have) because they needed to maintain compatibility with the main serialization API.

Summary of changes:
* `SerializationEnvironmentImpl` is restructured to make it clearer which kinds of serialization are supported in a given environment.
* `CheckpointSerializerFactory` and `CheckpointSerializationScheme` are deleted and replaced with a single `CheckpointSerializer` interface.
* References to `CordaSerializationMagic` are eliminated, as the two possible values - `amqpMagic` and `kryoMagic` - are now implied by the classes you are using to do serialization with. `SerializationScheme` lookup in `SerializationFactoryImpl` is now by use-case only.
* The ability to set a different "current `SerializationFactory`", and record this choice temporarily in a `ThreadLocal`, is not used anywhere in our codebase, and has been removed and the associated methods deprecated. Note that this is different from the ability to set a thread-local `SerializationContext`, which is still needed for some purposes and has been retained.


rick-r3:
I agree with @fenryka.  The probability of serialization versioning is pretty high, if only for the case where a bug fix changes something in an incompatible way.  I'm mostly thinking for our own AMQP serialization, but it might even apply to checkpoint serialization if a Kryo upgrade changes things, for example.

PR #3974, opened 2018-09-19
CORDA-2009 update to BC 1.60 (security fixes)

kchalkias:
Major Fixes/optimisations due to update to BC 1.60
CVE-2018-1000180 (RSA key gen M-R tests) **-> this is critical**
EC key generation and signing now use cache-timing resistant table lookups.
EC point pre-computations could fail due to race conditions in concurrent settings. Point pre-computation was reworked to fix this.
Further work has been done on improving SHA-3 performance -> 4 times faster (required in the future for hash agility).
SPHINCS keys are now correctly created off certificates by the BC provider.

I've added extra keystore/cert tests mainly to check if RSA and Sphincs (affected by the new BC version) work as expected, but also to ensure key combination works as well.


kchalkias:
It seems there are not significant performance implications (i.e., due to side channel security improvements).

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -88,10 +120,14 @@ class X509UtilitiesTest {
 
     @Test
     fun `create valid server certificate chain`() {
-        val caKey = generateKeyPair(X509Utilities.DEFAULT_TLS_SIGNATURE_SCHEME)
+        schemeCombinationList.forEach { createValidServerCertChain(it.first, it.second) }
```
shamsasari: `forEach { (rootScheme, childScheme) -> ...}` for better clarity, or because you use this pattern several times, perhaps instead rename `schemeCombinationList` to something like `certChainSchemeCombinations`

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -129,9 +169,13 @@ class X509UtilitiesTest {
 
     @Test
     fun `storing EdDSA key in java keystore`() {
```
shamsasari: This test is no longer just about EdDSA

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -315,22 +331,54 @@ class X509UtilitiesTest {
     }
 
     @Test
-    fun `get correct private key type from Keystore`() {
-        val keyPair = generateKeyPair(Crypto.ECDSA_SECP256R1_SHA256)
+    fun `get correct private EC key type from Keystore`() {
+        val keyPair = generateKeyPair(ECDSA_SECP256R1_SHA256)
+        val (keyFromKeystore, keyFromKeystoreCasted) = storeAndGetKeysFromKeystore(keyPair)
+        assertTrue(keyFromKeystore is java.security.interfaces.ECPrivateKey) // by default JKS returns SUN EC key.
```
shamsasari: Use AssertJ: `assertThat(keyFromKeystore).isInstanceOf(ECPrivateKey::class.java)`. You'll get better information if the assertion fails.

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -315,22 +331,54 @@ class X509UtilitiesTest {
...
+        assertTrue(keyFromKeystoreCasted is org.bouncycastle.jce.interfaces.ECPrivateKey)
```
shamsasari: ditto

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -315,22 +331,54 @@ class X509UtilitiesTest {
...
+    @Test
+    fun `get correct private EdDSA key type from Keystore`() {
+        val keyPair = generateKeyPair(EDDSA_ED25519_SHA512)
+        val (keyFromKeystore, keyFromKeystoreCasted) = storeAndGetKeysFromKeystore(keyPair)
+        assertTrue(keyFromKeystore is EdDSAPrivateKey)
+        assertTrue(keyFromKeystoreCasted is EdDSAPrivateKey)
+    }
+
+    @Test
+    fun `get correct private Sphincs key type from Keystore`() {
```
shamsasari: This set of tests have the same structure. You could reduce the repetition by have a `List<Triple<String, PrivateKey, PrivateKey>>` global list, similar to `schemeCombinationList`.

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -357,12 +409,44 @@ class X509UtilitiesTest {
...
+        schemeCombinationList.forEach { signCertWithOtherKeyType(it.first, it.second) }
+    }
+
+    private fun signCertWithOtherKeyType(signatureSchemeRoot: SignatureScheme, signatureSchemeChild: SignatureScheme) {
+        val tmpKeyStore = tempFile("keystore.jks")
+        val rootKey = generateKeyPair(signatureSchemeRoot)
+        val testName = X500Principal("CN=Test,O=R3 Ltd,L=London,C=GB")
+        val rootCert = X509Utilities.createSelfSignedCACertificate(testName, rootKey)
+        val childKey = generateKeyPair(signatureSchemeChild)
+        val childCert = X509Utilities.createCertificate(CertificateType.TLS, rootCert, rootKey, BOB.name.x500Principal, childKey.public)
```
shamsasari: Is it possible to use `createValidServerCertChain` to reduce the duplication?

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -357,12 +409,44 @@ class X509UtilitiesTest {
...
+
+        // Save the EdDSA private key with cert chains.
+        val keyStore = loadOrCreateKeyStore(tmpKeyStore, "keystorepass")
+        keyStore.setKeyEntry("Key", childKey.private, "password".toCharArray(), arrayOf(rootCert, childCert))
+        keyStore.save(tmpKeyStore, "keystorepass")
+
+        // Load the keystore from file and make sure keys are intact.
+        val keyStore2 = loadOrCreateKeyStore(tmpKeyStore, "keystorepass")
```
shamsasari: `reloadedKeyStore`

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -357,12 +409,44 @@ class X509UtilitiesTest {
...
+        val privateKey = keyStore2.getKey("Key", "password".toCharArray())
```
shamsasari: `realoadedPrivateKey`. The variable names need changing, it's not clear which variable means what.

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
```diff
@@ -357,12 +409,44 @@ class X509UtilitiesTest {
...
         val actual: CertPath = serialized.deserialize(factory, context)
         assertEquals(expected, actual)
     }
+
+    @Test
+    fun `signing a key type with another key type certificate`() {
+        schemeCombinationList.forEach { signCertWithOtherKeyType(it.first, it.second) }
+    }
+
+    private fun signCertWithOtherKeyType(signatureSchemeRoot: SignatureScheme, signatureSchemeChild: SignatureScheme) {
```
shamsasari: The name here doesn't seem to reflect what the method is testing. It seems to be doing more.

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
kchalkias: @shamsasari I haven't really changed these tests, just made them more generic to work for any algorithm, but will do according to your recommendations.

node-api/src/test/kotlin/net/corda/nodeapi/internal/crypto/X509UtilitiesTest.kt
kchalkias: The main difference is for some key types, i.e., EC and RSA, default keystores unfortunately store them with java's default key types and we need to cast to BC types to use them (I wanted the comment to be well-visible for ECDSA and RSA). But yes, I will add this comment to the Triple generation.

kchalkias:
@shamsasari back to you

PR #3981, opened 2018-09-21
Corda 2006 serialization refactor

r3domfox:
A follow-up to [this PR](https://github.com/corda/corda/pull/3922) implementing [Corda-1391](https://r3-cev.atlassian.net/browse/CORDA-1391), this PR implements [Corda-2006](https://r3-cev.atlassian.net/browse/CORDA-2006) and is mainly focused on simplifying the checkpoint serialization classes and interfaces to eliminate features that they only had (or pretended to have) because they needed to maintain compatibility with the main serialization API.

Summary of changes:
* `CheckpointSerializerFactory` and `CheckpointSerializationScheme` are deleted and replaced with a single `CheckpointSerializer` interface.
* `SerializationEnvironmentImpl` is restructured to make it clearer which kinds of serialization are supported in a given environment.


fenryka:
As discussed, the reasons for some of te refactoring here have actually been undone underneath us, there isn't two factories living on the context object anymore with the simplication of the checkpoin steriliser... so lets close this and just mkae those changes

PR #3985, opened 2018-09-24
remove requirement to override default progress tracker for interacti…

roastario:


roastario:
Yup, will actually require a bit more logic. We should not use the UNSTARTED / DONE steps if the user has specified a custom progressChecker

PR #3989, opened 2018-09-25
Corda-1931 Asynchronous logging

blsemo:
Switch corda to use asynchronous logging (using a custom logging context selector to avoid usage of thread local storage as this causes issues with quasar).
Default behaviour is switched to async via log4j properties, documentation how to override on the command line or config has been provided.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/utilities/logging/AsyncLoggerContextSelectorNoThreadLocal.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/logging/AsyncLoggerContextSelectorNoThreadLocal.kt
@@ -10,7 +10,6 @@ import org.apache.logging.log4j.util.PropertiesUtil
 import java.net.URI
 
 class AsyncLoggerContextSelectorNoThreadLocal : ClassLoaderContextSelector() {
-
     companion object {
         /**
          * Returns `true` if the user specified this selector as the Log4jContextSelector, to make all loggers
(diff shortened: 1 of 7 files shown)


node/src/main/kotlin/net/corda/node/utilities/logging/AsyncLoggerContextSelectorNoThreadLocal.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.core.LoggerContext
+import org.apache.logging.log4j.core.async.AsyncLoggerContext
+import org.apache.logging.log4j.core.selector.ClassLoaderContextSelector
+import org.apache.logging.log4j.core.util.Constants
+import org.apache.logging.log4j.util.PropertiesUtil
+import java.net.URI
+
+class AsyncLoggerContextSelectorNoThreadLocal : ClassLoaderContextSelector() {
+
```
rick-r3: remove blank line

blsemo:
I don't think further changes for Enterprise are required - that's it to turn on async logging by default.

rick-r3:
I think we need to change some defaults / system properties then.  The ring buffer defaults to 4KB unless you set the property to tell Log4J to not use thread locals globally or another system property for configuring the ring buffer size.  Maybe we should set the ring buffer size in `log4j2.component.properties` ?  Their own code sets it to 256KB if the other property is set, so maybe go with that?

rick-r3:
Also, do you want to increase the log file size from 10MB ?

PR #3993, opened 2018-09-26
CORDA-1726 Make sure test frameworks and nodes refuse to start on any…

szymonsztuka:
…thing except the right Java versions

Move isValidJavaVersion(), hasMinimumJavaVersion() to Node as it can be use for check in DriverDSL for starting in-process node,
in-process node doesn't go via NodeStartup class where the correct Java version is checked.

Added safeguard to hasMinimumJavaVersion as some JDKs distributions may not have the update version number (e.g. AdoptOpenJDK) which causes unintended parse exception.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/Node.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/Node.kt
@@ -156,7 +156,7 @@ open class Node(configuration: NodeConfiguration,
                 val update = SystemUtils.JAVA_VERSION.substringAfter("_").toLong()
                 major == 1.8F && update >= 171
             } catch (e: NumberFormatException) { // custom JDKs may not have the update version (e.g. 1.8.0-adoptopenjdk)
-                return false
+                false
             }
         }
     }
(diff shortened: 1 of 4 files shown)


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -138,6 +139,26 @@ open class Node(configuration: NodeConfiguration,
...
+        }
+
+        private fun hasMinimumJavaVersion(): Boolean {
+            // when the ext.java8_minUpdateVersion gradle constant changes, so must this check
+            val major = SystemUtils.JAVA_VERSION_FLOAT
+            return try {
+                val update = SystemUtils.JAVA_VERSION.substringAfter("_").toLong()
+                major == 1.8F && update >= 171
+            } catch (e: NumberFormatException) { // custom JDKs may not have the update version (e.g. 1.8.0-adoptopenjdk)
+                return false
```
shamsasari: You don't need the "return" keyword here

PR #3997, opened 2018-09-27
CORDA-2031 put "AlgorithmParameters.SHA256WITHECDSA" to BC

kchalkias:
There was an issue reported after updating to BC 1.60 related to `X509CRL.verify()`. Added a new algorithm params entry to BC to resolve this.


core/src/main/kotlin/net/corda/core/crypto/internal/ProviderMap.kt
```diff
@@ -30,6 +30,8 @@ internal val cordaBouncyCastleProvider = BouncyCastleProvider().apply {
         override fun generatePublic(keyInfo: SubjectPublicKeyInfo) = decodePublicKey(EDDSA_ED25519_SHA512, keyInfo.encoded)
         override fun generatePrivate(keyInfo: PrivateKeyInfo) = decodePrivateKey(EDDSA_ED25519_SHA512, keyInfo.encoded)
     })
+    // Required due to [X509CRL].verify() reported issues in network-services after BC 1.60 update.
+    put("AlgorithmParameters.SHA256WITHECDSA", "org.bouncycastle.jcajce.provider.asymmetric.ec.AlgorithmParametersSpi")
```
patrickkuo: better use `AlgorithmParametersSpi::class.java.name` instead of string? incase bc decided to move this class in future release.

kchalkias: back to you @patrickkuo

PR #4001, opened 2018-09-28
ENT-2500: Introduce optional BridgeMetricsService in `nodeApi` module

vkolomeyko:
This is a split-out from the work done on the Enterprise Jira.

Apparently, all the outbound connections are handled by the Bridge which is part of NodeAPI module in Corda OS.
Therefore it makes sense to introduce an optional `BridgeMetricsService` here such that it will be naturally integrated into Enterprise repo and reduces merging conflicts in the future.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/bridging/BridgeAuditService.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/bridging/BridgeAuditService.kt
@@ -1,9 +1,0 @@
-package net.corda.nodeapi.internal.bridging
-
-import net.corda.nodeapi.internal.protonwrapper.messages.SendableMessage
-import org.apache.activemq.artemis.api.core.client.ClientMessage
-
-interface BridgeAuditService {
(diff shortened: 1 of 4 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/bridging/BridgeAuditService.kt
```diff
@@ -0,0 +1,9 @@
+package net.corda.nodeapi.internal.bridging
+
+import net.corda.nodeapi.internal.protonwrapper.messages.SendableMessage
+import org.apache.activemq.artemis.api.core.client.ClientMessage
+
+interface BridgeAuditService {
```
mnesbit: Should we add a few other traps? Not sure Audit is right term, perhaps Metrics?

vkolomeyko: Sure, will add/change.

PR #4007, opened 2018-10-01
Updates tutorial to match template format.

joeldudleyr3:


docs/source/example-code/src/main/kotlin/net/corda/docs/kotlin/tutorial/twoparty/IOUContract.kt
```diff
@@ -9,9 +9,11 @@ import net.corda.core.transactions.LedgerTransaction
 import net.corda.core.contracts.*
 
 // Replace IOUContract's contract ID and definition with:
-const val IOU_CONTRACT_ID = "com.template.IOUContract"
-
 class IOUContract : Contract {
+    companion object {
+        val ID = "com.template.IOUContract"
```
shamsasari: `const`

joeldudleyr3:
I don't think I do? The Java side matches the template.

PR #4008, opened 2018-10-01
CORDA-1915 Update to Network Bootstrapper for signed JARs

szymonsztuka:
Network Bootstrapper doesn't adds contracts classes from signed CorDapp JARs to whitelist.
If contracts from signed JARs need to be added, they can be specified in `include_whitelist.txt` file.

The cordapp and cordformation plugins in the incoming cord-gradle-plugins will have ability to sign JARs (in cordformation signing will be by default), to enable signature constraints to work out of box Network Bootstrapper will not  whitelist contracts form signed JARs.
For unsigned JARs the Network Bootstrapper behaviour is unchanged.

### DIFF
--- a/docs/source/network-bootstrapper.rst
+++ b/docs/source/network-bootstrapper.rst
@@ -87,9 +87,10 @@ Any CorDapps provided when bootstrapping a network will be scanned for contracts
 The CorDapp JARs will be hashed and scanned for ``Contract`` classes. These contract class implementations will become part
 of the whitelisted contracts in the network parameters (see ``NetworkParameters.whitelistedContractImplementations`` :doc:`network-map`).
 
-By default the bootstrapper will whitelist all the contracts found in all the CorDapp JARs. To prevent certain
-contracts from being whitelisted, add their fully qualified class name in the ``exclude_whitelist.txt``. These will instead
+By default the bootstrapper will whitelist all the contracts found in all the unsiged CorDapp JARs (a JAR file not signed bu jarSinger tool).
+To prevent certain contracts from being whitelisted, add their fully qualified class name in the ``exclude_whitelist.txt``. These will instead
 use the more restrictive ``HashAttachmentConstraint``.
+To add certain contracts to whitelisting, add their fully qualified class name in the ``include_whitelist.txt``.
 
 For example:
 
--- a/docs/source/network-bootstrapper.rst
+++ b/docs/source/network-bootstrapper.rst
@@ -87,7 +87,7 @@ Any CorDapps provided when bootstrapping a network will be scanned for contracts
 The CorDapp JARs will be hashed and scanned for ``Contract`` classes. These contract class implementations will become part
 of the whitelisted contracts in the network parameters (see ``NetworkParameters.whitelistedContractImplementations`` :doc:`network-map`).
 
-By default the bootstrapper will whitelist all the contracts found in all the unsiged CorDapp JARs (a JAR file not signed bu jarSinger tool).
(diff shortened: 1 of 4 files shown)


docs/source/network-bootstrapper.rst
```diff
@@ -87,9 +87,10 @@ Any CorDapps provided when bootstrapping a network will be scanned for contracts
 The CorDapp JARs will be hashed and scanned for ``Contract`` classes. These contract class implementations will become part
 of the whitelisted contracts in the network parameters (see ``NetworkParameters.whitelistedContractImplementations`` :doc:`network-map`).
 
-By default the bootstrapper will whitelist all the contracts found in all the CorDapp JARs. To prevent certain
-contracts from being whitelisted, add their fully qualified class name in the ``exclude_whitelist.txt``. These will instead
+By default the bootstrapper will whitelist all the contracts found in all the unsigned CorDapp JARs (a JAR file not signed by jarSinger tool).
```
shamsasari: Typo: jarSinger

shamsasari: Can you elaborate more here. This is about signature constraints so perhaps at least mention that and link to the relevant docsite page.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -398,4 +400,9 @@ internal constructor(private val initSerEnv: Boolean,
             return magic == amqpMagic && target == SerializationContext.UseCase.P2P
         }
     }
+
+    private fun isSigned(file: Path): Boolean =
+            JarInputStream(Files.newInputStream(file)).use {
```
shamsasari: Use `net.corda.core.internal.PathUtilsKt#read` extension method.

docs/source/network-bootstrapper.rst
```diff
@@ -87,9 +87,10 @@ Any CorDapps provided when bootstrapping a network will be scanned for contracts
...
+To prevent certain contracts from being whitelisted, add their fully qualified class name in the ``exclude_whitelist.txt``. These will instead
 use the more restrictive ``HashAttachmentConstraint``.
+To add certain contracts to whitelisting, add their fully qualified class name in the ``include_whitelist.txt``.
```
shamsasari: I think you need to re-state why you might need this file.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/WhitelistGenerator.kt
```diff
@@ -40,4 +45,9 @@ fun generateWhitelist(networkParameters: NetworkParameters?,
 fun readExcludeWhitelist(directory: Path): List<String> {
     val file = directory / EXCLUDE_WHITELIST_FILE_NAME
     return if (file.exists()) file.readAllLines().map(String::trim) else emptyList()
+}
+
+fun readIncludeWhitelist(directory: Path): List<String> {
```
shamsasari: Can you merge `readExcludeWhitelist` and `readExcludeWhitelist` so that it takes in an extra filename parameter to remove the code duplication.

szymonsztuka:
I've dropped include_whitelist.txt file (which I wanted to add) as whenever signature constraints can be used (a contract is from signed JAR), this is preferable over Zone Constraint (a whitelisted contract), so there's no need to force adding to whitelist.
As before for unsigned JARs network bootstrapper works as before (including exclude_whitelist.txt)

PR #4011, opened 2018-10-02
[CORDA-1926] Make CordappInfoResolver.withCordappInfoResolution @VisibleForTesting.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1926


core/src/main/kotlin/net/corda/core/internal/cordapp/CordappInfoResolver.kt
```diff
@@ -49,6 +50,7 @@ object CordappInfoResolver {
      * Temporarily switch out the internal resolver for another one. For use in testing.
      */
     @Synchronized
+    @VisibleForTesting
```
shamsasari: And the `clear` method below as well.

PR #4013, opened 2018-10-02
Added more docs for reference states to key concepts and api docs.

roger3cev:
Nothing special to add.

### DIFF
--- a/docs/source/api-states.rst
+++ b/docs/source/api-states.rst
@@ -152,3 +152,41 @@ Where:
 * ``encumbrance`` points to another state that must also appear as an input to any transaction consuming this
   state
 * ``constraint`` is a constraint on which contract-code attachments can be used with this state
+
+Reference States
+----------------
+
+A reference input state is a ``ContractState`` which can be referred to in a transaction by the contracts of input and
+output states but whose contract is not executed as part of the transaction verification process. Furthermore,
+reference states are not consumed when the transaction is committed to the ledger but they are checked for
+"current-ness". In other words, the contract logic isn't run for the referencing transaction only. It's still a normal
+state when it occurs in an input or output position.
+
+Reference data states enable many parties to "reuse" the same state in their transactions as reference data whilst
+still allowing the reference data state owner the capability to update the state. A standard example would be the
+creation of financial instrument reference data and the use of such reference data by parties holding the related
+financial instruments.
+
+Just like regular input states, the chain of provenance for reference states is resolved and all dependency transactions
+verified. This is because users of reference data must be satisfied that the data they are referring to is valid as per
+the rules of the contract which governs it and that all previous participants of teh state assented to updates of it.
+
+**Known limitations:**
+
+*Notary change:* It is likely the case that users of reference states do not have permission to change the notary
+assigned to a reference state. Even if users *did* have this permission the result would likely be a bunch of
+notary change races. As such, if a reference state is added to a transaction which is assigned to a
+different notary to the input and output states then all those inputs and outputs must be moved to the
+notary which the reference state uses.
+
+If two or more reference states assigned to different notaries are added to a transaction then it follows
+that this transaction likely *cannot* be committed to the ledger as it unlikely that the party using the
+reference state can change the assigned notary for one of the reference states.
+
+As such, if reference states assigned to multiple different notaries are added to a transaction builder
+then the check below will fail.
+
+*Encumbrances:* In the near future, additional logic must be added to handle references of encumbered states. In the
+case where a state is encumbered by an encumbrance state. The encumbrance state should also be referenced in the same
+transaction that references the encumbered state. This is because the data contained within the encumbered state may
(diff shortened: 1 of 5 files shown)


docs/source/api-states.rst
```diff
@@ -151,4 +151,42 @@ Where:
...
+Reference States
+----------------
+
+A reference input state is a ``ContractState`` which can be referred to in a transaction by the contracts of input and
+output states but whose contract is not executed as part of the transaction verification process. Furthermore,
+reference states are not consumed when the transaction is committed to the ledger but they are checked for
+"current-ness". In other words, the contract logic isn't run for the referencing transaction only. It's still a normal
+state when it occurs in an input or output position.
+
+Reference data states enable many parties to "reuse" the same state in their transactions as reference data whilst
```
mikehearn: I guess reuse doesn't need quotes - it's totally correct.

docs/source/api-states.rst
```diff
@@ -151,4 +151,42 @@ Where:
...
+
+**Known limitations:**
+
+*Notary change:* It is likely the case that users of reference states do not have permission to change the notary
+assigned to a reference state. Even if users *did* have this permission the result would likely be a bunch of
+notary change races. As such, if a reference state is added to a transaction which is assigned to a
+different notary to the input and output states then all those inputs and outputs must be moved to the
+notary which the reference state uses.
+
+If two or more reference states assigned to different notaries are added to a transaction then it follows
```
mikehearn: This paragraph applies already - the system won't automatically move states between notaries for you. Probably it can be taken out.

docs/source/api-states.rst
```diff
@@ -151,4 +151,42 @@ Where:
...
+that this transaction likely *cannot* be committed to the ledger as it unlikely that the party using the
+reference state can change the assigned notary for one of the reference states.
+
+As such, if reference states assigned to multiple different notaries are added to a transaction builder
+then the check below will fail.
+
+*Encumbrances:* In the near future, additional logic must be added to handle references of encumbered states. In the
```
mikehearn: Maybe this should be a `.. warning::` and phrased as "encumbrances should not be combined with reference states" rather than like this, which is almost a todo note to ourselves.

docs/source/api-transactions.rst
```diff
@@ -123,20 +114,32 @@ obtained from a ``StateAndRef`` by calling the ``StateAndRef.referenced()`` meth
...
-that this transaction likely *cannot* be committed to the ledger as it unlikely that the party using the
-reference state can change the assigned notary for one of the reference states.
+At the point where the transaction updating the state being used as a reference is committed to storage and the vault
+update occurs, then the ``WithReferencedStatesFlow`` will wake up and re-execute the provided flow.
 
-As such, if reference states assigned to multiple different notaries are added to a transaction builder
-then the check below will fail.
+.. warning:: Caution should be taken when using this flow as it facilitates automated re-running of flows which use
+reference states. The flow using reference states should include checks to ensure that the reference data is
+   reasonable, especially if some economics transaction depends upon it.
```
mikehearn: especially if _the economics of the_ transaction depends upon it.

Current phrasing is a bit odd.

docs/source/index.rst
```diff
@@ -79,6 +79,7 @@ We look forward to seeing what you can do with Corda!
    design/kafka-notary/design.md
    design/monitoring-management/design.md
    design/sgx-integration/design.md
+      design/reference-states/design.md
```
mikehearn: Indent error.

docs/source/key-concepts-transactions.rst
```diff
@@ -33,7 +33,7 @@ Here is an example of an update transaction, with two inputs and two outputs:
    :scale: 25%
    :align: center
 
-A transaction can contain any number of inputs and outputs of any type:
+   A transaction can contain any number of inputs, outputs and references of any type:
```
mikehearn: Indent error

docs/source/api-states.rst
roger3cev: Ah what i meant here was that this cannot ever be committed to the ledger, unless the parties which "own" the reference states change the notary.

docs/source/api-states.rst
roger3cev: Makes sense. I should add this as a JIRA for future work.

docs/source/api-states.rst
roger3cev: amended

docs/source/index.rst
roger3cev: amended

docs/source/key-concepts-transactions.rst
roger3cev: amended

docs/source/api-transactions.rst
roger3cev: amended

roger3cev:
Over to you @mikehearn

PR #4020, opened 2018-10-03
[CORDA-2063] Ensure signatures and BC operations always use newSecureRandom

kchalkias:
Note that this change mainly impacts nodes where ECDSA keys are used to sign transactions. Also, by setting the default BC PRNG to newSecureRandom() we ensure that other BC PRNG requests that might be invoked in the background and we don't know (i.e., cert signing might use `CryptoServicesRegistrar`) are non-blocking anymore. 
This change caused a domino effect on disallowing all signature operations on DJVM, but because we want to check sig. verification in DJVM, part of the signing code has been ported to `deterministic.crypto`.

### DIFF
--- a/core-deterministic/testing/src/test/kotlin/net/corda/deterministic/crypto/CryptoSignUtils.kt
+++ b/core-deterministic/testing/src/test/kotlin/net/corda/deterministic/crypto/CryptoSignUtils.kt
@@ -1,0 +1,78 @@
+@file:JvmName("CryptoSignUtils")
+
+package net.corda.deterministic.crypto
+
+import net.corda.core.crypto.*
+import net.corda.core.crypto.Crypto.SPHINCS256_SHA256
+import net.corda.core.crypto.Crypto.findSignatureScheme
+import net.corda.core.crypto.Crypto.isSupportedSignatureScheme
+import net.corda.core.serialization.serialize
+import java.security.*
+
+/**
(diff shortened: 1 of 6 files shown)


tlil:
@kchalkias  3 unit tests are failing.... ?

kchalkias:
Hm, it seems this is BC issue and I will raise a ticket to BC's github.
```
protected void engineInitSign(PrivateKey privateKey)
        throws InvalidKeyException
    {
        if (privateKey instanceof BCSphincs256PrivateKey)
        {
            CipherParameters param = ((BCSphincs256PrivateKey)privateKey).getKeyParams();
            // HERE IS THE PROBLEM, SPHINCS is deterministic by its nature, 
            // they should just omit the PRNG, similarly to what we do for EdDSA.
            if (random != null)
            {
                param = new ParametersWithRandom(param, random); 
            }
......
```
```
public void init(boolean var1, CipherParameters var2) {
        if (var1) {
            this.keyData = ((SPHINCSPrivateKeyParameters)var2).getKeyData(); // This will FAIL.
        } else {
            this.keyData = ((SPHINCSPublicKeyParameters)var2).getKeyData();
        }
    }
```
So, until this is resolved on BC, I will add an if statement to handle Sphincs differently (without a PRNG when signing).

kchalkias:
@tlil here is the ticket to BC's github https://github.com/bcgit/bc-java/issues/415

kchalkias:
back to you @tlil

PR #4026, opened 2018-10-03
Moved the PLATFORM_VERSION constant to core and added some missing usages

shamsasari:


constants.properties
```diff
@@ -1,7 +1,9 @@
 gradlePluginsVersion=4.0.29
 kotlinVersion=1.2.51
-# When adjusting platformVersion upwards please also modify CordaRPCClientConfiguration.minimumServerProtocolVersion \
-# if there have been any RPC changes. Also please modify InternalMockNetwork.kt:MOCK_VERSION_INFO and NodeBasedTest.startNode
+# ***************************************************************#
+# When incrementing platformVersion make sure to update          #
+# net.corda.core.internal.CordaUtilsKt.PLATFORM_VERSION as well. #
+# ***************************************************************#
```
josecoll: Is there no way of easily reading this value from the code ?

shamsasari: Yeah I been thinking of a way to remove the duplication. Something for another PR I think.

shamsasari:
>Is this likely to cause binary compatibility problems with pre-existing Corda versions using PLATFORM_VERSION ?

No, because this has only just been introduced and not in V3.*. Further, this is internal API which we don't provide any guarantees for.

josecoll:
Cool - the issue I fixed was our bad in the 1st place for using an internal package from the `corda-finance` app in V3.0/V3.1.

constants.properties
josecoll: Fair enough

josecoll: Although interestingly you have actually removed a piece of code that does exactly that:

```
        val expectedPlatformVersion = (ProjectStructure.projectRootDir / "constants.properties").read {
            val constants = Properties()
            constants.load(it)
            constants.getProperty("platformVersion").toInt()
        }
```

josecoll: Can we not use the above in the definition of PLATFORM constant?

shamsasari: No, that only works in tests where the git repo is checked out (`ProjectStructure.projectRootDir` searches for the `.git` dir)

shamsasari: And incidentally, I've removed that code to use `PLATFORM_VERSION` as well.

PR #4035, opened 2018-10-05
ENT-1906: Modify the DJVM to wrap Java primitive types.

chrisr3:
Rework how we load classes into the `sandbox.*` namespace:
- Create new "boxed" classes for Java primitives and `String` inside the sandbox. Ensure that all of these classes belong to the `SandboxClassLoader`.
- The only `sandbox.*` classes that do _not_ belong to the `SandboxClassLoader` are now `RuntimeCostAccounter`, `RuleViolationError` and `ThresholdViolationError`, because these are also needed outside the sandbox.
- Refactor the `SandboxClassLoader` so that it _always_ provides the class that it has been asked to load. The `SandboxExecutor` now adds `sandbox.` to its target class name, and then requests an instance of that. This allows `SandboxClassLoader` still to load instances of (e.g.) `java.lang.Throwable` when necessary.


djvm/src/test/kotlin/net/corda/djvm/costing/RuntimeCostTest.kt
```diff
@@ -16,15 +17,14 @@ class RuntimeCostTest {
...
     fun `cannot increment cost beyond threshold`() {
-        Thread {
+        thread {
             val cost = RuntimeCost(10) { "failed in ${it.name}" }
             assertThatExceptionOfType(ThresholdViolationError::class.java)
                     .isThrownBy { cost.increment(11) }
                     .withMessage("failed in Foo")
             assertThat(cost.value).isEqualTo(11)
         }.apply {
             name = "Foo"
```
exFalso: If you use `thread` then the name should be set in the function call i.e. `thread(name = ..)`

djvm/src/main/java/sandbox/java/lang/Object.java
```diff
@@ -0,0 +1,71 @@
...
+        for (java.lang.Object arg : args) {
+            unwrapped[i] = unwrap(arg);
+            ++i;
+        }
+        return unwrapped;
+    }
+
+    private static java.lang.Object unwrap(java.lang.Object arg) {
+        if (arg instanceof Object) {
+            return ((Object) arg).fromDJVM();
```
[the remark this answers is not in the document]
chrisr3: :+1:

djvm/src/main/kotlin/net/corda/djvm/code/instructions/DropKotlinMetadata.kt
```diff
@@ -0,0 +1,4 @@
+package net.corda.djvm.code.instructions
+
+class DropKotlinMetadata {
```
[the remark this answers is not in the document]
chrisr3: Nope. See above... :point_up:

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/ArgumentUnwrapper.kt
```diff
@@ -0,0 +1,35 @@
...
+class ArgumentUnwrapper : Emitter {
+    override fun emit(context: EmitterContext, instruction: Instruction) = context.emit {
+        if (instruction is MemberAccessInstruction && context.whitelist.matches(instruction.owner)) {
+            fun unwrapString() = invokeStatic("sandbox/java/lang/String", "fromDJVM", "(Lsandbox/java/lang/String;)Ljava/lang/String;")
+
+            if (hasStringArgument(instruction)) {
+                unwrapString()
+            } else if (instruction.owner == "java/lang/Class" && instruction.signature.startsWith("(Ljava/lang/String;ZLjava/lang/ClassLoader;)")) {
+                /**
+                 * [kotlin.jvm.internal.Intrinsics.checkHasClass] invokes [Class.forName], so I'm
```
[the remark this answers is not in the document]
chrisr3: The parent of `SandboxClassLoader` is the DJVM's application-level classloader. And we need that, because it contains the DJVM's own code as well as the "pinned" classes.

djvm/src/main/kotlin/sandbox/java/lang/DJVM.kt
```diff
@@ -0,0 +1,145 @@
...
+    return if (startsWith(SANDBOX_PREFIX)) {
+        this
+    } else {
+        SANDBOX_PREFIX + this
+    }
+}
+
+private fun kotlin.String.fromSandboxPackage(): kotlin.String {
+    return if (startsWith(SANDBOX_PREFIX)) {
+        substring(8)
```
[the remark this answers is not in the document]
chrisr3: Yes, that's a tweak I've made since. (Amongst others... :wink:)

djvm/src/test/kotlin/net/corda/djvm/costing/RuntimeCostTest.kt
chrisr3: This _is_ only a unit test ... :wink:

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/ArgumentUnwrapper.kt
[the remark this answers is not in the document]
chrisr3: Ah, I see what you mean. Unfortuately, Kotin calls `Class.forName()`:
```java
package kotlin.jvm.internal;

public class Intrinsics {
    public static void checkHasClass(String internalName) throws ClassNotFoundException;
    public static void checkHasClass(String internalName, String requiredVersion) throws ClassNotFoundException
}
```

djvm/src/main/java/sandbox/java/lang/System.java
```diff
@@ -0,0 +1,28 @@
...
+    /*
+     * This class is duplicated into every sandbox, where everything is single-threaded.
+     */
+    private static final java.util.Map<java.lang.Integer, java.lang.Integer> objectHashCodes = new java.util.LinkedHashMap<>();
+    private static int objectCounter = 0;
+
+    public static int identityHashCode(java.lang.Object obj) {
+        int nativeHashCode = java.lang.System.identityHashCode(obj);
+        // TODO Instead of using a magic offset below, one could take in a per-context seed
+        return objectHashCodes.computeIfAbsent(nativeHashCode, i -> ++objectCounter + 0xfed_c0de);
```
tlil: Worth using `ConcurrentHashMap` instead of `LinkedHashMap` ?

djvm/src/main/kotlin/net/corda/djvm/analysis/ClassAndMemberVisitor.kt
```diff
@@ -86,7 +86,7 @@ open class ClassAndMemberVisitor(
     /**
      * Process class after it has been fully traversed and analyzed.
      */
-    open fun visitClassEnd(clazz: ClassRepresentation) {}
+    open fun visitClassEnd(classVisitor: ClassVisitor, clazz: ClassRepresentation) {}
```
tlil: Worth adding a short comment to the Kdoc about what state the class visitor is in when this method is being called?

djvm/src/main/kotlin/net/corda/djvm/code/ClassMutator.kt
```diff
@@ -2,26 +2,48 @@ package net.corda.djvm.code
...
+import org.objectweb.asm.Opcodes.*
 
 /**
  * Helper class for applying a set of definition providers and emitters to a class or set of classes.
  *
  * @param classVisitor Class visitor to use when traversing the structure of classes.
- * @property definitionProviders A set of providers used to update the name or meta-data of classes and members.
+ * @property configuration The configuration to use for class analysis.
+ * @param definitionProviders A set of providers used to update the name or meta-data of classes and members.
  * @property emitters A set of code emitters used to modify and instrument method bodies.
```
tlil: `@property definitionProviders`
`@param emitters`

djvm/src/main/kotlin/net/corda/djvm/code/ClassMutator.kt
```diff
@@ -2,26 +2,48 @@ package net.corda.djvm.code
...
+            if (instruction is MethodEntry
+                    && instruction.method.memberName == "<clinit>" && instruction.method.signature == "()V"
+                    && initializers.isNotEmpty()) {
+                writeByteCode(initializers)
+                initializers.clear()
+            }
+        }
+    }
+
+    private val emitters: List<Emitter> = emitters + PrependClassInitializer()
```
tlil: Do you want to flip the order around in case future emitters would want to override or extend the behaviour of `PrependClassInitializer`? (Minor point as you're doing the `initializers` business below)

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
```diff
@@ -69,69 +75,68 @@ class SandboxClassLoader(
...
 
         // Check if the class has already been loaded.
-        val loadedClass = loadedClasses[name]
+        val loadedClass = loadedClasses[requestedPath]
         if (loadedClass != null) {
-            logger.trace("Class {} already loaded", source.qualifiedClassName)
+            logger.trace("Class {} already loaded", request.qualifiedClassName)
             return loadedClass
+        } else if (analysisConfiguration.isPinnedClass(requestedPath)) {
+            logger.debug("Class {} is loaded unmodified", request.qualifiedClassName)
```
tlil: #nitpick, but is the mix between `trace`- and `debug`-level logging here intentional?

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassWriter.kt
```diff
@@ -8,13 +8,13 @@ import org.objectweb.asm.ClassWriter.COMPUTE_MAXS
...
 /**
- * Class writer for sandbox execution, with configurable a [classLoader] to ensure correct deduction of the used class
+ * Class writer for sandbox execution, with a configurable classloader to ensure correct deduction of the used class
  * hierarchy.
  *
  * @param classReader The [ClassReader] used to read the original class. It will be used to copy the entire constant
  * pool and bootstrap methods from the original class and also to copy other fragments of original byte code where
  * applicable.
- * @property classLoader The class loader used to load the classes that are to be rewritten.
+ * @param cloader The class loader used to load the classes that are to be rewritten.
```
tlil: Still a (private) property below #nit

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/ArgumentUnwrapper.kt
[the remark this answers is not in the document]
tlil: :+1:

djvm/src/main/kotlin/sandbox/java/lang/DJVM.kt
```diff
@@ -0,0 +1,152 @@
...
+    if (bannedClasses.any { it.matches(className) }) {
+        throw ClassNotFoundException(className)
+    }
+    return SANDBOX_PREFIX + className
+}
+
+private val bannedClasses = setOf(
+    "^java\\.lang\\.DJVM(.*)?\$".toRegex(),
+    "^net\\.corda\\.djvm\\..*\$".toRegex(),
+    "^Task\$".toRegex()
```
tlil: `"^sandbox.Task\$"`?

djvm/src/test/kotlin/net/corda/djvm/costing/RuntimeCostTest.kt
```diff
@@ -16,15 +17,13 @@ class RuntimeCostTest {
...
     @Test
     fun `cannot increment cost beyond threshold`() {
-        Thread {
+        thread(name = "Foo") {
             val cost = RuntimeCost(10) { "failed in ${it.name}" }
             assertThatExceptionOfType(ThresholdViolationError::class.java)
                     .isThrownBy { cost.increment(11) }
                     .withMessage("failed in Foo")
             assertThat(cost.value).isEqualTo(11)
         }.apply {
```
tlil: `}.join()` ? :)

djvm/src/main/java/sandbox/java/lang/System.java
chrisr3: No, because the sandbox is single-threaded. I think I may also have deleted `ConcurrentHashMap` from `deterministic-rt.jat` ... :wink:

djvm/src/main/kotlin/net/corda/djvm/code/ClassMutator.kt
chrisr3: :+1:

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
chrisr3: Yes. I found `trace` logging to be too verbose and so promoted the logging that I found helpful when debugging to `debug`.

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassWriter.kt
chrisr3: Strictly speaking, Kotlin still considers `classLoader` to be a function. The following does not compile:
```kotlin
override val classLoader: ClassLoader get() = cloader
```

djvm/src/main/kotlin/sandbox/java/lang/DJVM.kt
chrisr3: The DJVM _always_ prepends `sandbox.*`, so a user would need to write:
```java
Class.forName("Task");
```
in order to attempt to load `sandbox.Task`. Writing
```
Class.forName("sandbox.Task");
```
would result in the DJVM trying to load `sandbox.sandbox.Task`, which would obviously fail anyway.

djvm/src/main/kotlin/net/corda/djvm/analysis/ClassAndMemberVisitor.kt
chrisr3: :+1:

djvm/src/main/kotlin/net/corda/djvm/code/ClassMutator.kt
chrisr3: Does it make any difference? We sort the emitters anyway so that the Trace ones are executed first:
```kotlin
// We need to apply the tracing emitters before the non-tracing ones.
Processor.processEntriesOfType<Emitter>(emitters.sortedByDescending(Emitter::isTracer), analysisContext.messages) {
    it.emit(context, instruction)
}
```

chrisr3: FYI, a future change will replace `isTracer` by a `priority` property so that I can order emitters more precisely.

djvm/src/main/java/sandbox/java/lang/System.java
tlil: Duh 😂

djvm/src/main/kotlin/net/corda/djvm/code/ClassMutator.kt
tlil: Kk, that should suffice. Thanks

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
tlil: Okay :)

djvm/src/main/kotlin/sandbox/java/lang/DJVM.kt
tlil: Ah, okay, I see what you mean. That makes sense. Perhaps a short comment on top here to make that 100% clear?

PR #4039, opened 2018-10-05
Assorted set of clean ups

shamsasari:


shamsasari:
These are cleanups I've made along the way in my current work.

PR #4041, opened 2018-10-05
CORDA-1915 node rejects CorDapps signed by our dev keys in prod mode

szymonsztuka:
Related to Signing CorDapp JARs - Corda node rejects CorDapps signed by our development keys when running in production mode - this is to prevent Cordapps signed by our dev key (will be by default) running  in production (node devMode=false).

Full context of this PR described in https://r3-cev.atlassian.net/browse/CORDA-1915


distributedleetravis:
Other than my query above, this all LGTM.

PR #4042, opened 2018-10-08
CORDA-2006: Simplify checkpoint serialization

distributedleetravis:
A follow-up to [this PR](https://github.com/corda/corda/pull/3922) implementing [Corda-1391](https://r3-cev.atlassian.net/browse/CORDA-1391), this PR implements [Corda-2006](https://r3-cev.atlassian.net/browse/CORDA-2006) and is mainly focused on simplifying the checkpoint serialization classes and interfaces to eliminate features that they only had (or pretended to have) because they needed to maintain compatibility with the main serialization API. Accordingly, `CheckpointSerializerFactory` and `CheckpointSerializationScheme` are deleted and replaced with a single `CheckpointSerializer` interface.


core-deterministic/testing/verifier/src/main/kotlin/net/corda/deterministic/verifier/LocalSerializationRule.kt
```diff
@@ -58,13 +58,11 @@ class LocalSerializationRule(private val label: String) : TestRule {
...
     }
 
-    private fun createTestSerializationEnv(): SerializationEnvironmentImpl {
+    private fun createTestSerializationEnv(): SerializationEnvironment {
         val factory = SerializationFactoryImpl(mutableMapOf()).apply {
             registerScheme(AMQPSerializationScheme(emptySet(), AccessOrderLinkedHashMap(128)))
         }
-        return object : SerializationEnvironmentImpl(factory, AMQP_P2P_CONTEXT) {
-            override fun toString() = "testSerializationEnv($label)"
-        }
```
distributedleetravis: A particular bugbear of mine is `open` classes that are only open so that something in test can override something in them. It's like `@VisibleForTesting` only worse. In general I think that many more things should be interfaces rather than abstract classes, and concrete implementations should be final unless there's a really good reason for them not to be (e.g. we are actually, thoughtfully and deliberately, implementing the template pattern...)

I've made `SerializationEnvironmentImpl` final and private. If it's really necessary for something in test to provide its own `SerializationEnvironment`, we can define a `TestSerializationEnvironment` implementation, but I doubt it's really needed.

core/src/main/kotlin/net/corda/core/serialization/internal/SerializationEnvironment.kt
```diff
@@ -11,38 +11,63 @@ import net.corda.core.serialization.SerializationFactory
...
 
-    init {
-        rpcServerContext?.let { this.rpcServerContext = it }
-        rpcClientContext?.let { this.rpcClientContext = it }
-        storageContext?.let { this.storageContext = it }
-        checkpointContext?.let { this.checkpointContext = it }
-        checkpointSerializationFactory?.let { this.checkpointSerializationFactory = it  }
-    }
+    override val rpcServerContext: SerializationContext get() = optionalRpcServerContext ?:
+            throw UnsupportedOperationException("RPC server serialization not supported in this environment")
```
distributedleetravis: Meaningful error messages, rather than the `UninitializedPropertyAccessException`you get if something tries to read a `lateinit` property that hasn't been init-ed.

core-deterministic/testing/verifier/src/main/kotlin/net/corda/deterministic/verifier/LocalSerializationRule.kt
fenryka: I tend to agree, if we cant test the contract the calsses interface promises without injecting things into it... making it special is probably tesitng the implementation too tightly.. not that I haven't fallen prey to that before

PR #4046, opened 2018-10-08
Python 3 compat, ENM version name, skip closed tests

tlil:
* Make the script runnable on Python 3
* Change base name of versions for Network Services, using the ENM abbreviation
* Skip parent test cases marked as "Pass" / "Fail" / "Descope" when creating new release candidate tests


release-tools/testing/test-manager
```diff
@@ -36,7 +36,7 @@ except:
 product_map = {
     'OS'   : 'Corda',
     'ENT'  : 'Corda Enterprise',
-    'NS'   : 'Corda Network Services',
```
GavT: Would be nice for ENM to mean ENM rather than NS mean ENM, i.e. `ENM`  :  `ENM`

tlil: Fixed.

PR #4055, opened 2018-10-10
Early Discussion Of 'Maximus' Scope and Provisional High Level Design

mnesbit:
This PR is to open to the wider community the state of thinking on the Maximus proposal.


triconic:
You mention in the description document that Maximus is not about managing identities - Marco Polo is. Can you provide a link in the documents to a discussion of Marco Polo? Or perhaps put a brief summary in the Maximus document to understand the difference / overlap?

I'm interested in the ability to support large numbers of users - but with that will probably come some form of permissioning based on identity i.e. possibly classes of users - not all first class. I'm interested in learning more about how this might be achieved / questioning my assumptions around it.

Thanks.

mnesbit:
There have been lots of internal discussions in R3 about massive numbers of identities and Marco Polo is our internal umbrella term for some projects considering this.
Unfortunately, none of these discussions are at a point where we have a concrete way of doing this, plus there are lots of trade offs that probably aren't what people want in reality. So for now I do not have much more that can be said on this, just that we are aware it is a common request.
What I will say is that one must consider if you actually want separate first class identities for every user? If is perfectly possible right now to manage account views using data stored in QueryableState objects and direct custom vault/jdbc queries.  In most cases what is being asked for is account specific views rather than the full isolation and public advertisement of a unique and authorised X500 Name that node identity gives.

rickcrook:
@mnesbit It might just be me but I get an uneasy feeling about this one.  It feels like MAXIMUS is working on two concerns - multiple identities/single node and HA (single identity/multi-node). A functional and non-functional requirement (NFR) respectively. I only lean in because I have not seen this type of conflation end well.

If I understand the Work Proposal, the proposed solution to running multiple identities is to run multiple nodes. A reasonable work around until when/if we are able to move beyond one identity per JVM. It closes out the functional requirement with zero code changes. My favourite type of solution! I would propose writing a short summary on why we have a design limit of one identity per JVM and closing out the functional requirement and let us focus on the NFR.

The NFR, as I understand it, the work proposed is solving is running a large number of Corda nodes.

The Work Proposal introduces both Docker and Kubernetes. I am supportive of the choice to use Docker and Kubernetes to manage large scale workloads such as Corda. This has proved a good choice by many and is increasingly looking like a de-facto standard for the industry.

May I propose that the Work Proposal focus on what Corda changes (now and future) need to occur to make Corda easier to deploy using Docker and Kubernetes including:
   1. able to use the config features of Docker/Kubernetes  eg. configMap; Secrets
   2. able to use the HA features of Docker and Kubernetes eg. restart, replica

This will force clear separation of responsibilities between the different layers - app (Corda/JVM) & infra (docker/kube) and make it easier to manage a large collection of Corda nodes (with or without multiple identities)

mikehearn:
Hey Rick,

The gist is we want to avoid large API changes to overload lots of stuff with identity parameters. Currently the node is a single process, and the current identity is an ambient property of that process, and we'd like to keep the work scoped  such that we're not deprecating lots of API and adding new work for app developers.

Now, as the design doc notes, Matthew and his team have explored supporting multiple identities concurrently in a single process, and it looks more feasible than originally believed. It may well be what we ship first. But it's important to understand that in the future the word "node" will stop meaning "JVM process that hosts an identity" and start meaning "a site installation". Therefore a node may end up being composed of multiple servers working together, coordinating via MQ and the database, and we'll need an alternative term for the processes where flows and services run (maybe "app worker", maybe "flow worker" ... still figuring that out).

With respect to how those processes are managed, that's independent of this effort. There are different people experimenting with Kubernetes and Docker. The MAXIMUS project does not say anything how servers are started or stopped anymore, only what their roles are and how they switch identities.

Whilst I appreciate your enthusiasm for Kubernetes/Docker, we probably won't be taking hard dependencies on these technologies in future - just supporting them as options for those who want them. We'll continue to be able to operate in environments without Docker/K8s and indeed even recommending them in some cases.

triconic:
Hi @mikehearn - if R3 is not going to directly support implementations using Docker and Kubernetes, would you be willing to connect interested parties? There may be useful collective feedback for R3 from an effort like this - and even if you don't support the specific technologies, supporting that pattern of deployment will be necessary.

akashmodi1508:
> Hey Rick,
> 
> The gist is we want to avoid large API changes to overload lots of stuff with identity parameters. Currently the node is a single process, and the current identity is an ambient property of that process, and we'd like to keep the work scoped such that we're not deprecating lots of API and adding new work for app developers.
> 
> Now, as the design doc notes, Matthew and his team have explored supporting multiple identities concurrently in a single process, and it looks more feasible than originally believed. It may well be what we ship first. But it's important to understand that in the future the word "node" will stop meaning "JVM process that hosts an identity" and start meaning "a site installation". Therefore a node may end up being composed of multiple servers working together, coordinating via MQ and the database, and we'll need an alternative term for the processes where flows and services run (maybe "app worker", maybe "flow worker" ... still figuring that out).
> 
> With respect to how those processes are managed, that's independent of this effort. There are different people experimenting with Kubernetes and Docker. The MAXIMUS project does not say anything how servers are started or stopped anymore, only what their roles are and how they switch identities.
> 
> Whilst I appreciate your enthusiasm for Kubernetes/Docker, we probably won't be taking hard dependencies on these technologies in future - just supporting them as options for those who want them. We'll continue to be able to operate in environments without Docker/K8s and indeed even recommending them in some cases.

Hi Mike, 

Your explanation is pretty useful in the case where we are working upon. We are struggling with managing multiple identities per node. Would you be willing to advise the on the progress of above topic?

PR #3664, opened 2018-07-22
Ordered starting of node components and clear dependencies between them

shamsasari:
Moved start up logic of the various node components out of their c’tors and into “start” methods, which are called from Node.start(). The components themselves are created in the Node’s c’tor with minimal initialisation logic.

Certain things are not immediately available at construction time, which are instead given to the components at start time in an orderly fashion:

* Certs from the node’s key store and trust store
* The network parameters, and thus access to things like maxMessageSize and the contracts whitelist
* A running database - several components were doing database stuff their c’tors
* The node’s NodeInfo, and thus access to things like identities and addresses

The messaging service couldn’t be created in the Node’s c’tor due to initialisation issues with MockNode. This should be fixed in a later commit.

PR #4056, opened 2018-10-11
ENT-2569: Perform better clean-up of DB, Hibernate mappings and IdentityService.

vkolomeyko:
This is not super-important for `AbstractNode` as when node shuts down - JVM shuts down,
but for more light-weight node-like constructs developed as part of project Maximus this can be an important improvement allowing to tidy-up the heap upon shutdown.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/services/WellKnownPartyTranslator.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/WellKnownPartyTranslator.kt
@@ -1,0 +1,26 @@
+package net.corda.core.node.services
+
+import net.corda.core.identity.AbstractParty
+import net.corda.core.identity.CordaX500Name
+import net.corda.core.identity.Party
+
+interface WellKnownPartyTranslator {
+    /**
+     * Resolves a party name to the well known identity [Party] instance for this name. Where possible well known identity
+     * lookup from name should be done from the network map (via [NetworkMapCache]) instead, as it is the authoritative
+     * source of well known identities.
+     *
+     * @param name The [CordaX500Name] to determine well known identity for.
+     * @return If known the canonical [Party] with that name, else null.
+     */
+    fun wellKnownPartyFromX500Name(name: CordaX500Name): Party?
+
+    /**
+     * Resolves a (optionally) confidential identity to the corresponding well known identity [Party].
+     * It transparently handles returning the well known identity back if a well known identity is passed in.
+     *
+     * @param party identity to determine well known identity for.
+     * @return well known identity, if found.
+     */
+    fun wellKnownPartyFromAnonymous(party: AbstractParty): Party?
+}
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
@@ -3,6 +3,8 @@ package net.corda.nodeapi.internal.persistence
 import co.paralleluniverse.strands.Strand
 import net.corda.core.schemas.MappedSchema
 import net.corda.core.utilities.contextLogger
+import org.hibernate.type.descriptor.java.AbstractTypeDescriptor
+import org.hibernate.type.descriptor.java.JavaTypeDescriptorRegistry
 import rx.Observable
 import rx.Subscriber
 import rx.subjects.UnicastSubject
@@ -52,7 +54,8 @@ class CordaPersistence(
         databaseConfig: DatabaseConfig,
         schemas: Set<MappedSchema>,
         val jdbcUrl: String,
-        attributeConverters: Collection<AttributeConverter<*, *>> = emptySet()
+        private val attributeConverters: Collection<AttributeConverter<*, *>> = emptySet(),
+        private val typeDescriptors: Collection<AbstractTypeDescriptor<*>> = emptySet()
 ) : Closeable {
     companion object {
         private val log = contextLogger()
@@ -177,6 +180,13 @@ class CordaPersistence(
     override fun close() {
         // DataSource doesn't implement AutoCloseable so we just have to hope that the implementation does so that we can close it
         (_dataSource as? AutoCloseable)?.close()
+
+        fun Collection<*>.closeAutoCloseables() {
+            mapNotNull { it as? AutoCloseable }.forEach { it.close() }
+        }
+
+        attributeConverters.closeAutoCloseables()
+        typeDescriptors.closeAutoCloseables()
     }
 }
 
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
@@ -61,6 +61,10 @@ class CordaPersistence(
         private val log = contextLogger()
     }
 
+    init {
+        typeDescriptors.forEach { JavaTypeDescriptorRegistry.INSTANCE.addDescriptor(it) }
+    }
+
     private val defaultIsolationLevel = databaseConfig.transactionIsolationLevel
     val hibernateConfig: HibernateConfiguration by lazy {
         transaction {
@@ -182,7 +186,7 @@ class CordaPersistence(
         (_dataSource as? AutoCloseable)?.close()
 
         fun Collection<*>.closeAutoCloseables() {
-            mapNotNull { it as? AutoCloseable }.forEach { it.close() }
+            forEach { (it as? AutoCloseable)?.close() }
         }
 
         attributeConverters.closeAutoCloseables()
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -12,7 +12,6 @@ import net.corda.core.context.InvocationContext
 import net.corda.core.crypto.newSecureRandom
 import net.corda.core.crypto.sign
 import net.corda.core.flows.*
-import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.identity.PartyAndCertificate
@@ -146,8 +145,7 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
     val identityService = PersistentIdentityService(cacheFactory).tokenize()
     val database: CordaPersistence = createCordaPersistence(
             configuration.database,
-            identityService::wellKnownPartyFromX500Name,
-            identityService::wellKnownPartyFromAnonymous,
+            identityService,
             schemaService,
             configuration.dataSourceProperties
     )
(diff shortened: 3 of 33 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -177,6 +180,13 @@ class CordaPersistence(
     override fun close() {
         // DataSource doesn't implement AutoCloseable so we just have to hope that the implementation does so that we can close it
         (_dataSource as? AutoCloseable)?.close()
+
+        fun Collection<*>.closeAutoCloseables() {
+            mapNotNull { it as? AutoCloseable }.forEach { it.close() }
```
shamsasari: Avoid the extra list that's created by the map by simply doing `(it as? AutoCloseable)?.close()`

node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
```diff
@@ -194,4 +194,8 @@ class PersistentIdentityService(cacheFactory: NamedCacheFactory) : SingletonSeri
         }
     }
 
-}
+    override fun close() {
+        log.info("Closing PersistentIdentityService")
```
shamsasari: This is more of a debug level message

node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
```diff
@@ -132,7 +132,7 @@ class PersistentIdentityService(cacheFactory: NamedCacheFactory) : SingletonSeri
 
     @Throws(CertificateExpiredException::class, CertificateNotYetValidException::class, InvalidAlgorithmParameterException::class)
     override fun verifyAndRegisterIdentity(identity: PartyAndCertificate, isNewRandomIdentity: Boolean): PartyAndCertificate? {
-        return database.transaction {
+        return database!!.transaction {
```
shamsasari: `database!!.transaction` is used several times. Perhaps extract out into a helper method with a useful message if `database` is null.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -52,7 +54,8 @@ class CordaPersistence(
         databaseConfig: DatabaseConfig,
         schemas: Set<MappedSchema>,
         val jdbcUrl: String,
-        attributeConverters: Collection<AttributeConverter<*, *>> = emptySet()
+        private val attributeConverters: Collection<AttributeConverter<*, *>> = emptySet(),
+        private val typeDescriptors: Collection<AbstractTypeDescriptor<*>> = emptySet()
```
shamsasari: This Set is not used by `CordaPersistence` except in the `close` method, which implies it doesn't belong in here. Or perhaps you've forgotten to do something with it as there's no equivalent code to `JavaTypeDescriptorRegistry.INSTANCE.addDescriptor` that you deleted.

node/src/main/kotlin/net/corda/node/services/persistence/AbstractPartyDescriptor.kt
```diff
@@ -10,15 +10,20 @@ import org.hibernate.type.descriptor.java.AbstractTypeDescriptor
 import org.hibernate.type.descriptor.java.ImmutableMutabilityPlan
 import org.hibernate.type.descriptor.java.MutabilityPlan
 
-class AbstractPartyDescriptor(private val wellKnownPartyFromX500Name: (CordaX500Name) -> Party?,
-                              private val wellKnownPartyFromAnonymous: (AbstractParty) -> Party?) : AbstractTypeDescriptor<AbstractParty>(AbstractParty::class.java) {
+class AbstractPartyDescriptor(private var wellKnownPartyFromX500Name: ((CordaX500Name) -> Party?)?,
```
shamsasari: This is the wrong approach - it makes the code messy and that more harder to reason about. Simply null out the reference where this class is used.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
vkolomeyko: Sure - will change.

node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
vkolomeyko: It would not happen that often, also it is an important milestone in the life cycle of the `IdentityService` I would leave it as `info` to always see in the logs clear indication when `IdentityService` shutting down.

node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
vkolomeyko: sure - will do.

node/src/main/kotlin/net/corda/node/services/persistence/AbstractPartyDescriptor.kt
vkolomeyko: Unfortunately, this is not quite possible as `AbstractPartyDescriptor` is held inside Hibernate's `JavaTypeDescriptorRegistry.INSTANCE` moreover it is only possible to `addDescriptor()` and not remove :-(

These callbacks passed into `AbstractPartyDescriptor` represent inner classes which hook into a larger object, all this object hierarchy cannot be GC'd due to Hibernate's registration.
To improve that, I decided to add ability to cut ties with callbacks, which will effectively make `AbstractPartyDescriptor` no longer functional even though it will continue to be registered with Hibernate.

szymonsztuka: I'm not sure why we need to null `wellKnownPartyFromX500Name ` and `wellKnownPartyFromAnonymous`,  if the enclosing `CordaPersistence` object is no longer available itself, then it's  GC, but anyway Hibernate still holds refs to descriptors (due to  `JavaTypeDescriptorRegistry.INSTANCE.addDescriptor()` so they will remain regardless of nulling them here or not.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -1002,28 +999,26 @@ class ConfigurationException(message: String) : CordaException(message)
 // TODO This is no longer used by AbstractNode and can be moved elsewhere
 fun configureDatabase(hikariProperties: Properties,
                       databaseConfig: DatabaseConfig,
-                      wellKnownPartyFromX500Name: (CordaX500Name) -> Party?,
-                      wellKnownPartyFromAnonymous: (AbstractParty) -> Party?,
+                      wellKnownPartyTranslator: WellKnownPartyTranslator,
```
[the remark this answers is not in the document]
vkolomeyko: No `IdentityService` was always created *before* call to `createCordaPersistence()` is made.
Please see in current `master`: https://github.com/corda/corda/blob/master/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt#L147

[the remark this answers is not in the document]
mnesbit: That changed at some point, though really it should be network map cache that provides the well known identity translation. Step me through it tomorrow if we are both in office.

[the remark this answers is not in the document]
vkolomeyko: Yes, before @shamsasari changes (corda/corda#3664) it was:
```
        // Wrapped in an atomic reference just to allow setting it before the closure below gets invoked.
        val identityServiceRef = AtomicReference<IdentityService>()

        // Do all of this in a database transaction so anything that might need a connection has one.
        val database = initialiseDatabasePersistence(
                schemaService,
                { name -> identityServiceRef.get().wellKnownPartyFromX500Name(name) },
                { party -> identityServiceRef.get().wellKnownPartyFromAnonymous(party) })
```
Which is pretty much same thing to workaround cyclic dependency problem between `PersistentIdentityService` and `CordaPersistence`.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
```diff
@@ -52,12 +54,17 @@ class CordaPersistence(
...
 ) : Closeable {
     companion object {
         private val log = contextLogger()
     }
 
+    init {
+        typeDescriptors.forEach { JavaTypeDescriptorRegistry.INSTANCE.addDescriptor(it) }
```
shamsasari: Shouldn't we also be removing these descriptors in `close`?

core/src/main/kotlin/net/corda/core/node/services/WellKnownPartyTranslator.kt
```diff
@@ -0,0 +1,26 @@
+package net.corda.core.node.services
+
+import net.corda.core.identity.AbstractParty
+import net.corda.core.identity.CordaX500Name
+import net.corda.core.identity.Party
+
+interface WellKnownPartyTranslator {
```
shamsasari: This is definitely better than the two nullable lamdbas that we had before, but we're introducing this into the public API for purely internal reasons. Can `WellKnownPartyTranslator` instead be moved into `node` and wrap an `IdentityService`.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
vkolomeyko: Unfortunately Hibernate's: `org.hibernate.type.descriptor.java.JavaTypeDescriptorRegistry` doesn't provide any way of removing descriptors... unless we go via reflection and promote private `descriptorsByClass` to be accessible and then call `remove()` on it.

core/src/main/kotlin/net/corda/core/node/services/WellKnownPartyTranslator.kt
vkolomeyko: This is a good idea!
I will make `WellKnownPartyTranslator` an internal interface, create an adaptor for `IdentityService` and remove any public API changes I made.

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/CordaPersistence.kt
shamsasari: When stuck, add another level of indirection! 

Can we make the type descriptors that we register global singletons (and we register them once in a static block), and when we move the management of the "real" descriptor implementations into the global singleton?

vkolomeyko: We can do that, but I feel that we just  trying to outgame (incorrectly designed?) Hibernate internals.
With current code when we close `PersistentIdentityService` it becomes rather lightweight in terms of memory footprint. `AbstractPartyDescriptor` still has a reference to it but again memory footprint is small and the most important thing is that `AbstractNode` is eligible for GC.

If and when subsequent registration is performed and call to `JavaTypeDescriptorRegistry.INSTANCE.addDescriptor()` is made with another instance of `AbstractPartyDescriptor` it will knock the old instance out from the `JavaTypeDescriptorRegistry.descriptorsByClass`, so at most one instance of `AbstractPartyDescriptor` will be registered at any given time.

That said, I would refrain from creating another level of indirection for very limited benefit.

mnesbit: What I want is that we can lookup the identity service lazily based upon the db context. I am not sure we should be adding any sort of stateful objects to the type registry. It should be able to look back at the scope to determine the servicehub and thence the converter.
In Maximus several isolated identities will be around and need to convert according to context. Ideally, the DB stuff will be pushed into the permanently resident part. Therefore I want this to be as stateless as we can make it.

vkolomeyko:
@mnesbit and I discussed and we might need to go back to the drawing board before making code changes in this space.
Closing this PR without merging. Will leave the branch hanging around for sometime.

PR #4064, opened 2018-10-12
CORDA-2088: Simplified the TestCordapp public API

shamsasari:
The entry point to the API has been simplified to just requireing a list of packages to scan, with sensible defaults provided for the metadata. Because of the wither methods, having parameters for the metadata (with default values) seems unnecessary. Also the ability to scan just individual classes has been made internal, as it seems unlikely app developers would need that level of control when testing their apps.

TestCordappImpl is a data class and thus just acts as a natural key for the Jar caching, where previously the key was the package names. This fixes an issue where it was not possible to create two CorDapp Jars of the same package but different metadata.


shamsasari:
It's all green finally. I had to fix some tests.

PR #4068, opened 2018-10-12
[ENT-1774] FlowAsyncOperation deduplication ID

thschroeter:
A deduplication ID is passed as parameter to the `execute` method of `FlowAsyncOperation`.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/FlowAsyncOperation.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/FlowAsyncOperation.kt
@@ -13,7 +13,7 @@ import net.corda.core.serialization.CordaSerializable
 @CordaSerializable
 interface FlowAsyncOperation<R : Any> {
     /** Performs the operation in a non-blocking fashion. */
-    fun execute(): CordaFuture<R>
+    fun execute(deduplicationId: String): CordaFuture<R>
 }
 // DOCEND FlowAsyncOperation
 
--- a/core/src/main/kotlin/net/corda/core/internal/FlowAsyncOperation.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/FlowAsyncOperation.kt
@@ -12,7 +12,14 @@ import net.corda.core.serialization.CordaSerializable
  */
 @CordaSerializable
 interface FlowAsyncOperation<R : Any> {
-    /** Performs the operation in a non-blocking fashion. */
+    /**
+     * Performs the operation in a non-blocking fashion.
+     * @param deduplicationId  If the flow restarts from a checkpoint (due to node restart, or via a visit to the flow
+     * hospital following an error) the execute method might be called more than once by the Corda flow state machine.
+     * For each duplicate call, the deduplicationId is guaranteed to be the same allowing duplicate requests to be
+     * de-duplicated if necessary inside the execute method.
+     */
+
     fun execute(deduplicationId: String): CordaFuture<R>
 }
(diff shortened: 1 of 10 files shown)


thschroeter:
This will require some doc changes before merging.

core/src/main/kotlin/net/corda/core/internal/FlowAsyncOperation.kt
```diff
@@ -13,7 +13,7 @@ import net.corda.core.serialization.CordaSerializable
 @CordaSerializable
 interface FlowAsyncOperation<R : Any> {
     /** Performs the operation in a non-blocking fashion. */
```
rick-r3: Let's add a little more to the kdoc here.  We should document the behaviour with respect to the `deduplicationId`.  So add a `@param deduplicationId` line to the kdoc.  That is, if the flow restarts from a checkpoint (due to node restart, or via a visit to the flow hospital following an error) the `execute` method might be called more than once by the Corda flow state machine.  For each duplicate call, the `deduplicationId` is guaranteed to be the same allowing duplicate requests to be de-duplicated if necessary inside the `execute` method.

rick-r3: Can we also build a test to show the de-duplication ID is the same?  I have an idea how this might be done.  This PR is extending the `FlowRetryTest` https://github.com/corda/enterprise/pull/1448/files (I'm bringing this over to OS today).

Maybe you could copy the `RetryFlow` and modify it to instead have it do some async thing that put's the `deduplicationId` into a `Set` (the Set will need to be in a companion object to avoid being reset from a checkpoint).  If the `deduplicationId` is not in the set when adding it (to the set), throw the exception that causes the retry and if it is in the set let the flow complete.  The test checks the flow completes normally (should return the `String`, rather than throw `CordaRuntimeException`).  It's fine for it to be in the same `FlowRetryTest`.

core/src/main/kotlin/net/corda/core/internal/FlowAsyncOperation.kt
```diff
@@ -12,8 +12,15 @@ import net.corda.core.serialization.CordaSerializable
...
-    /** Performs the operation in a non-blocking fashion. */
-    fun execute(): CordaFuture<R>
+    /**
+     * Performs the operation in a non-blocking fashion.
+     * @param deduplicationId  If the flow restarts from a checkpoint (due to node restart, or via a visit to the flow
+     * hospital following an error) the execute method might be called more than once by the Corda flow state machine.
+     * For each duplicate call, the deduplicationId is guaranteed to be the same allowing duplicate requests to be
+     * de-duplicated if necessary inside the execute method.
+     */
+
```
adagys: unneeded line

node/src/main/kotlin/net/corda/node/services/statemachine/transitions/StartedFlowTransition.kt
```diff
@@ -411,7 +411,8 @@ class StartedFlowTransition(
 
     private fun executeAsyncOperation(flowIORequest: FlowIORequest.ExecuteAsyncOperation<*>): TransitionResult {
         return builder {
-            actions.add(Action.ExecuteAsyncOperation(flowIORequest.operation))
+            val deduplicationId = context.id.toString() +":"+ currentState.checkpoint.numberOfSuspends.toString()
```
adagys: can you add a comment why the number of suspends is needed?
also (minor) spaces are missing around ":"

PR #4073, opened 2018-10-15
CORDA-2030: Prevent kotlin-stdlib-jre8 from becoming an accidental transitive dependency.

chrisr3:
Add `kotlin-stdlib-jre8` to the `runtimeOnly` configuration to keep it from creeping into some other module's compile classpath.


josecoll:
Sounds good - according to the docs the `runtime` is deprecated anyway.
https://docs.gradle.org/current/userguide/java_plugin.html

PR #4074, opened 2018-10-15
StatePointer

roger3cev:
Introduces the `StatePointer` concept. See design doc for more info.

Outstanding bits:

1. Tests.
2. Examples.

Will do these once the design is approved.

going forward, I'd like to experiment with making resolved pointers easier to work with.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
@@ -89,7 +89,7 @@ class LinearPointer<T : LinearState>(override val pointer: UniqueIdentifier, ove
     @Suppress("UNCHECKED_CAST")
     override fun resolve(services: ServiceHub): StateAndRef<T> {
         // Return the latest version of the linear state.
-        // TODO: Add paging as chains of provenance for token types could be quite long!
+        // This query will only ever return one or zero states.
         val query = QueryCriteria.LinearStateQueryCriteria(
                 linearId = listOf(pointer),
                 status = Vault.StateStatus.UNCONSUMED,
--- a/core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
@@ -1,0 +1,121 @@
+package net.corda.core.contracts
+
+import net.corda.core.node.ServiceHub
+import net.corda.core.node.services.Vault
+import net.corda.core.node.services.queryBy
+import net.corda.core.node.services.vault.QueryCriteria
+import net.corda.core.serialization.CordaSerializable
+import net.corda.core.transactions.LedgerTransaction
+
+/**
+ * A [StatePointer] contains a [pointer] to a [ContractState]. The [StatePointer] can be included in a [ContractState]
+ * or included in an off-ledger data structure. [StatePointer]s can be resolved to a [StateAndRef] by performing a
+ * vault query. There are two types of pointers; linear and static. [LinearPointer]s are for use with [LinearState]s.
+ * [StaticPointer]s are for use with any type of [ContractState].
+ */
+@CordaSerializable
+sealed class StatePointer<T : ContractState> {
+    /**
+     * An identifier for the [ContractState] that this [StatePointer] points to.
+     */
+    abstract val pointer: Any
+
+    /**
+     * Type of the state which is being pointed to.
+     */
+    abstract val type: Class<T>
+
+    /**
+     * Resolves a [StatePointer] to a [StateAndRef] via a vault query. This method will either return a [StateAndRef]
+     * or return an exception.
+     *
+     * @param services a [ServiceHub] implementation is required to resolve the pointer.
+     */
+    abstract fun resolve(services: ServiceHub): StateAndRef<T>
+
+    /**
+     * Resolves a [StatePointer] to a [StateAndRef] from inside a [LedgerTransaction]. The intuition here is that all
+     * of the pointed-to states will be included in the transaction as reference states.
+     *
+     * @param ltx the [LedgerTransaction] containing the [pointer] and pointed-to states.
+     */
+    abstract fun resolve(ltx: LedgerTransaction): StateAndRef<T>
+}
+
+/**
+ * A [StaticPointer] contains a [pointer] to a specific [StateRef] and can be resolved by looking up the [StateRef] via
+ * [ServiceHub]. There are a number of things to keep in mind when using [StaticPointer]s:
+ * - The [ContractState] being pointed to may be spent or unspent when the [pointer] is resolved
+ * - The [ContractState] may not be known by the node performing the look-up in which case the [resolve] method will
+ *   throw a [TransactionResolutionException]
+ */
+class StaticPointer<T : ContractState>(override val pointer: StateRef, override val type: Class<T>) : StatePointer<T>() {
+    /**
+     * Resolves a [StaticPointer] to a [StateAndRef] via a [StateRef] look-up.
+     */
+    @Throws(TransactionResolutionException::class)
+    @Suppress("UNCHECKED_CAST")
+    override fun resolve(services: ServiceHub): StateAndRef<T> {
+        val transactionState = services.loadState(pointer) as TransactionState<T>
+        val castState: T = type.cast(transactionState.data)
+        val castTransactionState: TransactionState<T> = transactionState.copy(data = castState)
+        return StateAndRef(castTransactionState, pointer)
+    }
+
+    override fun resolve(ltx: LedgerTransaction): StateAndRef<T> {
+        return ltx.referenceInputRefsOfType(type).single { pointer == it.ref }
+    }
+}
+
+/**
+ * [LinearPointer] allows a [ContractState] to "point" to another [LinearState] creating a "many-to-one" relationship
+ * between all the states containing the pointer to a particular [LinearState] and the [LinearState] being pointed to.
+ * Using the [LinearPointer] is useful when one state depends on the data contained within another state that evolves
+ * independently. When using [LinearPointer] it is worth noting:
+ * - The node performing the resolution may not have seen any [LinearState]s with the specified [linearId], as such the
+ *   vault query in [resolve] will return null and [resolve] will throw an exception
+ * - The node performing the resolution may not have the latest version of the [LinearState] and therefore will return
+ *   an older version of the [LinearState]. As the pointed-to state will be added as a reference state to the transaction
+ *   then the transaction with such a reference state cannot be committed to the ledger until the most up-to-date version
+ *   of the [LinearState] is available. See reference states documentation on docs.corda.net for more info.
+ */
+class LinearPointer<T : LinearState>(override val pointer: UniqueIdentifier, override val type: Class<T>) : StatePointer<T>() {
+    /**
+     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. Returns a
+     * [StateAndRef] containing the latest version of the [LinearState] that the node calling [resolve] is aware of.
+     *
+     * @param services a [ServiceHub] implementation is required to perform a vault query.
+     */
+    @Suppress("UNCHECKED_CAST")
+    override fun resolve(services: ServiceHub): StateAndRef<T> {
+        // Return the latest version of the linear state.
+        // TODO: Add paging as chains of provenance for token types could be quite long!
+        val query = QueryCriteria.LinearStateQueryCriteria(
+                linearId = listOf(pointer),
+                status = Vault.StateStatus.UNCONSUMED,
+                relevancyStatus = Vault.RelevancyStatus.ALL
+        )
+        val result: List<StateAndRef<LinearState>> = services.vaultService.queryBy<LinearState>(query).states
+
+        check(result.isNotEmpty()) {
+            // Here either one of two things has happened:
+            // 1. The pointed-to state has not been seen by the resolver node. It is unlikely that this is the case.
+            //    The state can probably be obtained via subscribing to the data distribution group which created and
+            //    and maintains this data.
+            // 2. Uh oh... The pointed-to state has been exited from the ledger!
+            //    It is unlikely this would ever happen as most reference data states will be created such that they cannot
+            //    be exited from the ledger. At this point there are two options; use an old consumed version of the state,
+            //    or don't use it at all.
+            "The LinearState with ID ${pointer.id} is unknown to this node or it has been exited from the ledger."
+        }
+
+        val stateAndRef = result.single()
+        val castState: T = type.cast(stateAndRef.state.data)
+        val castTransactionState = stateAndRef.state.copy(data = castState) as TransactionState<T>
+        return StateAndRef(castTransactionState, stateAndRef.ref)
+    }
+
+    override fun resolve(ltx: LedgerTransaction): StateAndRef<T> {
+        return ltx.referenceInputRefsOfType(type).single { pointer == it.state.data.linearId }
+    }
+}
--- a/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
@@ -11,6 +11,8 @@ import net.corda.core.flows.FlowLogicRef
 import net.corda.core.flows.FlowLogicRefFactory
 import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.Party
+import net.corda.core.node.ServiceHub
+import net.corda.core.node.ServicesForResolution
 import net.corda.core.serialization.CordaSerializable
 import net.corda.core.serialization.serialize
 import net.corda.core.transactions.LedgerTransaction
@@ -340,3 +342,25 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
  */
 @KeepForDJVM
 data class StateAndContract(val state: ContractState, val contract: ContractClassName)
+
+/**
+ * [LinearPointer] allows a [ContractState] to "point" to another [LinearState] creating a "many-to-one" relationship
+ * between all the states containing the pointer to a particular [LinearState] and the [LinearState] being pointed to.
+ *
+ * Using the [LinearPointer] is useful when one state depends on the data contained within another state that evolves
+ * independently.
+ */
+interface LinearPointer {
+
+    /** The [UniqueIdentifier] of the [LinearState] that this [LinearPointer] points to. */
+    val pointer: UniqueIdentifier
+
+    /**
+     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. returns a
+     * [StateAndRef] containing the latest version of the [LinearState] that the node calling [resolve] is aware of.
+     * There are two issues to note with [LinearPointer]:
+     *
+     * @param services a minimal [ServiceHub] implementation to resolve the
+     */
+    fun resolve(services: ServicesForResolution): StateAndRef<LinearState>
+}
--- a/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
@@ -371,7 +371,7 @@ interface StatePointer {
  * A [StaticPointer] contains a [pointer] to a specific [StateRef] and can be resolved by looking up the [StateRef] via
  * [ServiceHub]. There are a number of things to keep in mind when using [StaticPointer]s:
  * - The [ContractState] being pointed to may be spent or unspent when the [pointer] is resolved
- * - The [ContractState] may not the known by the node performing the look-up in which case the [resolve] method will
+ * - The [ContractState] may not be known by the node performing the look-up in which case the [resolve] method will
  *   throw a [TransactionResolutionException]
  */
 class StaticPointer(override val pointer: StateRef) : StatePointer {
--- a/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
@@ -397,7 +397,7 @@ class StaticPointer(override val pointer: StateRef) : StatePointer {
  */
 class LinearPointer(override val pointer: UniqueIdentifier) : StatePointer {
     /**
-     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. returns a
+     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. Returns a
      * [StateAndRef] containing the latest version of the [LinearState] that the node calling [resolve] is aware of.
      *
      * @param services a [ServiceHub] implementation is required to perform a vault query.
--- a/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
@@ -12,6 +12,7 @@ import net.corda.core.flows.FlowLogicRefFactory
 import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.Party
 import net.corda.core.node.ServiceHub
+import net.corda.core.node.services.Vault
 import net.corda.core.node.services.queryBy
 import net.corda.core.node.services.vault.QueryCriteria
 import net.corda.core.serialization.CordaSerializable
@@ -350,6 +351,7 @@ data class StateAndContract(val state: ContractState, val contract: ContractClas
  * vault query. There are two types of pointers; linear and static. [LinearPointer]s are for use with [LinearState]s.
  * [StaticPointer]s are for use with any type of [ContractState].
  */
+@CordaSerializable
 interface StatePointer {
     /**
      * An identifier for the [ContractState] that this [StatePointer] points to.
@@ -401,7 +403,13 @@ class LinearPointer(override val pointer: UniqueIdentifier) : StatePointer {
      * @param services a [ServiceHub] implementation is required to perform a vault query.
      */
     override fun resolve(services: ServiceHub): StateAndRef<LinearState> {
-        val query = QueryCriteria.LinearStateQueryCriteria(linearId = listOf(pointer))
+        // Returns the latest version of the LinearState,
+        // even if it is not "relevant" for the resolver node.
+        val query = QueryCriteria.LinearStateQueryCriteria(
+                linearId = listOf(pointer),
+                status = Vault.StateStatus.UNCONSUMED,
+                relevancyStatus = Vault.RelevancyStatus.ALL
+        )
         val result = services.vaultService.queryBy<LinearState>(query).states
         check(result.isNotEmpty()) { "LinearPointer $pointer cannot be resolved." }
         return result.single()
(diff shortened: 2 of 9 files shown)


docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
+# LinearPointer
+
+## Problem
+
+Occasionally there is a need to create a link from one `ContractState` to another. This has the effect of creating a uni-directional "one-to-one" relationship between a pair of `ContractState`s.
+
+A first attempt to do this might be to link one `ContractState` to another by including a `StateRef` or a `StateAndRef<T>` as follows:
+
+```kotlin
```
gendal: java first, please!

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+// StateRef.
+data class FooState(val ref: StateRef)
+// StateAndRef.
+data class FooState(val ref: StateAndRef<BarState>)
```
gendal: this example is pretty terse.  I'm not actually sure I follow it tbh!

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+Linking to a `StateRef` or `StateAndRef<T>` is only recommended if a specific version of a state is required in perpetuity. Clearly, linking to a `StateAndRef` embeds the data directly so there is no need to "resolve" the `StateRef`.
```
gendal: Is it also the case that these linkages necessarily imply the referenced states are included in (or referenced in) the associated transaction? Or can they be in other transactions that the node just happens to know about?

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+
+But what if the linked state is updated? The the `StateRef` will be pointing to an older version of the data and this could be a problem for the `ContractState` which requires the link.
+
+## Solution
+
+The solution is to include a `linearId` which references a `LinearState`. This is because all `LinearState`s contain a `linearId` which refers to a particular lineage of `LinearState`.
```
gendal: Maybe elaborate on the requirement first? "What is needed is a way to tell the node that I would like this reference to point to "the most current version of this state" or "the version of this state at a point in time", rather than be locked into the version that happened to exist at the time the transaction was created. This requirement only makes sense for states that can evolve linearly (no splitting or merging) and so the solution is...."

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+
+The trade-off here is that the data being pointed to (in the`LinearState`) cannot be immediately seen. To see the data contained within the `LinearState`, it must be "resolved".
+
+## Design
+
+Introduce a `LinearPointer` interface.
+
+The `LinearPointer` contains the `linearId` of the `LinearState` being pointed to and a `resolve` method. Resolving a `LinearPointer` returns a `StateAndRef<T>` containing the latest version of the `LinearState` that the node calling `resolve` is aware of.
+
+```kotlin
```
gendal: and java ?

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+    
+    /* 
+     * Resolves the [LinearPointer] to the latest known version of the pointed-to 
+     * [LinearState].
+     */
+    fun resolve(services: ServiceHub): StateAndRef<LinearState>?
+}
+```
+
+To use `LinearPointer`s developers must implemnent the above interface. A typical implementation of `resolve` would look like:
```
gendal: typo

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+
+```kotlin
+fun resolve(services: ServiceHub): StateAndRef<LinearState>? {
+	val query = QueryCriteria.LinearStateQueryCriteria(linearId = listOf(pointer))
+	val result = services.vaultService.queryBy<LinearState>(query)
+    return result.states.singleOrNull()?
+}
```
gendal: Mike also proposed a variation of resolve() that enabled "the version as of xxx time".  Do we not need that?

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+
+Uncoupling token type definitions from the notion of ownership. Using the `LinearPointer`, `Token` states can include an `Amount` of some pointed-to type. The pointed-to type can evolve independently from the `Token` state which should just be concerned with teh question of ownership.
+
+## Issues and resolutions
+
+Some issue to be aware of and their resolutions:
+
+| Problem                                                      | Resolution                                                   |
+| :----------------------------------------------------------- | ------------------------------------------------------------ |
+| If the node calling `resolve` has not seen any transactions containing a `LinearState` with the specified `linearId` then `resolve` will return `null`. Here, the node calling `resolve` might be missing some crucial data. | Use data distribution groups. Assuming the creator of the reference data in the `LinearState` publishes updates to a data distribution group, subscribing to that group ensures that the node calling resolve will eventually have the required data. |
```
gendal: use them _when available_?

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,125 @@
...
+| The node calling `resolve` has seen and stored transactions containing a `LinearState` with the specified `linearId`. However, there is no guarantee the `StateAndRef<T>` returned by `resolve` is the most recent version of the `LinearState`. | Embed the pointed-to `LinearState` in transactions containing the `LinearPointer` as a reference state. The reference states feature will ensure the pointed-to state is the latest version. |
+| The creator of the pointed-to `LinearState` exits the state from the ledger. If the pointed-to state is included a reference state then notaries will reject transactions containing it. | Add a `ContractState` interface called `IrremovableState` (or something). Platform logic checks for existence of this interface and if so, ensures the state cannot be exited from the ledger. |
+
+All of the noted resolutions rely on additional paltform features:
+
+* Reference states which will be available in V4
+* Data distribution groups which are not currently available. However, there is an early prototype
```
gendal: ah - ok.

roger3cev:
Had an interesting chat with @roastario and @josecoll ... There are two types of pointers which we need: static pointers and dynamic pointers. 

This PR implements dynamic pointers. Dynamic pointers must reference a `LinearState`.

Another approach can be to reference a `StateRef`, directly... This is for data which you don't want to change. e.g. a token definition.

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +343,28 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+ * [LinearPointer] allows a [ContractState] to "point" to another [LinearState] creating a "many-to-one" relationship
+ * between all the states containing the pointer to a particular [LinearState] and the [LinearState] being pointed to.
+ *
+ * Using the [LinearPointer] is useful when one state depends on the data contained within another state that evolves
+ * independently.
+ */
+interface LinearPointer {
+
+    /** The [UniqueIdentifier] of the [LinearState] that this [LinearPointer] points to. */
+    val pointer: UniqueIdentifier
```
kchalkias: As per our discussion and related slack chats, we have to ensure that the UUID is actually a function of stateRef that will ensure "on-ledger" uniqueness, otherwise you cannot tell the difference between two LinearStates with the same arbitrary-chosen UUID. I'm pretty sure there are applications that this UUID is only used for efficient Vault searching, but if a strict linking with the evolved version of a particular LinearState is required (for security purposes), we need to change the UUID generation logic behind LinearState first.

docs/source/design/linear-pointer/design.md
roger3cev: I would add Java but this is platform code which is Kotlin :)

docs/source/design/linear-pointer/design.md
roger3cev: I'll add that as a TODO.

roger3cev:
This PR implements the base types needed to make this work as well as a short design document and some necessary documentation.

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
roger3cev: Thanks for the heads up!

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +344,74 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+     * @param services a [ServiceHub] implementation is required to resolve the pointer.
+     */
+    fun resolve(services: ServiceHub): StateAndRef<ContractState>
+}
+
+/**
+ * A [StaticPointer] contains a [pointer] to a specific [StateRef] and can be resolved by looking up the [StateRef] via
+ * [ServiceHub]. There are a number of things to keep in mind when using [StaticPointer]s:
+ * - The [ContractState] being pointed to may be spent or unspent when the [pointer] is resolved
+ * - The [ContractState] may not the known by the node performing the look-up in which case the [resolve] method will
```
henrikr3: "may not the known by the node" -> typo
```suggestion
 * - The [ContractState] may not be known by the node performing the look-up in which case the [resolve] method will
```

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +344,74 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+ *   throw a [TransactionResolutionException]
+ */
+class StaticPointer(override val pointer: StateRef) : StatePointer {
+    /**
+     * Resolves a [StaticPointer] to a [StateAndRef] via a a [StateRef] look-up.
```
henrikr3: "a a"
```suggestion
     * Resolves a [StaticPointer] to a [StateAndRef] via a [StateRef] look-up.
```

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +344,74 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+ * Using the [LinearPointer] is useful when one state depends on the data contained within another state that evolves
+ * independently. When using [LinearPointer] it is worth noting:
+ * - The node performing the resolution may not have seen any [LinearState]s with the specified [linearId], as such the
+ *   vault query in [resolve] will return null
+ * - The node performing the resolution may not have the latest version of the [LinearState] and therefore will return
+ *   an older version of the [LinearState]
+ */
+class LinearPointer(override val pointer: UniqueIdentifier) : StatePointer {
+    /**
+     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. returns a
```
henrikr3: Capitalize
```suggestion
     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. Returns a
```

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,144 @@
...
+```kotlin
+// StateRef.
+data class FooState(val ref: StateRef) : ContractState
+// StateAndRef.
+data class FooState(val ref: StateAndRef<BarState>) : ContractState
+```
+
+Linking to a `StateRef` or `StateAndRef<T>` is only recommended if a specific version of a state is required in perpetuity. Clearly, adding a `StateAndRef` embeds the data directly. This type of pointer is compatible with any `ContractState` type.
+
+But what if the linked state is updated? The the `StateRef` will be pointing to an older version of the data and this could be a problem for the `ContractState` which contains the pointer.
```
henrikr3: "The the"
```suggestion
But what if the linked state is updated? The `StateRef` will be pointing to an older version of the data and this could be a problem for the `ContractState` which contains the pointer.
```

roger3cev:
Thanks @henrikr3 ! The "apply changes" feature is quite cool!

josecoll:
Great piece of work Roger!
With my Release Manager hat on, please can you add a `changelog` entry and also reference a JIRA story in the title (and associated commit). Wouldn't want to miss such a crucial change from the release notes.

docs/source/design/linear-pointer/design.md
```diff
@@ -0,0 +1,144 @@
...
+
+Symmetrical relationships can be modelled by embedding a `LinearPointer` in the pointed-to `LinearState` which points in the "opposite" direction. **Note:** this can only work if both states are `LinearState`s.
+
+## Use-cases
+
+It is important to note that this design only standardises a pattern which is currently possible with the platform. In other words, this design does not enable anything new.
+
+#### Tokens
+
+Uncoupling token type definitions from the notion of ownership. Using the `LinearPointer`, `Token` states can include an `Amount` of some pointed-to type. The pointed-to type can evolve independently from the `Token` state which should just be concerned with teh question of ownership.
```
henrikr3: "teh"
```suggestion
Uncoupling token type definitions from the notion of ownership. Using the `LinearPointer`, `Token` states can include an `Amount` of some pointed-to type. The pointed-to type can evolve independently from the `Token` state which should just be concerned with the question of ownership.
```

docs/source/api-states.rst
```diff
@@ -225,3 +225,32 @@ then the check below will fail.
                      transaction that references the encumbered state. This is because the data contained within the
                      encumbered state may take on a different meaning, and likely would do, once the encumbrance state
                      is taken into account.
+
+State Pointers
+--------------
+
+A `StatePointer` contains a pointer to a `ContractState`. The `StatePointer` can be included in a `ContractState` as a
```
mikehearn: I think you need double backticks here - does the docsite render well with this?

docs/source/api-states.rst
```diff
@@ -225,3 +225,32 @@ then the check below will fail.
...
+   points to the same `ContractState`.
+2. The `LinearPointer` is for use with `LinearState` s. They are particularly useful because due to the way `LinearState` s
+   work, the pointer will automatically point you to the latest version of a `LinearState` that the node performing `resolve`
+   is aware of. In effect, the pointer "moves" as the `LinearState` is updated.
+
+ `StatePointer` s do not enable a feature in Corda which was unavailable before. Rather, they help to formalise a pattern
+ which was already possible. In that light it is worth nothing some issues which you may encounter with `StatePointer` s:
+
+* If the node calling `resolve` has not seen any transactions containing a `ContractState` which the `StatePointer`
+  points to, then `resolve` will return an exception. Here, the node calling `resolve` might be missing some crucial data.
```
mikehearn: return -> throw

docs/source/design/linear-pointer/design.md
mikehearn: yeah design docs don't need java samples imo - this is not docs, it's design background for the curious

node/src/test/kotlin/net/corda/node/services/transactions/ResolveStatePointersTest.kt
```diff
@@ -0,0 +1,114 @@
...
+                addCommand(Command(DummyContract.Commands.Create(), myself.party.owningKey))
+            })
+            recordTransactions(listOf(tx))
+            tx.tx.outRefsOfType<Bar>().single()
+        }
+    }
+
+    @Test
+    fun `resolve state pointers and check reference state is added to transaction`() {
+        val stateAndRef = createPointedToState(barOne)
```
mikehearn: Can you add a test or two for what happens when the pointer can't resolve properly?

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +345,81 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
 data class StateAndContract(val state: ContractState, val contract: ContractClassName)
+
+/**
+ * A [StatePointer] contains a [pointer] to a [ContractState]. The [StatePointer] can be included in a [ContractState]
+ * or included in an off-ledger data structure. [StatePointer]s can be resolved to a [StateAndRef] by performing a
+ * vault query. There are two types of pointers; linear and static. [LinearPointer]s are for use with [LinearState]s.
+ * [StaticPointer]s are for use with any type of [ContractState].
+ */
+@CordaSerializable
+interface StatePointer<T : ContractState> {
```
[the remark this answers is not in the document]
roger3cev: Good point. Will convert to sealed class

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +345,81 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+ *   throw a [TransactionResolutionException]
+ */
+class StaticPointer<T : ContractState>(override val pointer: StateRef, override val type: Class<T>) : StatePointer<T> {
+    /**
+     * Resolves a [StaticPointer] to a [StateAndRef] via a [StateRef] look-up.
+     */
+    @Throws(TransactionResolutionException::class)
+    override fun resolve(services: ServiceHub): StateAndRef<T> {
+        val transactionState = uncheckedCastTransactionState<ContractState, T>(services.loadState(pointer))
+        // Type inference fail.
```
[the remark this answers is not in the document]
roger3cev: Resolved. Can't remember why I did this... Just added an annotation to the method.

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +345,81 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+     */
+    override fun resolve(services: ServiceHub): StateAndRef<T> {
+        // Returns the latest version of the LinearState,
+        // even if it is not "relevant" for the resolver node.
+        val query = QueryCriteria.LinearStateQueryCriteria(
+                linearId = listOf(pointer),
+                status = Vault.StateStatus.UNCONSUMED,
+                relevancyStatus = Vault.RelevancyStatus.ALL
+        )
+        val result = services.vaultService.queryBy<LinearState>(query).states.single()
```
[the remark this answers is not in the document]
roger3cev: Added

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -340,3 +345,81 @@ class PrivacySalt(bytes: ByteArray) : OpaqueBytes(bytes) {
...
+    /**
+     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. Returns a
+     * [StateAndRef] containing the latest version of the [LinearState] that the node calling [resolve] is aware of.
+     *
+     * @param services a [ServiceHub] implementation is required to perform a vault query.
+     */
+    override fun resolve(services: ServiceHub): StateAndRef<T> {
+        // Returns the latest version of the LinearState,
+        // even if it is not "relevant" for the resolver node.
+        val query = QueryCriteria.LinearStateQueryCriteria(
```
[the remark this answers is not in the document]
roger3cev: If the vault query returns nothing then either:
- the node hasn't seen a state with this linear ID
- the node has seen all states pertaining to this linearID, the last transaction exited from the ledger
... and the resolve method will return an exception. I can probably add a bit more colour around what happened with more instructive exceptions...

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -209,6 +210,47 @@ open class TransactionBuilder @JvmOverloads constructor(
 
     private fun checkReferencesUseSameNotary() = referencesWithTransactionState.map { it.notary }.toSet().size == 1
 
+    /**
+     * If any inputs or outputs added to the [TransactionBuilder] contain [StatePointer]s, then this method can be
+     * optionally called to resolve those [StatePointer]s to [StateAndRef]s. The [StateAndRef]s are then added as
+     * reference states to the transaction. The effect is that the referenced data is carried along with the
+     * transaction. This may or may not be appropriate in all circumstances, which is why calling this method is
```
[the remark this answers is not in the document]
roger3cev: The issue here is that then you would need overloaded add state methods for when calling from a unit test and service hub is not available.

docs/source/api-states.rst
roger3cev: Ah damn :( I always forget this

node/src/test/kotlin/net/corda/node/services/transactions/ResolveStatePointersTest.kt
roger3cev: Will add some

roger3cev:
Over to you @mikehearn  !

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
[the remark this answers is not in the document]
mikehearn: Can't we pass in the ServiceHub in the class c'tor, default it to null, and use it only when there's a hub? Some unit tests do have ServiceHubs anyway.

core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * [LinearPointer] allows a [ContractState] to "point" to another [LinearState] creating a "many-to-one" relationship
+ * between all the states containing the pointer to a particular [LinearState] and the [LinearState] being pointed to.
+ * Using the [LinearPointer] is useful when one state depends on the data contained within another state that evolves
+ * independently. When using [LinearPointer] it is worth noting:
+ * - The node performing the resolution may not have seen any [LinearState]s with the specified [linearId], as such the
+ *   vault query in [resolve] will return null and [resolve] will throw an exception
+ * - The node performing the resolution may not have the latest version of the [LinearState] and therefore will return
+ *   an older version of the [LinearState]
```
mikehearn: Can you add a sentence here observing that if the pointer is being resolved to a reference state as intended, this will mean the tx can't commit as the notary won't agree to it, and the user should read the ref state docs to learn more?

core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Resolves a [LinearPointer] using the [UniqueIdentifier] contained in the [pointer] property. Returns a
+     * [StateAndRef] containing the latest version of the [LinearState] that the node calling [resolve] is aware of.
+     *
+     * @param services a [ServiceHub] implementation is required to perform a vault query.
+     */
+    @Suppress("UNCHECKED_CAST")
+    override fun resolve(services: ServiceHub): StateAndRef<T> {
+        // Return all versions of the LinearState over time.
+        // TODO: Add paging as chains of provenance for token types could be quite long!
```
mikehearn: Ah. Yes, we should do this otherwise things will break in production. We're actually planning on deprecating the API you're using here because it's dangerous.

Can we do a query to retrieve always the latest, instead of needing paging? Stuff like selecting only unconsumed should be doable via the API.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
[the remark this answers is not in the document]
roger3cev: Yeah I guess we can do that, i feel a little bad doing so because I made a fuss that it wasn't the right way to deal with checking minimum platform versions! :)

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -11,6 +11,11 @@ import net.corda.core.flows.FlowLogicRef
 import net.corda.core.flows.FlowLogicRefFactory
 import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.Party
+import net.corda.core.node.ServiceHub
```
mikehearn: Unnecessary imports.

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -4,6 +4,9 @@ package net.corda.core.internal
 
 import net.corda.core.DeleteForDJVM
 import net.corda.core.KeepForDJVM
+import net.corda.core.contracts.ContractState
```
mikehearn: Unnecessary imports

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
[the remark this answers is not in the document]
roger3cev: but as you note... there are only really two ways to do this, pass into method or pass into constructor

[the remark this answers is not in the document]
roger3cev: Done

core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
roger3cev: Added!

core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
roger3cev: Changed. Now `resolve` throws an exception to say that the state is unknown OR has been exited from the ledger.

core/src/main/kotlin/net/corda/core/contracts/Structures.kt
roger3cev: removed

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
roger3cev: removed

roger3cev:
Back to you @mikehearn - thanks again for review!

core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
mikehearn: The TODO still talks about the need for paging? If we don't implement this we are guaranteed to see crashes in production and an emergency fix be required.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -209,6 +217,42 @@ open class TransactionBuilder @JvmOverloads constructor(
...
+     * If this method is called outside the context of a flow, a [ServiceHub] instance must be passed to this method
+     * for it to be able to resolve [StatePointer]s. Usually for a unit test, this will be an instance of mock services.
+     *
+     * @param serviceHub a [ServiceHub] instance needed for performing vault queries.
+     *
+     * @throws IllegalStateException if no [ServiceHub] is provided and no flow context is available.
+     */
+    private fun resolveStatePointers(transactionState: TransactionState<*>) {
+        if (serviceHub == null) {
+            logger.warn("You must pass in a ServiceHub reference to TransactionBuilder to resolve state pointers " +
```
mikehearn: Can we only log a warning if there are actually any pointers to resolve?

roger3cev: Sure

core/src/main/kotlin/net/corda/core/contracts/StatePointer.kt
roger3cev: Whoops need to remove that. The query now only asks for a single linear state.

PR #4075, opened 2018-10-15
CORDA-1707 Tests to prove bug doesn't exist.

rick-r3:
We'd theorised a bug to do with re-generating flow IDs but the tests prove it doesn't exist.  Going to commit the tests to ensure we do not regress.


rick-r3:
BTW, the RPC client gets a `CordaRuntimeException` if a flow constructor throws during startFlow.

PR #4077, opened 2018-10-15
Documents default node tables.

joeldudleyr3:


docs/source/node-database.rst
```diff
@@ -1,20 +1,20 @@
 Node database
 =============
 
-Default in-memory database
---------------------------
+.. contents::
+
+Configuring the node database
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
szymonsztuka: I'm afraid that we are loosing main info that OS works with H2, there's still a link but previously it was a paragraph "Default in-memory database" as the same level with PostgresSQL, SQL Server, but we can address it later

PR #4083, opened 2018-10-17
CORDA-2111: Upgrade to Kotlin 1.2.71.

chrisr3:
Kotlin 1.2.71 was released on 24th September 2018, and our Gradle plugins are using it successfully.
Here are the [release notes](https://blog.jetbrains.com/kotlin/2018/09/kotlin-1-2-70-is-out)


josecoll:
Any specific reason why we should upgrade to 1.2.71 now? (security fixes, other bugs that affect our codebase)

PR #4086, opened 2018-10-17
CORDA-2113 - Include PNM ID in CSR

fenryka:
If Compatibility Zone operator is using private networks and the node
should be joining one, optionally the ID (a UUID) of that network can be
included as part of the node's CSR to to the Doorman.


node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
```diff
@@ -200,7 +199,7 @@ open class NetworkRegistrationHelper(private val certificatesDirectory: Path,
                 if (idlePeriodDuration != null) {
                     Thread.sleep(idlePeriodDuration.toMillis())
                 } else {
-                    throw NodeRegistrationException(e)
+                    throw NodeRegistrationException(null, e)
```
oknowles: It might be a bit of an edge case where this error occurs and gets printed, but it would be more helpful to have a message here rather than `null`.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -152,7 +152,7 @@ open class NodeStartup : CordaCliWrapper("corda", "Runs a Corda Node") {
 
     private val handleRegistrationError = { error: Exception ->
         when (error) {
-            is NodeRegistrationException -> error.logAsExpected("Node registration service is unavailable. Perhaps try to perform the initial registration again after a while.")
+            is NodeRegistrationException -> error.logAsExpected("Issue with Node registration: ${error.message}")
```
oknowles: This is much nicer!

node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt
fenryka: technically this causes it to print the origional *catch all* error message... but that behaviour is a bit magic so let's, as you say, actually return the error message here

oknowles:
LGTM

PR #4088, opened 2018-10-17
ENT-1906: Allow DJVM code to throw and catch sandbox exceptions.

chrisr3:
- Support for `throw`, `catch`, `finally` and stack traces.
- Create a template `sandbox.java.lang.Throwable` class that will become the non-throwable base class for sandbox exceptions.
- Create a two-way mapping being non-throwable sandbox exceptions and throwable exception wrappers. For the most part, this involves generating a throwable synthetic counterpart for each exception type. However, exceptions which can be thrown by the JVM itself (including _all_ of their parent exception types) are handled natively. This results in an exception hierarchy like:
```
sandbox.java.lang.Throwable <-> java.lang.Throwable
sandbox.java.lang.Exception <-> java.lang.Exception
sandbox.java.lang.RuntimeException <-> java.lang.RuntimeException
sandbox.java.time.DateTimeException <-> sandbox.java.time.DateTimeException.1DJVM
sandbox.net.corda.core.CordaRuntimeException <-> sandbox.net.corda.core.CordaRuntimeException.1DJVM
...
```
- Ensure that synthetic `.1DJVM` exception classes are generated automatically, immediately after their "owner" classes are loaded. If the JVM attempts to load an unknown `.1DJVM` class then the "owner" class is loaded, which triggers the generation of the synthetic class.
- Exception stack traces are truncated to exclude elements from outside the sandbox.
- Emitters are now ordered by integer "priority", for greater control over which emitters execute first.

### DIFF
--- a/djvm/src/main/java/sandbox/java/lang/String.java
+++ b/djvm/src/main/java/sandbox/java/lang/String.java
@@ -7,6 +7,7 @@ import sandbox.java.util.Locale;
 
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Constructor;
 
 @SuppressWarnings("unused")
 public final class String extends Object implements Comparable<String>, CharSequence, Serializable {
@@ -21,6 +22,17 @@ public final class String extends Object implements Comparable<String>, CharSequ
 
     private static final String TRUE = new String("true");
     private static final String FALSE = new String("false");
+
+    private static final Constructor SHARED;
+
+    static {
+        try {
+            SHARED = java.lang.String.class.getDeclaredConstructor(char[].class, java.lang.Boolean.TYPE);
+            SHARED.setAccessible(true);
+        } catch (NoSuchMethodException e) {
+            throw new NoSuchMethodError(e.getMessage());
+        }
+    }
 
     private final java.lang.String value;
 
@@ -86,6 +98,17 @@ public final class String extends Object implements Comparable<String>, CharSequ
 
     public String(StringBuilder builder) {
         this.value = builder.toString();
+    }
+
+    String(char[] value, boolean share) {
+        java.lang.String newValue;
+        try {
+            // This is (presumably) an optimisation for memory usage.
+            newValue = (java.lang.String) SHARED.newInstance(value, share);
+        } catch (Exception e) {
+            newValue = new java.lang.String(value);
+        }
+        this.value = newValue;
     }
 
     @Override
--- a/djvm/src/main/kotlin/net/corda/djvm/analysis/PrefixTree.kt
+++ b/djvm/src/main/kotlin/net/corda/djvm/analysis/PrefixTree.kt
@@ -1,7 +1,7 @@
 package net.corda.djvm.analysis
 
 /**
- * Trie data structure to make prefix matching more efficient.
+ * Tree data structure to make prefix matching more efficient.
  */
 class PrefixTree {
 
--- a/djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
+++ b/djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
@@ -3,6 +3,9 @@ package net.corda.djvm.rewiring
 import net.corda.djvm.SandboxConfiguration
 import net.corda.djvm.analysis.AnalysisContext
 import net.corda.djvm.analysis.ClassAndMemberVisitor
+import net.corda.djvm.analysis.ExceptionResolver.Companion.getDJVMExceptionOwner
+import net.corda.djvm.analysis.ExceptionResolver.Companion.isDJVMException
+import net.corda.djvm.code.asPackagePath
 import net.corda.djvm.code.asResourcePath
 import net.corda.djvm.references.ClassReference
 import net.corda.djvm.source.ClassSource
@@ -33,7 +36,7 @@ class SandboxClassLoader(
     /**
      * The analyzer used to traverse the class hierarchy.
      */
-    val analyzer: ClassAndMemberVisitor
+    private val analyzer: ClassAndMemberVisitor
         get() = ruleValidator
 
     /**
@@ -57,6 +60,18 @@ class SandboxClassLoader(
     private val rewriter: ClassRewriter = ClassRewriter(configuration, supportingClassLoader)
 
     /**
+     * We need to load this class up front, so that we can identify sandboxed exception classes.
+     */
+    private val throwableClass: Class<*>
+
+    init {
+        // Bootstrap the loading of the sandboxed Throwable class.
+        loadClassAndBytes(ClassSource.fromClassName("sandbox.java.lang.Object"), context)
+        loadClassAndBytes(ClassSource.fromClassName("sandbox.java.lang.StackTraceElement"), context)
+        throwableClass = loadClassAndBytes(ClassSource.fromClassName("sandbox.java.lang.Throwable"), context).type
+    }
+
+    /**
      * Given a class name, provide its corresponding [LoadedClass] for the sandbox.
      */
     fun loadForSandbox(name: String, context: AnalysisContext): LoadedClass {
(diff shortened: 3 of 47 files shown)


djvm/src/main/kotlin/net/corda/djvm/analysis/PrefixTree.kt
```diff
@@ -1,7 +1,7 @@
 package net.corda.djvm.analysis
 
 /**
- * Trie data structure to make prefix matching more efficient.
+ * Tree data structure to make prefix matching more efficient.
```
tlil: [Trie](https://en.wikipedia.org/wiki/Trie) = prefix tree ;)

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
```diff
@@ -78,13 +93,38 @@ class SandboxClassLoader(
...
+        return if (analysisConfiguration.isSandboxClass(source.internalClassName)) {
+            loadSandboxClass(source, context).type
         } else {
             super.loadClass(name, resolve)
         }
     }
 
+    private fun loadSandboxClass(source: ClassSource, context: AnalysisContext): LoadedClass {
+        return if (isDJVMException(source.internalClassName)) {
+            /**
```
tlil: Why kdoc comments here? :-)

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/HandleExceptionUnwrapper.kt
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Converts an exception from [java.lang.Throwable] to [sandbox.java.lang.Throwable]
+ * at the beginning of either a catch block or a finally block.
+ */
+class HandleExceptionUnwrapper : Emitter {
+    private val handlers = mutableMapOf<Label, String>()
+
+    override fun emit(context: EmitterContext, instruction: Instruction) = context.emit {
+        if (instruction is TryBlock) {
+            handlers[instruction.handler] = instruction.typeName
```
tlil: Looking at this, perhaps `handler` should be called `handlerLabel` instead for clarity? Wdyt?

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/HandleExceptionUnwrapper.kt
```diff
@@ -0,0 +1,46 @@
...
+                     * This is a finally block; the wrapping function MUST NOT throw exceptions.
+                     */
+                    invokeStatic("sandbox/java/lang/DJVM", "finally", "(Ljava/lang/Throwable;)Lsandbox/java/lang/Throwable;")
+                }
+            }
+        }
+    }
+
+    override val priority: Int
+        get() = EMIT_TRAPPING_EXCEPTIONS + 1
```
tlil: When there's an explicit ordering dependency between two or more emitters within a single "category" (tracers, trapping exceptions, etc.), perhaps it'd be good to express them based on each other in ascending order?

E.g.,

`Emitter1.priority = EMIT_TRAPPING_EXCEPTIONS`
`Emitter2.priority = Emitter1.priority + 1`
`...`

tlil:
Looks great - just done a 1st pass. Will go through in more detail tomorrow.

djvm/src/main/kotlin/net/corda/djvm/analysis/PrefixTree.kt
chrisr3: OK, I assumed it was a typo :wink:.

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
chrisr3: Probably because I was originally thinking of "linking" the `DJVMException` reference. Does my use of KDoc threaten any kittens :cat2:?

djvm/src/test/kotlin/net/corda/djvm/analysis/ExceptionResolverTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import net.corda.djvm.analysis.AnalysisConfiguration.Companion.JVM_EXCEPTIONS
+import net.corda.djvm.analysis.AnalysisConfiguration.Companion.SANDBOX_PREFIX
+import net.corda.djvm.code.ruleViolationError
+import net.corda.djvm.code.thresholdViolationError
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import org.objectweb.asm.Type
+
+class ExceptionResolverTest {
```
exFalso: No tests?

djvm/src/main/java/sandbox/java/lang/String.java
```diff
@@ -310,6 +335,8 @@ public String trim() {
         return toDJVM(value.trim());
     }
 
+    public String intern() { return INTERNAL.computeIfAbsent(value, s -> this); }
```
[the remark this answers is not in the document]
chrisr3: Something was complaining that `String.intern()` wasn't implemented, and so I added it. The `sandbox.java.lang.String` contains
```java
static final Map<java.lang.String, sandbox.java.lang.String> INTERNAL;
```
which will be destroyed when the sandbox is destroyed. I never actually invoke `java.lang.String.intern()` here.

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/HandleExceptionUnwrapper.kt
```diff
@@ -0,0 +1,46 @@
...
+                    /**
+                     * When catching exceptions, we also need to tell the verifier which
+                     * which kind of [sandbox.java.lang.Throwable] to expect this to be.
+                     */
+                    castObjectTo(exceptionType)
+                } else {
+                    /**
+                     * This is a finally block; the wrapping function MUST NOT throw exceptions.
+                     */
+                    invokeStatic("sandbox/java/lang/DJVM", "finally", "(Ljava/lang/Throwable;)Lsandbox/java/lang/Throwable;")
```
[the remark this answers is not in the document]
chrisr3: Exceptions entering `finally` blocks are rethrown at the end via `ATHROW`. The DJVM is invoking
```java
java.lang.Throwable sandbox.java.lang.DJVM.toDJVM(sandbox.java.lang.Throwable t):
```
before _every_ `ATHROW`, so the exception inside the `finally` still needs to be wrapped. However, it does _not_ need to be exposed, and so I wrap it inside `DJVMThrowableWrapper` instead which the DJVM can trivially unwrap at the end.

Aborting the `finally` block prematurely by rethrowing the exception is _not_ an option.

djvm/src/main/kotlin/sandbox/java/lang/DJVM.kt
```diff
@@ -154,5 +168,130 @@ private fun toSandbox(className: kotlin.String): kotlin.String {
...
+private fun Class<*>.createJavaThrowable(t: Throwable): kotlin.Throwable {
+    return (try {
+        getDeclaredConstructor(kotlin.String::class.java).newInstance(String.fromDJVM(t.message))
+    } catch (e: NoSuchMethodException) {
+        newInstance()
+    } as kotlin.Throwable).apply {
+        t.cause?.also {
+            initCause(fromDJVM(it))
+        }
+        stackTrace = copyFromDJVM(t.stackTrace)
```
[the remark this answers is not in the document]
chrisr3: I don't think so. Where would the "dirty" information be coming from?

djvm/src/main/kotlin/sandbox/java/lang/DJVM.kt
```diff
@@ -154,5 +168,130 @@ private fun toSandbox(className: kotlin.String): kotlin.String {
...
+ */
+private fun kotlin.Throwable.toDJVMThrowable(): Throwable {
+    return (this as? DJVMException)?.getThrowable() ?: javaClass.toDJVMType().createDJVMThrowable(this)
+}
+
+/**
+ * Creates a new [sandbox.java.lang.Throwable] from a [java.lang.Throwable],
+ * which was probably thrown by the JVM itself.
+ */
+private fun Class<*>.createDJVMThrowable(t: kotlin.Throwable): Throwable {
```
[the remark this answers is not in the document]
chrisr3: Yes, the synthetic `.1DJVM` exceptions all have trivial constructors that accept `sandbox.java.lang.Throwable` only. It's only the JVM exceptions where we need to hunt for a suitable constructor via reflection. And almost all of those support `(java.lang.String)`, although some just have no-arg constructors.

djvm/src/test/kotlin/net/corda/djvm/analysis/ExceptionResolverTest.kt
chrisr3: No; there used to be but I refactored everything into a brick wall :face_with_head_bandage:.

djvm/src/main/kotlin/sandbox/java/lang/DJVM.kt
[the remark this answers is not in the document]
exFalso: Won't the stacktrace include stack trace elements from outside the sandbox?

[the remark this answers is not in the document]
chrisr3: Stack traces are created by native code, so  `sandbox.Throwable` must create an actual `java.Throwable` in its `fillStackTrace()` and then clean the stack trace up. So `sandbox.Throwable` stack traces are already "clean".

PR #4089, opened 2018-10-17
[CORDA-738] Ensure encumbrances are bi-directional

kchalkias:
Encumbrances should form a complete directed cycle, otherwise one can spend the "encumbrance" state, which would freeze the "encumbered" state for ever. This should also apply to encumbrance chains with more than two elements. Changes on `NotaryChangeLedgerTransaction` were also required. 
*Note: I didn't touch `ContractUpgradeLedgerTransaction`

For instance, the following settings are valid: 
_a -> b, b -> a_
- - -
_a -> b, b -> c, c -> a_ 
- - -
_a -> c, b -> a, c -> b_

while the following are not:
_a -> b_    (because one can spend _b_, which will freeze _a_ for ever)
- - -
_a -> c, b -> c_    (because, among the others, one can spend _a_ and _c_, which will freeze _b_ for ever)
- - -
_a -> b, b -> c_   (because, among the others, one can spend _c_, which will freeze _a_ and _b_ for ever)

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
@@ -123,7 +123,7 @@ sealed class TransactionVerificationException(val txId: SecureHash, message: Str
     @KeepForDJVM
     class TransactionDuplicateEncumbranceException(txId: SecureHash, index: Int)
         : TransactionVerificationException(txId, "The bi-directionality property of encumbered output states " +
-            "is not satisfied. Index $index is referenced more than one times", null)
+            "is not satisfied. Index $index is referenced more than once", null)
 
     /**
      * An encumbered state should also be referenced as the encumbrance of another state in order to satisfy the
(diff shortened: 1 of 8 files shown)


kchalkias:
@adagys wrt to UX, this PR will focus on the underlying logic before it's too late for security purposes (so, in Corda 4.0). I will then create another PR for UX.
Good point on the Cash-TimeLock tests, I would advise against doing things in the background as we might not know if application-wise all of the Cash states are encumbered together or not. One has to create multiple timelock states and encumber each state separately with its TimeLock. This will also be part of the UI PR.

kchalkias:
back to you @adagys. `statesAndEncumbrance` is now destructured and I also added two tests for independent chains in the same tx, to capture how `Cash -> its own TimeLock` should be implemented.

adagys:
We also need to make a decision what to do with contract upgrades. The answer may be that we don't support it for encumbered states.. as it might get really complex

gendal:
There don't seem to be any updates to the docs or updates to the changelog/release notes, etc?  Is this something users need to know about?

kchalkias:
This is mostly a major security fix for encumbrances. I guess it makes sense to add a few lines, however more work is required to make encumbrances fully functional (ie contract upgrades, coin selection, UX etc). I can mention this change in an upcoming docs PR, but I don’t want to incentivize users to use it yet.

gendal:
ok - thanks.

PR #4090, opened 2018-10-18
ENT-2610: Separate passwords for store and for private keys in Corda OS.

vkolomeyko:
When it comes to KeyStores there are *2* passwords: 1 for the keyStore as a whole and separately there is one private keys within this keyStore.
Unfortunately, those 2 passwords have to be the same due to Artemis limitation, for more details please see:
`org.apache.activemq.artemis.core.remoting.impl.ssl.SSLSupport.loadKeyManagerFactory`
where it is calling `KeyManagerFactory.init()` with store password.

Before change in this PR, throughout our codebase there are multiple places where we assume that storePassword is the same as keyPassword, even in the classes that have nothing to do with Artemis.
This is of course less than ideal as TLS communication may be used not only for Artemis connectivity (e.g. Bridge/Float interaction in Ent) and it is unfair to impose same passwords constraint on that communication channel.
Therefore this PR is removing this limitation and properly separating storePassword from keyPassword.

Linked Jira(https://r3-cev.atlassian.net/browse/ENT-2610) has for more background info.

Suggest to start review from `net.corda.core.crypto.X509NameConstraintsTest` to get an idea about the nature of the changes made.


vkolomeyko:
@sollecitom and @kchalkias - thanks for your input for this PR!

I have incorporated some of the changes and pushed them.
Both of you mentioned that a keystore may have multiple keys which may be protected by several distinct passwords. So having a single `keyPassword` is somewhat restrictive.

I looked at the API methods and there is: https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/KeyManagerFactory.html which has `init()` method that is taking a *single* password.
The Javadoc for it says:
`password - the password for recovering keys in the KeyStore`, Note: keys is in the plural form.

Are you absolutely sure that distinct passwords for distinct entries are allowed?
If "yes" - in your opinion, which of those passwords would be appropriate to pass into aforementioned `KeyManagerFactory.init()` method? (we do use it in several places in our codebase through extension method: `KeyManagerFactory.init(keyStore: CertificateStore)`)

vkolomeyko:
Thanks @kchalkias.

I understood that our `CertificateStore` and `X509KeyStore` are for dual use at the moment, they may act as SSL related key store as well as general key store.
That said, I would refrain from doing more refactoring in this space for now, especially given that it will not yield much of a benefit w.r.t. the immediate task I am working on: https://r3-cev.atlassian.net/browse/ENT-2610

Could you kindly have a second look at my PR and let me know whether you are happy with it?

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStoreSupplier.kt
```diff
@@ -18,7 +18,7 @@ interface CertificateStoreSupplier {
 }
 
 // TODO replace reference to FileBasedCertificateStoreSupplier with CertificateStoreSupplier, after coming up with a way of passing certificate stores to Artemis.
-class FileBasedCertificateStoreSupplier(val path: Path, val password: String) : CertificateStoreSupplier {
+class FileBasedCertificateStoreSupplier(val path: Path, val storePassword: String, val keyPassword: String) : CertificateStoreSupplier {
```
kchalkias: change to `entryPassword`

PR #4091, opened 2018-10-18
CORDA-1838: Add subcommands to node

anthonykeenan:
* Replaced `--install-shell-extensions` flag with `install-shell-extensions` command which ignores required parameters
* Replaced `--clear-network-map-cache` flag with `clear-network-cache` command
* Replaced `--just-generate-node-info` flag with `generate-node-info` command
* Replaced `--just-generate-rpc-ssl-settings` flag with `generate-rpc-ssl-settings` command
* Replaced `--bootstrap-raft-cluster` flag with `bootstrap-raft-cluster` command
* Replaced `--initial-registration` flag with `initial-registration` command

All previous flags are still there, but don't show up in the help usage. If a deprecated flag is used it logs a WARN message to the log file and to the screen.


docs/source/corda-configuration-file.rst
```diff
@@ -125,7 +125,7 @@ absolute path to the node's base directory.
         .. note:: The RPC SSL certificate is used by RPC clients to authenticate the connection.
             The Node operator must provide RPC clients with a truststore containing the certificate they can trust.
             We advise Node operators to not use the P2P keystore for RPC.
-            The node ships with a command line argument "--just-generate-rpc-ssl-settings", which generates a secure keystore
+            The node can be ran with the "generate-rpc-ssl-settings" command, which generates a secure keystore
```
sollecitom: "can be run"

docs/source/network-bootstrapper.rst
```diff
@@ -255,14 +255,18 @@ The network bootstrapper can be started with the following command-line options:
 
 .. code-block:: shell
 
-    bootstrapper [-hvV] [--install-shell-extensions] [--no-copy] [--dir=<dir>]
-                 [--logging-level=<loggingLevel>]
+    bootstrapper [-hvV] [--no-copy] [--dir=<dir>] [--logging-level=<loggingLevel>]
+                 [--minimum-platform-version=<minimumPlatformVersion>] [COMMAND]
```
sollecitom: Is this a new option? Happy if it is, commenting just to ensure it's not a pasting mistake.

sollecitom: If it's a legit new option, it should be documented below, before the `Sub-commands` section.

docs/source/network-map.rst
```diff
@@ -63,7 +63,7 @@ be used to supplement or replace the HTTP network map. If the same node is adver
 latest one is taken.
 
 On startup the node generates its own signed node info file, filename of the format ``nodeInfo-${hash}``. It can also be
-generated using the ``--just-generate-node-info`` command line flag without starting the node. To create a simple network
+generated using the ``generate-node-info`` command line flag without starting the node. To create a simple network
```
sollecitom: Replace "command line flag" with "sub-command".

docs/source/running-a-node.rst
```diff
@@ -48,27 +48,38 @@ Command-line options
...
 
+Sub-commands
+^^^^^^^^^^^^
+
+``bootstrap-raft-cluster``: Bootstraps Raft cluster. The node forms a single node cluster (ignoring otherwise configured peer
+addresses), acting as a seed for other nodes to join the cluster.
+
+``clear-network-cache``: Clears local copy of network map, on node startup it will be restored from server or file system.
+
+``initial-registration``: Start initial node registration with the compatibility zone to obtain a certificate from the Doorman.
```
sollecitom: I know it's nothing about this PR, but "Starts" here, rather than "Start".

docs/source/running-a-node.rst
```diff
@@ -48,27 +48,38 @@ Command-line options
...
+
+Parameters:
+
+* ``--network-root-truststore``, ``-t`` **required**: Network root trust store obtained from network operator.
+* ``--network-root-truststore-password``, ``-p``: Network root trust store password obtained from network operator.
+
+``generate-node-info``: Perform the node start-up task necessary to generate its nodeInfo, save it to disk, then quit.
```
sollecitom: "Generates the `node-info` file then quits"?

docs/source/running-a-node.rst
```diff
@@ -48,27 +48,38 @@ Command-line options
...
+
+``generate-rpc-ssl-settings``: Generate the ssl keystore and truststore for a secure RPC connection.
```
sollecitom: "Generates"

sollecitom: SSL

docs/source/running-a-node.rst
```diff
@@ -48,27 +48,38 @@ Command-line options
...
+
+``install-shell-extensions``: Install ``corda`` alias and auto completion for bash and zsh. See :doc:`cli-application-shell-extensions` for more info.
```
sollecitom: "Installs".

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -27,6 +27,63 @@ class NodeCmdLineOptions {
...
+    open fun loadConfig(): NodeConfiguration {
+        return getRawConfig().parseAsNodeConfiguration(unknownConfigKeysPolicy::handle)
+    }
+
+    protected fun getRawConfig(): Config {
+        val rawConfig = ConfigHelper.loadConfig(
+                baseDirectory,
+                configFile
+        )
+        if (devMode == true) {
```
sollecitom: Not sure side-effects like these should be here.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -2,18 +2,18 @@ package net.corda.node
...
 import net.corda.node.services.config.NodeConfiguration
+import net.corda.node.services.config.NodeConfigurationImpl
 import net.corda.node.services.config.parseAsNodeConfiguration
 import net.corda.nodeapi.internal.config.UnknownConfigKeysPolicy
 import picocli.CommandLine.Option
 import java.nio.file.Path
 import java.nio.file.Paths
 
-class NodeCmdLineOptions {
+open class SharedNodeCmdLineOptions {
```
sollecitom: Not sure how "shared" can this be, existing under `node` and using "node.conf" as a default. Might be fine for now but, as an example, the notary configuration is already changing not to be part of "node.conf" anymore.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -46,89 +103,72 @@ class NodeCmdLineOptions {
...
+            hidden = true
     )
-    var justGenerateNodeInfo: Boolean = false
+    var clearNetworkMapCache: Boolean = false
 
     @Option(
-            names = ["--just-generate-rpc-ssl-settings"],
-            description = ["Generate the SSL key and trust stores for a secure RPC connection."]
+            names = ["--initial-registration"],
+            description = ["DEPRECATED. Start initial node registration with Corda network to obtain certificate from the permissioning server."],
```
sollecitom: "Starts"

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -46,89 +103,72 @@ class NodeCmdLineOptions {
...
+            hidden = true
     )
-    var networkRootTrustStorePassword: String? = null
+    var justGenerateNodeInfo: Boolean = false
 
     @Option(
-            names = ["--on-unknown-config-keys"],
-            description = ["How to behave on unknown node configuration. \${COMPLETION-CANDIDATES}"]
+            names = ["--just-generate-rpc-ssl-settings"],
+            description = ["DEPRECATED. Generate the SSL key and trust stores for a secure RPC connection."],
```
sollecitom: "Generates"

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -717,7 +717,7 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
             requireNotNull(getCertificateStores()) {
                 "One or more keyStores (identity or TLS) or trustStore not found. " +
                         "Please either copy your existing keys and certificates from another node, " +
-                        "or if you don't have one yet, fill out the config file and run corda.jar --initial-registration. " +
+                        "or if you don't have one yet, fill out the config file and run corda.jar Finitialinitial-registration. " +
```
sollecitom: "Finitialinitial-registration" has a typo.

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -111,7 +112,7 @@ open class Node(configuration: NodeConfiguration,
 
         fun printWarning(message: String) {
             Emoji.renderIfSupported {
-                println("${Emoji.warningSign} ATTENTION: $message")
+                println("${ShellConstants.RED}${Emoji.warningSign} ATTENTION: $message${ShellConstants.RESET}")
```
sollecitom: Can we create a function `printlnInRed` that does the above? Also, `Node` referencing `ShellConstants` sounds wrong.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -47,206 +36,144 @@ import java.nio.file.Path
...
-     * @return exit code based on the success of the node startup. This value is intended to be the exit code of the process.
-     */
     override fun runProgram(): Int {
-        val startTime = System.currentTimeMillis()
+        //deal with legacy flags and redirect to subcommands
+        if (InitialRegistration.checkRegistrationMode(cmdLineOptions.baseDirectory)) {
+            println("Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.")
+            return initialRegistrationCli.runProgram()
+        }
+        if (cmdLineOptions.isRegistration) {
```
sollecitom: Maybe use `return when(cmdLineOptions) { ... }`.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -47,206 +36,144 @@ import java.nio.file.Path
...
+        const val LOGS_CAN_BE_FOUND_IN_STRING = "Logs can be found in"
+    }
+
+    lateinit var cmdLineOptions: SharedNodeCmdLineOptions
+
+    fun initialiseAndRun(cmdLineOptions: SharedNodeCmdLineOptions, afterNodeInitialisation: RunAfterNodeInitialisation): Int {
+        this.cmdLineOptions = cmdLineOptions
+
         // Step 1. Check for supported Java version.
         if (!isValidJavaVersion()) return ExitCodes.FAILURE
```
sollecitom: `if(isInvalidJavaVersion)`?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -594,4 +405,67 @@ open class NodeStartup : CordaCliWrapper("corda", "Runs a Corda Node") {
...
+/** Provide some common logging methods for node startup commands. */
+interface NodeStartupLogging {
+    companion object {
+        val logger by lazy { contextLogger() }
+        val startupErrors = setOf(MultipleCordappsForFlowException::class, CheckpointIncompatibleException::class, AddressBindingException::class, NetworkParametersReader::class, DatabaseIncompatibleException::class)
+    }
+
+    fun <RESULT> attempt(action: () -> RESULT): Try<RESULT> = Try.on(action)
+
+    fun Exception.logAsExpected(message: String? = this.message, print: (String?) -> Unit = logger::error) = print(message)
```
sollecitom: Mind passing `, error` here too, and collapse these 2 to a `fun Exception.log(...)`? Nothing to do with your PR, but needed and agreed on, and it sounds convenient to bundle it :)

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
```diff
@@ -169,7 +152,28 @@ abstract class CordaCliWrapper(val alias: String, val description: String) : Cal
...
+ * Picocli will prioritise versionProvider from the `@Command` annotation on the subclass, see: https://picocli.info/#_reuse_combinations
+ */
+abstract class CordaCliWrapper(alias: String, description: String) : CliWrapperBase(alias, description) {
+    companion object {
+        private val logger by lazy { contextLogger() }
+    }
+
+    val installShellExtensionsParser = InstallShellExtensionsParser(this.javaClass.name)
+
+    val subCommands = mutableListOf<CliWrapperBase>(installShellExtensionsParser)
```
sollecitom: We should try to avoid non-private mutable fields. Can we replace this with an alternative? In order of preference:

1. an `additionalSubCommands: Set<CliWrapperBase>` passed to the constructor, then joined into `subCommands: Set<CliWrapperBase> = additionalSubCommands + installShellExtensionsParser`.
2. a `protected abstract fun additionalSubCommands(): Set<CliWrapperBase>`, used as per above.
3. a `protected open subCommands(): Set<CliWrapperBase> = setOf(installShellExtensionsParser)`

docs/source/network-bootstrapper.rst
anthonykeenan: I just ran the `bootstrapper --help` command and then copy and pasted it, but you're right someone has added a new option and not documented it. How rude!

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -47,206 +36,145 @@ import java.nio.file.Path
...
+    override fun initLogging() = this.initLogging(cmdLineOptions.baseDirectory)
 
-    /**
-     * @return exit code based on the success of the node startup. This value is intended to be the exit code of the process.
-     */
     override fun runProgram(): Int {
-        val startTime = System.currentTimeMillis()
+        return when {
+            InitialRegistration.checkRegistrationMode(cmdLineOptions.baseDirectory) -> {
+                println("Node was started before with `--initial-registration`, but the registration was not completed.\nResuming registration.")
```
tlil: Rephrase given that this is now deprecated?

tlil: Ah, I see.. Further down.

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -46,89 +103,72 @@ class NodeCmdLineOptions {
...
+            hidden = true
     )
-    var networkRootTrustStorePassword: String? = null
+    var justGenerateNodeInfo: Boolean = false
 
     @Option(
-            names = ["--on-unknown-config-keys"],
-            description = ["How to behave on unknown node configuration. \${COMPLETION-CANDIDATES}"]
+            names = ["--just-generate-rpc-ssl-settings"],
+            description = ["DEPRECATED. Generates the SSL key and trust stores for a secure RPC connection."],
```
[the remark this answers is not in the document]
anthonykeenan: These help messages are no longer displayed, I've just added deprecated in the description in case a developer is looking at the code.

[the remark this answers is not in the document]
tlil: Ah, yeah. OK

node/src/main/kotlin/net/corda/node/internal/subcommands/GenerateRpcSslCerts.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private fun generateRpcSslCertificates(conf: NodeConfiguration) {
+        val (keyPair, cert) = createKeyPairAndSelfSignedTLSCertificate(conf.myLegalName.x500Principal)
+
+        val keyStorePath = conf.baseDirectory / "certificates" / "rpcsslkeystore.jks"
+        val trustStorePath = conf.baseDirectory / "certificates" / "export" / "rpcssltruststore.jks"
+
+        if (keyStorePath.exists() || trustStorePath.exists()) {
+            println("Found existing RPC SSL keystores. Command was already run. Exiting..")
```
tlil: Inconsistent use of: `Exiting..`, `Exiting` :) #nit

node/src/main/kotlin/net/corda/node/internal/subcommands/GenerateRpcSslCerts.kt
```diff
@@ -0,0 +1,103 @@
...
+                println("Not connected to console. Exiting")
+                exitProcess(1)
+            }
+            // Otherwise we can proceed normally.
+            else -> {
+                while (true) {
+                    val keystorePassword1 = console.readPassword("Enter the RPC keystore password => ")
+                    // TODO: consider adding a password strength policy.
+                    if (keystorePassword1.isEmpty()) {
+                        println("The RPC keystore password cannot be an empty String.")
```
tlil: s/String/string/

node/src/main/kotlin/net/corda/node/internal/subcommands/GenerateRpcSslCerts.kt
```diff
@@ -0,0 +1,103 @@
...
+                        continue
+                    }
+
+                    val keystorePassword2 = console.readPassword("Re-enter the RPC keystore password => ")
```
tlil: ` => ` -> `:`?

node/src/main/kotlin/net/corda/node/internal/subcommands/GenerateRpcSslCerts.kt
```diff
@@ -0,0 +1,103 @@
...
+                    if (!keystorePassword1.contentEquals(keystorePassword2)) {
+                        println("The RPC keystore passwords don't match.")
+                        continue
+                    }
+
+                    saveToKeyStore(keyStorePath, keyPair, cert, String(keystorePassword1), "rpcssl")
+                    println("The RPC keystore was saved to: $keyStorePath .")
```
tlil: Space between `$keyStorePath` and `.`

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistration.kt
```diff
@@ -0,0 +1,111 @@
...
+            // If the node was started with `--initial-registration`, create marker file.
+            // We do this here to ensure the marker is created even if parsing the args with NodeArgsParser fails.
+            val marker = baseDirectory / INITIAL_REGISTRATION_MARKER
+            if (!marker.exists()) {
+                return false
+            }
+            try {
+                marker.createFile()
+            } catch (e: Exception) {
+                logger.warn("Could not create marker file for `--initial-registration`.", e)
```
tlil: Refer to the sub-command here?

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -131,13 +136,13 @@ open class Node(configuration: NodeConfiguration,
         // TODO: make this configurable.
         const val MAX_RPC_MESSAGE_SIZE = 10485760
 
-        fun isValidJavaVersion(): Boolean {
+        fun isInvalidJavaVersion(): Boolean {
```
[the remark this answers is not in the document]
anthonykeenan: Fixed the test that I broke!

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
anthonykeenan: It was previously in the `NodeStartup`, and the `loadConfig` was passing the raw config file out, purely so it could be logged, which also didn't seem right to me. Couldn't really think of a better way of doing it but open to suggestions

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
anthonykeenan: Yes that's a lot better, I agree

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
anthonykeenan: Well, it's shared based on the way it works now, those options have some effect in all the different modes. It should be easy enough to create a seperate `SharedNotaryCmdLineOptions` class for any notary startup options. But you'll probably need to do some more refactoring anyway as the `NodeStartup` class depends on all of those options existing

anthonykeenan:
I think that's everything for now

anthonykeenan:
Just as a general point, would it be possible to have a test or two, preferably the version based tests, using an actual config file? I think that would aid understanding.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -57,22 +57,20 @@ abstract class NodeCliCommand(alias: String, description: String, val startup: N
 /** Main corda entry point. */
 open class NodeStartupCli : CordaCliWrapper("corda", "Runs a Corda Node") {
     val startup = NodeStartup()
-    private val networkCacheCli = ClearNetworkCache(startup)
-    private val justGenerateNodeInfoCli = GenerateNodeInfo(startup)
+    private val networkCacheCli = ClearNetworkCacheCli(startup)
+    private val justGenerateNodeInfoCli = GenerateNodeInfoCli(startup)
     private val justGenerateRpcSslCerts = GenerateRpcSslCertsCli(startup)
```
tlil: justGenerateRpcSslCerts -> justGenerateRpcSslCertsCli? :) @anthonykeenan

node/src/main/kotlin/net/corda/node/internal/subcommands/GenerateRpcSslCerts.kt
tlil: @anthonykeenan ^

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -414,12 +414,12 @@ interface NodeStartupLogging {
 
     fun <RESULT> attempt(action: () -> RESULT): Try<RESULT> = Try.on(action)
 
-    fun Exception.log(message: String? = this.message, print: (String?, Throwable) -> Unit = logger::error) = print("$message${this.message?.let { ": $it" } ?: ""}", this)
+    fun Exception.log(message: String? = this.message, exception: Throwable? = null, print: (String?, Throwable?) -> Unit = logger::error) = print("$this${exception?.message?.let { ": $it" } ?: ""}", exception)
```
tlil: Why `$this` there? That's the receiver of type `Exception`, and then you have an argument of type `Throwable` as well...?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -414,12 +414,12 @@ interface NodeStartupLogging {
...
 
     fun handleRegistrationError(error: Exception) {
         when (error) {
             is NodeRegistrationException -> error.log("Issue with Node registration: ${error.message}")
-            else -> error.log("Exception during node registration")
+            else -> error.log("Exception during node registration", error)
```
tlil: See above; you're essentially passing in `error` twice here :)

tlil:
>Replaced `--bootstrap-raft-cluster` flag with `bootstrap-raft-cluster` command
That flag was removed by N&P, right?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: So this is part of the changes that @sollecitom asked for.
The old logAsUnexpected function had an error parameter that defaulted to this, basically if the extra error parameter is supplied it adds it to the error log. But I agree it’s confusing - it took me most of Monday to work out what I’d originally done that was causing logging problems. 

However the $this at the beginning of the message should be $message, that’s a typo. I’ll fix it when I’m in.

anthonykeenan:
Yes the raft cluster bootstrapping has been deleted by N&P, I had some merge pain yesterday ;-)

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: Thinking about it I should probably get rid of the receiver parameter - it’s just confusing? Make it explicit which bits are getting logged.

tlil: Thanks.

Look at the usages further down, it seems like the `error` parameter is completely redundant though :-)

```kt
error.log("<message>", error)
```

`error` passed in twice. :-)

anthonykeenan: No it’s actually doing something - this is always not null, but error is sometimes null, so

error.log(‘message’, error)

Logs the message and the stacktrace and

error.log(message)

Just logs the message and

error.log()

Just logs the error message, but not the stack trace.

anthonykeenan: Possibly a logStackTrace Boolean is more sensible if there are no instances where the error is being overridden (I don’t think there are)

tlil: Yes, exactly  😝

docs/source/blob-inspector.rst
```diff
@@ -109,6 +109,10 @@ The blob inspector can be started with the following command-line options:
 * ``--schema``: Print the blob's schema first.
 * ``--verbose``, ``--log-to-console``, ``-v``: If set, prints logging to the console as well as to a file.
 * ``--logging-level=<loggingLevel>``: Enable logging at this level and higher. Possible values: ERROR, WARN, INFO, DEBUG, TRACE. Default: INFO.
-* ``--install-shell-extensions``: Install ``blob-inspector`` alias and auto completion for bash and zsh. See :doc:`cli-application-shell-extensions` for more info.
 * ``--help``, ``-h``: Show this help message and exit.
-* ``--version``, ``-V``: Print version information and exit.
\ No newline at end of file
+* ``--version``, ``-V``: Print version information and exit.
```
sollecitom: "Prints" .. and "exits"

docs/source/network-bootstrapper.rst
```diff
@@ -255,14 +255,19 @@ The network bootstrapper can be started with the following command-line options:
...
 * ``--dir=<dir>``: Root directory containing the node configuration files and CorDapp JARs that will form the test network.
   It may also contain existing node directories. Defaults to the current directory.
 * ``--no-copy``: Don't copy the CorDapp JARs into the nodes' "cordapps" directories.
 * ``--verbose``, ``--log-to-console``, ``-v``: If set, prints logging to the console as well as to a file.
 * ``--logging-level=<loggingLevel>``: Enable logging at this level and higher. Possible values: ERROR, WARN, INFO, DEBUG, TRACE. Default: INFO.
-* ``--install-shell-extensions``: Install ``bootstrapper`` alias and auto completion for bash and zsh. See :doc:`cli-application-shell-extensions` for more info.
 * ``--help``, ``-h``: Show this help message and exit.
-* ``--version``, ``-V``: Print version information and exit.
\ No newline at end of file
+* ``--version``, ``-V``: Print version information and exit.
```
sollecitom: "Prints" and "exits"

docs/source/network-bootstrapper.rst
```diff
@@ -255,14 +255,19 @@ The network bootstrapper can be started with the following command-line options:
...
+    bootstrapper [-hvV] [--no-copy] [--dir=<dir>] [--logging-level=<loggingLevel>]
+                 [--minimum-platform-version=<minimumPlatformVersion>] [COMMAND]
 
 * ``--dir=<dir>``: Root directory containing the node configuration files and CorDapp JARs that will form the test network.
   It may also contain existing node directories. Defaults to the current directory.
 * ``--no-copy``: Don't copy the CorDapp JARs into the nodes' "cordapps" directories.
 * ``--verbose``, ``--log-to-console``, ``-v``: If set, prints logging to the console as well as to a file.
 * ``--logging-level=<loggingLevel>``: Enable logging at this level and higher. Possible values: ERROR, WARN, INFO, DEBUG, TRACE. Default: INFO.
-* ``--install-shell-extensions``: Install ``bootstrapper`` alias and auto completion for bash and zsh. See :doc:`cli-application-shell-extensions` for more info.
 * ``--help``, ``-h``: Show this help message and exit.
```
sollecitom: "Shows" and "exits"

docs/source/network-bootstrapper.rst
```diff
@@ -255,14 +255,19 @@ The network bootstrapper can be started with the following command-line options:
...
-    bootstrapper [-hvV] [--install-shell-extensions] [--no-copy] [--dir=<dir>]
-                 [--logging-level=<loggingLevel>]
+    bootstrapper [-hvV] [--no-copy] [--dir=<dir>] [--logging-level=<loggingLevel>]
+                 [--minimum-platform-version=<minimumPlatformVersion>] [COMMAND]
 
 * ``--dir=<dir>``: Root directory containing the node configuration files and CorDapp JARs that will form the test network.
   It may also contain existing node directories. Defaults to the current directory.
 * ``--no-copy``: Don't copy the CorDapp JARs into the nodes' "cordapps" directories.
 * ``--verbose``, ``--log-to-console``, ``-v``: If set, prints logging to the console as well as to a file.
 * ``--logging-level=<loggingLevel>``: Enable logging at this level and higher. Possible values: ERROR, WARN, INFO, DEBUG, TRACE. Default: INFO.
```
sollecitom: "Enables"

docs/source/running-a-node.rst
```diff
@@ -48,25 +48,38 @@ Command-line options
...
+``clear-network-cache``: Clears local copy of network map, on node startup it will be restored from server or file system.
+
+``initial-registration``: Starts initial node registration with the compatibility zone to obtain a certificate from the Doorman.
+
+Parameters:
+
+* ``--network-root-truststore``, ``-t`` **required**: Network root trust store obtained from network operator.
+* ``--network-root-truststore-password``, ``-p``: Network root trust store password obtained from network operator.
+
+``generate-node-info``: Perform the node start-up tasks necessary to generate the nodeInfo file, save it to disk, then exit.
```
sollecitom: "Performs"... "saves" ... "exits"

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -27,6 +27,53 @@ class NodeCmdLineOptions {
...
 
+    @Option(
+            names = ["--on-unknown-config-keys"],
+            description = ["How to behave on unknown node configuration. \${COMPLETION-CANDIDATES}"]
+    )
+    var unknownConfigKeysPolicy: UnknownConfigKeysPolicy = UnknownConfigKeysPolicy.FAIL
+
+    @Option(
+            names = ["-d", "--dev-mode"],
+            description = ["Run the node in development mode. Unsafe for production."]
```
sollecitom: "Runs"

node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
sollecitom: Alright, let's keep it for now, as it will look different after I merge my changes.

node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -109,9 +110,13 @@ open class Node(configuration: NodeConfiguration,
             LoggerFactory.getLogger(loggerName).info(msg)
         }
 
+        fun printInRed(message: String) {
+            println("${ShellConstants.RED}$message${ShellConstants.RESET}")
```
sollecitom: Still not sure `RED` should be a ShellConstant. Also, might be worth to pull in the `Emoji.warningSign` padding, along with "ATTENTION:" into a warn function.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -45,201 +36,145 @@ import java.nio.file.Path
...
-            return
-        }
+    open fun createNode(conf: NodeConfiguration, versionInfo: VersionInfo): Node = Node(conf, versionInfo)
 
-        if (conf.devMode) {
+    fun startNode(node: Node, startTime: Long) {
+        if (node.configuration.devMode) {
             Emoji.renderIfSupported {
-                Node.printWarning("This node is running in developer mode! ${Emoji.developer} This is not safe for production deployment.")
+                Node.printWarning("This node is running in development mode! ${Emoji.developer} This is not safe for production deployment.")
```
sollecitom: This we change this in the docs as well?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -577,3 +396,64 @@ open class NodeStartup : CordaCliWrapper("corda", "Runs a Corda Node") {
...
+/** Provide some common logging methods for node startup commands. */
+interface NodeStartupLogging {
+    companion object {
+        val logger by lazy { contextLogger() }
+        val startupErrors = setOf(MultipleCordappsForFlowException::class, CheckpointIncompatibleException::class, AddressBindingException::class, NetworkParametersReader::class, DatabaseIncompatibleException::class)
+    }
+
+    fun <RESULT> attempt(action: () -> RESULT): Try<RESULT> = Try.on(action)
+
+    fun Exception.log(message: String? = this.message, logStackTrace: Boolean = false, print: (String?, Throwable?) -> Unit = logger::error) = print("$message${this.message?.let { ": $it" } ?: ""}", if (logStackTrace) this else null)
```
sollecitom: Can we remove `logStackTrace` and always do pass the throwable? You can change `print: (String?, Throwable?)` to be `print: (String?, Throwable)`.
This is taken care by `log4j2.xml` now for the console, so we can just remove this check, which is creating issues in some cases.

node/src/main/kotlin/net/corda/node/internal/Node.kt
anthonykeenan: Is it not a constant of the shell? Is it more a unicode constant? I'm not really sure what to call it. And if we're moving the warning sign and the attention into the `printInRed` function, then it's a `printWarning` function like it was originally surely?

tlil: It is a shell constant. I'm happy with it being like it is at the moment.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: I tried changing it to `fun Exception.log(message: String? = null, print: (String?, Throwable) -> Unit = logger::error) = print(message?.let { "$it : $this.message" } ?: this.message, this)` but it causes tests to fail in `FlowCheckpointVersionNodeStartupCheckTest` because stack traces are being logged to the console. I've also tried modifying the `log4j2.xml` to remove the throwable from the console logs but that also doesn't seem to help.

docs/source/blob-inspector.rst
```diff
@@ -108,9 +108,9 @@ The blob inspector can be started with the following command-line options:
 * ``--full-parties``: Display the owningKey and certPath properties of Party and PartyAndReference objects respectively.
 * ``--schema``: Print the blob's schema first.
 * ``--verbose``, ``--log-to-console``, ``-v``: If set, prints logging to the console as well as to a file.
-* ``--logging-level=<loggingLevel>``: Enable logging at this level and higher. Possible values: ERROR, WARN, INFO, DEBUG, TRACE. Default: INFO.
-* ``--help``, ``-h``: Show this help message and exit.
-* ``--version``, ``-V``: Print version information and exit.
+* ``--logging-level=<loggingLevel>``: Enables logging at this level and higher. Possible values: ERROR, WARN, INFO, DEBUG, TRACE. Default: INFO.
```
tlil: This is now inconsistent with things like `--schema` and `--no-copy` etc. further up / down. Sorry to be a pain :P

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -421,18 +421,18 @@ interface NodeStartupLogging {
     fun handleStartError(error: Exception) {
         when {
             error.isExpectedWhenStartingNode() -> error.log()
-            error is CouldNotCreateDataSourceException -> error.log(logStackTrace = true)
+            error is CouldNotCreateDataSourceException -> error.log()
```
tlil: This will now print `An error occurred` to the console, instead of the error message from the thrown exception - so seems like we're loosing some context for the console user here....

docs/source/blob-inspector.rst
anthonykeenan: ``--schema`` and ``--no-copy`` read fine to me but I've corrected a couple of others.

anthonykeenan: *googles what the imperative mood is*

anthonykeenan:
@tlil I've reverted the logging changes (which weren't part of the original ticket anyway) as they obviously need more thought, and I've reverted the doc changes.

docs/source/running-a-node.rst
```diff
@@ -48,25 +48,38 @@ Command-line options
...
 * ``--verbose``, ``--log-to-console``, ``-v``: If set, prints logging to the console as well as to a file.
 * ``--logging-level=<loggingLevel>``: Enable logging at this level and higher. Possible values: ERROR, WARN, INFO, DEBUG, TRACE. Default: INFO.
-* ``--install-shell-extensions``: Install ``corda`` alias and auto completion for bash and zsh. See :doc:`cli-application-shell-extensions` for more info.
 * ``--help``, ``-h``: Show this help message and exit.
 * ``--version``, ``-V``: Print version information and exit.
 
+Sub-commands
+^^^^^^^^^^^^
+
+``bootstrap-raft-cluster``: Bootstraps Raft cluster. The node forms a single node cluster (ignoring otherwise configured peer
```
tlil: Let's remove this in a separate PR.

PR #4099, opened 2018-10-19
[CORDA-2011] [CORDA-2057] CryptoService interface and BC HSM simulation

kchalkias:
Goal: allow support for `CryptoService` implementations (i.e., HSM) and minimise dependency on Java Keystores for legal identity keys. Among the others, this PR includes the following major changes/updates:
- a `CryptoService` interface as a new module to provide the basic cryptographic operations an HSM implementation should support:  `generateKeyPair`,  `containsKey`, `getPublicKey`, `sign`, `getSigner`. ~The sole dependency of this module is BouncyCastle (due to `ContentSigner` requirement), thus there are no dependencies on corda-core and thus allowing lighter implementation jars.~
It's now part of `nodeapi.internal`.
- a `BCCryptoService` basic implementation of a `CryptoService` that uses BouncyCastle for cryptographic operations and a Java `KeyStore` to store private keys. This is the default if no CryptoService is defined in NodeConfiguration.
- a `BasicHSMKeyManagementService` that is loaded with a `CryptoService` for legal identity key. Anonymous keys are out of scope for this PR so we handle them the same way as before (without CryptoService).
- `NodeConfiguration` optionally accepts a `cryptoServiceName` for the desired CryptoService implementation and a `cryptoServiceConf` to locate the `CryptoService` configuration file.
- `SupportedCryptoServices` is an `enum` that enlists supported and tested `CryptoService` implementations, at the moment only BC_SIMPLE (BouncyCastle using Java KeyStores).
- `AliasPrivateKey` a `PrivateKey` implementation that actually carries the alias only and not private key material. This was required for minimal changes in the codebase that used to return `KeyPair`s.
- Major refactoring on many classes, mainly `AbstractNode` and `NetworkRegistrationHelper`.
- `signingCertStore` is still there due to backwards compatibility purposes (we want nodes to just import their old KeyStore and work out of the box). However, the main role of the `signingCertStore` is now to store certPaths; the only exception is BCCryptoService which reuses it to also store keys.
- Some Dev Identity generation functions do still use keystores internally (on purpose to keep this PR as small as possible) - I will update on upcoming PRs.
- Same for extensive kdocs, I will add a lot more when the final version is ready (after original review).

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
@@ -199,6 +199,12 @@ object Crypto {
             + signatureSchemeMap.values.map { Pair(it.signatureOID, it) })
             .toMap()
 
+    /**
+     * Map of supported digital signature schemes associated by [SignatureScheme.schemeNumberID].
+     * SchemeNumberID is the scheme identifier attached to [SignatureMetadata].
+     */
+    private val signatureSchemeNumberIDMap: Map<Int, SignatureScheme> = Crypto.supportedSignatureSchemes().associateBy { it.schemeNumberID }
+
     @JvmStatic
     fun supportedSignatureSchemes(): List<SignatureScheme> = ArrayList(signatureSchemeMap.values)
 
@@ -222,6 +228,13 @@ object Crypto {
     fun findSignatureScheme(algorithm: AlgorithmIdentifier): SignatureScheme {
         return algorithmMap[normaliseAlgorithmIdentifier(algorithm)]
                 ?: throw IllegalArgumentException("Unrecognised algorithm: ${algorithm.algorithm.id}")
+    }
+
+    /** Find [SignatureScheme] by platform specific schemeNumberID. */
+    @JvmStatic
+    fun findSignatureScheme(schemeNumberID: Int): SignatureScheme {
+        return signatureSchemeNumberIDMap[schemeNumberID]
+                ?: throw IllegalArgumentException("Unsupported key/algorithm for schemeNumberID: $schemeNumberID")
     }
 
     /**
@@ -458,9 +471,14 @@ object Crypto {
     @JvmStatic
     @Throws(InvalidKeyException::class, SignatureException::class)
     fun doSign(keyPair: KeyPair, signableData: SignableData): TransactionSignature {
-        val sigKey: SignatureScheme = findSignatureScheme(keyPair.private)
-        val sigMetaData: SignatureScheme = findSignatureScheme(keyPair.public)
-        require(sigKey == sigMetaData) {
+        val sigKey: SignatureScheme = Crypto.findSignatureScheme(keyPair.private)
+        val sigMetaData: SignatureScheme = Crypto.findSignatureScheme(signableData.signatureMetadata.schemeNumberID)
+        // Special handling if the advertised SignatureScheme is CompositeKey.
+        // TODO fix notaries that advertise [CompositeKey] in their signature Metadata. Currently, clustered notary nodes
+        //      mention Crypto.COMPOSITE_KEY in their SignatureMetadata, but they are actually signing with a leaf-key
+        //      (and if they refer to it as a Composite key, then we lose info about the actual type of their signing key).
+        //      In short, their metadata should be the leaf key-type, until we support CompositeKey signatures.
+        require(sigKey == sigMetaData || sigMetaData == Crypto.COMPOSITE_KEY) {
             "Metadata schemeCodeName: ${sigMetaData.schemeCodeName} is not aligned with the key type: ${sigKey.schemeCodeName}."
         }
         val signatureBytes = doSign(sigKey.schemeCodeName, keyPair.private, signableData.serialize().bytes)
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStore.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStore.kt
@@ -1,5 +1,6 @@
 package net.corda.nodeapi.internal.config
 
+import net.corda.core.crypto.AliasPrivateKey
 import net.corda.core.internal.outputStream
 import net.corda.nodeapi.internal.crypto.X509KeyStore
 import net.corda.nodeapi.internal.crypto.X509Utilities
@@ -8,6 +9,7 @@ import java.io.InputStream
 import java.io.OutputStream
 import java.nio.file.OpenOption
 import java.nio.file.Path
+import java.security.PrivateKey
 import java.security.cert.X509Certificate
 
 interface CertificateStore : Iterable<Pair<String, X509Certificate>> {
@@ -42,7 +44,6 @@ interface CertificateStore : Iterable<Pair<String, X509Certificate>> {
     }
 
     operator fun set(alias: String, certificate: X509Certificate) {
-
         update {
             internal.addOrReplaceCertificate(X509Utilities.CORDA_ROOT_CA, certificate)
         }
@@ -64,7 +65,6 @@ interface CertificateStore : Iterable<Pair<String, X509Certificate>> {
      * @throws IllegalArgumentException if no certificate for the alias is found, or if the certificate is not an [X509Certificate].
      */
     operator fun get(alias: String): X509Certificate {
-
         return query {
             getCertificate(alias)
         }
@@ -77,6 +77,21 @@ interface CertificateStore : Iterable<Pair<String, X509Certificate>> {
         certificateStore.update {
             this@CertificateStore.forEach(::setCertificate)
         }
+    }
+
+    fun setCertPathOnly(alias: String, certificates: List<X509Certificate>) {
+        // In case CryptoService and CertificateStore share the same KeyStore (i.e., when BCCryptoService is used),
+        // extract the existing key from the Keystore and store it again along with the new certificate chain.
+        // This is because KeyStores do not support updateKeyEntry and thus we cannot update the certificate chain
+        // without overriding the key entry.
+        // Note that if the given alias already exists, the keystore information associated with it
+        // is overridden by the given key (and associated certificate chain).
+        val privateKey: PrivateKey = if (this.contains(alias)) {
+            this.value.getPrivateKey(alias, entryPassword)
+        } else {
+            AliasPrivateKey(alias)
+        }
+        this.value.setPrivateKey(alias, privateKey, certificates, entryPassword)
     }
 }
 
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStore.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStore.kt
@@ -1,6 +1,6 @@
 package net.corda.nodeapi.internal.config
 
-import net.corda.core.crypto.AliasPrivateKey
+import net.corda.core.crypto.internal.AliasPrivateKey
 import net.corda.core.internal.outputStream
 import net.corda.nodeapi.internal.crypto.X509KeyStore
 import net.corda.nodeapi.internal.crypto.X509Utilities
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -9,21 +9,19 @@ import net.corda.confidential.SwapIdentitiesHandler
 import net.corda.core.CordaException
 import net.corda.core.concurrent.CordaFuture
 import net.corda.core.context.InvocationContext
+import net.corda.core.crypto.AliasPrivateKey
+import net.corda.core.crypto.DigitalSignature
 import net.corda.core.crypto.SecureHash
 import net.corda.core.crypto.newSecureRandom
-import net.corda.core.crypto.sign
 import net.corda.core.flows.*
 import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.identity.Party
 import net.corda.core.identity.PartyAndCertificate
-import net.corda.core.internal.FlowStateMachine
-import net.corda.core.internal.NamedCacheFactory
-import net.corda.core.internal.VisibleForTesting
+import net.corda.core.internal.*
 import net.corda.core.internal.concurrent.map
 import net.corda.core.internal.concurrent.openFuture
 import net.corda.core.internal.notary.NotaryService
-import net.corda.core.internal.uncheckedCast
 import net.corda.core.messaging.*
 import net.corda.core.node.*
 import net.corda.core.node.services.*
@@ -56,8 +54,9 @@ import net.corda.node.services.config.shouldInitCrashShell
 import net.corda.node.services.events.NodeSchedulerService
 import net.corda.node.services.events.ScheduledActivityObserver
 import net.corda.node.services.identity.PersistentIdentityService
+import net.corda.node.services.keys.BasicHSMKeyManagementService
 import net.corda.node.services.keys.KeyManagementServiceInternal
-import net.corda.node.services.keys.PersistentKeyManagementService
+import net.corda.node.internal.cryptoservice.BCCryptoService
 import net.corda.node.services.messaging.DeduplicationHandler
 import net.corda.node.services.messaging.MessagingService
 import net.corda.node.services.network.NetworkMapClient
@@ -75,14 +74,15 @@ import net.corda.node.utilities.*
 import net.corda.nodeapi.internal.NodeInfoAndSigned
 import net.corda.nodeapi.internal.SignedNodeInfo
 import net.corda.nodeapi.internal.config.CertificateStore
+import net.corda.nodeapi.internal.crypto.CertificateType
 import net.corda.nodeapi.internal.crypto.X509Utilities
 import net.corda.nodeapi.internal.crypto.X509Utilities.CORDA_CLIENT_CA
 import net.corda.nodeapi.internal.crypto.X509Utilities.CORDA_CLIENT_TLS
 import net.corda.nodeapi.internal.crypto.X509Utilities.CORDA_ROOT_CA
+import net.corda.nodeapi.internal.crypto.X509Utilities.DEFAULT_VALIDITY_WINDOW
 import net.corda.nodeapi.internal.crypto.X509Utilities.DISTRIBUTED_NOTARY_ALIAS_PREFIX
 import net.corda.nodeapi.internal.crypto.X509Utilities.NODE_IDENTITY_ALIAS_PREFIX
 import net.corda.nodeapi.internal.persistence.*
-import net.corda.nodeapi.internal.storeLegalIdentity
 import net.corda.tools.shell.InteractiveShell
 import org.apache.activemq.artemis.utils.ReusableLatch
 import org.hibernate.type.descriptor.java.JavaTypeDescriptorRegistry
@@ -90,7 +90,6 @@ import org.slf4j.Logger
 import rx.Observable
 import rx.Scheduler
 import java.io.IOException
-import java.lang.UnsupportedOperationException
 import java.lang.reflect.InvocationTargetException
 import java.nio.file.Paths
 import java.security.KeyPair
@@ -167,6 +166,7 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
     val transactionStorage = makeTransactionStorage(configuration.transactionCacheSizeBytes).tokenize()
     val networkMapClient: NetworkMapClient? = configuration.networkServices?.let { NetworkMapClient(it.networkMapURL, versionInfo) }
     val attachments = NodeAttachmentService(metricRegistry, cacheFactory, database).tokenize()
+    val cryptoService = configuration.makeCryptoService()
     val cordappProvider = CordappProviderImpl(cordappLoader, CordappConfigFileProvider(configuration.cordappDirectories), attachments).tokenize()
     @Suppress("LeakingThis")
     val keyManagementService = makeKeyManagementService(identityService).tokenize()
@@ -262,6 +262,11 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
     private fun initKeyStores(): X509Certificate {
         if (configuration.devMode) {
             configuration.configureWithDevSSLCertificate()
+            // configureWithDevSSLCertificate is a devMode process that writes directly to keystore files, so
+            // we should re-synchronise BCCryptoService with the updated keystore file.
+            if (cryptoService is BCCryptoService) {
+                cryptoService.resyncKeystore()
+            }
         }
         return validateKeyStores()
     }
@@ -300,7 +305,6 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
         log.info("Node starting up ...")
 
         val trustRoot = initKeyStores()
-        val nodeCa = configuration.signingCertificateStore.get()[CORDA_CLIENT_CA]
         initialiseJVMAgents()
 
         schemaService.mappedSchemasWarnings().forEach {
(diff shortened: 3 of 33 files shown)


crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
```diff
@@ -0,0 +1,15 @@
...
+import java.security.PublicKey
+
+interface CryptoService {
+
+    /** schemeNumberID is Corda specific. */
+    fun generateKeyPair(alias: String, schemeNumberID: Int): PublicKey
+    fun containsKey(alias: String): Boolean
+    fun getPublicKey(alias: String): PublicKey?
+    fun sign(alias: String, data: ByteArray): ByteArray
+    fun signer(alias: String): ContentSigner
```
tlil: `getSigner` to be consistent with `getPublicKey`?

tlil: `getSigner` to be consistent with `getPublicKey`?

tlil: `getSigner` to be consistent with the naming of the other methods, e.g., `getPublicKey` ?

tlil: `getSigner` to be consistent with the naming of the other methods, e.g., `getPublicKey` ?

tlil: `getSigner` to be consistent with the naming of the other methods, e.g., `getPublicKey` ?

mikehearn:
Let me know when this is ready for me to take a look!

kchalkias:
@tlil and @mikehearn I just managed to rebase and fix conflicts, there were a few requiring extra caution (Victor's extra entryPassword for keystores, Andrius' extract notary module that affected obtain identity and then when I pushed and thought I was ready, Anthony's NodeStartup subcommands cleanup today required extra conflict resolution. I will get back to you for the 1st round of reviews when TC eventually gets green.

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
```diff
@@ -0,0 +1,37 @@
+package net.corda.cryptoservice
+
+import org.bouncycastle.operator.ContentSigner
+import java.security.KeyPair
+import java.security.PublicKey
+
+interface CryptoService {
```
mikehearn: Is this intended to be new public API? If so, it should be under net.corda.core and have interface-level kdocs, and be marked `@DoNotImplement` or be a base class.

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Generate and store a new [KeyPair].
+     * Note that schemeNumberID is Corda specific. Cross-check with the network operator for supported schemeNumberID
+     * and their corresponding signature schemes. The main reason for using schemeNumberID and not algorithm OIDs is
```
mikehearn: Are "scheme numbers" already a part of the public API? To observe for future, we do own our own OID and can put things under it if we want to get ahead of standardisation.

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
```diff
@@ -0,0 +1,37 @@
...
+     * because some schemes might not be standardised and thus an official OID might for this scheme not exist yet.
+     *
+     * @return the [PublicKey] of the generated [KeyPair].
+     */
+    fun generateKeyPair(alias: String, schemeNumberID: Int): PublicKey
+
+    /** Check if this [CryptoService] has a private key record for the input alias. */
```
mikehearn: record_ed_?

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
```diff
@@ -0,0 +1,37 @@
...
+    fun containsKey(alias: String): Boolean
+
+    /**
+     * @return [PublicKey] of the input alias or null if it doesn't exist.
+     */
+    fun getPublicKey(alias: String): PublicKey?
+
+    /**
+     * Sign a [ByteArray] using the private key identified by the input alias.
+     * @return the signature bytes.
```
mikehearn: Is the returned signature in any particular format or does it depend on the backend? ASN.1?

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
```diff
@@ -0,0 +1,37 @@
...
+     */
+    fun sign(alias: String, data: ByteArray): ByteArray
+
+    /**
+     * @return [ContentSigner] for the key identified by the input alias.
```
mikehearn: Please use "Returns" rather than `@return` here:

https://docs.corda.net/head/codestyle.html#comments

(second bullet in the second list). Annotation only kdocs/javadocs don't render that great.

core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
```diff
@@ -458,9 +471,11 @@ object Crypto {
     @JvmStatic
     @Throws(InvalidKeyException::class, SignatureException::class)
     fun doSign(keyPair: KeyPair, signableData: SignableData): TransactionSignature {
-        val sigKey: SignatureScheme = findSignatureScheme(keyPair.private)
-        val sigMetaData: SignatureScheme = findSignatureScheme(keyPair.public)
-        require(sigKey == sigMetaData) {
+        val sigKey: SignatureScheme = Crypto.findSignatureScheme(keyPair.private)
+        val sigMetaData: SignatureScheme = Crypto.findSignatureScheme(signableData.signatureMetadata.schemeNumberID)
+        // Special handling if the advertised SignatureScheme is CompositeKey.
+        // TODO fix notaries that advertise [CompositeKey] in their signature Metadata.
```
mikehearn: What's this TODO about? Fix how? Aren't notaries meant to use composite keys? I could use more explanation in the code comments here.

core/src/main/kotlin/net/corda/core/crypto/NullKeys.kt
```diff
@@ -3,6 +3,7 @@ package net.corda.core.crypto
 import net.corda.core.KeepForDJVM
 import net.corda.core.identity.AnonymousParty
 import net.corda.core.serialization.CordaSerializable
+import java.security.PrivateKey
```
mikehearn: Useless import?

core/src/main/kotlin/net/corda/core/crypto/PlatformContentSigner.kt
```diff
@@ -0,0 +1,25 @@
+package net.corda.core.crypto
+
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier
+import org.bouncycastle.operator.ContentSigner
+import java.io.ByteArrayOutputStream
+import java.io.OutputStream
+import java.security.PrivateKey
+
+class PlatformContentSigner(private val privateKey: PrivateKey) : ContentSigner {
```
mikehearn: If this is meant to be internal it should go into an internal package namespace, if it's meant to be public please fully document. But it looks like it can be internal.

core/src/main/kotlin/net/corda/core/crypto/PlatformContentSigner.kt
```diff
@@ -0,0 +1,25 @@
...
+    override fun getAlgorithmIdentifier(): AlgorithmIdentifier {
+        return signatureOID
+    }
+
+    override fun getOutputStream(): OutputStream {
+        return outputStream
+    }
+
+    override fun getSignature(): ByteArray {
+        return Crypto.doSign(privateKey, outputStream.toByteArray())
```
mikehearn: This is a weird pattern. Don't you need to flush or reset the stream in some manner? Or is this object single use - if so maybe add assertions to ensure that is the case and _document_

node-api/src/main/kotlin/net/corda/nodeapi/internal/config/CertificateStore.kt
```diff
@@ -78,6 +77,16 @@ interface CertificateStore : Iterable<Pair<String, X509Certificate>> {
             this@CertificateStore.forEach(::setCertificate)
         }
     }
+
+    fun setCertPathOnly(alias: String, certificates: List<X509Certificate>) {
+        // In case CryptoService and CertificateStore share the same KeyStore, extract and store the key again.
+        val privateKey = if (this.contains(alias)) {
+            this.value.getPrivateKey(alias, entryPassword)
+        } else {
+            DummyKeysAndCerts.DUMMY_ECDSAR1_KEYPAIR.private
```
mikehearn: I could use more comments here - it isn't clear to me at the moment what this is about or why dummy keys are suddenly appearing outside of test packages. Can you add an explicit type annotation to privateKey also? (alt-enter, x, enter)

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/DummyKeysAndCerts.kt
```diff
@@ -0,0 +1,11 @@
+package net.corda.nodeapi.internal.crypto
+
+import net.corda.core.crypto.Crypto
+import net.corda.nodeapi.internal.crypto.X509Utilities.createSelfSignedCACertificate
+import java.math.BigInteger
+import javax.security.auth.x500.X500Principal
+
+object DummyKeysAndCerts {
```
mikehearn: What's this about? KDocs!

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -262,7 +264,11 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
 
     private fun initKeyStores(): X509Certificate {
         if (configuration.devMode) {
+            // TODO use CryptoService in devMode as well, at the moment we reuse the signingCertificateStore to generate dev keys.
```
mikehearn: Slightly confusing comment, cryptoService is referred to just two lines below.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -780,62 +788,93 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
                                                  nodeInfo: NodeInfo,
                                                  myNotaryIdentity: PartyAndCertificate?,
                                                  networkParameters: NetworkParameters)
-
+    /**
+     * Loads or generates the node's legal identity and key-pair.
+     * Note that obtainIdentity returns a KeyPair with an [AliasPrivateKey].
+     */
```
mikehearn: This comment is now duplicating the kdoc below - copy/paste error?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -780,62 +788,93 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
...
     private fun obtainIdentity(): Pair<PartyAndCertificate, KeyPair> {
-        val keyStore = configuration.signingCertificateStore.get()
-        val legalName = configuration.myLegalName
-
-        // TODO: Integrate with Key management service?
         val privateKeyAlias = "$NODE_IDENTITY_ALIAS_PREFIX-private-key"
-        if (privateKeyAlias !in keyStore) {
+
+        if (!cryptoService.containsKey(privateKeyAlias)) {
             log.info("$privateKeyAlias not found in key store, generating fresh key!")
```
mikehearn: I wonder if we should be explicitly checking here that we're doing `--initial-registration` as it can be dangerous to just randomly create new keys outside of the onboarding/setup process.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -78,6 +81,11 @@ interface NodeConfiguration {
     val cordappDirectories: List<Path>
     val flowOverrides: FlowOverrideConfig?
 
+    // TODO At the moment this is just an identifier for the desired CryptoService engine. Consider using a classname to
+    //      to allow for pluggable implementations.
+    val cryptoServiceName: SupportedCryptoServices?
+    val cryptoServiceConf: String? // Location for the cryptoService conf file.
```
mikehearn: Do we really need a separate config file? Is this something forced on us by HSM vendors?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -96,6 +104,14 @@ interface NodeConfiguration {
 
         val defaultJmxReporterType = JmxReporterType.JOLOKIA
     }
+
+    // TODO remove this when we use classname instead of cryptoServiceName enum.
```
mikehearn: Let's remove this TODO, it's unclear we'll ever open this up.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -272,6 +290,17 @@ data class NodeConfigurationImpl(
         return errors
     }
 
+    private fun validateCryptoService(): List<String> {
+        val errors = mutableListOf<String>()
+        if (cryptoServiceName != null) {
```
mikehearn: Hm? So you can't specify BC_SIMPLE as an explicit config at all?

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -272,6 +290,17 @@ data class NodeConfigurationImpl(
...
+            errors += "cryptoServiceName $cryptoServiceName is not supported"
+        }
+        if (cryptoServiceName == null && cryptoServiceConf != null) {
```
mikehearn: This is hard-coding knowledge about the BC_SIMPLE backend, right?

node/src/main/kotlin/net/corda/node/services/keys/cryptoServices/AliasPrivateKey.kt
```diff
@@ -0,0 +1,10 @@
+package net.corda.node.services.keys.cryptoServices
```
mikehearn: Lower case package names please. Also do we need this new package?

node/src/main/kotlin/net/corda/node/services/keys/cryptoServices/AliasPrivateKey.kt
```diff
@@ -0,0 +1,10 @@
...
+
+import java.security.PrivateKey
+
+/** [PrivateKey] wrapper to just store the alias of a private key. */
```
mikehearn: Could use a longer explanation of why it's necessary/what it's about. Not many people have worked with HSMs.

node/src/main/kotlin/net/corda/node/services/keys/cryptoServices/AliasPrivateKey.kt
```diff
@@ -0,0 +1,10 @@
...
+class AliasPrivateKey(val alias: String): PrivateKey {
+    override fun getAlgorithm() = "ALIAS"
+    override fun getEncoded() = ByteArray(0)
```
mikehearn: Should we be throwing an exception here rather than returning an empty byte array? I can imagine callers would be confused by that.

node/src/main/kotlin/net/corda/node/services/keys/cryptoServices/BCCryptoService.kt
```diff
@@ -0,0 +1,66 @@
...
+ * Basic implementation of a [CryptoService] that uses BouncyCastle for cryptographic operations
+ * and a Java KeyStore to store private keys.
+ */
+class BCCryptoService(private val nodeConf: NodeConfiguration) : CryptoService {
+
+    // TODO check if keyStore exists.
+    // TODO make it work with nodeConf.cryptoServiceConf (if it exists). I.e., read the signingCertificateStore
+    //      keyStore file name from there.
+    // TODO make it private when E2ETestKeyManagementService does not require direct access to the private key.
+    internal var certificateStore = nodeConf.signingCertificateStore.get(true)
```
mikehearn: It's unclear what the type of certificateStore is here - add an explicit type (alt-enter, x, enter)

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
kchalkias: As mentioned in the comments of this PR, I wanted to minimise dependencies in Corda-core, because actually you don't need it.

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
kchalkias: This is aligned with the Signature metadata, which is not the OID. Otherwise we have to change this everywhere. Also, we might decide in the future that RSA 2048 and RSA3072 are different schemes.

node/src/main/kotlin/net/corda/node/services/keys/cryptoServices/AliasPrivateKey.kt
kchalkias: I've explained why. The sole dependency of this module is BouncyCastle (due to ContentSigner requirement), thus, there are no dependencies on corda-core and this will allow for lighter implementation jars.

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
kchalkias: Correct, for all Java's signature byte, the format of the signature depends on the underlying
signature scheme.

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
kchalkias: I can even remove dependencies in BC (only needed for ContentSigner), by creating a custom ContentSigner object, so all HSM jars will be super light.

kchalkias: and why @DoNotImplement?, we want people to implement it, no? but we can only accept it in the future after we thoroughly run all the required tests.

core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
kchalkias: This is a bug we have (I had a related Metadata PR - we've discussed about this in the past, it's parked until this CryptoService is merged). In short, yes notaries do advertise their cluster's CompositeKey, but each one's signature (along with it's metadata) corresponds to the leaf-key. We still don't utilise the CompositeSignature object.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
kchalkias: I've mentioned this to @tlil a while ago as well. I didn't change the logic in this PR, iiuc this is here to speed up the boarding process (and I don't agree either).

core/src/main/kotlin/net/corda/core/crypto/PlatformContentSigner.kt
kchalkias: Indeed, this is one time, similarly to `ContentSignerBuilder` and others. Hm, I will probably remove this class entirely though and create one using a Builder.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
kchalkias: Correct, it refers to the fact that `configureDevKeyAndTrustStores` does not take cryptoService as input, but there is a similar TODO comment inside the aforementioned method, so it's fine to remove this one.

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
kchalkias: Hm.. leftover due to initial tests. I will rewrite this one.

node/src/main/kotlin/net/corda/node/services/keys/cryptoServices/AliasPrivateKey.kt
kchalkias: I tried this, then if we ever store it to a Keystore it will break as it internally invokes this method. I will change it though as atm we don't store it to Keystores and if we did usually Keystores check for PKCS8 format.

core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
mikehearn: OK, can you put that explanation in the comments themselves?

node/src/main/kotlin/net/corda/node/services/keys/BasicHSMKeyManagementService.kt
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * A persistent re-implementation of [E2ETestKeyManagementService] to support CryptoService for initial keys and
+ * database storage for anonymous fresh keys.
+ *
+ * This is not the long-term implementation.  See the list of items in the above class.
+ *
+ * This class needs database transactions to be in-flight during method calls and init.
+ */
+class BasicHSMKeyManagementService(cacheFactory: NamedCacheFactory, val identityService: PersistentIdentityService,
```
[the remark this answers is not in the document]
kchalkias: As I explain, anonymous keys are out of scope for this PR due to increased cost and complexity (i.e., some HSMs only support a few hundreds of thousands stored keys)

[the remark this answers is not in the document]
gendal: 👍

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
kchalkias: it applies to any CryptoService. We cannot define a `cryptoServiceConf` without a cryptoServiceName. Note that BCCryptoService reuses the `signingCertificateStore`, so it does not need any extra configuration file anyway.

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
```diff
@@ -0,0 +1,38 @@
...
+     * Returns the [PublicKey] of the input alias or null if it doesn't exist.
+     */
+    fun getPublicKey(alias: String): PublicKey?
+
+    /**
+     * Sign a [ByteArray] using the private key identified by the input alias.
+     * @return the signature bytes whose format depends on the underlying signature scheme and it should
+     * be Java BouncyCastle compatible (i.e., ASN.1 DER-encoded for ECDSA).
+     */
+    fun sign(alias: String, data: ByteArray): ByteArray
```
florian-f: Using plain old file-based KeyStores, it is reasonable to assume that this operation will succeed in a timely manner. Now, we have different failure modes. Signing can take significantly longer, due to e.g. network latencies, signing might be (temporarily) not possible due to connectivity issues/ timeouts. Where and how do we address these problems best?
Example: We want to sign a transaction, but the (cloud-based) HSM of our node temporarily responds with 5xx. Do we want the signing operation to fail hard, should we retry with exponential back-off?
Another example: We want to sign a transaction, but the (cloud-based) HSM of our node is down, and does not even respond with an HTTP error code. Do we want to wait for a timeout to occur? Should we block until that happens?
@mikehearn what do you think?

kchalkias: Btw all key related methods in `KeyManagementService` are `@Suspendable` anyway. Defining a timeout policy is a different thing though. Wrt retry, IMO this should be implemented internally to each `CryptoServiceImpl` because there might be different requirements/guidelines per vendor (i.e., they can only accept certain requests per second - especially remote HSMs).

crypto-service/src/main/kotlin/net/corda/cryptoservice/CryptoService.kt
kchalkias: Moved it to `nodeapi.internal` 🗸

node-api/src/main/kotlin/net/corda/nodeapi/internal/crypto/DummyKeysAndCerts.kt
mikehearn: OK, so maybe these should be called NOT_YET_REGISTERED_MARKER_KEY or something more clear?

PR #4100, opened 2018-10-21
CORDA-1621: The finance CorDapp uses the app config feature rather than the node's config

shamsasari:


shamsasari:
The behave framework needs updating but it's very unclear how to proceed. I've added TODOs, but if you feel they should be tackled in this PR then please advise to how to proceed.

PR #4103, opened 2018-10-22
ENT-1906: Allow sandboxes to share a parent classloader.

chrisr3:
Allow `deterministic-rt.jar` and the template Java classes to be loaded into a parent `SandboxClassLoader` that can be shared by multiple sandboxes.

Child sandbox configurations inherit the same whitelist and pinned classes as the parent.

This has required the `System.objectHashCodes` and `String.INTERNAL` data structures to be migrated to `SandboxRuntimeContext` so that each sandbox still has its own private copy. It also means that the `DJVM` class must now fetch the correct classloader from `SandboxRuntimeContext` in order to locate user-supplied `Enum` and `Throwable` classes.

Should any other "static state" still exist in the Java API classes then would also be shareable among the child classloaders. But this change has knocked over 2 minutes off the unit test times.

Note that the `SandboxClassLoader` contains a cache of the sandboxed byte-code:
```kotlin
private val loadedClasses = mutableMapOf<String, LoadedClass>()
```
It wouldn't be difficult to "replay" this cache into a new `SandboxClassLoader` instance, c.f..
```kotlin
for (loadedClass in loadedClasses) {
    val byteCode = loadedClass.value.byteCode
    defineClass(loadedClass.key.asQualifiedName, byteCode.bytes, 0, byteCode.bytes.size)
}
```
since `loadedClasses` preserves insertion order.

### DIFF
--- a/djvm/cli/src/main/kotlin/net/corda/djvm/tools/cli/WhitelistGenerateCommand.kt
+++ b/djvm/cli/src/main/kotlin/net/corda/djvm/tools/cli/WhitelistGenerateCommand.kt
@@ -33,7 +33,7 @@ class WhitelistGenerateCommand : CommandBase() {
     override fun validateArguments() = paths.isNotEmpty()
 
     override fun handleCommand(): Boolean {
-        val entries = AnalysisConfiguration().use { configuration ->
+        val entries = AnalysisConfiguration.createRoot().use { configuration ->
             val entries = mutableListOf<String>()
             val visitor = object : ClassAndMemberVisitor(configuration, null) {
                 override fun visitClass(clazz: ClassRepresentation): ClassRepresentation {
--- a/djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
+++ b/djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
@@ -1,6 +1,7 @@
 package net.corda.djvm.rewiring
 
 import net.corda.djvm.SandboxConfiguration
+import net.corda.djvm.analysis.AnalysisConfiguration
 import net.corda.djvm.analysis.AnalysisContext
 import net.corda.djvm.analysis.ClassAndMemberVisitor
 import net.corda.djvm.analysis.ExceptionResolver.Companion.getDJVMExceptionOwner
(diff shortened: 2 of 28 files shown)


djvm/cli/src/main/kotlin/net/corda/djvm/tools/cli/WhitelistGenerateCommand.kt
```diff
@@ -33,7 +33,7 @@ class WhitelistGenerateCommand : CommandBase() {
     override fun validateArguments() = paths.isNotEmpty()
 
     override fun handleCommand(): Boolean {
-        val entries = AnalysisConfiguration().use { configuration ->
+        val entries = AnalysisConfiguration.createRoot().use { configuration ->
```
tlil: With the deprecation of whitelists, we can probably just remove this CLI command by the way :-)

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
```diff
@@ -50,36 +53,62 @@ class SandboxClassLoader(
...
+        loadClassAndBytes(ClassSource.fromClassName("sandbox.java.lang.Object"), context)
+        loadClassAndBytes(ClassSource.fromClassName("sandbox.java.lang.StackTraceElement"), context)
+        loadClassAndBytes(ClassSource.fromClassName("sandbox.java.lang.Throwable"), context).type
+    }
 
     /**
-     * The re-writer to use for registered classes.
+     * Creates an empty [SandboxClassLoader] with exactly the same
+     * configuration as this one, but with the given [AnalysisContext].
+     * @param newContext
```
tlil: Strip or describe

djvm/src/test/kotlin/net/corda/djvm/TestBase.kt
```diff
@@ -203,4 +277,77 @@ abstract class TestBase {
         }
     }
 
+    @Suppress("MemberVisibilityCanBePrivate")
+    protected class DJVM(private val classLoader: ClassLoader) {
+        private val djvm: Class<*> = classFor("sandbox.java.lang.DJVM")
+        val sandboxObject: Class<*> by lazy { classFor("sandbox.java.lang.Object") }
+        val string: Class<*> by lazy { classFor("sandbox.java.lang.String") }
```
tlil: No biggie, but with the dev preview coming up, etc., do you envision people being able to reuse this test base class from Java as well? Or only Kotlin? If the former, we should watch out for using reserved keywords.

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
```diff
@@ -50,36 +53,62 @@ class SandboxClassLoader(
...
+    fun loadForSandbox(className: String): LoadedClass {
+        val sandboxClass = loadClassForSandbox(className)
+        val sandboxName = Type.getInternalName(sandboxClass)
+        var loader = this
+        while(true) {
+            val loaded = loader.loadedClasses[sandboxName]
+            if (loaded != null) {
+                return loaded
+            }
+            loader = loader.parent as? SandboxClassLoader ?: return LoadedClass(sandboxClass, UNMODIFIED)
```
tlil: Would debug- or trace-level logging for when unmodified and modified classes get loaded be useful? Or do you log sufficient info elsewhere?

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
chrisr3: :+1:

djvm/src/test/kotlin/net/corda/djvm/TestBase.kt
chrisr3: Good point, I'll rename them.

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
chrisr3: The existing logging should be enough. All that's happening here is that we're returning either the existing contents of the `loadedClasses` maps or an `UNMODIFIED` placeholder. Or throwing `ClassNotFoiundException` if the class isn't there at all.

tlil: Coolio 👍

djvm/cli/src/main/kotlin/net/corda/djvm/tools/cli/WhitelistGenerateCommand.kt
chrisr3: Heh, maybe in the next PR... ;-)

djvm/src/main/kotlin/net/corda/djvm/rewiring/SandboxClassLoader.kt
```diff
@@ -50,36 +53,65 @@ class SandboxClassLoader(
...
-     * Given a class name, provide its corresponding [LoadedClass] for the sandbox.
-     */
-    fun loadForSandbox(name: String, context: AnalysisContext): LoadedClass {
-        return loadClassAndBytes(ClassSource.fromClassName(analysisConfiguration.classResolver.resolveNormalized(name)), context)
+    private fun loadClassForSandbox(className: String): Class<*> {
+        val sandboxName = analysisConfiguration.classResolver.resolveNormalized(className)
+        return try {
+            loadClass(sandboxName)
+        } finally {
+            context.messages.acceptProvisional()
```
exFalso: In what situation do we get provisional messages?

chrisr3: The only provisional message is created by `AbstractSourceClassLoader.classReader()`.

What happens is that the JVM first tries to load a class from the parent `SandboxClassLoader`. If this classloader cannot satisfy the request then it is expected to throw an exception and delegate to the child `SandboxClassLoader` instead. (This is typical classloader behaviour.) However, a `SandboxClassLoader` also wants to store an error inside its `AnalysisContext` whenever this happens, and finding any errors here is considered lethal to sandboxing. Hence `SandboxClassLoader.loadClass()` will drop this provisional message when it catches the exception that caused it.

chrisr3:
A lot of this code is good to have anyway, if only for the more realistic test coverage and better use of `SandboxRuntimeContext`. There's nothing here that _forces_ us to run with parent & child SandboxClassLoaders.

Issue #3800, opened 2018-08-17
In Demobench, saving a profile, opening it, and adding a node skips next name


joeldudleyr3:
There is a standard ordering of default names for nodes started in DemoBench (Notary, Bank of Breakfast Tea, Bank of Big Apples...).

However, if you do the following:

1. Start DemoBench
2. Start Notary
3. Start Bank Of Breakfast Tea
4. Save profile
5. Open that profile

When you next click add node, instead of suggesting Bank Of Big Apples, it wants to create Bank Of Fondue.

Tracked in JIRA here: https://r3-cev.atlassian.net/browse/CORDA-1480.

joeldudleyr3:
Thanks @vardan10 !

PR #4113, opened 2018-10-24
CORDA-1480: In Demobench, saving a profile, opening it, and adding a node skips next name

vardan10:
Simple fix for #3800
Jira issue: CORDA-1480

Now Selecting the name based on number of active nodes open.


tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeData.kt
```diff
@@ -25,8 +25,13 @@ object SuggestedDetails {
             "Bank of Golden Gates" to "San Francisco"
     )
 
-    var cursor = 0
+    private var cursor = 0
     val nextBank: Pair<String, String> get() = banks[cursor++ % banks.size]
+
+    fun excludeNames(exists: (String) -> Boolean) {
+        cursor = 0
```
chrisr3: I don't think you need to set `cursor = 0` here. In the "happy" case, `cursor` will already be pointing to an unused name and so your loop should exit immediately.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeData.kt
```diff
@@ -25,8 +25,13 @@ object SuggestedDetails {
...
+        for (bank in banks) if (!exists(bank.first)) break else cursor++
```
chrisr3: I think you want `cursor++ % banks.size` rather than `cursor++`, to ensure that the value is always a valid index into `banks`. In fact, you could even create a new property:
```kotlin
private val nextIndex: Int get() = cursor++ % banks.size
```
and then use this both here and in `nextBank`.

tools/demobench/src/main/kotlin/net/corda/demobench/model/NodeData.kt
vardan10: @chrisr3 
As per issue #3800 the ordering of banks as per list of banks is important.
If I don't set cursor=0, then after we load a profile, cursor will be incremented again by the number of banks already present in the profile loaded.

chrisr3: OK, I see. But my issue here is that this particular code is being invoked _every time_ we create a new node, and not just when we load a profile.

I've dug deeper into this, and I think the underlying issue is that we really want to combine `excludeNames` and `nextBank` into a _single_ function. Effectively, we make `nextBank` _smarter_:
```kotlin
private var cursor = 0

fun nextBank(exists: (String) -> Boolean): Pair<String, String> {
    for (bank in banks) if (!exists(bank.first)) return bank else cursor++ % banks.size
    return banks[cursor % banks.size]
}

fun reset() {
    cursor = 0
}
```
And then we can invoke `SuggestedDetails.reset()` when we reset the `NodeController`:
```diff
@@ -165,6 +165,7 @@ class NodeController(check: atRuntime = ::checkExists) : Con
troller() {
         nodeInfoFilesCopier.reset()
         notaryIdentity = null
         networkParametersCopier = null
+        SuggestedDetails.reset()
     }
 
     /**
```

chrisr3: Possibly something more like this:
```kotlin
fun nextBank(exists: (String) -> Boolean): Pair<String, String> {
    for (i in 0 until banks.size) {
        val bank = banks[cursor]
        if (!exists(bank.first)) {
            return bank
        }
        cursor = (cursor + 1) % banks.size
    }
    return banks[cursor]
}
```

chrisr3: OK, thanks for doing that. Do we still need `SuggestedDetails.reset()` when we reset `NodeComtroller` (i.e. whenever we load a profile)?

vardan10: I don't think we need ```SuggestedDetails.reset()```.
I tried the steps mentioned in #3800, and its working perfectly fine.

chrisr3: Try this:
- Create a profile containing just the Notary.
- Now create nodes for all banks up to and including "Bank of Golden Gates"
- Reload your notary profile.
- Create a new bank node.
Without the `reset()`, this new node is "Bank of Golden Gates" whereas we would expect "Bank of Breakfast Tea".

vardan10: @CaisR3 -- I have added ```SuggestedDetails.reset()```

vardan10:
I acknowledge that my contribution is in accordance with the DCO: https://developercertificate.org/

PR #4115, opened 2018-10-25
ENT-1906: Don't transform java.Object[] to sandbox.Object[].

chrisr3:
An array of objects is itself still `java.lang.Object` inside a sandbox, and so it cannot be added to `sandbox.java.lang.Object[]`. So keep all object arrays as `java.lang.Object[]` inside the sandbox.

### DIFF
--- a/djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysInheritFromSandboxedObject.kt
+++ b/djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysInheritFromSandboxedObject.kt
@@ -24,8 +24,12 @@ class AlwaysInheritFromSandboxedObject : ClassDefinitionProvider, Emitter {
 
     override fun emit(context: EmitterContext, instruction: Instruction) = context.emit {
         if (instruction is TypeInstruction &&
-                instruction.typeName == OBJECT_NAME) {
+                instruction.typeName == OBJECT_NAME &&
+                instruction.operation != Opcodes.ANEWARRAY &&
+                instruction.operation != Opcodes.MULTIANEWARRAY) {
             // When creating new objects, make sure the sandboxed type gets used.
+            // However, an array is always [java.lang.Object] so we must exclude
+            // arrays from this so that we can still support arrays of arrays.
             new(SANDBOX_OBJECT_NAME, instruction.operation)
             preventDefault()
         }
--- a/djvm/src/main/kotlin/net/corda/djvm/rules/implementation/RewriteObjectMethods.kt
+++ b/djvm/src/main/kotlin/net/corda/djvm/rules/implementation/RewriteObjectMethods.kt
@@ -1,0 +1,29 @@
+package net.corda.djvm.rules.implementation
+
+import net.corda.djvm.code.Emitter
+import net.corda.djvm.code.EmitterContext
+import net.corda.djvm.code.Instruction
+import net.corda.djvm.code.instructions.MemberAccessInstruction
+import org.objectweb.asm.Opcodes.*
+
+/**
+ * We cannot wrap Java array objects - and potentially others - and so these would still
+ * use the non-deterministic [java.lang.Object.hashCode] by default. Therefore we intercept
+ * these invocations and redirect them to our [sandbox.java.lang.DJVM] object.
+ */
+class RewriteObjectMethods : Emitter {
+    override fun emit(context: EmitterContext, instruction: Instruction) = context.emit {
+        if (instruction is MemberAccessInstruction && instruction.owner == "java/lang/Object") {
+            when (instruction.operation) {
+                INVOKEVIRTUAL -> if (instruction.memberName == "hashCode" && instruction.signature == "()I") {
+                    invokeStatic(
+                        owner = "sandbox/java/lang/DJVM",
+                        name = "hashCode",
+                        descriptor = "(Ljava/lang/Object;)I"
+                    )
+                    preventDefault()
+                }
+            }
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


chrisr3:
Done :vulcan_salute:

djvm/src/main/kotlin/net/corda/djvm/rules/implementation/AlwaysInheritFromSandboxedObject.kt
```diff
@@ -24,8 +24,12 @@ class AlwaysInheritFromSandboxedObject : ClassDefinitionProvider, Emitter {
 
     override fun emit(context: EmitterContext, instruction: Instruction) = context.emit {
         if (instruction is TypeInstruction &&
-                instruction.typeName == OBJECT_NAME) {
+                instruction.typeName == OBJECT_NAME &&
+                instruction.operation != Opcodes.ANEWARRAY &&
+                instruction.operation != Opcodes.MULTIANEWARRAY) {
```
tlil: So what happens if you then call `.hashCode()` on the `[]` or `[][]`? Source of non-determinism?

exFalso: doesn't it dispatch through sandbox.System?

tlil: Don't think so if it hits `java.lang.Object.hashCode:()I`...? Unless we always rewrite those calls and there are no instances where that happens for arrays in the `java.lang` classes, etc. Not sure with the recent changes, so ... @chrisr3, will need you to confirm :-)

Either way, we should at least add an explicit unit test to check that this doesn't happen. That way  we can guarantee that behaviour moving forward regardless of what changes get introduced.

tlil: (unless there is already such a test - haven't checked)

chrisr3: I am happy to include this test:
```java
public static class ObjectArray implements Function<Object, Integer> {
    @Override
    public Integer apply(Object o) {
        Object[] obj = new Object[10];
        obj[0] = new Object();
        return obj[0].hashCode();
    }
}
```
which would be invoked as:
```java
@Test
public void testObjectArrays() {
    parentedSandbox(WARNING, true, ctx -> {
        SandboxExecutor<Object, Integer> executor = new DeterministicSandboxExecutor<>(ctx.getConfiguration());
        ExecutionSummaryWithResult<Integer> output = WithJava.run(executor, ObjectArray.class, null);
        assertThat(output.getResult()).isEqualTo(0xfed_c0de + 1);
        return null;
    });
}
```
When you run this, :green_heart: IT PASSES :green_heart: :tada:

More specifically, you create `java.lang.Object[10]` and populate it with `sandbox.java.lang.Object`:
```
  public strictfp sandbox.java.lang.Integer apply(java.lang.Object);
    descriptor: (Ljava/lang/Object;)Lsandbox/java/lang/Integer;
    flags: ACC_PUBLIC, ACC_STRICT
    Code:
      stack=4, locals=3, args_size=2
         0: bipush        10
         2: dup
         3: ldc           #61                 // String java/lang/Object
         5: invokestatic  #65                 // Method sandbox/net/corda/djvm/costing/RuntimeCostAccounter.recordArrayAllocation:(ILjava/lang/String;)V
         8: anewarray     #2                  // class java/lang/Object
        11: astore_2
        12: aload_2
        13: iconst_0
        14: ldc           #61                 // String java/lang/Object
        16: invokestatic  #69                 // Method sandbox/net/corda/djvm/costing/RuntimeCostAccounter.recordAllocation:(Ljava/lang/String;)V
        19: new           #48                 // class sandbox/java/lang/Object
        22: dup
        23: invokestatic  #57                 // Method sandbox/net/corda/djvm/costing/RuntimeCostAccounter.recordInvocation:()V
        26: invokespecial #58                 // Method sandbox/java/lang/Object."<init>":()V
        29: aastore
        30: aload_2
        31: iconst_0
        32: aaload
        33: invokestatic  #57                 // Method sandbox/net/corda/djvm/costing/RuntimeCostAccounter.recordInvocation:()V
        36: invokevirtual #3                  // Method java/lang/Object.hashCode:()I
        39: invokestatic  #57                 // Method sandbox/net/corda/djvm/costing/RuntimeCostAccounter.recordInvocation:()V
        42: invokestatic  #74                 // Method sandbox/java/lang/Integer.valueOf:(I)Lsandbox/java/lang/Integer;
        45: areturn
```

tlil: Ah, great! That's what I wanted to know. Thanks, Chris!

tlil: Actually, Chris and I just talked. And it's not that easy; That test is calling `hashCode()` on the array element. Calling `hashCode()` on the array itself leads to a nondeterministic output. @exFalso, Chris is looking into it.

chrisr3: Ah, I understand :bulb:. The _real_ concern here is about `Object[].hashCode()` still being the `native` variety. However, that's a consequence of `anewarray` returning `java.lang.Object` whereas this PR is concerned with that array's component type.

I'll try intercepting `Object.hashCode()` in the same way that I already intercept certain methods on `Class` and `ClassLoader`.

exFalso: Ah! Good catch!!

tlil: Yep, exactly :+1: Thanks, Chris!

PR #4116, opened 2018-10-25
CORDA 2131 - Extend Network Bootstrapper to enable registration of Java Package Namespaces.

josecoll:
As a Corda Network operator,
I wish to register a Java Package Namespace against the Public Key of the verified owner of that namespace using standard Corda bootstrapping and network registration tooling.

NOTE: recommend that a new tool be created for the sole purpose of Network Parameters administration (adding/removing/updating field values)


josecoll:
Adding the documentation in the meantime.

anthonykeenan:
I haven't had a proper look yet because the tests are still failing, but it's going to need the documentation updating please!

tlil:
Agree, something along those lines would be good, @anthonykeenan.

josecoll:
@anthonykeenan Agree with your suggestions, but as I mentioned in the PR description, we really need a separate tool to update Network Parameters (within the bootstrapping part), so I would suggest this new tool use the configuration file based approach.
In the meantime, I would ask that this PR is approved based on following existing implementation pattern of adding new CLI arguments.

josecoll:
@tlil Any update on progressing this? (we need to start using the tooling in our QA tests).

tools/bootstrapper/src/test/kotlin/net/corda/bootstrapper/PackageOwnerParsingTest.kt
```diff
@@ -0,0 +1,194 @@
...
+        val aliceKeyStorePath = dirAlice / "_teststore"
+        val args = arrayOf("--register-package-owner", "com.example.stuff=$aliceKeyStorePath:$ALICE_PASS:$ALICE",
+                                        "--unregister-package-owner", "com.example.stuff2")
+        commandLine.parse(*args)
+        assertThat(packageOwner.registerPackageOwnership).containsKey(JavaPackageName("com.example.stuff"))
+        assertThat(packageOwner.unregisterPackageOwnership).contains(JavaPackageName("com.example.stuff2"))
+    }
+}
+
+class PackageOwner {
```
tlil: Why is this duplicated in the test class?

tlil:
@josecoll sorry, ran out of time today, but will look at this properly tomorrow, promise.

tools/bootstrapper/src/test/kotlin/net/corda/bootstrapper/PackageOwnerParsingTest.kt
josecoll: Good spot - originally it was for complete testability isolation.
I've now moved this back into the mainline codebase but this required some slight restructuring of the main `NetworkBootstrapperRunner` class hierarchy.  Happy to apply suggested improvements.

docs/source/network-bootstrapper.rst
```diff
@@ -256,7 +256,10 @@ The network bootstrapper can be started with the following command-line options:
 .. code-block:: shell
 
     bootstrapper [-hvV] [--no-copy] [--dir=<dir>] [--logging-level=<loggingLevel>]
-                 [--minimum-platform-version=<minimumPlatformVersion>] [COMMAND]
+                 [--minimum-platform-version=<minimumPlatformVersion>]
+                 [--register-package-owner java-package-namespace=keystore-file:password:alias]
```
anthonykeenan: I think the choice of separators here is not going to work. Picocli allows = or a space between option, also `:` is valid in filenames for us pesky windows users. So we could have:

`bootstrapper --register-package-owner=net.corda.bank=d:\store:pass:bank`

(Which gives this error: `Invalid value for option '--register-package-owner' (<JavaPackageName=PublicKey>): cannot convert 'd:\store:pass:bank' to PublicKey (java.lang.IllegalArgumentException: keystore argument must specify 3 elements separated by colon: 'keyStorePath:password:alias')`)


I think using something like semicolons, and having all the seperators the same might work better?

`bootstrapper --register-package-owner=net.corda.bank;d:\store;pass;bank`

docs/source/network-bootstrapper.rst
```diff
@@ -266,8 +269,60 @@ The network bootstrapper can be started with the following command-line options:
 * ``--help``, ``-h``: Show this help message and exit.
 * ``--version``, ``-V``: Print version information and exit.
 * ``--minimum-platform-version``: The minimum platform version to use in the generated network-parameters.
+* ``--register-package-owner``: Register a java package namespace with its owners public key.
```
anthonykeenan: Would be nice to have a link to the section further down that has the info

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -20,16 +24,66 @@ class NetworkBootstrapperRunner : CordaCliWrapper("bootstrapper", "Bootstrap a l
...
+ * Converter to extract PublicKey from KeyStore specification (keyStorePath:password:alias)
+ */
+class PublicKeyConverter : CommandLine.ITypeConverter<PublicKey> {
+    override fun convert(keyStore: String): PublicKey {
+        if (!keyStore.isBlank()) {
+            val keyStoreSpec = keyStore.split(":")
+            if (keyStoreSpec.size != 3)
+                throw IllegalArgumentException("keystore argument must specify 3 elements separated by colon: 'keyStorePath:password:alias'")
+            val ks = loadKeyStore(Paths.get(keyStoreSpec[0]), keyStoreSpec[1])
+            return ks.getCertificate(keyStoreSpec[2]).publicKey
```
anthonykeenan: If this fails for some reason, you get an error: `Invalid value for option '--register-package-owner' (<JavaPackageName=PublicKey>): cannot convert 'teststorex.jks:password:alice' to PublicKey (java.nio.file.NoSuchFileException: teststorex.jks)`
Can we make the error a bit clearer? I.e. check for the existence of the file first, and throw that error. Throw a corrupt keystore error if it exists but isn't accessible, throw an invalid password error if the password is wrong etc... 
I also worry that we might need to allow delimiters, i.e. if the password contains a semicolon:
`--register-package-owner net.something;keystore.jks;passw;rd;blah` would fail
`--register-package-owner net.something;keystore.jks;"passw;rd";blah` would pass
`--register-package-owner "net.something;keystore.jks;passw;rd;blah"` would fail
`--register-package-owner "net.something;keystore.jks;'passw;rd';blah"` to fix it or maybe
`--register-package-owner "net.something;keystore.jks;""passw;rd"";blah"`

docs/source/network-bootstrapper.rst
```diff
@@ -266,8 +269,60 @@ The network bootstrapper can be started with the following command-line options:
...
+* ``--unregister-package-owner``: Unregister a java package namespace.
 
 Sub-commands
 ^^^^^^^^^^^^
 
-``install-shell-extensions``: Install ``bootstrapper`` alias and auto completion for bash and zsh. See :doc:`cli-application-shell-extensions` for more info.
\ No newline at end of file
+``install-shell-extensions``: Install ``bootstrapper`` alias and auto completion for bash and zsh. See :doc:`cli-application-shell-extensions` for more info.
+
+Package owner registration by example
```
anthonykeenan: I think this section should go below the `Updating the contract whitelist for bootstrapped networks` section and above the `Command-line options` section, which is essentially the reference manual at the top.

I also think  the section should just be called `Package owner registration`, all the other sections have examples but don't say they are `by example`, and I think an opening paragraph which talks about what the feature is, what it is for, possibly a link to somwhere else that documents it in more detail? To be honest, I'm not entirely sure myself what it is for!

docs/source/network-bootstrapper.rst
josecoll: Adjusted as discussed and described above.

docs/source/network-bootstrapper.rst
josecoll: Added a link to the Package Namespace docs.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
josecoll: I've improved the exception handling but not sure what to do about password parsing.

anthonykeenan: It might be the best we can do is put a big warning in the bootstrapper docs?

docs/source/network-bootstrapper.rst
tlil: Sorry to be a pain, but be aware that semicolons can be tricky in bash unless the argument is passed as a quoted string.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
josecoll: OK - I've added some additional parsing logic to deal with passwords including the delimiter itself.

docs/source/network-bootstrapper.rst
josecoll: Updated to reflect these recommendations.

josecoll:
@anthonykeenan I hope I've addressed all your concerns.

docs/source/network-bootstrapper.rst
```diff
@@ -247,6 +247,64 @@ To give the following:
...
+2. Register the package namespace to be claimed by the public key generated above
+
+.. code-block:: shell
+
+    # Use a prebuilt bootstrapper jar (from the corda maven repository) or build it from source:
+    ./gradlew tools:bootstrapper:clean tools:bootstrapper:jar
+
+    # Register the java package namespace using the bootstrapper tool
+    java -jar tools/bootstrapper/build/libs/network-bootstrapper.jar \
+         --dir samples/bank-of-corda-demo/build/nodes
```
anthonykeenan: Do you need an extra `\` at the end of this line for the multiline?

docs/source/network-bootstrapper.rst
```diff
@@ -247,6 +247,64 @@ To give the following:
...
+         --register-package-owner net.corda.bank;./_teststore;MyStorePassword;MyKeyAlias
+
+3. Unregister the package namespace
+
+.. code-block:: shell
+
+    # Unregister the java package namespace using the bootstrapper tool
+    java -jar tools/bootstrapper/build/libs/network-bootstrapper.jar \
+         --dir samples/bank-of-corda-demo/build/nodes
```
anthonykeenan: Do you need an extra `\` at the end of this line for the multiline?

docs/source/network-bootstrapper.rst
```diff
@@ -247,6 +247,64 @@ To give the following:
...
+`Java keytool <https://docs.oracle.com/javase/8/docs/technotes/tools/windows/keytool.html>`_.
+
+The following four items are passed as a semi-colon separated string to the ``--register-package-owner`` command:
+
+    1. Java package name (e.g com.my_company).
+    2. keystore-file refers to the location of key store file containing the signed certificate.
+    3. password refers tp the key store password (not to be confused with the key password)
+    4. alias refers to the name associated with a certificate containing the public key to be associated with the package namespace
+
+Lets use the Bank of Corda example to initialise a simple network:
```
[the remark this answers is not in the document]
josecoll: "Let us" to be on the safe side

docs/source/network-bootstrapper.rst
josecoll: removed

docs/source/network-bootstrapper.rst
josecoll: removed

docs/source/network-bootstrapper.rst
```diff
@@ -247,6 +247,64 @@ To give the following:
...
+
+2. Register the package namespace to be claimed by the public key generated above:
+
+.. code-block:: shell
+
+    # Use a prebuilt bootstrapper jar (from the corda maven repository) or build it from source:
+    ./gradlew tools:bootstrapper:clean tools:bootstrapper:jar
+
+    # Register the java package namespace using the bootstrapper tool
+    java -jar tools/bootstrapper/build/libs/network-bootstrapper-|version|.jar
```
anthonykeenan: Sorry, I mean to remove the reference to repository location, since the binaries will be downloaded, like it is further up, i.e. ``java -jar network-bootstrapper-VERSION.jar --option1 etc...``

anthonykeenan: https://ci-artifactory.corda.r3cev.com/artifactory/corda-releases/net/corda/corda-tools-network-bootstrapper/3.3-corda/ is the most recent version.
I think the documentation where possible should be written so that people don't have to download the repo. Would it be a massive pain to change the example so it uses the example-cordapp instead?

tlil:
@anthonykeenan @josecoll - as mentioned on CORDA-2055, I'll address that with my two tickets.

docs/source/network-bootstrapper.rst
josecoll: Done! and raised a separate story to provide a downloadable reference such that the reader can follow the documentation instructions.
https://r3-cev.atlassian.net/browse/CORDA-2185

PR #4126, opened 2018-10-29
CORDA-1838: A few misc fixes

anthonykeenan:
* Remove `joptsimple` from node/shell (as it uses picocli now)
** Add `joptsimple` to explorer/irs-demo as they were inheriting through a dependency on node
* Fix some grammar
* Remove `AbstractArgsParser` as no longer used


samples/irs-demo/web/build.gradle
```diff
@@ -70,7 +70,11 @@ dependencies {
     compile project(":client:rpc")
     compile project(":client:jackson")
     compile project(":test-utils")
-    compile project(path: ":samples:irs-demo:cordapp", configuration: "demoArtifacts")
+	compile project(path: ":samples:irs-demo:cordapp", configuration: "demoArtifacts")
```
sollecitom: Is the formatting OK here? I see some indentation, not sure if it's MS GitHub playing a trick on me.

samples/irs-demo/web/build.gradle
```diff
@@ -70,7 +70,11 @@ dependencies {
...
+
+	// JOpt: for command line flags.
+	compile "net.sf.jopt-simple:jopt-simple:$jopt_simple_version"
```
sollecitom: Any reason why we still need this here? Is it because we haven't applied PicoCLI to these yet?

samples/irs-demo/web/build.gradle
anthonykeenan: Yeah the formatting is fine it's just github

samples/irs-demo/web/build.gradle
anthonykeenan: Yes the irs-demo previously had a dependency on joptsimple via its dependency on node. Adding picocli to the demo apps is out of scope atm.

samples/irs-demo/web/build.gradle
tlil: tabs vs spaces?

PR #4128, opened 2018-10-29
CORDA-1633 - Check Java version before launch

tlil:
- [JIRA: CORDA-1633](https://r3-cev.atlassian.net/browse/CORDA-1633) - Present the user with a sensible message if the node is attempted started on Java 9/10/11 instead of Java 8


tlil:
Running with Java 11 now yields:

```bash
$ /usr/lib/jvm/java-11-openjdk-amd64/bin/java -jar build/libs/corda-4.0-SNAPSHOT.jar 
Error: Unsupported Java version 10.0.2; currently only version 1.8 is supported.
```

PR #4131, opened 2018-10-29
CORDA-2109: Fix a bug that prevents consecutive multiparty contract upgrades

adagys:
The contract upgrade handler assumes that the state to be upgraded is
created by a WireTransaction. This breaks the upgrade process if it was
in fact issued by a ContractUpgradeWireTransactions or a NotaryChangeWireTransaction.

Backport of https://github.com/corda/corda/pull/4082


adagys:
I've verified that the updated test fails without this fix, and the fix makes it happy

PR #4134, opened 2018-10-30
ENT-1858: Notary cleanup

adagys:
Migrated all non-BFT notary implementations to use async commits.


adagys:
I've added a change to MockNetwork which needs a re-review

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
```diff
@@ -34,14 +35,35 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
...
+            commitInputStates(parts, txId, requestPayload)
             signTransactionAndSendResponse(txId)
         } catch (e: NotaryInternalException) {
             throw NotaryException(e.error, txId)
         }
         return null
     }
 
+    @Suspendable
+    private fun commitInputStates(parts: TransactionParts, txId: SecureHash, requestPayload: NotarisationPayload) {
```
rick-r3: Why put this back into the shipped code in Corda?  If it remained in the CorDapp as it is now, we can change this outside of a full Corda release and/or not have to quite so carefully align the versions of the CorDapps with Corda releases.

e.g. if we fix that TODO, all the CorDapps need updating / become incompatible.  If we left it in the CorDapp, not so much...

core/src/main/kotlin/net/corda/core/internal/notary/UniquenessProvider.kt
```diff
@@ -21,5 +21,13 @@ interface UniquenessProvider {
...
             references: List<StateRef> = emptyList()
-    )
+    ): CordaFuture<Result>
+
+    /** The outcome of committing a transaction. */
+    sealed class Result {
+        /** Indicates that all input states have been committed successfully. */
+        object Success : Result()
+        /** Indicates that the transaction has not been committed. */
+        data class Failure(val error: NotaryError) : Result()
```
rick-r3: So now we don't throw an exception ?

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
```diff
@@ -17,7 +18,7 @@ import net.corda.core.utilities.unwrap
  * Additional transaction validation logic can be added when implementing [validateRequest].
  */
 // See AbstractStateReplacementFlow.Acceptor for why it's Void?
-abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service: TrustedAuthorityNotaryService) : FlowLogic<Void?>() {
+abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service: SinglePartyNotaryService) : FlowLogic<Void?>() {
```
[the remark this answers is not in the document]
adagys: We might want to create an abstract BFT notary service class at some point. I would've called this one CFT, but it's used by both single-node and HA notaries.

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
adagys: This logic already was in the core module, I just moved it from AsyncCFTNotaryService

core/src/main/kotlin/net/corda/core/internal/notary/UniquenessProvider.kt
adagys: No, it gets a bit messy if we want to catch it through the async operation boundary

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
adagys: Also note that the SimpleNotaryService is in a "virtual" CorDapp, so we'd always need Corda releases to update it

rick-r3: But the NotaryService impl. was in the CorDapp, so could be overridden.  Even though we generally just used one in the Corda codebase.  I'd rather we had maximum flexibility in the CorDapps.

adagys: Okay, fair, I'll move it to SinglePartyNotaryService

PR #4137, opened 2018-10-30
CORDA-2099 make serializer factory interface

distributedleetravis:
Part of [CORDA-2099](https://r3-cev.atlassian.net/browse/CORDA-2099). As a prerequisite for developing the new `TypeModellingSerializerFactory`, we pull out a `SerializerFactory` interface, renaming the old concrete class to `DefaultSerializerFactory`, and move initialisation to a `SerializerFactoryBuilder`. Other changes in this PR mostly concern decoupling tests from the implementation class, so we can use existing tests as a handrail for development of the modified implementation.

### DIFF
--- a/serialization-deterministic/build.gradle
+++ b/serialization-deterministic/build.gradle
@@ -50,6 +50,7 @@ task patchSerialization(type: Zip, dependsOn: serializationJarTask) {
         exclude 'net/corda/serialization/internal/amqp/AMQPSerializerFactories*'
         exclude 'net/corda/serialization/internal/amqp/AMQPStreams*'
         exclude 'net/corda/serialization/internal/amqp/AMQPSerializationThreadContext*'
+        exclude 'net/corda/serialization/internal/model/DefaultCacheProvider*'
     }
 
     reproducibleFileOrder = true
--- a/serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
+++ b/serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
@@ -15,7 +15,7 @@ fun createSerializerFactoryFactory(): SerializerFactoryFactory = DeterministicSe
 
 private class DeterministicSerializerFactoryFactory : SerializerFactoryFactory {
     override fun make(context: SerializationContext) =
-            SerializerFactoryBuilder.build(
+            SerializerFactoryBuilder.buildDeterministic(
             whitelist = context.whitelist,
             classCarpenter = DummyClassCarpenter(context.whitelist, context.deserializationClassLoader))
 }
--- a/serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
+++ b/serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
@@ -15,7 +15,7 @@ fun createSerializerFactoryFactory(): SerializerFactoryFactory = DeterministicSe
 
 private class DeterministicSerializerFactoryFactory : SerializerFactoryFactory {
     override fun make(context: SerializationContext) =
-            SerializerFactoryBuilder.buildWithCarpenter(
+            SerializerFactoryBuilder.build(
             whitelist = context.whitelist,
             classCarpenter = DummyClassCarpenter(context.whitelist, context.deserializationClassLoader))
 }
--- a/serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
+++ b/serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
@@ -15,15 +15,9 @@ fun createSerializerFactoryFactory(): SerializerFactoryFactory = DeterministicSe
 
 private class DeterministicSerializerFactoryFactory : SerializerFactoryFactory {
     override fun make(context: SerializationContext) =
-        SerializerFactory(
+            SerializerFactoryBuilder.buildWithCarpenter(
             whitelist = context.whitelist,
-            classCarpenter = DummyClassCarpenter(context.whitelist, context.deserializationClassLoader),
-            serializersByType = mutableMapOf(),
-            serializersByDescriptor = mutableMapOf(),
-            customSerializers = ArrayList(),
-            customSerializersCache = mutableMapOf(),
-            transformsCache = mutableMapOf()
-        )
+            classCarpenter = DummyClassCarpenter(context.whitelist, context.deserializationClassLoader))
 }
 
 private class DummyClassCarpenter(
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactoryBuilder.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactoryBuilder.kt
@@ -1,6 +1,5 @@
 package net.corda.serialization.internal.amqp
 
-import net.corda.core.DeleteForDJVM
 import net.corda.core.KeepForDJVM
 import net.corda.core.serialization.ClassWhitelist
 import net.corda.serialization.internal.carpenter.ClassCarpenter
@@ -10,8 +9,9 @@ import net.corda.serialization.internal.carpenter.ClassCarpenterImpl
 object SerializerFactoryBuilder {
 
     @JvmStatic
+    // Has to be named differently, otherwise serialization-deterministic:determinise fails mysteriously.
     fun buildDeterministic(whitelist: ClassWhitelist, classCarpenter: ClassCarpenter) =
-            SerializerFactory(
+            makeFactory(
                     whitelist,
                     classCarpenter,
                     DefaultEvolutionSerializerProvider,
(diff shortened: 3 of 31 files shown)


serialization/src/test/java/net/corda/serialization/internal/amqp/JavaPrivatePropertyTests.java
```diff
@@ -147,10 +147,7 @@ public void singlePrivateWithConstructor() throws NotSerializableException, NoSu
         //
         // Now ensure we actually got a private property serializer
         //
-        Field f = SerializerFactory.class.getDeclaredField("serializersByDescriptor");
```
distributedleetravis: This is if anything more evil than `@VisibleForTesting`.

serialization-deterministic/build.gradle
```diff
@@ -50,6 +50,7 @@ task patchSerialization(type: Zip, dependsOn: serializationJarTask) {
         exclude 'net/corda/serialization/internal/amqp/AMQPSerializerFactories*'
         exclude 'net/corda/serialization/internal/amqp/AMQPStreams*'
         exclude 'net/corda/serialization/internal/amqp/AMQPSerializationThreadContext*'
+        exclude 'net/corda/serialization/internal/model/DefaultCacheProvider*'
```
chrisr3: And another kitten meets a gruesome end... :cat2:

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactoryBuilder.kt
```diff
@@ -0,0 +1,59 @@
...
+    // Has to be named differently, otherwise serialization-deterministic:determinise fails mysteriously.
+    fun buildDeterministic(whitelist: ClassWhitelist, classCarpenter: ClassCarpenter) =
+            makeFactory(
+                    whitelist,
+                    classCarpenter,
+                    DefaultEvolutionSerializerProvider,
+                    ::SerializerFingerPrinter,
+                    false)
+
+    @JvmStatic
```
chrisr3: Add `@DeleteForDJVM` here, for clarity.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactoryBuilder.kt
```diff
@@ -0,0 +1,59 @@
...
+import net.corda.core.KeepForDJVM
+import net.corda.core.serialization.ClassWhitelist
+import net.corda.serialization.internal.carpenter.ClassCarpenter
+import net.corda.serialization.internal.carpenter.ClassCarpenterImpl
+
+@KeepForDJVM
+object SerializerFactoryBuilder {
+
+    @JvmStatic
+    // Has to be named differently, otherwise serialization-deterministic:determinise fails mysteriously.
```
chrisr3: Not _quite_ so mysterious - the `@DeleteForDJVM ClassCarpenterImpl` reference requires that one of the `build()` methods is deleted. This `build()` method has default parameters, which means that Kotlin synthesises a `build$default()` companion function for it. And this function must be deleted too, of course.

The JarFilter step would probably work if the second `build()` method didn't share a pair of default parameter lambdas with the first method. Kotlin implements these lambdas as inner classes `$build$1` and `$build$2`. The JarFilter notices that these lambdas are associated with a deleted synthetic `build()` method and so decides to delete them both as well, and this then kills the second synthetic `build$default()` method whose default parameter values have suddenly vanished. And then the second `build()` method is left referencing a synthetic companion function which no longer exists :cry:.

OK, I suppose that is all _slightly_ mysterious...

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaPrivatePropertyTests.java
```diff
@@ -147,10 +147,7 @@ public void singlePrivateWithConstructor() throws NotSerializableException, NoSu
...
-        f.setAccessible(true);
-
-        Map<?, AMQPSerializer<?>> serializersByDescriptor = uncheckedCast(f.get(factory));
+        Map<Object, AMQPSerializer<Object>> serializersByDescriptor = factory.getSerializersByDescriptor();
```
chrisr3: Do we want to use `Object` here?

tools/shell/src/test/kotlin/net/corda/tools/shell/InteractiveShellTest.kt
```diff
@@ -57,7 +58,7 @@ class InteractiveShellTest {
         }, input, FlowA::class.java, om)
         assertEquals(expected, output!!, input)
     }
-
+
```
chrisr3: Meh

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaPrivatePropertyTests.java
distributedleetravis: That's the actual type of the field as it stands, and I'm not changing it here / yet.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactoryBuilder.kt
chrisr3: If you rebase against the `master` branch and upgrade the Gradle plugins to 4.0.34 then you should now be able to remove the `buildDeterministic()` function. Obviously the `build()` variant that uses `ClassCarpenterImpl` should still be annotated as `@DeleteForDJVM`.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactoryBuilder.kt
```diff
@@ -0,0 +1,51 @@
...
+object SerializerFactoryBuilder {
+    
+    @JvmStatic
+    @JvmOverloads
+    fun build(
+            whitelist: ClassWhitelist,
+            classCarpenter: ClassCarpenter,
+            evolutionSerializerProvider: EvolutionSerializerProvider = DefaultEvolutionSerializerProvider,
+            fingerPrinterProvider: (SerializerFactory) -> FingerPrinter = ::SerializerFingerPrinter,
+            onlyCustomSerializers: Boolean = false) =
```
shamsasari: Let's use a block body here. It's not immediately clear where the method body starts.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactoryBuilder.kt
```diff
@@ -0,0 +1,51 @@
...
+    @JvmStatic
+    @JvmOverloads
+    @DeleteForDJVM
+    fun build(
+            whitelist: ClassWhitelist,
+            carpenterClassLoader: ClassLoader,
+            lenientCarpenterEnabled: Boolean = false,
+            evolutionSerializerProvider: EvolutionSerializerProvider = DefaultEvolutionSerializerProvider,
+            fingerPrinterProvider: (SerializerFactory) -> FingerPrinter = ::SerializerFingerPrinter,
+            onlyCustomSerializers: Boolean = false) =
```
shamsasari: likewise

serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
```diff
@@ -15,15 +15,9 @@ fun createSerializerFactoryFactory(): SerializerFactoryFactory = DeterministicSe
...
+            SerializerFactoryBuilder.build(
             whitelist = context.whitelist,
-            classCarpenter = DummyClassCarpenter(context.whitelist, context.deserializationClassLoader),
-            serializersByType = mutableMapOf(),
-            serializersByDescriptor = mutableMapOf(),
-            customSerializers = ArrayList(),
-            customSerializersCache = mutableMapOf(),
-            transformsCache = mutableMapOf()
-        )
+            classCarpenter = DummyClassCarpenter(context.whitelist, context.deserializationClassLoader))
```
shamsasari: ident issues

serialization/src/test/kotlin/net/corda/serialization/internal/amqp/CorDappSerializerTests.kt
```diff
@@ -14,11 +15,9 @@ class CorDappSerializerTests {
 
     private fun proxyFactory(
             serializers: List<SerializationCustomSerializer<*, *>>
-    ) = SerializerFactory(
-            AllWhitelist,
-            ClassLoader.getSystemClassLoader(),
-            onlyCustomSerializers = true
-    ).apply {
+    ) = SerializerFactoryBuilder.build(AllWhitelist,
```
shamsasari: This single expression is very busy. Can we split it up. Easiest thing to do would be to have a block body for this method.

serialization/src/test/kotlin/net/corda/serialization/internal/amqp/SerializationOutputTests.kt
```diff
@@ -207,9 +208,8 @@ class SerializationOutputTests(private val compression: CordaSerializationEncodi
     }
 
     private fun defaultFactory(): SerializerFactory {
-        return SerializerFactory(
-                AllWhitelist,
-                ClassLoader.getSystemClassLoader(),
+        return SerializerFactoryBuilder.build(AllWhitelist,
+                ClassCarpenterImpl(AllWhitelist, ClassLoader.getSystemClassLoader()),
```
shamsasari: Does it make sense to have this as the default value for the class carpenter? Most, if not all, of the changes to this class is this.

distributedleetravis: Having it as a default value breaks `serialization-deterministic`, because `ClassCarpenterImpl` is marked as `@DeleteForDJVM`, so any method which includes it in its parameters (including as a default parameter value) is also deleted.

It might make sense to introduce a `TestSerializerFactoryBuilder` which could provide convenient options for constructing test factories.

shamsasari: Will you be doing that in this PR?

distributedleetravis: `SerializerFactoryBuilder` is a bit of a moving target, so I'd prefer to defer this to a tidy-up phase later on.

PR #4142, opened 2018-10-30
CORDA-2164: Out-of-process node from driver does not include caller's dependencies

shamsasari:


testing/node-driver/build.gradle
```diff
@@ -24,6 +39,47 @@ sourceSets {
...
+        
+        // Get all the (transitive) dependencies of node
+        def dependencies = project(':node')
+                .configurations
+                .runtime
+                .incoming
+                .resolutionResult
+                .allDependencies
+                .collect { ((ResolvedDependencyResult)it).selected.id }
+                .unique()
```
chrisr3: So you want the contents of node's capsule? Couldn't you just unpack that? Or _run_ the node? Exactly like the smoke tests do?

shamsasari:
So not only is this overly complicated, it also doesn't work. I'm going to package `corda.jar` into node-driver's resources. The reason I didn't do this originally was that it would bloat the node-driver jar to ~60MB and it already pulls in the node module as a dependency. But I guess that's second fiddle to something that actually works (and is sane).

chrisr3:
The external Node dependencies could be generated like this:
```gradle
def node = project(':node')
task generateNodeDepsFiles {
    def externalDeps = file("$generatedResources/node-external-deps.txt")

    outputs.files externalDeps

    doLast {
        def dependencies = node.configurations.runtimeClasspath
            .resolvedConfiguration
            .resolvedArtifacts
        
        dependencies.forEach {
            if (it.moduleVersion.id.group != project.group) {
                externalDeps << it.moduleVersion.id.module << System.lineSeparator()
            }
        }
    }
}
```
By declaring what the output of this task is, we allow Gradle to skip this task once the output has been created. I'm also preferring the node's `configuration.runtimeClasspath` over the deprecated `configuration.runtime`.

But using `ClassGraph` to deduce the packages is _*SCARY*_!!!! :ghost:

shamsasari:
Yes, that's what I'm doing now. I'll close this and raise another PR.

PR #4147, opened 2018-10-31
Fix CLI verbosity issue with subcommands

anthonykeenan:
If `verbose` is set on a subcommand the `CordaCliWrapper` wasn't properly throwing the stack trace to the console.


tlil:
@anthonykeenan was it intentional to include commit 67c9304?

tlil:
Thanks

PR #4151, opened 2018-11-01
[CORDA-2174] Fix demo node startup

vstoykov:
During investigation of [CORDA-2174](https://r3-cev.atlassian.net/browse/CORDA-2174) was found that running Nodes Explorer's demo nodes fails because of multiple problems.

The first problem was caused by #4100. When app config was implemented for Finance app `issuableCurrencies` was made required. This config was missing from demo setup. With the first commit I made it optional.

The second problem is that by using this app config feature if we want to add one cordapp for all nodes with driver DSL by using `cordappsForAllNodes` and then we want for some of the nodes to have different config for that app we need to pass `additionalCordapps` to `runNode` and the node will complain that there are two different jars for the same app. This is fixed by my second commit.

### DIFF
--- a/finance/src/main/kotlin/net/corda/finance/internal/CashConfigDataFlow.kt
+++ b/finance/src/main/kotlin/net/corda/finance/internal/CashConfigDataFlow.kt
@@ -24,7 +24,12 @@ class ConfigHolder(services: AppServiceHub) : SingletonSerializeAsToken() {
     val issuableCurrencies: List<Currency>
 
     init {
-        val issuableCurrenciesStringList: List<String> = uncheckedCast(services.getAppContext().config.get("issuableCurrencies"))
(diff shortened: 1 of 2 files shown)


vkolomeyko:
There is another PR #4064 which could have contributed the problems we were seeing.
@shamsasari and @sollecitom: @vstoykov 's  changes looks fine to me, but I would appreciate if you could have a second look.

vstoykov:
I squashed my commits in one and rebased against latest changes in `master`

vkolomeyko:
I was more thinking of using "Squash and merge" button when integrating this PR. GitHub provides such convenient functionality and there is no need to do it manually and force push commits.

vstoykov:
Sorry that I misunderstood you but I don't have permissions to merge.

PR #4153, opened 2018-11-01
[ENT-2678] PluginRegistrationTest is failing (fix) (OS part)

sollecitom:
https://r3-cev.atlassian.net/browse/ENT-2678

These are the necessary OS changes. Second PR to follow for Enterprise after porting these across.

### DIFF
--- a/tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaLog4j2ConfigFactory.kt
+++ b/tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaLog4j2ConfigFactory.kt
@@ -10,9 +10,8 @@ import org.apache.logging.log4j.core.config.xml.XmlConfiguration
 import org.apache.logging.log4j.core.impl.LogEventFactory
 
 @Plugin(name = "CordaLog4j2ConfigFactory", category = "ConfigurationFactory")
-@Order(10)
+@Order(Integer.MAX_VALUE)
 class CordaLog4j2ConfigFactory : ConfigurationFactory() {
-
     private companion object {
         private val SUPPORTED_TYPES = arrayOf(".xml", "*")
     }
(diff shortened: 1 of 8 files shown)


sollecitom:
Hey Rich, I didn't manage to make it work with the context provider that disables the thread context.

Apart from the fact that the error codes generation predates the async logging changes and got borked by them, we need the error codes to appear on master these days because of the testing, our perfect and only chance to collect data to populate the error codes data source.

So we should merge this, which fixes the error codes and adds a test to avoid another regression, and then figure out how to make both work together.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaLog4j2ConfigFactory.kt
```diff
@@ -22,13 +21,17 @@ class CordaLog4j2ConfigFactory : ConfigurationFactory() {
     override fun getSupportedTypes() = SUPPORTED_TYPES
 
     private class ErrorCodeAppendingConfiguration(loggerContext: LoggerContext, source: ConfigurationSource) : XmlConfiguration(loggerContext, source) {
+        init {
+            println("Log4j2: Initialising ConfigurationFactory of type CordaLog4j2ConfigFactory.")
```
anthonykeenan: Did you mean to leave this `println` in?

sollecitom: Yes, it's quite useful and logging isn't available here. The downside is that it appears in shell, but it's quite early on, so you have to scroll back up to notice.
Happy to remove it if we find it inconvenient.

sollecitom:
@rick-r3 and @blsemo : I'll give another go at making them work together.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaLog4j2ConfigFactory.kt
tlil: It'd be best to remove I think.

sollecitom:
@blsemo I managed to make it work with async logging. Dismissed @rick-r3 's review, because he's on PTO today, and we need to move forward. I still need a couple of changes in Enterprise, before getting back at my current work.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaLog4j2ConfigFactory.kt
sollecitom: Removed

PR #4154, opened 2018-11-01
Updated network docs

joeldudleyr3:
Seeks to restructure this section and create a clear split between bootstrapping vs. joining an existing network vs. creating a dynamic (i.e. non-bootstrapped) network.

If the new structure is agreed, we can then check whether there is any required information missing from each section (esp. around dynamic networks).


docs/source/bootstrapping-a-compatibility-zone.rst
```diff
@@ -0,0 +1,11 @@
+.. highlight:: kotlin
+.. raw:: html
+
+   <script type="text/javascript" src="_static/jquery.js"></script>
+   <script type="text/javascript" src="_static/codesets.js"></script>
+
+Bootstrapping a compatibility zone
+==================================
+
+You can easily bootstrap a compatibility zone for testing and pre-production use with either the
```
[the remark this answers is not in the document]
joeldudleyr3: Addressed.

docs/source/compatibility-zones.rst
```diff
@@ -4,54 +4,12 @@
    <script type="text/javascript" src="_static/jquery.js"></script>
    <script type="text/javascript" src="_static/codesets.js"></script>
 
-Compatibility zones
-===================
+What is a compatibility zone?
```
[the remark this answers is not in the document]
joeldudleyr3: Addressed.

docs/source/setting-up-a-dynamic-compatibility-zone.rst
```diff
@@ -0,0 +1,189 @@
...
+You can read more about the reasons for the creation of the bootstrapper tool
+`in a blog post on the design thinking behind Corda's network map infrastructure <https://medium.com/corda/cordas-new-network-map-infrastructure-8c4c248fd7f3>`__.
+
+**Permissioning.** And creating a zone is also unnecessary for imposing permissioning requirements beyond that of the
+base Corda network. You can control who can use your app by creating a *business network*. A business network is what we
+call a coalition of nodes that have chosen to run a particular app within a given commercial context. Business networks
+aren't represented in the Corda API at this time, partly because the technical side is so simple. You can create one
+via a simple three step process:
+
+1. Distribute a list of X.500 names that are members of your business network, e.g. a simple way to do this is by
```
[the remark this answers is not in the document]
joeldudleyr3: Addressed.

docs/source/setting-up-a-dynamic-compatibility-zone.rst
```diff
@@ -0,0 +1,189 @@
...
+resulting file is just a few lines of code. A full example can be found in ``NetworkParametersCopier.kt`` in the source
+tree, but a flavour of it looks like this:
+
+.. container:: codeset
+
+   .. sourcecode:: java
+
+      NetworkParameters networkParameters = new NetworkParameters(
+                4,                        // minPlatformVersion
+                Collections.emptyList(),  // notaries
```
gendal: what goes in the list?

docs/source/connecting-to-a-compatibility-zone.rst
```diff
@@ -0,0 +1,50 @@
...
+  entities that have the same or similar brand names.
+
+* **emailAddress** e.g. "admin@company.com"
+
+* **devMode** must be set to false
+
+* **networkServices** or **compatibilityZoneURL** The Corda compatibility zone services must be configured. This must be either:
+
+  * **compatibilityZoneURL** The Corda compatibility zone network management service root URL.
+  * **networkServices** Replaces the ``compatibilityZoneURL`` when the doorman and network map services
```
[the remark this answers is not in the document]
joeldudleyr3: Addressed.

docs/source/setting-up-a-dynamic-compatibility-zone.rst
joeldudleyr3: Addressed.

docs/source/connecting-to-a-compatibility-zone.rst
```diff
@@ -0,0 +1,50 @@
...
+    are configured to operate on different URL endpoints. The ``doorman`` entry is used for registration.
+
+A new pair of private and public keys generated by the Corda node will be used to create the request.
```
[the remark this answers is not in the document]
kchalkias: @gendal key pairs are generated from the CryptoServiceImpl (i.e., HSM) obviously before you send the request, as you need to sign the CSR. This process has not been changed.
However, CryptoService does not deal with certificates; when we create our NodeCA key pair for the first time, we typically store a self-signed cert (this will never be used as certificate, but it's required to extract the public key). So key-entries are of the form `(alias, privateKey, certificate)`.

The main difference is how we store the certificate chain, which is received from Doorman. Usually you would expect that we update the certificate part of the already stored key. However, many underlying keystore implementations (including PKCS12 and HSMs) CANNOT store more than one certificate (so no chain is supported, while some of them do not allow updates on the cert). Due to above, we still use the `signingCertStore` JKS keystore, but its purpose now is to store certificate chains under the same key alias. 
Practically, CryptoServices store `KeyPair`s and in Corda we additionally have a JKS keystore to store their certificate chains under the same alias.

Fortunately, JKS allows storing certificate chains of any size, so this works.
Note that the only exception is the default `BCCryptoService` (BouncyCastle) which for backwards compatibility purposes reuses the `signingCertStore` to store and update both keys and certs.

docs/source/joining-a-compatibility-zone.rst
```diff
@@ -0,0 +1,56 @@
...
+    network map service are operating on the same URL endpoint, where:
+
+    * **doormanURL** is the root address of the doorman. This is the address used for initial registration
+    * **networkMapURL** is the root address of the network map service
+
+A new pair of private and public keys generated by the Corda node will be used to create the request.
+
+The utility will submit the request to the doorman server and poll for a result periodically to retrieve the
+certificates. Once the request has been approved and the certificates downloaded from the server, the node will create
+the keystore and trust store using the certificates and the generated private key.
```
kchalkias: With the CryptoService, this is not entirely true, as the underlying CryptoService might not use Keystores. As mentioned in my previous comment. The private key has already been generated ans stored inside the CryptoService (Corda does not know if this service uses an underlying keystore or not). However, the cert-chain received from Doorman is stored in the keystore, you never need to update the private key entry in the CryptoService.

docs/source/setting-up-a-dynamic-compatibility-zone.rst
```diff
@@ -0,0 +1,189 @@
...
+
+Do you need to create your own dynamic compatibility zone?
+----------------------------------------------------------
+
+By *dynamic compatibility zone*, we mean a compatibility zone that relies on a network map server to allow nodes to
+join dynamically, instead of requiring each node to be bootstrapped and have the node-infos distributed manually. While
+this may sound appealing, think twice before going down this route:
+
+1. It isn't necessary for testing.
+2. It isn't necessary for adding another layer of permissioning or 'know your customer' requirements onto your app.
```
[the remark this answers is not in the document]
joeldudleyr3: Addressed.

docs/source/joining-a-compatibility-zone.rst
joeldudleyr3: Addressed.

docs/source/compatibility-zones.rst
```diff
@@ -4,54 +4,29 @@
    <script type="text/javascript" src="_static/jquery.js"></script>
    <script type="text/javascript" src="_static/codesets.js"></script>
 
-Compatibility zones
```
kchalkias: QQ: although we mention how NodeCA key is created during `--initial-registration`, we never mention how legal identity keys are generated, i.e., to update our NodeInfo to network map, we sign with the legal identity key, not NodeCA, but we never mention the two difference between these two key types.

docs/source/joining-a-compatibility-zone.rst
```diff
@@ -0,0 +1,58 @@
...
+       * **doormanURL** is the root address of the doorman. This is the address used for initial registration
+       * **networkMapURL** is the root address of the network map service
+
+2. It generates a new private/public keypair to sign the certificate signing request
+
+3. It submits the request to the doorman server and polls periodically to retrieve the corresponding certificates
+
+4. It creates the node's keystore and trust store using the received certificates
+
+5. It creates and stores the node's TLS keys and legal identity key
```
kchalkias: just a final comment:

> It creates and stores the node's TLS and legal identity keys along with their corresponding certificate-chains

joeldudleyr3: Addressed.

PR #4158, opened 2018-11-02
[CORDA-2130] Encumbered states should always be assigned to the same notary

kchalkias:
Although we already check that ALL output states should be assigned to the same Notary in `LedgerTransaction.checkNoNotaryChange`, this only applies if transaction's notary is NOT null.
However, in special cases, like issuing transactions, a transaction might have a null Notary.

Previously, it was possible that an issuer could create encumbered output states assigned to different notaries which would result to freezing all encumbered states as we don't yet support multi-notary transactions. Thus, such encumbered states could NOT be consumed in the same transaction in the future.

This PR checks if all related encumbered states are assigned to the same notary (if transaction's notary is null). The word "related" is important as a transaction can have multiple encumbrance chains and we handle each chain/group separately.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
@@ -137,9 +137,9 @@ sealed class TransactionVerificationException(val txId: SecureHash, message: Str
             "a full cycle. Offending indices $nonMatching", null)
 
     /**
-     * If two encumbered states are assigned to different notaries, then these states cannot be encumbered. This is
-     * due to the fact that multi-notary transactions are not supported yet and thus two encumbered states with different
-     * notaries cannot be consumed in the same transaction.
+     * All encumbered states should be assigned to the same notary. This is due to the fact that multi-notary
+     * transactions are not supported yet and thus two encumbered states with different notaries cannot be consumed
+     * in the same transaction.
      */
     @KeepForDJVM
     class TransactionNotaryMismatchEncumbranceException(txId: SecureHash, encumberedIndex: Int, encumbranceIndex: Int, encumberedNotary: Party, encumbranceNotary: Party)
--- a/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
@@ -136,6 +136,16 @@ sealed class TransactionVerificationException(val txId: SecureHash, message: Str
             "is not satisfied. Encumbered states should also be referenced as an encumbrance of another state to form " +
             "a full cycle. Offending indices $nonMatching", null)
 
+    /**
+     * If two encumbered states are assigned to different notaries, then these states cannot be encumbered. This is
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
```diff
@@ -136,6 +136,16 @@ sealed class TransactionVerificationException(val txId: SecureHash, message: Str
             "is not satisfied. Encumbered states should also be referenced as an encumbrance of another state to form " +
             "a full cycle. Offending indices $nonMatching", null)
 
+    /**
+     * All encumbered states should be assigned to the same notary. This is due to the fact that multi-notary
+     * transactions are not supported yet and thus two encumbered states with different notaries cannot be consumed
```
adagys: Maybe remove the "yet" to avoid confusion 😄 We may end up never implementing multi-notary transactions

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -229,10 +230,39 @@ data class LedgerTransaction @JvmOverloads constructor(
...
+        }
+    }
+
+    private fun checkNotary(index: Int, indicesAlreadyChecked: HashSet<Int>) {
+        if (indicesAlreadyChecked.add(index)) {
+            val encumbranceIndex = outputs[index].encumbrance!!
+            if (outputs[index].notary != outputs[encumbranceIndex].notary) {
+                throw TransactionVerificationException.TransactionNotaryMismatchEncumbranceException(id, index, encumbranceIndex, outputs[index].notary, outputs[encumbranceIndex].notary)
+            } else {
+                checkNotary(encumbranceIndex, indicesAlreadyChecked)
```
adagys: maybe use a loop instead of recursion? if we have say 10_000 states in an encumbrance chain it might have to go quite deep

adagys: actually you can probably just mark the function as `tailrec`

core/src/test/kotlin/net/corda/core/transactions/TransactionEncumbranceTests.kt
```diff
@@ -316,4 +314,41 @@ class TransactionEncumbranceTests {
...
+                    .addOutputState(stateWithNewOwner, Cash.PROGRAM_ID, DUMMY_NOTARY, 2, AutomaticHashConstraint)
+                    .addCommand(Cash.Commands.Issue(), MEGA_CORP.owningKey)
+                    .toLedgerTransaction(ledgerServices)
+        }
+
+        // Two different encumbrance chains, where only one fails due to mismatched notary.
+        // 0 -> 1, 1 -> 0, 2 -> 3, 3 -> 2 where encumbered states with indices 2 and 3, respectively, are assigned
+        // to different notaries.
+        assertFailsWith<TransactionVerificationException.TransactionNotaryMismatchEncumbranceException> {
+            TransactionBuilder()
```
adagys: is it worth also checking that the mismatch happens on the expected indices?

PR #4166, opened 2018-11-05
CORDA-2187 reverting maxTransactionSize <= maxMessageSize

mkit:
This PR removes the `maxTransactionSize` constraint from the network parameters.


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -375,7 +375,7 @@ internal constructor(private val initSerEnv: Boolean,
                     notaries = notaryInfos,
                     modifiedTime = Instant.now(),
                     maxMessageSize = 10485760,
-                    maxTransactionSize = 10485760,
+                    maxTransactionSize = 104857600,
```
shamsasari: Actually, make this `50 * maxMessageSize`

PR #4173, opened 2018-11-06
CORDA-2191: Ignoring invalid test (works on Linux not on Windows)

anthonykeenan:


tools/bootstrapper/src/test/kotlin/net/corda/bootstrapper/PackageOwnerParsingTest.kt
```diff
@@ -124,6 +121,7 @@ class PackageOwnerParsingTest {
         assertThat(networkBootstrapper.registerPackageOwnership).hasSize(3)
     }
 
+    @Ignore("Ignoring this test as the delimiters don't work correctly, see CORDA-2191")
```
josecoll: According to Christian, the problem isn't with the delimiters per se.
Its the parsing of the parameters passed to the keytool (which subsequently fails the latter test).

anthonykeenan: Christian is right, the reason the windows test fails is because the keytool is stripping the outer quotes from the password.

For example, in Linux, ``keytool --password "password"`` sets the password on the keystore to ``"password"``

In Windows, ``keytool --password "password"`` sets the password on the keystore to `password`, as windows automatically strips quotes from commandline options.

However, the test is testing the wrong thing. For example, if the following is sent:

``bootstrapper --register-password-owner=net.something;keystore.jks;"pasword";blah``

The password that should be expected on the keystore is ``password``, not ``"password"``. The `"` should be stripped from the input.

PR #4174, opened 2018-11-06
CORDA-2191 - Revert Windows password parsing edge cases.

josecoll:
Remove password parsing edge cases failing on Windows (due to keytool execution mix-up)

De-scoping further edge cases parsing changed in lieue of https://r3-cev.atlassian.net/browse/CORDA-2141


anthonykeenan:
I've already created a PR to ignore the test case here: https://github.com/corda/corda/pull/4173

I think it's probably better to ignore rather than revert in case we decide to keep this method and we need to fix it, to help understand the underlying problem.

josecoll:
Covered by other PR #4173

PR #4178, opened 2018-11-06
CORDA-2183: Show root cause of error in console to aid with debugging

anthonykeenan:
Was:

`[ERROR] 15:29:08+0000 [main] internal.Node.installCordaServices - Corda service net.corda.finance.internal.ConfigHolder failed to instantiate
Cordapp configuration is incorrect due to exception`

Now displays (well it would if the bug still existed):

``[ERROR] 17:07:26+0000 [main] internal.Node.installCordaServices - Corda service net.corda.finance.internal.ConfigHolder failed to instantiate (No configuration setting found for key 'issuableCurrencies') [errorCode=ut8i8s]``

Also fixed an issue with newlines in the default log4j confg when printing throwables in the console.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -554,7 +554,11 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
                 log.error("${it.name}, as a Corda service, must have a constructor with a single parameter of type " +
                         ServiceHub::class.java.name)
             } catch (e: ServiceInstantiationException) {
-                log.error("Corda service ${it.name} failed to instantiate. Reason was: ${e.cause?.rootMessage}", e.cause)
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -554,7 +554,7 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
                 log.error("${it.name}, as a Corda service, must have a constructor with a single parameter of type " +
                         ServiceHub::class.java.name)
             } catch (e: ServiceInstantiationException) {
-                log.error("Corda service ${it.name} failed to instantiate", e.cause)
+                log.error("Corda service ${it.name} failed to instantiate (${e.cause?.rootMessage})", e.cause)
```
tlil: Perhaps have a fallback here in case `e.cause` is `null`?

PR #4180, opened 2018-11-06
add documentation for initiating flow modification

roastario:


roastario:
Intended to be used alongside: https://github.com/corda/extendable-example

PR #4181, opened 2018-11-06
CORDA-2129: Attempts to initiate a flow that's unknown is paused and instead sent to the flow hospital

shamsasari:
This allows for the case of a missing CorDapp to be installed on node restart and recover the original flow initiate request.


gendal:
I love how far you're taking the hospital analogies...  I guess you've had recent experience :)

What's the documentation plan for this? Seems like some pretty clear operational docs will be needed...  eg should people deploying Corda be employing a registrar/consultant/matron(?!) to walk the wards of the flow hospital each day and taking action?

PR #4183, opened 2018-11-07
[CORDA-2195] Replace @return with Returns due to rendering issues.

kchalkias:
Annotation only kdocs/javadocs don't render that great.
See https://docs.corda.net/head/codestyle.html#comments (second bullet in the second list).


core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
```diff
@@ -562,7 +572,8 @@ object Crypto {
      * It returns true if it succeeds, but it always throws an exception if verification fails.
      * @param txId transaction's id.
      * @param transactionSignature the signature on the transaction.
-     * @return true if verification passes or throw exception if verification fails.
+     *
+     * true if verification passes or throw exception if verification fails.
```
mikehearn: Seems "Returns" went missing here?

core/src/main/kotlin/net/corda/core/crypto/Crypto.kt
```diff
@@ -492,7 +499,8 @@ object Crypto {
      * @param publicKey the signer's [PublicKey].
      * @param signatureData the signatureData on a message.
      * @param clearData the clear data/message that was signed (usually the Merkle root).
-     * @return true if verification passes or throws an exception if verification fails.
+     *
+     * Returns true if verification passes or throws an exception if verification fails.
```
mikehearn: Does text that appears after `@param` work right? It's a bit unusual but I know this was done automatically so no problem.

mikehearn:
So technically the guidance is for kdocs/javadocs that *only* contain a `@return` tag. Not all uses of it! Was this PR much work to create? Is it easy to restrict the script to only looking at Kdocs that are exclusively `@return` tags i.e. match against `/** (\n +*|) @return` or something like that.

mikehearn (changes requested):

kchalkias:
@mikehearn It was originally automatic yes, but then I had to rewrite/rephrase a few of them (fixed some typos while checking as well, missing periods etc), so eventually it was mixed work and did it while waiting for CryptoService TC tests.
I will change according to your recommendation, during another TC test round :)

kchalkias:
Back to you @mikehearn

shamsasari:
Unfortunately this has been neglected. Best to close and retry later.

PR #4184, opened 2018-11-07
CORDA-2106: Print node info gen log on bootstrap failure

roastario:


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -96,21 +97,30 @@ internal constructor(private val initSerEnv: Boolean,
...
-            check(process.exitValue() == 0) { "Error while generating node info file. Please check the logs in $logsDir." }
+            printNodeInfoGenLogToConsole(nodeInfoGenFile) { process.exitValue() == 0 }
             return nodeDir.list { paths ->
                 paths.filter { it.fileName.toString().startsWith(NODE_INFO_FILE_NAME_PREFIX) }.findFirst().get()
             }
         }
+
+        private fun printNodeInfoGenLogToConsole(nodeInfoGenFile: File, check: (() -> Boolean) = { true }) {
+            if (!check.invoke()) {
+                System.err.println("Error while generating node info file")
```
anthonykeenan: You need to say which node it is that failed first I think, I know it normally says it once the node starts up but it might not have got that far

roastario:
Added logic to print out either the node's legal name, or the node's directory depending on which is available.

PR #4132, opened 2018-10-29
[CORDA-1993]: Replace reflection-based NodeConfiguration parsing with versioned property-based parsing mechanism.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1993

Notes:

- Extracted `NodeConfigurationImpl` into its own file.
- Still use `NodeConfigurationImpl.validate` for validation, as there's no point to do this differently for version 1.
- If the configuration is valid, the actual configuration will be logged with DEBUG level and sensitive data masked.
- At the moment, the assumption is that the node conf version is `1`, we can add the logic to parse the value after we come up with a version 2.

PR #4191, opened 2018-11-07
[CORDA-1879]: Prevent Corda nodes from trying to recover from unrecoverable errors.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1879

Notes:

- I've been lenient in same cases where the `catch(t: Throwable)` was inside tests.
- I have branched from another branch of mine by mistake. Only commit https://github.com/corda/corda/pull/4191/commits/ca3de43be5a55cb14ed9c0bbced79122cdf0b6e6 should be reviewed. I'll wait for https://github.com/corda/corda/pull/4132 to go in and then rebase this one.


tlil:
Why does this come up with so many commits? Rebase off of `master`? Also, gates are red.

core/src/main/kotlin/net/corda/core/contracts/Attachment.kt
```diff
@@ -36,10 +37,11 @@ interface Attachment : NamedByHash {
...
     fun openAsJAR(): JarInputStream {
         val stream = open()
-        try {
-            return JarInputStream(stream)
-        } catch (t: Throwable) {
-            stream.use { throw t }
+        return try {
+            JarInputStream(stream)
+        } catch (e: IOException) {
+            stream.close()
```
tlil: I would keep this `stream.use { throw t }`

From `use`'s kdoc:
> Executes the given [block] function on this resource and then closes it down correctly whether an exception is thrown or not.

See also its implementation - it does more than just calling `.close()`

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -554,7 +554,7 @@ class RPCClientProxyHandler(
             observationExecutorPool.run(k) {
                 try {
                     m[k]?.onError(ConnectionFailureException())
-                } catch (th: Throwable) {
```
tlil: `s/th/e/`?

settings.gradle
```diff
@@ -21,7 +21,6 @@ include 'experimental'
 include 'experimental:avalanche'
 include 'experimental:behave'
 include 'experimental:quasar-hook'
-include 'experimental:kryo-hook'
```
tlil: Why is this suddenly being removed from this PR?

tlil:
What testing have you done with this? Can you run some tests where you trigger an OOM in the state machine for instance and show that it behaves as expected, please?

settings.gradle
exFalso: I recommended this. I wrote that tool for analyzing the contents of kryo-serialized messages to debug abnormally large blobs. We're not using it anymore, and if we do need it (e.g. checkpoints are getting too large) we can just dig it up from git history. Otherwise it just creates development drag

tlil: Kk, fine by me. Just see the relevance to CORDA-1879. But fine

exFalso:
The statemachine related changes look fine.

Also I've never seen such a git history, what sequence of commands results in such duplication of commit messages?

core/src/main/kotlin/net/corda/core/contracts/Attachment.kt
sollecitom: Good one, fixed.

client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
sollecitom: Done

sollecitom:
@exFalso no GIT commands involved. I was absent-minded and branched off another branch I was working on.

Given that this carries like 2 commits on its own, we can merge it and GIT will ignore the duplicated commits from history.

PR #4195, opened 2018-11-08
ENT-2695 Restore async logging

rick-r3:
Replace error code generation technique with custom event re-writer. Switch to RandomAccessFile appenders because they supposedly give higher throughput.

Also added a small unit test for async logging.


testing/test-common/src/main/resources/log4j2-test.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Configuration status="info">
+<Configuration status="info" packages="net.corda.cliutils">
```
sollecitom: Nice :) out of curiosity, why do we need to different `log4j2.xml` files and, in particular, one for the driver?

tools/cliutils/src/main/kotlin/net/corda/cliutils/ErrorCodeRewritePolicy.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.logging.log4j.core.config.plugins.PluginFactory
+import org.apache.logging.log4j.core.impl.Log4jLogEvent
+
+@Plugin(name = "ErrorCodeRewritePolicy", category = Core.CATEGORY_NAME, elementType = "rewritePolicy", printObject = false)
+class ErrorCodeRewritePolicy : RewritePolicy {
+
+    override fun rewrite(source: LogEvent?): LogEvent? {
+        val newMessage = source?.message?.withErrorCodeFor(source.thrown, source.level)
+        if (newMessage == source?.message) {
+            return source
```
sollecitom: You can lift the `return` outside of the `if` block, and maybe replace it with `when`.

testing/test-common/src/main/resources/log4j2-test.xml
rick-r3: Something to do with team city not working well with the other config, apparently.

tools/cliutils/src/main/kotlin/net/corda/cliutils/ErrorCodeRewritePolicy.kt
rick-r3: Will pull out return.  Will stick with `if` when it's a binary choice.

PR #4211, opened 2018-11-10
[CORDA-2210] Add matching party on the first X500 attribute match

kchalkias:
Previously, a `partiesFromName` `query` used to traverse all X500 attributes, even if a match had already been found. It's true that because we use a `Set` there were not duplicate entries, but we should indeed stop after the first match.
The following has been applied and the function was renamed to `x500Matches`:
PS: We could also refactor this method to just return true or false and not update the `result`. I didn't do as it's not marked as internal, although its name states so ???


shamsasari:
Yeah, it probably makes sense to have the method return a boolean. It's internal so it's fine.

kchalkias:
Back to you @shamsasari, renamed the function as well.

PR #4213, opened 2018-11-12
[CORDA-1879]: Ensure Node dies on unrecoverable errors.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1879

@tlil tested it with `ErrorCodeLoggingTests.OOM exception()`, which I'm going to remove in a sec.
(removed, still visible in the first commit, for reference)


tlil:
Provided that tests pass :)

PR #4214, opened 2018-11-12
Fixed missing entries for common modules in parent `build.gradle` file.

sollecitom:


build.gradle
```diff
@@ -366,7 +366,9 @@ bintrayConfig {
             'corda-tools-network-bootstrapper',
             'corda-tools-cliutils',
             'corda-notary-raft',
-            'corda-notary-bft-smart'
+            'corda-notary-bft-smart',
+            'common-configuration-parsing',
+            'common-validation'
```
chrisr3: Don't these violate our naming convention of Corda artifacts starting with `corda-`?

sollecitom: Added the "corda-" prefix. Mind another look?

PR #4215, opened 2018-11-12
CORDA-1817: Fix issue with misleading capsule error messages

anthonykeenan:
If there are errors creating the cordapp directory, rather than throw an error when modifying the classpath, just return the existing classpath. When the main method is invoked in the `Corda.jar` it will check again and fail gracefully.

Also bring capsule in line with the options accepted by the main Corda jar.

### DIFF
--- a/node/build.gradle
+++ b/node/build.gradle
@@ -151,6 +151,7 @@ dependencies {
     // Capsule is a library for building independently executable fat JARs.
     // We only need this dependency to compile our Caplet against.
     compileOnly "co.paralleluniverse:capsule:$capsule_version"
+    testCompile "co.paralleluniverse:capsule:$capsule_version"
 
     // OkHTTP: Simple HTTP library.
     compile "com.squareup.okhttp3:okhttp:$okhttp_version"
--- a/node/src/main/java/CordaCaplet.java
+++ b/node/src/main/java/CordaCaplet.java
@@ -20,8 +20,7 @@ public class CordaCaplet extends Capsule {
     }
 
     private Config parseConfigFile(List<String> args) {
-        String baseDirOption = getOption(args, "--base-directory");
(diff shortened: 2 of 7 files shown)


node/build.gradle
```diff
@@ -151,6 +151,7 @@ dependencies {
     // Capsule is a library for building independently executable fat JARs.
     // We only need this dependency to compile our Caplet against.
     compileOnly "co.paralleluniverse:capsule:$capsule_version"
+    testCompile "co.paralleluniverse:capsule:$capsule_version"
```
sollecitom: I thought `compileOnly` dependencies were available within test scope as well. Am I wrong?

node/src/main/java/CordaCaplet.java
```diff
@@ -152,17 +181,20 @@ private static void checkJavaVersion() {
...
                 logOnFailedCordappDir();
-                throw new RuntimeException("Cordapps dir could not be created"); // Let Capsule handle the error (log error, clean up, die).
+                //throw new RuntimeException("Cordapps dir could not be created"); // Let Capsule handle the error (log error, clean up, die).
+                return false;
             }
         }
         catch (SecurityException | NullPointerException e) {
             logOnFailedCordappDir();
-            throw e; // Let Capsule handle the error (log error, clean up, die).
+            //throw e; // Let Capsule handle the error (log error, clean up, die).
```
sollecitom: What about this commented line?

node/build.gradle
anthonykeenan: Well it didn't compile without that dependency, so I guess so. Although I'm not sure whether there is a better way of doing it.

node/src/main/java/CordaCaplet.java
anthonykeenan: Removed

PR #4222, opened 2018-11-12
CORDA-1965 Auto-accept network parameter changes that modify whitelisted map or package ownership map

oknowles:
This PR enables certain network parameters to be automatically accepted by the node. The annotation `@AutoAcceptable` can be used within the `NetworkParameters` class to specify which parameters are eligible for this. This behaviour is switched on by default but can be turned off by setting `autoAcceptNetworkParameterChanges` to false within the node configuration file. Any change that modifies a parameter that is not auto-acceptable will require manual acceptance as before.

This _doesn't_ affect flag day behaviour - nodes will still require a manual restart. I investigated hot swapping of the network parameters and implemented a solution that seemed to achieve the desired effect, but upon further thought removed it. See below for more info.

**Hot Swap Investigation**
I originally investigated hot swapping the network parameters by adding a wrapper class (called `NetworkParameters`) around the old NetworkParameters data class (renamed to `NetworkParameterValues`). This would mean that externally the NetworkParameters object is interacted with in an identical way to previously but it would be possible to hot swap the underlying parameters by swapping out the internal data class. Although this is okay in some respects (e.g. if an object asks for the whitelist map then it will either get the old one or the new one, never a mixture), doing it this way means that if any object reads the map and then keeps hold of it for a long time then any subsequent hot swaps of this will not be detected. This is because the underlying object was swapped out rather than modified. If this was an acceptable consequence then it should not bottle neck the application as much as the alternative approach of using threadsafe objects for certain parameters (such as ConcurrentHashMaps etc)/locks. 

However there are certain parameters which are only read on start up and make hot swapping not very viable. For example the `messageBroker` and `bridgeControlListener` are created using `maxMessageSize` once at node start up. Hence we cannot blindly hot swap out the network parameters as the changes would not necessarily be propagated to all services. 

The currently auto-acceptable parameters (whitelist and package name map) appear to currently be hot-swappable using the above method. That is, they are only used temporarily via reading them from the `networkParameters` object. Therefore when a flag day occurs it would be possible to check if the changes only include the auto acceptable parameters and hot swap them if so. However I worry that in doing this we would create an easy route for someone to introduce a hard to spot bug. For example adding the `@AutoAcceptable` annotation to something like the `maxMessageSize` would cause it to be '_hot swapped_' with no immediate errors but, as it wouldn't get updated in the `messageBroker`, unexpected behaviour would occur down the line.


core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -7,10 +7,16 @@ import net.corda.core.node.services.AttachmentId
...
+import java.lang.annotation.Inherited
 import java.security.PublicKey
 import java.time.Duration
 import java.time.Instant
+import kotlin.reflect.KProperty1
+import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.full.findAnnotation
+import kotlin.reflect.jvm.javaGetter
 
+// DOCSTART 1
```
fenryka: Probably should mention which properties have been modified to be auto accept in the jdoc

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -34,12 +40,12 @@ data class NetworkParameters(
...
-        val modifiedTime: Instant,
-        val epoch: Int,
-        val whitelistedContractImplementations: Map<String, List<AttachmentId>>,
+        @AutoAcceptable val modifiedTime: Instant,
+        @AutoAcceptable val epoch: Int,
+        @AutoAcceptable val whitelistedContractImplementations: Map<String, List<AttachmentId>>,
         val eventHorizon: Duration,
-        val packageOwnership: Map<JavaPackageName, PublicKey>
-) {
+        @AutoAcceptable val packageOwnership: Map<JavaPackageName, PublicKey>) {
```
fenryka: ) {    on next line

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -149,6 +155,28 @@ data class NetworkParameters(
      * Returns the public key of the package owner of the [contractClassName], or null if not owned.
      */
     fun getOwnerOf(contractClassName: String): PublicKey? = this.packageOwnership.filterKeys { it.owns(contractClassName) }.values.singleOrNull()
+
+    /**
+     * Returns true if all non-matching variables have the annotation @AutoAcceptable
```
fenryka: This comment doesn't make a great deal of sense :)

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -149,6 +155,28 @@ data class NetworkParameters(
...
+     */
+    fun canAutoAccept(newNetworkParameters: NetworkParameters): Boolean {
+        for (property in this::class.declaredMemberProperties) {
```
fenryka: Reflection is slow, could probably pre-capture all the auto-accept fields and then just invoke them

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -149,6 +155,28 @@ data class NetworkParameters(
...
+
+            if (propertyValue != newPropertyValue && !property.isAutoAcceptable()) {
+                return false
+            }
+        }
+        return true
+    }
+
+    fun getAutoAcceptableParamNames(): List<String> {
+        return this::class.declaredMemberProperties
```
fenryka: should probably wrap this as a sequence before the filter and map?

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -194,3 +222,12 @@ private fun isPackageValid(packageName: String): Boolean = packageName.isNotEmpt
 private fun noOverlap(packages: Collection<JavaPackageName>) = packages.all { currentPackage ->
     packages.none { otherPackage -> otherPackage != currentPackage && otherPackage.name.startsWith("${currentPackage.name}.") }
 }
+
+private fun KProperty1<out NetworkParameters, Any?>.isAutoAcceptable(): Boolean {
+    return this.findAnnotation<AutoAcceptable>() != null
+}
+
+@Target(AnnotationTarget.PROPERTY)
```
fenryka: should be in it's own file
should be jdoc'd

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -45,26 +45,43 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+        this.keyManagementService = keyManagementService
+        this.autoAcceptNetworkParameters = autoAcceptNetworkParameters
         watchForNodeInfoFiles()
         if (networkMapClient != null) {
             watchHttpNetworkMap()
         }
     }
 
+    fun getNetworkParameters(): NetworkParameters {
+        return networkParameters
```
fenryka: make single expression func!

docs/source/corda-configuration-file.rst
```diff
@@ -244,6 +244,9 @@ absolute path to the node's base directory.
                                        The option takes effect only in production mode and defaults to Corda development keys (``["56CA54E803CB87C8472EBD3FBC6A2F1876E814CEEBF74860BD46997F40729367",
                                        "83088052AF16700457AE2C978A7D8AC38DD6A7C713539D00B897CD03A5E5D31D"]``), in development mode any key is allowed to sign Cordpapp JARs.
 
+:autoAcceptNetworkParameterChanges: This flag toggles auto accepting of network parameter changes and is enabled by default. If a network operator issues a network parameter change which modifies
```
fenryka: we did talk about adding finer grained control of what to auto-accept *as well as* disabling and enabling it as a whole

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -335,7 +335,10 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
         val (nodeInfo, signedNodeInfo) = nodeInfoAndSigned
         identityService.ourNames = nodeInfo.legalIdentities.map { it.name }.toSet()
         services.start(nodeInfo, netParams)
-        networkMapUpdater.start(trustRoot, signedNetParams.raw.hash, signedNodeInfo.raw.hash)
+        networkMapUpdater.start(trustRoot, signedNetParams.raw.hash, signedNodeInfo, netParams, keyManagementService, configuration.autoAcceptNetworkParameterChanges)
+        if (configuration.autoAcceptNetworkParameterChanges) {
```
fenryka: shouldn't we log this in the network map updater?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -45,26 +45,43 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+              currentParametersHash: SecureHash,
+              ourNodeInfo: SignedNodeInfo,
+              networkParameters: NetworkParameters,
+              keyManagementService: KeyManagementService,
+              autoAcceptNetworkParameters: Boolean) {
         require(fileWatcherSubscription == null) { "Should not call this method twice." }
         this.trustRoot = trustRoot
         this.currentParametersHash = currentParametersHash
-        this.ourNodeInfoHash = ourNodeInfoHash
+        this.ourNodeInfo = ourNodeInfo
```
fenryka: all this "this" makes my spidy senses tingle... seems wrong somewhow, should this be an object that wraps the thread? it's fine, but it seems "un-neat" if that makes sense

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
oknowles: Good point, will update to remove repeated use of reflection

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
oknowles: This temporary getter was accidentally left in. Will remove

docs/source/corda-configuration-file.rst
oknowles: I started implementing this but initially worried that it might be a bit fragile to change, as well as slightly less straightforward for the user. I will revisit today

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
oknowles: Yes that would make more sense. I'll move it

oknowles:
@fenryka it should be ready for another review now.

I addressed most of your comments in commit [bf94552](https://github.com/corda/corda/pull/4222/commits/bf94552fdeeb64f8c1032edbdf87451b50da8b2a) and made the change to add per parameter configurability in commit [9230501](https://github.com/corda/corda/pull/4222/commits/9230501feddf865bc365f8f8cdb701f79d8a83b2). It feels slightly brittle and makes the code slightly clunkier, so still not 100% convinced if the extra granularity is worth it, but let me know your thoughts.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -335,7 +335,14 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
...
         identityService.ourNames = nodeInfo.legalIdentities.map { it.name }.toSet()
         services.start(nodeInfo, netParams)
-        networkMapUpdater.start(trustRoot, signedNetParams.raw.hash, signedNodeInfo.raw.hash)
+        networkMapUpdater.start(
+                trustRoot,
+                signedNetParams.raw.hash,
+                signedNodeInfo,
+                netParams,
+                keyManagementService,
+                configuration.autoAcceptNetworkParameterChanges,
```
fenryka: these two should probably be on the same object

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
oknowles: The cause of this is that this class gets instantiated in two parts. There are a few `lateinit` variables that only get instantiated when `start()` is called (which is where all these `this` are).

PR #4223, opened 2018-11-12
Add Docker image as output of build

roastario:


docker/src/docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    useradd corda -g corda -m -d /opt/corda
+
+# Copy corda jar
+
+ADD --chown=corda:corda corda.jar /opt/corda/corda.jar
+ADD --chown=corda:corda start.sh /start.sh
+RUN chmod +x /start.sh
+
+WORKDIR /opt/corda
+CMD ["start.sh"]
```
tlil: Can you make sure you add a way for passing explicit `java` arguments as well, say for instance for controlling heap memory, JITing and GC.

tlil:
Can you also refenence the originating JIRA ticket, please @roastario?

docker/src/config/node.conf
```diff
@@ -0,0 +1,25 @@
...
+                    permissions=[
+                        ALL
+                    ]
+                    user=rpcUser
+                }
+            ]
+        }
+    }
+}
+compatibilityZoneURL = ${COMPATIBILITY_ZONE}
```
wawrzek: What about network where NetworkMap is separated from Doorman?

docker/src/docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM azul/zulu-openjdk:8u192
```
wawrzek: Not sure if we should specify a version here or latest.

docker/src/docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+RUN apt-get update && apt-get install bash
```
wawrzek: Do we need bash?

docker/src/docker/Dockerfile
przemolb: IMHO by specifying version we make images predictable in production and testing. If we use `latest` and in the meantime azul changes their image we may get different behavior of the image.

docker/src/docker/Dockerfile
cburlinchon: entrypoint i think

docker/src/docker/Dockerfile
wawrzek: That's true. On the other hand I wonder what's the impact of this line for release process. Are we going to keep u192 forever for version 4? Are we going to update underling image. How are we going to mark various 'upgrades'. How does other address this problem?

wawrzek:
> Can you also refenence the originating JIRA ticket, please @roastario?

FYI I moved ticket to CORDA so it's now: https://r3-cev.atlassian.net/browse/CORDA-2214

docker/src/docker/Dockerfile
mikehearn: We should bump it when we do a point release.

przemolb: If we stick with particular numbers we control when to upgrade. If we relay on latest the upgrade process is out of our control. So if we control it and after i.e.: a month from a new jdk release we decide to go for it - it is up to us. Do we have to decide now how often we will upgrade ?

docker/src/docker/Dockerfile
```diff
@@ -1,22 +1,40 @@
 FROM azul/zulu-openjdk:8u192
 
-RUN apt-get update && apt-get install bash
+RUN apt-get update && apt-get -y install bash curl unzip
 
 # Create dirs
-RUN mkdir -p /opt/corda/cordapps
-RUN mkdir -p /opt/corda/certificates
-RUN mkdir -p /opt/corda/drivers
+RUN mkdir -p /opt/corda/mount-point/cordapps
```
przemolb: What is the `mount-point` folder for ?

docker/src/docker/Dockerfile
roastario: if you use latest, you will get java 11.

docker/src/docker/Dockerfile
roastario: entrypoint limits you to a single executable. 
by using CMD you can then specify a different executable, I use this to allow generate-config.

docker/src/config/node.conf
roastario: yah, I've split them

docker/src/docker/Dockerfile
roastario: yes, it's used by the generate-config script.

docker/src/docker/Dockerfile
roastario: nothing, it was just an investigation.

docker/src/bash/example-generate-testnet.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+docker run -ti \
+        -e MY_PUBLIC_ADDRESS=stefano.azure.io \
```
wawrzek: Is this address right?

roastario: It's an example.

docker/src/docker/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+VOLUME ["/opt/corda/cordapps"]
+##PERSISTENCE FOLDER
+VOLUME ["/opt/corda/persistence"]
+##CERTS FOLDER
+VOLUME ["/opt/corda/certificates"]
+##OPTIONAL JDBC DRIVERS FOLDER
+VOLUME ["/opt/corda/drivers"]
+##CONFIG LOCATION
+VOLUME ["/etc/corda"]
+##LOG FOLDER
```
wawrzek: minor thing, but I would move logs volume lines  (39-40) about config volume - to be consistent in order we create those dirs and then set env.

docker/src/bash/run-corda.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+: ${JVM_ARGS='-Xmx15g -Xms1g -XX:+UseG1GC'}
```
wawrzek: Do we need any extra Java flags for properly handling memory in container? For example anyhing from: https://royvanrijn.com/blog/2018/05/java-and-docker-memory-limits/

docker/src/docker/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+VOLUME ["/opt/corda/logs"]
+
+##CORDA JAR
+ADD --chown=corda:corda corda.jar /opt/corda/bin/corda.jar
+##CONFIG MANIPULATOR JAR
+ADD --chown=corda:corda config-exporter.jar /opt/corda/config-exporter.jar
+##CONFIG GENERATOR SHELL SCRIPT
+ADD --chown=corda:corda generate-config.sh /opt/corda/bin/config-generator
```
wawrzek: I think this file name is unnecessary. I suggest to keep script names consistences between source and docker image. If we have `geneate-config.sh` in code, lets put `generate-config.sh` into image. (Unless there is a reason for this name change).

docker/src/docker/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+##CORDA RUN SCRIPT
+ADD --chown=corda:corda run-corda.sh /opt/corda/bin/run-corda
```
wawrzek: Lets keep .sh extension - to have consistence filename.

docker/src/bash/example-generate-testnet.sh
wawrzek: I suggest to change it to 'example.com' or something similar and add comment line at the top of the script `# PLEASE REMEMBER THAT YOU NEED TO REPLACE ALL ENVIRONMENT VARIABLE IN THE SCRIPT`

docker/src/bash/generate-config.sh
```diff
@@ -0,0 +1,117 @@
...
+
+function generateTestnetConfig() {
+    RPC_PASSWORD=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1) \
+    DB_PASSWORD=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1) \
+    MY_PUBLIC_ADDRESS=${MY_PUBLIC_ADDRESS} \
+    MY_P2P_PORT=${MY_P2P_PORT} \
+    MY_RPC_PORT=${MY_RPC_PORT} \
+    MY_RPC_ADMIN_PORT=${MY_RPC_ADMIN_PORT} \
+    COMPATIBILITY_ZONE='https://map.testnet.corda.network' \
+    DOORMAN_URL='https://doorman.testnet.corda.network' \
```
wawrzek: Does it work for TestNet. AFAIR doorman is hidden.

docker/src/bash/example-generate-testnet.sh
roastario: done.

docker/src/docker/Dockerfile
roastario: done

docker/src/bash/run-corda.sh
roastario: added! with documentation about flags.

docker/src/bash/generate-config.sh
roastario: Testnet is not configured in this way - you have to use the --testnet option for configurator.

If testnet doorman was open, it would be possible to configure using normal means.

docker/src/bash/example-join-generic-cz.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+##in this example the doorman will be running on the host machine on port 8080
+##so the container must be launched with "host" networking
+docker run -ti --net="host" \
+        -e  MY_LEGAL_NAME="O=EXAMPLE,L=Berlin,C=DE"     \
```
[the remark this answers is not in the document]
roastario: I've not found a reliable means to test if a volume is mounted from within the container. 

The node itself does a good job printing errors if the required files are not found. 

Will have a think about what to do if mounts are missing for config-generator.

docker/src/bash/generate-config.sh
```diff
@@ -84,7 +81,7 @@ while :; do
             if [[ ${GENERATE_TEST_NET} = 0 ]]; then
                 GENERATE_GENERIC=1
             else
-                die 'ERROR: "cannot generate config for multiple networks'
+                die 'ERROR: cannot generate config for multiple networks'
```
tlil: Similar fix is still needed on line 77

docker/src/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM azul/zulu-openjdk:8u192
 
-RUN apt-get update && apt-get -y install bash curl unzip
+RUN apt-get update && apt-get -y upgrade && apt-get -y install bash curl unzip
```
tlil: Why `upgrade`?

roastario: this will pickup any "fixed" packages. Ubuntu 18.04 has several CVE's fixed by doing apt-get upgrade.

I'm working on adding container scanning to our TC.

docker/src/bash/generate-config.sh
roastario: yarp.

docker/src/docker/Dockerfile
tlil: Rightio... Okay

docs/source/building-a-cordapp-index.rst
```diff
@@ -15,6 +15,7 @@ CorDapps
    upgrade-notes
    upgrading-cordapps
    secure-coding-guidelines
+   flow-overriding
```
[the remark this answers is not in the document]
tlil: Okay, I was just surprised to see it as part of the PR. But I see, that's fine then.

docker/src/bash/run-corda.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+: ${JVM_ARGS='-XX:+UseG1GC'}
+##IT IS IMPORTANT THAT WE DOCUMENT THE FACT THAT IN DOCKER, USERS SHOULD NOT SPECIFY -Xmx AS JAVA WILL TRANSPARENTLY SET MAXIMUM BASED ON CONTAINER LIMITS
+JVM_ARGS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap "${JVM_ARGS}
+
+java -Djava.security.egd=file:/dev/./urandom -Dcapsule.jvm.args="${JVM_ARGS}" -jar /opt/corda/bin/corda.jar --base-directory=/opt/corda --config-file=/etc/corda/node.conf
```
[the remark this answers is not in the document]
roastario: This might be one of those self-propagating errors, but almost all blogs around this (like: http://ruleoftech.com/2016/avoiding-jvm-delays-caused-by-random-number-generation) all have the . in the middle of the path. 

I can't see a difference.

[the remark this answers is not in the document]
tlil: Well, dot is just a no-op really in path resolution sense... Fine by me.. Was just curious if there was anything that I wasn't aware of.

PR #4224, opened 2018-11-13
[CORDA-2172]: Add link to error-codes-webapp in Corda.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2172

### DIFF
--- a/tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
+++ b/tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
@@ -10,7 +10,7 @@ import picocli.CommandLine
 class CordaVersionProvider : CommandLine.IVersionProvider {
     companion object {
         const val current_major_release = "4.0"
-        const val source_edition = "OS"
+        const val source_edition = "Open Source"
         private val editionCodes = mapOf("Open Source" to "OS", "Enterprise" to "ENT")
 
         private fun manifestValue(name: String): String? = if (Manifests.exists(name)) Manifests.read(name) else null
--- a/tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
+++ b/tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
@@ -9,29 +9,18 @@ import picocli.CommandLine
  */
 class CordaVersionProvider : CommandLine.IVersionProvider {
     companion object {
-        const val current_major_release = "4.0"
-        const val source_edition = "Open Source"
-        private val editionCodes = mapOf("Open Source" to "OS", "Enterprise" to "ENT")
+        private const val UNKNOWN = "Unknown"
+        const val current_major_release = "4.0-SNAPSHOT"
+        const val platformEditionCode = "OS"
 
         private fun manifestValue(name: String): String? = if (Manifests.exists(name)) Manifests.read(name) else null
 
-        val releaseVersion: String by lazy { manifestValue("Corda-Release-Version") ?: current_major_release }
-        val edition: String by lazy { manifestValue("Corda-Edition") ?: source_edition }
-        val revision: String by lazy { manifestValue("Corda-Revision") ?: "Unknown" }
-        val vendor: String by lazy { manifestValue("Corda-Vendor") ?: "Unknown" }
+        val releaseVersion: String by lazy { manifestValue("Corda-Release-Version") ?: UNKNOWN }
+        val revision: String by lazy { manifestValue("Corda-Revision") ?: UNKNOWN }
+        val vendor: String by lazy { manifestValue("Corda-Vendor") ?: UNKNOWN }
         val platformVersion: Int by lazy { manifestValue("Corda-Platform-Version")?.toInt() ?: 1 }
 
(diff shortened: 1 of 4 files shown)


tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
```diff
@@ -9,12 +9,29 @@ import picocli.CommandLine
...
+            val raw = releaseVersion
+            val parts = raw.removeSuffix("-SNAPSHOT").split(".")
+            try {
+                "${parts[0]}.${parts[1]}${parts.getOrNull(2)?.let { ".$it" } ?: ""}"
+            } catch (e: Exception) {
+                current_major_release
+            }
+        }
+
+        internal val platformEditionCode: String by lazy { editionCodes[edition] ?: source_edition }
```
tlil: Perhaps clarify this a bit more. Currently `source_edition` defaults to a short version of the edition (i.e., `OS`). However, if I read this correctly, `manifestValue("Corda-Edition")` gets a long version (`Open Source` or `Enterprise`).

So here you end up looking up `editionCodes["OS"]` if the manifest value is non-existent. But the map maps long editions to short ones. So they're inconsistent. I know you end up with the same result. But .....

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
```diff
@@ -9,12 +9,29 @@ import picocli.CommandLine
  */
 class CordaVersionProvider : CommandLine.IVersionProvider {
     companion object {
+        const val current_major_release = "4.0"
```
tlil: Why did you have to hard-code a fallback value here? Is there a missing manifest somewhere?

tools/cliutils/src/main/kotlin/net/corda/cliutils/ExceptionsErrorCodeFunctions.kt
```diff
@@ -8,11 +8,13 @@ import java.util.*
 internal fun Message.withErrorCodeFor(error: Throwable?, level: Level): Message {
 
     return when {
-        error != null && level.isInRange(Level.FATAL, Level.WARN) -> CompositeMessage("$formattedMessage [errorCode=${error.errorCode()}]", format, parameters, throwable)
+        error != null && level.isInRange(Level.FATAL, Level.WARN) -> CompositeMessage("$formattedMessage [errorCode=${error.errorCode()}, errorDescription=${error.errorCodeLocationUrl()}]", format, parameters, throwable)
```
tlil: `errorDescription` = URL? Perhaps use a different name for the key here.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
sollecitom: It wouldn't work from Intellij, etc.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
```diff
@@ -9,12 +9,29 @@ import picocli.CommandLine
...
         private fun manifestValue(name: String): String? = if (Manifests.exists(name)) Manifests.read(name) else null
 
-        val releaseVersion: String by lazy { manifestValue("Corda-Release-Version") ?: "Unknown" }
+        val releaseVersion: String by lazy { manifestValue("Corda-Release-Version") ?: current_major_release }
+        val edition: String by lazy { manifestValue("Corda-Edition") ?: source_edition }
         val revision: String by lazy { manifestValue("Corda-Revision") ?: "Unknown" }
         val vendor: String by lazy { manifestValue("Corda-Vendor") ?: "Unknown" }
         val platformVersion: Int by lazy { manifestValue("Corda-Platform-Version")?.toInt() ?: 1 }
+
+        internal val semanticVersion: String by lazy {
```
anthonykeenan: As discussed, I think the release version should be used in the error codes url to distinguish between released binaries and locally compiled/snapshot binaries, as we are likely to grep the website logs for frequently ocurring issues in our released versions, which we wouldn't want to get confused with development issues.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
sollecitom: Fixeed

tools/cliutils/src/main/kotlin/net/corda/cliutils/ExceptionsErrorCodeFunctions.kt
sollecitom: Renamed to `moreInformationAt`.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaVersionProvider.kt
sollecitom: Fixed

sollecitom: Also, modified the error codes web app to handle "-SNAPSHOT" suffix.

PR #4228, opened 2018-11-13
CORDA-2089 - network parameters tags - part

kasiastreich:
Data structures changes, storage and notarisation.

Tag transactions with network parameters hash that was in force when tx
was created. Add component group on all core transactions and resolved
parameters on full transactions. The hash should be always visible on
the filtered versions of transactions. Add
notarisation check that the parameters are current.
Implement network parameters storage on services for resolution.

This is only part of the work, next PR will include changes to
ResolveTransactionsFlow to make sure that parameters in the transaction
graph are ordered (this is to prevent the downgrade attack, when the
malicious notary and participants sign transaction that shouldn't be
notarised otherwise).
Probably on network services side we need the default parameters
endpoint for the
transactions that were created before this change - for now it's default
to the current ones.

### DIFF
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
@@ -188,7 +188,8 @@ private class WireTransactionSerializer : JsonSerializer<WireTransaction>() {
                 value.timeWindow,
                 value.attachments,
                 value.privacySalt,
-                value.networkParametersHash
+                value.networkParametersHash,
+                value.references
         ))
     }
 }
@@ -203,7 +204,8 @@ private class WireTransactionDeserializer : JsonDeserializer<WireTransaction>() 
                 wrapper.attachments,
                 wrapper.notary,
                 wrapper.timeWindow,
-                networkParametersHash = wrapper.networkParametersHash
+                wrapper.references,
+                wrapper.networkParametersHash
         )
         return WireTransaction(componentGroups, wrapper.privacySalt)
     }
@@ -217,7 +219,8 @@ private class WireTransactionJson(val id: SecureHash,
                                   val timeWindow: TimeWindow?,
                                   val attachments: List<SecureHash>,
                                   val privacySalt: PrivacySalt,
-                                  val networkParametersHash: SecureHash?)
+                                  val networkParametersHash: SecureHash?,
+                                  val references: List<StateRef>)
 
 private interface TransactionStateMixin {
     @get:JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
@@ -190,7 +190,7 @@ private class WireTransactionSerializer : JsonSerializer<WireTransaction>() {
                 value.attachments,
                 value.references,
                 value.privacySalt,
-                value.networkParametersHash,
+                value.networkParametersHash
         ))
     }
 }
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
@@ -187,7 +187,8 @@ private class WireTransactionSerializer : JsonSerializer<WireTransaction>() {
                 value.commands,
                 value.timeWindow,
                 value.attachments,
-                value.privacySalt
+                value.privacySalt,
+                value.networkParametersHash
         ))
     }
 }
@@ -201,7 +202,8 @@ private class WireTransactionDeserializer : JsonDeserializer<WireTransaction>() 
                 wrapper.commands,
                 wrapper.attachments,
                 wrapper.notary,
-                wrapper.timeWindow
+                wrapper.timeWindow,
+                networkParametersHash = wrapper.networkParametersHash
         )
         return WireTransaction(componentGroups, wrapper.privacySalt)
     }
@@ -214,7 +216,8 @@ private class WireTransactionJson(val id: SecureHash,
                                   val commands: List<Command<*>>,
                                   val timeWindow: TimeWindow?,
                                   val attachments: List<SecureHash>,
-                                  val privacySalt: PrivacySalt)
+                                  val privacySalt: PrivacySalt,
+                                  val networkParametersHash: SecureHash?)
 
 private interface TransactionStateMixin {
     @get:JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
--- a/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
@@ -212,6 +212,11 @@ abstract class SignTransactionFlow @JvmOverloads constructor(val otherSideSessio
         progressTracker.currentStep = RECEIVING
         // Receive transaction and resolve dependencies, check sufficient signatures is disabled as we don't have all signatures.
         val stx = subFlow(ReceiveTransactionFlow(otherSideSession, checkSufficientSignatures = false))
+        // TODO ENT-2666: Have time period for checking the parameters (because of the delay of propagation of new network data).
+        check(stx.networkParametersHash == serviceHub.networkParametersStorage.currentParametersHash) {
+            "Received transaction for signing with invalid network parameters hash: ${stx.networkParametersHash}." +
+                    "Node's parameters hash: ${serviceHub.networkParametersStorage.currentParametersHash}"
+        }
         // Receive the signing key that the party requesting the signature expects us to sign with. Having this provided
         // means we only have to check we own that one key, rather than matching all keys in the transaction against all
         // keys we own.
--- a/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
@@ -213,9 +213,9 @@ abstract class SignTransactionFlow @JvmOverloads constructor(val otherSideSessio
         // Receive transaction and resolve dependencies, check sufficient signatures is disabled as we don't have all signatures.
         val stx = subFlow(ReceiveTransactionFlow(otherSideSession, checkSufficientSignatures = false))
         // TODO ENT-2666: Have time period for checking the parameters (because of the delay of propagation of new network data).
-        check(stx.networkParametersHash == serviceHub.networkParametersStorage.currentParametersHash) {
+        check(stx.networkParametersHash == serviceHub.networkParametersStorage.currentHash) {
             "Received transaction for signing with invalid network parameters hash: ${stx.networkParametersHash}." +
-                    "Node's parameters hash: ${serviceHub.networkParametersStorage.currentParametersHash}"
+                    "Node's parameters hash: ${serviceHub.networkParametersStorage.currentHash}"
         }
         // Receive the signing key that the party requesting the signature expects us to sign with. Having this provided
         // means we only have to check we own that one key, rather than matching all keys in the transaction against all
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -100,9 +100,11 @@ class NotaryFlow {
             val ctx = stx.coreTransaction
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
-                // TODO networkParametersHash type overlaps with the attachments - it is problematic in filtering.
+                // TODO networkParametersHash type overlaps with the attachments type - it is problematic in filtering.
                 //  Also, we want to be sure in this case that we always include the componentGroup for the parameters hash.
-                is WireTransaction -> ctx.buildFilteredTransaction(Predicate { it is StateRef || it is TimeWindow || it == notaryParty || it is Pair<*, *> && it.first is SecureHash && it.second == ComponentGroupEnum.PARAMETERS_GROUP.ordinal })
+                is WireTransaction -> ctx.buildFilteredTransaction(Predicate {
+                    it is StateRef || it is TimeWindow || it == notaryParty || it is Pair<*, *> && it.first is SecureHash && it.second == ComponentGroupEnum.PARAMETERS_GROUP.ordinal
+                })
                 else -> ctx
             }
             return session.sendAndReceiveWithRetry(NotarisationPayload(tx, signature))
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -13,10 +13,7 @@ import net.corda.core.internal.TimedFlow
 import net.corda.core.internal.notary.generateSignature
 import net.corda.core.internal.notary.validateSignatures
 import net.corda.core.internal.pushToLoggingContext
-import net.corda.core.transactions.ContractUpgradeWireTransaction
-import net.corda.core.transactions.ReferenceStateRef
-import net.corda.core.transactions.SignedTransaction
-import net.corda.core.transactions.WireTransaction
+import net.corda.core.transactions.*
 import net.corda.core.utilities.ProgressTracker
 import net.corda.core.utilities.UntrustworthyData
 import net.corda.core.utilities.unwrap
@@ -101,12 +98,8 @@ class NotaryFlow {
             val ctx = stx.coreTransaction
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
-                // TODO networkParametersHash type overlaps with the attachments type - it is problematic in filtering.
-                //  We want to be sure that attachment with the same hash as network parameters isn't revealed by accident (or vice versa).
-                //  The same applies to reference and input states.
-                //  Also, we want to ensure in this case that the componentGroup for the parameters hash is always included.
                 is WireTransaction -> ctx.buildFilteredTransaction(Predicate {
-                    it is StateRef || it is ReferenceStateRef || it is TimeWindow || it == notaryParty || it is Pair<*, *> && it.first is SecureHash && it.second == ComponentGroupEnum.PARAMETERS_GROUP.ordinal
+                    it is StateRef || it is ReferenceStateRef || it is TimeWindow || it == notaryParty || it is NetworkParametersHash
                 })
                 else -> ctx
             }
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -101,7 +101,9 @@ class NotaryFlow {
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
                 // TODO networkParametersHash type overlaps with the attachments type - it is problematic in filtering.
-                //  Also, we want to be sure in this case that we always include the componentGroup for the parameters hash.
+                //  We want to be sure that attachment with the same hash as network parameters isn't revealed by accident (or vice versa).
+                //  The same applies to reference and input states.
+                //  Also, we want to ensure in this case that the componentGroup for the parameters hash is always included.
                 is WireTransaction -> ctx.buildFilteredTransaction(Predicate {
                     it is StateRef || it is TimeWindow || it == notaryParty || it is Pair<*, *> && it.first is SecureHash && it.second == ComponentGroupEnum.PARAMETERS_GROUP.ordinal
                 })
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -2,6 +2,7 @@ package net.corda.core.flows
 
 import co.paralleluniverse.fibers.Suspendable
 import net.corda.core.DoNotImplement
+import net.corda.core.contracts.ComponentGroupEnum
 import net.corda.core.contracts.StateRef
 import net.corda.core.contracts.TimeWindow
 import net.corda.core.crypto.SecureHash
@@ -29,7 +30,8 @@ class NotaryFlow {
      * signatures will be returned – one from each replica that accepted the input state commit.
      *
      * @throws NotaryException in case the any of the inputs to the transaction have been consumed
-     *                         by another transaction or the time-window is invalid.
+     *                         by another transaction or the time-window is invalid or
+     *                         the parameters used for this transaction are no longer in force in the network.
      */
     @DoNotImplement
     @InitiatingFlow
@@ -98,7 +100,9 @@ class NotaryFlow {
             val ctx = stx.coreTransaction
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
-                is WireTransaction -> ctx.buildFilteredTransaction(Predicate { it is StateRef || it is TimeWindow || it == notaryParty })
+                // TODO networkParametersHash type overlaps with the attachments - it is problematic in filtering.
+                //  Also, we want to be sure in this case that we always include the componentGroup for the parameters hash.
+                is WireTransaction -> ctx.buildFilteredTransaction(Predicate { it is StateRef || it is TimeWindow || it == notaryParty || it is Pair<*, *> && it.first is SecureHash && it.second == ComponentGroupEnum.PARAMETERS_GROUP.ordinal })
                 else -> ctx
             }
             return session.sendAndReceiveWithRetry(NotarisationPayload(tx, signature))
--- a/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
@@ -39,6 +39,7 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
 
             verifyTransaction(requestPayload)
 
+            // TODO Should it commit all parts?
             service.commitInputStates(
                     tx.inputs,
                     tx.id,
@@ -46,7 +47,6 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
                     requestPayload.requestSignature,
                     tx.timeWindow,
                     tx.references)
-
         } catch (e: NotaryInternalException) {
             logError(e.error)
             // Any exception that's not a NotaryInternalException is assumed to be an unexpected internal error
@@ -63,6 +63,7 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
             val transaction = extractParts(requestPayload)
             transactionId = transaction.id
             checkNotary(transaction.notary)
+            checkParametersHash(transaction.networkParametersHash)
             checkInputs(transaction.inputs + transaction.references)
             return transaction
         } catch (e: Exception) {
@@ -87,6 +88,21 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
         require(inputs.size < maxAllowedInputsAndReferences) {
             "A transaction cannot have more than $maxAllowedInputsAndReferences " +
                     "inputs or references, received: ${inputs.size}"
+        }
+    }
+
+    /**
+     * Check that network parameters hash on this transaction is the current hash for the network.
+     */
+     // TODO  ENT-2666 Implement network parameters fuzzy checking. By design in Corda network we have propagation time delay.
+     //     We will never end up in perfect synchronization with all the nodes. However, network parameters update process
+     //     lets us predict what is the reasonable time window for changing parameters on most of the nodes.
+    @Suspendable
+    protected fun checkParametersHash(networkParametersHash: SecureHash?) {
+        if (networkParametersHash == null && serviceHub.networkParameters.minimumPlatformVersion < 4) return
+        val notaryParametersHash = serviceHub.networkParametersStorage.currentParametersHash
+        require (notaryParametersHash == networkParametersHash) {
+            "Transaction for notarisation was tagged with parameters with hash: $networkParametersHash, but current network parameters are: $notaryParametersHash"
         }
     }
 
@@ -118,7 +134,8 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
             val inputs: List<StateRef>,
             val timeWindow: TimeWindow?,
             val notary: Party?,
-            val references: List<StateRef> = emptyList()
+            val references: List<StateRef> = emptyList(),
+            val networkParametersHash: SecureHash?
     )
 
     private fun logError(error: NotaryError) {
--- a/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
@@ -39,7 +39,6 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
 
             verifyTransaction(requestPayload)
 
-            // TODO Should it commit all parts?
             service.commitInputStates(
                     tx.inputs,
                     tx.id,
@@ -99,7 +98,7 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
      //     lets us predict what is the reasonable time window for changing parameters on most of the nodes.
     @Suspendable
     protected fun checkParametersHash(networkParametersHash: SecureHash?) {
-        if (networkParametersHash == null && serviceHub.networkParameters.minimumPlatformVersion < 4) return
+        if (networkParametersHash == null && serviceHub.networkParametersStorage.currentParameters.minimumPlatformVersion < 4) return
         val notaryParametersHash = serviceHub.networkParametersStorage.currentParametersHash
         require (notaryParametersHash == networkParametersHash) {
             "Transaction for notarisation was tagged with parameters with hash: $networkParametersHash, but current network parameters are: $notaryParametersHash"
--- a/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
@@ -98,7 +98,7 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
      //     lets us predict what is the reasonable time window for changing parameters on most of the nodes.
     @Suspendable
     protected fun checkParametersHash(networkParametersHash: SecureHash?) {
-        if (networkParametersHash == null && serviceHub.networkParametersStorage.currentParameters.minimumPlatformVersion < 4) return
+        if (networkParametersHash == null && serviceHub.networkParameters.minimumPlatformVersion < 4) return
         val notaryParametersHash = serviceHub.networkParametersStorage.currentParametersHash
         require (notaryParametersHash == networkParametersHash) {
             "Transaction for notarisation was tagged with parameters with hash: $networkParametersHash, but current network parameters are: $notaryParametersHash"
--- a/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
@@ -99,7 +99,7 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
     @Suspendable
     protected fun checkParametersHash(networkParametersHash: SecureHash?) {
         if (networkParametersHash == null && serviceHub.networkParameters.minimumPlatformVersion < 4) return
-        val notaryParametersHash = serviceHub.networkParametersStorage.currentParametersHash
+        val notaryParametersHash = serviceHub.networkParametersStorage.currentHash
         require (notaryParametersHash == networkParametersHash) {
             "Transaction for notarisation was tagged with parameters with hash: $networkParametersHash, but current network parameters are: $notaryParametersHash"
         }
--- a/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
@@ -5,10 +5,7 @@ import net.corda.core.DoNotImplement
 import net.corda.core.contracts.*
 import net.corda.core.cordapp.CordappContext
 import net.corda.core.cordapp.CordappProvider
-import net.corda.core.crypto.Crypto
-import net.corda.core.crypto.SignableData
-import net.corda.core.crypto.SignatureMetadata
-import net.corda.core.crypto.TransactionSignature
+import net.corda.core.crypto.*
 import net.corda.core.flows.ContractUpgradeFlow
 import net.corda.core.node.services.*
 import net.corda.core.serialization.SerializeAsToken
@@ -41,7 +38,11 @@ interface ServicesForResolution {
     /** Provides access to anything relating to cordapps including contract attachment resolution and app context */
     val cordappProvider: CordappProvider
 
+    /** Provides access to storage of historical network parameters that are used in transaction resolution */
+    val networkParametersStorage: NetworkParametersStorage
+
     /** Returns the network parameters the node is operating under. */
+    // TODO We don't need this one anymore on ServiceHub, could use the networkParametersStorage instead.
     val networkParameters: NetworkParameters
 
     /**
--- a/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
@@ -42,7 +42,7 @@ interface ServicesForResolution {
     val networkParametersStorage: NetworkParametersStorage
 
     /** Returns the network parameters the node is operating under. */
-    // TODO We don't need this one anymore on ServiceHub, could use the networkParametersStorage instead.
+    @Deprecated("Use: ServiceHub.networkParametersStorage.currentParameters instead")
     val networkParameters: NetworkParameters
 
     /**
--- a/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
+++ b/core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
@@ -42,7 +42,6 @@ interface ServicesForResolution {
     val networkParametersStorage: NetworkParametersStorage
 
     /** Returns the network parameters the node is operating under. */
-    @Deprecated("Use: ServiceHub.networkParametersStorage.currentParameters instead")
     val networkParameters: NetworkParameters
 
     /**
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
@@ -1,0 +1,46 @@
+package net.corda.core.node.services
+
+import net.corda.core.DoNotImplement
+import net.corda.core.crypto.SecureHash
+import net.corda.core.internal.SignedDataWithCert
+import net.corda.core.node.NetworkParameters
+
+/**
+ * Interface for handling network parameters storage used for resolving transactions according to parameters that were
+ * historically in force in the network.
+ */
+@DoNotImplement //TODO check DeleteFromDJVM
+interface NetworkParametersStorage {
+    /**
+     * Hash of the current parameters for the network.
+     */
+    val currentParametersHash: SecureHash
+
+    /**
+    * Hash of the current parameters for the network.
+    */
+    val currentParameters: NetworkParameters
+
+    /**
+     * For backwards compatibility, these parameters will be used for resolving historical transactions in the chain.
+     */
+    val defaultParametersHash: SecureHash
+
+    /**
+     * For backwards compatibility, these parameters will be used for resolving historical transactions in the chain.
+     */
+    val defaultParameters: NetworkParameters
+
+    /**
+     * Return network parameters for the given hash. Null if there are no parameters for this hash in the storage and we are unable to
+     * get them from network map.
+     */
+    fun readParametersFromHash(hash: SecureHash): NetworkParameters?
+
+    /**
+     * Save signed network parameters data. Internally network parameters bytes should be stored with the signature.
+     * It's because of ability of older nodes to function in network where parameters were extended with new fields.
+     * Hash should always be calculated over the serialized bytes.
+     */
+    fun saveParameters(signedNetworkParameters: SignedDataWithCert<NetworkParameters>)
+}
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
@@ -9,7 +9,7 @@ import net.corda.core.node.NetworkParameters
  * Interface for handling network parameters storage used for resolving transactions according to parameters that were
  * historically in force in the network.
  */
-@DoNotImplement //TODO check DeleteFromDJVM
+@DoNotImplement
 interface NetworkParametersStorage {
     /**
      * Hash of the current parameters for the network.
@@ -17,12 +17,12 @@ interface NetworkParametersStorage {
     val currentParametersHash: SecureHash
 
     /**
-    * Hash of the current parameters for the network.
+    * Current parameters for the network.
     */
     val currentParameters: NetworkParameters
 
     /**
-     * For backwards compatibility, these parameters will be used for resolving historical transactions in the chain.
+     * For backwards compatibility, this parameters hash will be used for resolving historical transactions in the chain.
      */
     val defaultParametersHash: SecureHash
 
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
@@ -38,6 +38,12 @@ interface NetworkParametersStorage {
     fun readParametersFromHash(hash: SecureHash): NetworkParameters?
 
     /**
+     * Return parameters epoch for the given parameters hash. Null if there are no parameters for this hash in the storage and we are unable to
+     * get them from network map.
+     */
+    fun getEpochFromHash(hash: SecureHash): Int?
+
+    /**
      * Save signed network parameters data. Internally network parameters bytes should be stored with the signature.
      * It's because of ability of older nodes to function in network where parameters were extended with new fields.
      * Hash should always be calculated over the serialized bytes.
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
@@ -15,16 +15,16 @@ interface NetworkParametersStorage {
     /**
      * Hash of the current parameters for the network.
      */
-    val currentParametersHash: SecureHash
+    val currentHash: SecureHash
 
     /**
      * For backwards compatibility, this parameters hash will be used for resolving historical transactions in the chain.
      */
-    val defaultParametersHash: SecureHash
+    val defaultHash: SecureHash
 
     /**
      * Return network parameters for the given hash. Null if there are no parameters for this hash in the storage and we are unable to
      * get them from network map.
      */
-    fun readParametersFromHash(hash: SecureHash): NetworkParameters?
+    fun lookup(hash: SecureHash): NetworkParameters?
 }
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
@@ -1,5 +1,6 @@
 package net.corda.core.node.services
 
+import net.corda.core.CordaInternal
 import net.corda.core.DoNotImplement
 import net.corda.core.crypto.SecureHash
 import net.corda.core.internal.SignedDataWithCert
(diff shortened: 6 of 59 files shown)


node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -98,3 +115,128 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
+
+    private val hashToParameters = createParametersMap(cacheFactory)
+
+    override fun readParametersFromHash(hash: SecureHash): NetworkParameters? {
+        return database.transaction {
+            hashToParameters[hash]?.raw?.deserialize() ?: tryDownloadUnknownParameters(hash)
+        }
+    }
+
+    fun readSignedParametersFromHash(hash: SecureHash): SignedDataWithCert<NetworkParameters>? {
```
adagys: is this used anywhere? does it also need a fallback downlod

node/src/test/kotlin/net/corda/node/services/transactions/ValidatingNotaryServiceTests.kt
```diff
@@ -96,6 +101,41 @@ class ValidatingNotaryServiceTests {
         assertEquals(setOf(expectedMissingKey), missingKeys)
     }
 
+    @Test
+    fun `should sign transaction with valid network parameters`() {
```
adagys: Duplicates `should sign a unique transaction with a valid time-window`

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -117,16 +117,17 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
...
-        return toLedgerTransactionInternal(resolveIdentity, resolveAttachment, resolveStateRef, null)
+        return toLedgerTransactionInternal(resolveIdentity, resolveAttachment, resolveStateRef, resolveParameters)
     }
 
     private fun toLedgerTransactionInternal(
             resolveIdentity: (PublicKey) -> Party?,
             resolveAttachment: (SecureHash) -> Attachment?,
             resolveStateRef: (StateRef) -> TransactionState<*>?,
-            networkParameters: NetworkParameters?
+            resolveParameters: (SecureHash?) -> NetworkParameters?
```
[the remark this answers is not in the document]
kasiastreich: When they cannot be downloaded. This is tricky, you can have hash present.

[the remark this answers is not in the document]
adagys: But that would mean we can't verify the transaction. Should it just throw an exception then?

core/src/main/kotlin/net/corda/core/transactions/NotaryChangeTransactions.kt
```diff
@@ -35,6 +36,15 @@ data class NotaryChangeWireTransaction(
     override val inputs: List<StateRef> = serializedComponents[INPUTS.ordinal].deserialize()
     override val references: List<StateRef> = emptyList()
     override val notary: Party = serializedComponents[NOTARY.ordinal].deserialize()
+    override val networkParametersHash: SecureHash?
+        get() {
+            return try {
+                serializedComponents[PARAMETERS_HASH.ordinal].deserialize()
+            } catch (e: IndexOutOfBoundsException) {
```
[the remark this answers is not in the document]
kasiastreich: It's because of backwards compatibility, old transactions won't have this component group.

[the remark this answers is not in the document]
adagys: oh I'm just saying we can check `serializedComponents.size` before attempting to access the element

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -97,7 +100,9 @@ class NotaryFlow {
             val ctx = stx.coreTransaction
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
-                is WireTransaction -> ctx.buildFilteredTransaction(Predicate { it is StateRef || it is TimeWindow || it == notaryParty })
+                // TODO networkParametersHash type overlaps with the attachments - it is problematic in filtering.
+                //  Also, we want to be sure in this case that we always include the componentGroup for the parameters hash.
+                is WireTransaction -> ctx.buildFilteredTransaction(Predicate { it is StateRef || it is TimeWindow || it == notaryParty || it is Pair<*, *> && it.first is SecureHash && it.second == ComponentGroupEnum.PARAMETERS_GROUP.ordinal })
```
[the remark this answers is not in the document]
kasiastreich: This is question for @kchalkias we discussed that some time ago. The same issue is with reference states, we need new API, wrapping component with ordinal is one way of doing that.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -45,7 +45,8 @@ data class LedgerTransaction @JvmOverloads constructor(
         override val notary: Party?,
         val timeWindow: TimeWindow?,
         val privacySalt: PrivacySalt,
-        private val networkParameters: NetworkParameters? = null,
+        /** Network parameters that were in force when the trasnaction was notarised. */
+        override val networkParameters: NetworkParameters? = null,
```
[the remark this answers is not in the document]
kasiastreich: I can move throwing an exception a level up (before the resolution to Ledger transaction, now it's on verify). The question is, won't it break compatibility when deserializing?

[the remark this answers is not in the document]
adagys: We don't send LedgerTx across the wire or store it, so that should be fine. The only place it could get serialized is in a checkpoint, but I'm not sure if we support maintaining checkpoints cross versions

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
kasiastreich: Hm, true, it's not used now, can be added later.

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -296,7 +296,7 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
         for ((data) in outputs) buf.appendln("${Emoji.leftArrow}OUTPUT:     $data")
         for (command in commands) buf.appendln("${Emoji.diamond}COMMAND:    $command")
         for (attachment in attachments) buf.appendln("${Emoji.paperclip}ATTACHMENT: $attachment")
-        // TODO The most important question is what emoji should I choose for network parameters hash
+        if (networkParametersHash != null) buf.appendln("PARAMETERS HASH: $networkParametersHash")
```
adagys: no emoji!!??

kasiastreich: I can't test it on my system! :sob:

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -98,3 +115,130 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
         operator fun component2() = signed
     }
 }
+
+// TODO NetworkParametersReader + NodeParametersStorage refactor
```
josecoll: Please split this out into a separate PR as discussed (eg. storage including refactoring so we don't have duplication / redundancy in the database).

core/src/main/kotlin/net/corda/core/node/ServiceHub.kt
```diff
@@ -41,7 +38,11 @@ interface ServicesForResolution {
     /** Provides access to anything relating to cordapps including contract attachment resolution and app context */
     val cordappProvider: CordappProvider
 
+    /** Provides access to storage of historical network parameters that are used in transaction resolution */
+    val networkParametersStorage: NetworkParametersStorage
+
     /** Returns the network parameters the node is operating under. */
+    @Deprecated("Use: ServiceHub.networkParametersStorage.currentParameters instead")
```
josecoll: I recommend *not* deprecating this (just default it to `networkParametersStorage.currentParameters` in the SH implementations). This ensures a lot of the original code does not need to change (and is less verbose).

adagys: After seeing the required changes I have to agree with you

kasiastreich: eh... ok, I will revert

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -16,19 +18,26 @@ abstract class CoreTransaction : BaseTransaction() {
...
+     * was created on older version of Corda (before 4), resolution will default to initial parameters.
+     */
+    abstract val networkParametersHash: SecureHash?
 }
 
 /** A transaction with fully resolved components, such as input states. */
 abstract class FullTransaction : BaseTransaction() {
     abstract override val inputs: List<StateAndRef<ContractState>>
     abstract override val references: List<StateAndRef<ContractState>>
+    abstract val networkParameters: NetworkParameters?
```
josecoll: should never be nullable right?

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -16,19 +18,26 @@ abstract class CoreTransaction : BaseTransaction() {
...
 
     override fun checkBaseInvariants() {
         super.checkBaseInvariants()
         checkInputsAndReferencesHaveSameNotary()
     }
 
     private fun checkInputsAndReferencesHaveSameNotary() {
+        // TODO Add check for the notary in whitelist.
```
josecoll: if not part of this PR please reference the JIRA story it should be completed in.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -40,6 +40,11 @@ data class ContractUpgradeWireTransaction(
     val legacyContractAttachmentId: SecureHash by lazy { serializedComponents[LEGACY_ATTACHMENT.ordinal].deserialize<SecureHash>() }
     val upgradedContractClassName: ContractClassName by lazy { serializedComponents[UPGRADED_CONTRACT.ordinal].deserialize<ContractClassName>() }
     val upgradedContractAttachmentId: SecureHash by lazy { serializedComponents[UPGRADED_ATTACHMENT.ordinal].deserialize<SecureHash>() }
+    override val networkParametersHash: SecureHash? by lazy {
```
josecoll: nullable ?

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -77,6 +82,13 @@ data class ContractUpgradeWireTransaction(
                 ?: throw AttachmentResolutionException(legacyContractAttachmentId)
         val upgradedContractAttachment = services.attachments.openAttachment(upgradedContractAttachmentId)
                 ?: throw AttachmentResolutionException(upgradedContractAttachmentId)
+        val resolvedNetworkParameters = networkParametersHash.let {
+            if (it == null) {
+                services.networkParametersStorage.defaultParameters
```
josecoll: unnecessary nullability checking

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -134,6 +147,12 @@ data class ContractUpgradeFilteredTransaction(
         visibleComponents[NOTARY.ordinal]?.component?.deserialize<Party>()
                 ?: throw IllegalArgumentException("Notary not specified")
     }
+    // TODO Do we ever store filteredTransactions anywhere? Or is it safe to assume that apart from checkpoints it won't break wire compat.
```
josecoll: I don't think we do ... but would ask @exFalso to confirm on this point.

core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
```diff
@@ -48,6 +48,8 @@ abstract class TraversableTransaction(open val componentGroups: List<ComponentGr
         timeWindows.firstOrNull()
     }
 
+    override val networkParametersHash: SecureHash? = deserialiseComponentGroup(SecureHash::class, PARAMETERS_GROUP).firstOrNull()
```
josecoll: singleOrNull ?

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -135,7 +135,9 @@ open class TransactionBuilder @JvmOverloads constructor(
                             (allContractAttachments + attachments).toSortedSet().toList(), // Sort the attachments to ensure transaction builds are stable.
                             notary,
                             window,
-                            referenceStates),
+                            referenceStates
+                    ,
```
josecoll: move comma to previous line

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -279,7 +281,7 @@ open class TransactionBuilder @JvmOverloads constructor(
         val defaultOutputConstraint = selectAttachmentConstraint(contractClassName, inputStates, attachmentToUse, services)
 
         // Sanity check that the selected attachment actually passes.
-        val constraintAttachment = AttachmentWithContext(attachmentToUse, contractClassName, services.networkParameters.whitelistedContractImplementations)
+        val constraintAttachment = AttachmentWithContext(attachmentToUse, contractClassName, services.networkParametersStorage.currentParameters.whitelistedContractImplementations)
```
josecoll: `services.networkParameters`

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -310,9 +312,9 @@ open class TransactionBuilder @JvmOverloads constructor(
             attachmentToUse: ContractAttachment,
             services: ServicesForResolution): AttachmentConstraint = when {
         inputStates != null -> attachmentConstraintsTransition(inputStates.groupBy { it.constraint }.keys, attachmentToUse)
-        attachmentToUse.signers.isNotEmpty() && services.networkParameters.minimumPlatformVersion < 4 -> {
-            log.warnOnce("Signature constraints not available on network requiring a minimum platform version of 4. Current is: ${services.networkParameters.minimumPlatformVersion}.")
-            if (useWhitelistedByZoneAttachmentConstraint(contractClassName, services.networkParameters)) {
+        attachmentToUse.signers.isNotEmpty() && services.networkParametersStorage.currentParameters.minimumPlatformVersion < 4 -> {
```
josecoll: `services.networkParameters`

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -310,9 +312,9 @@ open class TransactionBuilder @JvmOverloads constructor(
...
+            log.warnOnce("Signature constraints not available on network requiring a minimum platform version of 4. Current is: ${services.networkParametersStorage.currentParameters.minimumPlatformVersion}.")
+            if (useWhitelistedByZoneAttachmentConstraint(contractClassName, services.networkParametersStorage.currentParameters)) {
```
josecoll: `services.networkParameters`

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -321,7 +323,7 @@ open class TransactionBuilder @JvmOverloads constructor(
             }
         }
         attachmentToUse.signers.isNotEmpty() -> makeSignatureAttachmentConstraint(attachmentToUse.signers)
-        useWhitelistedByZoneAttachmentConstraint(contractClassName, services.networkParameters) -> WhitelistedByZoneAttachmentConstraint
+        useWhitelistedByZoneAttachmentConstraint(contractClassName, services.networkParametersStorage.currentParameters) -> WhitelistedByZoneAttachmentConstraint
```
josecoll: `services.networkParameters`

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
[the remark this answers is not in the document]
josecoll: Agree - this should throw an exception.
Also, can the (`SecureHash`) input ever be null?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -169,16 +169,18 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
...
     val keyManagementService = makeKeyManagementService(identityService).tokenize()
-    val servicesForResolution = ServicesForResolutionImpl(identityService, attachments, cordappProvider, transactionStorage).also {
+    val servicesForResolution = ServicesForResolutionImpl(identityService, attachments, cordappProvider, networkParametersStorage, transactionStorage).also {
         attachments.servicesForResolution = it
     }
     @Suppress("LeakingThis")
     val vaultService = makeVaultService(keyManagementService, servicesForResolution, database).tokenize()
     val nodeProperties = NodePropertiesPersistentStore(StubbedNodeUniqueIdProvider::value, database, cacheFactory)
     val flowLogicRefFactory = FlowLogicRefFactoryImpl(cordappLoader.appClassLoader)
+    // TODO Cancelling parameters updates - if we do that, how we ensure that no one uses cancelled parameters in the transactions?
```
josecoll: What do you mean? A node cannot stop listening to network updates right?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -967,7 +969,7 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
         private lateinit var _myInfo: NodeInfo
         override val myInfo: NodeInfo get() = _myInfo
 
-        private lateinit var _networkParameters: NetworkParameters
+        private lateinit var _networkParameters: NetworkParameters // TODO delegate to network parameters storage
```
josecoll: TODO please ...

node/src/main/kotlin/net/corda/node/internal/CordaRPCOpsImpl.kt
```diff
@@ -83,7 +83,7 @@ internal class CordaRPCOpsImpl(
         return snapshot
     }
 
-    override val networkParameters: NetworkParameters get() = services.networkParameters
+    override val networkParameters: NetworkParameters get() = services.networkParametersStorage.currentParameters
```
josecoll: revert change

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -98,3 +115,130 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
+class NodeParametersStorage(
```
josecoll: Move out to separate file so it aligns more closely with the interface.
Should it not be called `NetworkParametersService` ?

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -98,3 +115,130 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
+        saveParameters(currentSignedParameters)
+        _currentHash = currentSignedParameters.raw.hash
+    }
+
+    private lateinit var _currentHash: SecureHash
+    override val currentParametersHash: SecureHash get() = _currentHash
+    override val currentParameters: NetworkParameters
+        get() = readParametersFromHash(currentParametersHash)
+                ?: throw IllegalAccessException("No current parameters for the network provided")
+    // TODO Have network map serve special "starting" parameters as parameters for resolution for older transactions?
```
josecoll: Yes - need to decide what are the UNTAGGED txn NP's.

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -98,3 +115,130 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
+
+    override fun saveParameters(signedNetworkParameters: SignedNetworkParameters) {
+        log.trace { "Saving new network parameters to network parameters storage." }
+        val networkParameters = signedNetworkParameters.verified()
+        val hash = signedNetworkParameters.raw.hash
+        log.trace { "Parameters to save $networkParameters with hash $hash" }
+        hashToParameters.addWithDuplicatesAllowed(hash, signedNetworkParameters)
+    }
+
+    // TODO For the future we could get them also as signed (by network operator) attachments on transactions.
```
josecoll: Please add or reference a JIRA story for this.

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -98,3 +115,130 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
+    }
+
+    @Entity
+    @javax.persistence.Table(name = "${NODE_DATABASE_PREFIX}network_parameters")
+    class PersistentNetworkParameters(
+            @Id
+            @Column(name = "hash", length = MAX_HASH_HEX_SIZE, nullable = false)
+            val hash: String = "",
+
+            // TODO For notary change transaction needed for Andrius work
```
josecoll: remove reference to "Andrius work" (perhaps a JIRA reference instead)

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
```diff
@@ -98,3 +115,130 @@ class NetworkParametersReader(private val trustRoot: X509Certificate,
...
+            @Column(name = "cert", nullable = false)
+            private val certificate: ByteArray = ArrayUtils.EMPTY_BYTE_ARRAY,
+
+            // Parent certificate path (the first one is stored separately), so node is agnostic to certificate hierarchy.
+            @Lob
+            @Column(name = "parent_cert_path", nullable = false)
+            private val certPath: ByteArray = ArrayUtils.EMPTY_BYTE_ARRAY
+    ) {
+        val networkParameters: NetworkParameters get() = networkParametersBytes.deserialize()
+        val signedNetworkParameters: SignedDataWithCert<NetworkParameters>
```
josecoll: @kchalkias please can you review this piece of code (eg. storing certificate, cert path & signatures with NPs)

node/src/test/kotlin/net/corda/node/services/persistence/ExposeJpaToFlowsTests.kt
```diff
@@ -4,29 +4,25 @@ import co.paralleluniverse.fibers.Suspendable
...
-import java.time.Duration
 import javax.persistence.Column
 import javax.persistence.Entity
 import javax.persistence.Id
 import javax.persistence.Table
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 
-
 class ExposeJpaToFlowsTests {
```
josecoll: why has this code changed?

node/src/test/kotlin/net/corda/node/services/persistence/NodeAttachmentServiceTest.kt
```diff
@@ -69,8 +68,6 @@ class NodeAttachmentServiceTest {
         database = configureDatabase(dataSourceProperties, DatabaseConfig(), { null }, { null })
         fs = Jimfs.newFileSystem(Configuration.unix())
 
-        doReturn(testNetworkParameters()).whenever(services).networkParameters
```
josecoll: interesting ... that this has not failed existing tests.

node/src/test/kotlin/net/corda/node/services/transactions/MaxTransactionSizeTests.kt
```diff
@@ -40,6 +40,7 @@ class MaxTransactionSizeTests {
         mockNet = MockNetwork(listOf("net.corda.testing.contracts"), networkParameters = testNetworkParameters(maxTransactionSize = 3_000_000))
         aliceNode = mockNet.createNode(MockNodeParameters(legalName = ALICE_NAME))
         bobNode = mockNet.createNode(MockNodeParameters(legalName = BOB_NAME))
+        bobNode.registerInitiatedFlow(ReceiveLargeTransactionFlow::class.java)
```
josecoll: ???

node/src/test/kotlin/net/corda/node/services/transactions/ResolveStatePointersTest.kt
```diff
@@ -58,6 +52,18 @@ class ResolveStatePointersTest {
...
 
+    @Before
+    fun setUp() {
+        val databaseAndServices = MockServices.makeTestDatabaseAndMockServices(
+                cordappPackages = cordapps,
+                identityService = makeTestIdentityService(notary.identity, myself.identity),
+                initialIdentity = myself,
+                networkParameters = testNetworkParameters(minimumPlatformVersion = 4)
+        )
+
```
josecoll: empty line

node/src/test/kotlin/net/corda/node/services/vault/VaultQueryExceptionsTests.kt
```diff
@@ -16,6 +16,10 @@ import org.junit.rules.ExpectedException
 class VaultQueryExceptionsTests : VaultQueryParties by rule {
```
josecoll: why has this code changed?

serialization/src/test/kotlin/net/corda/serialization/internal/AttachmentsClassLoaderTests.kt
```diff
@@ -63,10 +64,15 @@ class AttachmentsClassLoaderTests {
...
+    private val networkParametersStorage get() = rigorousMock<NetworkParametersStorage>().also {
+        doReturn(networkParameters.serialize().hash).whenever(it).currentParametersHash
+        doReturn(networkParameters).whenever(it).currentParameters
+    }
+
+    private val serviceHub get() = rigorousMock<ServiceHub>().also {
         doReturn(attachments).whenever(it).attachments
         doReturn(cordappProvider).whenever(it).cordappProvider
-        doReturn(networkParameters).whenever(it).networkParameters
+        doReturn(networkParametersStorage).whenever(it).networkParametersStorage
```
josecoll: revert

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -252,15 +257,15 @@ open class MockServices private constructor(
         }
     override val transactionVerifierService: TransactionVerifierService get() = InMemoryTransactionVerifierService(2)
     private val mockCordappProvider: MockCordappProvider = MockCordappProvider(cordappLoader, attachments).also {
-        it.start( networkParameters.whitelistedContractImplementations)
+        // TODO Remove deprecated usage
```
josecoll: revert

josecoll:
In your description:
```
This is only part of the work, next PR will include changes to
ResolveTransactionsFlow to make sure that parameters in the transaction
graph are ordered (this is to prevent the downgrade attack, when the
malicious notary and participants sign transaction that shouldn't be
notarised otherwise).
```
What JIRA story does this refer to?

josecoll:
With reference to:

```
Probably on network services side we need the default parameters
endpoint for the transactions that were created before this change - for now it's default
to the current ones.
```

Do we actually need this? 
Can we not just define and use a PRE_NP_TXN_TAGGING default NP set.

node/src/test/kotlin/net/corda/node/services/transactions/MaxTransactionSizeTests.kt
kasiastreich: This test was wrong, it threw an exception on unregistered flow.

kasiastreich:
No, there are no generic untagged network parameters pre-Corda 4. What will you put in notary whitelist? What will you put in `whitelistedContractImplementations`? I can't put that now anywhere.

testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
kasiastreich: I think it's sufficient to mention once that I need to revert the change that Andrius asked me for, at the top.

node/src/test/kotlin/net/corda/node/services/vault/VaultQueryExceptionsTests.kt
kasiastreich: Because of the SerializationEnvironmentRule nonsense and initialisation order.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -97,7 +100,9 @@ class NotaryFlow {
             val ctx = stx.coreTransaction
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
-                is WireTransaction -> ctx.buildFilteredTransaction(Predicate { it is StateRef || it is TimeWindow || it == notaryParty })
+                // TODO networkParametersHash type overlaps with the attachments - it is problematic in filtering.
```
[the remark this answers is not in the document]
kasiastreich: I will extend the comment, I discussed it with @kchalkias already.
It's because `networkParametersHash` type overlaps with the attachments type, you want to be sure that you don't reveal by accident attachment that has same hash as network parameters or vice versa . I think it is worse for reference states, so this API needs changing either way.

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
```diff
@@ -90,6 +90,21 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
         }
     }
 
+    /**
+     * Check that network parameters hash on this transaction is the current hash for the network.
+     */
+     // TODO  ENT-2666 Implement network parameters fuzzy checking. By design in Corda network we have propagation time delay.
```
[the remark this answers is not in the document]
kasiastreich: It's not related to this PR and it needs extending network parameters itself with tooling on network services side and design decision related to network merging. I will let @adagys comment on that.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
kasiastreich: Old transactions

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
kasiastreich: Yes, CORDA-2146. This TODO is not needed here, will remove.

core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+    * Current parameters for the network.
+    */
+    val currentParameters: NetworkParameters
+
+    /**
+     * For backwards compatibility, this parameters hash will be used for resolving historical transactions in the chain.
+     */
+    val defaultParametersHash: SecureHash
```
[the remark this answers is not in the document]
josecoll: Right - interesting, it would obviously be specific to whatever Network Parameters were in operation at the time (for a given pre-Corda 4 Corda Network). But how would we ever know? (hence do we not just default to an EMPTY_NP_SET).

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -16,19 +18,26 @@ abstract class CoreTransaction : BaseTransaction() {
     abstract override val inputs: List<StateRef>
     /** The reference inputs of this transaction, containing the state references only. **/
     abstract override val references: List<StateRef>
+    /**
+     * Hash of the network parameters that were in force when the transaction was notarised. Null means, that the transaction
+     * was created on older version of Corda (before 4), resolution will default to initial parameters.
+     */
+    abstract val networkParametersHash: SecureHash?
```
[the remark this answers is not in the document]
kasiastreich: It can be done also with the UNTAGGED that refers to zero hash or something like that. But... It requires more work with adding special versioned constructor for deserialization, while nullability works out of the box. Every new non-nullable field on the transaction will require versioned constructor. I also like elvis operator for checking nullability ;)

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
kasiastreich: Network map can cancel update. This means that these parameters were never functioning in the network, but nodes still can download the signed network parameters data. It needs more thought.

node/src/main/kotlin/net/corda/node/internal/NetworkParametersReader.kt
kasiastreich: I followed the convention of AttachmentStorage naming, but I guess it can be called that too.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
josecoll: But by cancelling an update, does that not also mean it re-publishes the correct/previous set of NPs? (so we're all good, as we'll receive those and update ourselves accordingly).

node/src/main/kotlin/net/corda/node/internal/NodeParametersStorage.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    override fun readParametersFromHash(hash: SecureHash): NetworkParameters? {
+        return database.transaction {
+            hashToParameters[hash]?.raw?.deserialize() ?: tryDownloadUnknownParameters(hash)
+        }
+    }
+
+    override fun getEpochFromHash(hash: SecureHash): Int? = readParametersFromHash(hash)?.epoch
+
+    override fun saveParameters(signedNetworkParameters: SignedNetworkParameters) {
```
adagys: what happens on a network parameter update, do they get save to the storage automatically?

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
```diff
@@ -212,6 +212,11 @@ abstract class SignTransactionFlow @JvmOverloads constructor(val otherSideSessio
         progressTracker.currentStep = RECEIVING
         // Receive transaction and resolve dependencies, check sufficient signatures is disabled as we don't have all signatures.
         val stx = subFlow(ReceiveTransactionFlow(otherSideSession, checkSufficientSignatures = false))
+        // TODO ENT-2666: Have time period for checking the parameters (because of the delay of propagation of new network data).
+        check(stx.networkParametersHash == serviceHub.networkParametersStorage.currentParametersHash) {
```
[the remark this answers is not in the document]
josecoll: this is to avoid deprecating the existing `networkParameters` (which is used ubiquitously throughout the current codebase).

node/src/test/kotlin/net/corda/node/services/vault/VaultQueryExceptionsTests.kt
kasiastreich: You can't imagine how many what the hell moments I had with this and rigorous mock after a year of not looking at core :D

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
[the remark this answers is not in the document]
kasiastreich: There is one problem with the extension on `networkParameters` you can't calculate hash on the fly (unless you store serialised form, as I do now - you can't simply call `serialize().hash` as it changes between versions). I prefer my current approach. Deprecation was only after addressing @adagys comment, which I reverted.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
kasiastreich: This comment is about something different. It causes potential problem as cancelled parameters were never in force in the network, but were signed and distributed by the operator. If these parameters included a notary that then was decided to be removed, well, no one can stop this notary from signing transactions with these (invalid) parameters.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
kasiastreich: Yes, I forgot to remove it. It can be non-nullable only if `minimumPlatformVersion` increases to 4.

node/src/main/kotlin/net/corda/node/internal/NodeParametersStorage.kt
```diff
@@ -0,0 +1,155 @@
...
+    // TODO For the future we could get them also as signed (by network operator) attachments on transactions.
+    private fun tryDownloadUnknownParameters(parametersHash: SecureHash): NetworkParameters? {
+        return if (networkMapClient != null) {
+            try {
+                val signedParams = networkMapClient.getNetworkParameters(parametersHash)
+                val networkParameters = signedParams.verifiedNetworkMapCert(trustRoot)
+                saveParameters(signedParams)
+                networkParameters
+            } catch (e: Exception) {
+                log.warn("Failed to downolad historical network parameters with hash $parametersHash", e)
```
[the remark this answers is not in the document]
kasiastreich: :police_car:

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
```diff
@@ -199,7 +200,8 @@ private class WireTransactionDeserializer : JsonDeserializer<WireTransaction>()
                 wrapper.commands,
                 wrapper.attachments,
                 wrapper.notary,
-                wrapper.timeWindow
+                wrapper.timeWindow,
+                networkParametersHash = wrapper.networkParametersHash
```
kchalkias: ditto

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
```diff
@@ -185,7 +185,8 @@ private class WireTransactionSerializer : JsonSerializer<WireTransaction>() {
                 value.commands,
                 value.timeWindow,
                 value.attachments,
-                value.privacySalt
+                value.privacySalt,
+                value.networkParametersHash
```
kchalkias: NOt sure if this is related to this PR or we should take the opportunity to fix it, but maybe we should handle referenceStates as well? I.e., by adding it to `private class WireTransactionJson`?

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
```diff
@@ -212,7 +214,8 @@ private class WireTransactionJson(val id: SecureHash,
                                   val commands: List<Command<*>>,
                                   val timeWindow: TimeWindow?,
                                   val attachments: List<SecureHash>,
-                                  val privacySalt: PrivacySalt)
+                                  val privacySalt: PrivacySalt,
+                                  val networkParametersHash: SecureHash?)
```
kchalkias: maybe add `references` as well?

core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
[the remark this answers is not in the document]
kasiastreich: My suggestion here is, that default ones can be advertised on the network level, by the operator, but we need to ensure that they never change. And defaulting it to empty something doesn't make sense, as mentioned above, it will break validation of historical chain for notaries.

node/src/test/kotlin/net/corda/node/services/persistence/ExposeJpaToFlowsTests.kt
kasiastreich: Serialisation initialisation again

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -97,7 +100,11 @@ class NotaryFlow {
             val ctx = stx.coreTransaction
             val tx = when (ctx) {
                 is ContractUpgradeWireTransaction -> ctx.buildFilteredTransaction()
-                is WireTransaction -> ctx.buildFilteredTransaction(Predicate { it is StateRef || it is TimeWindow || it == notaryParty })
+                // TODO networkParametersHash type overlaps with the attachments type - it is problematic in filtering.
+                //  Also, we want to be sure in this case that we always include the componentGroup for the parameters hash.
+                is WireTransaction -> ctx.buildFilteredTransaction(Predicate {
+                    it is StateRef || it is TimeWindow || it == notaryParty || it is Pair<*, *> && it.first is SecureHash && it.second == ComponentGroupEnum.PARAMETERS_GROUP.ordinal
```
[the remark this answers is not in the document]
kasiastreich: One for @kchalkias, how he sees that. It's not strictly part of this PR and API needs to change either way.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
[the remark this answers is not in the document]
kchalkias: The case of having conflicting component types has been discussed in the past. At that time we said we should ensure no identical types should exist, but it seems this assumption is too strict. I'm also concerned about the usability if we use Pair<>, I mean it might seem difficult for clients to build filters. I would personally prefer using another wrapped type to avoid dealing with this now. Wdyt? I've already mentioned this to @roger3cev about reference states (same issue as Kasia mentioned) and we have a related Jira task. wdyt?

core/src/main/kotlin/net/corda/core/transactions/NotaryChangeTransactions.kt
```diff
@@ -35,9 +36,16 @@ data class NotaryChangeWireTransaction(
     override val inputs: List<StateRef> = serializedComponents[INPUTS.ordinal].deserialize()
     override val references: List<StateRef> = emptyList()
     override val notary: Party = serializedComponents[NOTARY.ordinal].deserialize()
+
     /** Identity of the notary service to reassign the states to.*/
     val newNotary: Party = serializedComponents[NEW_NOTARY.ordinal].deserialize()
 
+    override val networkParametersHash: SecureHash? by lazy {
+        if (serializedComponents.size >= 4) {
```
kchalkias: ditto, add a comment about "4" and update the `Component` list (comments/kdocs), similarly to `ComponentGroupEnum` for clarity.

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -40,6 +40,11 @@ data class ContractUpgradeWireTransaction(
     val legacyContractAttachmentId: SecureHash by lazy { serializedComponents[LEGACY_ATTACHMENT.ordinal].deserialize<SecureHash>() }
     val upgradedContractClassName: ContractClassName by lazy { serializedComponents[UPGRADED_CONTRACT.ordinal].deserialize<ContractClassName>() }
     val upgradedContractAttachmentId: SecureHash by lazy { serializedComponents[UPGRADED_ATTACHMENT.ordinal].deserialize<SecureHash>() }
+    override val networkParametersHash: SecureHash? by lazy {
+        if (serializedComponents.size >= 6) {
```
kchalkias: Maybe add a comment here why "6" or better update kdocs and add comments about ordinals to ContractUpgradeWireTransaction.Component, similarly to `ComponentGroupEnum`.

core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
```diff
@@ -148,6 +156,7 @@ class FilteredTransaction internal constructor(
                 if (wtx.notary != null) filter(wtx.notary, NOTARY_GROUP.ordinal, 0)
                 if (wtx.timeWindow != null) filter(wtx.timeWindow, TIMEWINDOW_GROUP.ordinal, 0)
                 wtx.references.forEachIndexed { internalIndex, it -> filter(it, REFERENCES_GROUP.ordinal, internalIndex) }
+                wtx.networkParametersHash?.let { filter(Pair(it, PARAMETERS_GROUP.ordinal), PARAMETERS_GROUP.ordinal, 0) }
```
kchalkias: This should be well-documented, so we know how to filter-in netParams in a `FilteredTx`. At the moment it works for the custom hand-crafted types of transactions (i..e, non-validating Notaries), but let's assume one wants to create a regular FilteredTx, i.e., to an Oracle that for some reasons should see netParams. That said, we should always use the `Pair` logic in our `Predicate`s to achieve that.

exFalso:
Re filtering, we could add an alternative API and deprecate the old one:
```kotlin
diff --git a/core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt b/core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
index 811b3dfab6..3a5202a4fb 100644
--- a/core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/MerkleTransaction.kt
@@ -12,6 +12,7 @@ import net.corda.core.serialization.*
 import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.utilities.lazyMapped
+import java.util.function.BiFunction
 import java.util.function.Predicate
 import kotlin.reflect.KClass
 
@@ -147,8 +148,16 @@ class FilteredTransaction internal constructor(
          * @param wtx WireTransaction to be filtered.
          * @param filtering filtering over the whole WireTransaction.
          */
+
         @JvmStatic
+        @Deprecated(message = "Don't use this lol, SecureHash is ambiguous")
         fun buildFilteredTransaction(wtx: WireTransaction, filtering: Predicate<Any>): FilteredTransaction {
+            val filteredComponentGroups = filterWithFun(wtx, BiFunction { _, a -> filtering.test(a) })
+            return FilteredTransaction(wtx.id, filteredComponentGroups, wtx.groupHashes)
+        }
+
+        @JvmStatic
+        fun buildFilteredTransaction(wtx: WireTransaction, filtering: BiFunction<Int, Any, Boolean>): FilteredTransaction {
             val filteredComponentGroups = filterWithFun(wtx, filtering)
             return FilteredTransaction(wtx.id, filteredComponentGroups, wtx.groupHashes)
         }
@@ -159,14 +168,14 @@ class FilteredTransaction internal constructor(
          * @param filtering filtering over the whole WireTransaction.
          * @return a list of [FilteredComponentGroup] used in PartialMerkleTree calculation and verification.
          */
-        private fun filterWithFun(wtx: WireTransaction, filtering: Predicate<Any>): List<FilteredComponentGroup> {
+        private fun filterWithFun(wtx: WireTransaction, filtering: BiFunction<Int, Any, Boolean>): List<FilteredComponentGroup> {
             val filteredSerialisedComponents: MutableMap<Int, MutableList<OpaqueBytes>> = hashMapOf()
             val filteredComponentNonces: MutableMap<Int, MutableList<SecureHash>> = hashMapOf()
             val filteredComponentHashes: MutableMap<Int, MutableList<SecureHash>> = hashMapOf() // Required for partial Merkle tree generation.
             var signersIncluded = false
 
             fun <T : Any> filter(t: T, componentGroupIndex: Int, internalIndex: Int) {
-                if (!filtering.test(t)) return
+                if (!filtering.apply(componentGroupIndex, t)) return
 
                 val group = filteredSerialisedComponents[componentGroupIndex]
```

Or perhaps create an explicit
```kotlin

@FunctionalInterface
interface FilteringFunction {
    fun filter(componentGroupEnum: ComponentGroupEnum, component: Any): Boolean
}
```

This API should be much clearer

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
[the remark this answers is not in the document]
josecoll: Agree - this needs revisiting as a separate story.

node/src/main/kotlin/net/corda/node/internal/NodeParametersStorage.kt
```diff
@@ -0,0 +1,145 @@
...
+import net.corda.nodeapi.internal.crypto.X509Utilities
+import net.corda.nodeapi.internal.network.SignedNetworkParameters
+import net.corda.nodeapi.internal.network.verifiedNetworkMapCert
+import net.corda.nodeapi.internal.persistence.CordaPersistence
+import net.corda.nodeapi.internal.persistence.NODE_DATABASE_PREFIX
+import org.apache.commons.lang.ArrayUtils
+import java.security.cert.X509Certificate
+import javax.persistence.*
+
+// TODO NetworkParametersReader + NodeParametersStorage refactor
```
josecoll: ? (they are distinct services doing different things)

node/src/main/kotlin/net/corda/node/internal/NodeParametersStorage.kt
```diff
@@ -0,0 +1,145 @@
...
+                val signedParams = networkMapClient.getNetworkParameters(parametersHash)
+                val networkParameters = signedParams.verifiedNetworkMapCert(trustRoot)
+                saveParameters(signedParams)
+                networkParameters
+            } catch (e: Exception) {
+                log.warn("Failed to download historical network parameters with hash $parametersHash", e)
+                null
+            }
+        } else {
+            log.warn("Tried to download historical network parameters with hash $parametersHash, but network map url isn't configured")
```
josecoll: Isn't there always a Network Map client URL configured? Of course, this may not be reachable (which is a different problem).

node/src/main/kotlin/net/corda/node/internal/ServicesForResolutionImpl.kt
```diff
@@ -5,21 +5,19 @@ import net.corda.core.cordapp.CordappProvider
...
+        override val networkParametersStorage: NetworkParametersStorage,
         private val validatedTransactions: TransactionStorage
 ) : ServicesForResolution {
-    private lateinit var _networkParameters: NetworkParameters
-    override val networkParameters: NetworkParameters get() = _networkParameters
-
-    fun start(networkParameters: NetworkParameters) {
-        _networkParameters = networkParameters
-    }
+    override val networkParameters: NetworkParameters get() = networkParametersStorage.readParametersFromHash(networkParametersStorage.currentParametersHash) ?:
```
josecoll: Is there a timing synchronisation issue here?
What happens if the latest NP's have been pushed to the node (but not yet saved into the NP database) ?
Should be not get the latest using the file reader first?

node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
```diff
@@ -71,6 +66,7 @@ import kotlin.test.assertTrue
  *
  * We assume that Alice and Bob already found each other via some market, and have agreed the details already.
  */
+// TODO These tests need serious cleanup.
```
josecoll: I don't quite understand why you needed to change these tests as there is nothing related to your changes ???

node/src/test/kotlin/net/corda/node/services/network/NodeParametersStorageTest.kt
```diff
@@ -0,0 +1,132 @@
...
+import net.corda.testing.node.MockServices
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.io.PrintStream
+import kotlin.streams.toList
+
+class NodeParametersStorageTest {
```
josecoll: rename `DBNetworkParametersStorageTest` to align

node/src/test/kotlin/net/corda/node/services/network/NodeParametersStorageTest.kt
```diff
@@ -0,0 +1,132 @@
...
+            assertThat(nodeParametersStorage.readParametersFromHash(hash2)).isEqualTo(netParams)
+            verify(networkMapClient, times(1)).getNetworkParameters(hash2)
+
+        }
+    }
+
+    @Test
+    fun `try save parameters with incorrect signature`() {
+        database.transaction {
+            val consoleOutput = interceptConsoleOutput {
```
josecoll: interesting ;-) 
can you not directly assert on a failed Exception type?

node/src/test/kotlin/net/corda/node/services/persistence/ExposeJpaToFlowsTests.kt
josecoll: Evil!

node/src/main/kotlin/net/corda/node/internal/NodeParametersStorage.kt
kasiastreich: No, for test deployments using Bootstraper there is no NetworkMapClient URL configured.

node/src/main/kotlin/net/corda/node/internal/ServicesForResolutionImpl.kt
kasiastreich: That is interesting. I wrote this code when there was no auto-approve of network parameters (merged in last two days I think, and I wasn't aware of that). Before that node updated the parameters always on restart, so there were no problems with synchronisation.

node/src/test/kotlin/net/corda/node/services/network/NodeParametersStorageTest.kt
kasiastreich: It shouldn't throw when data from network map is corrupted, for now warning is logged and null returned.

node/src/main/kotlin/net/corda/node/internal/DBNetworkParametersStorage.kt
```diff
@@ -0,0 +1,144 @@
...
+                null
+            }
+        } else {
+            log.warn("Tried to download historical network parameters with hash $parametersHash, but network map url isn't configured")
+            null
+        }
+    }
+
+    @Entity
+    @Table(name = "${NODE_DATABASE_PREFIX}network_parameters")
```
adagys: do you need to add any schema migration stuff?

core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
```diff
@@ -26,17 +27,4 @@ interface NetworkParametersStorage {
      * get them from network map.
      */
     fun readParametersFromHash(hash: SecureHash): NetworkParameters?
```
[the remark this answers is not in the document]
kasiastreich: nah, that's fine :smile:

PR #4229, opened 2018-11-13
[CORDA-2219] Fix Network Bootstrapper error messages

gatsinski:
This pull request attempts to fix [CORDA-2219](https://r3-cev.atlassian.net/browse/CORDA-2219).

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -194,7 +194,9 @@ internal constructor(private val initSerEnv: Boolean,
         // Don't accidently include the bootstrapper jar as a CorDapp!
         val bootstrapperJar = javaClass.location.toPath()
         val cordappJars = directory.list { paths ->
-            paths.filter { it.toString().endsWith(".jar") && !it.isSameAs(bootstrapperJar) && it.fileName.toString() != "corda.jar" }.toList()
+            paths.filter { it.toString().endsWith(".jar") &&
+                           !it.isSameAs(bootstrapperJar) &&
+                           it.fileName.toString() !in listOf("corda.jar", "runnodes.jar") }.toList()
         }
         bootstrap(directory, cordappJars, copyCordapps, fromCordform = false, minimumPlatformVersion = minimumPlatformVersion, packageOwnership = packageOwnership)
     }


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -194,7 +194,9 @@ internal constructor(private val initSerEnv: Boolean,
         // Don't accidently include the bootstrapper jar as a CorDapp!
         val bootstrapperJar = javaClass.location.toPath()
         val cordappJars = directory.list { paths ->
-            paths.filter { it.toString().endsWith(".jar") && !it.isSameAs(bootstrapperJar) && it.fileName.toString() != "corda.jar" }.toList()
+            paths.filter { it.toString().endsWith(".jar") &&
+                           !it.isSameAs(bootstrapperJar) &&
+                           it.fileName.toString() !in listOf("corda.jar", "runnodes.jar") }.toList()
```
vkolomeyko: Do we know which step created `runnodes.jar`?
Is it something new for Corda 4?

gatsinski: No. I don't think it is something new. `runnodes.jar` is created by `deployNodes` task. Bank of Corda Demo is a good example - `./gradlew samples:bank-of-corda-demo:deployNodes`.

vkolomeyko: Possibly, this is related: https://github.com/corda/corda/blob/master/docs/source/running-a-node.rst#native
Where `deployNodes` generates `runnodes.jar`.
Then as far as BootStrapper testing is concerned it might be that `runnodes.jar` is just an arteface that was left behind from previous testing...

gatsinski: What are you suggesting? To ignore the problem?

vkolomeyko: Let me try to perform testing by following test instructions and then I might have a better idea what can be done.
But I am not 100% sure the code change is right. We are including `runnodes.jar` which is created during testing by samples code into *production* class of `node-api`.

vkolomeyko: I have put additional information with my findings on: CORDA-2219

PR #4230, opened 2018-11-13
CORDA-2221: Fix clustered notary identity generation

adagys:
- Don't generate a composite key certificate for CFT notaries
- Don't require a composite key certificate for CFT notaries on startup


adagys:
Tested that it works with Raft and BFT notary samples

PR #4231, opened 2018-11-14
[CORDA-2224]: Removed field `value` from `Validated`. Renamed function `orThrow()` to `value()`.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2224


roastario:
yes, I was caught out by the old mechanism!

PR #4234, opened 2018-11-14
[CORDA-2225] Use platform's non-blocking PRNG when invoking SecureRandom()

kchalkias:
Although we've set the default algorithm for BC operations, other classes and methods like `UUID.randomUUID` directly invoke `SecureRandom()` which internally gets a default PRNG algorithm by looking through all registered providers. Then it returns the first PRNG algorithm of the first provider that has registered a SecureRandom implementation, or null if none of the registered providers supplies a SecureRandom implementation.

This means, that some operations in Corda might still use a blocking RNG.

This PR implements and uses a `PlatformSecureRandomService` that is inserted to `CordaSecurityProvider`, which is the first registered `Provider` in Corda, ensuring that all invocations to `SecureRandom()` use our `platformSecureRandom` (which is the same as `newSecureRandom`).

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
@@ -4,8 +4,11 @@ import net.corda.core.KeepForDJVM
 import net.corda.core.StubOutForDJVM
 import net.corda.core.crypto.CordaObjectIdentifier.COMPOSITE_KEY
 import net.corda.core.crypto.CordaObjectIdentifier.COMPOSITE_SIGNATURE
+import net.corda.core.crypto.internal.platformSecureRandom
 import org.bouncycastle.asn1.ASN1ObjectIdentifier
 import java.security.Provider
+import java.security.SecureRandom
+import java.security.SecureRandomSpi
 
 @KeepForDJVM
 class CordaSecurityProvider : Provider(PROVIDER_NAME, 0.1, "$PROVIDER_NAME security provider wrapper") {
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
```diff
@@ -18,9 +21,30 @@ class CordaSecurityProvider : Provider(PROVIDER_NAME, 0.1, "$PROVIDER_NAME secur
         put("Signature.${CompositeSignature.SIGNATURE_ALGORITHM}", CompositeSignature::class.java.name)
         put("Alg.Alias.Signature.$COMPOSITE_SIGNATURE", CompositeSignature.SIGNATURE_ALGORITHM)
         put("Alg.Alias.Signature.OID.$COMPOSITE_SIGNATURE", CompositeSignature.SIGNATURE_ALGORITHM)
+        putService(PlatformSecureRandomService(this))
     }
 }
 
+
+class PlatformSecureRandomService(provider: Provider)
```
shamsasari: Does this need to be public?

core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
```diff
@@ -18,9 +21,30 @@ class CordaSecurityProvider : Provider(PROVIDER_NAME, 0.1, "$PROVIDER_NAME secur
...
+    : Provider.Service(provider, "SecureRandom", algorithm, PlatformSecureRandomSpi::javaClass.name, null, null) {
+
+    companion object {
+        const val algorithm = "platformPRNG"
```
shamsasari: Does this have any particular meaning? Where does it come from?

kchalkias: Well, we have to define an algorithm name for this SecureRandom impl. What should I put?

core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
kchalkias: I made it public for tests only (to check the algorithm name and ensure that my SecureRandom is invoked). I can annotate it's visible for tests only. Does this work? or you want me to use reflection in my tests?

kchalkias:
@shamsasari wrt to `newSecureRandom()` they are practically the same. Iirc, this was required to keep ENT and OS closer. But as we use newSecureRandom() everywhere, I will change to that here as well.

core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
shamsasari: I see. How about `CordaPRNG`?

core/src/main/kotlin/net/corda/core/crypto/CordaSecurityProvider.kt
shamsasari: Make it `internal` and also add `@CordaInternal`.

shamsasari:
`newSecureRandom()` is public API helper method so that we get the right random object since creating a new `SecureRandom()` wasn't correct. Now that it is, this method feels out of place, because it's not really needed, hence my suggestion for deprecation. If we deprecate then we should replace usages in the codebase with `SecureRandom()`. WDYT?

W.r.t ENT / OS, the function that keeps them together is `platformSecureRandom()`.

kchalkias:
Thanks @shamsasari , so everything is in place now. Deprecating `newSecureRandom()` will be handled in another PR.

PR #4236, opened 2018-11-14
[CORDA-2219] Show message if CorDapp already exists

gatsinski:
This pull request attempts to fix [CORDA-2219](https://r3-cev.atlassian.net/browse/CORDA-2219)

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -33,6 +33,7 @@ import net.corda.serialization.internal.amqp.AbstractAMQPSerializationScheme
 import net.corda.serialization.internal.amqp.amqpMagic
 import java.io.File
 import java.io.InputStream
+import java.nio.file.FileAlreadyExistsException
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import java.security.PublicKey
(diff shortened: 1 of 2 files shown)


gatsinski:
Looks like the tests are failing. I'll examine the problem.

vkolomeyko:
Looks fine to me, looping in @shamsasari for second opinion.
This API stability TC job failure looks rather odd. Let me re-run it.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -226,7 +227,13 @@ internal constructor(private val initSerEnv: Boolean,
             println("Copying CorDapp JARs into node directories")
             for (nodeDir in nodeDirs) {
                 val cordappsDir = (nodeDir / "cordapps").createDirectories()
-                cordappJars.forEach { it.copyToDirectory(cordappsDir) }
+                cordappJars.forEach {
+                    try {
+                        it.copyToDirectory(cordappsDir)
+                    } catch (e: FileAlreadyExistsException) {
+                        println("WARNING: ${it.fileName} already exists in $cordappsDir")
```
shamsasari: Add "..., ignoring and leaving existing CorDapp untouched"

gatsinski: Done.

shamsasari:
Why are we changing the api file? The change is trivial and does not need such a change. I can't approve this modification. What was the original issue?

vkolomeyko:
Yes, API change is a bit unexpected.

Looks like `diff` tool got very confused and was claiming that `equals()` method for one of the `core` classes been removed and added back.

I discussed with @chrisr3 offline and we could not think of any better solution than updating `api-current.txt` for the class in question with current state in `master` branch. Earlier on people been adding methods to this class but did not update `api-current.txt`.

PR #4239, opened 2018-11-15
[CORDA-1826]: Enable large attachments to be uploaded.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1826

### DIFF
--- a/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
+++ b/client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
@@ -17,6 +17,7 @@ import net.corda.core.context.Trace
 import net.corda.core.context.Trace.InvocationId
 import net.corda.core.identity.CordaX500Name
 import net.corda.core.internal.*
+import net.corda.core.messaging.CordaRPCOps
 import net.corda.core.messaging.RPCOps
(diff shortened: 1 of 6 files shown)


client/rpc/src/main/kotlin/net/corda/client/rpc/internal/RPCClientProxyHandler.kt
```diff
@@ -254,16 +256,13 @@ class RPCClientProxyHandler(
...
-            val request = RPCApi.ClientToServer.RpcRequest(
-                    clientAddress,
-                    method.name,
-                    serialisedArguments,
-                    replyId,
-                    sessionId,
-                    externalTrace,
-                    impersonatedActor
-            )
+            val request: RPCApi.ClientToServer = if (method in methodsToUploadAttachment) {
```
exFalso: Why do we need this? This is is now specialized to CordaRPCOps. Can't we decide what to do by looking at the arguments?

exFalso (changes requested):
A test should be added to test whether large messages still deadlock. This is a different lock condition to the Piped* one. It was intermittent and happened while several clients were uploading at the same time. Iirc it wasn't actually a proper deadlock, rather a slowdown to a halt.
Can you add a test that uses out-of-process nodes and uploads from several clients in parallel repeatedly?

sollecitom:
Descoping this for Corda 4.

PR #4241, opened 2018-11-15
CORDA-2197: Update picocli to latest version

anthonykeenan:
Have done some testing on windows and linux and everything seems fine.


anthonykeenan:
Also important...
[#547] Bugfix: Fix infinite loop when print help. Thanks to Patrick Kuo for the pull request.

PR #4242, opened 2018-11-15
ENT-2356 NotaryServiceFlow backpressure

thschroeter:
Implementation of application level notary backpressure, based on the request queue size in the uniqueness provider.

Clients are asked to further delay their timed flows if the ETA of request processing in the notary exceeds the default retry timeout.

When the bottleneck is transport level (Artemis) messaging, not the request processing in the UniquenessProvider, the clients will still retry.

We tested with artificial delay in the uniqueness provider.


node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -86,6 +90,20 @@ class PersistentUniquenessProvider(val clock: Clock, val database: CordaPersiste
     private val commitLog = createMap(cacheFactory)
 
     private val requestQueue = LinkedBlockingQueue<CommitRequest>(requestQueueSize)
+    private val nrQueuedStates = AtomicInteger(0)
```
thschroeter: Issue transactions with time window (without inputs or references) don't contribute to the `nrQueuedStates` metric. Therefore it might be better to just use the number of requests in the queue and measure the throughput in number of requests per second.

blsemo: I am not so sure - tx without states and reference states don't cause much work in this uniqueness provider, either - they pretty much just go through a time-window check. I have a feeling that the number of states to check is a good approximation for the length of the queue - much better than the number of requests that could be very inhomogeneous in number of states per request.

gendal:
Hi Thomas - does this need docs? Anything to be added to changleog/releasenotes?

blsemo:
@gendal This could maybe documented as a feature, but it should be completely transparent to the user/CorDapp implementor.
We probably should add it to the release notes to document from which version on this behaviour change can be expected.

core/src/main/kotlin/net/corda/core/flows/NotaryWireFormat.kt
```diff
@@ -80,4 +80,8 @@ data class NotarisationPayload(val transaction: Any, val requestSignature: Notar
 
 /** Payload returned by the notary service flow to the client. */
 @CordaSerializable
-data class NotarisationResponse(val signatures: List<TransactionSignature>)
\ No newline at end of file
+data class NotarisationResponse(val signatures: List<TransactionSignature>)
+
+/** Sent by the notary when the notary detects it will unlikely respond before the client retries. */
+@CordaSerializable
+data class WaitTimeUpdate(val waitTimeSeconds: Long)
```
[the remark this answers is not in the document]
blsemo: These back-pressure messages only are sent if there is a serious delay (in the order of 30s or more) - no need for millisecond granularity. On the other hand 300 seconds is easier to parse as 5 minutes than 300000 ms.

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -86,6 +90,20 @@ class PersistentUniquenessProvider(val clock: Clock, val database: CordaPersiste
...
+
+    private val throughputMeter = Meter()
+
+    // Estimated time of request processing.
+    override fun eta(): Duration {
+        val rate = throughputMeter.oneMinuteRate
+        val nrStates = nrQueuedStates.get()
+        log.debug { "rate: $rate, queueSize: $nrStates" }
+        if (rate > 1e-5 && nrStates > 5) {
```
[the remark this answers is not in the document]
thschroeter: I'll comment and adjust the logic slightly to send 1 second if the queue is small and a bounded eta when the queue is large and the estimated throughput small.

.ci/api-current.txt
```diff
@@ -2266,7 +2266,7 @@ public final class net.corda.core.flows.NotaryFlow extends java.lang.Object
 ##
 @DoNotImplement
 @InitiatingFlow
-public static class net.corda.core.flows.NotaryFlow$Client extends net.corda.core.flows.FlowLogic implements net.corda.core.internal.TimedFlow
+public static class net.corda.core.flows.NotaryFlow$Client extends net.corda.core.internal.BackpressureAwareTimedFlow
```
rick-r3: Yep.  This appears to extend all of the prior interfaces / super classes, just in a slightly different hierarchy.  So if anyone was directly calling this, and for some bizarre reason doing some casting to one of them it should be fine.

adagys: We could in theory do away without this abstract class, but I think this is a preferrable solution

core/src/main/kotlin/net/corda/core/flows/NotaryWireFormat.kt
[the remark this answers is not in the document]
rick-r3: Use `Duration` here ?  All the `java.time` stuff has custom serializers so is definitely supported.

core/src/main/kotlin/net/corda/core/internal/BackpressureAwareTimedFlow.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Implementation of TimedFlow that can handle WaitTimeUpdate messages. Any flow talking to the notary should implement this and use
+ * explicit send and this class's receiveResultOrTiming to receive the response to handle cases where the notary sends a timeout update.
+ *
+ * This is handling the special case of the notary where the notary service will have an internal queue on the uniqueness provider and we
+ * want to stop retries overwhelming that internal queue. As the TimedFlow mechanism and the notary service back-pressure are very specific
+ * to this use case at the moment, this implementation is internal and not for general use.
+ */
+abstract class BackpressureAwareTimedFlow<ResultType, ReceiveType>(private val receiveType: Class<ReceiveType>) : FlowLogic<ResultType>(), TimedFlow {
```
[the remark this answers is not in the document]
blsemo: No I can't - a member function cannot be reified, and a non-member function wouldn't have access to the stateMachine field.

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -86,6 +90,20 @@ class PersistentUniquenessProvider(val clock: Clock, val database: CordaPersiste
...
+            return Duration.ofSeconds((1.5 * nrStates / rate).toLong())
```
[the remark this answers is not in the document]
blsemo: Maybe let's do that in a different PR - that would require to introduce an `extraConfig` class for the SimpleNotary, so all doable but fair amount of work.

core/src/main/kotlin/net/corda/core/internal/notary/UniquenessProvider.kt
```diff
@@ -25,9 +25,14 @@ interface UniquenessProvider {
             references: List<StateRef> = emptyList()
     ): CordaFuture<Result>
 
-    // Estimated time of request processing.
+    /**
+     *    Estimated time of request processing. A uniqueness provider that is aware of his own throughput can return
+     *    an estimate how long requests will be queued before they can be processed. Notary services use this information
+     *    to potentially update clients with an expected wait time in order to avoid spamming by retries when the notary
+     *    gets busy
```
thschroeter: nitpick: missing a full stop at the end

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -95,15 +92,20 @@ class PersistentUniquenessProvider(val clock: Clock, val database: CordaPersiste
     /** Measured in states per second. */
     private val throughputMeter = Meter()
 
-    /** Estimated time of request processing. */
+    /** Estimated time of request processing.
+     *  This uses performance metrics to gauge how long the wait time for a newly queued state will probably be.
+     *  It checkst that there is actual traffic going on (i.e. a non-trivial number of states are queued and there
```
thschroeter: s/checkst/checks/

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -147,6 +148,10 @@ data class NotaryConfig(
         val serviceLegalName: CordaX500Name? = null,
         /** The name of the notary service class to load. */
         val className: String = "net.corda.node.services.transactions.SimpleNotaryService",
+        /**If the wait time estimate on the internal queue exceeds this value, the notary may send
+         * a wait time update to the client (implementation specific and dependent on the counter
+         * party version) */
```
adagys: missing '.' and newlines before & after text

node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -147,6 +148,10 @@ data class NotaryConfig(
...
+        val etaMessageThresholdSeconds: Int = NotaryServiceFlow.defaultEstimatedWaitTime,
```
adagys: shoulde the waittime default also be suffixed with "Seconds"?

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -86,6 +87,26 @@ class PersistentUniquenessProvider(val clock: Clock, val database: CordaPersiste
     private val commitLog = createMap(cacheFactory)
 
     private val requestQueue = LinkedBlockingQueue<CommitRequest>(requestQueueSize)
+    private val nrQueuedStates = AtomicInteger(0)
+
+    /** Measured in states per second. */
+    private val throughputMeter = Meter()
+
+    /** Estimated time of request processing.
```
adagys: newline before text

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -86,6 +87,26 @@ class PersistentUniquenessProvider(val clock: Clock, val database: CordaPersiste
...
+     *  It checkst that there is actual traffic going on (i.e. a non-trivial number of states are queued and there
+     *  is actual throughput) and then returns the expected wait time scaled up by a factor to give a probable
+     *  upper bound.
+     */
+    override fun eta(): Duration {
+        val rate = throughputMeter.oneMinuteRate
+        val nrStates = nrQueuedStates.get()
+        log.debug { "rate: $rate, queueSize: $nrStates" }
+        if (rate > 1e-5 && nrStates > 100) {
+            return Duration.ofSeconds((2 * nrStates / rate).toLong())
```
adagys: we discussed with @thschroeter to add an upper bound to the wait time estimate. If it's more than several hours then something has gone wrong somewhere

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -587,6 +590,14 @@ class SingleThreadedStateMachineManager(
     }
 
     private fun resetCustomTimeout(flowId: StateMachineRunId, timeoutSeconds: Long) {
+        if (timeoutSeconds < serviceHub.configuration.flowTimeout.timeout.seconds) {
+            logger.warn("Ignoring request to set time-out on timed flow $flowId to $timeoutSeconds seconds which is shorter than default of ${serviceHub.configuration.flowTimeout.timeout.seconds} seconds.")
```
adagys: this can probably be info

node/src/test/kotlin/net/corda/node/services/TimedFlowTests.kt
```diff
@@ -194,6 +238,32 @@ class TimedFlowTests {
...
+                (issueTx + notarySignatures).verifyRequiredSignatures()
+                progressTrackerDone.get()
+            }
+        } finally {
+            waitETA = waitEtaThresholdSeconds.seconds
+        }
+    }
+
+
+
```
adagys: extra lines

adagys:
Should we run a perf test before merging?

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -579,22 +579,49 @@ class SingleThreadedStateMachineManager(
...
                 flow.fiber.scheduleEvent(event)
-            }, jitteredDelaySeconds, TimeUnit.SECONDS)
+            }, delay, TimeUnit.SECONDS)
+        }
+    }
+
+    private fun calculateDefaultTimeoutSeconds(retryCount: Int): Long {
+        return with(serviceHub.configuration.flowTimeout) {
+            val timeoutDelaySeconds = timeout.seconds * Math.pow(backoffBase, retryCount.toDouble()).toLong()
+            maxOf(1L, timeoutDelaySeconds / 2 + (Math.random() * timeoutDelaySeconds / 2).toLong())
```
rick-r3: `((1.0 + Math.random()) * timeoutDelaySeconds) / 2.0).toLong()` is more readable I think.

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -579,22 +579,49 @@ class SingleThreadedStateMachineManager(
...
-            val scheduledFuture = scheduleTimeoutException(flow, retryCount)
+            val scheduledFuture = scheduleTimeoutException(flow, calculateDefaultTimeoutSeconds(retryCount))
             timedFlows[flowId] = ScheduledTimeout(scheduledFuture, retryCount + 1)
         } else {
             logger.warn("Unable to schedule timeout for flow $flowId – flow not found.")
         }
     }
 
+    private fun resetCustomTimeout(flowId: StateMachineRunId, timeoutSeconds: Long) {
+        mutex.locked {
```
rick-r3: Did we defend against a very short timeout being sent to us?  I feel like we should do a `max(timeoutSeconds, calculateDefaultTimeoutSeconds(0))` on the timeout here to prevent it going lower than our lowest configured value.

rick-r3: Should we also have a maximum?  We could calculate the maximum from `calculateDefaultTimeoutSeconds(serviceHub.configuration.maxRestartCount)`

thschroeter: We discussed in person that we remove the maximum on the client side. Restarting a node will restart the idempotent notarisation flows.

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
rick-r3: I reckon debug.  It's going to happen a lot isn't it?.... because the notary wants to send the ETA to stop the timeout, but doesn't know what the client has configured so will always be somewhat generous and say low values, potentially.

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
```diff
@@ -15,16 +18,30 @@ import net.corda.core.utilities.unwrap
...
+         * Also used as default eta message threshold so that a default wait time/default threshold will never
+         * lead to an update message being sent.
+         */
+        val defaultEstimatedWaitTime: Duration = 10.seconds
     }
 
     private var transactionId: SecureHash? = null
 
+    @Suspendable
+    private fun counterpartyCanHandleBackPressure() = otherSideSession.getCounterpartyFlowInfo(true).flowVersion >= MIN_PLATFORM_VERSION_FOR_BACKPRESSURE_MESSAGE
```
adagys: guess would be technically more correct to say "notary" rather than "counterparty"

core/src/main/kotlin/net/corda/core/internal/notary/NotaryService.kt
```diff
@@ -17,9 +18,12 @@ abstract class NotaryService : SingletonSerializeAsToken() {
     abstract fun start()
     abstract fun stop()
 
+    /** Estimated wait time until incoming requests are processed. */
+    open fun getEstimatedWaitTime(): Duration = NotaryServiceFlow.defaultEstimatedWaitTime
```
adagys: this abstract class is used by bft notaries as well, for which this backpressure mechanism doesn't apply yet. Should we move this function down?

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -579,22 +589,53 @@ class SingleThreadedStateMachineManager(
...
         }
     }
 
+    private fun resetCustomTimeout(flowId: StateMachineRunId, timeoutSeconds: Long) {
+        if (timeoutSeconds < serviceHub.configuration.flowTimeout.timeout.seconds) {
+            logger.debug { "Ignoring request to set time-out on timed flow $flowId to $timeoutSeconds seconds which is shorter than default of ${serviceHub.configuration.flowTimeout.timeout.seconds} seconds." }
+            return
+        }
+        mutex.locked {
+            resetCustomTimeout(flowId, timeoutSeconds)
```
adagys: do we log anywhere that the timeout is being reset? I think we should if not

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
rick-r3: No, it's checking that the calling counterparty can handle an ETA response.

adagys: Ah right

node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
rick-r3: We could log at debug, yeah.

core/src/main/kotlin/net/corda/core/internal/notary/SinglePartyNotaryService.kt
```diff
@@ -61,7 +65,12 @@ abstract class SinglePartyNotaryService : NotaryService() {
         }
     }
 
-    override fun getEstimatedWaitTime(): Duration = uniquenessProvider.eta()
+    /**
+     * Estimate the wait time to be notarised and start taking into account the new request size.
+     *
+     * @param numStates The number of states we're about to request be notarised.
+     */
+    fun getEstimatedWaitTimeAndAddPendingRequest(numStates: Int): Duration = uniquenessProvider.getEtaAndAddPendingRequest(numStates)
```
adagys: since there's no other getEta method we can keep the old name? the pending request bit is an implementation detail

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
```diff
@@ -111,15 +111,17 @@ class PersistentUniquenessProvider(val clock: Clock, val database: CordaPersiste
...
+     * @param numStates The number of states (input + reference) we're about to request be notarised.
      */
-    override fun eta(): Duration {
+    override fun getEtaAndAddPendingRequest(numStates: Int): Duration {
         val rate = throughput
-        val nrStates = nrQueuedStates.get()
+        val nrStates = nrQueuedStates.getAndAdd(numStates)
         log.debug { "rate: $rate, queueSize: $nrStates" }
-        if (rate > 0.0 && nrStates > 100) {
+        if (rate > 0.0 && nrStates > 0) {
```
adagys: this will now return an eta message for every request, is that desirable?

core/src/main/kotlin/net/corda/core/internal/notary/SinglePartyNotaryService.kt
```diff
@@ -61,7 +65,12 @@ abstract class SinglePartyNotaryService : NotaryService() {
         }
     }
 
-    override fun getEstimatedWaitTime(): Duration = uniquenessProvider.eta()
+    /**
+     * Estimate the wait time to be notarised and start taking into account the new request size.
```
adagys: 'and start' not needed

node/src/main/kotlin/net/corda/node/services/transactions/PersistentUniquenessProvider.kt
rick-r3: Only if the ETA is greater than the threshold.  Yes.

adagys: Right, ok

PR #4252, opened 2018-11-19
CORDA-2239: Clean up Driver.startNode overload which specifies node parameters with default values

shamsasari:
This overload is not useful for Java devs which is why NodeParameters was created in the first place. The startNode(NodeParameters) overload is now used instead.

NodeParameters.logLevel removed as it doesn't do anything and is not in V3.

### DIFF
--- a/node/src/integration-test/kotlin/net/corda/node/flows/AsymmetricCorDappsTests.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/flows/AsymmetricCorDappsTests.kt
@@ -11,6 +11,7 @@ import net.corda.testing.core.ALICE_NAME
 import net.corda.testing.core.BOB_NAME
 import net.corda.testing.core.singleIdentity
 import net.corda.testing.driver.DriverParameters
+import net.corda.testing.driver.NodeParameters
 import net.corda.testing.driver.driver
 import net.corda.testing.node.internal.cordappForClasses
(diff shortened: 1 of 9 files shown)


shamsasari:
The key change is in DriverDSL.kt

mikehearn:
What's the rationale for this? The old method seems useful and cleaner, plus every app uses it?

shamsasari:
Each time we introduce a new node parameter it has to be added to that `startNode` overload and `NodeParameters`, which seems redundant. Or are you suggesting we don't deprecate `startNode`, keep its node parameters fixed and new parameters only go in `NodeParameters`.

mikehearn:
That would seem OK to me, yeah, or just keep adding named parameters (we don't have to add all the overloads?).

sollecitom:
Well, if we do not do this, and Java developers use `startNode(a, b, c)`, we need to add an overloaded method not to break the API, so `startNode(a,b,c)` and `startNode(a,b,c,d)`.

shamsasari:
Thanks @sollecitom for that, I completely forgot to mention that issue. Each new release that adds default parameters will force us to add a new startNode overload to preserve the API, which is completely ironic as this overload only benefits kotlin users.

As discussed @mikehearn I will un-deprecate but add a note to say that new params will only be introduced in `NodeParameters`.

PR #4257, opened 2018-11-19
CORDA-2099 serialisation rewrite

distributedleetravis:
A large-scale rewrite which addresses [CORDA-2147](https://r3-cev.atlassian.net/browse/CORDA-2147), [CORDA-1702](https://r3-cev.atlassian.net/browse/CORDA-1702) and [CORDA-2139](https://r3-cev.atlassian.net/browse/CORDA-2139)

We break `SerializerFactory` down into two interfaces representing separate code paths:

* `LocalSerializerFactory`, which provides serialisers into and out of local `Type`s
* `RemoteSerializerFactory` which provides deserialisers for remote data for which we have AMQP `Schema` information.

`RemoteSerializerFactory` has the tasks of finding local `Type`s that reflect the remote schema it has been presented with, carpenting up any classes that are missing in the local classloader, and determining whether evolution is necessary between mismatched versions of types.

`LocalSerializerFactory` builds and caches serializers for local types, which `RemoteSerializerFactory` will use for deserialisation if no evolution is required.

The primary mechanism for determining whether remote and local types are compatible / require evolution is to compare their type descriptors. A rewritten `FingerPrinter` (using `LocalTypeInformation` already gathered, rather than reflection) generates descriptors for composable types.

The serialisers for composable types are rewritten to use `LocalTypeInformation`. Both normal and evolution serialisers make use of a common interface, `ObjectBuilder`, which provides a numbered series of "slots" into which values can be written. Once all of the slots have been filled, the `ObjectBuilder` can `build` an object, either by calling a constructor with an array of parameters or by calling a no-argument constructor followed by setters for all of the properties it holds values for. The difference between normal deserialisation and evolution deserialisation is that in the evolution case the slots may be re-ordered, and values placed in some slots may be discarded.

A new `EvolutionSerializerFactory` determines whether evolution is necessary in cases where the fingerprints of local and remote types do not match, by comparing the type and property information we hold for both. It should eliminate spurious evolution caused by type erasure, or other changes to the signature of a type (such as adding a new interface) which do not alter how it is serialised and deserialised.

A lot of old code-paths are removed, and a few pieces of code here and there that still used them are rewritten to use the new framework.


serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+            fingerprintType(type).writer.fingerprint
+
+    // This method concatenates various elements of the types recursively as unencoded strings into the hasher,
+    // effectively creating a unique string for a type which we then hash in the calling function above.
+    private fun fingerprintType(type: LocalTypeInformation): CustomisableLocalTypeInformationFingerPrintingState = apply {
+        // Don't go round in circles.
+        when {
+            hasSeen(type.typeIdentifier) -> writer.writeAlreadySeen()
+            type is LocalTypeInformation.Cycle -> fingerprintType(type.follow)
+            else -> ifThrowsAppend({ type.observedType.typeName }, {
```
mikehearn: Is reading this string so expensive it really needs a lambda?

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+            is LocalTypeInformation.Cycle ->
+                throw IllegalStateException("Cyclic references must be dereferenced before fingerprinting")
+            is LocalTypeInformation.Unknown,
+            is LocalTypeInformation.Top -> writer.writeAny()
+            is LocalTypeInformation.AnArray -> {
+                fingerprintType(type.componentType)
+                writer.writeArray()
+            }
+            is LocalTypeInformation.ACollection -> fingerprintCollection(type)
+            is LocalTypeInformation.AMap -> fingerprintMap(type)
```
mikehearn: I guess you could have used overloading here instead of a big switch, but it's no big deal, leave it as is.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+
+    private fun fingerprintOpaque(type: LocalTypeInformation) =
+            fingerprintWithCustomSerializerOrElse(type) {
+                fingerprintName(type)
+            }
+
+    private fun fingerprintInterface(type: LocalTypeInformation.AnInterface) =
+            fingerprintWithCustomSerializerOrElse(type) {
+                fingerprintName(type)
+                writer.writeAlreadySeen() // replicate behaviour of old fingerprinter
```
mikehearn: I assume from the comment this curious behaviour was just a bug that we have to remain compatible with?

Are there other places like this? If so maybe we should mark them with a special marker comment in some way, so if/when we next rev the wire protocol we can clean these things up.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+
+    private fun fingerprintProperties(properties: Map<String, LocalPropertyInformation>) =
+            properties.asSequence().sortedBy { it.key }.forEach { (propertyName, propertyType) ->
+                val (neverMandatory, adjustedType) = adjustType(propertyType.type)
+                fingerprintType(adjustedType)
+                writer.write(propertyName)
+                if (propertyType.isMandatory && !neverMandatory) writer.writeNotNullable() else writer.writeNullable()
+            }
+
+    // Compensate for the serialisation framework's forcing of char to Character
```
mikehearn: Same thing I guess - let's mark these cases with FIXUP or some other easily greppable word to let us find the places we'd like to fix, in future.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+    // generate different fingerprints for class Outer<T>(val a: Inner<T>) when serialising
+    // and deserializing (assuming deserialization is occurring in a factory that didn't
+    // serialise the object in the  first place (and thus the cache lookup fails). This is also
+    // true of Any, where we need  Example<A, B> and Example<?, ?> to have the same fingerprint
+    private fun hasSeen(type: TypeIdentifier) = (type in typesSeen)
+            && (type != TypeIdentifier.UnknownType)
+}
+
+// region Utility functions
+internal fun fingerprintForDescriptors(vararg typeDescriptors: String): String =
```
mikehearn: This is only used in one place. Maybe just inline it into the call site?

serialization/src/main/kotlin/net/corda/serialization/internal/model/RemoteTypeReflector.kt
```diff
@@ -0,0 +1,46 @@
+package net.corda.serialization.internal.model
+
+/**
+ * Represents the reflection of some [RemoteTypeInformation] by some [LocalTypeInformation], which we use to make
+ * decisions about evolution.
+ */
+data class ReflectedTypeInformation(
```
mikehearn: We have a (soft) convention that these sorts of pair classes are called "FooAndBar", so this would be `LocalAndRemoteTypeInformation` which maybe is a bit more descriptive.

serialization/src/main/kotlin/net/corda/serialization/internal/model/RemoteTypeReflector.kt
```diff
@@ -0,0 +1,46 @@
...
+        val remoteTypeInformation: RemoteTypeInformation,
+        val localTypeInformation: LocalTypeInformation)
+
+/**
+ * Given [RemoteTypeInformation], a [RemoteTypeReflector] will find (or, if necessary, create) a local type reflecting
+ * the remote type.
+ */
+ interface RemoteTypeReflector {
```
mikehearn: Is it worth having these sorts of interfaces with only one implementation in internal code? If we just renamed TypeLoadingRemoteTypeReflector to `class RemoteTypeReflector` and dropped this interface, would the code suffer?

serialization/src/main/kotlin/net/corda/serialization/internal/model/RemoteTypeReflector.kt
```diff
@@ -0,0 +1,46 @@
...
+ * whether evolution is necessary.
+ *
+ * @param typeLoader The [TypeLoader] to use to load local types reflecting remote types.
+ * @param localTypeModel The [LocalTypeModel] to use to obtain [LocalTypeInformation] for the local [Type]s so obtained.
+ */
+class TypeLoadingRemoteTypeReflector(
+        private val typeLoader: TypeLoader,
+        private val localTypeModel: LocalTypeModel): RemoteTypeReflector {
+
+    override fun reflect(remoteInformation: Map<TypeDescriptor, RemoteTypeInformation>):
```
mikehearn: This method is itself only used in one place. Are we sure this pair of class+interface are carrying their conceptual weight? It seems like a method a few lines long has expanded quite a bit here.

serialization-deterministic/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPSerializerFactories.kt
```diff
@@ -16,8 +16,8 @@ fun createSerializerFactoryFactory(): SerializerFactoryFactory = DeterministicSe
 private class DeterministicSerializerFactoryFactory : SerializerFactoryFactory {
     override fun make(context: SerializationContext) =
             SerializerFactoryBuilder.build(
-                whitelist = context.whitelist,
-                classCarpenter = DummyClassCarpenter(context.whitelist, context.deserializationClassLoader))
+            whitelist = context.whitelist,
```
mikehearn: Bad whitespace change, in a few other places too. You can leave them, we'll do a global reformat after C4.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/CollectionSerializer.kt
```diff
@@ -33,40 +35,56 @@ class CollectionSerializer(private val declaredType: ParameterizedType, factory:
...
         ))
 
-        private fun findConcreteType(clazz: Class<*>): (List<*>) -> Collection<*> {
-            return supportedTypes[clazz] ?: throw AMQPNotSerializableException(
-                    clazz,
-                    "Unsupported collection type $clazz.",
-                    "Supported Collections are ${supportedTypes.keys.joinToString(",")}")
+        private val supportedTypeIdentifiers = supportedTypes.keys.asSequence().map { TypeIdentifier.forClass(it) }.toSet()
+
+        fun resolveDeclared(declaredTypeInformation: LocalTypeInformation.ACollection): LocalTypeInformation.ACollection {
```
mikehearn: I could use a comment explaining what this logic is doing - something related to type erasure?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
```diff
@@ -0,0 +1,265 @@
...
+interface PropertyReadStrategy {
+
+    companion object {
+        /**
+         * Select the correct strategy for reading properties, based on the property type.
+         */
+        fun make(name: String, typeIdentifier: TypeIdentifier, type: Type): PropertyReadStrategy =
+                if (AMQPTypeIdentifiers.isPrimitive(typeIdentifier)) {
+                    when (typeIdentifier) {
+                        in characterTypes -> AMQPCharPropertyReadStrategy
```
mikehearn: // FIXUP

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
```diff
@@ -0,0 +1,265 @@
...
+    companion object {
+        /**
+         * Make a [PropertyReader] based on the provided [LocalPropertyInformation].
+         */
+        fun make(propertyInformation: LocalPropertyInformation) = when(propertyInformation) {
+            is LocalPropertyInformation.GetterSetterProperty -> GetterReader(propertyInformation.observedGetter)
+            is LocalPropertyInformation.ConstructorPairedProperty -> GetterReader(propertyInformation.observedGetter)
+            is LocalPropertyInformation.ReadOnlyProperty -> GetterReader(propertyInformation.observedGetter)
+            is LocalPropertyInformation.CalculatedProperty -> GetterReader(propertyInformation.observedGetter)
+            is LocalPropertyInformation.PrivateConstructorPairedProperty -> FieldReader(propertyInformation.observedField)
```
mikehearn: Do we really need to be able to reflect private fields like this? Where in Corda do we depend on this? Maybe Kat or Rick remembers why this feature was added?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * Get the value of the property from the supplied instance, or null if the instance is itself null.
+     */
+    abstract fun read(obj: Any?): Any?
+
+    /**
+     * Reads a property using a getter [Method].
+     */
+    class GetterReader(_getter: Method): PropertyReader() {
+        val getter = _getter.also { it.isAccessible = true }
```
mikehearn: You could just write this as: 
```
class GetterReader(val getter: Method) : PropertyReader() {
    init { getter.isAccessible = true }
}
```

Introducing two names seems a bit overkill for that.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
```diff
@@ -0,0 +1,265 @@
...
+
+        override fun read(obj: Any?): Any? = if (obj == null) null else getter.invoke(obj)
+    }
+
+    /**
+     * Reads a property using a backing [Field].
+     */
+    class FieldReader(_field: Field): PropertyReader() {
+        val field = _field.also { it.isAccessible = true }
```
mikehearn: same

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/custom/ThrowableSerializer.kt
```diff
@@ -19,15 +22,32 @@ class ThrowableSerializer(factory: SerializerFactory) : CustomSerializer.Proxy<T
 
     override val additionalSerializers: Iterable<CustomSerializer<out Any>> = listOf(StackTraceElementSerializer(factory))
 
+    private val LocalTypeInformation.properties: Map<String, LocalPropertyInformation> get() = when(this) {
```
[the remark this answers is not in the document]
distributedleetravis: Yep. Removed.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/AMQPPrimitiveSerializer.kt
```diff
@@ -12,7 +12,7 @@ import java.lang.reflect.Type
  * [ByteArray] is automatically marshalled to/from the Proton-J wrapper, [Binary].
  */
 class AMQPPrimitiveSerializer(clazz: Class<*>) : AMQPSerializer<Any> {
-    override val typeDescriptor = Symbol.valueOf(SerializerFactory.primitiveTypeName(clazz)!!)!!
+    override val typeDescriptor = Symbol.valueOf(AMQPTypeIdentifiers.primitiveTypeName(clazz)!!)!!
```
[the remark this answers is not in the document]
distributedleetravis: Agreed - it's only ever used in cases where a non-null result is expected. Changed it to throw a meaningful exception if called with a non-primitive type.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+        if (debugBuffer != null) logger.info("$fingerprint from $debugBuffer")
+        return fingerprint
+    }
+}
+
+/**
+ * Representation of the current state of fingerprinting, which keeps track of which types have already been visited
+ * during fingerprinting.
+ */
+private class CustomisableLocalTypeInformationFingerPrintingState(
```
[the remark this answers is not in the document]
distributedleetravis: Changed to `FingerPrintingState`.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+    fun writeNullable() = append(NULLABLE_HASH)
+    fun writeNotNullable() = append(NOT_NULLABLE_HASH)
+    fun writeAny() = append(ANY_TYPE_HASH)
+
+    private fun append(chars: CharSequence) = apply {
+        debugBuffer?.append(chars)
+        hasher = hasher.putUnencodedChars(chars)
+    }
+
+    val fingerprint: String get() {
```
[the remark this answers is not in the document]
distributedleetravis: Agreed; changed.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+
+    companion object {
+        private const val ARRAY_HASH: String = "Array = true"
+        private const val ENUM_HASH: String = "Enum = true"
+        private const val ALREADY_SEEN_HASH: String = "Already seen = true"
+        private const val NULLABLE_HASH: String = "Nullable = true"
+        private const val NOT_NULLABLE_HASH: String = "Nullable = false"
+        private const val ANY_TYPE_HASH: String = "Any type = true"
+
+        val ANY = FingerprintWriter(false).writeAny().fingerprint
```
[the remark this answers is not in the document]
distributedleetravis: Removed.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+
+    /**
+     * Fingerprint the type recursively, and return the encoded fingerprint written into the hasher.
+     */
+    fun fingerprint(type: LocalTypeInformation): String =
+            fingerprintType(type).writer.fingerprint
+
+    // This method concatenates various elements of the types recursively as unencoded strings into the hasher,
+    // effectively creating a unique string for a type which we then hash in the calling function above.
+    private fun fingerprintType(type: LocalTypeInformation): CustomisableLocalTypeInformationFingerPrintingState = apply {
```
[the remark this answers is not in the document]
distributedleetravis: Purely to make it return itself for further method chaining, as in `fingerprintType(type).writer.fingerprint`.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+        // Don't go round in circles.
+        when {
+            hasSeen(type.typeIdentifier) -> writer.writeAlreadySeen()
+            type is LocalTypeInformation.Cycle -> fingerprintType(type.follow)
```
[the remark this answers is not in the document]
distributedleetravis: It was also a way to avoid having to identify the cycles within the generated graph and patch the mutable variable with the referred-to type information.

I've found a way to keep track of generated cycles during type information building, and then patch them as soon as the builder's done. This saves us having to keep around a lambda with references to the state of the builder - it's just a `lateinit` var now.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
distributedleetravis: Definitely not, but that's the interface of `ifThrowsAppend`, which the original fingerprinter used, which builds helpful lists of types so we can identify where things have gone wrong. Perhaps it's used that way for good reason elsewhere.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
distributedleetravis: The idea is that `LocalTypeInformation` shouldn't have to know about things like fingerprinting. The other way to do this sort of thing is with a visitor pattern, of course, but an exhaustive `when` clause is less trouble to implement.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
distributedleetravis: That's right - `CompatibilityTest` breaks if we don't do this. Marked `// FIXUP`.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
distributedleetravis: Agreed - inlined.

serialization/src/main/kotlin/net/corda/serialization/internal/model/RemoteTypeReflector.kt
distributedleetravis: Agreed; changed.

serialization/src/main/kotlin/net/corda/serialization/internal/model/RemoteTypeReflector.kt
distributedleetravis: In this case, I agree that it's not really pulling its weight. Removed the interface and class, and moved the functionality to a method in `RemoteSerializerFactory`.

serialization/src/main/kotlin/net/corda/serialization/internal/model/RemoteTypeReflector.kt
distributedleetravis: Agreed (see above).

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
distributedleetravis: I believe it's deprecated functionality, but there are unit tests which test for it, so I've reproduced it (along with every other test-visible quirk of the existing framework...)

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/CollectionSerializer.kt
distributedleetravis: Comment added.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
distributedleetravis: @fenryka may remember why this exception was necessary...

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
distributedleetravis: Agreed, changed.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
distributedleetravis: As above.

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeModellingFingerPrinter.kt
```diff
@@ -0,0 +1,240 @@
...
+        private const val NOT_NULLABLE_HASH: String = "Nullable = false"
+        private const val ANY_TYPE_HASH: String = "Any type = true"
+
+        val ANY = FingerprintWriter(false).writeAny().fingerprint
+
+        private val logger = contextLogger()
+    }
+
+    private val debugBuffer: StringBuilder? = if (debugEnabled) StringBuilder() else null
+    private var hasher = Hashing.murmur3_128().newHasher()
```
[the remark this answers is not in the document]
distributedleetravis: Marked //FIXUP

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
```diff
@@ -0,0 +1,228 @@
...
+     */
+    fun getOrBuildTransform(name: String, builder: () -> EnumMap<TransformTypes, MutableList<Transform>>):
+            EnumMap<TransformTypes, MutableList<Transform>>
+}
+
+/**
+ * A [LocalSerializerFactory] equipped with a [LocalTypeModel] and a [FingerPrinter] to help it build fingerprint-based descriptors
+ * and serializers for local types.
+ */
+class DefaultLocalSerializerFactory(
```
mikehearn: What's the thread safety level of this class? The SerializerFactory kdocs say it's meant to be thread safe, if so, do we need to protect our caches here? At least, annotate it as `@ThreadSafe`. Is there a risk of it being invoked by multiple threads simultaneously during multi-threaded deserialisation and if so, how is that handled? If not, can it be marked as explicitly not thread safe?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
```diff
@@ -0,0 +1,228 @@
...
+            else -> customSerializerRegistry.findCustomSerializer(clazz, declaredType) ?:
+                makeNonCustomSerializer(type, typeInformation, clazz)
+        }
+    }
+
+    private fun makeNonCustomSerializer(type: Type, typeInformation: LocalTypeInformation, clazz: Class<*>): AMQPSerializer<Any> = when {
+        onlyCustomSerializers -> throw AMQPNotSerializableException(type, "Only allowing custom serializers")
+        type.isArray() ->
+            if (clazz.componentType.isPrimitive) PrimArraySerializer.make(type, this)
+            else ArraySerializer.make(type, this)
```
mikehearn: Are we missing a whitelist.requireWhitelisted call for the component type of the array?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/MapSerializer.kt
```diff
@@ -18,11 +20,10 @@ private typealias MapCreationFunction = (Map<*, *>) -> Map<*, *>
  * Serialization / deserialization of certain supported [Map] types.
  */
 @KeepForDJVM
-class MapSerializer(private val declaredType: ParameterizedType, factory: SerializerFactory) : AMQPSerializer<Any> {
-    override val type: Type = (declaredType as? DeserializedParameterizedType)
-            ?: DeserializedParameterizedType.make(SerializerFactory.nameForType(declaredType), factory.classloader)
-    override val typeDescriptor: Symbol = Symbol.valueOf(
-            "$DESCRIPTOR_DOMAIN:${factory.fingerPrinter.fingerprint(type)}")
+class MapSerializer(private val declaredType: ParameterizedType, factory: LocalSerializerFactory) : AMQPSerializer<Any> {
```
mikehearn: I noticed a TODO further down this class suggesting we don't actually check we're inserting the right object types into the map anywhere, asking if that's a security hole. Sounds like it'd at least be unexpected and we _should_ be checking?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/RemoteSerializerFactory.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    private fun RemoteTypeInformation.isEvolvableTo(localTypeInformation: LocalTypeInformation): Boolean = when(this) {
+        is RemoteTypeInformation.Composable -> localTypeInformation is LocalTypeInformation.Composable
+        is RemoteTypeInformation.AnEnum -> localTypeInformation is LocalTypeInformation.AnEnum
+        else -> false
+    }
+
+    private fun RemoteTypeInformation.isDeserialisableWithoutEvolutionTo(localTypeInformation: LocalTypeInformation) = when(this) {
```
mikehearn: mmm, nested whens ... maybe converting to block body and using braces a bit more would make this a bit easier to visually decode.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactory.kt
```diff
@@ -47,413 +29,13 @@ data class CustomSerializersCacheKey(val clazz: Class<*>, val declaredType: Type
 // TODO: need to support super classes as well as interfaces with our current code base... what's involved?  If we continue to ban, what is the impact?
 @KeepForDJVM
 @ThreadSafe
-interface SerializerFactory {
```
mikehearn: Are any of the TODOs above this interface resolved by any of this work? Some of them are pretty obscure, others sound more serious. I'm not sure they're up to date anymore. The Kdoc at least probably needs to be adjusted.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TypeNotationGenerator.kt
```diff
@@ -0,0 +1,73 @@
...
+            is LocalTypeInformation.AMap -> "*" to listOf(type.amqpTypeName)
+            else -> type.amqpTypeName to emptyList()
+        }
+
+        val defaultValue: String? = defaultValues[type.typeIdentifier]
+
+        return Field(name, typeName, requires, defaultValue, null, isMandatory, false)
+    }
+
+    private val defaultValues = sequenceOf(
```
mikehearn: I don't fully grok what this is about - where is this data used?

serialization/src/test/java/net/corda/serialization/internal/carpenter/JavaCalculatedValuesToClassCarpenterTest.java
```diff
@@ -17,6 +17,7 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+/*
```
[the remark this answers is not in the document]
distributedleetravis: It used some things that no longer exist, and needed rewriting. Have now rewritten it.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
```diff
@@ -0,0 +1,269 @@
...
+}
+
+class AMQPPropertyWriteStrategy(private val reader: PropertyReader) : PropertyWriteStrategy {
+    override fun writeClassInfo(output: SerializationOutput) {}
+
+    override fun writeProperty(obj: Any?, data: Data, output: SerializationOutput,
+                               context: SerializationContext, debugIndent: Int
+    ) {
+        val value = reader.read(obj)
+        if (value is ByteArray) {
```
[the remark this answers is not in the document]
distributedleetravis: I'm not sure I know! It's just what the framework did previously...

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/EvolutionSerializerFactory.kt
```diff
@@ -0,0 +1,165 @@
...
+                    else null
+                is RemoteTypeInformation.AnEnum ->
+                    if (local is LocalTypeInformation.AnEnum) remote.getEvolutionSerializer(local)
+                    else null
+                else -> null
+            }
+    
+    private fun RemoteTypeInformation.Composable.getEvolutionSerializer(
+            localTypeInformation: LocalTypeInformation.Composable): AMQPSerializer<Any>? {
+        // The no-op case: although the fingerprints don't match for some reason, we have compatible signatures.
```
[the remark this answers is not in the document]
distributedleetravis: Done.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/EvolutionSerializerFactory.kt
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    // Find the evolution constructor with the highest version number whose parameters are all assignable from the
+    // provided property types.
+    private fun findEvolverConstructor(constructors: List<EvolverConstructorInformation>,
+                                       properties: Map<String, RemotePropertyInformation>): EvolverConstructorInformation? {
+        val propertyTypes = properties.mapValues { (_, info) -> info.type.typeIdentifier.getLocalType(classLoader).asClass() }
+
+        // Evolver constructors are listed in ascending version order, so we just want the last that matches.
```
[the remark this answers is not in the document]
distributedleetravis: No, they are explicitly ordered by version number during `LocalTypeInformation` building - we're not reflecting at this point, but using information that has already been curated for us.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/EvolutionSerializerFactory.kt
```diff
@@ -0,0 +1,165 @@
...
+    private fun RemoteTypeInformation.Composable.validateEvolvability(localProperties: Map<PropertyName, LocalPropertyInformation>) {
+        val remotePropertyNames = properties.keys
+        val localPropertyNames = localProperties.keys
+        val deletedProperties = remotePropertyNames - localPropertyNames
+        val newProperties = localPropertyNames - remotePropertyNames
+
+        // Here is where we can exercise a veto on evolutions that remove properties.
+        if (deletedProperties.isNotEmpty() && mustPreserveDataWhenEvolving)
+            throw EvolutionSerializationException(this,
+                    "Property ${deletedProperties.first()} of remote ContractState type is not present in local type")
```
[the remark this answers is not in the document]
distributedleetravis: Done.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
```diff
@@ -0,0 +1,228 @@
...
+     */
+    fun get(actualClass: Class<*>, declaredType: Type): AMQPSerializer<Any>
+
+    /**
+     * Obtain an [AMQPSerializer] for the [declaredType].
+     */
+    fun get(declaredType: Type): AMQPSerializer<Any> {
+        val resolvedType = when(declaredType) {
+            is WildcardType -> if (declaredType.upperBounds.size == 1) declaredType.upperBounds[0]
+            else throw NotSerializableException("Cannot obtain upper bound for type $declaredType")
```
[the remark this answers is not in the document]
distributedleetravis: Fixed.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
distributedleetravis: It's thread-safe - all the caches are `ConcurrentHashMap`s, unless it's been instantiated inside the DJVM, and while it's possible that two threads may simultaneously try to create and cache the same serialiser, `getOrPut` is used to ensure that only one is cached, and the second-writer ends up with the serialiser created by the first. (We don't use `computeIfAbsent` in a lot of cases, because calls to read/populate the cache are often recursive and this would lock).

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
distributedleetravis: There's a test - `DeserializeNeedingCarpentry.arrayOfTypes` that blows up if we try this, although perhaps the test itself is trying to do something illegitimate. The original code did not make a `requireWhitelisted` call at this point - I'm guessing that at some point during array serialization there will be a call back into the factory to get a serialiser for the component type, and it will be checked then.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/MapSerializer.kt
distributedleetravis: Marked // FIXUP for now.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/RemoteSerializerFactory.kt
distributedleetravis: Reformatted to a simple logic expression.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactory.kt
distributedleetravis: None are resolved by this work (the goal was to match the existing framework warts-and-all, apart from the specific things we set out to fix). Probably a good idea to swipe the whole set of TODOs into a JIRA ticket for future consideration?

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/TypeNotationGenerator.kt
distributedleetravis: In a small number of cases, a default value is written into the TypeNotation for a property - this is a lookup for when this is the case, and what the value should be. Whether it's actually useful or not I couldn't say - I don't think it's used anywhere at all on the receiving end.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/SerializerFactory.kt
distributedleetravis: I've moved them to a new ticket: https://r3-cev.atlassian.net/browse/CORDA-2245

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Obtain an [AMQPSerializer] for an object of actual type [actualClass], and declared type [declaredType].
+     */
+    fun get(actualClass: Class<*>, declaredType: Type): AMQPSerializer<Any>
+
+    /**
+     * Obtain an [AMQPSerializer] for the [declaredType].
+     */
+    fun get(declaredType: Type): AMQPSerializer<Any> {
```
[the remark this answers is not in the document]
distributedleetravis: Moved as suggested.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/ComposableTypePropertySerializer.kt
```diff
@@ -0,0 +1,269 @@
...
+ */
+class DescribedTypeWriteStrategy(name: String,
+                                 private val propertyInformation: LocalPropertyInformation,
+                                 private val reader: PropertyReader,
+                                 private val serializerProvider: () -> AMQPSerializer<Any>) : PropertyWriteStrategy {
+
+    // Lazy to avoid getting into infinite loops when there are cycles.
+    private val serializer by lazy { serializerProvider() }
+
+    private val nameForDebug = "$name(${propertyInformation.type.typeIdentifier.prettyPrint(false)})"
```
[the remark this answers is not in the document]
distributedleetravis: Done.

tools/shell/src/test/kotlin/net/corda/tools/shell/InteractiveShellTest.kt
```diff
@@ -16,6 +16,7 @@ import net.corda.core.utilities.ProgressTracker
 import net.corda.node.services.identity.InMemoryIdentityService
 import net.corda.testing.core.TestIdentity
 import net.corda.testing.internal.DEV_ROOT_CA
+import org.junit.Ignore
```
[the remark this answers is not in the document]
distributedleetravis: Done.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/EvolutionSerializerFactory.kt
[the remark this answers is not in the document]
distributedleetravis: (I've added a couple of comments to `LocalTypeInformation` and `LocalTypeInformationBuilder` to clarify this)

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/MapSerializer.kt
mikehearn: Let's keep FIXUP to mean "things we'd like to change about the protocol" rather than just as a synonym for TODO. In this case, we can leave it as a TODO, file a JIRA ticket and maybe address it as a security improvement. I think it doesn't matter much right now because the receiving code will do a checkcast when retrieving the object from the map due to the generics, and it'll be caught then. But it'd be good to tighten this regardless so we can't deserialise arbitrary (annotated) objects unexpectedly.

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
mikehearn: Can we dig into this a bit? It makes me nervous. At the very least there should be a big fat comment explaining why this code that looks wrong is actually right.

mikehearn:
If we just investigate the array stuff a little bit, whilst it's on our mind, then I'm happy to approve this.

distributedleetravis:
OK: I think the outstanding areas of concern are about the serialisation of collections of heterogeneous things - `Object[]`, `List<?>` or `Map<?, ?>`.

The decision that seems to have been taken is that we decide on the safety of serialisation in such cases on an item-by-item basis. You want to serialize an `Object[]`? For each item, you'll need a serializer for whatever sort of thing that item is - its `actualClass` will be passed to the factory, which will look for an appropriate serialiser for that type. If it's whitelisted, good - if not, we'll refuse to serialise it. It's a similar story on deserialisation: we'll agree to deserialise a `Map<?, ?>`, but each value in the map is going to come with a type descriptor that we'll use to look for a serialiser for whatever kind of thing that is, and again we'll refuse to deserialise any _item_ whose declared type is not whitelisted.

The test that fails, if we check that the component type of an array is whitelisted, is one that serialises a collection of instances of a carpented class. The only declared type we _can_ have for these instances is `Object`. Whether the carpented class itself is whitelisted is up to the whitelist.

Now we could be a bit more pre-emptive about this: we could say that if the type is `Foo[]`, then it makes sense to check up-front whether `Foo` is whitelisted. We obviously don't whitelist `Object`, because that's _everything_, but if we're being more specific about the kinds of things that are in the array, it's unlikely that we'll say "`SubclassAOfFoo` and `SubclassBOfFoo` are whitelisted, but `Foo` itself is not". But this would be more of a "fail faster" mechanism than an "extra security" mechanism.

PR #4259, opened 2018-11-19
[CORDA-2185]: Added link to network-bootstrapper Maven page from the docs.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2185

Notes:

- Currently `https://mvnrepository.com/artifact/net.corda/corda-tools-network-bootstrapper` does not exist, because we renamed the module from `corda-network-bootstrapper` to `corda-tools-network-bootstrapper` (`https://mvnrepository.com/artifact/net.corda/corda-network-bootstrapper` works). It will work after the release though.


josecoll:
Just looking at https://www.corda.net/develop/resources.html 
I can see that all the resource links on that page point to our internal artifactory.
So I think the change should probably *not* be made here ... but on the Developer page (such that all links point to the latest released version of a tool).

PR #4260, opened 2018-11-19
CORDA-2114: SwapIdentitiesFlow is now inlined

shamsasari:
This is to fix the security issue whereby any counterparty is able to generate anonymous identities with a node at will without checks.


shamsasari:
There is an issue that CorDapps using confidential identities V3 will not work when the counterpart is running a V4 node. I'm not sure if that's an issue as this module is experimental.

mikehearn:
Hmm, technically we can do this because indeed the module is experimental. But it may cause pain for users who are using it. I wonder, can we renamespace it or something trivial to ensure the new and old can run alongside each other. Perhaps we should ask on corda-dev how many people are using this feature.

shamsasari:
The main issue here is the `SwapIdentitiesHandler` flow which is running on every node. Removing it would break apps using V3 confidential-identities. We could scan the installed CorDapps if they're using the old `SwapIdentitiesFlow` and not disable `SwapIdentitiesHandler` if any are found. But that would only make sense if we're preserving the wire format which would require leaving `IdentityWithSignature` as is and that we're happy with its signature format (@kchalkias doesn't seem to think so).

mikehearn:
What's the specific issue with the signature format? @kchalkias ?

shamsasari:
I think his concern is that anonymous identities should be tied down to a tx id so that it can't be used again in another transaction. But `SwapIdentitiesFlow` is used to build the tx so you can't pass in any id. That's my understanding, but best that @kchalkias confirm.

mikehearn:
Oh, that's disappointing, I remember specifically flagging the need for a challenge phase during the initial work on this feature several times.

I guess given that protocol issue, and the desire to better integrate the API, and we didn't commit to stability for this module specifically for these reasons, we should maybe just go ahead and change them. We may encounter some pain from people who used the module anyway despite its lack of stability (this is noted in the docs but would everyone have read them?). 

It'd be good to finish off and graduate this feature into the core API at some point.

shamsasari:
https://r3-cev.atlassian.net/browse/CORDA-2250 for adding the challenge in the protocol.

shamsasari:
Mentioned the incompatibilities in the changelog. Ready for another look.

shamsasari:
Yep. The question is whether we should try and get it in for v4 or leave for v5.

kchalkias:
IMO, because the naive approach will introduce extra network communication, we should better rethink the final design (not to degrade tps). We might consider an other deterministic way to produce challenges. So I'm leaning towards Corda5. We should also check for backwards compatibility consequences, so I'm leaning towards Corda 5 as this was never part of Corda 4 tasks and the impact of a misuse is not critical. Wdyt?

mikehearn:
Yes, it's not for 4.

shamsasari:
Can you approve @kchalkias, I had to fix some tests.

PR #4261, opened 2018-11-19
CORDA-2157 - Hash to Signature constraints migration V2

josecoll:
This PR supersedes the implementation given in https://github.com/corda/corda/pull/4237 which was breaching the attachments classloading no-overlap rule.

This implementation is based on the following rules:
1. Original unsigned contract jar (used for generating hash constrained states) must be signed and imported/resolved into the attachment store of the respective node. The signed version of the contract jar must be available for loading at node startup.
2. Java package namespace used in the Contract jar must be registered with the CZ network parameters (using the same key used to sign the contract jar).

Constraints checking to allow for Hash constrained states (unconsumed inputs in the nodes vault) to migrate to Signature constrained states (used as outputs in the creation of new states) performs additional classloader attachment checking to ensure that both signed and unsigned entries of the respective contract jars are bit-wise equal (using hashcode checking). 
The Attachments Classloader is thereby guaranteed to only load one attachment for any given contract classname, and thus not break the no-overlap rule.


gendal:
Wow... I'm so glad you, @tudor-malene and the team are working on this...  It's so mind-bendingly complex.  What's the documentation plan/strategy for this? How much needs to be explained in the docs? Let me know if you (and the team?) want to discuss your plans for what the docs should cover/how they should be structured... I am happy to play (well, I won't be playing) dumb.

josecoll:
@gendal given the complexity of

josecoll:
@tudor-malene as discussed on slack with @mikehearn, this implementation is performing a bit-wise comparison of attachment entries (when both unsigned and signed versions of same contract jar).
See what you think.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -67,14 +72,35 @@ class AttachmentsClassLoader(attachments: List<Attachment>, parent: ClassLoader
...
                         // Also convert to Unix path separators as all resource/class lookups will expect this.
-                        // If 2 entries have the same CRC, it means the same file is present in both attachments, so that is ok. TODO - Mike, wdyt?
+                        //
                         val path = entry.name.toLowerCase().replace('\\', '/')
                         if (shouldCheckForNoOverlap(path)) {
-                            if (path in classLoaderEntries) throw OverlappingAttachments(path)
-                            classLoaderEntries.add(path)
+                            if (path in classLoaderEntries.keys) {
+                                // If 2 entries have the same content hash, it means the same file is present in both attachments, so that is ok.
+                                val contentHash = readAttachment(attachment, path).sha256()
```
tudor-malene: Would be interesting to measure how slow this is approach is.

If it's slow it could be probably done in one go.
E.g: create a function:
`fun calculateEntriesHashes(jar:JarInputStream):Map<String (the path),SecureHash(hash of the file)>`
And then compare based on that.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -181,7 +180,7 @@ open class TransactionBuilder @JvmOverloads constructor(
         // For each contract, resolve the AutomaticPlaceholderConstraint, and select the attachment.
         val contractAttachmentsAndResolvedOutputStates: List<Pair<AttachmentId, List<TransactionState<ContractState>>?>> = allContracts.toSet()
                 .map { ctr ->
-                    handleContract(ctr, inputContractGroups[ctr], outputContractGroups[ctr], explicitAttachmentContractsMap[ctr], serializationContext, services)
+                    handleContract(ctr, inputContractGroups[ctr], outputContractGroups[ctr], explicitAttachmentContractsMap[ctr], services)
```
tudor-malene: Some logic is needed in `handleContract` as it currently forces the output to HashConstraint if the input was HashConstraint

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -155,14 +175,27 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
...
+                    log.debug { "Hash->Signature constraints migration: Ledger txn unsigned contractClassNames = $unsignedContractClassNames" }
+                    checkNotNull(attachmentStorage)
+                    // TODO: filter query by contract version
+                    val attachmentQueryCriteria = AttachmentsQueryCriteria(contractClassNamesCondition = Builder.equal(unsignedContractClassNames), isSignedCondition = Builder.equal(true))
+                    val signedAttachmentIds = attachmentStorage!!.queryAttachments(attachmentQueryCriteria)
+                    log.debug { "Hash->Signature constraints migration: Ledger txn signed attachmentIds = $signedAttachmentIds" }
+                    signedAttachmentIds
+                } else emptyList()
+
+        val resolvedAttachments = (attachments + extraAttachmentIds).lazyMapped { att, _ ->
```
tudor-malene: I don't think this logic belongs here.

I thought that the node building the transaction will attach both the signed and the un-signed jar to the transaction (in the TransactionBuilder) if it desires to make this transition.
That's why we decided to adjust the no-overlap rule.

@mikehearn , wdyt?

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -224,41 +226,54 @@ data class LedgerTransaction private constructor(
...
             if (!state.constraint.isSatisfiedBy(constraintAttachment)) {
                 throw TransactionVerificationException.ContractConstraintRejection(id, state.contract)
             }
         }
     }
 
-    private fun getUniqueContractAttachmentsByContract(): Map<ContractClassName, ContractAttachment> {
-        val result = mutableMapOf<ContractClassName, ContractAttachment>()
+    // TODO: revisit to include contract version information
+    private fun getUniqueContractAttachmentsByContract(): Map<Pair<ContractClassName,Boolean>, ContractAttachment> {
```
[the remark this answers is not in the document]
josecoll: The no-overlap rule is still enforced as it will check/verify that the contents of >1 attachment for a given contract are identical (or throw an `OverlappingAttachments` exception).

I will change the return type to be clearer as per above suggestion.

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
```diff
@@ -74,6 +82,39 @@ interface AttachmentConstraint {
             input is WhitelistedByZoneAttachmentConstraint && output is SignatureAttachmentConstraint ->
                 attachment.signers.isNotEmpty() && output.key.keys.containsAll(attachment.signers)
 
+            // Transition from Hash to Signature constraint (via CZ whitelisting) requires
+            // 1. Signer(s) of signature-constrained JAR is same as signer(s) of registered package namespace
+            // 2. Signature constraint key must be one of keys used to sign contract JAR.
+            input is HashAttachmentConstraint && output is SignatureAttachmentConstraint -> {
```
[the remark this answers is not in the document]
josecoll: There is no guarantee these checks have been made in the TransactionBuilder `handleContract` method which also calls this function.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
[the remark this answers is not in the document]
tudor-malene: Before we were double-enforcing it with this method.
But that's no longer what we're doing.

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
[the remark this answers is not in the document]
tudor-malene: Oh. I didn't say that the checks should be made in the builder. That's just building a valid transaction.

I'm saying that the checks are done in other places in `LedgerTransaction.verify`

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
tudor-malene: Anyway, there needs to be some logic added in that area - which is quite complex indeed.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
tudor-malene: Now that I think of it, that would have made a better attachmentID.
The MerkleTree root of all file hashes sorted by path.
That would have saved us so much trouble with unsigning.

tudor-malene: Now that I think of this even more, I think there's real potential to this, to make the rule more general.

- Let's say we calculate this `contentHash` for each attachment ( the function would  select the file that would matter - same as the no-overlap check, sort them, and get the root hash or the MTree)
- We can allow a transaction to contain multiple contract jars for the same contract as long as their `contentHash` is equal and they are signed by different parties ( this might come in handy later when we'll think about third-party signers).
- When passing to the AttachmentsClassloader, we only need to pass one of these jars, as we already compared their `contentHash`, so we don't have to make this check twice.

tudor-malene:
I was wondering if it would be a good idea to create a network parameter: `allowConstraintMigrations=true/false`
We could set it to true when we roll corda 4, and set it to false later on when everyone migrated their states.

This would allow us to eventually isolate or remove this extra complexity we're adding now for this purpose.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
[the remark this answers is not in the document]
josecoll: The No Overlap checking happens in the AttachmentsClassloader.
This function needs to collect all the relevant attachments (which in the case of this PR requires both the signed and unsigned versions for a given contract class)

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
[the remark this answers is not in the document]
josecoll: The TransactionBuilder `handleContract` method is running constraints checking verification logic.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -224,41 +226,54 @@ data class LedgerTransaction private constructor(
...
+
     /**
      * Verify that all contract constraints are passing before running any contract code.
      *
      * This check is running the [AttachmentConstraint.isSatisfiedBy] method for each corresponding [ContractAttachment].
      *
      * @throws TransactionVerificationException if the constraints fail to verify
      */
-    private fun verifyConstraints(internalTx: LedgerTransaction, contractAttachmentsByContract: Map<ContractClassName, ContractAttachment>) {
+    private fun verifyConstraints(internalTx: LedgerTransaction, contractAttachmentsByContract: Map<Pair<ContractClassName, Boolean>, ContractAttachment>) {
```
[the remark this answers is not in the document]
josecoll: Implemented as per above algo.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
josecoll: Good thinking. Maybe do this in a follow-up PR.

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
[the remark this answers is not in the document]
tudor-malene: The reason it's doing that is to fail early in case the wrong attachment was set by the user or the locally installed one does not match the constraints

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -217,34 +234,69 @@ private constructor(
...
+                // hash to to signature constraint migration logic:
+                // pass the unsigned attachment when verifying the constraint of the input state, and the signed attachment when verifying the constraint of the output state.
+                if (state.contract in hashToSignatureConstrainedContracts) {
+                    val unsignedAttachment = contractAttachmentsByContract[state.contract].unsigned
+                            ?: throw TransactionVerificationException.MissingAttachmentRejection(id, state.contract)
+                    val signedAttachment = contractAttachmentsByContract[state.contract].signed
+                            ?: throw TransactionVerificationException.MissingAttachmentRejection(id, state.contract)
+                    when {
+                        // use unsigned attachment if hash-constrained input state
+                        state.data in inputStates -> AttachmentWithContext(unsignedAttachment, state.contract, networkParameters!!)
```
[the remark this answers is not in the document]
josecoll: That check was already performed previously when collecting the:
`hashToSignatureConstrainedContracts`

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -213,11 +232,30 @@ open class TransactionBuilder @JvmOverloads constructor(
...
             explicitContractAttachment: AttachmentId?,
-            serializationContext: SerializationContext?,
             services: ServicesForResolution
-    ): Pair<AttachmentId, List<TransactionState<ContractState>>?> {
+    ): Pair<Set<AttachmentId>, List<TransactionState<ContractState>>?> {
         val inputsAndOutputs = (inputStates ?: emptyList()) + (outputStates ?: emptyList())
 
+        // identify if any input-output pairs are transitioning from hash to signature constraints.
+        val inputsHashConstraints = inputStates?.filter { it.constraint is HashAttachmentConstraint } ?: emptyList()
+        val outputsHashConstraints = outputStates?.filter { it.constraint is HashAttachmentConstraint } ?: emptyList()
```
tudor-malene: Is this correct? 
Should it be Signature?

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -213,11 +232,30 @@ open class TransactionBuilder @JvmOverloads constructor(
...
+        if (inputsHashConstraints.isNotEmpty() && outputsHashConstraints.isNotEmpty()) {
+            val attachmentQueryCriteria = AttachmentQueryCriteria.AttachmentsQueryCriteria(contractClassNamesCondition = Builder.equal(listOf(contractClassName)))
+            val attachmentIds = services.attachments.queryAttachments(attachmentQueryCriteria)
+            if (attachmentIds.isNotEmpty() && attachmentIds.size == 2)  {
```
tudor-malene: I don't quite understand how this works.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -234,11 +272,6 @@ open class TransactionBuilder @JvmOverloads constructor(
         require(hashAttachments.size <= 1) {
             "Transaction was built with $contractClassName states with multiple HashConstraints. This is illegal, because it makes it impossible to validate with a single version of the contract code."
         }
-        if (explicitContractAttachment != null && hashAttachments.singleOrNull() != null) {
```
tudor-malene: This check is still useful as you already handled the migration above and exited the function.

core/src/main/kotlin/net/corda/core/contracts/ContractAttachment.kt
```diff
@@ -22,7 +23,14 @@ class ContractAttachment @JvmOverloads constructor(
 
     val allContracts: Set<ContractClassName> get() = additionalContracts + contract
 
+    val isSigned: Boolean get() = signerKeys.isNotEmpty()
+
+    /**
+     * Contract version
+     */
+    val version: String = try { attachment.openAsJAR().manifest?.mainAttributes?.getValue(Attributes.Name.IMPLEMENTATION_VERSION) ?: "1.0" } catch (e: Exception) { "1.0" }
```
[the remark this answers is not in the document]
josecoll: Good suggestion!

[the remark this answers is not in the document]
tudor-malene: I think I suggested the same to Szymon.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
josecoll: Yes - we need to identify existing Hash constrained output states and switch them over.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
josecoll: Will simplify - this query just needs to obtain the "signed" attachment.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
josecoll: reverted

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
tudor-malene: As discussed, at this point the constraint on output states is the 'AutomaticPlaceholderConstraint'. 
This function is responsible for setting constraints on output states.

If the constraint on the output states was manually set to HashConstraint, you should leave it alone, as it looks like a clear indication that the user wanted it that way.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -45,38 +52,72 @@ class AttachmentsClassLoader(attachments: List<Attachment>, parent: ClassLoader
...
+                // into the storage service. So we don't need to repeat it here.
+                //
+                // We forbid files that differ only in case, or path separator to avoid issues for Windows/Mac developers where the
+                // filesystem tries to be case insensitive. This may break developers who attempt to use ProGuard.
+                //
+                // Also convert to Unix path separators as all resource/class lookups will expect this.
+                //
+                val filepath = entry.name.toLowerCase().replace('\\', '/')
+                if (shouldCheckForNoOverlap(filepath, targetPlatformVersion)) {
+                    val contentHash = readAttachment(attachment, filepath).sha256()
```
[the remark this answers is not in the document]
josecoll: Re-coded to use the hash of each jar entry (as opposed to calculating the hash of the contents of each entry as a file attached byte-stream).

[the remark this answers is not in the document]
josecoll: Not sure what you mean by "Go through each entry and calculate the hash".
The code needs to calculate the hash of the contents to disambiguate (which is what it does currently).

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
```diff
@@ -275,12 +277,24 @@ sealed class AttachmentQueryCriteria : GenericQueryCriteria<AttachmentQueryCrite
     /**
      * AttachmentsQueryCriteria:
      */
-    data class AttachmentsQueryCriteria @JvmOverloads constructor (val uploaderCondition: ColumnPredicate<String>? = null,
-                                                                   val filenameCondition: ColumnPredicate<String>? = null,
-                                                                   val uploadDateCondition: ColumnPredicate<Instant>? = null) : AttachmentQueryCriteria() {
+    data class AttachmentsQueryCriteria @JvmOverloads constructor(val uploaderCondition: ColumnPredicate<String>? = null,
```
shamsasari: Is API is no longer useful for Java users. If they want to specify a bunch of conditions at the end, then they're forced to specify the conditions before it. That's what `@JvmOverloads` does. I think we need to introduce wither methods for each condition property, like we do for `DriverParameters`.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -45,38 +52,72 @@ class AttachmentsClassLoader(attachments: List<Attachment>, parent: ClassLoader
...
-                        if (shouldCheckForNoOverlap(path, targetPlatformVersion)) {
-                            if (path in classLoaderEntries) throw TransactionVerificationException.OverlappingAttachmentsException(path)
-                            classLoaderEntries.add(path)
+                    val attachmentEntries = calculateEntriesHashes(attachment, jar)
+                    if (attachmentEntries.isNotEmpty()) {
+                        val contentHashMT = MerkleTree.getMerkleTree(attachmentEntries.toSortedMap().map{ it.value }).hash
+                        if (attachmentContentMTHashes.contains(contentHashMT)) {
+                            log.debug { "Duplicate entry: ${attachment.id} has same content hash $contentHashMT as previous attachment" }
+                        }
+                        else {
```
shamsasari: Codestyle is `} else {`

docs/source/changelog.rst
```diff
@@ -7,6 +7,11 @@ release, see :doc:`upgrade-notes`.
 Unreleased
 ----------
 
+* Automatic Constraints propagation for hash-constrained states to signature-constrained states.
+  This enables signed CorDapps that have registered their contract jar java package namespaces with the
+  network parameters to consume pre-Corda 4 hash constrained states created by unsigned CorDapps.
+  Requires usage of two new Corda 4 features: signature constraints and java package namespace registration.
```
shamsasari: It's not clear, as an end-user, what this is or why I'd want to take benefit from it. Also we do refer to a page that provides more step-by-step instructions.

node/src/main/kotlin/net/corda/node/serialization/kryo/DefaultKryoCustomizer.kt
```diff
@@ -222,6 +223,7 @@ object DefaultKryoCustomizer {
                 val additionalContracts = kryo.readClassAndObject(input) as Set<ContractClassName>
                 val uploader = input.readString()
                 val signers = kryo.readClassAndObject(input) as List<PublicKey>
+                val version = input.readString()
```
shamsasari: This has changed the checkpoint format. I assume the node doesn't upgrade if there are active flows?

testing/node-driver/src/main/kotlin/net/corda/testing/node/TestCordapp.kt
```diff
@@ -31,6 +32,9 @@ interface TestCordapp {
     /** Returns the set of package names scanned for this test CorDapp. */
     val packages: Set<String>
 
+    /** Returns whether the CorDapp should be jar signed. */
+    val signJar: Boolean
```
shamsasari: Looking at how this is all implemented (use of hardcoded aliases and pwds in`TestCordappDirectories.getJarDirectory`), this does not look like a public testing API. Can we move this to `TestCordappImpl`.

testing/test-utils/src/main/kotlin/net/corda/testing/core/ContractJarTestUtils.kt
```diff
@@ -0,0 +1,108 @@
...
+import java.security.PublicKey
+import java.util.jar.Attributes
+import java.util.jar.JarEntry
+import java.util.jar.JarOutputStream
+import javax.tools.JavaFileObject
+import javax.tools.SimpleJavaFileObject
+import javax.tools.StandardLocation
+import javax.tools.ToolProvider
+
+object ContractJarTestUtils {
```
shamsasari: This doesn't look like it should be public API.

testing/test-utils/src/main/kotlin/net/corda/testing/core/JarSignatureTestUtils.kt
```diff
@@ -1,16 +1,33 @@
...
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.security.PublicKey
+import java.util.jar.Attributes
 import java.util.jar.JarInputStream
+import java.util.jar.JarOutputStream
+import java.util.jar.Manifest
 import kotlin.test.assertEquals
 
+/**
```
shamsasari: ditto for this file as well.

.ci/api-current.txt
```diff
@@ -3465,22 +3465,45 @@ public static final class net.corda.core.node.services.vault.AttachmentQueryCrit
   public <init>(net.corda.core.node.services.vault.ColumnPredicate<String>)
   public <init>(net.corda.core.node.services.vault.ColumnPredicate<String>, net.corda.core.node.services.vault.ColumnPredicate<String>)
   public <init>(net.corda.core.node.services.vault.ColumnPredicate<String>, net.corda.core.node.services.vault.ColumnPredicate<String>, net.corda.core.node.services.vault.ColumnPredicate<java.time.Instant>)
+  public <init>(net.corda.core.node.services.vault.ColumnPredicate<String>, net.corda.core.node.services.vault.ColumnPredicate<String>, net.corda.core.node.services.vault.ColumnPredicate<java.time.Instant>, net.corda.core.node.services.vault.ColumnPredicate<java.util.List<String>>)
```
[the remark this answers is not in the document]
josecoll: No breaking changes - just additional attachment query criteria attributes.

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
```diff
@@ -32,20 +34,27 @@ interface AttachmentConstraint {
...
      *  * You can transition from the [WhitelistedByZoneAttachmentConstraint] to the [SignatureAttachmentConstraint] only if all signers of the JAR are required to sign in the future.
+     *  * You can transition from a [HashAttachmentConstraint] to a [SignatureAttachmentConstraint] when the following conditions are met:
+     *  *  1. Jar contents (per entry, by hashcode) of both original (unsigned) and signed contract jars are identical
+     *  *     Note: this step is enforced in the [AttachmentsClassLoader] no overlap rule checking.
+     *  *  2. Java package namespace of signed contract jar is registered in the CZ network map with same public keys (as used to sign contract jar)
      *
      * TODO - SignatureConstraint third party signers.
      */
-    fun canBeTransitionedFrom(input: AttachmentConstraint, attachment: ContractAttachment): Boolean {
+    fun canBeTransitionedFrom(input: AttachmentConstraint, attachment: AttachmentWithContext): Boolean {
```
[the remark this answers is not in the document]
josecoll: Isn't the @DoNotImplement annotation on the interface sufficient?
Note that none of the functions declared in this interface should be used by the public.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -185,7 +186,15 @@ data class NetworkParameters(
     /**
      * Returns the public key of the package owner of the [contractClassName], or null if not owned.
      */
-    fun getOwnerOf(contractClassName: String): PublicKey? = this.packageOwnership.filterKeys { packageName -> owns(packageName, contractClassName) }.values.singleOrNull()
+    fun getOwnerOf(contractClassName: ContractClassName): PublicKey? = this.packageOwnership.filterKeys { packageName -> owns(packageName, contractClassName) }.values.singleOrNull()
+
+    /**
+     * Returns the public key of the package owner if any of [contractClassName] match, or null if not owned.
+     */
+    fun getOwnerOf(contractClassNames: Set<ContractClassName>): PublicKey? {
```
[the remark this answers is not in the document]
josecoll: Note there are other public functions in this class that probably should also be internal (eg. `canAutoAccept)`

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
josecoll: We could do but then the constructor permutations pollute the api.
We don't currently do this for other VaultQueryCriteria specifications.

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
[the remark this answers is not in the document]
shamsasari: `@DoNotImplement` does not prevent a user from calling its method, just they can't extend it for their own implementation. If it's not meant to be used externally then let's make it internal. It can be made into an extension method residing in `n.c.core.internal`.

docs/source/changelog.rst
josecoll: This is just a changelog entry. See the `api-contract-constraints` for detail.
Let me rephrase to make it clearer.

node/src/main/kotlin/net/corda/node/serialization/kryo/DefaultKryoCustomizer.kt
josecoll: Interesting observation.
Will need to add a warning to state this specifically if it is indeed a problem (will add a scenario test to the CTS to cover this use case).

testing/node-driver/src/main/kotlin/net/corda/testing/node/TestCordapp.kt
josecoll: Probably better to do this outside of this PR (as it touches on a lot of other code).

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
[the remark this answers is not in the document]
shamsasari: Yes I noticed that. I have a branch where I've made them internal. I will raise a PR once we cut.

testing/test-utils/src/main/kotlin/net/corda/testing/core/ContractJarTestUtils.kt
josecoll: It may be useful for upgrade testing of 3rd party cordapps.

testing/test-utils/src/main/kotlin/net/corda/testing/core/JarSignatureTestUtils.kt
josecoll: as per above comment.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
shamsasari: The additional c'tor params are fine, but I think we will need to supplement them with wither methods. Do we have Java samples or tests using this/these APIs so we have an idea of what they're like?

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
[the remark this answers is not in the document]
josecoll: OK - will change.

node/src/main/kotlin/net/corda/node/serialization/kryo/DefaultKryoCustomizer.kt
shamsasari: I can't remember how flow draining is used for updates. Does the node enforce flow draining before it can be updated? Perhaps @sollecitom or someone from his team knows.

testing/test-utils/src/main/kotlin/net/corda/testing/core/ContractJarTestUtils.kt
shamsasari: Then this needs to be polished up with docs on how they are to be used, as it's not clear otherwise. Also we'd want to not use `Pair` as it makes the API (especially in Java) ugly.

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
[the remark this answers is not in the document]
shamsasari: Then `@CordaInternal` it is, though I think moving it into the internal package is a cleaner solution.

docs/source/changelog.rst
```diff
@@ -8,9 +8,8 @@ Unreleased
 ----------
 
 * Automatic Constraints propagation for hash-constrained states to signature-constrained states.
-  This enables signed CorDapps that have registered their contract jar java package namespaces with the
-  network parameters to consume pre-Corda 4 hash constrained states created by unsigned CorDapps.
-  Requires usage of two new Corda 4 features: signature constraints and java package namespace registration.
+  This allows Corda 4 signed CorDapps using signature constraints to consume existing hash constrained states generated
+  by unsigned CorDapps in previous versions of Corda.
```
shamsasari: Refer to the upgrade rst for further details.

.ci/api-current.txt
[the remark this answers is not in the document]
anthonykeenan: Yes, there's a problem with the @Null/@NotNull notations being matched with the wrong method by the diff, causing the API check to faill. We really need a delimiter around each method record.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
josecoll: Added wither methods and Java Unit tests.

core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -185,7 +186,15 @@ data class NetworkParameters(
     /**
      * Returns the public key of the package owner of the [contractClassName], or null if not owned.
      */
-    fun getOwnerOf(contractClassName: String): PublicKey? = this.packageOwnership.filterKeys { packageName -> owns(packageName, contractClassName) }.values.singleOrNull()
+    fun getOwnerOf(contractClassName: ContractClassName): PublicKey? = this.packageOwnership.filterKeys { packageName -> owns(packageName, contractClassName) }.values.singleOrNull()
```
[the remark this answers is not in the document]
josecoll: Had to move `NetworkParametersTests` as they were in the wrong module and lost visibility of the above when made internal.

docs/source/changelog.rst
josecoll: Changelog is not the place to add these references.

josecoll: We do need to add some upgrade notes for the general topic of Signature Constraints and Constraints propagation. I will add a JIRA story for this.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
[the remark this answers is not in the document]
josecoll: I am going to revert these premature optimisations, and perform some basic checking to avoid the expensive no duplicate checking (when not necessary):
1. single attachment jar passed into the Classloader
2. attachment jars with non-overlapping contract classes

testing/test-utils/src/main/kotlin/net/corda/testing/core/JarSignatureTestUtils.kt
josecoll: OK - I've moved these utils to internal (pending clean-up and documentation before exposing them as public).

josecoll:
@shamsasari everything addressed. Back to you!

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
```diff
@@ -262,12 +262,12 @@ data class ContractUpgradeLedgerTransaction(
     }
 
     private fun verifyConstraints() {
+        checkNotNull(networkParameters)
```
shamsasari: Provide an error message.

docs/source/api-contract-constraints.rst
```diff
@@ -29,6 +29,11 @@ attachment JARs may not provide the same file path. If they do, the transaction
 state specifies both a constraint over attachments *and* a Contract class name to use, the specified class must appear
 in only one attachment.
 
+.. note:: With the introduction of signature constraints in Corda 4, a new Attachments Classloader will verify that
```
shamsasari: no need to capitalise "attachments classloader"

core/src/main/kotlin/net/corda/core/transactions/ContractUpgradeTransactions.kt
josecoll: Removed redundant `checkNotNull(networkParameters)` checks now that these are always passed into the main constructor.

vkolomeyko:
This PR is likely to be breaking Windows build:
https://ci-master.corda.r3cev.com/viewLog.html?buildId=129930&buildTypeId=Corda_BuildWindowsTest

core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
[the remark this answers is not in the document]
mikehearn: Please do move it. As much internal stuff as possible should be in an internal package for Jigsaw reasons.

PR #4263, opened 2018-11-20
[CORDA-2211]: Added error for older bash versions when attempting to install shell extensions.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2211

### DIFF
--- a/tools/cliutils/src/main/kotlin/net/corda/cliutils/InstallShellExtensionsParser.kt
+++ b/tools/cliutils/src/main/kotlin/net/corda/cliutils/InstallShellExtensionsParser.kt
@@ -87,17 +87,21 @@ private class ShellExtensionsGenerator(val parent: CordaCliWrapper) {
     }
 
     fun installShellExtensions(): Int {
+        // Get jar location and generate alias command
+        val command = "alias ${parent.alias}='java -jar \"${jarLocation.toStringWithDeWindowsfication()}\"'"
+        var generateAutoCompleteFile = true
         if (SystemUtils.IS_OS_UNIX && installedShell() == ShellType.BASH) {
             val semanticParts = declaredBashVersion().split(".")
             semanticParts.firstOrNull()?.toIntOrNull()?.let { major ->
                 if (major < minSupportedBashVersion) {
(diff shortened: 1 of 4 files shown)


tools/cliutils/src/main/kotlin/net/corda/cliutils/InstallShellExtensionsParser.kt
```diff
@@ -80,7 +86,16 @@ private class ShellExtensionsGenerator(val parent: CordaCliWrapper) {
...
+            val semanticParts = declaredBashVersion().split(".")
+            semanticParts.firstOrNull()?.toIntOrNull()?.let { major ->
+                if (major < minSupportedBashVersion) {
+                    parent.printlnErr("Cannot install shell extension for Bash major version earlier than $minSupportedBashVersion. Please upgrade your Bash version.")
+                    return ExitCodes.FAILURE
+                }
+            }
+        }
         // Get jar location and generate alias command
         val command = "alias ${parent.alias}='java -jar \"${jarLocation.toStringWithDeWindowsfication()}\"'"
```
anthonykeenan: The aliasing should still work on older versions of bash though shouldn't it?
I think we should still add the alias and just print a warning saying that the alias is available but not auto-complete, and return with success?

tools/cliutils/src/main/kotlin/net/corda/cliutils/InstallShellExtensionsParser.kt
```diff
@@ -101,9 +119,34 @@ private class ShellExtensionsGenerator(val parent: CordaCliWrapper) {
         zshSettingsFile.addIfNotExists(completionFileCommand)
         zshSettingsFile.updateAndBackupIfNecessary()
 
-        println("Installation complete, ${parent.alias} is available in bash with autocompletion. ")
+        if (generateAutoCompleteFile) {
+            println("Installation complete, ${parent.alias} is available in bash with autocompletion.")
+        } else {
+            println("Installation complete, ${parent.alias} is available in bash, but autocompletion was not installed because of an old version of Bash.")
```
anthonykeenan: If bash has a capital B it needs to have a capital B in all the messages, I'm not sure which is right tbh, but as long as it's consistent that's fine.

anthonykeenan:
Looks good 👍

PR #4264, opened 2018-11-20
[CORDA-2243] Notary client should sign over NotarisationRequest serialised with object referencing turned off

thschroeter:
To make sure the notary service constructs the same serialised bytes as the client.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/notary/NotaryUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/notary/NotaryUtils.kt
@@ -41,7 +41,7 @@ fun NotarisationResponse.validateSignatures(txId: SecureHash, notary: Party) {
 }
 
 /** Creates a signature over the notarisation request using the legal identity key. */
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/net/corda/core/internal/notary/NotaryUtils.kt
```diff
@@ -41,7 +41,7 @@ fun NotarisationResponse.validateSignatures(txId: SecureHash, notary: Party) {
 }
 
 /** Creates a signature over the notarisation request using the legal identity key. */
-fun NotarisationRequest.generateSignature(serviceHub: ServiceHub): NotarisationRequestSignature {
+fun NotarisationRequest.generateRequestSignature(serviceHub: ServiceHub): NotarisationRequestSignature {
```
adagys: was this rename intended? Seems as if intellij was trying to be too clever!

thschroeter: right, i'll undo it

Issue #3941, opened 2018-09-14
FR: initialiseSchema property couples corda node and cordapp ddl handling


remmeier:
Corda itself makes use of Liquibase. The cordapps make use of Hibernate DDL generation. Setting the database.initialiseSchema schema property only allows to enable/disable both together. There should rather be two properties that allow to configure the two separately, allowing to handle DDLs of cordapps outside of corda. For example:

- initialiseNodeSchema 
- initialiseCordappSchema

whereas initialiseSchema can continue two set both.

joeldudleyr3:
@remmeier Would you be able to PR your fix back into the codebase? I've spoken to the DB guys and they're happy to implement your fix :)

remmeier:
yes, sure, I should have time this week. After that two weeks on holidays.

PR #4277, opened 2018-11-22
initialiseSchema property couples corda node and cordapp

remmeier:
New property database.initialiseAppSchema is introduced that allows to turn-off Hibernate DDL generation while leaving Node-specific Liquibase handling enabled.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/HibernateConfiguration.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/HibernateConfiguration.kt
@@ -73,12 +73,22 @@ class HibernateConfiguration(
         logger.info("Creating session factory for schemas: $schemas")
         val serviceRegistry = BootstrapServiceRegistryBuilder().build()
         val metadataSources = MetadataSources(serviceRegistry)
+
+        val hbm2dll: String =
+        if(databaseConfig.initialiseSchema && databaseConfig.initialiseAppSchema == SchemaInitializationType.UPDATE) {
+            "update"
+        } else if((!databaseConfig.initialiseSchema && databaseConfig.initialiseAppSchema == SchemaInitializationType.UPDATE)
+                || databaseConfig.initialiseAppSchema == SchemaInitializationType.VALIDATE) {
+            "validate"
+        } else {
+            "none"
+        }
+
         // We set a connection provider as the auto schema generation requires it.  The auto schema generation will not
         // necessarily remain and would likely be replaced by something like Liquibase.  For now it is very convenient though.
-        // TODO: replace auto schema generation as it isn't intended for production use, according to Hibernate docs.
         val config = Configuration(metadataSources).setProperty("hibernate.connection.provider_class", NodeDatabaseConnectionProvider::class.java.name)
-                .setProperty("hibernate.hbm2ddl.auto", if (databaseConfig.initialiseSchema) "update" else "validate")
                 .setProperty("hibernate.format_sql", "true")
+                .setProperty("hibernate.hbm2ddl.auto", hbm2dll)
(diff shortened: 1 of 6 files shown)


joeldudleyr3:
@remmeier Hi Remo!

We generally try and minimise config options because it makes things more complicated for node operators. So it would be great to hear what the use-case is for adding this option? What problem are you trying to avoid?

remmeier:
Hi @joeldudleyr3, the DB guys already accepted the proposal, see https://github.com/corda/corda/issues/3941. The general issue is that the Hibernate DDL generator is not really usable for many things. There was also a TODO in the code base. But Corda should continue to migrate its own schema elements and not touch the cordapp ones. or for people runnign Corda Enterprise Corda can continue to take care of it (we likely end up in a mix of Corda OSS/Enterprise). So I think those two things should be kept separate. Alternatively one could for example also keep a single property and make an enum out it (with true and false as possible values to keep backward compatiblity).

remmeier:
maybe the enum would be better, potentially it could also allow to enable/disable the validation flag of hibernate (in our case we actually dont' want that either...)

szymonsztuka:
@remmeier Let's go with:
```
initialiseSchema = true/false
initialiseCordappSchemas = true/false
```

If that's ok?

The other options we could have gone for were:
* Disjoint properties (but it's not backward compatible):
```
initialiseNodeSchema = true/false
initialiseCordappSchemas = true/false
```
* An enum, but it  would need to have still 4  values (below not proposed names, but cases):
```
initialiseSchema = true_for_all/false_for_all/true_for_node_olny/true_for_app_olny
```

node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/HibernateConfiguration.kt
```diff
@@ -75,9 +75,8 @@ class HibernateConfiguration(
         val metadataSources = MetadataSources(serviceRegistry)
         // We set a connection provider as the auto schema generation requires it.  The auto schema generation will not
         // necessarily remain and would likely be replaced by something like Liquibase.  For now it is very convenient though.
-        // TODO: replace auto schema generation as it isn't intended for production use, according to Hibernate docs.
         val config = Configuration(metadataSources).setProperty("hibernate.connection.provider_class", NodeDatabaseConnectionProvider::class.java.name)
-                .setProperty("hibernate.hbm2ddl.auto", if (databaseConfig.initialiseSchema) "update" else "validate")
+                .setProperty("hibernate.hbm2ddl.auto", if (databaseConfig.initialiseSchema && databaseConfig.initialiseAppSchema) "update" else "none")
```
szymonsztuka: Why change from `validate` to `none`, we would still want to validate?

joeldudleyr3:
@remmeier We've discussed it and decided to leave it in "validate" mode for now. We can discuss whether to allow disabling of validation in a separate PR.

If you're happy to switch it back and rename to `initialiseCordappSchemas`, we can go ahead and merge.

remmeier:
thx, sorry had a meeting day. 

well the validation will at least not work for us.

szymonsztuka:
@remmeier So the best would be to introduce enum, with `update`, `validate`, `none`, the reason to keep `validate` is that it's still useful setup (updating core database schema and just validating cordapps database schemas), it introduces some inconsistency (like `initialiseSchema` is a boolean but `initialiseCordappSchemas` is enum) but we could clean this up later.

remmeier:
give me a short moment, I will look a bit more closely into Hibernate what is going on in their.

But in general it is desirable to disable it. Not just for the sake of validation, but also in terms of startup. We have a micro service architecture running in Kubernetes. And so the persistence.xml is used by both the cordapps and the microservices. What we don't want is for the services to access the database on startup and crash the application in case of the service being faster than the database. we rather rely on health checks for this kind of things.

remmeier:
I updated the PR, the

remmeier:
thx for alll the input!

PR #4278, opened 2018-11-22
CORDA-2199 NetworkParameters certificate role

jamesbr3:
This change introduces a new NetworkParameters certificate role (CertRole.NETWORK_PARAMETERS). This will allow the network parameters to be signed with a different cert in the future (currently it is being signed by the NetworkMap cert).

The change is backward-compatible - it will still accept network-parameters to be signed by the networkmap cert, but will also accept a (yet to exist) network-params cert when that is created.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/CertRole.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/CertRole.kt
@@ -46,7 +46,10 @@ enum class CertRole(val validParents: NonEmptySet<CertRole?>, val isIdentity: Bo
     LEGAL_IDENTITY(NonEmptySet.of(DOORMAN_CA, NODE_CA), true, true),
 
     /** Confidential (limited visibility) identity of a legal entity. */
-    CONFIDENTIAL_LEGAL_IDENTITY(NonEmptySet.of(LEGAL_IDENTITY), true, false);
+    CONFIDENTIAL_LEGAL_IDENTITY(NonEmptySet.of(LEGAL_IDENTITY), true, false),
+
+    /** Signing certificate for Network Parameters. */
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/net/corda/core/internal/CertRole.kt
```diff
@@ -46,7 +46,10 @@ enum class CertRole(val validParents: NonEmptySet<CertRole?>, val isIdentity: Bo
     LEGAL_IDENTITY(NonEmptySet.of(DOORMAN_CA, NODE_CA), true, true),
 
     /** Confidential (limited visibility) identity of a legal entity. */
-    CONFIDENTIAL_LEGAL_IDENTITY(NonEmptySet.of(LEGAL_IDENTITY), true, false);
+    CONFIDENTIAL_LEGAL_IDENTITY(NonEmptySet.of(LEGAL_IDENTITY), true, false),
+
+    /** Signing certificate for Network Parameters. */
+    NETWORK_PARAMETERS(NonEmptySet.of(null), false, false);
```
kchalkias: Just to clarify, `NETWORK_PARAMETERS` is signed by the root (or intermediate) directly, similarly to Doorman and NetworkMap, right?

jamesbr3: yes that's right

jamesbr3: It would sit next to the NetworkMap cert in the hierarchy

mikehearn:
That looks ok to me but I haven't reviewed it deeply. Will bow out and let Shams or Kostas take the lead.

PR #4283, opened 2018-11-22
[CORDA-2249]: Specified unique applicationId for Capsule manifest.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2249

As per the section "The Capsule ID" in http://www.capsule.io/user-guide/#building-capsule, Capsule composite ID must be unique for each application, and it's composed by `applicationId` or `<group>.<artifact>` if unspecified, plus `_<version>`.

Hence, the fix is to specify `applicationId` inside Capsule's manifest for Corda Enterprise, and for it to be different from the one in Open Source. Even if for Open Source `<group>.<artifact>` is good as an implicit `applicationId`, I put it there explicitly to avoid confusion.

### DIFF
--- a/node/capsule/build.gradle
+++ b/node/capsule/build.gradle
@@ -48,7 +48,7 @@ task buildCordaJAR(type: FatCapsule, dependsOn: project(':node').tasks.jar) {
 
     capsuleManifest {
         applicationVersion = corda_release_version
-
+        applicationId = "net.corda.node.Corda"
         // See experimental/quasar-hook/README.md for how to generate.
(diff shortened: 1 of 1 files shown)


node/capsule/build.gradle
```diff
@@ -48,7 +48,7 @@ task buildCordaJAR(type: FatCapsule, dependsOn: project(':node').tasks.jar) {
 
     capsuleManifest {
         applicationVersion = corda_release_version
-
+        applicationId = "net.corda.node.Corda"
```
rick-r3: Will enterprise differ then?

sollecitom: Yes :)

rick-r3: Just saw ENT PR.  The answer is "Yes".

PR #4289, opened 2018-11-22
CORDA-2150 Non downgrade contract version - test infrastructure 3

szymonsztuka:
`internalDriver` can sign CorDapp JAR, disabled by default.


szymonsztuka:
Moved new flag to `internalDriver` for now.

PR #4290, opened 2018-11-23
CORDA-1825 user-friendly error when serializing non-composable types

distributedleetravis:
When an attempt is made to serialize a type with no public constructor (e.g. `java.util.Optional`), and no custom serializer, the following message is now returned:

```
Trying to build an object serializer for Optional (erased), but it is not constructible from its public properties, and so requires a custom serialiser.
```

### DIFF
--- a/serialization/src/test/kotlin/net/corda/serialization/internal/amqp/DeserializeSimpleTypesTests.kt
+++ b/serialization/src/test/kotlin/net/corda/serialization/internal/amqp/DeserializeSimpleTypesTests.kt
@@ -3,6 +3,7 @@ package net.corda.serialization.internal.amqp
 import net.corda.serialization.internal.amqp.testutils.*
 import org.junit.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFails
 
 // Prior to certain fixes being made within the [PropertySerializaer] classes these simple
 // deserialization operations would've blown up with type mismatch errors where the deserlized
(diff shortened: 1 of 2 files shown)


serialization/src/test/kotlin/net/corda/serialization/internal/amqp/DeserializeSimpleTypesTests.kt
```diff
@@ -527,5 +528,13 @@ class DeserializeSimpleTypesTests {
         DeserializationInput(sf2).deserialize(serializedA.obj)
     }
 
+    @Test
+    fun classHasNoPublicConstructor() {
+        assertFails("Trying to build an object serializer for Optional (erased), " +
+                "but it is not constructible from its public properties, and so requires a custom serialiser.") {
+            TestSerializationOutput(VERBOSE, sf1).serializeAndReturnSchema(java.util.Optional.of(1))
```
shamsasari: We have `OptionalSerializer` registered so we should never have this issue with `Optional`?

distributedleetravis: In the test case, the factory doesn't have that serialiser (or any custom serialiser) registered. `Optional` is a convenient choice for an example of the kind of class that won't serialise without a custom serialiser (it's the one used in the original ticket).

distributedleetravis: Added tests showing `@CordaSerializable` absence being picked up when the serializer factory is using a whitelist (which it always will be, anywhere outside of a test class).

PR #4292, opened 2018-11-23
Update to corda-gradle-plugins v4.0.36

szymonsztuka:
New version contains new development key for JAR signing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,7 +87,7 @@ buildscript {
         jcenter()
         // This repository is needed for Dokka until 0.9.16 is released.
         maven {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -87,7 +87,7 @@ buildscript {
         jcenter()
         // This repository is needed for Dokka until 0.9.16 is released.
```
chrisr3: You can remove this comment too - it only applies to the `kotlin-eap` repository.

szymonsztuka: Fixed

PR #4293, opened 2018-11-26
CORDA-2115: Notary whitelist verification changes

adagys:
- For regular and contract upgrade transactions: check that the notary is in the network parameter whitelist
- For notary change transactions: check the the new notary is in the network parameter whitelist. This enabled support for network merging: the old notary doesn't have to be in the current network's notary whitelist for re-pointing old states to another notary.

These checks are done during transaction construction/verification and also by the non-validating notary.

(Raising a PR against Kasia's branch but will rebase against master later).

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -2278,8 +2278,6 @@ public static class net.corda.core.flows.NotaryFlow$Client extends net.corda.cor
   @NotNull
   public java.util.List<net.corda.core.crypto.TransactionSignature> call()
   @NotNull
-  protected final net.corda.core.identity.Party checkTransaction()
-  @NotNull
   public net.corda.core.utilities.ProgressTracker getProgressTracker()
   @Suspendable
   @NotNull
@@ -6383,7 +6381,7 @@ public class net.corda.testing.node.MockServices extends java.lang.Object implem
   @NotNull
   public net.corda.core.node.services.NetworkMapCache getNetworkMapCache()
   @NotNull
-  public final net.corda.core.node.NetworkParameters getNetworkParameters()
+  public net.corda.core.node.NetworkParameters getNetworkParameters()
   @NotNull
   protected final net.corda.core.node.ServicesForResolution getServicesForResolution()
   @NotNull
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -2,7 +2,6 @@ package net.corda.core.flows
 
 import co.paralleluniverse.fibers.Suspendable
 import net.corda.core.DoNotImplement
-import net.corda.core.contracts.ComponentGroupEnum
 import net.corda.core.contracts.StateRef
 import net.corda.core.contracts.TimeWindow
 import net.corda.core.crypto.SecureHash
@@ -26,6 +25,8 @@ class NotaryFlow {
      *
      * In case of a single-node or Raft notary, the flow will return a single signature. For the BFT notary multiple
      * signatures will be returned – one from each replica that accepted the input state commit.
+     *
+     * The transaction to be notarised, [stx], should be fully verified before calling this flow.
      *
      * @throws NotaryException in case the any of the inputs to the transaction have been consumed
      *                         by another transaction or the time-window is invalid or
@@ -51,8 +52,8 @@ class NotaryFlow {
         override fun call(): List<TransactionSignature> {
             stx.pushToLoggingContext()
             val notaryParty = checkTransaction()
-            progressTracker.currentStep = REQUESTING
             logger.info("Sending transaction to notary: ${notaryParty.name}.")
+            progressTracker.currentStep = REQUESTING
             val response = notarise(notaryParty)
             logger.info("Notary responded.")
             progressTracker.currentStep = VALIDATING
@@ -63,6 +64,7 @@ class NotaryFlow {
          * Checks that the transaction specifies a valid notary, and verifies that it contains all required signatures
          * apart from the notary's.
          */
+        // TODO: [CORDA-2274] Perform full transaction verification once verification caching is enabled.
         protected fun checkTransaction(): Party {
             val notaryParty = stx.notary ?: throw IllegalStateException("Transaction does not specify a Notary")
             check(serviceHub.networkMapCache.isNotary(notaryParty)) { "$notaryParty is not a notary on the network" }
@@ -72,18 +74,35 @@ class NotaryFlow {
             stx.resolveTransactionWithSignatures(serviceHub).verifySignaturesExcept(notaryParty.owningKey)
             return notaryParty
         }
-
         /** Notarises the transaction with the [notaryParty], obtains the notary's signature(s). */
         @Throws(NotaryException::class)
         @Suspendable
         protected fun notarise(notaryParty: Party): UntrustworthyData<NotarisationResponse> {
             val session = initiateFlow(notaryParty)
             val requestSignature = generateRequestSignature()
-            return if (serviceHub.networkMapCache.isValidatingNotary(notaryParty)) {
+            return if (isValidating(notaryParty)) {
                 sendAndReceiveValidating(session, requestSignature)
             } else {
                 sendAndReceiveNonValidating(notaryParty, session, requestSignature)
             }
+        }
+
+        private fun isValidating(notaryParty: Party): Boolean {
+            val onTheCurrentWhitelist = serviceHub.networkMapCache.isNotary(notaryParty)
+            return if (!onTheCurrentWhitelist) {
+                /*
+                    Note that the only scenario where it's acceptable to use a notary not in the current network parameter whitelist is
+                    when performing a notary change transaction after a network merge – the old notary won't be on the whitelist of the new network,
+                    and can't be used for regular transactions.
+                */
+                check(stx.coreTransaction is NotaryChangeWireTransaction) {
+                    "Notary $notaryParty is not on the network parameter whitelist. A non-whitelisted notary can only be used for notary change transactions"
+                }
+                val historicNotary = serviceHub.networkParametersStorage.getHistoricNotary(notaryParty)
+                        ?: throw IllegalStateException("The notary party $notaryParty specified by transaction ${stx.id}, is not recognised as a current or historic notary.")
+                historicNotary.validating
+
+            } else serviceHub.networkMapCache.isValidatingNotary(notaryParty)
         }
 
         @Suspendable
--- a/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
@@ -9,6 +9,7 @@ import net.corda.core.crypto.TransactionSignature
 import net.corda.core.identity.Party
 import net.corda.core.internal.BackpressureAwareTimedFlow
 import net.corda.core.internal.FetchDataFlow
+import net.corda.core.internal.notary.HistoricNetworkParameterStorage
 import net.corda.core.internal.notary.generateSignature
 import net.corda.core.internal.notary.validateSignatures
 import net.corda.core.internal.pushToLoggingContext
@@ -98,7 +99,7 @@ class NotaryFlow {
                 check(stx.coreTransaction is NotaryChangeWireTransaction) {
                     "Notary $notaryParty is not on the network parameter whitelist. A non-whitelisted notary can only be used for notary change transactions"
                 }
-                val historicNotary = serviceHub.networkParametersStorage.getHistoricNotary(notaryParty)
+                val historicNotary = (serviceHub.networkParametersStorage as HistoricNetworkParameterStorage).getHistoricNotary(notaryParty)
                         ?: throw IllegalStateException("The notary party $notaryParty specified by transaction ${stx.id}, is not recognised as a current or historic notary.")
                 historicNotary.validating
 
--- a/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
@@ -4,17 +4,7 @@ import co.paralleluniverse.fibers.Suspendable
 import net.corda.core.contracts.StateRef
 import net.corda.core.contracts.TimeWindow
 import net.corda.core.crypto.SecureHash
-import net.corda.core.flows.FlowException
-import net.corda.core.flows.FlowLogic
-import net.corda.core.flows.FlowSession
-import net.corda.core.flows.NotarisationPayload
-import net.corda.core.flows.NotarisationRequest
-import net.corda.core.flows.NotarisationRequestSignature
-import net.corda.core.flows.NotarisationResponse
-import net.corda.core.flows.NotaryError
-import net.corda.core.flows.NotaryException
-import net.corda.core.flows.NotaryFlow
-import net.corda.core.flows.WaitTimeUpdate
+import net.corda.core.flows.*
 import net.corda.core.identity.Party
 import net.corda.core.internal.MIN_PLATFORM_VERSION_FOR_BACKPRESSURE_MESSAGE
 import net.corda.core.internal.checkParameterHash
@@ -55,9 +45,6 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
 
     @Suspendable
     override fun call(): Void? {
-        check(serviceHub.myInfo.legalIdentities.any { serviceHub.networkMapCache.isNotary(it) }) {
-            "We are not a notary on the network"
-        }
         val requestPayload = otherSideSession.receive<NotarisationPayload>().unwrap { it }
 
         try {
@@ -123,6 +110,21 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
         }
     }
 
+    /**
+     * Check that network parameters hash on this transaction is the current hash for the network.
+     */
+    // TODO: ENT-2666 Implement network parameters fuzzy checking. By design in Corda network we have propagation time delay.
+    //     We will never end up in perfect synchronization with all the nodes. However, network parameters update process
+    //     lets us predict what is the reasonable time window for changing parameters on most of the nodes.
+    @Suspendable
+    protected fun checkParametersHash(networkParametersHash: SecureHash?) {
+        if (networkParametersHash == null && serviceHub.networkParameters.minimumPlatformVersion < 4) return
+        val notaryParametersHash = serviceHub.networkParametersStorage.currentHash
+        require(notaryParametersHash == networkParametersHash) {
+            "Transaction for notarisation was tagged with parameters with hash: $networkParametersHash, but current network parameters are: $notaryParametersHash"
+        }
+    }
+
     /** Verifies that the correct notarisation request was signed by the counterparty. */
     private fun validateRequestSignature(request: NotarisationRequest, signature: NotarisationRequestSignature) {
         val requestingParty = otherSideSession.counterparty
@@ -133,8 +135,7 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
      * Override to implement custom logic to perform transaction verification based on validity and privacy requirements.
      */
     @Suspendable
-    protected open fun verifyTransaction(requestPayload: NotarisationPayload) {
-    }
+    abstract fun verifyTransaction(requestPayload: NotarisationPayload)
 
     @Suspendable
     private fun signTransactionAndSendResponse(txId: SecureHash) {
@@ -158,7 +159,7 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
     private fun logError(error: NotaryError) {
         val errorCause = when (error) {
             is NotaryError.RequestSignatureInvalid -> error.cause
-            is NotaryError.TransactionInvalid ->  error.cause
+            is NotaryError.TransactionInvalid -> error.cause
             is NotaryError.General -> error.cause
             else -> null
         }
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
@@ -1,10 +1,10 @@
 package net.corda.core.node.services
 
-import net.corda.core.CordaInternal
 import net.corda.core.DoNotImplement
 import net.corda.core.crypto.SecureHash
-import net.corda.core.internal.SignedDataWithCert
+import net.corda.core.identity.Party
 import net.corda.core.node.NetworkParameters
+import net.corda.core.node.NotaryInfo
 
 /**
  * Interface for handling network parameters storage used for resolving transactions according to parameters that were
@@ -27,4 +27,9 @@ interface NetworkParametersStorage {
      * get them from network map.
      */
     fun lookup(hash: SecureHash): NetworkParameters?
+
+    /**
+     * Returns the [NotaryInfo] for a notary [party] in the current or any historic network parameter whitelist, or null if not found.
+     */
+    fun getHistoricNotary(party: Party): NotaryInfo?
 }
--- a/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
@@ -27,9 +27,4 @@ interface NetworkParametersStorage {
      * get them from network map.
      */
     fun lookup(hash: SecureHash): NetworkParameters?
-
-    /**
-     * Returns the [NotaryInfo] for a notary [party] in the current or any historic network parameter whitelist, or null if not found.
-     */
-    fun getHistoricNotary(party: Party): NotaryInfo?
 }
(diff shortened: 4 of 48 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/MockServices.kt
```diff
@@ -356,9 +360,14 @@ class MockNetworkParametersStorage(private var currentParameters: NetworkParamet
         return getNotaryInfo(party)?.validating ?: throw IllegalStateException("Party $party is not a notary on the network")
     }
 
-    override val currentParametersHash: SecureHash get() = currentParameters.serialize().hash
+    override val currentParametersHash: SecureHash
+        get() {
+            return withTestSerializationEnvIfNotSet("networkParameters") {
```
kasiastreich: Oh noooooo :joy: I was thinking about that too

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -50,27 +51,19 @@ class NotaryFlow {
         @Throws(NotaryException::class)
         override fun call(): List<TransactionSignature> {
             stx.pushToLoggingContext()
-            val notaryParty = checkTransaction()
```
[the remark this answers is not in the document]
adagys: It is already checked in LedgerTransaction (FullTransaction) during construction. We assume here that the transaction has already been validated (which is the case for FinalityFlow and AbstractStateReplacementFlow).
Now, it might be better to re-verify the whole transaction here, but that would result in duplicate verification for the most common path – finality flow. Not sure which is the better approach.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -50,27 +51,19 @@ class NotaryFlow {
...
-            progressTracker.currentStep = REQUESTING
+            val notaryParty = stx.notary ?: throw IllegalStateException("Transaction does not specify a Notary")
             logger.info("Sending transaction to notary: ${notaryParty.name}.")
+            checkEnoughSignatures(notaryParty)
```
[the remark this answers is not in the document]
adagys: the other checks became redundant, so I renamed the method to be more accurate

testing/test-utils/src/main/kotlin/net/corda/testing/internal/InternalTestUtils.kt
```diff
@@ -183,14 +185,29 @@ fun configureDatabase(hikariProperties: Properties,
...
     JarOutputStream(bs, manifest).use { js ->
         js.putNextEntry(ZipEntry(filePath))
         js.writer().apply { append(content); flush() }
         js.closeEntry()
     }
     return bs.toByteArray()
+}
+
+/** If [effectiveSerializationEnv] is not set, runs the block with a new [SerializationEnvironmentRule]. */
+fun <R> withTestSerializationEnvIfNotSet(taskName: String, block: () -> R): R {
```
kasiastreich: This is the almost the same as magic piece of code I showed you the other day. `net.corda.testing.node.NodeTestUtils#ledger` Maybe we should unify it somehow?

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -41,4 +41,14 @@ abstract class FullTransaction : BaseTransaction() {
         check(notaries.size == 1) { "All inputs and reference inputs must point to the same notary" }
         check(notaries.single() == notary) { "The specified notary must be the one specified by all inputs and input references" }
     }
+
+    /** Make sure the assigned notary is part of the network parameter whitelist. */
+    protected fun checkNotaryWhitelisted() {
+        if (notary != null) {
+            val notaryWhitelist = networkParameters!!.notaries.map { it.identity }
```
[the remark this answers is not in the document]
adagys: Yes, the !! is no longer necessary after the rebase.

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -41,4 +41,14 @@ abstract class FullTransaction : BaseTransaction() {
...
+            check(notary!! in notaryWhitelist) {
```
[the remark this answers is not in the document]
adagys: i'll re-write it so this !! is not needed

testing/test-utils/src/main/kotlin/net/corda/testing/internal/InternalTestUtils.kt
kasiastreich: Ah just realised that you moved it :smile:

node/src/main/kotlin/net/corda/node/internal/DBNetworkParametersStorage.kt
```diff
@@ -103,6 +117,26 @@ class DBNetworkParametersStorage(
...
+     */
+    private fun getNotaryInfo(party: Party): NotaryInfo? {
+        val currentParameters = readParametersFromHash(currentParametersHash) ?:
+                throw IllegalStateException("Unable to obtain NotaryInfo – current network parameters not set.")
+        val inCurrentParams = currentParameters.notaries.singleOrNull { it.identity == party }
+        if (inCurrentParams == null) {
+            val inOldParams = hashToParameters.allPersisted().flatMap { (_, signedParams) ->
+                val parameters = signedParams.raw.deserialize()
+                parameters.notaries.asSequence()
+            }.singleOrNull { it.identity == party }
```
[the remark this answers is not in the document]
adagys: Yeah good point

testing/test-utils/src/main/kotlin/net/corda/testing/internal/InternalTestUtils.kt
adagys: Yeah, that's where I drew inspiration!

node/src/main/kotlin/net/corda/node/services/CoreFlowHandlers.kt
```diff
@@ -29,17 +29,11 @@ class NotaryChangeHandler(otherSideSession: FlowSession) : AbstractStateReplacem
     override fun verifyProposal(stx: SignedTransaction, proposal: AbstractStateReplacementFlow.Proposal<Party>) {
         val state = proposal.stateRef
         val proposedTx = stx.resolveNotaryChangeTransaction(serviceHub)
-        val newNotary = proposal.modification
+        // TODO: check if the proposed new notary is acceptable.
```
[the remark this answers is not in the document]
adagys: To be defined in the future, if we want to add extra trust lists etc

node/src/main/kotlin/net/corda/node/internal/DBNetworkParametersStorage.kt
```diff
@@ -85,6 +91,14 @@ class DBNetworkParametersStorage(
         }
     }
 
+    override fun isNotary(party: Party): Boolean {
+        return getNotaryInfo(party) != null
+    }
+
+    override fun isValidatingNotary(party: Party): Boolean {
+        return getNotaryInfo(party)?.validating ?: throw IllegalStateException("Party $party is not a notary on the network")
```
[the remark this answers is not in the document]
adagys: Well, if we don't know whether it's validating we don't know what type of payload to send it, and the flow might fail anyway.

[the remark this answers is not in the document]
kasiastreich: What I am trying to say is that method name indicates boolean check, but behaviour is as if it was `checkSomething` (when party isn't a notary at all). Maybe document that.

testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
```diff
@@ -28,17 +22,19 @@ fun ServiceHub.ledger(
         notary: Party = TestIdentity.fresh("ledger notary").party,
         script: LedgerDSL<TestTransactionDSLInterpreter, TestLedgerDSLInterpreter>.() -> Unit
 ): LedgerDSL<TestTransactionDSLInterpreter, TestLedgerDSLInterpreter> {
-    val serializationExists = try {
-        effectiveSerializationEnv
-        true
-    } catch (e: IllegalStateException) {
-        false
+    val currentParameters = networkParametersStorage.run {
+        readParametersFromHash(currentParametersHash)!!
```
kasiastreich: Add better error. Theoretically null shouldn't happen?

testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
```diff
@@ -28,17 +22,19 @@ fun ServiceHub.ledger(
...
     }
-    return LedgerDSL(TestLedgerDSLInterpreter(this), notary).apply {
-        if (serializationExists) {
+    if (currentParameters.notaries.none { it.identity == notary }) {
+        // Add the notary to the whitelist. Otherwise no constructed transactions will verify.
+        val newParameters = currentParameters.copy(notaries = currentParameters.notaries + NotaryInfo(notary, true))
+        (networkParametersStorage as MockNetworkParametersStorage).setCurrentParametersUnverified(newParameters)
```
kasiastreich: Is it always ran on MockServices? Then it should be extension `MockServices.ledger`

testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
```diff
@@ -28,17 +22,19 @@ fun ServiceHub.ledger(
...
-    } catch (e: IllegalStateException) {
-        false
+    val currentParameters = networkParametersStorage.run {
+        readParametersFromHash(currentParametersHash)!!
     }
-    return LedgerDSL(TestLedgerDSLInterpreter(this), notary).apply {
-        if (serializationExists) {
+    if (currentParameters.notaries.none { it.identity == notary }) {
+        // Add the notary to the whitelist. Otherwise no constructed transactions will verify.
+        val newParameters = currentParameters.copy(notaries = currentParameters.notaries + NotaryInfo(notary, true))
```
kasiastreich: Document that it will add **validating** notary

adagys: well technically no actual nodes are being used with this, so it doesn't matter which type of notary it is

testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
adagys: Yes it should. But I think this is now public api

node/src/main/kotlin/net/corda/node/services/CoreFlowHandlers.kt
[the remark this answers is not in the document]
kasiastreich: Maybe extend comment, because for me it isn't clear and if I look at this TODO in 2 months I will forget.

node/src/main/kotlin/net/corda/node/services/transactions/NonValidatingNotaryFlow.kt
```diff
@@ -19,25 +25,78 @@ import java.time.Duration
...
+                }
+                else -> throw unexpectedTransactionType(tx)
             }
+        } catch (e: Exception) {
+            throw NotaryInternalException(NotaryError.TransactionInvalid(e))
         }
     }
+
+    /** Make sure the transaction notary is part of the network parameter whitelist. */
+    private fun checkNotaryWhitelisted(notary: Party?, attachedParameterHash: SecureHash?) {
```
kasiastreich: when notary can be null in this case?

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -29,13 +29,14 @@ abstract class CoreTransaction : BaseTransaction() {
 abstract class FullTransaction : BaseTransaction() {
     abstract override val inputs: List<StateAndRef<ContractState>>
     abstract override val references: List<StateAndRef<ContractState>>
-    // TODO NetworkParameters field is nullable only because of the API stability and the fact that our ledger transactions exposed constructors
-    //  (they were data classes). This should be revisited.
+    // TODO: NetworkParameters field is nullable only because of the API stability and the fact that our ledger transactions exposed constructors
```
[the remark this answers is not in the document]
shamsasari: Where you do the security checks, throw an exception rather than ignore if the network parameters are null. That would be part of the checks.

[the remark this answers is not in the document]
kasiastreich: But it is effectively API break, by changing semantics. Why construct LedgerTransaction with null parameters then?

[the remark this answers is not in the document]
adagys: But it will end up always throwing if the parameters are null. So it's still a break technically.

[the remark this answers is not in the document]
adagys: (These checks are during construction)

testing/node-driver/src/main/kotlin/net/corda/testing/node/NodeTestUtils.kt
adagys: Maybe we should cache the current network parameters somewhere. It can't ever be null really.

.ci/api-current.txt
```diff
@@ -2278,8 +2278,6 @@ public static class net.corda.core.flows.NotaryFlow$Client extends net.corda.cor
   @NotNull
   public java.util.List<net.corda.core.crypto.TransactionSignature> call()
   @NotNull
-  protected final net.corda.core.identity.Party checkTransaction()
```
adagys: the class has a DoNotImplement annotation

.ci/api-current.txt
```diff
@@ -6385,7 +6383,7 @@ public class net.corda.testing.node.MockServices extends java.lang.Object implem
   @NotNull
   public net.corda.core.node.services.NetworkMapCache getNetworkMapCache()
   @NotNull
-  public final net.corda.core.node.NetworkParameters getNetworkParameters()
+  public net.corda.core.node.NetworkParameters getNetworkParameters()
```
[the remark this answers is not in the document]
adagys: yes, instead of being passed to the constructor, it's not a separate val:
```
  override val networkParameters: NetworkParameters
        get() = networkParametersStorage.run { lookup(currentHash)!! }
```

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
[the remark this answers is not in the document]
rick-r3: I agree these are duplicate.  I'd rather we did a full verify and have that be a no-op (i.e. have tx remember they have been verified) but think making sure that really works needs some time.  In the meantime, let's retain these checks to try and prevent accidents.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
```diff
@@ -79,13 +72,28 @@ class NotaryFlow {
...
 
+        private fun isValidating(notaryParty: Party): Boolean {
+            val onTheCurrentWhitelist = serviceHub.networkMapCache.isNotary(notaryParty)
+            return if (!onTheCurrentWhitelist) {
+                /*
+                    Note that the only scenario where it's acceptable to use a notary not in the current network parameter whitelist is
+                    when performing a notary change transaction after a network merge – the old notary won't be on the whitelist of the new network,
+                    and can't be used for regular transactions.
+                */
+                val historicNotary = serviceHub.networkParametersStorage.getHistoricNotary(notaryParty)
```
rick-r3: Based on the comment, should we be checking this is a notary change tx ?

core/src/main/kotlin/net/corda/core/internal/notary/NotaryServiceFlow.kt
```diff
@@ -54,9 +44,6 @@ abstract class NotaryServiceFlow(val otherSideSession: FlowSession, val service:
 
     @Suspendable
     override fun call(): Void? {
-        check(serviceHub.myInfo.legalIdentities.any { serviceHub.networkMapCache.isNotary(it) }) {
-            "We are not a notary on the network"
-        }
```
rick-r3: Is this because we might have been removed from the current whitelist, but need to process requests for notary change (potentially) ?

adagys: Yes. Not sure this check makes much sense in general.

core/src/main/kotlin/net/corda/core/flows/NotaryFlow.kt
adagys: Sure

core/src/main/kotlin/net/corda/core/transactions/BaseTransactions.kt
```diff
@@ -47,4 +46,17 @@ abstract class FullTransaction : BaseTransaction() {
         check(notaries.size == 1) { "All inputs and reference inputs must point to the same notary" }
         check(notaries.single() == notary) { "The specified notary must be the one specified by all inputs and input references" }
     }
+
+    /** Make sure the assigned notary is part of the network parameter whitelist. */
+    protected fun checkNotaryWhitelisted() {
+        notary?.let { notaryParty ->
+            // Network parameters will never be null of the transaction is resolved from a CoreTransaction rather than constructed directly.
```
rick-r3: null of -> null if

node/src/test/kotlin/net/corda/node/services/transactions/NotaryWhitelistTests.kt
```diff
@@ -0,0 +1,249 @@
...
+     * it's not whitelisted for regular use).
+     */
+    @Test
+    fun `can perform notary change on a de-listed notary`() {
+        // Issue a state using the old notary. It is currently whitelisted.
+        val stateFakeNotary = issueStateOnOldNotary(oldNotary)
+
+        // Remove old notary from the whitelist
+        val parameters = aliceNode.services.networkParameters
+        val newParameters = parameters.copy(notaries = parameters.notaries.drop(1))
```
rick-r3: Can we add an `assert` that the remaining notary is indeed newNotaryName?

node/src/test/kotlin/net/corda/node/services/transactions/NotaryWhitelistTests.kt
```diff
@@ -0,0 +1,249 @@
...
+     */
+    @Test
+    fun `can't perform a regular transaction on a de-listed notary`() {
+        // Issue a state using the old notary. It is currently whitelisted.
+        val state = issueStateOnOldNotary(oldNotary)
+
+        // Remove old notary from the whitelist
+        val parameters = aliceNode.services.networkParameters
+        val newParameters = parameters.copy(notaries = parameters.notaries.drop(1))
+        mockNet.nodes.forEach {
```
rick-r3: Can we add an `assert` that the remaining notary is indeed newNotaryName?

node/src/test/kotlin/net/corda/node/services/transactions/NotaryWhitelistTests.kt
rick-r3: Maybe make `removeOldNotary` helper function, since you use in the next test too.

rick-r3: Actually, better to express as no notaries have the old name.

samples/irs-demo/cordapp/src/test/kotlin/net/corda/irs/contract/IRSTests.kt
```diff
@@ -238,14 +226,20 @@ class IRSTests {
...
-    private val notaryServices = MockServices(listOf("net.corda.irs.contract"), DUMMY_NOTARY.name, rigorousMock(), DUMMY_NOTARY_KEY)
-    private val ledgerServices
-        get() = MockServices(emptyList(), MEGA_CORP.name, rigorousMock<IdentityServiceInternal>().also {
-            doReturn(MEGA_CORP).whenever(it).partyFromKey(MEGA_CORP_PUBKEY)
-            doReturn(null).whenever(it).partyFromKey(ORACLE_PUBKEY)
-        })
+    private val cordappPackages = listOf("net.corda.irs.contract")
+    private val networkParameters = testNetworkParameters().addNotary(dummyNotary.party)
+    private val megaCorpServices = MockServices(cordappPackages, megaCorp, rigorousMock(), networkParameters, megaCorp.keyPair)
+    private val miniCorpServices = MockServices(listOf("net.corda.irs.contract"), miniCorp, rigorousMock(), networkParameters,  miniCorp.keyPair)
```
kasiastreich: Instead of `listOf(...)` use `cordappPackages` variable defined above. Same for line below.

node/src/test/kotlin/net/corda/node/services/persistence/HibernateColumnConverterTests.kt
```diff
@@ -25,6 +35,20 @@ class HibernateColumnConverterTests {
     private lateinit var bankOfCorda: Party
     private lateinit var notary: Party
 
+    class TestCashIssueFlow(private val amount: Amount<Currency>,
```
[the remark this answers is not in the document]
rick-r3: It's to dodge the notary in whitelist check.  We put a random notary in to check the caching logic works, but then the random notary isn't in the whitelist so the test would be failing for that reason instead.

node/src/main/kotlin/net/corda/node/services/transactions/NonValidatingNotaryFlow.kt
```diff
@@ -19,25 +25,78 @@ import java.time.Duration
...
+    }
+
+    /** Make sure the transaction notary is part of the network parameter whitelist. */
+    private fun checkNotaryWhitelisted(notary: Party, attachedParameterHash: SecureHash?) {
+        if (minPlatformVersion >= 4) {
+            // Expecting network parameters to be attached for platform version 4 or later.
+            if (attachedParameterHash == null) {
+                throw IllegalArgumentException("Transaction must contain network parameters.")
             }
+            // TODO: If not found, the notary should resolve network parameters from the network map server or the counterparty.
```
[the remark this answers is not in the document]
thschroeter: I've removed the todo.

node/src/main/kotlin/net/corda/node/services/transactions/NonValidatingNotaryFlow.kt
```diff
@@ -19,25 +25,78 @@ import java.time.Duration
...
+            val attachedParameters = serviceHub.networkParametersStorage.lookup(attachedParameterHash)
+                    ?: throw IllegalStateException("Unable to resolve network parameters from hash: $attachedParameterHash")
+
+            checkInWhitelist(attachedParameters, notary)
+        } else {
+            // Using default network parameters for platform versions 3 or earlier
```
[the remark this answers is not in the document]
thschroeter: Changing to `currentHash`.

core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
```diff
@@ -27,4 +27,9 @@ interface NetworkParametersStorage {
      * get them from network map.
      */
     fun lookup(hash: SecureHash): NetworkParameters?
+
+    /**
+     * Returns the [NotaryInfo] for a notary [party] in the current or any historic network parameter whitelist, or null if not found.
+     */
+    fun getHistoricNotary(party: Party): NotaryInfo?
```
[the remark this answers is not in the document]
thschroeter: Moved to `HistoricNetworkParameterStorage` in `core.internal`, and casting in the `NotaryFlow`, see https://github.com/corda/corda/pull/4346/commits/b18497329792ce3f5fcf195b199831101082cb80

node/src/main/kotlin/net/corda/node/services/CoreFlowHandlers.kt
```diff
@@ -29,17 +29,11 @@ class NotaryChangeHandler(otherSideSession: FlowSession) : AbstractStateReplacem
     override fun verifyProposal(stx: SignedTransaction, proposal: AbstractStateReplacementFlow.Proposal<Party>) {
         val state = proposal.stateRef
         val proposedTx = stx.resolveNotaryChangeTransaction(serviceHub)
-        val newNotary = proposal.modification
+        // TODO: Right now all nodes will automatically approve the notary change. We need to figure out if stricter controls are necessary.
```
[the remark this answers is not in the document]
rick-r3: It does seem like this check should be retained.  We have talked about the old one not being required to be in the current whitelist, but the new one should be.  

@thschroeter I think this file can mostly be reverted, but the original TODO (below) can be removed now because the network params are part of the isNotary check.

vkolomeyko:
Looks like these changes are breaking the `master` build: https://ci-master.corda.r3cev.com/viewLog.html?buildId=129739&buildTypeId=Corda_CordaBuild

thschroeter:
Fixing.

node/src/main/kotlin/net/corda/node/services/CoreFlowHandlers.kt
```diff
@@ -34,6 +34,12 @@ class NotaryChangeHandler(otherSideSession: FlowSession) : AbstractStateReplacem
         if (state !in proposedTx.inputs.map { it.ref }) {
             throw StateReplacementException("The proposed state $state is not in the proposed transaction inputs")
         }
+
+        val newNotary = proposal.modification
+        val isNotary = serviceHub.networkMapCache.isNotary(newNotary)
```
adagys: this check is redundant btw, we always check it in the notary change ledger transaction during construction.

node/src/main/kotlin/net/corda/node/services/transactions/NonValidatingNotaryFlow.kt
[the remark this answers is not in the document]
adagys: mm why? I thought that's what the default parameters were for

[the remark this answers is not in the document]
adagys: although yeah, current makes more sense here

PR #4294, opened 2018-11-26
CORDA-2253 Avoid leaking unpatched Cycles via the global type info cache

distributedleetravis:
This PR fixes the bug described in [CORDA-2253](https://r3-cev.atlassian.net/browse/CORDA-2253), by using a local cache of constructed `LocalTypeInformation` during traversal of the type graph, and merging the local cache into the global cache once all type information has been constructed and cycles have been "patched" with the type information for the types they refer to.

### DIFF
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/model/LocalTypeModel.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/model/LocalTypeModel.kt
@@ -41,12 +41,6 @@ interface LocalTypeModel {
      * @param type The [Type] to get [LocalTypeInformation] for.
      */
     fun inspect(type: Type): LocalTypeInformation
-
-    /**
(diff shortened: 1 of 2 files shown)


serialization/src/main/kotlin/net/corda/serialization/internal/model/LocalTypeModel.kt
```diff
@@ -41,12 +41,6 @@ interface LocalTypeModel {
      * @param type The [Type] to get [LocalTypeInformation] for.
      */
     fun inspect(type: Type): LocalTypeInformation
-
-    /**
-     * Get [LocalTypeInformation] directly from the registry by [TypeIdentifier], returning null if no type information
-     * is registered for that identifier.
-     */
-    operator fun get(typeIdentifier: TypeIdentifier): LocalTypeInformation?
```
distributedleetravis: Removed because no longer used.

thschroeter:
Tested locally.

PR #4306, opened 2018-11-27
CORDA-2255 suppress warnings on readonly properties

distributedleetravis:
Sometimes the `LocalTypeInformationBuilder` will observe types that are non-composable (i.e. we can't find a unique deserialisation constructor for them, or we can't match their constructor's signature with their observable properties) while analysing read-only properties of abstract classes or interfaces. Because we will never deserialize these properties directly (they will typically be handled by a custom serializer, or else we will be using the deserializer for a concrete implementation of that abstract type), it doesn't matter that they can't be deserialized by an `ObjectSerializer`, and we don't need to warn the user that we've encountered a non-composable type.

This PR extends warning-suppression so that it covers the following cases:

* The type is flagged as `Opaque`.
* The type is abstract or an interface, so we are just looking at its read-only properties.
* The type extends `Exception`, in which case it will be handled by the special Exception serialization code.

This removes the spurious warnings observed when running `samples:bank-of-corda-demo:deployNodes` (see [Corda-2255](https://r3-cev.atlassian.net/browse/CORDA-2255) for details).

### DIFF
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/model/LocalTypeInformationBuilder.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/model/LocalTypeInformationBuilder.kt
@@ -35,11 +35,15 @@ import kotlin.reflect.jvm.javaType
 internal data class LocalTypeInformationBuilder(val lookup: LocalTypeLookup,
                                                 val resolutionContext: Type? = null,
                                                 val visited: Set<TypeIdentifier> = emptySet(),
-                                                val cycles: MutableList<LocalTypeInformation.Cycle> = mutableListOf()) {
+                                                val cycles: MutableList<LocalTypeInformation.Cycle> = mutableListOf(),
+                                                val warnIfNonComposable: Boolean = true) {
 
     companion object {
         private val logger = contextLogger()
     }
+
+    fun <T> suppressingWarnings(block: LocalTypeInformationBuilder.() -> T): T =
+            copy(warnIfNonComposable = false).block()
 
     /**
      * Recursively build [LocalTypeInformation] for the given [Type] and [TypeIdentifier]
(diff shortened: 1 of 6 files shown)


serialization/src/main/kotlin/net/corda/serialization/internal/model/LocalTypeInformationBuilder.kt
```diff
@@ -35,12 +35,20 @@ import kotlin.reflect.jvm.javaType
...
     companion object {
         private val logger = contextLogger()
     }
 
+    /**
+     * If we are examining the type of a read-only property, or a type flagged as [Opaque], then we do not need to warn
+     * if the [LocalTypeInformation] for that type (or any of its related types) is [LocalTypeInformation.NonComposable].
+     */
+    private fun <T> suppressingWarnings(block: LocalTypeInformationBuilder.() -> T): T =
+            copy(warnIfNonComposable = false).block()
```
shamsasari: Let's not make unnecessary objects like this. A private `var` that we reset in a `finally` block would be better.

serialization/src/main/kotlin/net/corda/serialization/internal/model/LocalTypeInformationBuilder.kt
```diff
@@ -35,12 +35,20 @@ import kotlin.reflect.jvm.javaType
...
 
     companion object {
         private val logger = contextLogger()
     }
 
+    /**
+     * If we are examining the type of a read-only property, or a type flagged as [Opaque], then we do not need to warn
+     * if the [LocalTypeInformation] for that type (or any of its related types) is [LocalTypeInformation.NonComposable].
+     */
+    private fun <T> suppressingWarnings(block: LocalTypeInformationBuilder.() -> T): T =
```
shamsasari: `suppressWarnings`?

shamsasari: inline

tools/shell/src/main/java/net/corda/tools/shell/FlowShellCommand.java
```diff
@@ -37,7 +37,7 @@ public void start(
             @Usage("The data to pass as input") @Argument(unquote = false) List<String> input
     ) {
         logger.info("Executing command \"flow start {} {}\",", name, (input != null) ? input.stream().collect(joining(" ")) : "<no arguments>");
-        startFlow(name, input, out, ops(), ansiProgressRenderer(), objectMapper());
+        startFlow(name, input, out, ops(), ansiProgressRenderer(), objectMapper(null));
```
shamsasari: these last set of changes are already in master

PR #4308, opened 2018-11-27
ENT-2320 Fix comment

distributedleetravis:
Fix a rogue comment.


distributedleetravis:
Rebased against master so that unit tests don't fail with the error introduced into master by a previous PR.

PR #4309, opened 2018-11-27
[CORDA-2055] [CORDA-2236] Bootstrapper cordapp copying

anthonykeenan:
- Deprecate `--no-copy` flag in favour of `--copy-cordapps` option 
- `--copy-cordapps` option can be `OnFirstRun`, `Yes` or `No`. Default is `OnFirstRun`.
- - `OnFirstRun` is the default and only copies cordapps across if this is the first time the network has been bootstrapped
- - `Yes` always copies cordapps across
- - `No` never copies cordapps across

Also, `runnodes.jar` will never be copied into the cordapps directory.

Hopefully this will be a more natural user experience

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
@@ -213,9 +213,6 @@ internal constructor(private val initSerEnv: Boolean,
     ) {
         directory.createDirectories()
         println("Bootstrapping local test network in $directory")
-        if (!fromCordform) {
-            println("Found the following CorDapps: ${cordappJars.map { it.fileName }}")
-        }
         val networkAlreadyExists = createNodeDirectoriesIfNeeded(directory, fromCordform)
         val nodeDirs = gatherNodeDirectories(directory)
 
@@ -226,19 +223,9 @@ internal constructor(private val initSerEnv: Boolean,
 
         val configs = nodeDirs.associateBy({ it }, { ConfigFactory.parseFile((it / "node.conf").toFile()) })
         checkForDuplicateLegalNames(configs.values)
-        if ((copyCordapps == CopyCordapps.Yes || (copyCordapps == CopyCordapps.OnFirstRun && !networkAlreadyExists)) && cordappJars.isNotEmpty()) {
-            println("Copying CorDapp JARs into node directories")
-            for (nodeDir in nodeDirs) {
-                val cordappsDir = (nodeDir / "cordapps").createDirectories()
-                cordappJars.forEach {
-                    try {
-                        it.copyToDirectory(cordappsDir)
-                    } catch (e: FileAlreadyExistsException) {
-                        println("WARNING: ${it.fileName} already exists in $cordappsDir, ignoring and leaving existing CorDapp untouched")
-                    }
-                }
-            }
-        }
+
+        cordappJars.copy(nodeDirs, copyCordapps, networkAlreadyExists, fromCordform)
+
         generateServiceIdentitiesForNotaryClusters(configs)
         if (initSerEnv) {
             initialiseSerialization()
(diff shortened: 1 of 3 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -75,6 +75,8 @@ internal constructor(private val initSerEnv: Boolean,
 
         private const val LOGS_DIR_NAME = "logs"
 
+        private val jarsThatArentCordapps = setOf("corda.jar", "runnodes.jar")
```
sollecitom: I would probably use a predicate here, something like:

```
private val cordaArtifacts = setOf("corda.jar", "runnodes.jar")

private fun Path.isJar() = fileName.toString().endsWith(".jar")

private fun Path.isCorDappJar() = isJar() && !fileName.toString().toLowercase() in cordaArtifacts
```

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -482,5 +503,7 @@ data class NetworkParametersOverrides(
 )
 
 interface NetworkBootstrapperWithOverridableParameters {
-    fun bootstrap(directory: Path, copyCordapps: Boolean, networkParameterOverrides: NetworkParametersOverrides = NetworkParametersOverrides())
-}
\ No newline at end of file
+    fun bootstrap(directory: Path, copyCordapps: CopyCordapps, networkParameterOverrides: NetworkParametersOverrides = NetworkParametersOverrides())
+}
+
+enum class CopyCordapps { OnFirstRun, Yes, No }
```
sollecitom: I'd rename `CopyCordapps.OnFirstRun` to `CopyCordapps.FirstRunOnly`.

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
```diff
@@ -268,7 +257,33 @@ internal constructor(private val initSerEnv: Boolean,
         }
     }
 
-    private fun createNodeDirectoriesIfNeeded(directory: Path, fromCordform: Boolean) {
+    private fun List<Path>.copy(nodeDirs: List<Path>, copyCordapps: CopyCordapps, networkAlreadyExists: Boolean, fromCordform: Boolean) {
+        if (!fromCordform) {
+            println("Found the following CorDapps: ${this.map { it.fileName }}")
+        }
+        if (copyCordapps == CopyCordapps.OnFirstRun && networkAlreadyExists) {
```
sollecitom: Replace this with polymorphism?

node-api/src/main/kotlin/net/corda/nodeapi/internal/network/NetworkBootstrapper.kt
sollecitom: Also, polymorphism might good for this enum.

tools/bootstrapper/src/main/kotlin/net/corda/bootstrapper/Main.kt
```diff
@@ -85,12 +84,28 @@ class NetworkBootstrapperRunner(private val bootstrapper: NetworkBootstrapperWit
...
         )
         return ExitCodes.SUCCESS
     }
 }
+
+enum class CopyCordapps {
+    FirstRunOnly, Yes, No;
+}
+
+fun CopyCordapps.toCordappCopierOption() : CordappCopierOption {
```
sollecitom: Could we not put this directly in the `enum`? Or are these getting serialised somehow?

anthonykeenan: I didn't realise you could do that, I've changed it now.

PR #4318, opened 2018-11-28
CORDA-1456: Change output format in the shell

dimosr:
Added support for YAML as an output format and an option for choosing between JSON/YAML, when running an RPC command via the node's shell.

Context on some decisions:
- Collections were handled customly until now, but it seems that the serializer could handle them equally well, so removed the custom handling to simplify code. Synced with @sollecitom who had previously touched the code to make sure this is safe.
- The `runFlowByNameFragment` method is currently receiving an object mapper, so plugging variable formats there would be a bit more intrusive. For now, I kept it as-is with, keeping JSON as the default one, since users won't have the option to switch between JSON/YAML and they might have integrated with the json format now.
- I initially went for adding XML support, as well. However, this is not achievable just by using another factory and would require a custom object mapper (see: https://github.com/FasterXML/jackson-dataformat-xml). In the current state, it would require additional refactoring, because object mapper currently is done through `JacksonSupport.createNonRpcMapper`.

# Testing
- Run the unit/integration tests for `tools`
- Run the `bank-of-corda` demo with the new changes and verified the option is handled correctly. Below are the results of it.

```
>>> run --format json networkMapSnapshot
[ {
  "addresses" : [ "localhost:10008" ],
  "legalIdentitiesAndCerts" : [ "O=BigCorporation, L=New York, C=US" ],
  "platformVersion" : 4,
  "serial" : 1543403840475
}, {
  "addresses" : [ "localhost:10005" ],
  "legalIdentitiesAndCerts" : [ "O=BankOfCorda, L=London, C=GB" ],
  "platformVersion" : 4,
  "serial" : 1543403840587
}, {
  "addresses" : [ "localhost:10002" ],
  "legalIdentitiesAndCerts" : [ "O=Notary Service, L=Zurich, C=CH" ],
  "platformVersion" : 4,
  "serial" : 1543403838479
} ]

>>> run --format yaml networkMapSnapshot (or run networkMapSnapshot)
- addresses:
  - "localhost:10008"
  legalIdentitiesAndCerts:
  - "O=BigCorporation, L=New York, C=US"
  platformVersion: 4
  serial: 1543403840475
- addresses:
  - "localhost:10005"
  legalIdentitiesAndCerts:
  - "O=BankOfCorda, L=London, C=GB"
  platformVersion: 4
  serial: 1543403840587
- addresses:
  - "localhost:10002"
  legalIdentitiesAndCerts:
  - "O=Notary Service, L=Zurich, C=CH"
  platformVersion: 4
  serial: 1543403838479
```


tools/shell/src/test/kotlin/net/corda/tools/shell/InteractiveShellTest.kt
```diff
@@ -1,30 +1,67 @@
...
+    }
+
     companion object {
         private val megaCorp = TestIdentity(CordaX500Name("MegaCorp", "London", "GB"))
+
+        private val ALICE = getTestPartyAndCertificate(ALICE_NAME, generateKeyPair().public)
+        private val BOB = getTestPartyAndCertificate(BOB_NAME, generateKeyPair().public)
+        private val ALICE_NODE_INFO = NodeInfo(listOf(NetworkHostAndPort("localhost", 8080)), listOf(ALICE), 1, 1)
+        private val BOB_NODE_INFO = NodeInfo(listOf(NetworkHostAndPort("localhost", 80)), listOf(BOB), 1, 1)
+        private val NODE_INFO_JSON_PAYLOAD = this::class.java.getResource("/node_info_payload.json").readText()
```
shamsasari: Wouldn't it make more sense to have the expected output string as a literal in the test. It makes it easier to read the test. Kotlin supports raw strings so you wouldn't have to worry about new lines.

tools/shell/src/test/kotlin/net/corda/tools/shell/InteractiveShellTest.kt
```diff
@@ -57,6 +94,15 @@ class InteractiveShellTest {
         }, input, FlowA::class.java, om)
         assertEquals(expected, output!!, input)
     }
+
+    private fun objectMapperWithClassLoader(classLoader: ClassLoader?): ObjectMapper {
+        val objectMapper = ObjectMapper()
+        val tf = TypeFactory.defaultInstance()
+                .withClassLoader(classLoader)
```
shamsasari: Move up

tools/shell/src/test/kotlin/net/corda/tools/shell/InteractiveShellTest.kt
```diff
@@ -125,6 +171,32 @@ class InteractiveShellTest {
     @Test
     fun party() = check("party: \"${megaCorp.name}\"", megaCorp.name.toString())
 
+    @Test
+    fun runRpcFromStringWithCustomTypeResult() {
+        val command = listOf("nodeInfo")
+        whenever(cordaRpcOps.nodeInfo())
+                .thenReturn(ALICE_NODE_INFO)
```
shamsasari: likwise

tools/shell/src/test/kotlin/net/corda/tools/shell/InteractiveShellTest.kt
dimosr: That's how I started actually, I thought but I was afraid that it might become clunky as we add more tests. I agree though, I'll switch to raw strings.

docs/source/shell.rst
```diff
@@ -195,6 +195,14 @@ node by running:
 
 Some RPCs return a stream of events that will be shown on screen until you press Ctrl-C.
 
+You can choose the format of the output, using the ``--format`` option, i.e. by running:
+
+``run --format json networkMapSnapshot``
+
+The output formats currently supported are: ``json``, ``yaml``. The default format is ``yaml``.
+
+.. note:: The output formats do not apply to RPCs that return a stream of events.
```
shamsasari: How comes? It would be nice if the output format was consistent. Can you take a look and see if you can address it in this PR?

dimosr: Sorry, I think the phrasing is deceiving. The format of each record is still adjustable, but the overall text is not valid JSON/YAML. As an example:
```
run --format json networkMapFeed
Snapshot:
[ {
  "addresses" : [ "localhost:10002" ],
  "legalIdentitiesAndCerts" : [ "O=Notary Service, L=Zurich, C=CH" ],
  "platformVersion" : 4,
  "serial" : 1543403838479
}, {
  "addresses" : [ "localhost:10005" ],
  "legalIdentitiesAndCerts" : [ "O=BankOfCorda, L=London, C=GB" ],
  "platformVersion" : 4,
  "serial" : 1543403840587
}, {
  "addresses" : [ "localhost:10008" ],
  "legalIdentitiesAndCerts" : [ "O=BigCorporation, L=New York, C=US" ],
  "platformVersion" : 4,
  "serial" : 1543403840475
} ]
Updates:
...
```

Thinking it twice, I suspect that this is not the expectation though, so perhaps this note can be removed completely.

dimosr:
> i wonder if it makes sense to have a separate command that switches the default format

Makes sense, I think the CX will be much better (and indeed, if we do it, I think the `--format` option  should go away). The main thing I am contemplating on (prob. because of lack of context) is:
- If I understand correctly, till now the shell from a user-perspective was stateless and this will be the first modification that will change this. Is this true and are we fine with this?
- Also, I assume there's no expectation for this format setting to persist across restarts etc. (it can be simple in-memory), right ?

shamsasari:
I think for the shell to be more useful than it is it will have to be stateful, and persisted. For now, just an in-memory setting is sufficient.

dimosr:
Agree, I'll amend with an in-memory one for now and we can follow-up on this for persistence.

dimosr:
Small note: Having another look at `runFlowByNameFragment`, it seems that the objectMapper that's passed as argument is actually used as mapper for the input, not the output. So, I removed the default value, which was not used by any client anyway.

PR #4319, opened 2018-11-28
CORDA-2266 match old fingerprinter on Any type

distributedleetravis:
As per [CORDA-2266](https://r3-cev.atlassian.net/browse/CORDA-2266), notaries using pre-v4 code will fall into evolution serialisation (and fail) if the type descriptor of a `NotarisationPayload` sent to them doesn't match their own fingerprint. Prior to v4, the fingerprinter treated `Any` differently to `?`, and as `NotarisationPayload` has a parameter of type `Any` this results in an incompatible fingerprint being generated. The solution is to make the new fingerprinter do the same thing as the old one with these two types.


vkolomeyko:
Would it be possible to have a unit test that if not exposes the problem ensures that `Unknown` and `Any` types produce expected fingerprint?

distributedleetravis:
Added.

PR #4321, opened 2018-11-28
CORDA-2269: Removed DriverParameters.initialiseSerialization as it's not needed

shamsasari:
Instead we detect if a serialisation environment already exists and initialise one if not.


.ci/api-current.txt
```diff
@@ -5893,8 +5893,6 @@ public interface net.corda.testing.driver.DriverDSL
 public final class net.corda.testing.driver.DriverParameters extends java.lang.Object
   public <init>()
   public <init>(boolean, java.nio.file.Path, net.corda.testing.driver.PortAllocation, net.corda.testing.driver.PortAllocation, java.util.Map<String, String>, boolean, boolean, boolean, java.util.List<net.corda.testing.node.NotarySpec>, java.util.List<String>, net.corda.testing.driver.JmxPolicy, net.corda.core.node.NetworkParameters)
```
shamsasari: This is the only c'tor that existed in V3, so deleting the others is OK: https://github.com/corda/corda/blob/1c4542fed771e765a7b2a89a8672dbf52266cce3/.ci/api-current.txt#L5974

PR #4322, opened 2018-11-29
ENT-2666: Don't check for network parameter currentness for now

adagys:
At flag day, not all nodes may update their network parameters at the same time. There will be a period of time where the previous version of the parameters should still be accepted. That requires e.g. some timewindow encoded in the parameters, but it needs more thinking.
Until the design work is done, it's better to not check for parameter  "currentness" at all to not cause flow failures right after flag day.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/TransactionUtils.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/TransactionUtils.kt
@@ -166,6 +166,8 @@ fun FlowLogic<*>.checkParameterHash(networkParametersHash: SecureHash?) {
     if (networkParametersHash == null) {
         if (serviceHub.networkParameters.minimumPlatformVersion < 4) return
         else throw IllegalArgumentException("Transaction for notarisation doesn't contain network parameters hash.")
+    } else {
+        serviceHub.networkParametersStorage.lookup(networkParametersHash) ?: throw IllegalArgumentException("Transaction for notarisation contains unknown parameters hash: $networkParametersHash")
     }
 
     // TODO: [ENT-2666] Implement network parameters fuzzy checking. By design in Corda network we have propagation time delay.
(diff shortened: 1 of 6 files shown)


mikehearn:
Can we just accept any params that have been advertised by the zone for this? Happy to put the currentness check off to later, just wondering if there's a simple fix here.

adagys:
Yes that's the idea. Relaxing this check means older versions of the zone parameters will be accepted too.

kasiastreich:
Yes we can do that too, they are stored as `SignedDataWithCert` so it's easy to do.

core/src/main/kotlin/net/corda/core/internal/TransactionUtils.kt
```diff
@@ -157,3 +158,18 @@ fun createComponentGroups(inputs: List<StateRef>,
...
+/** Check that network parameters hash on this transaction is the current hash for the network. */
+fun FlowLogic<*>.checkParameterHash(networkParametersHash: SecureHash?) {
+    // Transactions created on Corda 3.x or below do not contain network parameters,
+    // so no checking is done until the minimum platform version is at least 4.
+    if (networkParametersHash == null) {
+        if (serviceHub.networkParameters.minimumPlatformVersion < 4) return
+        else throw IllegalArgumentException("Transaction for notarisation doesn't contain network parameters hash.")
+    }
+
+    // TODO: [ENT-2666] Implement network parameters fuzzy checking. By design in Corda network we have propagation time delay.
```
rick-r3: So does mean any hash is okay, or any hash that our network/node knows is okay?

node/src/test/kotlin/net/corda/node/services/transactions/NotaryServiceTests.kt
```diff
@@ -62,9 +63,10 @@ class NotaryServiceTests {
         mockNet.runNetwork()
         val ex = assertFailsWith<NotaryException> { future.getOrThrow() }
         val notaryError = ex.error as NotaryError.TransactionInvalid
-        assertThat(notaryError.cause).hasMessageContaining("Transaction for notarisation was tagged with parameters with hash: null")
+        assertThat(notaryError.cause).hasMessageContaining("Transaction for notarisation doesn't contain network parameters hash.")
     }
 
+    @Ignore("Re-enable the test when parameters currentness checks are in place, ENT-2666.")
```
rick-r3: Are we able to have a test to check we reject an unknown hash ?

kasiastreich: yes ;)

PR #4331, opened 2018-11-30
CORDA-2277 Shorten table name

thschroeter:
Shorten notary table name for older Oracle deployments. The names can't be longer than 30 characters.


szymonsztuka:
The change is for a new table being added in C4 (https://github.com/corda/corda/commit/8f15cfaa2849faadd9d382f9ce868cc0984afef2), so no need for migration from C3.

gendal:
ah -ok.. thanks.

PR #4332, opened 2018-11-30
[CORDA-2278]: KillFlow RPC operation is not releasing soft locks.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2278

Seems to be working. Leaving the test to ensure no future regressions.


testing/test-utils/build.gradle
```diff
@@ -21,14 +21,6 @@ dependencies {
...
     compile "com.google.guava:guava:$guava_version"
 
-    // Unit testing helpers.
-    compile "junit:junit:$junit_version"
-    compile 'org.hamcrest:hamcrest-library:1.3'
-    compile "com.nhaarman:mockito-kotlin:$mockito_kotlin_version"
-    compile "org.mockito:mockito-core:$mockito_version"
-    compile "org.assertj:assertj-core:$assertj_version"
-    compile "com.natpryce:hamkrest:$hamkrest_version"
-
```
thschroeter: @sollecitom, don't think we can remove those dependencies.

thschroeter: scratch that, it works, sorry.

PR #4333, opened 2018-11-30
CORDA-2128: makeTestIdentityService no longer declared to return internal InMemoryIdentityService

shamsasari:
While technically a break in the test API, it has no practical implications as InMemoryIdentityService does not provide any extra APIs.


shamsasari:
If they declare the local variable as `InMemoryIdentityService`, rather than `IdentityService`, then they'll need to change it to `IdentityService`.

PR #4334, opened 2018-11-30
CORDA-2128: Removed SerializationEnvironmentRule.run

shamsasari:
It exposed the internal SerializationEnvironment class. It previously documented that the SerializationEnvironmentRule JUnit rule should be used instead, and so any existing test code have a migration path.


shamsasari:
Refactored out the method completely.

shamsasari:
All green now.

PR #4342, opened 2018-12-02
Node startup minor refactor

anthonykeenan:


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -284,6 +284,7 @@ open class NodeStartup : NodeStartupLogging {
             println("Corda Node process in now exiting. Please check directory permissions and try starting the Node again.")
             System.exit(1)
```
sollecitom: Move this side effect out.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -267,7 +267,7 @@ open class NodeStartup : NodeStartupLogging {
             if (pidFileLock == null) {
                 println("It appears there is already a node running with the specified data directory $baseDirectory")
```
sollecitom: Even the logging I would say, but up to you.

anthonykeenan: There are two different possible errors, the true/false return wouldn't be enough, and also mean adding more cruft into NodeStartup, or another layer. I think I'd rather leave it as is.

PR #4343, opened 2018-12-03
CORDA-1497 enum evolution

distributedleetravis:
PR addressing [CORDA-1497](https://r3-cev.atlassian.net/browse/CORDA-1497).

When the transform annotations for an enum included both renaming and defaulting of enum constants, the validation was failing if we wanted to rename a defaulted constant.

```kotlin
@CordaSerializationTransformEnumDefaults(
        CordaSerializationTransformEnumDefault(old = "C", new = "D"),
        CordaSerializationTransformEnumDefault(old = "D", new = "E")
)
@CordaSerializationTransformRename(to = "BOB", from = "E")
enum class MultiOperations { A, B, C, D, BOB }
```

In this PR, we clean up the handling of enum transforms a bit, eliminating the transforms cache from `LocalSerializerFactory` and storing transforms directly in `LocalTypeInformation.AnEnum`, and move validation into the `EnumTransforms` class where it is easier to perform.

### DIFF
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
@@ -62,14 +62,6 @@ interface LocalSerializerFactory {
      * Use the [FingerPrinter] to create a type descriptor for the given [typeInformation].
      */
     fun createDescriptor(typeInformation: LocalTypeInformation): Symbol
-
-    /**
-     * Obtain or register [Transform]s for the given class [name].
-     *
-     * Eventually this information should be moved into the [LocalTypeInformation] for the type.
-     */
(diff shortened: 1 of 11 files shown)


serialization/src/main/kotlin/net/corda/serialization/internal/amqp/LocalSerializerFactory.kt
```diff
@@ -62,14 +62,6 @@ interface LocalSerializerFactory {
      * Use the [FingerPrinter] to create a type descriptor for the given [typeInformation].
      */
     fun createDescriptor(typeInformation: LocalTypeInformation): Symbol
-
-    /**
-     * Obtain or register [Transform]s for the given class [name].
-     *
-     * Eventually this information should be moved into the [LocalTypeInformation] for the type.
```
distributedleetravis: And now it has been.

PR #4344, opened 2018-12-03
[CORDA-2285]: Enabled jUnit parallel execution.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2285

Gradle property "test.parallel" controls this feature, default to `false`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -265,6 +265,14 @@ allprojects {
     }
 }
 
+subprojects {
+    tasks.withType(Test) {
+        if (project.getProperty('test.parallel').toBoolean()) {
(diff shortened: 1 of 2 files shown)


mikehearn:
This was previously activated already via a TeamCity configuration (`-Dtest.maxParallelForks=4`). I believe it was de-activated by someone due to someone making a code change that wasn't compatible with it, and then they never investigated. Rather than merge this PR, why not just tweak that TeamCity build and investigate? You can use the performance-experiments branch (just force push over it from master) and the equivalent TC build to do whatever perf experiments you'd like.

PR #4345, opened 2018-12-03
CORDA-2254: Replaced JmxPolicy.httpPort with the original PortAllocation

shamsasari:
The httpPort parameter is fundamentally wrong as it gives the same monitoring port to each node!

Added JmxPolicy.defaultEnabled() to workaround the (incorrect) default false value of startJmxHttpServer.


shamsasari:
This is a follow up to https://github.com/corda/corda/pull/4298.

shamsasari:
`jmxHttpServerPortAllocation` has been made non-null. With `startJmxHttpServer` it feels like an unnecessary complication.

PR #4349, opened 2018-12-03
CORDA-1497 replace missing validation

distributedleetravis:
Reviewing the [previous PR](https://github.com/corda/corda/pull/4343) for [CORDA-1497](https://r3-cev.atlassian.net/projects/CORDA/issues/CORDA-1497), I noticed that I'd accidentally removed a couple of validation rules (they weren't tested, so it didn't break anything, but it's still good to have them). This PR restores those rules, together with tests.

### DIFF
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/model/EnumTransforms.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/model/EnumTransforms.kt
@@ -53,7 +53,7 @@ data class EnumTransforms(
     }
 
     private fun validate(constants: Map<String, Int>): EnumTransforms {
-        validateNoCycles()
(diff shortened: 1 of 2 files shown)


serialization/src/main/kotlin/net/corda/serialization/internal/model/EnumTransforms.kt
```diff
@@ -105,13 +109,17 @@ data class EnumTransforms(
             }
 
             // Either update, or create, the chains in both directions.
-            chainStartsToEnds[from] = newEnd
-            chainEndsToStarts[to] = newStart
-
```
distributedleetravis: Turns out we don't need to do this - the two lines immediately following ensure that the longest possible chain is always recorded.

PR #4351, opened 2018-12-03
CORDA-2334 - Net Params non-downgrade verification in transaction resolution / building.

kasiastreich:
This is a successor to the story ["Tag transactions with the set of network parameters which govern them"](CORDA-2089), and consists of NP epoch verification in 2 places:

1) Addition of no downgrade rule for parameters in ResolveTransactionFlow
Make sure that parameters in the transaction graph are ordered (this is to prevent the downgrade attack, when the malicious notary and participants sign transaction that shouldn't be notarised otherwise). We ensure that by checking that epochs of network parameters in the transaction chain are ordered.

2) Verification in transaction building that the NP's associated with a new Txns inputs (eg. previous txns outputs) are equal or older than the current node NP's building the transaction.


core/src/main/kotlin/net/corda/core/internal/FetchDataFlow.kt
```diff
@@ -193,3 +191,37 @@ class FetchTransactionsFlow(requests: Set<SecureHash>, otherSide: FlowSession) :
...
+    override fun convert(wire: SignedDataWithCert<NetworkParameters>): SignedParametersByHash = SignedParametersByHash(wire)
+
+    override fun maybeWriteToDisk(downloaded: List<SignedParametersByHash>) {
+        for (parameters in downloaded) {
+            with(serviceHub.networkParametersStorage) {
+                if (!hasParameters(parameters.id)) {
+                    // This will perform the signature check too and throws with SignatureVerificationException
+                    saveParameters(parameters.signedParameters)
+                } else {
+                    logger.debug("Network parameters ${parameters.id} already exists in storage, skipping.")
```
rick-r3: Debug should use `{` not `(` to make it lazy.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +77,73 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
+                        "for dependency transaction: ${stx.id} and parent parameters hash: $parentHash")
+                    }
+                }
+            }
+            stx.verify(serviceHub)
+            serviceHub.recordTransactions(StatesToRecord.NONE, listOf(stx))
         }
     }
 
+    // At the beginning transactions from the constructor should be tagged with parameters with epoch less equal than than the current ones for the network.
```
rick-r3: less than or equal

core/src/main/kotlin/net/corda/core/node/services/NetworkParametersStorage.kt
```diff
@@ -27,4 +27,29 @@ interface NetworkParametersStorage {
      * get them from network map.
      */
     fun lookup(hash: SecureHash): NetworkParameters?
+
+    // TODO The methods below should be internal, they are used only in ResolveTransactionsFlow and FetchParametersFlow
```
rick-r3: We should move (the old form of) `NetworkParametersStorageInternal` to `core` and merge with `HistoricNetworkParameterStorage` coming in https://github.com/corda/corda/pull/4293

core/src/main/kotlin/net/corda/core/internal/FetchDataFlow.kt
```diff
@@ -139,7 +140,7 @@ sealed class FetchDataFlow<T : NamedByHash, in W : Any>(
 }
 
 /**
- * Given a set of hashes either loads from from local storage  or requests them from the other peer. Downloaded
+ * Given a set of hashes either loads from from local storage or requests them from the other peer. Downloaded
```
rick-r3: from from -> from

core/src/main/kotlin/net/corda/core/internal/FetchDataFlow.kt
```diff
@@ -193,3 +194,37 @@ class FetchTransactionsFlow(requests: Set<SecureHash>, otherSide: FlowSession) :
...
+    override fun convert(wire: SignedDataWithCert<NetworkParameters>): SignedParametersByHash = SignedParametersByHash(wire)
+
+    override fun maybeWriteToDisk(downloaded: List<SignedParametersByHash>) {
+        for (parameters in downloaded) {
+            with(serviceHub.networkParametersStorage as NetworkParametersStorageInternal) {
+                if (!hasParameters(parameters.id)) {
+                    // This will perform the signature check too and throws with SignatureVerificationException
+                    saveParameters(parameters.signedParameters)
+                } else {
+                    logger.debug{ "Network parameters ${parameters.id} already exists in storage, skipping." }
```
rick-r3: add a space between `debug` and `{`

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,75 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
-        result.forEach {
+        result.forEach { stx ->
             // For each transaction, verify it and insert it into the database. As we are iterating over them in a
             // depth-first order, we should not encounter any verification failures due to missing data. If we fail
             // half way through, it's no big deal, although it might result in us attempting to re-download data
             // redundantly next time we attempt verification.
-            it.verify(serviceHub)
-            serviceHub.recordTransactions(StatesToRecord.NONE, listOf(it))
+
+            if (minimumPlatformVersion >= 4) {
```
rick-r3: Could move the body of this `if` into a function. e.g. `checkNetworkParametersOrdered`.  Like we did for `fetchMissingParameters`.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,79 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
         for (pageNumber in 0..(txHashes.size - 1) / RESOLUTION_PAGE_SIZE) {
             val page = page(pageNumber, RESOLUTION_PAGE_SIZE)
 
             newTxns += downloadDependencies(page)
-            val txsWithMissingAttachments = if (pageNumber == 0) signedTransaction?.let { newTxns + it } ?: newTxns else newTxns
+            val txsWithMissingAttachments = if (pageNumber == 0) signedTransaction?.let { newTxns + it }
+                    ?: newTxns else newTxns
             fetchMissingAttachments(txsWithMissingAttachments)
+            // Fetch missing parameters flow was added in version 4.
+            if (minimumPlatformVersion >= 4) {
```
shamsasari: Isn't this turning off the network parameters validation for the entire network until it's on at least 4?

I don't think we need any other type of check here either because if we receive a tx from an old node then it won't have the `PARAMETERS_GROUP` component group, we'll receive it with `networkParametersHash` as null and thus the check in `fetchMissingParameters` means we won't do the parameters fetching.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,79 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
+            } else {
+                stx.networkParametersHash ?: defaultHash
+            }
+        }
+        for (txId in txHashes) {
+            stxRootParametersMap.computeIfAbsent(txId) { mutableSetOf() }.add(rootParametersHash)
         }
     }
 
+    private fun SignedTransaction.getHashOrDefault(): SecureHash = networkParametersHash
```
shamsasari: `netParamsHashOrDefault`

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -146,7 +200,17 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
 
             // Add all input states and reference input states to the work queue.
-            val inputHashes = downloads.flatMap { it.inputs + it.references }.map { it.txhash }
+            val inputHashes = mutableListOf<SecureHash>()
+            for (stx in downloads) {
+                val rootHash = stx.getHashOrDefault()
+                val depsIds = (stx.inputs + stx.references).map { it.txhash }
+                // For all inputs and references transaction hashes record the network parameters hash that is parent in the graph.
+                depsIds.forEach { depId ->
+                    stxRootParametersMap.computeIfAbsent(depId) { mutableSetOf() }.add(rootHash)
```
shamsasari: I think this logic is wrong. According to the comment `stxRootParametersMap` is a map of `tx id -> set of net param hashes of parents`. If that's the case then `depId` is the parent and we should be mapping `stx.id` to net param hashes of the the `depIds`.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,79 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
+        // We should have all dependent parameters resolved at this point using FetchParametersFlow, if not we fallback to network map.
+        // Check that epochs are ordered in the transaction chain.
+        with(serviceHub.networkParametersStorage as NetworkParametersStorageInternal) {
+            val currentEpoch = stx.getHashOrDefault().let { hash ->
+                getEpochFromHash(hash) ?: throw IllegalArgumentException("Couldn't find root parameters epoch with hash $hash")
+            }
+            stxRootParametersMap[stx.id]?.forEach { parentHash ->
+                val parentEpoch = getEpochFromHash(parentHash)
+                        ?: throw IllegalArgumentException("Couldn't find root parameters epoch with hash $parentHash")
+                if (currentEpoch > parentEpoch) {
```
shamsasari: Shouldn't we be checking that the epoch of the parent tx isn't greater than the child's?

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -29,6 +30,12 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
 
     // Need it ordered in terms of iteration. Needs to be a variable for the check-pointing logic to work.
     private val txHashes = txHashesArg.toList()
+    // Hash map of transaction id and set of all parent network parameters hashes it is direct child of.
+    // This is to prevent downgrade attacks, where we obtain transaction chain with network parameters out of order.
+    // We need to make an assumption that parameters epoch can only increase in the dependency chain and verify that.
+    private val stxRootParametersMap = HashMap<SecureHash, MutableSet<SecureHash>>()
```
shamsasari: I think there's something wrong with the usage of `stxRootParametersMap` (see comments below). It would be conceptually easier if this was a `HashMap<SecureHash, MutableSet<Int>>`, i.e. a map of tx id to set of parent epochs.

core/src/test/kotlin/net/corda/core/internal/ParametersResoultionTest.kt
```diff
@@ -0,0 +1,224 @@
...
+    @Test
+    fun `resolve with current`() {
+        // Epoch 2 and 4 but current is 3
+        val currentHash = megaCorpNode.services.networkParametersStorage.currentHash
+        val (stx1, stx2) = makeTransactions(params2, params4)
+        // This version of flow will use current parameters for the network in root node.
+        val p = ResolveTransactionsFlowTest.TestFlow(setOf(stx1.id, stx2.id), megaCorp)
+        val future = miniCorpNode.startFlow(p)
+        mockNet.runNetwork()
+        Assertions.assertThatIllegalArgumentException().isThrownBy {
```
shamsasari: static import

core/src/main/kotlin/net/corda/core/internal/FetchDataFlow.kt
```diff
@@ -51,6 +50,8 @@ sealed class FetchDataFlow<T : NamedByHash, in W : Any>(
 
     class HashNotFound(val requested: SecureHash) : FlowException()
 
+    class MissingNetworkParameters(val requested: SecureHash) : FlowException()
```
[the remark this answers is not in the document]
josecoll: I would say yes to differentiate between different types of hash referenced types: General attachments, Contract Attachments, Network Parameters.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
josecoll: I agree the logic looks rather odd.
Should we not just be storing the stx.id -> stx.net param hash, as per the description of the map data structure: "Hash map of transaction id and set of all parent network parameters hashes it is direct child of"
(why reference the inputs? these are StateRef hashes, not txn hashes)

```
            for (stx in downloads) {
                check(resultQ.putIfAbsent(stx.id, stx) == null)   // Assert checks the filter at the start.

                // Record network parameter hash for each downloaded stx
                stxRootParametersMap.computeIfAbsent(stx.id) { mutableSetOf() }.add(stx.netParamsHashOrDefault())
            }
```

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
josecoll: I agree - looks like it is the wrong way around.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
josecoll: Agree - this will be clearer.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
josecoll: agree - I don't think we need this check.
@rick-r3 ?

josecoll:
@rick-r3 Please take a look at the last commit with changes discussed.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,89 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
+        // We should have all dependent parameters resolved at this point using FetchParametersFlow, if not we fallback to network map.
+        // Check that epochs are ordered in the transaction chain.
+        with(serviceHub.networkParametersStorage as NetworkParametersStorageInternal) {
+            // only check txn if tagged with NP's and
+            // NP's specifies a minPlatformVersion >=4 (otherwise we can assume it was sent from a pre-Corda 4 node)
+            stx.networkParametersHash?.let { hash ->
+                val stxnNetworkParams = lookup(hash) ?: throw IllegalStateException("Unable to lookup network parameters for hash $hash")
+                if (stxnNetworkParams.minimumPlatformVersion >= 4) {
+                    val stxnEpoch = stxnNetworkParams.epoch
+                    stxRootParametersMap[stx.id]?.forEach { parentHash ->
```
rick-r3: I feel like it would be safer to just make sure there was an entry for every stx.  Otherwise we haven't prepared the data correctly.  Remove the `?`

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -146,7 +210,21 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
+            val inputHashes = mutableListOf<SecureHash>()
+            for (stx in downloads) {
+                val rootHash = stx.networkParametersHash
+                // only Corda 4+ txns will have NP's
+                rootHash?.let {
+                    val depsIds = (stx.inputs + stx.references).map { it.txhash }
+                    // For all inputs and references transaction hashes record the network parameters hash that is parent in the graph.
+                    // TODO: is this logic correct? Should we not be using the NP's of the original depIds txns (not the NP's of the stx)
+                    depsIds.forEach { depId ->
+                        stxRootParametersMap.computeIfAbsent(depId) { mutableSetOf() }.add(rootHash)
```
rick-r3: Should be the network params hash of the input/reference transactions, plus the network params hash of the current transaction (just to cover the case of zero inputs).

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -146,7 +210,21 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
             // Add all input states and reference input states to the work queue.
-            val inputHashes = downloads.flatMap { it.inputs + it.references }.map { it.txhash }
+            val inputHashes = mutableListOf<SecureHash>()
+            for (stx in downloads) {
+                val rootHash = stx.networkParametersHash
+                // only Corda 4+ txns will have NP's
+                rootHash?.let {
+                    val depsIds = (stx.inputs + stx.references).map { it.txhash }
+                    // For all inputs and references transaction hashes record the network parameters hash that is parent in the graph.
+                    // TODO: is this logic correct? Should we not be using the NP's of the original depIds txns (not the NP's of the stx)
```
rick-r3: This TODO is basically correct.  See comment below.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,89 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
             // half way through, it's no big deal, although it might result in us attempting to re-download data
             // redundantly next time we attempt verification.
-            it.verify(serviceHub)
-            serviceHub.recordTransactions(StatesToRecord.NONE, listOf(it))
+
+            // Considerations in checking ordering:
+            // C4 Node => C4 Node (valid checking)
+            // C3 Node => C4 Node (checking not required as C3 Node txns will not be tagged
+            // C4 Node => C3 Node (no checking as C3 codebase will not know about NP tagging, but here is where a malicious downgrade attack could be attempted by the C4 sender)
+            checkNetworkParametersOrdered(stx)
```
rick-r3: Actually, I hadn't (when I reviewed this file) spotted that you'd moved a check into verify due to ordering of the source code files in github.  Is this even needed now (and thus can ignore most of my comments on this file), and delete lots of logic in here ?

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -162,6 +169,20 @@ private constructor(
         }
     }
 
+    private fun verifyNetworkParameters() {
+        if (networkParameters != null) {
+            networkParametersForInputs?.forEach { inputStateRef, inputNetworkParameters ->
```
rick-r3: Perhaps worth a comment or rename of `resolveNetworkParametersForInputs` to include reference state inputs.

rick-r3: We should check that the set of all `networkParametersForInputs` keys is the same as the combined set of inputs and reference state inputs, provided `networkParametersForInputs` is not null.  i.e. it should be missing, or complete, but not partial.

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -176,6 +180,8 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
 
         val resolvedNetworkParameters = resolveParameters(networkParametersHash) ?: throw TransactionResolutionException(id)
 
+        val resolveNetworkParametersForInputs = (inputs + references).map { resolveNetworkParameters(it) }.mapNotNull { it }.toMap()
```
rick-r3: Group by transaction ID first, and only fetch each transaction once, since all inputs on the same transaction share the same network params.  I'd rather put the burden here (on the grouping, optimisation) so that the verify code is simpler.

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,89 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
+
+            stx.verify(serviceHub)
+            serviceHub.recordTransactions(StatesToRecord.NONE, listOf(stx))
+        }
+    }
+
+    private fun checkNetworkParametersOrdered(stx: SignedTransaction) {
```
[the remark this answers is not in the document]
rick-r3: Ignore this.  I later discovered that we do populate an entry for the inputs, even if they were not fetched.  It's populated wrong, but I commented on those lines of code.

adagys:
Just thinking if we say have this scenario:
`T1(v4, e3) -> T2(v3, e0) -> T3(v4, e2)`

`T1(v4, e3)` = Transaction 1, created with platform version 4, network parameter epoch 3

Let's say T1 is an issue transaction created on a version 4 node, with network parameters epoch 3. Now the node builds a transaction together with an older node (runnig version 3), but the version 3 node is the coordinator, so a new valid transaction T2 is built with epoch 0 (null). This node then transacts with another version 4 node (the latter is then the coordinator), and T3 is created. The version 4 node can set any parameters it wants, since the counterparty doesn't check it. So it picks epoch 2.

Do we consider this a downgrade attack we should guard against?

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -74,29 +78,89 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
...
+            // only check txn if tagged with NP's and
+            // NP's specifies a minPlatformVersion >=4 (otherwise we can assume it was sent from a pre-Corda 4 node)
+            stx.networkParametersHash?.let { hash ->
+                val stxnNetworkParams = lookup(hash) ?: throw IllegalStateException("Unable to lookup network parameters for hash $hash")
+                if (stxnNetworkParams.minimumPlatformVersion >= 4) {
+                    val stxnEpoch = stxnNetworkParams.epoch
+                    stxRootParametersMap[stx.id]?.forEach { parentHash ->
+                        val parentEpoch = getEpochFromHash(parentHash)
+                                ?: throw IllegalArgumentException("Couldn't find root parameters epoch with hash $parentHash")
+                        if (stxnEpoch > parentEpoch) {
```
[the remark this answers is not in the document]
kasiastreich: This condition is correct, and it's because how the hash map is constructed - parentEpoch is epoch of consuming transaction  (you also commented on other lines here, so I assume it is a little unexpected/needs better documentation).

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -140,6 +148,8 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
             if (stx == null) {
                 currentHash
             } else {
+                // TODO: defaultHash should refer to a genesis hash (not currentHash)
```
kasiastreich: This TODO isn't needed here, it was already captured in `DBNetworkParametersStorage`

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -146,7 +210,21 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
                 check(resultQ.putIfAbsent(stx.id, stx) == null)   // Assert checks the filter at the start.
 
             // Add all input states and reference input states to the work queue.
-            val inputHashes = downloads.flatMap { it.inputs + it.references }.map { it.txhash }
+            val inputHashes = mutableListOf<SecureHash>()
+            for (stx in downloads) {
+                val rootHash = stx.networkParametersHash
+                // only Corda 4+ txns will have NP's
+                rootHash?.let {
```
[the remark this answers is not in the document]
kasiastreich: yeah, this refactor is wrong

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
kasiastreich: This hash map is constructed in other direction (that's also why code where you commented on the line that check > is wrong, actually works ;).

core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
josecoll: Please leave it until we actually fix this properly.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -162,6 +169,20 @@ private constructor(
         }
     }
 
+    private fun verifyNetworkParameters() {
```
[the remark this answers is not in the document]
kasiastreich: What with NotaryChangeWireTransaction, ContractUpgradeWireTransaction?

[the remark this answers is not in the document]
adagys: they have equivalent methods, `resolve` i think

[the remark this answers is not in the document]
kasiastreich: Yeah, but then it's repeated.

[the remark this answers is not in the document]
adagys: yep, but they're quite repeated already anyway, could have a helper util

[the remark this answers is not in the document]
adagys: we'd need to repeat this logic on other ledger transaction types anyway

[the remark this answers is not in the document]
kasiastreich: I am now digging into this refactor. I am not sure if I 100% agree with it. And as you noticed, it will be repeated for all types of transactions as resolve/verification differs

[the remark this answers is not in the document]
josecoll: @kasiastreich talk to Rick as this was discussed and agreed as correct.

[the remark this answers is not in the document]
adagys: yeah I'm not sure whats the best approach, still a bit confused whether we should do it on resolution, on verify or on both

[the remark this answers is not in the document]
kasiastreich: @josecoll what is implemented now is entirely correct. It doesn't take into account other types of wire transactions, also it's not nicely extensible

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -162,6 +169,20 @@ private constructor(
...
+        if (networkParameters != null) {
```
kasiastreich: This also isn't entirely correct, if it's null the downgrade check should still be there, but with default parameters.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -162,6 +169,20 @@ private constructor(
...
+            networkParametersForInputs?.forEach { inputStateRef, inputNetworkParameters ->
+                if (inputNetworkParameters == null) {
```
kasiastreich: Also not correct, this can be maliciously constructed this way.

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
```diff
@@ -171,13 +171,30 @@ data class SignedTransaction(val txBits: SerializedBytes<CoreTransaction>,
     @DeleteForDJVM
     @Throws(SignatureException::class, AttachmentResolutionException::class, TransactionResolutionException::class, TransactionVerificationException::class)
     fun verify(services: ServiceHub, checkSufficientSignatures: Boolean = true) {
+        resolveAndCheckNetworkParameters(services)
```
adagys: hmm didn't we discuss that it's sufficient to check the ordering during transaction resolution only? This verify is also sometimes called after building a transaction. Or does caching make this quite cheap?

kasiastreich: ok, I moved it **back** to resolve transactions flow

kasiastreich:
@shamsasari  could you take a look please?

core/src/main/kotlin/net/corda/core/internal/DigitalSignatureWithCert.kt
```diff
@@ -43,7 +45,9 @@ Cert path: $fullCertPath
 
 /** Similar to [SignedData] but instead of just attaching the public key, the certificate for the key is attached instead. */
 @CordaSerializable
-class SignedDataWithCert<T : Any>(val raw: SerializedBytes<T>, val sig: DigitalSignatureWithCert) {
+class SignedDataWithCert<T : Any>(val raw: SerializedBytes<T>, val sig: DigitalSignatureWithCert): NamedByHash {
+    override val id: SecureHash = raw.hash
```
shamsasari: Use `get`

core/src/main/kotlin/net/corda/core/internal/NetworkParametersServiceInternal.kt
```diff
@@ -1,12 +1,40 @@
...
+     */
+    fun hasParameters(hash: SecureHash): Boolean
+
     /**
      * Returns the [NotaryInfo] for a notary [party] in the current or any historic network parameter whitelist, or null if not found.
      */
     fun getHistoricNotary(party: Party): NotaryInfo?
 }
+
+interface NetworkParametersStorage : NetworkParametersServiceInternal {
```
shamsasari: Why do you need the separate interface? Have these methods in `NetworkParametersServiceInternal` should be fine.

kasiastreich: It is after merge. For some reason we had two interfaces in master too. I will refactor this code.

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
adagys: Okay, I talked to Rick and thought about it some more, we do need to also check for param downgrade when building a transaction (as you state in the PR description). So your previous version was correct, sorry!

There's one catch that for notary change and contract upgrade verification is done via stx.verify(), but for regular transactions in the finality flow it's "resolve ltx, ltx.verify()", so I think it would be bypassed.

The other thing that needs doing is testing the performance impact of this, and making sure this param resolution & checking hits caches.

core/src/main/kotlin/net/corda/core/transactions/SignedTransaction.kt
```diff
@@ -157,6 +157,8 @@ data class SignedTransaction(val txBits: SerializedBytes<CoreTransaction>,
         } else {
             checkSignaturesAreValid()
         }
+        // We need parameters check here, because finality flow calls stx.toLedgerTransaction() and then verify.
+        resolveAndCheckNetworkParameters(services)
```
adagys: are there any uses of wtx.toLedgerTransaction + .verify that would bypass this?

PR #4354, opened 2018-12-04
[ENT-2821] Make the default schemes for TLS and identity available from the CryptoService interface

florian-f:
https://r3-cev.atlassian.net/browse/ENT-2821


florian-f:
Right. We could make it return an Int instead, since that's what generateKeyPair accepts anyways. Actually, I think I would have preferred to have `fun generateKeyPair(alias: String, scheme: SignatureScheme)`

kchalkias:
I completely agree, if we still have dependency to corda core anyway, let's better change the signature of `generateKeyPair`. The original idea is it should be in its own module, that;s why it is still an `Int`.

PR #4356, opened 2018-12-04
Fix smoke tests after merge of notary whitelist verification

thschroeter:
A notary party is required in the network parameters after the merge of https://github.com/corda/corda/pull/4346.

### DIFF
--- a/testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
+++ b/testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
@@ -3,9 +3,12 @@ package net.corda.smoketesting
 import net.corda.client.rpc.CordaRPCClient
 import net.corda.client.rpc.CordaRPCConnection
 import net.corda.client.rpc.internal.serialization.amqp.AMQPClientSerializationScheme
+import net.corda.core.identity.Party
 import net.corda.core.internal.*
+import net.corda.core.node.NotaryInfo
 import net.corda.core.utilities.NetworkHostAndPort
 import net.corda.core.utilities.contextLogger
+import net.corda.nodeapi.internal.DevIdentityGenerator
 import net.corda.nodeapi.internal.network.NetworkParametersCopier
 import net.corda.testing.common.internal.asContextEnv
 import net.corda.testing.common.internal.checkNotOnClasspath
@@ -59,11 +62,11 @@ class NodeProcess(
         private companion object {
             val javaPath: Path = Paths.get(System.getProperty("java.home"), "bin", "java")
             val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmss").withZone(systemDefault())
-            val defaultNetworkParameters = run {
+            fun defaultNetworkParameters(notaryInfo: NotaryInfo) = run {
                 AMQPClientSerializationScheme.createSerializationEnv().asContextEnv {
                     // TODO There are no notaries in the network parameters for smoke test nodes. If this is required then we would
                     // need to introduce the concept of a "network" which predefines the notaries, like the driver and MockNetwork
-                    NetworkParametersCopier(testNetworkParameters())
+                    NetworkParametersCopier(testNetworkParameters(notaries = listOf(notaryInfo)))
                 }
             }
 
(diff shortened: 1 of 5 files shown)


testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
```diff
@@ -76,14 +79,33 @@ class NodeProcess(
 
         private val nodesDirectory = (buildDirectory / formatter.format(Instant.now())).createDirectories()
 
+        private var notaryParty: Party? = null
+
         fun baseDirectory(config: NodeConfig): Path = nodesDirectory / config.commonName
 
+        fun createNotary(config: NodeConfig): NodeProcess {
```
shamsasari: This method is almost identity to the other one. Why not simply use `config.isNotary` to determine what to do and merge the methods together?

testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
```diff
@@ -59,11 +62,11 @@ class NodeProcess(
         private companion object {
             val javaPath: Path = Paths.get(System.getProperty("java.home"), "bin", "java")
             val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmss").withZone(systemDefault())
-            val defaultNetworkParameters = run {
+            fun defaultNetworkParameters(notaryInfo: NotaryInfo) = run {
```
shamsasari: rename to `createNetworkParameter`, and get rid of the `run`

testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
```diff
@@ -59,11 +62,11 @@ class NodeProcess(
...
                 AMQPClientSerializationScheme.createSerializationEnv().asContextEnv {
                     // TODO There are no notaries in the network parameters for smoke test nodes. If this is required then we would
```
shamsasari: Remove this TODO

testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
shamsasari: You still need to check that `config.isNotary` is true, and in `create` the opposite. Then you might as well not have a separate `createNotary` method.

thschroeter:
@shamsasari, not elegant but passes "build branches" in CI, see https://ci-master.corda.r3cev.com/viewLog.html?buildId=129956&buildTypeId=Corda_Branches.

testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
```diff
@@ -58,32 +62,46 @@ class NodeProcess(
...
-
+            val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss.SSS").withZone(systemDefault())
             init {
                 checkNotOnClasspath("net.corda.node.Corda") {
                     "Smoke test has the node in its classpath. Please remove the offending dependency."
                 }
             }
         }
 
+        var networkParametersCopier: NetworkParametersCopier? = null
```
vkolomeyko: can we use `private lateinit` modifier here?

testing/smoke-test-utils/src/main/kotlin/net/corda/smoketesting/NodeProcess.kt
```diff
@@ -58,32 +62,46 @@ class NodeProcess(
...
+
         private val nodesDirectory = (buildDirectory / formatter.format(Instant.now())).createDirectories()
 
+        private var notaryParty: Party? = null
```
vkolomeyko: Can we use `lateinit` modifier here?
(to ensure the variable assigned once)

thschroeter: I prefer leaving the notaryParty nullable, because we have to make sure it is set when creating nodes and it can't be set multiple times. The user has to call create first with a notary config, and has the ability to call create multiple times to add nodes to the network, as discussed with Shams.

PR #4366, opened 2018-12-05
[CORDA-2247]: Fixed a bug causing Hibernate to produce cross joins.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2247

Notes:

1. Fixed a problem that was preventing Hibernate from generating left joins instead of cross joins.
2. Added an additional sort column on `stateRef` to avoid unstable sorts on fields with possible duplicate values. This is added only if no sort columns are based on `stateRef` already.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -20,7 +20,7 @@ Unreleased
 
   The ``confidential-identities`` dependency in your CorDapp must now be ``compile`` and not ``cordaCompile``.
 
-* Fixed a bug causing Hibernate to generate cross join statements. This was evident in poor performance and incorrect results for vault queries with sorting.
+* Fixed a bug resulting in poor vault query performance and incorrect results when sorting.
 
 * Marked the ``Attachment`` interface as ``@DoNotImplement`` because it is not meant to be extended by CorDapp developers. If you have already
   done so, please get in contact on the usual communication channels.
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -20,6 +20,8 @@ Unreleased
 
   The ``confidential-identities`` dependency in your CorDapp must now be ``compile`` and not ``cordaCompile``.
 
+* Fixed a bug causing Hibernate to generate cross join statements. This was evident in poor performance and incorrect results for vault queries with sorting.
+
 * Marked the ``Attachment`` interface as ``@DoNotImplement`` because it is not meant to be extended by CorDapp developers. If you have already
   done so, please get in contact on the usual communication channels.
 
--- a/node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
+++ b/node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
@@ -649,7 +649,6 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
             when (direction) {
                 Sort.Direction.ASC -> {
                     if (entityStateAttributeChild != null) {
-                        // TODO sollecitom add asc/desc by stateRef, if not there already
                         orderCriteria.add(criteriaBuilder.asc(sortEntityRoot.get<String>(entityStateAttributeParent).get<String>(entityStateAttributeChild)))
                         orderCriteria.add(criteriaBuilder.asc(sortEntityRoot.get<String>(PersistentState::stateRef.name)))
                     }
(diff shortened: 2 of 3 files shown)


docs/source/changelog.rst
```diff
@@ -20,6 +20,8 @@ Unreleased
 
   The ``confidential-identities`` dependency in your CorDapp must now be ``compile`` and not ``cordaCompile``.
 
+* Fixed a bug causing Hibernate to generate cross join statements. This was evident in poor performance and incorrect results for vault queries with sorting.
```
josecoll: Why not just state what you fixed? (this is a specific fix to an identified issue).
Hibernate cross join's are fine in most circumstances.

node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
```diff
@@ -622,11 +624,16 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
         return emptySet()
     }
 
-    private fun parse(sorting: Sort) {
-        log.trace { "Parsing sorting specification: $sorting" }
+    private fun parse(sortingArg: Sort) {
```
josecoll: no need to rename this variable (use a different one in your new logic).

node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
```diff
@@ -638,24 +645,31 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
...
+                        joinPredicates.add(joinPredicate)
                         entityRoot
                     }
             when (direction) {
                 Sort.Direction.ASC -> {
-                    if (entityStateAttributeChild != null)
+                    if (entityStateAttributeChild != null) {
                         orderCriteria.add(criteriaBuilder.asc(sortEntityRoot.get<String>(entityStateAttributeParent).get<String>(entityStateAttributeChild)))
-                    else
+                    }
```
josecoll: remove braces for single-line if/else statements.

node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
```diff
@@ -622,11 +624,16 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
...
     }
 
-    private fun parse(sorting: Sort) {
-        log.trace { "Parsing sorting specification: $sorting" }
+    private fun parse(sortingArg: Sort) {
+        log.trace { "Parsing sorting specification: $sortingArg" }
 
         val orderCriteria = mutableListOf<Order>()
 
+        val sorting = if (sortingArg.columns.none { it.sortAttribute == SortAttribute.Standard(Sort.CommonStateAttribute.STATE_REF) }) {
```
josecoll: why are we force sorting here?

docs/source/changelog.rst
sollecitom: Fixed

node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
sollecitom: If by force sorting you mean adding a sort criteria on the primary key, if not there already, then it's because otherwise sorting is unstable. As an example, if you have 50 states all with 100 as value for the same field, and you sort by that field, then pagination might trip, because when you sort for each page, the order might be different, resulting in states appearing multiple times or never.

josecoll: OK - thanks for explanation.

docs/source/changelog.rst
```diff
@@ -20,6 +20,8 @@ Unreleased
 
   The ``confidential-identities`` dependency in your CorDapp must now be ``compile`` and not ``cordaCompile``.
 
+* Fixed a bug resulting in poor vault query performance and incorrect results when sorting.
```
josecoll: This should state that you have "fixed a pagination bug manifested with sorting".
It doesn't specifically fix poor query performance.

josecoll: Or just reference the Issue description:
"VaultService queryBy returned duplicate rows in StateMetadata"

PR #4372, opened 2018-12-06
[CORDA-1676] Adjust progress tracking rendering for skipped elements

dimosr:
## Changes
I adjusted the rendering for flow stages that have been skipped, so that user is not misled into thinking they failed (by the X symbol). 

Notes:
* I initially went for strikethrough, but realised it's not working for most terminals, so went for the faint effect. Also, could not find a more appropriate symbol, so used the ballot box check. Open to suggestions.
* I took the liberty of removing the `treeIndex == tree.lastIndex` condition, since I believe it's dead code and I think it's also confusing. The reason is it's comparing the index of the steps in progress with the `tree.lastIndex`, which corresponds to the number of times the structure of the progress tracker has changed. I run a few more tests, as shown below to make sure this does not introduce regressions. Let me know if I am missing something.

## Testing
* Run the `heartbeat`, `bank-of-corda-demo`, `trader-demo` projects and the output is the same (content-wise)
* The style of the new output can be seen below:
success

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/Emoji.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/Emoji.kt
@@ -54,6 +54,8 @@ object Emoji {
     val CODE_WARNING_SIGN: String = codePointsString(0x26A0, 0xFE0F)
     @JvmStatic
     val CROSS_MARK_BUTTON: String = codePointsString(0x274E)
+    @JvmStatic
+    val BALLOT_BOX_WITH_CHECK_BUTTON: String = codePointsString(0x2611)
 
     /**
      * When non-null, toString() methods are allowed to use emoji in the output as we're going to render them to a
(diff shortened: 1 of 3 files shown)


shamsasari:
The ballotbox tick is a bit misleading. I think it'll be clearer without it.

shamsasari:
I'm going to had this over to @mikehearn to review in case he has better ideas.

dimosr:
> The ballotbox tick is a bit misleading. I think it'll be clearer without it.

I agree, we could remove it and stick with only the faint effect.

mikehearn:
Thanks!

Hmm, from the code I can't see why a tickbox still appears next to the faint text? It looks weird to have two kinds of tickbox, one which is emoji and one which isn't.

dimosr:
> Thanks!
> 
> Hmm, from the code I can't see why a tickbox still appears next to the faint text? It looks weird to have two kinds of tickbox, one which is emoji and one which isn't.

Sorry my bad, forgot to update the image. Here's the new look.
without_check

PR #4387, opened 2018-12-07
CORDA-2227 Emit java.lang.Object, not *, in TypeNotation

distributedleetravis:
If a type parameter is `*`, we should write it as `java.lang.Object`, so that Corda V3 can interpret it correctly.

### DIFF
--- a/serialization/src/test/kotlin/net/corda/serialization/internal/amqp/AMQPRemoteTypeModelTests.kt
+++ b/serialization/src/test/kotlin/net/corda/serialization/internal/amqp/AMQPRemoteTypeModelTests.kt
@@ -45,10 +45,10 @@ class AMQPRemoteTypeModelTests {
     @Test
     fun `round-trip some types through AMQP serialisations`() {
         arrayOf("").assertRemoteType("String[]")
-        listOf(1).assertRemoteType("List<*>")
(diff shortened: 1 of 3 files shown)


serialization/src/test/kotlin/net/corda/serialization/internal/amqp/AMQPRemoteTypeModelTests.kt
```diff
@@ -45,10 +45,10 @@ class AMQPRemoteTypeModelTests {
     @Test
     fun `round-trip some types through AMQP serialisations`() {
         arrayOf("").assertRemoteType("String[]")
-        listOf(1).assertRemoteType("List<*>")
+        listOf(1).assertRemoteType("List<Object>")
```
shamsasari: Does this mean we're no longer matching the V3 fingerprint of `List`?

distributedleetravis: No, it _should_ be `java.lang.Object` (which the pretty-printer used in this test abbreviates to `Object`, but it's `java.lang.Object` that's actually in the type notation sent over the wire). V3 can't interpret `*`.

PR #4390, opened 2018-12-10
[CORDA-2303]: Create seperate webserver caplet

anthonykeenan:
The corda caplet no longer adds jar dirs to the classpath, but the webserver requires it, so taking a copy of the caplet just for the webserver as further changes are likely to the node caplet.

### DIFF
--- a/webserver/build.gradle
+++ b/webserver/build.gradle
@@ -55,6 +55,10 @@ dependencies {
     // For rendering the index page.
     compile "org.jetbrains.kotlinx:kotlinx-html-jvm:0.6.3"
 
+    // Capsule is a library for building independently executable fat JARs.
+    // We only need this dependency to compile our Caplet against.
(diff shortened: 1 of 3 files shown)


josecoll:
Also requires a changelog entry.

anthonykeenan:
@josecoll You can't do that without some bigger changes. The `CordaCapsule` code is in the `node` module, and the webserver doesn't depend on that. It actually just copies the compiled `CordaCapsule.class` file into it's jar. So in order to depend on the existing one, you'd have to either make the webserver depend on the node (which would add a lot of cruft to access one file), or move the caplet into the `core` module. However, I think an easier way is to add the jars to the `ClassLoader` at runtime, the same way the node now does. I'm just working on that now.

anthonykeenan:
Additionally, this is just fixing behaviour that has broken, but hasn't changed since the last release. I don't believe it warrants an entry in the changelog?

PR #4392, opened 2018-12-10
[CORDA-2007]: Fixed problematic max page size.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2007


core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteriaUtils.kt
```diff
@@ -140,7 +140,7 @@ const val DEFAULT_PAGE_SIZE = 200
 /**
  * Note: use [PageSpecification] to correctly handle a number of bounded pages of a pre-configured page size.
  */
-const val MAX_PAGE_SIZE = Int.MAX_VALUE
+const val MAX_PAGE_SIZE = Int.MAX_VALUE - 1
```
mikehearn: Could you add in a comment here why there's a `- 1`?

sollecitom: I will

PR #4393, opened 2018-12-10
Reverted incorrect deprecation of JacksonSupport.createInMemoryMapper

shamsasari:


docs/source/changelog.rst
```diff
@@ -206,29 +208,11 @@ Unreleased
 
 * Changes to the JSON/YAML serialisation format from ``JacksonSupport``, which also applies to the node shell:
 
-  * ``Instant`` and ``Date`` objects are serialised as ISO-8601 formatted strings rather than timestamps
```
shamsasari: These changes were ported to V3 and so are no longer relevant to V4

docs/source/changelog.rst
```diff
@@ -7,6 +7,8 @@ release, see :doc:`upgrade-notes`.
 Unreleased
 ----------
 
+* ``JacksonSupport.createInMemoryMapper`` was incorrectly marked as deprecated and is no longer so.
```
mikehearn: Was it marked as deprecated in the 3.x cycle?

shamsasari: yes

shamsasari: I think what happened was that the changes were backported into V3 to get the blob inspector out in 3.3/3.2.

PR #4394, opened 2018-12-10
[CORDA-2162]: `Cash.generateSpend` cannot be used twice to generate two cash moves in the same tx (fix).

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2162

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
@@ -628,7 +628,7 @@ with @BelongsToContract, or supply an explicit contract parameter to addOutputSt
     fun outputStates(): List<TransactionState<*>> = ArrayList(outputs)
 
     /** Returns an immutable list of [Command]s, grouping by [CommandData] and joining signers. */
-    fun commands(): List<Command<*>> = commands.groupBy { cmd -> cmd.value }.entries.map { (data, cmds) -> Command(data, cmds.flatMap(Command<*>::signers)) }
+    fun commands(): List<Command<*>> = commands.groupBy { cmd -> cmd.value }.entries.map { (data, cmds) -> Command(data, cmds.flatMap(Command<*>::signers).toSet().toList()) }
 
     /**
      * Sign the built transaction and return it. This is an internal function for use by the service hub, please use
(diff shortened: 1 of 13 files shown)


core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -64,13 +64,15 @@ open class TransactionBuilder @JvmOverloads constructor(
     /**
      * Creates a copy of the builder.
      */
-    fun copy(): TransactionBuilder {
+    fun copy(): TransactionBuilder = copy(includeCommand = { true })
+
+    fun copy(includeCommand: (command: Command<*>) -> Boolean = { true }): TransactionBuilder {
```
mikehearn: Adding new public API like this **requires** docs - and best to avoid Kotlin specific type signatures and default arguments!

For such a specific thing, it's probably better to just copy the TransactionBuilder in the cash code rather than add a new public API that's special cased for commands.

sollecitom: I'm removing the filter, but I cannot do it outside. It's not only about Cash, and I can't access some private variables from the outside.

sollecitom:
Unfortunately I don't think this can be fixed without (non-breaking) API changes.

One option would be waiting to introduce a relationship between states and commands.

finance/src/test/kotlin/net/corda/finance/contracts/asset/CashTests.kt
```diff
@@ -269,8 +272,8 @@ class CashTests {
             output(Cash.PROGRAM_ID, inState.copy(amount = inState.amount * 2))
             command(megaCorp.publicKey, Cash.Commands.Issue())
             tweak {
-                command(megaCorp.publicKey, Cash.Commands.Issue())
-                this `fails with` "there is only a single issue command"
+                command(miniCorp.publicKey, Cash.Commands.Issue())
```
tlil: Why change the participant here?

sollecitom: I planned to check that the signers included both, but there's no way to do this with this mini-DSL.
The test works regardless.

I created a new test to check signers.

PR #4398, opened 2018-12-11
CORDA-2134 - re-enabling ignored tests due to signature constrains non-downgrade rule

szymonsztuka:
* `TwoPartyTradeFlowTests` - restructured - saving transaction just after creating a state, as it's needed for the next TransactionState to be verified (for non-downgrade rule), enable more tests which were set as `@Ignored`,  left 3 TODOs as 2 tests (expectations) were slightly modified, need further investigation
* `MockService` method `loadContractAttachment` always returns a dummy Attachment which means it has a default Contract Class Version number `1` used for signature constraints non-downgrade rule.
* `TransactionSerializationTests` - added a fake transaction which created fake stated used as input state and using real (not mocked one) implementation of `loadContractAttachment`

### DIFF
--- a/node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
+++ b/node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
@@ -48,7 +48,6 @@ import net.corda.testing.node.ledger
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.After
 import org.junit.Before
-import org.junit.Ignore
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.Parameterized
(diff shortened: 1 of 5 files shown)


node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
```diff
@@ -650,11 +649,16 @@ class TwoPartyTradeFlowTests(private val anonymous: Boolean) {
         }
     }
 
-    private fun LedgerDSL<TestTransactionDSLInterpreter, TestLedgerDSLInterpreter>.fillUpForBuyer(
+    private fun LedgerDSL<TestTransactionDSLInterpreter, TestLedgerDSLInterpreter>.fillUpForBuyerAndInsertFakeTrnactions(
```
josecoll: typo.

PR #4409, opened 2018-12-13
[CORDA-2330] Turn on lenient mode in rpc serialization context.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-2330

### DIFF
--- a/serialization/src/main/kotlin/net/corda/serialization/internal/ClientContexts.kt
+++ b/serialization/src/main/kotlin/net/corda/serialization/internal/ClientContexts.kt
@@ -21,5 +21,6 @@ val AMQP_RPC_CLIENT_CONTEXT = SerializationContextImpl(
         emptyMap(),
         true,
         SerializationContext.UseCase.RPCClient,
-        null
+        null,
+        lenientCarpenterEnabled = true
 )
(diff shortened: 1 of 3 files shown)


fenryka:
As discussed with Florien, this needs more thought because this is addressing a symptom without understanding the cause

florian-f:
@fenryka The only alternative I see would be, as discussed, to change the serialization logic to not include type metadata for interfaces that cannot be carpented (e.g. have methods that don't start with "get"). This may cause problems elsewhere. I am not sufficiently familiar with our serialization engine to reason about the implications such a change may have, maybe @distributedleetravis can weigh in. 

Some observations:

If we were to decide that we don't include interface metadata when serializing, that would change the semantics of CordaRPCOps.vaultQuery. This call `rpcProxy.vaultQuery(QueryableState::class.java)` would not return any concrete `QueryableState`s that are not on the client's class path (or otherwise present in the class loader used for deserialization).  

Making the carpenter lenient would result in the above query returning states of type `QueryableState` that do not actually implement the interface -- calling for example `supportedSchemas()` would result in an AbstractMethodError. 

If we address this issue by changing the serialization logic, we will have a problem with states that were serialized by nodes running earlier versions of Corda. 

In my opinion, making the carpenter that is used by the RPC client lenient when deserializing is the better option here.

PR #4411, opened 2018-12-13
Docs: improve upgrade notes and versioning page

mikehearn:
* Adds some more information on target/min platform versioning (prior info was only the design doc)
* Add a missing design doc to the toctree
* Move the versioning page to the API section
* Add more info to the upgrade notes
* Delete old upgrade notes that are no longer relevant

### DIFF
--- a/docs/source/upgrade-notes.rst
+++ b/docs/source/upgrade-notes.rst
@@ -160,7 +160,7 @@ This is no longer necessary with ``ReceiveFinalityFlow`` and the call to ``waitF
 Step 4. Upgrade your use of SwapIdentitiesFlow
 ----------------------------------------------
 
-The ``confidential-identities`` module is experimental in Corda 3 and remains so in Corda 4. In this release, the ``SwapIdentitiesFlow``
+The :ref:`confidential_identities_ref` API is experimental in Corda 3 and remains so in Corda 4. In this release, the ``SwapIdentitiesFlow``
 has been adjusted in the same way as ``FinalityFlow`` above, to close problems with confidential identities being injectable into a node
 outside of other flow context. Old code will still work, but it is recommended to adjust your call sites so a session is passed into
 the ``SwapIdentitiesFlow``.
@@ -172,6 +172,24 @@ Step 5. Possibly, adjust unit test code
 Use a ``MockNetworkConfigOverrides`` object instead. This is an API change we regret, but unfortunately in Corda 3 we accidentally exposed
 large amounts of the node internal code through this one API entry point. We have now insulated the test API from node internals and
 reduced the exposure.
+
+If you are constructing a MockServices for testing contracts, and your contract uses the Cash contract from the finance app, you
+now need to explicitly add ``net.corda.finance.contracts`` to the list of ``cordappPackages``. This is a part of the work to disentangle
+the finance app (which is really a demo app) from the Corda internals. Example::
+
+    val ledgerServices = MockServices(
+        listOf("net.corda.examples.obligation", "net.corda.testing.contracts"),
+        identityService = makeTestIdentityService(),
+        initialIdentity = TestIdentity(CordaX500Name("TestIdentity", "", "GB"))
+    )
+
+becomes::
+
+    val ledgerServices = MockServices(
+        listOf("net.corda.examples.obligation", "net.corda.testing.contracts", "net.corda.finance.contracts"),
+        identityService = makeTestIdentityService(),
+        initialIdentity = TestIdentity(CordaX500Name("TestIdentity", "", "GB"))
+    )
 
 Step 6. Security: refactor to avoid violating sealed packages
 -------------------------------------------------------------
@@ -229,3 +247,12 @@ that causes transaction details to be converted to a PDF and sent to a particula
 into shared business logic, but it makes perfect sense to put into a user-specific app they developed themselves.
 
 If your flows could benefit from being extended in this way, read ":doc:`flow-overriding`" to learn more.
+
+Step 10. Explore other new features that may be useful
+------------------------------------------------------
+
+Corda 4 adds several new APIs that help you build applications. Why not explore:
+
+* The `new withEntityManager API <api/javadoc/net/corda/core/node/ServiceHub.html#withEntityManager-block->`_ for using JPA inside your flows and services.
+* :ref:`reference_states`, that let you use an input state without consuming it.
+* :ref:`state_pointers`, that make it easier to 'point' to one state from another and follow the latest version of a linear state.
--- a/docs/source/upgrade-notes.rst
+++ b/docs/source/upgrade-notes.rst
@@ -40,11 +40,13 @@ Step 2. Add a "cordapp" section to your Gradle build file
 This is used by the Corda Gradle build plugin to populate your app JAR with useful information. It should look like this::
 
     cordapp {
-        info {
+        targetPlatformVersion 4
+        minimumPlatformVersion 4
+        contract {
             name "MegaApp"
             vendor "MegaCorp"
-            targetPlatformVersion 4
-            minimumPlatformVersion 4
+            license "A really expensive license"
+            versionId 1
         }
     }
 
@@ -169,24 +171,11 @@ Step 5. Security: refactor to avoid violating sealed packages
 Hardly any apps will need to do anything in this step.
 
 App isolation has been improved. Version 4 of the finance CorDapp (*corda-finance.jar*) is now built as a sealed and signed JAR file.
-This means classes in your own CorDapps cannot be placed under the following packages:
+This means classes in your own CorDapps cannot be placed under the following package namespace:  ``net.corda.finance``
 
-  .. sourcecode:: java
-
-     net.corda.finance
-     net.corda.finance.contracts
-     net.corda.finance.contracts.asset.cash.selection
-     net.corda.finance.contracts.asset
-     net.corda.finance.contracts.math
-     net.corda.finance.flows
-     net.corda.finance.internal
-     net.corda.finance.plugin
-     net.corda.finance.schemas
-     net.corda.finance.utils
-
-In the unlikely event that you were injecting code into ``net.corda.*`` package namespaces from your own apps, you will need to move them
-into a new package, e.g. *net/corda/finance/flows.MyClass.java* can be moved to *net/corda/finance/flows/company/MyClass.java*.
-Also your classes are no longer able to access non-public members of finance CorDapp classes.
+In the unlikely event that you were injecting code into ``net.corda.finance.*`` package namespaces from your own apps, you will need to move them
+into a new package, e.g. ``net/corda/finance/flows/MyClass.java`` can be moved to ``com/company/corda/finance/flows/MyClass.java``.
+As a consequence your classes are no longer able to access non-public members of finance CorDapp classes.
 
 When recompiling your JARs for Corda 4, your own apps will also become sealed, meaning other JARs cannot place classes into your own packages.
 This is a security upgrade that ensures package-private visibility in Java code works correctly.
--- a/docs/source/upgrade-notes.rst
+++ b/docs/source/upgrade-notes.rst
@@ -43,10 +43,15 @@ This is used by the Corda Gradle build plugin to populate your app JAR with usef
         targetPlatformVersion 4
         minimumPlatformVersion 4
         contract {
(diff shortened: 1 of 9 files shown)


docs/source/upgrade-notes.rst
```diff
@@ -4,48 +4,66 @@
...
+---------------------------------------------------------
 
-    ext.kotlin_version = 'x.y.z'
-    ext.quasar_version = 'x.y.z'
+This is used by the Corda Gradle build plugin to populate your app JAR with useful information. It should look like this::
 
-Please consult the relevant release notes of the release in question. If not specified, you may assume the
-versions you are currently using are still in force.
+    cordapp {
+        info {
```
shamsasari: This is out-of-date. @josecoll can provide the details.

docs/source/upgrade-notes.rst
```diff
@@ -137,63 +155,21 @@ finalised transaction. If the initiator is written in a backwards compatible way
...
-  No action is needed for default node tables as ``PersistentStateRef`` is used as Primary Key only and the backing columns are automatically not nullable
-  or custom Cordapp entities using ``PersistentStateRef`` as Primary Key.
-
-* MockNetwork: ``MockNodeParameters`` and functions creating it no longer use a lambda expecting a ``NodeConfiguration``
-  object. Use a ``MockNetworkConfigOverrides`` object instead.
-
-* Finance CorDapp (*corda-finance-XXX.jar*) is now build as a sealed and signed JAR file.
-  This means classes in your CorDapps cannot be placed under the following packages:
+App isolation has been improved. Version 4 of the finance CorDapp (*corda-finance.jar*) is now built as a sealed and signed JAR file.
+This means classes in your own CorDapps cannot be placed under the following packages:
```
shamsasari: It would be simpler to say `net.corda.finance` and any sub-packages, instead of the list below.

docs/source/upgrade-notes.rst
```diff
@@ -208,616 +184,51 @@ For H2:
...
-  * ``TransactionType.General.Builder(notary)`` becomes ``TransactionBuilder(notary)``
-
-Build 
-^^^^^
-
-* Gradle dependency reference changes:
-
-  * Module names have changed to include ``corda`` in the artifacts' JAR names:
+In the unlikely event that you were injecting code into ``net.corda.*`` package namespaces from your own apps, you will need to move them
+into a new package, e.g. *net/corda/finance/flows.MyClass.java* can be moved to *net/corda/finance/flows/company/MyClass.java*.
```
shamsasari: Surely the goal is move out of the `net.corda.finance` namespace completely? The suggestion is not consistent with the statement "In the unlikely event that you were injecting code into net.corda.*"

docs/source/versioning.rst
```diff
@@ -8,15 +8,62 @@ friendly for a developer working on the platform. It first has to be parsed and
...
+is called the *platform version*. It is similar to Android's `API Level <https://developer.android.com/guide/topics/manifest/uses-sdk-element.html>`_.
 It starts at 1 and will increment by exactly 1 for each release which changes any of the publicly exposed APIs in the
 entire platform. This includes public APIs on the node itself, the RPC system, messaging, serialisation, etc. API backwards
-compatibility will always be maintained, with the use of deprecation to migrate away from old APIs. In rare situations
-APIs may have to be removed, for example due to security issues. There is no relationship between the Platform Version
-and the release version - a change in the major, minor or patch values may or may not increase the Platform Version.
+compatibility will always be maintained, with the use of deprecation to suggest migration away from old APIs. In very rare
+situations APIs may have to be changed, for example due to security issues. There is no relationship between the platform version
+and the release version - a change in the major, minor or patch values may or may not increase the platform version. However
```
shamsasari: Our versioning scheme doesn't have patch.

docs/source/versioning.rst
```diff
@@ -8,15 +8,62 @@ friendly for a developer working on the platform. It first has to be parsed and
...
\ No newline at end of file
+registers with the network map it will check its own version against the minimum version requirement for the network.
+
+Minimum platform version
+------------------------
+
+Applications can advertise a *minimum platform version* they require. If your app uses new APIs that were added in (for example) Corda 5,
+you should specify a minimum version of 5. This will ensure the app won't be loaded by older nodes. If you can *optionally* use the new
+APIs, you can keep the minimum set to a lower number. Attempting to use new APIs on older nodes can cause ``NoSuchMethodError`` exceptions
+and similar problems, so it's best to advertise the correct requirements so problems are caught up front, instead of at runtime.
```
shamsasari: I would also say that you can conditionally use new APIs if they exist by checking the node's platform version from the ServiceHub before attempting to call them.

docs/source/versioning.rst
```diff
@@ -8,15 +8,62 @@ friendly for a developer working on the platform. It first has to be parsed and
...
+For example, consider an app that uses new features introduced in Corda 4, but which has passed regression testing on Corda 5. It will
+advertise a minimum platform version of 4 and a target version of 5. These numbers are published in the JAR manifest file.
+
+If this app is loaded into a Corda 6 node, that node may implement backwards compatibility workarounds for your app that make it slower,
+less secure, or less featureful. You can opt-in to getting the full benefits of the upgrade by changing your target version to 6. By doing
+this, you promise that you understood all the changes in Corda 6 and have thoroughly tested your app to prove it works.
+
+Target versioning is one of the mechanisms we have to keep the platform evolving and improving, without being permanently constrained to
+being bug-for-bug compatible with old versions. When no apps are loaded that target old versions, any emulations of older bugs or problems
+can be disabled.
```
shamsasari: An example of this in this release is the old insecure FinalityFlow API.

docs/source/versioning.rst
```diff
@@ -8,15 +8,62 @@ friendly for a developer working on the platform. It first has to be parsed and
...
+
+Publishing versions in your JAR manifest
+----------------------------------------
+
+In your ``build.gradle`` file, add a block like this::
+
+    cordapp {
+        info {
```
shamsasari: ditto

docs/source/upgrade-notes.rst
mikehearn: Good spot. I didn't review this part carefully enough.

mikehearn:
I wonder if we can fix the finance app to have backwards compatible configuration, to get rid of that issue. Sounds a bit gratuitous, surely we can have a default value.

mikehearn:
@shamsasari How does it look now?

docs/source/upgrade-notes.rst
```diff
@@ -137,687 +157,75 @@ finalised transaction. If the initiator is written in a backwards compatible way
...
-  * The node’s name needs to be given as an ``CordaX500Name``, instead of using ``getX509Name``
+Step 4. Upgrade your use of SwapIdentitiesFlow
+----------------------------------------------
 
-:ref:`From Milestone 12 (First Public Beta) <changelog_m12>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ``confidential-identities`` module is experimental in Corda 3 and remains so in Corda 4. In this release, the ``SwapIdentitiesFlow``
+has been adjusted in the same way as ``FinalityFlow`` above, to close problems with confidential identities being injectable into a node
+outside of other flow context. Old code will still work, but it is recommended to adjust your call sites so a session is passed into
+the ``SwapIdentitiesFlow``.
```
anthonykeenan: Could do with a link to the Confiential identities section in `api-identity.rst`.

anthonykeenan:
With regards to the `issuableCurrencies`, in the O/S 3.3 I didn't need to add any configuration to get issuing currency to work. In ENT 3.2 you need to add `issuableCurrencies` to the `node.conf`, now you add it to the `cordapp.conf`. So there's a bit of a three way backwards compatibility issue.

mikehearn:
Have you taken a look at what it'd take to make that backwards compatible? Would be good to see if we can avoid this annoying thing.

anthonykeenan:
How do we want to make it backwards compatible though? Do we want to make it so that if there are no `issuableCurrencies` set it allows any? Surely the whole point of adding `issuableCurrencies` config was to stop that. I'm sure making the 'global' configuration version in enterprise compatible with the 'cordapp' configuration version on master would be easier, but the reason it was introduced was to seperate the finance app from corda internals - so we would have to re-introduce some coupling in order to make it work.

mikehearn:
Yeah, it could default to USD, CHF, GBP, EUR or something. Yes it's "insecure" but this app is basically a demo app anyway. We could log an error or warning about it.

anthonykeenan:
Ok, I've raised a JIRA to fix it: https://r3-cev.atlassian.net/browse/CORDA-2332

However, I would still argue if their cordapp is going to throw warnings about it, it's worth referring to it in the upgrade notes.

docs/source/upgrade-notes.rst
```diff
@@ -137,687 +157,75 @@ finalised transaction. If the initiator is written in a backwards compatible way
...
 
-Core data structures
-^^^^^^^^^^^^^^^^^^^^
+Step 5. Possibly, adjust unit test code
+---------------------------------------
```
anthonykeenan: If you are constructing a `MockServices` for testing contracts, and your contract uses the `Cash` contract from the finance app, you now need to explicitly add `net.corda.finance.contracts` to the list of `cordappPackages`.

i.e. (from the obligation cordapp)

val ledgerServices = MockServices(
            listOf("net.corda.examples.obligation", "net.corda.testing.contracts"),
            identityService = makeTestIdentityService(),
            initialIdentity = TestIdentity(CordaX500Name("TestIdentity", "", "GB")))

becomes

val ledgerServices = MockServices(
            listOf("net.corda.examples.obligation", "net.corda.testing.contracts", "net.corda.finance.contracts"),
            identityService = makeTestIdentityService(),
            initialIdentity = TestIdentity(CordaX500Name("TestIdentity", "", "GB")))

mikehearn:
Addressed review comments, added a brief mention of some of the new API features. PTAL.

docs/source/upgrade-notes.rst
```diff
@@ -4,48 +4,68 @@
...
-    ext.kotlin_version = 'x.y.z'
-    ext.quasar_version = 'x.y.z'
+This is used by the Corda Gradle build plugin to populate your app JAR with useful information. It should look like this::
 
-Please consult the relevant release notes of the release in question. If not specified, you may assume the
-versions you are currently using are still in force.
+    cordapp {
+        targetPlatformVersion 4
+        minimumPlatformVersion 4
+        contract {
```
shamsasari: Add the `workflow` section as well. We should be encouraging the contract vs flow separation immediately.

docs/source/upgrade-notes.rst
```diff
@@ -137,687 +157,102 @@ finalised transaction. If the initiator is written in a backwards compatible way
...
-^^^^^^^
-
-* MockNetwork testing:
+Step 4. Upgrade your use of SwapIdentitiesFlow
+----------------------------------------------
 
-  * Mock nodes in node tests are now of type ``StartedNode<MockNode>``, rather than ``MockNode``
+The :ref:`confidential_identities_ref` API is experimental in Corda 3 and remains so in Corda 4. In this release, the ``SwapIdentitiesFlow``
+has been adjusted in the same way as ``FinalityFlow`` above, to close problems with confidential identities being injectable into a node
+outside of other flow context. Old code will still work, but it is recommended to adjust your call sites so a session is passed into
```
shamsasari: Old code will not work - the c'tor has been removed. I've been debating whether to add backwards compatibility for old apps, like I did with FinalityFlow, but decided against it since it's experimental. I'm thinking now that was the wrong decision. WDYT?

mikehearn:
@josecoll Mentioned, please take another look! @shamsasari you too sir!

docs/source/upgrade-notes.rst
mikehearn: Well, we know we want to overhaul the API anyway if we can make it more automatic, so I'd suggest adding back to old c'tor so there's only one dollop of user pain around this module and the rest is smooth. Otherwise people will really learn to hate this feature ...

shamsasari: https://r3-cev.atlassian.net/browse/CORDA-2347. This is cutting it fine this late in the V4 dev cycle, but I'll do what I can.

PR #4415, opened 2018-12-14
[CORDA-2327]: Unable to verify transactions involving contract states depending on other CorDapps (fixed)

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2327

**Notes**

- This is a workaround, at best. Apart from being quite unclean in terms of implementation, the `AttachmentsClassLoader` should **not** use the CorDapps classloader as a parent, because that's not using historical versions of those CorDapps.

- A proper solution would involve modifying the `CorDapp` plugin to add to the JAR manifest information about dependent CorDapps, with hashes. After that, we could propagate that to the attachments we store in the database, and look up an attachment with its dependencies for the purpose of verifying transactions.


josecoll:
Please can you add an integration test that covers the scenario we are fixing.

sollecitom:
I can't reproduce it with an integration test. I tried even the asymmetric CorDapp setup, but I'm afraid our driver tried to fix these things.

josecoll:
You can use the DriverDSL to setup a similar scenario.

josecoll:
Please can you confirm this actually works by re-running the failed IRS Demo test scenario.

sollecitom:
IRS demo works. I created a test, but the driver doesn't work as it should and this setup is impossible to reproduce and I add to ignore it (the driver currently passes the classpath of the running test to out-of-process nodes, instead of using corda.jar). After we get rid of capsule, we can fix the driver and un-ignore the test.

screenshot from 2018-12-17 11-28-27

node/src/integration-test/kotlin/net/corda/node/TransactionAttachmentsTests.kt
```diff
@@ -0,0 +1,99 @@
...
+import net.corda.testing.driver.DriverParameters
+import net.corda.testing.driver.NodeParameters
+import net.corda.testing.driver.driver
+import net.corda.testing.node.internal.cordappForClasses
+import org.assertj.core.api.Assertions.assertThatCode
+import org.junit.Ignore
+import org.junit.Test
+import java.security.PublicKey
+
+class TransactionAttachmentsTests {
```
josecoll: Please rename to `ContractAttachmentTests`, and make it representative of the case of one Contract depending on another Contract (eg. MessageContract depends on AnotherMessageContract).

node/src/integration-test/kotlin/net/corda/node/TransactionAttachmentsTests.kt
```diff
@@ -0,0 +1,99 @@
...
+    // TODO un-ignore after fixing the driver in terms of classpath for external nodes.
+    @Ignore("This test always passes because the driver adds the classpath of the running test to out of process nodes.")
+    @Test
+    fun `state inheriting from type in another CorDapp works in terms of attachments resolution`() {
+        val cordapp1 = cordappForClasses(MessageContract.MessageState::class.java, Initiator::class.java, Responder::class.java, MessageContract::class.java, MessageContract.Commands::class.java, MessageContract.Commands.Create::class.java)
```
josecoll: only need to specify `MessageContract::class.java, Initiator::class.java, Responder::class.java`

josecoll: rename to `messageContractCordapp`

node/src/integration-test/kotlin/net/corda/node/TransactionAttachmentsTests.kt
```diff
@@ -0,0 +1,99 @@
...
+        val cordapp2 = cordappForClasses(WithText::class.java)
```
josecoll: rename to `dependentMessageContractCordapp` (or similar)

node/src/integration-test/kotlin/net/corda/node/TransactionAttachmentsTests.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    class MessageContract : Contract, WithText {
+        class MessageState(override val text: String, override val participants: List<AbstractParty>, override val linearId: UniqueIdentifier = UniqueIdentifier()) : LinearState, WithText
+
+        interface Commands : CommandData {
+            class Create : TypeOnlyCommandData(), Commands
+        }
+
+        override fun verify(tx: LedgerTransaction) {
+            logger.warn("MICHELE - VERIFYING!")
```
josecoll: remove MICHELE

node/src/integration-test/kotlin/net/corda/node/TransactionAttachmentsTests.kt
```diff
@@ -0,0 +1,99 @@
...
+            val command = tx.commandsOfType<MessageContract.Commands.Create>().singleOrNull()
+            require(command != null) { "Create command must be there" }
+            val state: WithText = tx.outputsOfType<MessageContract.MessageState>().single()
+            require(state.text.isNotEmpty()) { "Message text cannot be empty!" }
+            logger.warn("MICHELE - VERIFIED!")
```
josecoll: ditto

node/src/integration-test/kotlin/net/corda/node/TransactionAttachmentsTests.kt
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        override val text: String = "Contract!"
+
+        companion object {
+            const val PROGRAM_ID: ContractClassName = "net.corda.node.TransactionAttachmentsTests\$MessageContract"
```
josecoll: `MessageContractClassName`
`DependentMessageContractClassName`

node/src/integration-test/kotlin/net/corda/stubs/WithText.kt
```diff
@@ -0,0 +1,6 @@
+package net.corda.stubs
+
+interface WithText {
```
josecoll: remove

node/src/integration-test/kotlin/net/corda/node/TransactionAttachmentsTests.kt
sollecitom: Renamed. The IRS demo issue is exactly what it is in this test. A state depending on an interface, no contracts involved.

sollecitom: the types need to be in separate cordapp jars, because otherwise things work. The issue arises when a cordapp depending on a type from another cordapp is in play, and that cordapp is not directly involved in the same transaction.

node/src/integration-test/kotlin/net/corda/stubs/WithText.kt
josecoll: in that case, localize it to the test. Don't need another source file for this.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -45,6 +45,16 @@ class AttachmentsClassLoader(attachments: List<Attachment>, parent: ClassLoader
             URL.setURLStreamHandlerFactory(AttachmentURLStreamHandlerFactory)
         }
 
+        // Mighty unclean, but we need a quick stopgap to the bug it's addressing.
+        // TODO Remove ASAP after proper handling of dependent CorDapps with regards to attachments.
+        object CorDappsClassLoaderHolder {
+            var instance: ClassLoader? = null
+
```
josecoll: remove blank line

josecoll:
@mikehearn please can you comment on this temporary patch.
Do we ship Corda 4 with it, or fix it properly by parsing the Contract jar attachment to correctly identify any dependent classes (and ensuring these are classloaded by the Attachments Classloader)?

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: Removed, but I thought we did have one in between variables and functions.

sollecitom:
@josecoll , @tudor-malene and @mikehearn . What do we want to do then? We could wait for a proper solution, true, and this problem is probably not that common anyway, but IRS demo doesn't work without this fix, and there's another class-loading related bug preventing finance built vs 3.3 from working with Corda 4.

Shall we revert the classloading isolation for cordapps and attachments? Shall we merge this workaround and try to fix the other one? Shall we ignore this one? Mind that the finance one can hardly be ignored.

tudor-malene:
This workaround reverts the classloader isolation, but I don't think it's what we want.

I would recommend going for the hardcoding approach:
`If transaction has contract attachment with hash h1 also add attachment with h2.`

This would fix our demo, and we could do the same for third parties who might have the same issue. (unlikely)

sollecitom:
@tudor-malene mind implementing that then? This approach would not retro-fix cordapps with the IRS situation, because even if we made `addCordappDependencyHash()` public API (and we shouldn't), they haven't done it yet. Also, in terms of user experience is pretty poor, and we have that information in the cordapp plugin anyway.

tudor-malene:
I didn't want to imply any api change. 
Just an internal hardcoding mechanism to address invalid transactions that were created with v3.x in a case by case way.

@mikehearn , wdyt?

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -174,6 +174,28 @@ internal object AttachmentsClassLoaderBuilder {
...
+    override fun loadClass(name: String?): Class<*> {
+        for (classLoader in classLoaders) {
+            try {
+                return classLoader.loadClass(name)
+            } catch (e: ClassNotFoundException) {
+                // Keep iterating without failing.
+            }
+        }
+        val contextClassLoader = Thread.currentThread().contextClassLoader
```
mikehearn: I suspect what we want to do here is actually chain to the parent, rather than use the CCL (which can be modified by arbitrary code?).

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -157,12 +157,12 @@ internal object AttachmentsClassLoaderBuilder {
...
 
     fun <T> withAttachmentsClassloaderContext(attachments: List<Attachment>, block: (ClassLoader) -> T): T {
-
-
         // Create classloader from the attachments.
         // TODO This should not default to the CorDapps classloader, but it's needed for now to stop a bug preventing CorDapps with dependencies on other CorDapps from working as attachments.
         // TODO A proper fix would require gathering information about dependent CorDapps with hashes at build time, and importing these as attachments as well.
-        val transactionClassLoader = AttachmentsClassLoaderBuilder.build(attachments, AttachmentsClassLoader.Companion.CorDappsClassLoaderHolder.instance)
+        val attachmentsClassLoader = AttachmentsClassLoaderBuilder.build(attachments)
+        val cordappsClassLoader = AttachmentsClassLoader.Companion.CorDappsClassLoaderHolder.instance
```
mikehearn: Why `.Companion.` here - it's not meant to be needed in Kotlin.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: Well, arguably we can, but not trying with the context class loader as last step means if anyone do some logic involving that, it will get broken.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: It's an object within the companion object. Doesn't compile without it. Moved it to be top-level within the same file, which avoids this altogether.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: Fixed to use both parent and contextClassLoader

sollecitom:
@mikehearn I cannot write an automated test for this scenario, because the driver passes the classpath of the running test to out-of-process nodes, meaning there's no way to reproduce it.
After we switch to Java 9, we can get rid of capsule, and use the `corda.jar` to start out-of-process nodes from the driver instead.

mikehearn:
How so? Start a node with v1 of an app, wait for it to be installed. Stop the node, start it with v2 of an app, to whitelist that that too. Now you have a node with v2 installed and v1+v2 in the attachment stores. Now start an app that attaches v1 and expects v1 behaviour, verify a tx, and ensure the v1 behaviour applied, not the v2 behaviour. Seems like it doesn't matter what the node classpath is there?

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -145,7 +146,11 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
         }
     }
 
-    val cordappLoader: CordappLoader = makeCordappLoader(configuration, versionInfo)
+    val cordappLoader: CordappLoader = makeCordappLoader(configuration, versionInfo).also {
+        // Mighty unclean, but we need a quick stopgap to the bug it's addressing.
+        // TODO Remove ASAP after proper handling of dependent CorDapps with regards to attachments.
+        CorDappsClassLoaderHolder.set(it.appClassLoader)
```
josecoll: surely by setting this you are automatically switching off Attachments Classloading ?

sollecitom:
@mikehearn the test should reproduce the issue, so a cordapp attachment depending on a type defined in another cordapp contract JAR, which does not appear in the transaction itself. This is currently not possible to test with the driver, because the type will always be on the system classloader, given that the driver augments the classpath of all out of process nodes with the classpath of the running test.

It's pointless to test the cascading classloader, especially as an integration test.

mikehearn:
I don't follow. In the scenario I described it doesn't matter what's on the node classpath, as long as you can change from v1 to v2 across runs, which is possible already (or it used to be).

I disagree it's pointless to test this. @josecoll Want to take a look at possible approaches from a testing POV?

sollecitom:
@mikehearn I can easily test the cascading classloader, no nodes involved, but we can't test whether this solves the original issue.

tudor-malene:
@mikehearn, I think this is a mistake.

Using this approach we will be masking the real problem, and allow people to continue building invalid transactions for which we will have to find solutions in the future when the cordapp classloader will no longer have the required jars.

At least we should disable this cascading behaviour when building transactions so that it is constrained to the past, and print some warning when we actually cascade to the cordapp classloader.

We can instruct users to send us the warning and hardcode all the existing missing dependencies in core in version 5 and get rid of cascading.

node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
sollecitom: No

mikehearn:
We can't assume global knowledge of what all platform users are doing. There are private apps out there that we'll never hear about, and we can't break those either.

I agree this doesn't solve the real problem but see the discussion in #development, there is a plan to re-attack this for the next release after C4, which will involve removing the delegation link when apps are indicating it's safe to do so via their target version. Eventually old apps will be upgraded or phased out and the delegation will trigger less and less often until one day it never does.

josecoll:
Add https://r3-cev.atlassian.net/browse/CORDA-2356 for Corda 5.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -180,15 +180,27 @@ private class CascadingClassLoader(classLoaders: Sequence<ClassLoader>) : ClassL
     private val classLoaders = classLoaders.toList()
 
     override fun loadClass(name: String?): Class<*> {
-        for (classLoader in (classLoaders + parent + Thread.currentThread().contextClassLoader).filterNotNull()) {
+        for (candidate in candidates) {
```
josecoll: why rename this to `candidate` `candidates` (if was clearer before).

josecoll:
sure - over to @mikehearn 

The test scenario I suggested as discussed is:

```
start Node with Attachment X + Attachment Y (where X depends on Y)
create a transaction (issue)
stop Node
restart Node with Attachment X only (Y will now be stored in nodes attachments store).
create a transaction (spend) with previous input
```

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: Because I can't reuse `classLoaders` here. `allClassLoaders` compiles but kind of sucks. Those are candidates after all.

josecoll: Sorry - I don't follow.
What was wrong with the code before this latest commit?

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -141,16 +150,19 @@ internal object AttachmentsClassLoaderBuilder {
     private val cache: MutableMap<List<SecureHash>, AttachmentsClassLoader> = createSimpleCache<List<SecureHash>, AttachmentsClassLoader>(ATTACHMENT_CLASSLOADER_CACHE_SIZE)
             .toSynchronised()
 
-    fun build(attachments: List<Attachment>): AttachmentsClassLoader {
+    fun build(attachments: List<Attachment>, parentClassLoader: ClassLoader? = null): AttachmentsClassLoader {
```
tudor-malene: This is not really a parentClassloader.
It's an alternative classloader to look for dependencies.
I think some comments should be added to explain why this is required

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: Before, `(classLoaders + parent + Thread.currentThread().contextClassLoader).filterNotNull())` was used directly but, since the last commit, it's used in 2 places because resources need to be handled the same way classes are, in a cascade.
At that point, using `(classLoaders + parent + Thread.currentThread().contextClassLoader).filterNotNull())` twice is a duplication, so a `private val candidates get() = (classLoaders + parent + Thread.currentThread().contextClassLoader).filterNotNull())` was introduced.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -162,6 +174,35 @@ internal object AttachmentsClassLoaderBuilder {
...
+        for (candidate in candidates) {
+            val url = candidate.getResource(name)
+            if (url != null) {
+                return url
+            }
+        }
+        return null
+    }
+
+    private val candidates: List<ClassLoader> get() = (classLoaders + parent + Thread.currentThread().contextClassLoader).filterNotNull()
```
tudor-malene: I don't think you need the parent and the contextClassLoader  for our use case.

All you need is the attachmentsClassloader and the alternative cordappsClassloader.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: This is a `parentClassLoader`, not the one I use in the cascade. I'll remove this addition, since it's confusing.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
josecoll: well something with classloader in the name then,
`overridingClassloaders`, `fallbackClassloaders`

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
sollecitom: I created a general-purpose cascading classloader. Whether we need things or not, the closer this is to a normal class loader, the better.

tudor-malene: Not sure what you mean by "normal classloader".

All it has to do - for this use case -  is try to load the class ( or resource) in order from the classloaders it was initialised with.
Attempting to load from anything else is just a potential source of problems.

remmeier:
Hi, we upgraded our sources to the most recent version (release/v4) and now we sometimes see :

``` 
java.lang.IllegalArgumentException: Attempted to find dependent attachment for class net.corda.core.contracts.TransactionState (erased) -> 
data(net.corda.core.contracts.ContractState) -> net.corda.core.contracts.TransactionState (erased) -> 
data(net.corda.core.contracts.ContractState) -> net.corda.core.contracts.TransactionState (erased) -> 
data(net.corda.core.contracts.ContractState) -> net.corda.core.contracts.TransactionState (erased) ->
 data(net.corda.core.contracts.ContractState) -> ch/test/identity/api/schema/OwnedState,
 but could not find a suitable candidate. 
at net.corda.core.transactions.TransactionBuilder.addMissingDependency(TransactionBuilder.kt:179) 
at net.corda.core.transactions.TransactionBuilder.toWireTransactionWithContext$core(TransactionBuilder.kt:160) 
at net.corda.core.transactions.TransactionBuilder.toWireTransactionWithContext$core$default(TransactionBuilder.kt:128) 
at net.corda.core.transactions.TransactionBuilder.toWireTransaction(TransactionBuilder.kt:125) 
at net.corda.core.transactions.TransactionBuilder.toLedgerTransaction(TransactionBuilder.kt:477) 
at net.corda.core.transactions.TransactionBuilder.verify(TransactionBuilder.kt:485) 
at ch.test.corda.base.flow.commit.CommitTransactionFlow$Initiator.call(CommitTransactionFlow.java:130) 
at ch.test.corda.base.flow.commit.CommitTransactionFlow$Initiator.call(CommitTransactionFlow.java:29) 
at net.corda.node.services.statemachine.FlowStateMachineImpl.subFlow(FlowStateMachineImpl.kt:291) 
at net.corda.core.flows.FlowLogic.subFlow(FlowLogic.kt:311) 
at ch.test.corda.base.flow.simulator.SimulationFlow$InitiatorBase.call(SimulationFlow.java:88) 
at ch.test.corda.base.flow.simulator.SimulationFlow$InitiatorBase.call(SimulationFlow.java:26) 
at net.corda.node.services.statemachine.FlowStateMachineImpl.subFlow(FlowStateMachineImpl.kt:291) 
at net.corda.core.flows.FlowLogic.subFlow(FlowLogic.kt:311) 
at com.example.flow.event.EventFlow$EventFlowInitiater.call(EventFlow.java:53) 
at com.example.flow.event.EventFlow$EventFlowInitiater.call(EventFlow.java:38) 
at net.corda.node.services.statemachine.FlowStateMachineImpl.subFlow(FlowStateMachineImpl.kt:291) 
at net.corda.core.flows.FlowLogic.subFlow(FlowLogic.kt:311) 
at com.example.flow.event.Form1320aFlow$EventFlowInitiater.call(Form1320aFlow.java:38)
at com.example.flow.event.Form1320aFlow$EventFlowInitiater.call(Form1320aFlow.java:25) 
at net.corda.node.services.statemachine.FlowStateMachineImpl.run(FlowStateMachineImpl.kt:228)  
```

That particular OwnedState is from another Cordapp. Both cordapps are installed on the server. Is that something expected or an error? If expected, could the error be improved to give a hint of what should be done?

tudor-malene:
Hi @remmeier,
In Corda 4 we haven't yet implemented dependencies between corDapps - as it is quite a complex topic.
For now you should just bundle everything you depend on in your own contract jar. (Including dependant cordapps)

We have added a very basic way of helping users auto-detect dependencies, and the message that you posted is basically from that mechanism. 
It would be interesting to find out why it's not working properly for you if the dependant cordapp is installed.
Can you set a breakpoint there and take a look ?

Thank you,
Tudor

remmeier:
well yes a I see it:

```
val missingClass = requireNotNull(e.message) { "Transaction $ltx is incorrectly formed." }
val attachment = requireNotNull(services.attachments.internalFindTrustedAttachmentForClass(missingClass)) {
```

taking the exception message as className gives things like this:

```
CET
ch/test/identity/api/schema/MethodPermission.class net.corda.core.contracts.TransactionState (erased) -> data(net.corda.core.contracts.ContractState) -> net.corda.core.contracts.TransactionState (erased) -> data(net.corda.core.contracts.ContractState) ->...
```

using that has classname obviously does not work. I guess the particular code has somewhere been working, so maybe the JDK makes a difference. In this case anapsix/alpine-java:8u192b12_server-jre_unlimited.

In general cordapp dependencies are fairly vital for many use cases (all of our applications make use of it), The workaround itself a bit on the hacky side and rather dangerous for production use cases. It might would have been good to have a feature flag for such major features to allow to roll them gradually. I could also imagine that some application may do not want that particular feature and rely on more low-tech means of maintaing compatibility.

tudor-malene:
The message is not great. Probably nobody payed attention to it during testing. I will raise a story to fix. Or if you want you can raise a PR.

Wrt cordapp dependencies: for now ( C3 and 4), you should fat-jar all dependencies into your cordapps. There should be no dangling dependency jars in the attachments, as there is not much protection for them.

The workaround is indeed hacky. It's there so that transactions created pre-v4 which were passing even though they weren't complete should still pass.
It's not dangerous as the logic is only looking through "trusted" attachments. This means attachments that were installed by the node operator. 
An attacker would not be able to plant some unsafe attachment on your node and then reference it later.

remmeier:
I workaround it locally with:

```
val sep = className1.lastIndexOf("->")
var className = if(sep != -1) className1.substring(sep + 2).trim() else className1
```

but obviously that should not be the solution. I did not follow up on the entire thread and scope of the workaround, but would probably have gone for a parent classloader or something similar rather than attempting to catch NoClassDefFoundError. May then also not too far away from a future solution with proper dependency management.

dangerous I meant less from a security perspective, more in terms of robustness. e.g. we make use of content hashes for versioning, so versions are not monotonic. but something else. so the ordering by versions as done in this PR will not give the latest one in our case. would have probably took us quite a while to notice something like this if we did not stumble upon this issue first. 

we already have multiple cordapps from different companies, not all have exactly the same setup, try to separate contracts from flows, etc.  so we will attempt to rather go with this workaround rather than do fat jars again. not sure whether I can help with this one, but could help e.g. with https://github.com/corda/corda-gradle-plugins/issues/167 which touches the same area.

PR #4417, opened 2018-12-14
CORDA-2301 JVM args documentation

blsemo:
Document the ways and behaviours of setting JVM args on a corda node process.

### DIFF
--- a/docs/source/corda-configuration-file.rst
+++ b/docs/source/corda-configuration-file.rst
@@ -266,6 +266,11 @@ The available config fields are listed below.
     :excludedAutoAcceptableParameters: List of auto-acceptable parameter names to explicitly exclude from auto-accepting. Allows a node operator to control the behaviour at a
                                        more granular level. Defaults to an empty list.
 
+:custom: Set custom command line attributes on the node process via the capsule launcher
+
+        :jvmArgs: A list of JVM arguments to apply to the node process. This removes any defaults specified in the fat jar, but can be overriden
+                  from the command line. See :ref:`setting_jvm_args` for details on arguments precedence.
+
 Examples
 --------
 
--- a/docs/source/corda-configuration-file.rst
+++ b/docs/source/corda-configuration-file.rst
@@ -198,9 +198,6 @@ The available config fields are listed below.
 
         .. note:: This is an unsupported configuration.
 
-:jvmArgs: An optional list of JVM args, as strings, which replace those inherited from the command line when launching via ``corda.jar``
-    only. e.g. ``jvmArgs = [ "-Xmx220m", "-Xms220m", "-XX:+UseG1GC" ]``
-
 :systemProperties: An optional map of additional system properties to be set when launching via ``corda.jar`` only.  Keys and values
     of the map should be strings. e.g. ``systemProperties = { visualvm.display.name = FooBar }``
 
@@ -266,10 +263,11 @@ The available config fields are listed below.
     :excludedAutoAcceptableParameters: List of auto-acceptable parameter names to explicitly exclude from auto-accepting. Allows a node operator to control the behaviour at a
                                        more granular level. Defaults to an empty list.
 
-:custom: Set custom command line attributes on the node process via the capsule launcher
+:custom: Set custom command line attributes (e.g. Java system properties) on the node process via the capsule launcher
 
-        :jvmArgs: A list of JVM arguments to apply to the node process. This removes any defaults specified in the fat jar, but can be overriden
-                  from the command line. See :ref:`setting_jvm_args` for details on arguments precedence.
+        :jvmArgs: A list of JVM arguments to apply to the node process. This removes any defaults specified from ``corda.jar``, but can be
+                  overriden from the command line. See :ref:`setting_jvm_args` for examples and details on the precedence of the different
+                  approaches to settings arguments.
 
 Examples
 --------
--- a/docs/source/deploying-a-node.rst
+++ b/docs/source/deploying-a-node.rst
@@ -56,6 +56,7 @@ handling, and ensures the Corda service is run at boot.
               ]
           }
       ]
+      custom { jvmArgs = [ '-Xmx2048m', '-XX:+UseG1GC' ] }
 
 7. Make the following changes to ``/opt/corda/node.conf``:
 
@@ -90,7 +91,7 @@ handling, and ensures the Corda service is run at boot.
        Type=simple
        User=corda
        WorkingDirectory=/opt/corda
-       ExecStart=/usr/bin/java -Xmx2048m -jar /opt/corda/corda.jar
+       ExecStart=/usr/bin/java -jar /opt/corda/corda.jar
        Restart=on-failure
 
        [Install]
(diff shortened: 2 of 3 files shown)


docs/source/deploying-a-node.rst
```diff
@@ -56,6 +56,7 @@ handling, and ensures the Corda service is run at boot.
               ]
           }
       ]
+      custom { jvmArgs = [ '-Xmx2048m', '-XX:+UseG1GC' ] }
```
gendal: weird how the minus is still expected/required...   what happens if it's omitted I wonder?  I mean: the fact it's an argument is obvious from the context, right?

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,40 @@ Optionally run the node's webserver as well by opening a terminal window in the
 
 .. warning:: The node webserver is for testing purposes only and will be removed soon.
 
+.. _setting_jvm_args:
+
+Setting JVM arguments
+~~~~~~~~~~~~~~~~~~~~~
+
+There are several ways of setting JVM arguments for the node process (particularly the garbage collector and the memory settings).
+They are listed here in order of increasing priority, i.e. if the same flag is set in a way lower in this list, it will override
```
gendal: do you mean later?

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,40 @@ Optionally run the node's webserver as well by opening a terminal window in the
...
+anything set earlier.
+
+:Default arguments in capsule: The capuled corda node has default flags set to ``-Xmx512m -XX:+UseG1GC`` - this gives the node (a relatively
```
gendal: typo

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,40 @@ Optionally run the node's webserver as well by opening a terminal window in the
...
+   low) 512 MB of heap space and turns on the G1 garbage collector, ensuring low pause times for garbage collection.
```
gendal: Think we should document a recommendation here.  "However, for typical production deployments, it is common to use xxxxx"

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,40 @@ Optionally run the node's webserver as well by opening a terminal window in the
...
+
+:Node configuration: The node configuration file can specify custom default JVM arguments by adding a section like::
+
+      custom = {
+         jvmArgs: [ '-Xmx1G', '-XX:+UseG1GC' ]
+      }
+
+   Note that this will completely replace any defaults set by capsule above, not just the flags that are set here, so if you use this
```
gendal: do we log if/when this happens so a puzzled administrator doesn't tear their hair out the first time this happens to them?

gendal: does it also completely replace any defaults set _below_?  (ie you outline another approach below... which makes it sound like there are THREE ways to set the node's Jim parameters... does node.conf always win?)

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,40 @@ Optionally run the node's webserver as well by opening a terminal window in the
...
+      launcher that actually starts it::
+
+      java -Dcapsule.jvm.args="-Xmx:1G" corda.jar
+
+   This will override any ``-Xmx`` value set previously, but not interfere with any other JVM arguments that are configured in a way above.
+   This is particarly useful for either setting large memory allowances that you don't want to give to the launcher or for setting values that
+   can only be set on one process at a time, e.g. a debug port.
+
+:Command line flag: You can set JVM args on the command line that apply to the launcher process and the node process as in the example
+      above. This will override any value for the same flag set any other way.
```
gendal: precedence again... I'm confused about what supersedes what!

docs/source/running-a-node.rst
blsemo: I don't think so - depending on what your do with your node and how you configure it, anything between 1G and 32G might be appropriate.

docs/source/deploying-a-node.rst
blsemo: These arguments are passed to the JVM verbatim, so the `-` is required. We could change it, but that has its own issues. 
- I am not sure if there are any arguments that take parameters as an entry separated by a space?
- We would break any configs that have `-` in place unless we add clever logic to check for that.
I think overall changing this is just not worth it.

gendal: fine... I was just wondering :)

docs/source/running-a-node.rst
blsemo: I have a ticket open for tuning a node in Enterprise (including memory, threads, RPC threads and so on). Getting these parameters for a node right is a non-trivial process.

docs/source/running-a-node.rst
blsemo: No. It works in the order of the list, anything further down the list always wins over anything further up.

docs/source/corda-configuration-file.rst
```diff
@@ -266,6 +266,11 @@ The available config fields are listed below.
     :excludedAutoAcceptableParameters: List of auto-acceptable parameter names to explicitly exclude from auto-accepting. Allows a node operator to control the behaviour at a
                                        more granular level. Defaults to an empty list.
 
+:custom: Set custom command line attributes on the node process via the capsule launcher
+
+        :jvmArgs: A list of JVM arguments to apply to the node process. This removes any defaults specified in the fat jar, but can be overriden
+                  from the command line. See :ref:`setting_jvm_args` for details on arguments precedence.
```
gendal: "for details on the precedence of the different approaches to setting arguments"

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,41 @@ Optionally run the node's webserver as well by opening a terminal window in the
...
+   launcher that actually starts it::
+
+      java -Dcapsule.jvm.args="-Xmx:1G" corda.jar
+
+   Setting a flag like this will override any value for this flag, but not interfere with any other JVM arguments that are configured in a
+   way above. In this example, it would reset the maximum heap memory to ``-Xmx1G`` but not touch the garbage collector settings.
+   This is particarly useful for either setting large memory allowances that you don't want to give to the launcher or for setting values that
+   can only be set on one process at a time, e.g. a debug port.
+
+:Command line flag: You can set JVM args on the command line that apply to the launcher process and the node process as in the example
```
gendal: So anything set for the launcher process is also applied to the node process (assuming not overridden)?  ie with nothing else set at all, the launcher and node JVMs will have the exact same settings?

gendal:
those minor changes really help make it clearer - thanks!

docs/source/corda-configuration-file.rst
rick-r3: jvmArgs is mentioned further up this file.

docs/source/corda-configuration-file.rst
```diff
@@ -266,6 +266,11 @@ The available config fields are listed below.
     :excludedAutoAcceptableParameters: List of auto-acceptable parameter names to explicitly exclude from auto-accepting. Allows a node operator to control the behaviour at a
                                        more granular level. Defaults to an empty list.
 
+:custom: Set custom command line attributes on the node process via the capsule launcher
```
rick-r3: I think custom is just an unenforced section of the config.

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,41 @@ Optionally run the node's webserver as well by opening a terminal window in the
...
+:Node configuration: The node configuration file can specify custom default JVM arguments by adding a section like::
+
+      custom = {
+         jvmArgs: [ '-Xmx1G', '-XX:+UseG1GC' ]
+      }
+
+   Note that this will completely replace any defaults set by capsule above, not just the flags that are set here, so if you use this
+   to set e.g. the memory, you also need to set the garbage collector, or it will revert to whatever default your JVM is using.
+
+:Capsule command line flag: You can use a special capsule command line flag to set JVM arguments only for the Corda process, not the
```
rick-r3: it's a Java system property.  I would say something like `You can use a special Java system property that Capsule understands to set....`

docs/source/running-a-node.rst
```diff
@@ -43,6 +43,41 @@ Optionally run the node's webserver as well by opening a terminal window in the
...
+   launcher that actually starts it::
+
+      java -Dcapsule.jvm.args="-Xmx:1G" corda.jar
+
+   Setting a flag like this will override any value for this flag, but not interfere with any other JVM arguments that are configured in a
```
rick-r3: `are configured in any way above` ?

docs/source/running-a-node.rst
blsemo: Yes. Anything set for the launcher process will be applied to the node process and I haven't found a way to override - it takes highest precedence.

PR #4422, opened 2018-12-14
CORDA-2331 Split Workflow and contracts of Finance App into separate Cordapps

szymonsztuka:
Split Workflow and contracts of Finance App into separate Cordapps.


shamsasari:
Shouldn't the remaining code be in a `finance:workflows` module?

mikehearn:
Is it possible to leave behind a fatjar artifact so any apps that are depending on finance don't break? I know it's experimental/demo code, so we can make this change without violating any guarantees, but if it's a small lift it'd be nice to provide this continuity. We can put stuff in the upgrade notes about the split.

finance/finance-contracts/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+}
+
+artifacts {
+    testArtifacts testJar
+}
+cordapp {
+    targetPlatformVersion corda_platform_version.toInteger()
+    minimumPlatformVersion 1
+    contract {
+        name "net/corda/finance/contracts"
```
mikehearn: Is this naming scheme really the one we're using? In the upgrade notes and other places we say this is a free text string that's meant to be human meaningful, perhaps "Corda Finance Demo" would make more sense here?

Also if we're renaming it and breaking people's gradle configs, we should rename it to "finance-demo" at the same time.

josecoll:
@mikehearn we probably should leave the original aswell (for the reasons you mention), but need to highlight in the original build.gradle that the path forward is to always use the split version (due to hardened security checks in Corda 4).

finance/finance-contracts/build.gradle
josecoll: Yes - I questioned this when I revisited all the samples (to add the new metadata attributes).
Not sure where this naming convention came in, but I'd prefer a simpler English short descriptive name (rather than a file path reference).

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Michele" />
```
shamsasari: What's this?

docs/source/example-code/build.gradle
```diff
@@ -42,7 +42,8 @@ dependencies {
     compile project(path: ":webserver:webcapsule", configuration: 'runtimeArtifacts')
 
     // Corda Plugins: dependent flows and services
```
shamsasari: Corda Plugins -> CorDapps

finance/finance-contracts/build.gradle
shamsasari: Can we address this here and change it.

.idea/codeStyles/codeStyleConfig.xml
josecoll: What the hell?

finance/finance-contracts/build.gradle
josecoll: "Corda Finance Demo" it is!

vkolomeyko:
Can we please make sure that with those changes Demobench still works?
(https://github.com/corda/corda/blob/master/tools/demobench/build.gradle#L129)

@chrisr3 - FYI

tools/demobench/build.gradle
```diff
@@ -50,7 +50,8 @@ dependencies {
     compile "org.controlsfx:controlsfx:$controlsfx_version"
 
     compile project(':client:rpc')
-    compile project(':finance')
+    compile project(':finance:finance-contracts')
+    compile project(':finance:finance-workflows')
```
chrisr3: So `:finance` is now two jars rather than one? Fair enough, but this means that you're going to need to _install_ two jars too. DemoBench cannot deploy CorDapps it doesn't have.

tools/demobench/build.gradle
```diff
@@ -126,8 +127,8 @@ distributions {
                 into 'corda'
                 fileMode = 0444
             }
-            from(project(':finance').tasks.jar) {
-                rename 'corda-finance-(.*)', 'corda-finance.jar'
+            from(project(':finance:finance-contracts').tasks.jar) {
+                rename 'corda-finance-contracts(.*)', 'corda-finance.jar'
```
chrisr3: This is `corda-finance-contracts.jar`. Don't you need to install `corda-finance-workflows.jar` too so that DemoBench can deploy _both_ CorDapps into new nodes that it creates? You've presumably kept the original `corda-finance.jar` name because DemoBench expects it, but I suspect this is hiding a bug and that no-one has tried to _run_ DemoBench.

josecoll: Testing this now ...

client/jfx/build.gradle
```diff
@@ -26,7 +26,8 @@ sourceSets {
 
 dependencies {
     compile project(':core')
-    compile project(':finance')
+    compile project(':finance:finance-contracts')
```
shamsasari: Can we not have the dependencies as `:finance:contracts` and `:finance: workflows`? The second "finance" is redundant.

tools/demobench/src/main/kotlin/net/corda/demobench/plugin/CordappController.kt
```diff
@@ -14,11 +14,13 @@ import kotlin.streams.toList
 class CordappController : Controller() {
     companion object {
         const val FINANCE_CORDAPP_FILENAME = "corda-finance"
```
chrisr3: You shouldn't need `FINANCE_CORDAPP_FILENAME` any more.

tools/demobench/build.gradle
josecoll: Fixed and re-tested.

client/jfx/build.gradle
josecoll: Not sure what you mean.
It is currently explicit enough I think.

tools/demobench/build.gradle
josecoll: Fixed.

tools/demobench/src/main/kotlin/net/corda/demobench/plugin/CordappController.kt
josecoll: It is still required for the node config file name.

client/jfx/build.gradle
shamsasari: :finance:contracts` instead of `:finance:finance-contracts`

chrisr3:
There's a problem here with issuing cash. A Bank node's `cordapps/config/corda-finance.conf` file is _supposed_ to enable `Issue` transactions in Node Explorer for all its listed currencies:
```
{
    "issuableCurrencies" : [
        "GBP"
    ]
}
```
However, I'm only seeing the `Pay` transaction listed.

@josecoll The good news is that [CORDA-2351](https://r3-cev.atlassian.net/browse/CORDA-2351) _isn't_ a bug after all because this PR hasn't been merged yet :relieved:.

josecoll:
thanks @chrisr3 
The `CashConfigDataFlow` was using the funky CordappProvider Stack walking to locate a config file with the same name as the loaded CorDapps jars.
So `cordapps/config/corda-finance.conf` was invalid.
Fixed to use `cordapps/config/corda-finance-workflows.conf` as only this CorDapp requires a config file.

PR #4429, opened 2018-12-17
[CORDA-2341]: Fixing ABI compatibility for TransactionBuilder vs Corda 3.3.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2341

This might break the API. If that's the case, we can decide not to release these overloaded constructors, or to add them back manually.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
+++ b/core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
@@ -73,13 +73,9 @@ open class TransactionBuilder @JvmOverloads constructor(
     }
 
     protected var references: MutableList<StateRef> = arrayListOf()
-        private set(value) {
-            field = value
-        }
+        private set
     protected var serviceHub: ServiceHub? = (Strand.currentStrand() as? FlowStateMachine<*>)?.serviceHub
-        private set(value) {
-            field = value
-        }
+        private set
 
     private val inputsWithTransactionState = arrayListOf<StateAndRef<ContractState>>()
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -50,15 +50,37 @@ open class TransactionBuilder @JvmOverloads constructor(
...
+            privacySalt: PrivacySalt = PrivacySalt(),
+            references: MutableList<StateRef> = arrayListOf(),
+            serviceHub: ServiceHub? = (Strand.currentStrand() as? FlowStateMachine<*>)?.serviceHub
+    ) : this(notary, lockId, inputs, attachments, outputs, commands, window, privacySalt) {
+        this.references = references
+        this.serviceHub = serviceHub
+    }
+
+    protected var references: MutableList<StateRef> = arrayListOf()
+        private set(value) {
```
chrisr3: I think `private set` is enough. The setter's actual implementation is standard and so the compiler can generate it by default.

core/src/main/kotlin/net/corda/core/transactions/TransactionBuilder.kt
```diff
@@ -50,15 +50,37 @@ open class TransactionBuilder @JvmOverloads constructor(
...
+            field = value
+        }
+    protected var serviceHub: ServiceHub? = (Strand.currentStrand() as? FlowStateMachine<*>)?.serviceHub
+        private set(value) {
```
chrisr3: Again, just `private set`

sollecitom:
@chrisr3 fixed that :)

PR #4430, opened 2018-12-18
Remove documentation of building container images with jib

thschroeter:
This is not inline with the official container image.


docs/source/index.rst
```diff
@@ -104,4 +103,4 @@ We look forward to seeing what you can do with Corda!
    corda-network/index.md
    corda-network/governance-structure.md
    corda-network/governance-guidelines.md
-   corda-network/joining-corda-network.md
\ No newline at end of file
+   corda-network/joining-corda-network.md
```
wawrzek: I think adding this line is OK, but is there any particual reason to bring it in this PR?

thschroeter: There was no newline at the end of the file. My setup is always adding a newline as the last byte of the file. And i tend to check in those changes. It would be good to decide on a convention.

I prefer having the newline at the end for cleaner diffs and working with unix tools, e.g. search and replace using `find ./ -name '*.kt' -type f -exec sed -i '' -e 's/committed_transactions/committed_txs/g' {} \;` to give an example.

PR #4433, opened 2018-12-18
ENT-2848 More contract attachment caching to avoid database queries slowing the node down

rick-r3:

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
+++ b/node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
@@ -54,7 +54,6 @@ class NodeAttachmentService(
         cacheFactory: NamedCacheFactory,
         private val database: CordaPersistence
 ) : AttachmentStorageInternal, SingletonSerializeAsToken() {
-
     // This is to break the circular dependency.
     lateinit var servicesForResolution: ServicesForResolution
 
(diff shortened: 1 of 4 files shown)


node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -403,32 +402,62 @@ class NodeAttachmentService(
...
-                val criteriaBuilder = session.criteriaBuilder
+        fun toList(): List<AttachmentId> =
+            if(signed != null) {
+                if(unsigned != null) {
+                    listOf(signed, unsigned)
+                } else listOf(signed)
+            } else listOf(unsigned!!)
+    }
+
+    private val contractsCache = InfrequentlyMutatedCache<String, NavigableMap<Int, AttachmentIds>>("NodeAttachmentService_contractAttachmentVersions", cacheFactory)
```
josecoll: Can you just add a comment here to explain the contents of this cache:
ContractClassName -> Versions -> AttachmentIds (where a single version of a ContractClassName may have one or both of a signed/unsigned version of the same contract jar code).

rick-r3: I've added a comment and some type aliases to make it clearer.

node/src/main/kotlin/net/corda/node/services/persistence/NodeAttachmentService.kt
```diff
@@ -45,6 +45,9 @@ import java.util.jar.JarInputStream
 import javax.annotation.concurrent.ThreadSafe
 import javax.persistence.*
 
+typealias ClassName = String
```
josecoll: actually we already have both of these typealias's ;-)

`typealias ContractClassName = String` in `TransactionState` (core)
`typealias Version = Int` in `Version.kt` (core)

rick-r3: So we do!  Switched to those.

PR #4434, opened 2018-12-18
CORDA-2345: Simplified TestCordapp to make it inline with the recent CorDapp versioning changes

shamsasari:
`TestCordapp` has now two implementations to clearly separate the two use cases it has in the Corda repo:

* `TestCordappImpl` which implements the revised public API of `TestCordapp`; namely that a TestCordapp instance references a real CorDapp jar on the classpath. This is either an external dependency jar in which case it’s taken as is and given to the node, or it’s a local gradle project in which case it’s compiled using the gradle `jar` task to generate the CorDapp jar. This approach means the jar has all the original CorDapp versioning information, which is important that it’s correct when testing. To this end, TestCordapp only needs to expose the ability to specify the app’s config. All the remaining properties have moved to `CustomCordapp`.

* `CustomCordapp` for creating arbitrary custom CorDapps, including specifying the jar’s MANIFEST values. This is internal API and only used for testing the platform. Technically this shouldn’t implement `TestCordapp` but does so to reduce the complexity of the driver and mock network.


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
```diff
@@ -0,0 +1,139 @@
...
+import java.util.jar.JarFile
+import java.util.jar.JarOutputStream
+import java.util.jar.Manifest
+import java.util.zip.ZipEntry
+
+/**
+ * Represents a completely custom CorDapp comprising of resources taken from packages on the existing classpath, even including individual
+ * disparate classes. The CorDapp metadata that's present in the MANIFEST can also be tailored.
+ */
+data class CustomCordapp(
```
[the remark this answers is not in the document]
shamsasari: I didn't see any uses of this, though it is a good idea. Perhaps add it when we need it.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
```diff
@@ -0,0 +1,139 @@
...
+    companion object {
+        private val logger = contextLogger()
+        private val epochFileTime = FileTime.from(Instant.EPOCH)
+        private val cordappsDirectory: Path
+        private val defaultJarSignerDirectory: Path
+        private val whitespace = "\\s".toRegex()
+        private val cache = ConcurrentHashMap<CustomCordapp, Path>()
+
+        init {
+            val buildDir = Paths.get("build").toAbsolutePath()
```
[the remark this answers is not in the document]
shamsasari: Yes, tested on both

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
```diff
@@ -0,0 +1,139 @@
...
+            val timeDirName = getTimestampAsDirectoryName()
+            cordappsDirectory = buildDir / "generated-custom-cordapps" / timeDirName
+            defaultJarSignerDirectory = buildDir / "jar-signer" / timeDirName
+        }
+
+        fun getJarFile(cordapp: CustomCordapp): Path {
```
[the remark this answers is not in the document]
shamsasari: That will complicate things without much benefit I think. The purpose of the directory is a cache, so for max benefit we'd want to use the same directory per JVM instance.

sollecitom:
OK then, as long as it's internal.

docs/source/upgrade-notes.rst
```diff
@@ -202,6 +202,9 @@ becomes::
         initialIdentity = TestIdentity(CordaX500Name("TestIdentity", "", "GB"))
     )
 
+You may need to use the new ``TestCordapp`` API when testing with the node driver or mock network, especially if you decide to stick with the
+old ``FinalityFlow`` API. The previous way of pulling in CorDapps into your tests does not honour CorDapp versioning.
```
josecoll: old ambiguous -> change to `pre-Corda 4`

testing/node-driver/src/main/kotlin/net/corda/testing/node/TestCordapp.kt
```diff
@@ -1,87 +1,41 @@
...
+ * Encapsulates a CorDapp that exists on the current classpath, which can be pulled in for testing. Use [TestCordapp.Factory.findCordapp]
+ * to locate an existing CorDapp.
+ *
+ * @see DriverParameters.cordappsForAllNodes
+ * @see NodeParameters.additionalCordapps
+ * @see MockNetworkParameters.cordappsForAllNodes
+ * @see MockNodeParameters.additionalCordapps
  */
 @DoNotImplement
 interface TestCordapp {
```
josecoll: personally I prefer to the `withXXX` DSL methods over `copy()`

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
josecoll: why introduce another type?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
```diff
@@ -85,20 +83,21 @@ class DriverDSLImpl(
...
         val waitForAllNodesToFinish: Boolean,
+        val extraCordappPackagesToScan: List<String>,
         val jmxPolicy: JmxPolicy,
         val notarySpecs: List<NotarySpec>,
         val compatibilityZone: CompatibilityZoneParams?,
         val networkParameters: NetworkParameters,
         val notaryCustomOverrides: Map<String, Any?>,
         val inMemoryDB: Boolean,
-        val cordappsForAllNodes: Collection<TestCordapp>,
-        val signCordapps: Boolean
```
josecoll: why has this been removed?

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
shamsasari: For the ability to create truly custom CorDapp jars, which some of our tests require. I've explained all in the Kdocs for this class and in this PR. Let me know if this hasn't been explained properly.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
shamsasari: We had two ways to sign a jar previously: this parameter on the driver which would sign all jars, and on the test cordapp instance. The way this was implemented made it hard to follow, plus it's a bit unnecessary when you can specify signing on the `CustomCordapp` level.

testing/node-driver/src/main/kotlin/net/corda/testing/node/TestCordapp.kt
shamsasari: I'll add a `withConfig` for the config option.

Otherwise for the other properties in the internal `CustomCordapp` I'd rather not add wither methods when our Kotlin tests can benefit from the default parameters in `copy`.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
```diff
@@ -41,7 +41,9 @@ data class CustomCordapp(
 
     override val scanPackage: String get() = throw UnsupportedOperationException()
 
-    override fun withoutMeta(): CustomCordapp = copy(sign = null, config = emptyMap())
+    override fun withConfig(config: Map<String, Any>): CustomCordapp = copy(config = config)
+
+    override fun withoutMeta(): CustomCordapp = CustomCordapp(packages = packages, classes = classes)
```
josecoll: can we rename this to something more meaningful (or explain what it means with a comment) ?

node/src/integration-test/kotlin/net/corda/node/CordappConstraintsTests.kt
```diff
@@ -35,6 +38,7 @@ class CordappConstraintsTests {
                 invokeRpc(CordaRPCOps::wellKnownPartyFromX500Name),
                 invokeRpc(CordaRPCOps::notaryIdentities),
                 invokeRpc("vaultTrackByCriteria")))
+        val UNSIGNED_FINANCE_CORDAPP = cordappWithPackages("net.corda.finance")
```
josecoll: I am confused here ...
previously we explicitly specified in a test when to use a Signed Vs Unsigned version of a CorDapp.
It is not clear what is going on now.

node/src/integration-test/kotlin/net/corda/node/CordappConstraintsTests.kt
```diff
@@ -44,9 +48,11 @@ class CordappConstraintsTests {
                 inMemoryDB = false
         )) {
 
-            val alice = startNode(NodeParameters(additionalCordapps = listOf(FINANCE_CORDAPP.signJar()),
+            val alice = startNode(NodeParameters(
+                    additionalCordapps = listOf(FINANCE_CORDAPP),
```
josecoll: I preferred explicitly specifying signed or unsigned (as its no longer clear from the context).

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
shamsasari: It's documented on the interface declaration, but I've also renamed this to be clearer.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
josecoll: OK but the name kind of suggests its public rather than internal.
after all it extends `TestCordappInternal`.

node/src/integration-test/kotlin/net/corda/node/CordappConstraintsTests.kt
shamsasari: Well, since all our samples are now signed, it's a bit redundant to add the SIGNED_ prefix, plus I've documented `FINANCE_CORDAPP` to say it's signed. I'll add a `SIGNED_FINANCE_CORDAPP` val which just points to `FINANCE_CORDAPP` to make this clearer.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
```diff
@@ -0,0 +1,141 @@
...
+ * Represents a completely custom CorDapp comprising of resources taken from packages on the existing classpath, even including individual
+ * disparate classes. The CorDapp metadata that's present in the MANIFEST can also be tailored.
+ */
+data class CustomCordapp(
+        val packages: Set<String>,
+        val name: String = "custom-cordapp",
+        val versionId: Int = 1,
+        val targetPlatformVersion: Int = PLATFORM_VERSION,
+        val classes: Set<Class<*>> = emptySet(),
+        val sign: Sign? = null,
```
josecoll: what is this?
previously we had 

```
    /** Returns whether the CorDapp should be jar signed. */
    val signJar: Boolean
```

which defaulted to a sensible value.
Why introduce a nullable type?

shamsasari: Yeah, I changed it to tighten the semantics of `keyStorePath`, which is only needed if the boolean is set, which otherwise was a bit confusing to deal with. I'll rename it to `SigningInfo`.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
shamsasari: This feature was not previously public, or rather was not designed to be public by @szymonsztuka in its previous state (after having spoken with him). You could need to flesh it out completely with keystore aliases, passwords, etc. Right now it's all hardcoded.

Secondly, I don't think developers would need signing capabilities anyway. Whether a CorDapp is signed or not is self-evident in the dependency jar file that's pulled in, or it's specified in the CorDapp project's build.gradle (if it's a local build). It either case the app developer doesn't need to tell the driver or mock network to sign the jar, as it's already signed.

shamsasari:
I'm going to wait for https://github.com/corda/corda/pull/4422 to be merged first.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/CustomCordapp.kt
josecoll: OK - fair enough.
We may expose this publicly if there is demand later on for more advanced testing.

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/DriverDSLImpl.kt
josecoll: Lets hope so ... assuming that are starting with a clean sheet Corda 4.
Otherwise they may wish to test Upgradability of their Corda 3 CorDapp (unsigned) to Corda 4 (signed). I guess we can always expose the internal `CustomCordapp` if there is a demand for more advanced testing.

PR #4435, opened 2018-12-18
Reduce the operation of ip autodetect to nodes specifying localhost

mnesbit:
The current IP autodetect can trigger with any 192.168.x.x address, or 10.x.x.x address even if the user has intentionally set it. Users also sometimes setup 0.0.0.0 in the p2pAddress field.
This change makes the autodetect only apply to localhost:<port> addresses and prints a clear message to say how to disable the lookup if desired.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/Node.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/Node.kt
@@ -43,7 +43,9 @@ import net.corda.node.services.api.FlowStarter
 import net.corda.node.services.api.ServiceHubInternal
 import net.corda.node.services.api.StartedNodeServices
 import net.corda.node.services.config.*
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/internal/Node.kt
```diff
@@ -311,6 +313,9 @@ open class Node(configuration: NodeConfiguration,
 
     private fun getAdvertisedAddress(): NetworkHostAndPort {
         return with(configuration) {
+            require(p2pAddress.host != "0.0.0.0") {
```
mikehearn: Probably using an assert to do this isn't the right exception type for user-config errors, but will approve for now to get this into C4. Maybe ask exp-prod-readyness what the right way to indicate configuration errors is?

PR #4443, opened 2018-12-19
[CORDA-1828] Documentation update for observer nodes

JamesHR3:
This PR provides a documentation update for Observer nodes, which states that they must be informed of all transactions in a given chain if they are to record the creation and deletion of states correctly. It also adds a test to catch the original issue reported in this ticket, which currently hasn't been reproduced.

See https://r3-cev.atlassian.net/browse/CORDA-1828 for more discussion.

Changes:
 - Extra bullet point in observer node tutorial
 - Extra test added to NodeStatePersistenceTests
 - Change to the test MessageStateContract to support the new test


gendal:
This is interesting... is it by design? ie why doesn't an observer node resolve transactions? Or is it that it does but it then doesn't pluck out everything it finds and records to vault? (Which I guess could be a massive thing for long/wide back chains?)

JamesHR3:
I think it's more that with the privacy model for transactions that's planned, the observer node has no visibility into the states it encounters in the chain, and so wouldn't be able to record the states even if it wanted to. Either way it sounds like some clarification would help, so I'll add that.

node/src/integration-test/kotlin/net/corda/node/persistence/NodeStatePersistenceTests.kt
```diff
@@ -99,8 +102,89 @@ class NodeStatePersistenceTests {
         val retrievedMessage = stateAndRef!!.state.data.message
         assertEquals(message, retrievedMessage)
     }
+
+
+    @Test
+    fun `Broadcasting an old transaction will cause 2 unconsumed states`() {
```
shamsasari: `will` -> `does not`, otherwise it reads as if we want this bug!

node/src/integration-test/kotlin/net/corda/node/persistence/NodeStatePersistenceTests.kt
```diff
@@ -99,8 +102,89 @@ class NodeStatePersistenceTests {
...
+        val user = User("mark", "dadada", setOf(startFlow<SendMessageFlow>(), startFlow<SendMessageFlowConsuming>(), startFlow<ReportToCounterparty>(), invokeRpc("vaultQuery") ,
+                invokeRpc("internalVerifiedTransactionsSnapshot")))
+        val regulatorUser = User("mark", "dadada", setOf(startFlow<ReceiveReportedTransaction>(), invokeRpc("vaultQuery")))
+
+        driver(DriverParameters(
```
shamsasari: It should be possible to test this using the `MockNetwork` (or `InternalMockNetwork`). It's faster than the driver, and for this bug test we don't need the real messaging layer that mock network forgoes.

node/src/integration-test/kotlin/net/corda/testMessage/MessageState.kt
```diff
@@ -55,7 +55,7 @@ open class MessageContract : Contract {
         val command = tx.commands.requireSingleCommand<Commands.Send>()
         requireThat {
             // Generic constraints around the IOU transaction.
-            "No inputs should be consumed when sending a message." using (tx.inputs.isEmpty())
+           // "No inputs should be consumed when sending a message." using (tx.inputs.isEmpty())
```
shamsasari: ?

JamesHR3: This is commented out as I've reused the MessageState and contract in a flow that consumes a message state. I was planning to just remove it, although I can add a new contract/state if that would be preferable?

node/src/integration-test/kotlin/net/corda/node/persistence/NodeStatePersistenceTests.kt
```diff
@@ -138,4 +199,75 @@ class SendMessageFlow(private val message: Message, private val notary: Party) :
...
+
+    @Suspendable
+    override fun call() {
+        // TODO: add error handling
+
+        subFlow(IdentitySyncFlow.Receive(otherSideSession))
+
+        val recorded = subFlow(ReceiveTransactionFlow(otherSideSession, true, StatesToRecord.ALL_VISIBLE))
+
+        otherSideSession.send(recorded)
```
shamsasari: You don't need to send back the tx that's just been received.

node/src/integration-test/kotlin/net/corda/node/persistence/NodeStatePersistenceTests.kt
```diff
@@ -138,4 +199,75 @@ class SendMessageFlow(private val message: Message, private val notary: Party) :
...
+    }
+}
+
+
+@InitiatedBy(ReportToCounterparty::class)
+class ReceiveReportedTransaction(private val otherSideSession: FlowSession) : FlowLogic<Unit>() {
+
+    @Suspendable
+    override fun call() {
+        // TODO: add error handling
```
shamsasari: Remove

node/src/integration-test/kotlin/net/corda/testMessage/MessageState.kt
shamsasari: Won't this affect other tests that use it? Best to revert and use your own one if this class isn't what you need. If you do write your own contract, then keep it as simple as possible. It shouldn't be ebalorate as it can distract from the real issue at hand.

node/src/test/kotlin/net/corda/node/services/persistence/ObserverNodeTransactionTests.kt
```diff
@@ -0,0 +1,195 @@
...
+@InitiatingFlow
+@StartableByRPC
+class ReportToCounterparty(
+        private val regulator: Party,
+        private val signedTx: SignedTransaction) : FlowLogic<Unit>() {
+
+    @Suspendable
+    override fun call() {
+        val session = initiateFlow(regulator)
+
```
shamsasari: remove these empty lines

node/src/test/kotlin/net/corda/node/services/persistence/ObserverNodeTransactionTests.kt
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+
+@InitiatedBy(ReportToCounterparty::class)
+class ReceiveReportedTransaction(private val otherSideSession: FlowSession) : FlowLogic<Unit>() {
+
+    @Suspendable
+    override fun call() {
+
```
shamsasari: ditto

node/src/test/kotlin/net/corda/node/services/persistence/ObserverNodeTransactionTests.kt
```diff
@@ -0,0 +1,195 @@
...
+        subFlow(IdentitySyncFlow.Receive(otherSideSession))
```
shamsasari: You're not using confidential identities so I think you can remove the use of `IdentitySyncFlow`

PR #4395, opened 2018-12-10
NOT FOR MERGE - V4 release notes

fenryka:
Starting to get the release notes for C4 into order, raising a PR against master for visibilty and comments, will eventually merge into the release branch once cut.

PR #4445, opened 2018-12-19
Refresh release notes for Corda 4

mikehearn:
Taking over from https://github.com/corda/corda/pull/4395

First round of updates done. I will go over the remaining comments from the old PR tomorrow.


mikehearn:
Here's a PDF printout of the web page for easier review:

[Release notes for Corda 4 — R3 Corda Master documentation.pdf](https://github.com/corda/corda/files/2695997/Release.notes.for.Corda.4.R3.Corda.Master.documentation.pdf)

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
+to read data from the ledger without simultaneously updating it. They're useful not only for any kind of
+reference data such as rates, healthcare codes, geographical information etc, but for anywhere
+you might have used a SELECT JOIN in a SQL based app.
 
-* **Corda Serialization fails with "Unknown constant pool tag"**
+A reference input state is a ``ContractState`` which can be referred to in a transaction by the contracts
+of input and output states but, significantly, whose contract is not executed as part of the transaction
+verification process and is not consumed when the transaction is committed to the ledger. Rather, it is checked
+for "current-ness". In other words, the contract logic isn't run for the referencing transaction only.
+It's still a normal state when it occurs in an input or output position.
```
shamsasari: Since they're normal states, if they occur in the input or output positions, they can evolve on the ledger, modeling reference data in the real world.

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
+For a given contract class, the contract attachment of the output states must be of the same or newer version than
+the contract attachment of the input states. See :ref:`Contract attachment non-downgrade rule <contract_non-downgrade_rule_ref>` for further information.
 
-    Ben's contribution can be found on GitHub
-    `here <https://github.com/corda/corda/commit/d17670c747d16b7f6e06e19bbbd25eb06e45cb93>`__
+Network parameters in transactions
+++++++++++++++++++++++++++++++++++
 
-  * Tomas Tauber for adding support for running Corda atop PostgresSQL in place of the in-memory H2 service
+Transactions created under a Corda 4+ node will have the currently valid signed ``NetworkParameter``
```
shamsasari: Missing 's' in `NetworkParameter`

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
 
-    Tomas's contribution can be found on GitHub
-    `here <https://github.com/corda/corda/commit/342090db62ae40cef2be30b2ec4aa451b099d0b7>`__
+RPC upgrades
+++++++++++++
 
-    .. warning:: This is an experimental feature that has not been tested as part of our standard release testing.
+**AMQP/1.0** is now default serialization framework across all of Corda (checkpointing aside), swapping the RPC
+framework from using the older Kryo implementation. This means Corda open source and Enterprise editions are
+now RPC wire compatible and either client library can be used.
```
shamsasari: Add a small note, as a reminder, that this was enabled for P2P in 3.

josecoll:
Section 2 in the App-Upgrade Notes:

Before the paragraph:

`"The duplication between contract and workflow blocks exists because you should split your app into two separate JARs/modules, one that contains on-ledger validation code like states and contracts, and one for the rest (called by convention the "workflows" module although it can contain a lot more than just flows: services would also go here, for instance)."`

We should state why we have introduced the Contract VersionId and where it is used (eg. the doc ref I gave you previously to the non-downgrade rule checking). Should also add a note to say that the Workflow VersionId will be used in future releases of Corda to improve flow evolution and upgrades.

josecoll:
PS: apologies but I wasn't aware you had already merged the **App Upgrade notes.**

mikehearn:
That's alright! I can make those fixes whilst I'm here too.

docs/source/app-upgrade-notes.rst
```diff
@@ -45,16 +54,18 @@ This is used by the Corda Gradle build plugin to populate your app JAR with usef
...
         workflow {
             name "MegaApp flows"
             vendor "MegaCorp"
-            license "A really expensive proprietary license"
+            licence "A really expensive proprietary licence"
         }
     }
 
+.. important:: Watch out for the UK spelling of the word licence (with a c).
+
```
gendal: The name and vendor stuff (line 69) is confusing.  Is versionId the thing we're referring to as targctversion in the text?

docs/source/release-notes.rst
gendal: We do verify the reference state's backchain, right? ie the state's contract itself is not run, but we do check it is the output of a chain of valid transactions?

docs/source/app-upgrade-notes.rst
```diff
@@ -45,16 +54,18 @@ This is used by the Corda Gradle build plugin to populate your app JAR with usef
...
         }
         workflow {
             name "MegaApp flows"
             vendor "MegaCorp"
-            license "A really expensive proprietary license"
+            licence "A really expensive proprietary licence"
         }
     }
 
+.. important:: Watch out for the UK spelling of the word licence (with a c).
```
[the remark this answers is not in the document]
mikehearn: Me too :) I bet this issue will bite us again and again. Serialisation is another fun one.

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
 
-  This issue is most often seen when running a CorDapp with a Rest API using / provided by ``Spring Boot``.
+Signature constraints
++++++++++++++++++++++
 
-  The fundamental cause was ``Corda 3.0`` shipping with an out of date dependency for the
-  `fast-classpath-scanner <https://github.com/lukehutch/fast-classpath-scanner>`_ library, where the manifesting
-  bug was already fixed in a released version newer than our dependant one. In response, we've updated our dependent
-  version to one including that bug fix.
+CorDapps built by the ``corda-gradle-plugins`` are now signed and sealed JAR files by default. This
```
gendal: how/where does this interact with target version, etc?  eg if I change nothing and just start compiling an old v3 app against whatever stuff we ship with c4, what happens?  or is the compatibility guarantee only applicable if v3 apps are compiled against v3?

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
+on the ledger by ``StateRef`` (a pair of transaction hash and output index that is sufficient to locate
+any information on the global ledger). State pointers work together with the reference states feature
+to make it easy for data to point to the latest version of any other piece of data, with the right
+version being automatically incorporated into transactions for you.
 
-.. _release_notes_v3_0:
+New network builder tool
+++++++++++++++++++++++++
+
+A new graphical tool for building test Corda networks has been added. It can build Docker images for local
```
gendal: hopefully redundant question but can somebody ( @wawrzek  ? ) confirm that the docker images created by this tool use the same dockerfiles or are otherwise aligned with the other docker 'stuff' we're doing?

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
+
+Corda 3 provides the ``jdbcConnection`` API on ``FlowLogic`` to give access to an active connection to your
+underlying database. It is fully intended that apps can store their own data in their own tables in the
+node database, so app-specific tables can be updated atomically with the ledger data itself. But JDBC is
+not always convenient, so in Corda 4 we are additionally exposing the *Java Persistence Architecture*, for
+object-relational mapping. The new ``ServiceHub.withEntityManager`` API lets you load and persist entity
+beans inside your flows and services.
+
+Please do write apps that read and write directly to tables running alongside the node's own tables. Using
+SQL is a convenient and robust design pattern for accessing data on or off the ledger.
```
gendal: I think we need to add a big scary warning here...  basically telling people what NOT to do (eg don't write to tables prefixed with NODE or whatever the rule actually is)

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
 
-    Ben's contribution can be found on GitHub
-    `here <https://github.com/corda/corda/commit/d17670c747d16b7f6e06e19bbbd25eb06e45cb93>`__
+Network parameters in transactions
+++++++++++++++++++++++++++++++++++
 
-  * Tomas Tauber for adding support for running Corda atop PostgresSQL in place of the in-memory H2 service
+Transactions created under a Corda 4+ node will have the currently valid signed ``NetworkParameter``
+file attached to each transaction. This will allow future introspection of states to ascertain what was
+the accepted global state of the network at the time they were notarised. Additionally, new signatures must
```
gendal: The last sentence is confusingly worded.  How about something like: "The notary signing a transaction will check that it does indeed reference the current in-force network parameters, meaning that old (and superseded) network parameters can not be used to create new transactions".  Something like that? Assuming notaries do actually check!

docs/source/release-notes.rst
gendal: have we tested this?  if not, is it strictly necessary for us to make this promise yet?

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
+code must be fully deterministic. Out of the box a standard JVM is not fully deterministic, thus we must
+make some modifications in order to satisfy our requirements.
 
-  Systems can thus be deployed safe in the knowledge that valuable and important information will always be accessible through
-  upgrade and change. Practically speaking this means from this point forward upgrading all, or part, of a Corda network
-  will not require the replaying of data; "it will just work".
+This version of Corda introduces a standalone :doc:`key-concepts-djvm`. It isn't yet integrated with
+the rest of the platform. It will eventually become a part of the node and enforce deterministic and
+secure execution of smart contract code, which is mobile and may propagate around the network without
+human intervention.
```
gendal: I don't see why it _shouldn't_ but have we checked in the last few months that it still works?!

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
-  which the nodes within a network need to agree in order to be assured of seamless inter-operation. Additional security
-  controls ensure that all network map data is now signed, thus reducing the power of the network operator to tamper with
-  the map.
+Errors generated in Corda are now hashed to produce a unique error code that can be
+used to perform a lookup into a knowledge base. The lookup URL will be printed to the logs when an error
+occur. Here's an example::
 
-  There is also support for a group of nodes to operate locally, which is achieved by copying each
-  node's signed info file to the other nodes' directories. We've added a bootstrapping tool to facilitate this use case.
+[ERROR] 2018-12-19T17:18:39,199Z [main] internal.NodeStartupLogging.invoke - Exception during node startup: The name 'O=Wawrzek Test C4, L=London, C=GB' for identity doesn't match what's in the key store: O=Wawrzek Test C4, L=Ely, C=GB [errorCode=wuxa6f, moreInformationAt=https://errors.corda.net/OS/4.0-SNAPSHOT/wuxa6f]
```
gendal: shame there's nothing there!

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
-  Zone constraints provide a less restrictive but more centralised control mechanism. This can be useful when you want
-  the ability to upgrade an app and you don’t mind the upgrade taking effect “just in time” when a transaction happens
-  to be required for other business reasons. These allow you to specify that the network parameters of a compatibility zone
-  (see :doc:`network-map`) is expected to contain a map of class name to hashes of JARs that are allowed to provide that
-  class. The process for upgrading an app then involves asking the zone operator to add the hash of your new JAR to the
-  parameters file, and trigger the network parameters upgrade process. This involves each node operator running a shell
-  command to accept the new parameters file and then restarting the node. Node owners who do not restart their node in
-  time effectively stop being a part of the network.
+We have open sourced the Liquibase schema upgrade feature from Corda Enterprise. The node now uses Liquibase to
+bootstrap and update itself automatically. This is a transparent change with pre Corda 4 nodes seamlessly
```
gendal: where is this tested?

docs/source/app-upgrade-notes.rst
mikehearn: Nope. It's different. I'll try and word it better.

mikehearn:
@josecoll @shamsasari Review comments addressed. I also made the release/upgrade notes much more prominent at the top of the toctree, as the don't really fit properly in any existing section. PTAL.

docs/source/release-notes.rst
josecoll: That's a good observation. 
There is currently no validation in the cordapp plugin to prevent signing a CorDapp when the targetversion is < 4. We could easily add this.
Not sure what would happen if we tried this .... but most likely an unhappy Corda Node

docs/source/release-notes.rst
josecoll: Thats odd - I already commented on Kat's PR about this paragraph.
Firstly, transactions are tagged with a hash of the network parameters (not a file) which is resolved internally using the new network parameters storage service.

josecoll: Now I know, I think Kat took over the PR before Kat has a change to review my comments and adjust accordingly.

docs/source/release-notes.rst
mikehearn: I added a discussion of this to the app upgrade notes.

docs/source/release-notes.rst
mikehearn: Not sure. Task for after merging this PR?

docs/source/release-notes.rst
mikehearn: I don't know, I'm just the messenger :) Tudor or Szymon know about this change.

mikehearn:
@josecoll @gendal done, PTAL

docs/source/release-notes.rst
mikehearn: It's not aligned. Official Docker post-dates this tool and came too late. We can align in a 4.x release.

docs/source/app-upgrade-notes.rst
```diff
@@ -45,38 +54,62 @@ This is used by the Corda Gradle build plugin to populate your app JAR with usef
...
+
+Name, vendor and licence can be set to any string you like, they don't have to be Corda identities.
+
+Target versioning is a new concept introduced in Corda 4. Learn more by reading :doc:`versioning`.
+Setting a target version of 4 opts in to changes that might not be 100% backwards compatible, such as
+API semantics changes or disabling workarounds for bugs that may be in your apps, so by doing this you
+are promising that you have thoroughly tested your app on the new version. Using a high target version is
+a good idea because some features and improvements are only available to apps that opt in.
+
+The minimum platform version is just the (major) version number of the node that you require, so if you
```
shamsasari: This isn't consistent with how platform version is defined. There's also the fear that with this throwaway statement we're committing to this. Perhaps caveat this by saying this may not be the case in future releases.

docs/source/app-upgrade-notes.rst
```diff
@@ -45,38 +54,62 @@ This is used by the Corda Gradle build plugin to populate your app JAR with usef
...
+do not know about this metadata and don't check it, so your app will still be loaded in such nodes and
+may exhibit undefined behaviour at runtime. However it's good to get in the habit of setting this
+properly for future releases.
+
+The new ``versionId`` number is a version code for **your** app, and is unrelated to Corda's own versions.
+It is used to block state downgrades: when a state constraint can be satisfied
+by multiple attachments, the version is tracked in the ledger and cannot decrement. This ensures security
+fixes in CorDapps stick and can't be reversed by downgrading states to an earlier version. See
+":ref:`contract_non-downgrade_rule_ref`" for more information. In future, the version attached to the
+workflow JAR will also be used to help implement smoother upgrade and migration features. You may directly
```
shamsasari: Workflow jars haven't been explained yet.

docs/source/app-upgrade-notes.rst
```diff
@@ -237,20 +255,48 @@ to be governed by a contract that is either:
...
 
 You can read more about signature constraints and what they do in :doc:`api-contract-constraints`. The ``TransactionBuilder`` class will
-automatically use them if your application JAR is signed. We recommend all JARs are signed. To start signing your JAR files, read
-:ref:`cordapp_build_system_signing_cordapp_jar_ref`.
+automatically use them if your application JAR is signed. **We recommend all JARs are signed**. To learn how to sign your JAR files, read
+:ref:`cordapp_build_system_signing_cordapp_jar_ref`. In dev mode, all JARs are signed by developer certificates. If a JAR that was signed
+with developer certificates is deployed to a production node, the node will refuse to start. Therefore to deploy apps built for COrda 4
+to production you will need to generate signing keys and integrate them with the build process.
+
+Step 8. Security: package namespace handling
```
shamsasari: Capital P

docs/source/app-upgrade-notes.rst
```diff
@@ -237,20 +255,48 @@ to be governed by a contract that is either:
...
+--------------------------------------------
+
+Almost no apps will be affected by these changes, but they're important to know about.
+
+There are two improvements to how Java package protection is handled in Corda 4:
+
+1. Package sealing
+2. Package namespace ownership
+
+**Sealing.** App isolation has been improved. Version 4 of the finance CorDapp (*corda-finance.jar*) is now built as a sealed and signed JAR file.
```
shamsasari: Technically there's three finance CorDapp jars in 4. As `corda-finance` is only there for backwards compatibility, perhaps only reference the workflows and contracts jars instead.

docs/source/release-notes.rst
```diff
@@ -1,995 +1,295 @@
...
+to read data from the ledger without simultaneously updating it. They're useful not only for any kind of
+reference data such as rates, healthcare codes, geographical information etc, but for anywhere
+you might have used a SELECT JOIN in a SQL based app.
 
-* **Corda Serialization fails with "Unknown constant pool tag"**
+A reference input state is a ``ContractState`` which can be referred to in a transaction by the contracts
+of input and output states but, significantly, whose contract is not executed as part of the transaction
+verification process and is not consumed when the transaction is committed to the ledger. Rather, it is checked
+for "current-ness". In other words, the contract logic isn't run for the referencing transaction only.
+Since they're normal states, if they occur in the input or output positions, they can evolve on the ledger,
```
shamsasari: if they do occur

docs/source/app-upgrade-notes.rst
mikehearn: Ah ha. And in fact that means, people should update their gradle files with the new dependencies - another upgrade step I forgot about.

mikehearn:
Sorry, I keep invalidating people's reviews because everyone keeps commenting.

@shamsasari PTAL. If you approve this one I'll merge, and we can continue in another PR.

docs/source/app-upgrade-notes.rst
```diff
@@ -75,29 +79,39 @@ API semantics changes or disabling workarounds for bugs that may be in your apps
...
-The minimum platform version is just the (major) version number of the node that you require, so if you
+The minimum platform version is the platform version of the node that you require, so if you
 start using new APIs and features in Corda 4, you should set this to 4. Unfortunately Corda 3 and below
 do not know about this metadata and don't check it, so your app will still be loaded in such nodes and
 may exhibit undefined behaviour at runtime. However it's good to get in the habit of setting this
 properly for future releases.
 
+.. note:: Whilst it's currently a convention that Corda releases have the platform version number as their
+   major version i.e. Corda 3.3 implements platform version 3, this is not actually required and may in
+   future not hold true. You should know the platform version of the node releases you wan to target.
```
shamsasari: typo: wan

mikehearn:
Thanks. I enabled the spelling inspector in IntelliJ for the docs and caught one more typo.

mikehearn:
Next steps:

* Merge if TeamCity passes
* Add info about the Docker images
* Review Tudor's comments and possibly add more info about the new rules e.g. non-downgrade, constraints propagation etc.

docs/source/release-notes.rst
```diff
@@ -1,995 +1,268 @@
...
+**Sealing.** Sealed JARs are a security upgrade that ensures JARs cannot define classes in each other's packages,
+thus ensuring Java's package-private visibility feature works. The Gradle plugins now seal your JARs
+by default.
 
-Special Thanks
-~~~~~~~~~~~~~~
+**BelongsToContract annotation.** CorDapps are currently expected to verify that the right contract
+is named in each state object. This manual step is easy to miss, which would make the app less secure
+in a network where you trade with potentially malicious counterparties. The platform now handles this
+for you by allowing you to annotate states with which contract governs them. If states are inner
```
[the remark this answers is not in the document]
mikehearn: I think it's only required if you target version 4? @tudor-malene ?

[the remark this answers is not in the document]
ulrikrasmussen: That's correct: https://github.com/corda/corda/blob/9cdda3bd7722d91f7ea54ad724ef86fe4c2a9217/core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt#L175

It will only generate a warning if the target version is <4. But if is present in a contract with target version <3, I guess it would be by mistake, so a more correct behavior is perhaps to also throw an exception in that case? I am not sure.

That being said, the current sentence reads as if the annotation is always optional.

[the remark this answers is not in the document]
tudor-malene: I agree with @ulrikrasmussen. 
We should be more explicit.

We allow the previous behaviour only for backwards compatibility reasons.
Any corDapp that is being developed now *should* target version 4 and *is required* to declare the state-contract relationship.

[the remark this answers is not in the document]
ulrikrasmussen: I think it is fine that the annotation is not required for <4 contracts, since that would otherwise break compatibility.

However, since this is a critical security feature, the mere presence of `@BelongsToContract` in <4 contracts should trigger an unconditional exception in my opinion, instead of the current behavior which merely generates a warning. The warning will almost always imply that the contract has lost all of the security guarantees that the programmer intended (i.e. anyone can basically forge the contract states, as in the current `Cash` contract), and such code should not be allowed to be deployed.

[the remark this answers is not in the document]
ulrikrasmussen: By the way, I realize that this discussion is out of scope of this PR which is only concerned with the release notes, so please let me know if it should be taken up elsewhere :).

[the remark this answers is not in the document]
mikehearn: Not at all, thanks for the discussion. I have a feeling you are right about this. At the very least, I accept the argument that it's only technically optional and not really optional. 

I will merge this PR anyway so I stop invalidating people's reviews with changes, and open a new PR to continue, which will update this language.

docs/source/release-notes.rst
```diff
@@ -1,995 +1,295 @@
...
+might otherwise cause confusion.
+
+**No downgrades.** Transaction building and verification enforces new contract attachment version non-downgrade rule.
+For a given contract class, the contract attachment of the output states must be of the same or newer version than
+the contract attachment of the input states. See :ref:`Contract attachment non-downgrade rule <contract_non-downgrade_rule_ref>` for further information.
+
+Network parameters in transactions
+++++++++++++++++++++++++++++++++++
+
+Transactions created under a Corda 4+ node will have the currently valid signed ``NetworkParameters``
```
kasiastreich: Transactions have network parameters hash as a component group. The file isn't attached itself to the transaction

mikehearn: Right, but that's what "attachment" in the context of Corda means - a hash of some resolvable data is included, so it's at least internally consistent. I see what you mean but I doubt the distinction matters for release notes. Thanks for the review!

PR #4446, opened 2018-12-20
ENT-2848 remove need for generated IDs on some frequently used tables

rick-r3:
The generated IDs slow things down, and shouldn't be necessary.  There's a natural primary key.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -4398,6 +4398,14 @@ public static class net.corda.core.schemas.CommonSchemaV1$LinearState extends ne
   public void setParticipants(java.util.Set<net.corda.core.identity.AbstractParty>)
   public void setUuid(java.util.UUID)
 ##
+public interface net.corda.core.schemas.DirectStatePersistable extends net.corda.core.schemas.StatePersistable
+  @Nullable
+  public abstract net.corda.core.schemas.PersistentStateRef getStateRef()
+##
+public interface net.corda.core.schemas.IndirectStatePersistable extends net.corda.core.schemas.StatePersistable
+  @NotNull
+  public abstract T getCompositeKey()
+##
 public class net.corda.core.schemas.MappedSchema extends java.lang.Object
   public <init>(Class<?>, int, Iterable<? extends Class<?>>)
   @NotNull
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -4410,7 +4410,7 @@ public class net.corda.core.schemas.MappedSchema extends java.lang.Object
 ##
 @MappedSuperclass
 @CordaSerializable
-public class net.corda.core.schemas.PersistentState extends java.lang.Object implements net.corda.core.schemas.StatePersistable
+public class net.corda.core.schemas.PersistentState extends java.lang.Object implements net.corda.core.schemas.DirectStatePersistable
   public <init>()
   public <init>(net.corda.core.schemas.PersistentStateRef)
   @Nullable
--- a/node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
+++ b/node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
@@ -544,9 +544,9 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
             }
             predicateSet.add(joinPredicate)
 
-                        // resolve general criteria expressions
-                        @Suppress("UNCHECKED_CAST")
-                        parseExpression(entityRoot as Root<L>, criteria.expression, predicateSet)
+            // resolve general criteria expressions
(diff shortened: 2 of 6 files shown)


rick-r3:
If @mnesbit is around, we need him to bless the API diff.

.ci/api-current.txt
```diff
@@ -4398,6 +4398,14 @@ public static class net.corda.core.schemas.CommonSchemaV1$LinearState extends ne
   public void setParticipants(java.util.Set<net.corda.core.identity.AbstractParty>)
   public void setUuid(java.util.UUID)
 ##
+public interface net.corda.core.schemas.DirectStatePersistable extends net.corda.core.schemas.StatePersistable
```
rick-r3: We have to add these as the API checker seems to flag up new interfaces.  Confirmed they should NOT be `@DoNotImplement` since they are intended to be used in advanced token use cases.

.ci/api-current.txt
```diff
@@ -4410,7 +4418,7 @@ public class net.corda.core.schemas.MappedSchema extends java.lang.Object
 ##
 @MappedSuperclass
 @CordaSerializable
-public class net.corda.core.schemas.PersistentState extends java.lang.Object implements net.corda.core.schemas.StatePersistable
+public class net.corda.core.schemas.PersistentState extends java.lang.Object implements net.corda.core.schemas.DirectStatePersistable
```
rick-r3: Slight change of hierarchy (`DirectStatePersistable` is inserted between this class and `StatePersistable`).  But should be no impact as the abstract methods on `DirectStatePersistable` are implemented here.

node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
```diff
@@ -536,12 +537,16 @@ class HibernateQueryCriteriaParser(val contractStateType: Class<out ContractStat
...
+                        criteriaBuilder.equal(vaultStates.get<PersistentStateRef>("stateRef"), entityRoot.get<IndirectStatePersistable<*>>("compositeKey").get<PersistentStateRef>("stateRef"))
+                    } else {
+                criteriaBuilder.equal(vaultStates.get<PersistentStateRef>("stateRef"), entityRoot.get<PersistentStateRef>("stateRef"))
+            }
             predicateSet.add(joinPredicate)
 
-            // resolve general criteria expressions
-            @Suppress("UNCHECKED_CAST")
-            parseExpression(entityRoot as Root<L>, criteria.expression, predicateSet)
+                        // resolve general criteria expressions
```
rick-r3: I need to fix weird indent here.

node/src/main/resources/migration/vault-schema.changelog-v8.xml
```diff
@@ -6,9 +6,6 @@
         <createTable tableName="state_party">
             <column name="output_index" type="INT"/>
             <column name="transaction_id" type="NVARCHAR(64)"/>
-            <column name="id" type="BIGINT">
```
rick-r3: This is the column I'm getting rid of.

.ci/api-current.txt
mnesbit: I am fine with diff, but I am no expert. I guess just test upgrading a cash state from 3.0 to this and then consuming

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -136,8 +136,8 @@ class NodeVaultService(
                 val constraintInfo = Vault.ConstraintInfo(stateAndRef.value.state.constraint)
                 // Save a row for each party in the state_party table.
                 // TODO: Perhaps these can be stored in a batch?
-                stateOnly.participants.forEach { participant ->
-                    val persistentParty = VaultSchemaV1.PersistentParty(persistentStateRef, participant)
+                stateOnly.participants.groupBy { it.owningKey }.forEach { participants ->
```
josecoll: I presume this saves some additional inserts now also.

rick-r3: Correct.  We were inserting duplicates unbeknownst to us previously.  Using it as a key revealed it.

roger3cev:
Looks good - thanks @rick-r3 !

node/src/main/kotlin/net/corda/node/services/vault/HibernateQueryCriteriaParser.kt
roger3cev: There is a weird intent for all uses of parseExpression in that file - not sure why!

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
roger3cev: Good spot

PR #4448, opened 2018-12-20
[CORDA-2311] Create a subclass of CordaCaplet for the web server.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-2311


florian-f:
I think right now there's a certain urgency, we need fix the webserver before we release. We can still remove the CordaWebServerCaplet from the node later, it's not like its part of the public API.

anthonykeenan:
I agree there’s an urgency, but it’s only a case of a bit of copy and paste isn’t it? It doesn’t seem like a big change?

PR #4449, opened 2018-12-20
Fixed missing endpoint in IRS demo.

anthonykeenan:


samples/irs-demo/web/src/test/kotlin/net/corda/irs/web/api/InterestRatesSwapDemoAPI.kt
```diff
@@ -41,15 +41,4 @@ class InterestRatesSwapDemoAPI {
         logger.error("Attempt to set demodate to $newDemoDate but $msg")
         return ResponseEntity.status(HttpStatus.CONFLICT).body(msg)
     }
-
-    @GetMapping("demodate")
-    fun fetchDemoDate(): LocalDate {
```
sollecitom: Should these 2 mappings be removed?

anthonykeenan: Yes, I think so. When the test API is loaded, it's loaded alongside the main one, they only need to be in one place.

Issue #3772, opened 2018-08-12
Improve startup exception/log handling


remmeier:
Using current master branch (4.0 snapshot)

Various exceptions on startup currently get suppressed, even if DEBUG log level is enabled from the command line. For example:

```
  [ERROR] 08:45:47+0000 [main] internal.Node.validateKeyStore - IO exception while trying to validate keystore
 [ERROR] 08:45:47+0000 [main] internal.Node.run - Exception during node startup 
```
The command line parameters are `--log-to-console` and  `--logging-level=DEBUG`. 
After those two lines, no further lines are provided.

An other example would when setting the H2 address to 0.0.0.0 while omitting to set a password. Or when providing an illegal database configuration and corda is no longer able to connect.

joeldudleyr3:
Thanks. Tracked here: https://r3-cev.atlassian.net/browse/CORDA-1897.

PR #4452, opened 2018-12-21
[CORDA-1897] Improve exception handling on startup

JamesHR3:
This PR contains some changes to the logging configuration to improve the handling of exceptions in log messages. Specifically:
 - Add an extra appender that prints the full exception stack trace to console if console logging is enabled and the logging level is at most DEBUG. (Note that by default, no exception stack trace is seen even if --log-to-console is set as the default console logging level is INFO.)
 - Add logic to select the correct appender depending on the settings provided.
 - Don't print both the message and the exception message in the logs. Log event messages nearly always contain the exception message, so this results in the same message being printed twice.


JamesHR3:
The original issue was filed at https://github.com/corda/corda/issues/3772. The change is as suggested there: give more exception information on the console if debug logging is requested. Apart from that, I think the only other requirements discussion was had between Shams and I earlier today, and this is what was suggested.

(Apologies for the reviewer noise, I got confused!)

tlil:
Since you're saying that you want to:
> ... give more exception information on the console if *debug logging* is requested.

And given that this happens only when `--log-to-console` is set, that's fine by me. @sollecitom, can you complete the review, please? :-)

PR #4460, opened 2018-12-24
CORDA-2128: Moved constraints and attachments stuff out of the public API that shouldn't be there

shamsasari:


core/src/main/kotlin/net/corda/core/contracts/AttachmentConstraint.kt
```diff
@@ -34,72 +29,6 @@ annotation class NoConstraintPropagation
...
-     *  * You can transition from the [WhitelistedByZoneAttachmentConstraint] to the [SignatureAttachmentConstraint] only if all signers of the JAR are required to sign in the future.
-     *  * You can transition from a [HashAttachmentConstraint] to a [SignatureAttachmentConstraint] when the following conditions are met:
-     *  *  1. Jar contents (per entry, by hashcode) of both original (unsigned) and signed contract jars are identical
-     *  *     Note: this step is enforced in the [AttachmentsClassLoader] no overlap rule checking.
-     *  *  2. Java package namespace of signed contract jar is registered in the CZ network map with same public keys (as used to sign contract jar)
-     *
-     * TODO - SignatureConstraint third party signers.
-     */
-    @CordaInternal
-    fun canBeTransitionedFrom(input: AttachmentConstraint, attachment: AttachmentWithContext): Boolean {
```
shamsasari: This has been moved in response to https://github.com/corda/corda/pull/4261#discussion_r239508313.

shamsasari:
Yes, this is just moving stuff, and a bit of clean up of some methods. There was no functional changes.

>Also, it seems to be a bit of a dark art to know where stuff belongs.
>Do you know if we have a guide?

Where internal stuff belongs or knowing if something should be public or not? For internal stuff, as long as it's in a `.internal.` package then it doesn't really matter. We can always move it later.

PR #4461, opened 2018-12-24
CORDA-2345: Do not re-install CorDapps if possible in MockNetwork

vkolomeyko:
Since CorDapps directory is shared for all nodes - there is no need to re-install CorDapps upon node's restart.
Moreover, on some OSes (Windows) this will fail as we will be trying to replace a file that is open for
reading by different node class loader.

This change should fix failing `master` Windows build:
https://ci-master.corda.r3cev.com/viewLog.html?buildId=134406&buildTypeId=Corda_BuildWindowsTest

### DIFF
--- a/node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
+++ b/node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
@@ -268,7 +268,11 @@ class TwoPartyTradeFlowTests(private val anonymous: Boolean) {
 
             // ... bring the node back up ... the act of constructing the SMM will re-register the message handlers
             // that Bob was waiting on before the reboot occurred.
-            bobNode = mockNet.createNode(InternalMockNodeParameters(bobAddr.id, BOB_NAME))
+            // NB: Since CorDapps directory is shared for all nodes - there is no need to re-install CorDapps upon
+            // Bob's restart.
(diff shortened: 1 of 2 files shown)


node/src/test/kotlin/net/corda/node/messaging/TwoPartyTradeFlowTests.kt
```diff
@@ -268,7 +268,11 @@ class TwoPartyTradeFlowTests(private val anonymous: Boolean) {
 
             // ... bring the node back up ... the act of constructing the SMM will re-register the message handlers
             // that Bob was waiting on before the reboot occurred.
-            bobNode = mockNet.createNode(InternalMockNodeParameters(bobAddr.id, BOB_NAME))
+            // NB: Since CorDapps directory is shared for all nodes - there is no need to re-install CorDapps upon
```
shamsasari: This comment is incorrect - nodes do not share the cordapps directory. Each one has its own "cordapps" in their baseDirectory.

vkolomeyko: Having looked more into the code - I agree that CorDapps for different nodes are different.

I think (much) more work will be needed to ensure that the node can be re-started within same VM with no "traces" left. Classloader is used in many contexts like Serialization initialisation and currently there is no way to un-initialise it.
I am closing this PR without merging.

PR #4479, opened 2018-12-31
Closing the app class loader on node shutdown

shamsasari:
This is an attempted follow-up to https://github.com/corda/corda/pull/4461.


shamsasari:
This fixes the broken windows build: https://ci-master.corda.r3cev.com/viewLog.html?buildId=134998&buildTypeId=Corda_BuildWindowsTest

PR #4482, opened 2018-12-31
CORDA-2331 - Code alignment to produce identical finance contracts JAR.

josecoll:
This small PR ensures that jar manifest metadata is identical across OS and ENT for the finance contracts jar.

### DIFF
--- a/finance/contracts/build.gradle
+++ b/finance/contracts/build.gradle
@@ -39,12 +39,19 @@ task testJar(type: Jar) {
 
 jar {
     baseName 'corda-finance-contracts'
+    manifest {
+        attributes('Corda-Revision': 'n/a')
+        attributes('Corda-Vendor': 'Corda Open Source')
+    }
 }
 
 artifacts {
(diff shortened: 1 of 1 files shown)


finance/contracts/build.gradle
```diff
@@ -39,12 +39,19 @@ task testJar(type: Jar) {
 
 jar {
     baseName 'corda-finance-contracts'
+    manifest {
+        attributes('Corda-Revision': 'n/a')
```
shamsasari: I don't like this. CORDA-2331 shouldn't have to worry about an ENT version of finance - we should just use the OS version for both ENT and OS nodes.

josecoll: This is required because the ENT commit version will always be different to the OS commit version, and thus, cause a difference JAR hash.

josecoll: Right but we've already had a long discussion about this - and the verdict was we are not removing samples code from the ENT repo right now.

PR #4483, opened 2019-01-01
CORDA-1942: StartedMockNode.registerResponderFlow simplified

shamsasari:
The ResponderFlowFactory parameter is not used and so removed. Also, instead of returning a Future it returns an Observable of responder flows, to support multiple invocations.


shamsasari:
Removed `ResponderFlowFactory` from the API file, which is not an issue as it didn't exist in V3 (https://github.com/corda/corda/blob/release-V3/.ci/api-current.txt). I also remembered that we already have a `registerInitiatedFlow` and so `registerResponderFlow` renamed to be an overload of that to keep the naming scheme consistent.

PR #4485, opened 2019-01-02
[CORDA-2374] Avoid using Unicode in the Progress Tracker output on Windows

JamesHR3:
Windows currently does not provide full Unicode support in terminals. As a result, when running a flow, the progress tracker will display a ? next to completed or in progress steps, making it difficult to see what is happening. This PR changes the marker to ASCII text, which ensures it is displayed correctly.

Changes:
 - Detect when a terminal is run on Windows and change the markers displayed to text.
 - Fix an issue where the current step would be displayed as completed. This means that the final "DONE" step in the progress tracker will be marked as current when the flow finishes.
 - Update the tests to account for the new output.


tools/shell/src/main/kotlin/net/corda/tools/shell/utlities/ANSIProgressRenderer.kt
```diff
@@ -226,7 +229,7 @@ object StdoutANSIProgressRenderer : ANSIProgressRenderer() {
 
     override fun setup() {
         AnsiConsole.systemInstall()
-
+        usingUnicode = !SystemUtils.IS_OS_WINDOWS
```
shamsasari: What about `Emoji.hasEmojiTerminal`?

PR #4491, opened 2019-01-02
CORDA-2345: Updated docs to use the new TestCordapp API, rather than the old scan cordapp packages

shamsasari:
Also made some improvements to the API, especially for Java users.


docs/source/api-contract-constraints.rst
```diff
@@ -311,70 +311,6 @@ For example:
 For Contracts that are annotated with ``@NoConstraintPropagation``, the platform requires that the Transaction Builder specifies
 an actual constraint for the output states (the ``AutomaticPlaceholderConstraint`` can't be used) .
 
-
```
shamsasari: I deleted this entire bit as it's redundant and doesn't inform the dev of anything new. If they're worrying about contract constraints then they probably already know how to use the MockNetwork and driver.

shamsasari:
Can you re-approve, the build should be green soon.

PR #4494, opened 2019-01-03
CORDA-2098: State pointers are now comparable.

roger3cev:
They were not data classes before... Not sure why but they probably should be!


sollecitom:
@roger3cev as discussed, mind avoiding data classes as part of the public API?

roger3cev:
Yup, don't merge this for now. I will update this PR to remove the data classes and explicitly make them comparable.

PR #4500, opened 2019-01-03
CORDA-2392 Align sha256 function between OS and ENT

josecoll:
Align code such that CorDapps that use this public API are guaranteed to generate the same JAR hash.


josecoll:
@rick-r3 @shamsasari @mikehearn 
I would like to merge this PR to align our `core` modules such that in a future version of Corda we can just completely remove the ENT `core` module (after extracting out any ENT specific code - which I pointed out in a separate JIRA story (https://r3-cev.atlassian.net/browse/ENT-2928).

josecoll:
Deferred until further notice.

PR #4503, opened 2019-01-03
ENT-2927 - Create test finance contract jar

roastario:
Break out the dummy contracts and schemas previously contained within the test-jar artifact of finance into a proper separate module.


build.gradle
```diff
@@ -366,6 +366,7 @@ bintrayConfig {
             'corda-finance',    // maintained for backwards compatibility only
             'corda-finance-workflows',
             'corda-finance-contracts',
+            'corda-finance-test-contracts',
```
sollecitom: Why do we have to publish this? Used in enterprise?

finance/contracts/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
     // cordapp project(':finance:workflows')
     // cordapp project(':finance:contracts')
     cordaCompile project(':core')
+    cordaCompile project(':core')
```
sollecitom: Remove duplicate.

build.gradle
roastario: yup

finance/contracts/build.gradle
roastario: done

josecoll:
That looks a lot better now.
I think we will be doing away with those SampleCashSchema's in the near future anyway (the concept of schema evolution through explicit object versioning was conceived before we had liquidbase for schema management)

PR #4506, opened 2019-01-04
[CORDA-1596] Improve error message when CorDapp schema changes

JamesHR3:
Add a new exception when a new CorDapp version includes schema changes, and the new schema has not been initialised. The error message indicates that the node should be run with the option to initialise the new schema set to true.

Changes:
- Add a new exception for the above, plus a generic exception if a different error is encountered while constructing Hibernate config
- Add a test case that checks the right error is thrown if the schema changes


node/src/test/kotlin/net/corda/node/services/persistence/HibernateConfigurationTest.kt
```diff
@@ -963,4 +964,26 @@ class HibernateConfigurationTest {
     private fun toStateRef(pStateRef: PersistentStateRef): StateRef {
         return StateRef(SecureHash.parse(pStateRef.txId), pStateRef.index)
     }
+
+    @Test(expected = HibernateSchemaChangeException::class)
+    fun `schema change`() {
+        fun createNewDB(schemas: Set<MappedSchema>, initialiseSchema: Boolean = true) {
```
szymonsztuka: What about returning database object from ``createNewDB ``, then we can properly close it ...

node/src/test/kotlin/net/corda/node/services/persistence/HibernateConfigurationTest.kt
```diff
@@ -963,4 +964,26 @@ class HibernateConfigurationTest {
...
+                }
+            }
+            database = configureDatabase(dataSourceProps, DatabaseConfig(initialiseSchema = initialiseSchema), identityService::wellKnownPartyFromX500Name, identityService::wellKnownPartyFromAnonymous, schemaService)
+            database.transaction {
+                vaultFiller.fillWithSomeTestCash(100.DOLLARS, issuerServices, 4, issuer.ref(1), rng = Random(0L))
+            }
+        }
+
+        createNewDB(setOf(CashSchemaV1, SampleCashSchemaV1))
+        createNewDB(setOf(CashSchemaV1, SampleCashSchemaV1, SampleCashSchemaV2), initialiseSchema = false)
```
szymonsztuka: ...and here something like that (``use { ... }``will close ``CordaPersistence`` objects) :
````
 createNewDB(setOf(CashSchemaV1, SampleCashSchemaV1)).use {
            it.transaction {
                     vaultFiller.fillWithSomeTestCash(100.DOLLARS, issuerServices, 4, issuer.ref(1), rng = Random(0L))
            }
            createNewDB(setOf(CashSchemaV1, SampleCashSchemaV1, SampleCashSchemaV2), initialiseSchema = false).use {
                assertThatThrownBy {
                    it.transaction {
                        vaultFiller.fillWithSomeTestCash(100.DOLLARS, issuerServices, 4, issuer.ref(1), rng = Random(0L))
                    }
                }.isInstanceOf(HibernateSchemaChangeException::class.java)
            }
        }
````

So the exact code when the error should be thrown is checked (the second db initlisation), also this would melas obsolete ``(expected = HibernateSchemaChangeException::class)`` at the top of the test.

JamesHR3: I think the first use needs to be an apply instead, as closing the database results in creating an entirely new one (rather than changing the schema on the old one). I've done this in the newest version - let me know if you think this is incorrect.

PR #4508, opened 2019-01-04
CORDA-2345: Removed TestCordapp.scanPackage to allow in future differ…

shamsasari:
…ent ways of finding CorDapps other can package scanning


shamsasari:
This should go green now, if you could re-approve.

PR #4130, opened 2018-10-29
CORDA-1489 Exposure of node internals in mock network

blsemo:
Introduce public subset of `NodeConfiguration` inside the mock network subproject to tweak config via mock net work without exposing internal node config.

PR #4509, opened 2019-01-04
ENT-2822: Move experimental raft and bft-smart notaries back into node, fix reference state support

adagys:
To maintain config backwards compatiblity.

Moved some of the notary setup logic from `AbstractNode` into a separate `NotaryLoader`

I ended up undoing some of the work from #4130 as I believe using `InternalMockNodeParameters` does not actually leak internal config.


node/src/main/kotlin/net/corda/node/services/config/NodeConfiguration.kt
```diff
@@ -141,15 +141,17 @@ data class NotaryConfig(
...
         /**
          * If the wait time estimate on the internal queue exceeds this value, the notary may send
          * a wait time update to the client (implementation specific and dependent on the counter
          * party version).
          */
         val etaMessageThresholdSeconds: Int = NotaryServiceFlow.defaultEstimatedWaitTime.seconds.toInt(),
         /** Notary implementation-specific configuration parameters. */
-        val extraConfig: Config? = null
+        val extraConfig: Config? = null,
+        val raft: Config? = null,
```
adagys: I'll fix this to be actual config types in a followup PR, this isn't needed for the release

core/src/test/kotlin/net/corda/core/node/NetworkParametersTest.kt
```diff
@@ -66,8 +67,14 @@ class NetworkParametersTest {
     // Notaries tests
     @Test
     fun `choosing notary not specified in network parameters will fail`() {
-        val fakeNotary = mockNet.createNode(InternalMockNodeParameters(legalName = BOB_NAME,
-                configOverrides = MockNodeConfigOverrides(notary = MockNetNotaryConfig(validating = false))))
+        val fakeNotary = mockNet.createNode(
```
rick-r3: Does this file need to change?  Looks like formatting.

adagys: It does, MockNetNotaryConfig needs to be replaced with a lambda. I restored the lambda parameter in the InternalMockNodeParameters so we can override any parts of the config.

rick-r3: Ah, okay.  It kinda looked the same.

docs/source/changelog.rst
```diff
@@ -104,34 +104,8 @@ Version 4.0
 
 * Introduced new optional network bootstrapper command line option (--minimum-platform-version) to set as a network parameter
 
-* BFT-Smart and Raft notary implementations have been extracted out of node into ``experimental`` CorDapps to emphasise
```
rick-r3: I think we should make sure we include `experimental` in the package names of the raft and bft notaries to indicate their experimental nature.  e.g. `net.corda.notary.experimental`.

Also put a line here saying they were moved to an experimental package in node to reflect their experimental nature.  The changelog can comment on code changes, even if they are not direct user facing.

adagys: Will do

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/InternalMockNetwork.kt
```diff
@@ -90,15 +87,15 @@ data class InternalMockNodeParameters(
         val forcedID: Int? = null,
         val legalName: CordaX500Name? = null,
         val entropyRoot: BigInteger = BigInteger.valueOf(random63BitValue()),
-        val configOverrides: MockNodeConfigOverrides? = null,
+        val configOverrides: (NodeConfiguration) -> Any? = {},
```
adagys: @blsemo just want to double-check whether this change is ok

blsemo: It should be ok in InternalMockNetwork.

PR #4514, opened 2019-01-07
[CORDA-2326] Ensure right step is rendered by AnsiProgressRenderer

JamesHR3:
The AnsiProgressRenderer maintains a set of indices in the step tree that have been completed. If the step tree changes (perhaps because a new subflow has been started under a step, changing the child progress tracker for that step), then the set of indices previously was not updated to account for the new tree. This PR ensures that the set of indices visited is updated on a step tree change.

 When a step tree change is made, the Progress Tracker will update the current step index immediately after pushing the step tree update. In theory, there is therefore an issue where the step index update is received before the step tree update, and this is caught up (incorrectly) in the remapping. In practice, I was unable to see this happening. As accounting for this would be significantly more complicated, this has been left for now.


shamsasari:
Do you have an example of this bug that I can see?

JamesHR3:
You can see the problem by running the trader demo. I have a screenshot that shows the problem:
corda-2326

I also have a clip that shows the problem that I've put on the JIRA ticket.

PR #4516, opened 2019-01-07
[CORDA-2394]: Fixed ABI issue with VaultQueryCriteria.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-2394

### DIFF
--- a/core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
+++ b/core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
@@ -273,9 +273,21 @@ sealed class QueryCriteria : GenericQueryCriteria<QueryCriteria, IQueryCriteriaP
     data class VaultCustomQueryCriteria<L : StatePersistable> @JvmOverloads constructor(
             val expression: CriteriaExpression<L, Boolean>,
             override val status: Vault.StateStatus = Vault.StateStatus.UNCONSUMED,
-            override val contractStateTypes: Set<Class<out ContractState>>? = null,
-            override val relevancyStatus: Vault.RelevancyStatus = Vault.RelevancyStatus.ALL
+            override val contractStateTypes: Set<Class<out ContractState>>? = null
     ) : CommonQueryCriteria() {
+        // These extra field is handled this way to preserve Kotlin wire compatibility wrt additional parameters with default values.
+        constructor(
+                expression: CriteriaExpression<L, Boolean>,
+                status: Vault.StateStatus = Vault.StateStatus.UNCONSUMED,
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
```diff
@@ -94,12 +94,39 @@ sealed class QueryCriteria : GenericQueryCriteria<QueryCriteria, IQueryCriteriaP
...
+                constraints: Set<Vault.ConstraintInfo> = emptySet(),
+                participants: List<AbstractParty>? = null
+        ) : this(status, contractStateTypes, stateRefs, notary, softLockingCondition, timeCondition) {
+            this.relevancyStatus = relevancyStatus
+            this.constraintTypes = constraintTypes
+            this.constraints = constraints
+            this.participants = participants
+        }
+
+        override var relevancyStatus: Vault.RelevancyStatus = Vault.RelevancyStatus.ALL
```
shamsasari: Why not leave the primary c'tor with all the properties and then have this additional V3 c'tor (with the same default values)? That should avoid the need to have these private setters.

Alex-Koller-R3:
The fix has resolved the problem I was seeing with BNMS

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
sollecitom: As discussed f2f, that doesn't work.

core/src/main/kotlin/net/corda/core/node/services/vault/QueryCriteria.kt
```diff
@@ -249,12 +296,14 @@ sealed class QueryCriteria : GenericQueryCriteria<QueryCriteria, IQueryCriteriaP
         fun copy(
                 expression: CriteriaExpression<L, Boolean> = this.expression,
                 status: Vault.StateStatus = this.status,
-                contractStateTypes: Set<Class<out ContractState>>? = this.contractStateTypes
+                contractStateTypes: Set<Class<out ContractState>>? = this.contractStateTypes,
+                relevancyStatus: Vault.RelevancyStatus = this.relevancyStatus
```
shamsasari: Why is this added?

sollecitom: Because it was removed by the primary constructor of the data class, so the "primary" copy is now the one without, and this is putting the one with additional fields back in place.

shamsasari: Yes, sorry I was confused by the other class.

PR #4518, opened 2019-01-07
CORDA-2398 - Include private network ID in node info publish request [MERGE POST C4 CUT]

oknowles:
This PR includes a small change to add in a node's private network UUID (if exists) via a header during the node info publish request to the Network Map service.


gendal:
what happens if a node doesn't upgrade to get this feature? they can't use a private network if the doorman has been upgraded? to which version?  if this will require existing customers (eg finastra) to do another node upgrade once SSZ is enabled on CN, for example, then this is likely to be a non-starter.....

shamsasari:
What's the status of this PR? It has been touched in over 3 months.

fenryka:
@shamsasari this still needs adding to master @oknowles can you revive this for C5

oknowles:
Apologies I did not know this was still open! I am closing this PR to avoid building on a deprecated feature. The Network Services changes were built out with the assumption that this header was not present so the change is not needed regardless.

PR #4519, opened 2019-01-07
CORDA-2395 Add cordapp code signing dev key to production blacklist.

josecoll:
Add the DEV Code Signing key to the node fingerprint blacklist such that a Node cannot use artifacts signed by this key in a production environment (eg. devMode = false).

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
@@ -9,7 +9,9 @@ import net.corda.core.identity.PartyAndCertificate
 import net.corda.core.internal.hash
 import net.corda.core.internal.toX500Name
 import net.corda.nodeapi.internal.config.CertificateStore
-import net.corda.nodeapi.internal.crypto.*
+import net.corda.nodeapi.internal.crypto.CertificateAndKeyPair
+import net.corda.nodeapi.internal.crypto.CertificateType
+import net.corda.nodeapi.internal.crypto.X509Utilities
 import org.bouncycastle.asn1.x509.GeneralName
 import org.bouncycastle.asn1.x509.GeneralSubtree
 import org.bouncycastle.asn1.x509.NameConstraints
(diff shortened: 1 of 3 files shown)


josecoll:
Tested as per description in https://r3-cev.atlassian.net/browse/CORDA-2375 comment.

node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
```diff
@@ -102,7 +104,12 @@ const val DEV_CA_TRUST_STORE_FILE: String = "cordatruststore.jks"
 const val DEV_CA_TRUST_STORE_PASS: String = "trustpass"
 const val DEV_CA_TRUST_STORE_PRIVATE_KEY_PASS: String = "trustpasskeypass"
 
-val DEV_PUB_KEY_HASHES: List<SecureHash.SHA256> get() = listOf(DEV_INTERMEDIATE_CA.certificate, DEV_ROOT_CA.certificate).map { it.publicKey.hash.sha256() }
+// A code signing policy is currently under discussion (https://r3-cev.atlassian.net/wiki/spaces/SEC/pages/859996195/Code+Signing+Certificates+-+options+for+discussion)
```
jamesbr3: do we really want to include links to internal documentation in the OS codebase?

node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
```diff
@@ -102,7 +104,12 @@ const val DEV_CA_TRUST_STORE_FILE: String = "cordatruststore.jks"
...
+// The following interim key represents a self-signed certificate produced using the Java keytool and located in the gradle cordapp plugins resources key store:
+// https://github.com/corda/corda-gradle-plugins/blob/master/cordapp/src/main/resources/certificates/cordadevcodesign.jks
+const val DEV_CORDAPP_CODE_SIGNING_STR = "AA59D829F2CA8FDDF5ABEA40D815F937E3E54E572B65B93B5C216AE6594E7D6B"
```
jamesbr3: I'd prefer to see a certificate in a trust-store rather than being hard-coded. Certificates can & do expire so being able to control them via configuration would be preferable imho.

jamesbr3: what does the DEV_CORDAPP_CODE_SIGNING_STR value represent, and how was it generated please?

node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
josecoll: I'll remove the link.

node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
josecoll: The reason is to *not* have multiple copies of the truststore file.
There should be one source and that is in corda-gradle-plugins (for now).
Once we get a proper policy & process in place this can be removed also.

josecoll: Please see the associated JIRA for instructions how how the string was generated.

josecoll:
@jamesbr3 please note this is a temporary solution until we have a proper code signing policy and process in place.

node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
jamesbr3: ok, understood. No problem - just wanted to flag this up as a potential todo for the future

node-api/src/main/kotlin/net/corda/nodeapi/internal/KeyStoreConfigHelpers.kt
```diff
@@ -102,7 +104,12 @@ const val DEV_CA_TRUST_STORE_FILE: String = "cordatruststore.jks"
...
+
+val DEV_PUB_KEY_HASHES: List<SecureHash.SHA256> get() = listOf(DEV_INTERMEDIATE_CA.certificate, DEV_ROOT_CA.certificate).map { it.publicKey.hash.sha256() } + SecureHash.parse(DEV_CORDAPP_CODE_SIGNING_STR).sha256()
```
jamesbr3: I just noticed this code (I realise it is not new btw) - 
is this calculating the hash twice? i.e. `publicKey.hash` will return a SHA256 hash, but then we make a second explicit call to `sha256()` ?

josecoll: That is a damn good observation.
Let me fix that now.

PR #4520, opened 2019-01-07
CORDA-2388 Deprecate `CommandWithParties.signingParties`

szymonsztuka:
https://r3-cev.atlassian.net/browse/CORDA-2388
`CommandWithParties.signingParties` should not be used in contract verification code as it is non-deterministic.
 Replaced one remaining usage of ``signingParties`` with ``signers`` in Obligation following other examples (e.g. around https://github.com/corda/corda/blob/4aaefb4fe9dcb3eb2fe55f0fb7cf2b3216cd8f31/finance/contracts/src/main/kotlin/net/corda/finance/contracts/asset/Cash.kt#L197).
 No other contracts use ``signingParties`` in their logic.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/Structures.kt
@@ -237,7 +237,7 @@ interface MoveCommand : CommandData {
 data class CommandWithParties<out T : CommandData>(
         val signers: List<PublicKey>,
         /** If any public keys were recognised, the looked up institutions are available here */
-        @Deprecated("Should not be used in contract verification code as it is non-deterministic, will be disabled in the future Corda release.")
+        @Deprecated("Should not be used in contract verification code as it is non-deterministic, will be disabled for some future target platform version onwards and will take effect only for CorDapps targeting those versions.")
         val signingParties: List<Party>,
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/net/corda/core/contracts/Structures.kt
```diff
@@ -237,6 +237,7 @@ interface MoveCommand : CommandData {
 data class CommandWithParties<out T : CommandData>(
         val signers: List<PublicKey>,
         /** If any public keys were recognised, the looked up institutions are available here */
+        @Deprecated("Should not be used in contract verification code as it is non-deterministic, will be disabled in the future Corda release.")
```
rick-r3: We should re-word this.  

The first part is fine and accurate.   We will never disable it though.  In reality, it will be disabled for CorDapps with a target platform that is some future target platform version.  i.e.

rick-r3: I'm not sure how to phrase that quite right though.

rick-r3: disabled by -> disabled for

effect only CorDapps -> effect only for CorDapps

PR #4521, opened 2019-01-07
CORDA-2399: Samples using public TestCordapp API rather than internal one

shamsasari:
Also moved the contents of TestCordappUtils.kt to InternalTestUtils.kt to make it more obvious they're internal.


joeldudleyr3:
@shamsasari Thanks. Why does the trader demo still depend on the internal `FINANCE_CORDAPPS`?

shamsasari:
I was thinking of whether to expand it out and do the same thing that `FINANCE_CORDAPPS` does in the test. Though the list is internal, it's created using public APIs. Let me know if you still want this removed.

PR #4532, opened 2019-01-09
CORDA-2406: FinalityFlow can support mix of participants using new and old APIs

shamsasari:
Otherwise it's impossible to finalise a transaction and the participants are a mix of those using the new API and those using the old.


core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
```diff
@@ -29,40 +30,56 @@ import net.corda.core.utilities.ProgressTracker
...
+     * @param transaction What to commit.
+     * @param sessions A collection of [FlowSession]s for each participant.
+     * @param oldParticipants An **optional** collection of parties for participants who are still using the old API.
+     *
+     * You will only need to use this parameter if you have upgraded your CorDapp from the V3 FinalityFlow API but are required to provide
+     * backwards compatibility with participants running V3 nodes. If you're writing a new CorDapp then this does not apply and this
+     * parameter should be ignored.
+     */
+    @JvmOverloads
+    constructor(
```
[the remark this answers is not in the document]
shamsasari: Yes, forgot to. I added this c'tor late.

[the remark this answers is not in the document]
shamsasari: Sorry, ignore me, that doesn't make sense.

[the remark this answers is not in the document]
shamsasari: C'tor deprecated.

shamsasari:
Back to you.

core/src/main/kotlin/net/corda/core/flows/FinalityFlow.kt
```diff
@@ -29,44 +30,74 @@ import net.corda.core.utilities.ProgressTracker
...
+     * backwards compatibility with participants running V3 nodes. If you're writing a new CorDapp then this does not apply and this
+     * parameter should be ignored.
+     */
+    @Deprecated(DEPRECATION_MSG)
+    constructor(
+            transaction: SignedTransaction,
+            sessions: Collection<FlowSession>,
+            oldParticipants: Collection<Party>,
+            progressTracker: ProgressTracker
+    ) : this(transaction, oldParticipants, progressTracker, sessions, true)
```
rick-r3: Why is a constructor using the new API deprecated already?

rick-r3: Oh, is this because the constructor allows a mixed network?

core/src/test/kotlin/net/corda/core/flows/FinalityFlowTests.kt
```diff
@@ -75,15 +76,29 @@ class FinalityFlowTests : WithFinality {
...
-        assertThat(bob.services.validatedTransactions.getTransaction(stx.id)).isNotNull()
+        assertThat(oldBob.services.validatedTransactions.getTransaction(stx.id)).isNotNull()
+    }
+
+    @Test
+    fun `broadcasting to both new and old participants`() {
+        val newCharlie = mockNet.createNode(InternalMockNodeParameters(legalName = CHARLIE_NAME))
+        val oldBob = createBob(cordapps = listOf(tokenOldCordapp()))
+        val stx = aliceNode.issuesCashTo(oldBob)
+        val resultFuture = aliceNode.startFlowAndRunNetwork(FinalityInvoker(
```
rick-r3: Is there a mechanism for actually figuring out who needs to use the old version of the flows?  You can see what version node is in the network map.  How do folks know what version the app on the other side is?  I guess they need to squirrel away some info from a `FlowSession` used during coordination of the transaction?   A little tricky.  I suppose observer nodes would be the most tricky.

shamsasari: You use flow versioning. I added an example in the migration guide as part of the original PR: https://docs.corda.net/head/app-upgrade-notes.html#step-3-security-upgrade-your-use-of-finalityflow.

shamsasari:
Back to you @rick-r3

PR #4536, opened 2019-01-09
CORDA-2404: SwapIdentitiesHandler is a proper responder to SwapIdentitiesFlow

shamsasari:
This enables V4 nodes to run V3 CorDapps that use confidential identities by installing the new V3 confidential-identities jar as a CorDapp.


shamsasari:
This will require an update to the V4 docs for those using confidential identities.

shamsasari:
The tests are failing because of `groovy.lang.MissingPropertyException: Could not get unknown property 'buildWebserverJar' for task set of type org.gradle.api.internal.tasks.DefaultTaskContainer.`. Has something changed on TC?

shamsasari:
Actually I think this is the wrong solution. Closing for now.

PR #4537, opened 2019-01-09
finance-workflows jar contains confidential identities

shamsasari:
As highlighted in https://github.com/corda/corda/pull/4296, confidential-identities should now be bundled in with the CorDapp. For some reason this was lost when finance was split.


shamsasari:
As discussed offline, confidential-identities is meant to be a CorDapp.

Issue #3267, opened 2018-05-30
Node names that contain commas don't work


lankydan:
Using Corda V3.1

Writing a node node that contains a comma does not work since it is the delimiter character used by `X500Name`. Using a backslash "\\" didnt work, which was mentioned as a escape character in https://tools.ietf.org/html/rfc1779 (X500Name documentation refers to this rfc).

Using a backslash (double backslash) due to gradle using groovy led to a node including a backslash in its name.

```groovy
node {
        name "O=ABC\\, DEF,L=Berlin,C=DE"
        p2pPort 10050
        rpcSettings {
            address("localhost:10005")
            adminAddress("localhost:10505")
        }
}
```
Running `deployNodes` lead to a node named `ABC\` which isn't what is desired.

Changes in CordaX500Name before passing the `name` field to `X500Name` might work, but I'm only guessing here.

joeldudleyr3:
This is a known limitation. However, I am not aware of a reason why it could not be lifted.

mikehearn:
What sort of node names need commas in them? Are you thinking of company names that have commas?

Yes, we could implement escaping everywhere, but we've tried to be restrictive when it comes to "special" characters in X.500 names because identity is so critical and such a sensitive attack surface. We're still not as tight as we should be, but, note that some rules are up to the zone operator to enforce.

mikehearn:
Reopening - we should at least document the limitations clearly and produce good error messages. And if there's a simple set of fixes, let's do that too.

X.500 is natively binary so it's probably only at the edges that commas become special.

lolo8304:
I take this one @joeldudleyr3

lolo8304:
on master. I have tested

String
`O=Sue\, Grabbit and Runn, L=VALID, C=CH
`
X500Principal
`O="Sue, Grabbit and Runn", L=VALID, C=CH
`
X500Name —> RDNs have 2 String representation
`RFC 1779 = "Sue, Grabbit and Runn"
`
`RFC 2253 = Sue\, Grabbit and Runn
`
in todays CordaC500Name we are using toString() instead of a standardised one based on RFC1779 as in X500Name from sun (internal library)

joeldudleyr3:
@lolo8304 So are you saying that commas in Corda X500 names are allowed and work?

joeldudleyr3:
@lolo8304 I'd suggest opening a PR so we can take a look at the changes, even if you're not sure they pass the tests for now. This will also get TeamCity to run the tests for you.

[...]

PR #4539, opened 2019-01-09
3267 - Node names that contain commas don't work - use RFC1779 for CordaX500Name to verify principles

lolo8304:
from bug https://github.com/corda/corda/issues/3267
 
special characters in X500Name are supported but today CordaX500Name does not supports it while validating the minimal requirements of a X500name for corda.
Using the RFC1779 and the representation of an attribute (e.g. O) we are able represent the x500name in corda own class as it is in the standard X500Name of our security library

# PR Checklist:
- [x] Have you run the unit, integration and smoke tests as described [here](https://docs.corda.net/head/testing.html)? only "unit" tests
- [x] If you added public APIs, did you write the JavaDocs?
- [x] If the changes are of interest to application developers, have you added them to the [changelog](https://github.com/corda/corda/blob/master/docs/source/changelog.rst), and potentially the [release notes](https://github.com/corda/corda/blob/master/docs/source/release-notes.rst)?
- [x] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/corda/corda/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/corda/corda/blob/master/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
+++ b/core/src/main/kotlin/net/corda/core/identity/CordaX500Name.kt
@@ -6,7 +6,9 @@ import net.corda.core.internal.toAttributesMap
 import net.corda.core.internal.toX500Name
 import net.corda.core.internal.unspecifiedCountry
 import net.corda.core.serialization.CordaSerializable
+import org.bouncycastle.asn1.ASN1Encodable
 import org.bouncycastle.asn1.x500.style.BCStyle
+import org.bouncycastle.asn1.x500.style.IETFUtils
 import java.util.*
 import javax.security.auth.x500.X500Principal
 
@@ -83,13 +85,21 @@ data class CordaX500Name(val commonName: String?,
         @JvmStatic
         fun build(principal: X500Principal): CordaX500Name {
             val attrsMap = principal.toAttributesMap(supportedAttributes)
-            val CN = attrsMap[BCStyle.CN]?.toString()
-            val OU = attrsMap[BCStyle.OU]?.toString()
-            val O = requireNotNull(attrsMap[BCStyle.O]?.toString()) { "Corda X.500 names must include an O attribute" }
-            val L = requireNotNull(attrsMap[BCStyle.L]?.toString()) { "Corda X.500 names must include an L attribute" }
-            val ST = attrsMap[BCStyle.ST]?.toString()
-            val C = requireNotNull(attrsMap[BCStyle.C]?.toString()) { "Corda X.500 names must include an C attribute" }
+            val CN = toRFC1779String(attrsMap[BCStyle.CN])
+            val OU = toRFC1779String(attrsMap[BCStyle.OU])
+            val O = requireNotNull(toRFC1779String(attrsMap[BCStyle.O])) { "Corda X.500 names must include an O attribute" }
+            val L = requireNotNull(toRFC1779String(attrsMap[BCStyle.L])) { "Corda X.500 names must include an L attribute" }
+            val ST = toRFC1779String(attrsMap[BCStyle.ST])
+            val C = requireNotNull(toRFC1779String(attrsMap[BCStyle.C])) { "Corda X.500 names must include an C attribute" }
             return CordaX500Name(CN, OU, O, L, ST, C)
+        }
+
+        /*
+         * use RFC1779 String representation based on encoded values and not default toString()
+         */
+        @JvmStatic
+        private fun toRFC1779String(rdnsValue: ASN1Encodable?): String? {
+            return if (rdnsValue != null) IETFUtils.valueToString(rdnsValue) else null
         }
 
         @JvmStatic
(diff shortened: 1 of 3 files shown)


mikehearn:
Thanks for the PR! The rule against commas is deliberate for now, partly because even if Corda does all the escaping correctly it's not clear other people's software always would, and reducing the chances of obscure bugs in third party UIs etc, seemed more valuable than allowing commas. 

Which names are you looking at where the comma is essential?

mikehearn:
Well, maybe we should revisit this and decide if commas should be supported. Legal firms in particular seem to often have them. @jamescarlyle @jamesbr3 What do you both think?

jamesbr3:
There is technically no reason why a comma cannot be used as part of a common/distinguished name. However note that the comma often used as a separator between fields of the distinguished name. i.e.:

`CN=Alice, O=Alice Corp, OU=Corda, L=London,C=GB`

Any comma used as part of the CN (for example) would need to be escaped whenever the CN is read/written:

`CN="Alice, Bob", O=Acme Corp, OU=Corda, L=London,C=GB`

or

`CN=Alice\, Bob, O=Acme Corp, OU=Corda, L=London,C=GB`

(see https://tools.ietf.org/html/rfc1779)

So it's technically possible, but any tooling that reads or writes x500 names will need to support escaping properly. This applies not only to Corda, but to any tooling and business process that is used by the node operator to issue and manage identities on the network. I can envisage multiple security issues where a node or operator could be tricked by cleverly placed commas in a x500 name.

lolo8304:
Hi @jamesbr3
I can see the point. 
The Standard is clear and the verification and escape detection too. The standard java classes are doing it. 
Except our cordaX500name. 

About security. We have to harden even more the cordax500name verification process using test cases to throw exception in any strange way. There already a lot but we could use tools to generate strange names to see how the system reacts. 

And we could also allow only escaped , in the O name and not others.

lolo8304:
But based on reasearch there are other solutions (like e.g. Microsoft) also not allowing special chars
But it’s fixed since 2004

https://support.microsoft.com/en-us/help/293485/fix-error-message-when-you-try-to-generate-a-certificate-request-in-ii

shamsasari:
This PR seems to be abandoned. Can we get a final decision on this please @jamesbr3 @mikehearn.

mikehearn:
I'll escalate this one to Richard or James Carlyle. I don't particularly care either way - both blocking and allowing have downsides and it's not really clear which one is worse. My guess is few firms (and mostly legal firms) use commas in their legal name *and* care about the comma always being present. But nonetheless, they will have a legitimate desire to see their brand nicely preserved and silly programmers with their odd text-based protocols be damned!

roastario:
Can one of the admins verify this patch?

roastario:
Can one of the admins verify this patch?

gendal:
I agree the current situation is suboptimal.. it seems we're neither one thing nor another right now... as in we ban commas but don't do so consistently and don't fail as early as we could?

My instinct is that enabling commas (and other punctuation?) is a complexity we would live to regret.  Like it or not, people routinely use string representations of X500 names in their code, displaying them directly on UIs and all the rest.  Aside from the fact that we'll end up with crappy-looking UIs all over the place with escape characters printed to the screen, annoying users and giving Corda a bad name ("another piece of horrible enterprise software"!), what's the chance everybody will get the escaping right?

It feels like just making the prohibition clearer and more effectively enforced keeps things simple and avoids a source of (potential) exploits should somebody figure out a way to use a poorly-escaped comma to confuse a user or do something else nefarious.

@jamescarlyle is there a demand from Corda Network to allow commas?

jamesbr3:
I'm OK with this change - fundamentally Corda should support X500 properly, and currently we do not it seems. If we identify a down-stream security issue then that can be fixed. 
This change brings in a dependency on bouncy castle - are we ok with that?

shamsasari:
>This change brings in a dependency on bouncy castle - are we ok with that?

The dependency already exists.

lolo8304:
Thanks for the great and long debate. The PR was done at times where 3.3 was the active release and 4 was coming. 
Stallmist validate it again with the current master and will you integrate it into other branches? 
Unser some of the checks for PR were not successful

vkolomeyko:
Do we still want to pursue this change? It has been open for a long time and `master` branch is no longer a valid target for merge.

mnesbit:
Unfortunately, because of the way that the network map communicates these names and interactions with other software components such as the doorman and network map we cannot make this class of change. We accidentally enlarged a DB column at one point and broke a sub zone, so we could only apply such changes as part of a major platform upgrade.

jamesbr3:
I'd still argue for making this change and being standards-compliant. NetworkMap and/or Doorman could maintain network compatibility by refusing to issue newer (compliant) x500 names until everyone was on the right version to support it?

mnesbit:
We really cannot deliver this and maintain any kind of compatibility promise. Sorry.

PR #4544, opened 2019-01-10
ENT-2822: Minor – add a note about experimental notary upgradability

adagys:
Do we need to expand on this in the release notes?


adagys:
Ping @mikehearn

PR #4546, opened 2019-01-10
[CORDA-2385] Ensure codesets are included in documentation

JamesHR3:
There are a number of documentation pages that are using codeset containers but not including the relevant .js file at the top of the file. This prevents those elements from being rendered (failing silently). This PR adds a set of headers to the top of these files that brings them in line with other documentation pages.


GavT:
@JamesHR3 I don't see this fix applied to Enterprise docs docs.corda.r3.com/head.

GavT:
Ah yes, the docs are correct at https://docs.corda.r3.com/releases/master/api-states.html

PR #4552, opened 2019-01-11
CORDA-2426 Traverse an object graph, avoiding cycles

r3domfox:
Introduce an `ObjectGraphTraverser` which uses the type analysis from the serialization framework to traverse an object graph in depth-first order, emitting a `Sequence` of objects in the graph:

```kotlin
val peopleInObjectGraph = ObjectGraphTraverser.traverse(object) { it is Person }.toList()
```

To support code which must run in `core`, the `LocalTypeModel` code and related classes have been separated from the `serialization` context and now stand alone, in `net.corda.core.internal.reflection`.

The remaining classes in `serialization.amqp` have been re-organised with some additional packages being introduced - the aim was to clarify the dependencies between groups of classes and eliminate cyclic dependencies between packages, but some cycles remain which it would be troublesome to remove.

By re-implementing the very small part of `TypeToken` that is actually used in `serialization`, we remove the dependency on `TypeToken` from both `core` and `serialization`.


serialization/src/main/kotlin/net/corda/serialization/internal/model/PropertyEnumerator.kt
```diff
@@ -0,0 +1,38 @@
...
+    override fun invoke(value: Any): Sequence<Pair<String, Any?>> =
+            readers.asSequence().map { (name, reader) -> name to reader.read(value) }
+}
+
+class ObjectGraphTraverser(private val typeInformationProvider: (Type) -> LocalTypeInformation) {
+
+    constructor(factory: LocalSerializerFactory) : this(factory::getTypeInformation)
+
+    private val propertyEnumeratorCache = DefaultCacheProvider.createCache<TypeIdentifier, PropertyEnumerator>()
+    private val alreadySeen: MutableSet<Any> = mutableSetOf()
```
shamsasari: This will need to be `Collections.newSetFromMap(IdentityHashMap<E, Boolean>())`

core/build.gradle
```diff
@@ -81,7 +81,7 @@ dependencies {
     testCompile "org.assertj:assertj-core:${assertj_version}"
 
     // Guava: Google utilities library.
-    testCompile "com.google.guava:guava:$guava_version"
+    compile "com.google.guava:guava:$guava_version"
```
shamsasari: This reverses https://github.com/corda/corda/pull/3569

r3domfox: The only solution here would be to re-implement the parts of guava (certain parts of `TypeToken`) that perform type resolution - which might not be a bad thing, in fact.

r3domfox: Done (the needed parts are quite small, and are in `TypeResolver`).

r3domfox:
To traverse an object graph, we need to identify _when_ an object has readable properties which we can traverse, and _what_ those properties are. There is logic to do this encapsulated in `LocalTypeModel` which obtains/caches this information for types and makes it available as `LocalTypeInformation`. This is what is used by `SerializerFactory` in `serialization` to build serializers for POJO-like classes.

`ObjectGraphTraverser` uses a `LocalTypeModel` to get `LocalTypeInformation` for the objects it traverses. It is not itself used by `serialization`.

In this PR, we move `LocalTypeModel` across from `serialization` to `core` so that `ObjectGraphTraverser` can live in `core` and be used by `StatePointerSearch`. `LocalTypeModel` knows about things like calculated properties, for example, and can distinguish between a method called `getX` which exposes a field value and a method called `getX` which performs some (possibly side-effecting) calculation, so it's preferable to make use of this information in object graph traversal, rather than doing ad hoc reflection (and, ultimately, duplicating a lot of the work that `LocalTypeInformationBuilder` does).

`serialization` will depend on `core` for as long as `core` defines the top-level serialization API. If we later choose to make it stand-alone, then the `net.corda.core.internal.reflection` package to which `LocalTypeModel` and friends have been moved can be migrated back into `serialization`, and `core` can depend on `serialization` (it can't at the moment, because that would create a circular dependency).

core/src/main/kotlin/net/corda/core/internal/reflection/PropertyEnumerator.kt
```diff
@@ -0,0 +1,46 @@
...
+        if (value == null || value in alreadySeen) return emptySequence()
+
+        alreadySeen += value
+        val typeInformation = localTypeModel.inspect(value::class.java)
+
+        val converter = propertyEnumeratorCache.getOrPut(typeInformation.typeIdentifier) {
+            PropertyEnumerator.forType(typeInformation)
+        }
+
+        return converter(value).flatMap { (_, propertyValue) -> traverse(propertyValue, filter) } +
```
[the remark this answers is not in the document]
r3domfox: Done.

shamsasari:
OK fair enough, I see the benefit.

PR #4558, opened 2019-01-11
Update Corda gradle plugins to 4.0.38

josecoll:
Includes following changes:

* `cordformation`: deployNodes does not attempt to sign any JARs unless explicitly configured to do so using `signing.enabled = true`
* `cordapp`: enforce CorDapp `workflow` and `contract` version identifier metadata to be an integer.
* `cordapp`: relax cordapp metadata checking (warn on missing metadata rather than failing-fast).
   This allows multi-module structured CorDapps to only define metadata for flows and/or contracts modules.

### DIFF
--- a/docs/source/generating-a-node.rst
+++ b/docs/source/generating-a-node.rst
@@ -140,13 +140,18 @@ To copy the same file to all nodes `ext.drivers` can be defined in the top level
 
 Signing Cordapp JARs
 ^^^^^^^^^^^^^^^^^^^^
-Cordform entry ``signing`` configures the signing of CorDapp JARs.
+The default behaviour of Cordform is to deploy CorDapp JARs "as built":
+
+ - prior to Corda 4 all CorDapp JARs were unsigned.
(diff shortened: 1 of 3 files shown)


docs/source/generating-a-node.rst
```diff
@@ -140,13 +140,18 @@ To copy the same file to all nodes `ext.drivers` can be defined in the top level
 
 Signing Cordapp JARs
 ^^^^^^^^^^^^^^^^^^^^
-Cordform entry ``signing`` configures the signing of CorDapp JARs.
+The default behaviour of Cordform is to deploy CorDapp JARs "as built":
+
+ - prior to Corda 4 all CorDapp JARs were unsigned.
+ - as of Corda 4, CorDapp JARs are signed by a Corda development certificate by default.
```
szymonsztuka: Maybe ``CorDapp JARs are signed``  -> 
``CorDapp JARs created by cordapp Gradle plugin are signed  ``

PR #4559, opened 2019-01-11
[CORDA-2402] Ensure out of order transactions result in correct vault state

JamesHR3:
Transactions processed by a node may not be processed in the order they are received. The node currently assumes that transactions are processed in order. When this doesn't happen, it is possible for states to be missing or for duplicate unconsumed states to be seen in the vault.

The fix proposed here is to ensure that all transactions are processed in order by recording any states relevant to the node that it doesn't know about during the transaction resolution process. Transaction resolution will happen whenever a transaction is sent to a node, and so even if multiple transactions are received at similar times, all threads processing transactions will attempt to store states from the earliest first.

Changes:
 - Pass the statesToRecord parameter through to ResolveTransactionFlow from ReceiveTransactionFlow, and record states in the ResolveTransactionFlow.
 - Ensure that states are always recorded, even if the ReceiveTransactionFlow passes StatesToRecord.NONE. This is necessary as a transaction will only be processed once, and after this further updates are filtered out. If, for example, a node is sent a transaction multiple times as part of a flow, and it hasn't received all prior transactions when it is sent the first time, then a transaction in the back chain may not be recorded correctly.
 - Add a test that sends transactions to a node out of order and checks that the correct number of states appear in the vault.
 - Add a test that records transactions in the wrong order and checks all expected states appear as unconsumed. (This may be superfluous.)


JamesHR3:
If I've read the description of the previous issues Mike was referring to correctly, then I think we might be ok with regard to the node explorer. This comes with the caveat that I haven't properly looked at it, so I'll dig a bit deeper this morning to double check.

My understanding is that we stopped recording states in the vault during transaction resolution because some of those transactions may be irrelevant to us, so we don't want to store those states. It looks to me that at the time the isRelevant check was missing, so states from those irrelevant transactions would be stored. This shouldn't happen today, so I think you'd end up seeing the right thing in the node explorer. (Indeed, before the fix I'd expect to see the wrong thing as there would be states missing/duplicated.)

On an unrelated note, I removed the second test I added, as I realised the fix operates before recordTransactions is called, and so the test cannot check the behaviour I want it to.

gendal:
Thanks James.

node/src/test/kotlin/net/corda/node/services/persistence/TransactionOrderingTests.kt
```diff
@@ -0,0 +1,114 @@
...
+    @Suspendable
+    override fun call() {
+        val session = initiateFlow(party)
+        subFlow(SendTransactionFlow(session, stx))
+    }
+}
+
+@InitiatedBy(SendTx::class)
+class ReceiveTx(private val otherSideSession: FlowSession) : FlowLogic<Unit>() {
+
```
rick-r3: Remove blank line.

node/src/test/kotlin/net/corda/node/services/persistence/TransactionOrderingTests.kt
```diff
@@ -0,0 +1,114 @@
...
+}
+
+@InitiatingFlow
+@StartableByRPC
+class SendTx(private val party: Party,
+             private val stx: SignedTransaction) : FlowLogic<Unit>() {
+    @Suspendable
+    override fun call() {
+        val session = initiateFlow(party)
+        subFlow(SendTransactionFlow(session, stx))
```
rick-r3: I think to be extra sure that in the tests the `ReceiveTx` has also finished `ReceiveTransactionFlow`,  we should do a receive of a token message after the `subFlow` here, and have `ReceiveTx` do a matching send of the token.  e.g. `session.receive<Unit>()`here and `session.send(Unit)` there.

rick-r3:
Hmmm.  @gendal You raise a good point.  During transaction resolution with `ALL_VISIBLE`, potentially a node will now see extra states appear in their vault.  `ALL_VISIBLE` is precisely what one of the impacted clients is using.  Now, fortunately, we now also compute the relevancy in the vault.  But for backwards compatibility the criteria for vault queries is `ALL`.  So, in summary, clients using `ALL_VISIBLE` to record states in the vault may now see addition results in vault queries that they did not before.  They can fix it, by setting the criteria "correctly" in their queries, but it will be a code change in their app.

However, this relevancy stuff is not in Corda 3.x.  So it could be a problem for the client(s) in a back port, who will also have no easy way to filter out.

JamesHR3:
Isn't a node recording transactions using ALL_VISIBLE an observer node? In which case, the node should not be both a participant in the ledger and an observer at the same time in V3 ([see here](https://docs.corda.net/releases/release-V3.3/tutorial-observer-nodes.html)). Doesn't that imply that an observer should never have any relevant states in V3 anyway? (And if they do, they're doing something unsupported.)

We don't explicitly say this isn't supported in V4. If a client is querying for relevant states before, then they'll definitely get the right thing now. If they're after all states, then they may get extra states back now, depending on exactly what transactions are sent when. This is because a transaction can only be recorded once, either as an observer or as a participant (see [here](https://docs.corda.net/head/tutorial-observer-nodes.html) for the documentation saying this and [here](https://r3-cev.atlassian.net/browse/CORDA-883) for the old JIRA ticket covering this). 

One thing I have spotted is that the last caveat in the observer node documentation for V4 is now not true - observer nodes will record states for transactions not explicitly sent to them. However, I guess in the future if someone opts into SGX this would change? Should we encourage people to always send all transactions they wish to be recorded by observers to them (so that moving to SGX things will continue to work as they expect), or should we remove that caveat and (potentially) reintroduce it further down the line?

JamesHR3:
In that case, I'll remove that last bullet point from the observer node documentation for V4. I think exactly what states are in the vault is somewhat nondeterministic. As an extreme example, if an observer node sees a transaction as a participant (recording using ONLY_RELEVANT) and as an observer (ALL_VISIBLE) at about the same time, and none of the back chain has been resolved, then some of the back chain will have all states recorded and some only relevant, more or less at random (I think). That might be worth noting for now?

For V3, as you say, this will need some discussion when the fix is backported.

rick-r3:
Maybe a better solution for the client(s) is to introduce and `AlwaysVisible` state type.

JamesHR3:
Shall I do the docs update here too? If I do I think you'll have to reapprove the PR.

shamsasari:
I'm just skimming through the discussion here, but please use `CordappResolver.currentTargetVersion` to preserve old vault querying behaviour for existing CorDapps.

docs/source/tutorial-observer-nodes.rst
```diff
@@ -47,8 +47,7 @@ Caveats
   later re-record the same transaction as an observer. This issue is tracked here:
   https://r3-cev.atlassian.net/browse/CORDA-883
 
-* Observer nodes will only record the states of the transactions sent to them, and not any states from any previous
-  transactions in the chain. If the observer node is required to follow the creation and deletion of states, then each
-  transaction in the chain involving those states must be sent individually. This is because the observer node does not
-  necessarily have any visibility into the states of intermediate transactions, and so cannot always determine whether
-  a previous state has been consumed when a new transaction is received.
+* When an observer node is sent a transaction with the ALL_VISIBLE flag set, any transactions in the transaction history
```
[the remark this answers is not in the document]
JamesHR3: Does the first bullet point in the docs sort of cover this?

"By default, vault queries do not differentiate between states you recorded as a participant/owner, and states you recorded as an observer. You will have to write custom vault queries that only return states for which you are a participant/owner. See https://docs.corda.net/api-vault-query.html#example-usage for information on how to do this. This also means that ``Cash.generateSpend`` should not be used when recording ``Cash.State`` states as an observer"

JamesHR3:
@shamsasari @roger3cev @rick-r3 There's been some discussion about preserving the behaviour of vault queries in V3 with this fix. I'd hoped coming back to this today would make things a bit clearer, but I'm still a bit stuck.

The problem is that with this fix, a node that recorded non-relevant states before (by setting ALL_VISIBLE) may now also record some extra non-relevant states as part of transaction resolution. Nodes that did not record non-relevant states before are unaffected. To preserve the old behaviour, vault queries would have to distinguish between "non relevant states that were recorded explicitly by being part of a transaction set to a node with ALL_VISIBLE set" and "non relevant states that were recorded because the transaction was discovered as part of transaction resolution for a transaction with ALL_VISIBLE set". I don't see how we can do this, particularly in V3 where the concept of relevancy for vault queries doesn't exist either.

What it boils down to is that the vault query behaviour hasn't changed with this fix. In V3, you get all states and you can't filter by relevancy. In V4 you can filter by relevancy, but the default is not to, which would do exactly what the V3 query does. In both cases, you get extra states with the fix because the recording behaviour has changed.

One thought I've just had - perhaps we should record ONLY_RELEVANT states always as part of transaction resolution? The result of that is that whether you get non-relevant states in your vault for a transaction is a bit more predictable, as you'll only get them if the transaction was explicitly received with ALL_VISIBLE set. This is something we could gate on version (i.e. if a cordapp has currentTargetVersion < 4 use ONLY_RELEVANT, otherwise use ALL_VISIBLE if it is set). Is this what you were getting at Shams?

rick-r3:
We can't hide this behind a target version, since the fix also needs to work on V3.  If the fix isn't acceptable (i.e. this change to states in the vault, some of which is intentional) then we need a different fix.

To be clear, I think we should be saying that transactions chains can't co-mingle transactions with ALL_VISIBLE and ONLY_RELEVANT in V4 unless you make sure your queries all specify the relevancy status.  In V3, the docs say you can't even mix ALL_VISIBLE and ONLY_RELEVANT in the same node.  Unfortunately I think we have clients doing so, and unfortunately we don't have a plan for preventing those from seeing changes to the content of their vaults, unless we back-track and go with another solution to resolve out of order inserts (requires database changes) for target version 3.

Some ideas going forward: I think we should deprecate StatesToRecord since it isn't on the ledger and thus is not deterministic and is causing us some pain.  A better solution would be e.g. reference data contract state type that all nodes that see it record (i.e. always relevant).  Actual observers should be in the `participants` list.  I don't know why in our privacy model other participants would not want to control who could observe.  But this is all perhaps for another time.

JamesHR3:
My suggestion was to use a slightly different version of the fix for cordapps expecting V3 behaviour that always records ONLY_RELEVANT states in resolution. I thought that would prevent the extra states in V3 as the node would only record non-relevant states for transactions received with the flag set. Does that help?

rick-r3:
Just added a comment that I put on slack: this fix does mean that the node behaves correctly under supported operation, whereas before you could have states missing.  Under unsupported operation, you now might get additional states rather than having states missing before.

mikehearn:
I'm not quite clear on why we don't just set ONLY_RELEVANT always for resolution here, why are we letting users control it?

JamesHR3:
I think the situation we were concerned with was the case where a chain of transactions was always sent to a node using ALL_VISIBLE, and two of these were received by a node close together in time. There would then be a race where the earlier transaction in the chain might have all states recorded, or it might have just the relevant ones. I think there's a concern this might break Finastra (@rick-r3 does that sound right?)

The fix ensures that if a chain of transactions is always received with ONLY_RELEVANT or ALL_VISIBLE, then you'll get the states you're expecting in your vault. If the chain is received using a mix of these, then there are situations where you might not get the (non-relevant) states you're expecting, or may get extra (non-relevant) states. This is still the case if transaction resolution always uses ONLY_RELEVANT, but in cases where the full chain should be received with ALL_VISIBLE non-relevant states could be missing.

mikehearn:
Hmm you can't pick how the target node processes transactions. It's up to the flows on the receiving side. The transactions are sent in the same way regardless.

We probably need to rethink ALL_VISIBLE completely. The whole observer node concept was a rush job for a single customer who didn't need it fast in the end anyway. I don't quite recall the reasoning behind why observers can't just be participants, would need to revisit the old PRs and emails I think. Unfortunately due to the rush this feature didn't go through the design doc process.

PR #4561, opened 2019-01-12
CORDA-2426 Fixed bug in state pointer search.

roger3cev:
This fixes some of the bugs in state pointer search, see JIRA: https://r3-cev.atlassian.net/browse/CORDA-2426

Pointers are now searched for inside iterables and maps but this fix still doesn't search for state pointers inside computed properties.

This code will just be used temporarily until Dominic has split all the basic type model stuff from serialisation to core but that won't happen until next week.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
@@ -26,9 +26,9 @@ class StatePointerSearch(val state: ContractState) {
     // Queue of fields to search.
     private val fieldQueue = ArrayDeque<FieldWithObject>().apply { addAllFields(state) }
 
-    private fun getAllFields(clazz: Class<*>, fields: List<Field> = emptyList()): List<Field> {
+    private fun getAllFields(clazz: Class<*>, fields: List<Field>): List<Field> {
         return if (clazz.superclass != null) {
-            getAllFields(clazz.superclass, fields + clazz.declaredFields)
+            getAllFields(clazz.superclass, fields + clazz.declaredFields.asList())
         } else {
             fields
         }
@@ -36,7 +36,7 @@ class StatePointerSearch(val state: ContractState) {
 
     // Helper for adding all fields to the queue.
     private fun ArrayDeque<FieldWithObject>.addAllFields(obj: Any) {
-        val fields = getAllFields(obj::class.java)
+        val fields = getAllFields(obj::class.java, emptyList())
 
         val fieldsWithObjects = fields.mapNotNull { field ->
             // Ignore classes which have not been loaded.
--- a/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
@@ -2,7 +2,6 @@ package net.corda.core.internal
 
 import net.corda.core.contracts.ContractState
 import net.corda.core.contracts.StatePointer
-import org.apache.commons.lang.ClassUtils
 import java.lang.reflect.Field
 import java.util.*
 
@@ -26,17 +25,20 @@ class StatePointerSearch(val state: ContractState) {
     // Queue of fields to search.
     private val fieldQueue = ArrayDeque<FieldWithObject>().apply { addAllFields(state) }
 
-    private fun getAllFields(clazz: Class<*>, fields: List<Field>): List<Field> {
-        return if (clazz.superclass != null) {
-            getAllFields(clazz.superclass, fields + clazz.declaredFields.asList())
-        } else {
-            fields
+    // Get fields of class and all super-classes.
+    private fun getAllFields(clazz: Class<*>): List<Field> {
+        val fields = mutableListOf<Field>()
+        var currentClazz = clazz
+        while (currentClazz.superclass != null) {
+            fields.addAll(currentClazz.declaredFields)
+            currentClazz = currentClazz.superclass
         }
+        return fields
     }
 
     // Helper for adding all fields to the queue.
     private fun ArrayDeque<FieldWithObject>.addAllFields(obj: Any) {
-        val fields = getAllFields(obj::class.java, emptyList())
+        val fields = getAllFields(obj::class.java)
 
         val fieldsWithObjects = fields.mapNotNull { field ->
             // Ignore classes which have not been loaded.
--- a/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
@@ -1,18 +1,17 @@
 package net.corda.core.internal
 
 import net.corda.core.contracts.ContractState
-import net.corda.core.contracts.LinearPointer
 import net.corda.core.contracts.StatePointer
-import net.corda.core.contracts.StaticPointer
 import java.lang.reflect.Field
 import java.util.*
 
 /**
  * Uses reflection to search for instances of [StatePointer] within a [ContractState].
+ * TODO: Doesn't handle calculated properties. Add support for this.
  */
 class StatePointerSearch(val state: ContractState) {
     // Classes in these packages should not be part of a search.
-    private val blackListedPackages = setOf("java.", "javax.")
+    private val blackListedPackages = setOf("java.", "javax.", "org.bouncycastle", "org.hibernate")
 
     // Type required for traversal.
     private data class FieldWithObject(val obj: Any, val field: Field)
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -1,18 +1,17 @@
...
 import java.util.*
 
 /**
  * Uses reflection to search for instances of [StatePointer] within a [ContractState].
+ * TODO: Doesn't handle calculated properties. Add support for this.
  */
 class StatePointerSearch(val state: ContractState) {
     // Classes in these packages should not be part of a search.
-    private val blackListedPackages = setOf("java.", "javax.")
+    private val blackListedPackages = setOf("java.", "javax.", "org.bouncycastle", "net.i2p.crypto")
```
shamsasari: The new packages need the dot at the end

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -36,39 +35,58 @@ class StatePointerSearch(val state: ContractState) {
...
+        return StatePointer::class.java.isAssignableFrom(obj.javaClass)
+    }
 
-                // Ignore nulls.
-                if (newObj in seenObjects) {
-                    return
-                }
+    private fun isMap(obj: Any): Boolean {
+        return java.util.Map::class.java.isAssignableFrom(obj.javaClass) ||
+                kotlin.collections.Map::class.java.isAssignableFrom(obj.javaClass)
```
shamsasari: I don't think you need this line. The kotlin collections map to Java's

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -36,39 +35,58 @@ class StatePointerSearch(val state: ContractState) {
...
+    }
+
+    private fun isIterable(obj: Any): Boolean {
+        return java.lang.Iterable::class.java.isAssignableFrom(obj.javaClass) ||
+                kotlin.collections.Iterable::class.java.isAssignableFrom(obj.javaClass)
```
shamsasari: likewise

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -36,39 +35,58 @@ class StatePointerSearch(val state: ContractState) {
...
+    private fun isStatePointer(obj: Any): Boolean {
+        return StatePointer::class.java.isAssignableFrom(obj.javaClass)
+    }
 
-                // Ignore nulls.
-                if (newObj in seenObjects) {
-                    return
-                }
+    private fun isMap(obj: Any): Boolean {
+        return java.util.Map::class.java.isAssignableFrom(obj.javaClass) ||
```
shamsasari: `obj is Map`?

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -36,39 +35,58 @@ class StatePointerSearch(val state: ContractState) {
...
+                kotlin.collections.Map::class.java.isAssignableFrom(obj.javaClass)
+    }
+
+    private fun isIterable(obj: Any): Boolean {
+        return java.lang.Iterable::class.java.isAssignableFrom(obj.javaClass) ||
```
shamsasari: likewise

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -36,39 +35,58 @@ class StatePointerSearch(val state: ContractState) {
...
-    private fun handleField(obj: Any, field: Field) {
-        when {
-            // StatePointer. Handles nullable StatePointers too.
-            field.type == LinearPointer::class.java -> statePointers.add(field.get(obj) as? LinearPointer<*> ?: return)
-            field.type == StaticPointer::class.java -> statePointers.add(field.get(obj) as? StaticPointer<*> ?: return)
-            // Not StatePointer.
-            else -> {
-                val newObj = field.get(obj) ?: return
+    private fun isStatePointer(obj: Any): Boolean {
+        return StatePointer::class.java.isAssignableFrom(obj.javaClass)
```
shamsasari: `obj is StatePointer`?

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -36,39 +35,58 @@ class StatePointerSearch(val state: ContractState) {
...
+        map.forEach { k, v ->
+            handleObject(k)
+            handleObject(v)
+        }
+    }
+
+    private fun handleObject(obj: Any?) {
+        if (obj == null) return
+        when {
+            isMap(obj) -> handleMap(obj as Map<*,*>)
```
shamsasari: You're losing out on kotlin's automatic type conversion here. Do this instead
`obj is Map<*, *> -> handleMap(obj)`
etc.

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -36,39 +35,58 @@ class StatePointerSearch(val state: ContractState) {
...
+                val packageName = obj.javaClass.`package`.name
+                val isBlackListed = blackListedPackages.any { packageName.startsWith(it) }
+                if (isBlackListed.not()) fieldQueue.addAllFields(obj)
             }
         }
     }
 
+    private fun handleField(obj: Any, field: Field) {
+        val newObj = field.get(obj) ?: return
+        if (newObj in seenObjects) return
```
shamsasari: I notice that `seenObjects` is a normal Map that's using object equality. These circular reference checks need to be based on object identity. Use this instead: `Collections.newSetFromMap(IdentityHashMap<Any, Boolean>())`

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: Done.

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: done

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: done

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: done - not sure why i didn't use is here... usually would do!

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: done - removed these methods.

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: done

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: Whoops - thanks

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
roger3cev: Done - thanks

roger3cev:
@shamsasari thanks for the comments - I think I have addressed them. Cheers!

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -21,54 +21,68 @@ class StatePointerSearch(val state: ContractState) {
...
+    private val seenObjects = Collections.newSetFromMap(IdentityHashMap<Any, Boolean>()).apply { add(state) }
 
     // Queue of fields to search.
     private val fieldQueue = ArrayDeque<FieldWithObject>().apply { addAllFields(state) }
 
+    private fun getAllFields(clazz: Class<*>, fields: List<Field> = emptyList()): List<Field> {
+        return if (clazz.superclass != null) {
+            getAllFields(clazz.superclass, fields + clazz.declaredFields)
+        } else {
+            fields
```
shamsasari: Shouldn't this be `clazz.declaredFields.asList()`?

Also, I'd advise you remove the default value on the `fields` parameter to avoid any confusion.

shamsasari: No, I meant this line for the null case. Following the code back to where it's called, this is just the empty list, no?

roger3cev: No worries - I changed it anyway. I was trying to be too clever :)

roger3cev:
@shamsasari oops I accidentally dismissed your review! Thanks for your time on this!

roger3cev:
@shamsasari oops I auto-dismissed your review again - apologies for the hassle. It was a pretty crucial change. When the search found an `Iterable` or `Map` it wasn't storing the encountered objects in the `seen` set, so if there was a cycle, the search would never halt.  The offending type was `Commodity` in `financeTypes.kt` where there is a circular reference. Seen objects should now be added to the seen set now matter what type the parent is now.

PR #4562, opened 2019-01-13
CORDA-2418: Fixed inability to use TestCordapp under gradle in Windows if referring to self cordapp

shamsasari:
The gradle process that runs the tests keeps an exclusive hold on built jars in Windows, which means if a test wants to get hold of the CorDapp, and it's requested from the same module, it will fail. Now instead of deleting the "libs" dir we just use the latest created jar.


shamsasari:
This issue will affect any app developer on Windows.

shamsasari:
The fix works on my home PC.

PR #4566, opened 2019-01-14
ENT-2927 Split Finance corDapp documentation updates

roastario:


gendal:
what is the dev experience?   existing c3 apps (compiled) will continue to work but if somebody tries to compile the c3 app again against c4 (asking for c3 behaviour) it will require gradle changes?

it's an interesting one.... i'm not sure that's actually included in our compatibility guarantee.  i mean: do we even support building a c3 app against c4?

eek... @mikehearn - your view?

josecoll:
I think we concluded that **re-compiling** a Corda 3 app against Corda 4 is considered a new version (as the JAR hashes will be different and thus incompatible with its previous version).

_why wouldn't we support building a c3 app against c4?_

gendal:
"why wouldn't we support building a c3 app against c4?"

and have it still work on c3? I agree it *should* work, but would it be a priority to fix it if it didn't, given there's an easy fix: compile against c3.  I suppose the argument would be: if a v3 app compiled against c4 doesn't work on c3, it's probably a sign that something _else_ isn't quite right somewhere?  and so it's a useful canary?

roastario:
I think the downsides of leaving this in (two versions of the Cash Contract) massively outweigh the downsides to requiring a two line change in gradle.

Effectively, at the moment you would

`compile net.corda:finance:$corda_version`

and after this, you must

`compile net.corda:finance-contracts:$corda_version
compile net.corda:finance-workflows:$corda_version
`

Everywhere in our documentation we tell users to split their cordapps into contracts and workflows - and then OUR flagship cordapp (in terms of usage) does not obey these rules?

josecoll:
Yes - I agree @roastario 
Just need to make sure out documentation is very clear about this.
Please can you update the docs as part of this PR 
eg. outstanding items from https://github.com/corda/enterprise/pull/1701

Please also:
- add a changelog entry
- update the upgrade notes (lines 146-149 in app-upgrade-notes.rst)
- perform a search for 'finance/contracts' and update doc refs.

josecoll:
>>"why wouldn't we support building a c3 app against c4?"

>>>>and have it still work on c3? I agree it *should* work, but would it be a priority to fix it if it didn't, >>>>given there's an easy fix: compile against c3.  I suppose the argument would be: if a v3 app >>>>compiled against c4 doesn't work on c3, it's probably a sign that something _else_ isn't quite >>>>right somewhere?  and so it's a useful canary?

Apologies - I mis-interpreted your question. 
I was referring to migrating a Corda 
probably didn't express myself succinctly enough.
I meant why would we support building a c3 app against c4

roastario:
on it.

PR #4567, opened 2019-01-14
CORDA-2434: Removed internal CordaRPCClient factory methods from the public API

shamsasari:


shamsasari:
This PR is completely wrong.

PR #4569, opened 2019-01-14
CORDA-2434: CordaRPCClient now only has c'tors rather a mix of c'tors and factory methods

shamsasari:


shamsasari:
It's more about being consistent. "If you don't need SSL then use the c'tor otherwise use these static methods." - it's not very obvious API.

PR #4571, opened 2019-01-14
Fixed AttachmentLoadingTests

shamsasari:
Port of #4565.

There's no new feature here, but we should make sure the release branch has this important test enabled.


shamsasari:
Closing this as the port needs to be in #4588 to fix the tests.

PR #4578, opened 2019-01-14
CORDA-2434: CordaRPCClient now only has c'tors rather a mix of c'tors…

shamsasari:
… and factory methods (#4569)

(cherry picked from commit 24a7821dceaa8cb02ff4d10a915880e435354f92)


shamsasari:
I'm not sure what `Build Pull Requests (Tests)` is, but all the other tasks are green.

PR #4582, opened 2019-01-15
[CORDA-2441] Fix Stdout progress renderer to swap out correct appender

JamesHR3:
The StdoutANSIProgressRenderer uses a bit of a hack to ensure the progress tracker information is printed at the bottom of the console - it looks for the console appender in the current log4j configuration and swaps this out for its own. To do this it looks for the name of this appender in the log4j config. This was changed in a previous fix ([CORDA-1897](https://r3-cev.atlassian.net/browse/CORDA-1897)), but the corresponding name the renderer was looking up was not changed.

This fix has been tested manually, by deploying the Bank of Corda demo and running a CashIssueFlow.

Changes:
 - Update the name of the appender the renderer looks for to swap out
 - Remove two superfluous return types the compiler was warning about

### DIFF
--- a/tools/shell/src/main/kotlin/net/corda/tools/shell/utlities/ANSIProgressRenderer.kt
+++ b/tools/shell/src/main/kotlin/net/corda/tools/shell/utlities/ANSIProgressRenderer.kt
@@ -250,7 +250,12 @@ object StdoutANSIProgressRenderer : ANSIProgressRenderer() {
             // than doing things the official way with a dedicated plugin, etc, as it avoids mucking around with all
             // the config XML and lifecycle goop.
             val manager = LogManager.getContext(false) as LoggerContext
-            val consoleAppender = manager.configuration.appenders.values.filterIsInstance<ConsoleAppender>().single { it.name == "Console-Selector" }
+            val consoleAppender: ConsoleAppender
+            try {
+                consoleAppender = manager.configuration.appenders.values.filterIsInstance<ConsoleAppender>().single { it.name == "Console-Selector" }
+            } catch (e: java.util.NoSuchElementException) {
+                throw AppenderNotFoundException("Failed to find Console-Selector appender - cannot display flow progress", e)
+            }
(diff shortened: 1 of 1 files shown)


tools/shell/src/main/kotlin/net/corda/tools/shell/utlities/ANSIProgressRenderer.kt
```diff
@@ -250,7 +250,7 @@ object StdoutANSIProgressRenderer : ANSIProgressRenderer() {
             // than doing things the official way with a dedicated plugin, etc, as it avoids mucking around with all
             // the config XML and lifecycle goop.
             val manager = LogManager.getContext(false) as LoggerContext
-            val consoleAppender = manager.configuration.appenders.values.filterIsInstance<ConsoleAppender>().single { it.name == "Console-Appender" }
+            val consoleAppender = manager.configuration.appenders.values.filterIsInstance<ConsoleAppender>().single { it.name == "Console-Selector" }
```
vkolomeyko: Should we be a bit more defensive here?

I.e. if appender is not found, produce a sensible warning, rather than having more technical exception in the log.

tools/shell/src/main/kotlin/net/corda/tools/shell/utlities/ANSIProgressRenderer.kt
```diff
@@ -250,7 +253,12 @@ object StdoutANSIProgressRenderer : ANSIProgressRenderer() {
             // than doing things the official way with a dedicated plugin, etc, as it avoids mucking around with all
             // the config XML and lifecycle goop.
             val manager = LogManager.getContext(false) as LoggerContext
-            val consoleAppender = manager.configuration.appenders.values.filterIsInstance<ConsoleAppender>().single { it.name == "Console-Appender" }
+            val consoleAppender: ConsoleAppender?
```
vkolomeyko: Really minor: Do we have to have declaration (with type) on a separate line?

PR #4598, opened 2019-01-17
[CORDA-2460] Add note to documentation warning about state relevancy update

JamesHR3:
[CORDA-2435](https://r3-cev.atlassian.net/projects/CORDA/issues/CORDA-2435) tracks an issue whereby the database upgrade process from V3 to V4 introduces a new column and a new table that are not appropriately backfilled by the migration process. This results in all states created in a V3 node in the vault being marked as relevant, which is incorrect. This may result in unexpected issues for an app that upgrades to V4, and starts running on an upgraded node, in the event that the app starts using V4 features.

The aim is to fix this issue in a patch release, but in the meantime a warning should be placed highlighting the problem to any party upgrading their app. This PR adds that warning.

Changes:
 * Add a note to the app upgrade notes indicating that unexpected results may occur if both an app and the node it runs on are upgraded to V4 and the app starts working with states created in V3 using V4 features.


JamesHR3:
I think this is superseded by the fix to CORDA-2435 in V4, so closing this PR.

PR #4606, opened 2019-01-18
CORDA-2464 - Enhance docker image, and improve usage documentation

roastario:
add example network
add ability to specify RPC password during config generation

### DIFF
--- a/docker/src/bash/example-mini-network.sh
+++ b/docker/src/bash/example-mini-network.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+NODE_LIST=("dockerNode1" "dockerNode2" "dockerNode3")
+NETWORK_NAME=mininet
+
+mkdir cordapps
+rm -f cordapps/*
+
+wget -O cordapps/finance-contracts.jar          https://ci-artifactory.corda.r3cev.com/artifactory/list/corda-dev/net/corda/corda-finance-contracts/4.0-SNAPSHOT/corda-finance-contracts-4.0-SNAPSHOT.jar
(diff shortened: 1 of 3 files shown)


docker/src/bash/example-mini-network.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+NODE_LIST=("dockerNode1" "dockerNode2" "dockerNode3")
+NETWORK_NAME=mininet
+CORDAPP_VERSION="4.0-SNAPSHOT"
+DOCKER_IMAGE_VERSION="corda-zulu-4.0-rc02"
+
+mkdir cordapps
+rm -f cordapps/*
+
+wget -O cordapps/finance-contracts.jar          https://ci-artifactory.corda.r3cev.com/artifactory/list/corda-dev/net/corda/corda-finance-contracts/${CORDAPP_VERSION}/corda-finance-contracts-4.0-SNAPSHOT.jar
```
tlil: There's still some `-4.0-SNAPSHOT` in here :)

tlil:
Unit and integration tests failed (failed to clean out working directory) .... Are you looking into it?

docker/src/bash/example-mini-network.sh
tlil: @roastario ^

arsalankhalid:
A suggestion would be to reduce the number of directories to get to this sample. Also should this be in a separate docker folder, why not in the samples directory?

arsalankhalid:
Hello there, chiming in just to help out. I tested your branch, and i got:
ERROR: Service 'Notary' failed to build: ADD failed: stat /var/lib/docker/tmp/docker-builder963460639/corda.jar: no such file or directory

Also, it seems this sample is in the "docker module"/folder, wouldn't this type of project go in the samples project instead?

roastario:
> Also, it seems this sample is in the "docker module"/folder, wouldn't this type of project go in the samples project instead?

Samples is for cordapps only. This bash file is solely there to provide a location to import into our docs build.

PR #4608, opened 2019-01-18
ENT-2967 Fix up use of various JDK performance contention points

rick-r3:
All code is in OS.  Each has a global mutex.

Stop repeated look up of KeyFactory.

Stop repeatedly asking for system class loader.

Replace ClassLoader.loadClass with Class.forName.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/crypto/SignatureScheme.kt
+++ b/core/src/main/kotlin/net/corda/core/crypto/SignatureScheme.kt
@@ -2,6 +2,7 @@ package net.corda.core.crypto
 
 import net.corda.core.KeepForDJVM
 import org.bouncycastle.asn1.x509.AlgorithmIdentifier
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/net/corda/core/crypto/SignatureScheme.kt
```diff
@@ -33,4 +34,15 @@ data class SignatureScheme(
         val algSpec: AlgorithmParameterSpec?,
         val keySize: Int?,
         val desc: String
-)
+) {
+    @Volatile
```
blsemo: If this field needs to be volatile, is it safe enough to set it without further synchronisation?

rick-r3: Yes.  Result of computing twice (or more) is stable.

PR #4611, opened 2019-01-21
[CORDA-2435] Print an error and exit if the vault contains V3 states

JamesHR3:
If a node is upgraded from V3 to V4, then any states currently in the vault are not migrated correctly. In particular, the relevancy column in the vault is filled in as relevant for all states (even if the node has recorded non-relevant states) and the state_party table is not populated with old states. The result is that certain V4 features will not work correctly with these states.

It turns out that fixing this for V4 is fairly involved. However, as the issue only affects nodes that are upgrading with V3 states in the vault, it is deemed acceptable to block these nodes from starting. This gives time to provide a proper fix for the problem in time for V4.1. This PR adds code to detect V3 states in the vault and block the node from starting in this case. See [CORDA-2435](https://r3-cev.atlassian.net/browse/CORDA-2435) for full details of the issue.

Changes:
 - Added function to NodeVaultService to check for V3 states in the vault
 - Added test to make sure this returns correctly
 - Check added at node startup using this that blocks the node starting if V3 states are detected.

Manual testing has been carried out to ensure the node does not start if V3 states are present, and does start if no V3 states are present.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -381,6 +381,10 @@ abstract class AbstractNode<S>(val configuration: NodeConfiguration,
             installCordaServices()
             contractUpgradeService.start()
             vaultService.start()
+            if (vaultService.oldStatesPresent()) {
+                stop()
+                throw OldStatesException()
+            }
             ScheduledActivityObserver.install(vaultService, schedulerService, flowLogicRefFactory)
 
             val frozenTokenizableServices = tokenizableServices!!
@@ -1073,6 +1077,9 @@ class FlowStarterImpl(private val smm: StateMachineManager, private val flowLogi
 }
 
 class ConfigurationException(message: String) : CordaException(message)
+
+class OldStatesException : Exception("Detected states created using Corda 3 in the vault. Certain Corda 4 features cannot work with Corda 3" +
+        " states. Exiting.")
 
 fun createCordaPersistence(databaseConfig: DatabaseConfig,
                            wellKnownPartyFromX500Name: (CordaX500Name) -> Party?,
--- a/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
@@ -1078,8 +1078,8 @@ class FlowStarterImpl(private val smm: StateMachineManager, private val flowLogi
 
 class ConfigurationException(message: String) : CordaException(message)
 
-class OldStatesException : Exception("Detected states created using Corda 3 in the vault. Certain Corda 4 features cannot work with Corda 3" +
-        " states. Exiting.")
+class OldStatesException : Exception("Detected states created using Corda 3 in the vault. Currently certain Corda 4 features cannot work with Corda 3" +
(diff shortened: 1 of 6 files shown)


node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt
```diff
@@ -1074,6 +1078,9 @@ class FlowStarterImpl(private val smm: StateMachineManager, private val flowLogi
 
 class ConfigurationException(message: String) : CordaException(message)
 
+class OldStatesException : Exception("Detected states created using Corda 3 in the vault. Certain Corda 4 features cannot work with Corda 3" +
```
rick-r3: Perhaps add some information to make clear this is temporary.  e.g. `Currently certain....` and perhaps ask them so see the release notes for a fuller explanation.

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -458,6 +458,68 @@ class NodeVaultService(
...
+            val query = session.createQuery(criteriaQuery)
+            val result = query.resultList.toSet()
+
+            log.debug("Found ${result.size} persistent party entries")
+            result
+        }
+    }
+
+    override fun oldStatesPresent(): Boolean {
+        log.info("Updating vault data from previous version to be compatible with the new version.")
```
rick-r3: I assume this message is now incorrect?

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -458,6 +458,68 @@ class NodeVaultService(
...
+        val session = getSession()
+        val persistentStates: List<VaultSchemaV1.VaultStates> = database.transaction {
+            getPersistentStates(
+                    session,
+                    QueryCriteria.VaultQueryCriteria(status = Vault.StateStatus.ALL),
+                    PageSpecification(),
+                    Sort(emptySet()),
+                    ContractState::class.java).filter { it[0] is VaultSchemaV1.VaultStates }.map { it[0] as VaultSchemaV1.VaultStates }
+        }
+        val stateParties = getPersistentParties(session)
```
rick-r3: I don't think we can or should be loading the states and comparing.  We should do all this in the database.  I think the condition holds at the transaction level.

select count(distinct transaction_id) from vault_states
select count(distinct transaction_id) from state_party

Something like that.  Happy to discuss.

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -458,6 +458,43 @@ class NodeVaultService(
...
+                .get<String>("txId")))
+        val query = session.createQuery(criteriaQuery)
+        val result = query.singleResult
+
+        log.debug("Found $result persistent party entries")
+        return result
+    }
+
+    override fun oldStatesPresent(): Boolean {
+        log.info("Checking for states in vault from a previous version")
```
rick-r3: I think we should log the result as well so we know when it ended.

JamesHR3:
Thanks Rick - also addressed.

JamesHR3:
Closing this after discussion, instead only committing this to release/4. Also adding a configuration option

PR #4620, opened 2019-01-22
add reference states to vault recording

roastario:


roger-that-dev (changes requested):
Need to do some further thinking on this. Cheers mate

PR #4631, opened 2019-01-24
[CORDA-2496] - Adjust CorDappResolver to handle same CorDapp registered multiple times

dimosr:
### Changes
* Allow `CordappResolver` to handle the case, where the same CorDapp is registered multiple times, by deduplicating. As an example, this happens, when running tests via `MockNetwork`, because a single resolver is shared between the several in-memory nodes, where the same CorDapp is regsitered for all of them.
* Update upgrade docs to be a bit more descriptive on the steps needed

### Testing
* Performed an upgrade of the [tests](https://github.com/corda/corda-solutions/blob/master/bn-apps/ledger-sync/ledger-sync-service/src/test/kotlin/com/r3/businessnetworks/ledgersync/LedgerConsistencyTests.kt) from a business networks app to ensure that the described steps are enough

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/cordapp/CordappResolver.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/cordapp/CordappResolver.kt
@@ -17,8 +17,12 @@ object CordappResolver {
     private var cordappResolver: () -> Cordapp? = {
         Exception().stackTrace
                 .mapNotNull { cordappClasses[it.className] }
-                // If there is more than one cordapp registered for a class name we can't determine the "correct" one and return null.
-                .firstOrNull { it.size == 1 }?.single()
+                // for each class, we filter out CorDapps that originate from the same file.
+                .map { it.distinctBy { it.jarHash } }
+                // in case there are multiple classes matched, we select the first one having a single CorDapp registered against it.
+                .firstOrNull { it.size == 1 }
+                // otherwise we return null, signalling we cannot reliably determine the current CorDapp.
+                ?.single()
     }
 
     /*
--- a/core/src/main/kotlin/net/corda/core/internal/cordapp/CordappResolver.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/cordapp/CordappResolver.kt
@@ -17,8 +17,6 @@ object CordappResolver {
     private var cordappResolver: () -> Cordapp? = {
         Exception().stackTrace
                 .mapNotNull { cordappClasses[it.className] }
-                // for each class, we filter out CorDapps that originate from the same file.
-                .map { it.distinctBy { it.jarHash } }
                 // in case there are multiple classes matched, we select the first one having a single CorDapp registered against it.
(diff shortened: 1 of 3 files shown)


docs/source/app-upgrade-notes.rst
```diff
@@ -261,7 +261,25 @@ becomes::
     )
 
 You may need to use the new ``TestCordapp`` API when testing with the node driver or mock network, especially if you decide to stick with the
-pre-Corda 4 ``FinalityFlow`` API. The previous way of pulling in CorDapps into your tests does not honour CorDapp versioning.
+pre-Corda 4 ``FinalityFlow`` API. The previous way of pulling in CorDapps into your tests (i.e. via using the ``cordappPackages`` parameter) does not honour CorDapp versioning.
```
ivanr3: Should we also mention that classes are picked up only from the main module? I.e. if you have flows defined in test module they wouldn't be picked up even though they are technically in the same package?

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappResolver.kt
```diff
@@ -17,8 +17,12 @@ object CordappResolver {
     private var cordappResolver: () -> Cordapp? = {
         Exception().stackTrace
                 .mapNotNull { cordappClasses[it.className] }
-                // If there is more than one cordapp registered for a class name we can't determine the "correct" one and return null.
-                .firstOrNull { it.size == 1 }?.single()
+                // for each class, we filter out CorDapps that originate from the same file.
+                .map { it.distinctBy { it.jarHash } }
```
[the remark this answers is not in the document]
dimosr: Done, I decided to de-dup based on the hash, instead of overriding equals/hashCode to reduce the scope of the change, given this is only needed for the testing infrastructure.

docs/source/app-upgrade-notes.rst
dimosr: thanks for reminding this, I added a note about it.

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappResolver.kt
```diff
@@ -29,8 +31,11 @@ object CordappResolver {
     fun register(cordapp: Cordapp) {
         cordapp.cordappClasses.forEach {
             if (cordappClasses.containsKey(it)) {
-                logger.warn("More than one CorDapp registered for $it.")
-                cordappClasses[it] = cordappClasses[it]!! + cordapp
+                // we do not register CorDapps that originate from the same file.
+                if (cordappClasses[it]!!.none { it.jarHash.equals(cordapp.jarHash) }) {
```
shamsasari: Can you replace the `containsKey` with the index operation that you're doing here.

shamsasari: `==` instead of equals

PR #4634, opened 2019-01-24
CORDA-2345: Fix Windows build

vkolomeyko:
Small change to fix: https://ci-master.corda.r3cev.com/viewLog.html?buildId=140738&buildTypeId=Corda_BuildWindowsTest


shamsasari:
I'd like to understand a bit more why this issues started happening. https://github.com/corda/corda/pull/4600 didn't touch `TestCordappImpl` nor `IdentitySyncFlowTests`, so I'm a bit confused.

Can you print out what URL is that we're trying to convert that Windows doesn't like.

vkolomeyko:
Sure.

I think it was #4434 that broke the Windows build.

URL on Windows looks as follows:
`file:Z%3a/corda/finance/contracts/build/libs/corda-finance-contracts-5.0-SNAPSHOT.jar`

PR #4638, opened 2019-01-25
CORDA-2514 Constraints checking failures do not provide indication of cause

josecoll:
Add logging to constraints checking to indicate reason of failure.


josecoll:
Yes -agreed. Clearer exception handling may be better going forward.

PR #4650, opened 2019-01-28
Fix link to errors.corda.net in release notes

tlil:


tlil:
As per discussion on Slack 26 Jan.

PR #4653, opened 2019-01-28
CORDA-2422: Ignore future interfaces from serialization

dimosr:
### Changes
The interface `FungibleState` - introduced in v4 - prevents interop between v3 and v4. Briefly, the reason is this interface is missing, which means carpenter is triggered, but it's not able to synthesize it. This fix ignores this specific interface, avoiding to create a serializer. This is not expected to cause any issue, since this interface will not be needed during serialization.

### Testing
* Executed the BoC demo, as specified in the ticket, making sure that it was not failing after incorporating the fix


r3domfox (approved):
This is OK as a tactical fix for now. We should look into optionally suppressing interfaces from the serialisation schema for outgoing messages from the V4 node, which could potentially be done via the `isExcluded` method on `LocalTypeModelConfiguration` which currently checks classes against the whitelist.

dimosr:
Since there will be work for 4.1 for a systemic solution, which might make this unnecessary, closing this for now.

PR #4663, opened 2019-01-28
dockerize docs building

roastario:
At the moment, building docs is quite hit and miss. We currently have developers on OSX, Windows, and various distros of linux. 

This new container will isolate the individual developer system from the required setup for building docs.

### DIFF
--- a/docs/build.gradle
+++ b/docs/build.gradle
@@ -1,3 +1,5 @@
+import java.nio.file.Files
+
 apply plugin: 'org.jetbrains.dokka'
 apply plugin: 'kotlin'
 
@@ -62,8 +64,14 @@ task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
 task buildDocs(dependsOn: ['apidocs', 'makeDocs'])
 task apidocs(dependsOn: ['dokka', 'dokkaJavadoc'])
 
-task makeDocs(type: Exec) {
-    commandLine "docker", "run", "-v", "${project.projectDir}:/opt/docs_builder", "roastario/docs-builder:latest", "bash",  "-c", "rm -rf build && rm -rf virtualenv && install-docsite-requirements.sh && make-docsite.sh"
+task makeDocs(type: Exec, dependsOn: ['ensureBuildFolderPresent']) {
+    commandLine "docker", "run", "-v", "${project.projectDir}:/opt/docs_builder", "corda/docs-builder:latest", "bash",  "-c", "make-docsite.sh"
+}
+
+task ensureBuildFolderPresent{
+    doLast{
+        Files.createDirectories(new File(project.buildDir, "html").toPath())
+    }
 }
 
 apidocs.shouldRunAfter makeDocs
--- a/docs/build.gradle
+++ b/docs/build.gradle
@@ -62,16 +62,8 @@ task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
 task buildDocs(dependsOn: ['apidocs', 'makeDocs'])
 task apidocs(dependsOn: ['dokka', 'dokkaJavadoc'])
 
-task makeDocs(type: Exec, dependsOn: ['installDocsiteRequirements']) {
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash make-docsite.sh' // Windows
-    commandLine 'bash', './make-docsite.sh' // Linux
-}
-
-task installDocsiteRequirements(type: Exec) {
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash install-docsite-requirements.sh' // Windows
-    commandLine 'bash', './install-docsite-requirements.sh' // Linux
+task makeDocs(type: Exec) {
+    commandLine "docker", "run", "-v", "${project.projectDir}:/opt/docs_builder", "roastario/docs-builder:latest", "bash",  "-c", "rm -rf build && rm -rf virtualenv && install-docsite-requirements.sh && make-docsite.sh"
 }
 
 apidocs.shouldRunAfter makeDocs
--- a/docs/build.gradle
+++ b/docs/build.gradle
@@ -1,3 +1,5 @@
+import org.apache.tools.ant.taskdefs.condition.Os
+
 import java.nio.file.Files
 
 apply plugin: 'org.jetbrains.dokka'
@@ -23,7 +25,7 @@ def internalPackagePrefixes(sourceDirs) {
 ext {
     // TODO: Add '../client/jfx/src/main/kotlin' and '../client/mock/src/main/kotlin' if we decide to make them into public API
     dokkaSourceDirs = files('../core/src/main/kotlin', '../client/rpc/src/main/kotlin', '../finance/workflows/src/main/kotlin', '../finance/contracts/src/main/kotlin', '../client/jackson/src/main/kotlin',
-                            '../testing/test-utils/src/main/kotlin', '../testing/node-driver/src/main/kotlin')
+            '../testing/test-utils/src/main/kotlin', '../testing/node-driver/src/main/kotlin')
     internalPackagePrefixes = internalPackagePrefixes(dokkaSourceDirs)
 }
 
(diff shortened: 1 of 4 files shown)


tlil:
TC less happy tho... looking into it?

docs/build.gradle
```diff
@@ -62,16 +62,8 @@ task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
...
-    commandLine 'cmd', '/c', 'bash make-docsite.sh' // Windows
-    commandLine 'bash', './make-docsite.sh' // Linux
-}
-
-task installDocsiteRequirements(type: Exec) {
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash install-docsite-requirements.sh' // Windows
-    commandLine 'bash', './install-docsite-requirements.sh' // Linux
+task makeDocs(type: Exec) {
+    commandLine "docker", "run", "-v", "${project.projectDir}:/opt/docs_builder", "roastario/docs-builder:latest", "bash",  "-c", "rm -rf build && rm -rf virtualenv && install-docsite-requirements.sh && make-docsite.sh"
```
tlil: `roastario` => `corda` ?

roastario: I will move it to corda once it's working - don't want to pollute our repo with loads of broken images :stuck_out_tongue:

roastario:
> TC less happy tho... looking into it?

Yep, mystery error with: 

    `org.gradle.api.UncheckedIOException: Failed to create parent directory '/data/BuildAgent/work/996a671061afbd68/docs/build/html/api' when creating directory '/data/BuildAgent/work/996a671061afbd68/docs/build/html/api/kotlin'`

wzur-r3:
This replicates (badly) what has been already done for `corda-docs-builder` - see `corda-builder.sh` in that repo.

Also it pollutes working directories on TC builders, as the documentation are executed as `root`. This causes subsequent TC build to fail for other branches, when TC tries to clean working directory and it cannot remove files owned by `root`.

docs/build.gradle
```diff
@@ -62,16 +64,14 @@ task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
 task buildDocs(dependsOn: ['apidocs', 'makeDocs'])
 task apidocs(dependsOn: ['dokka', 'dokkaJavadoc'])
 
-task makeDocs(type: Exec, dependsOn: ['installDocsiteRequirements']) {
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash make-docsite.sh' // Windows
-    commandLine 'bash', './make-docsite.sh' // Linux
+task makeDocs(type: Exec, dependsOn: ['ensureBuildFolderPresent']) {
```
wawrzek: We should not force people to use docker for docs. It should be an option, even if that default one, there should be a way to not use docker.

docs/docs_builder/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:2-stretch
+
+RUN apt-get update && apt-get install -y texlive preview-latex-style texlive-generic-extra texlive-latex-extra latexmk dos2unix
```
wawrzek: Maybe something like that (however, that's for Ubuntu)
RUN export DEBIAN_FRONTEND=noninteractive && apt-get -y update \
  && LC_ALL=C apt-get -y install \
     git \
     latexmk \
     make \
     python-minimal \
     python-git \
     python-jinja2 \
     python-virtualenv \
     python-yaml \
     python-pip \
     sudo \
     texlive-latex-extra \
     texlive-latex-recommended \
     texlive-fonts-recommended \
     unzip \
     wget \
  && apt-get -y clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

roastario: won't do.

docs/build.gradle
roastario: no, this is to make it the same across all systems and remove all variability around docs building.

roastario:
> This replicates (badly) what has been already done for `corda-docs-builder` - see `corda-builder.sh` in that repo.
> 
> Also it pollutes working directories on TC builders, as the documentation are executed as `root`. This causes subsequent TC build to fail for other branches, when TC tries to clean working directory and it cannot remove files owned by `root`.

incorrect - the linked repo does not replicate this behaviour - this allows developers to run it as part of a normal gradle build.

docs/docs_builder/Dockerfile
wzur-r3: at least this line should delete all deb downloaded, ie:
> && apt-get -y clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

that will make a docker images smaller

docs/build.gradle
```diff
@@ -62,16 +66,12 @@ task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
...
-    commandLine 'bash', './make-docsite.sh' // Linux
-}
-
-task installDocsiteRequirements(type: Exec) {
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash install-docsite-requirements.sh' // Windows
-    commandLine 'bash', './install-docsite-requirements.sh' // Linux
+task makeDocs(type: Exec) {
+    if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        commandLine "docker", "run", "-v", "${project.projectDir}:/opt/docs_builder", "corda/docs-builder:latest", "bash", "-c", "make-docsite.sh"
```
tlil: `--rm` for the Windows run as well?

roastario: oops, adding now.

docs/build.gradle
```diff
@@ -62,16 +66,12 @@ task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
...
-
-task installDocsiteRequirements(type: Exec) {
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash install-docsite-requirements.sh' // Windows
-    commandLine 'bash', './install-docsite-requirements.sh' // Linux
+task makeDocs(type: Exec) {
+    if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        commandLine "docker", "run", "--rm", "-v", "${project.projectDir}:/opt/docs_builder", "corda/docs-builder:latest", "bash", "-c", "make-docsite.sh"
+    } else {
+        commandLine "bash", "-c",  "docker run --rm --user \$(id -u):\$(id -g) -v ${project.projectDir}:/opt/docs_builder corda/docs-builder:latest bash -c make-docsite.sh"
```
wzur-r3: There is no need to start `docker` through `bash` at all here.

docs/build.gradle
```diff
@@ -62,16 +66,12 @@ task dokkaJavadoc(type: org.jetbrains.dokka.gradle.DokkaTask) {
...
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash make-docsite.sh' // Windows
-    commandLine 'bash', './make-docsite.sh' // Linux
-}
-
-task installDocsiteRequirements(type: Exec) {
-    // TODO: Non-msys Windows script
-    commandLine 'cmd', '/c', 'bash install-docsite-requirements.sh' // Windows
-    commandLine 'bash', './install-docsite-requirements.sh' // Linux
+task makeDocs(type: Exec) {
```
wzur-r3: I really would like to see this task to optionally *not* to start the docs builder inside a docker container.
That is for a case when the build environment that docker container provides has been already provided.
Most specifically, by another docker container for the whole build.

docs/build.gradle
wzur-r3: also when you drop the first `bash -c`, the `commandLine` can be an array, just like it is for `FAMILY_WINDOWS` and it makes it easier to manage common docker arguments in one place.

wzur-r3: ah - one more thing - prefixing a command to be executed inside a docker container with `bash -c` is obsolete. It would be the best to drop it completely.

docs/build.gradle
roastario: This is not a good suggestion


We are standardising the way to build docs. Introducing an "alternative" will not be standard.

docs/build.gradle
roastario: You are incorrect. 

I must do bash -c because I am performing `id -u` and `id -g` to figure out the correct user to run as. This is in turn launched as a subprocess of a subprocess (gradle -> shell -> id). To make it work, reliably across all shells, this is the way to do it. 

Simply checking out this PR and running it would have shown why this was done this way.

docs/build.gradle
wzur-r3: That is correct, and I wrote already I like that approach. However it won't play very well if a build is running *already* inside a docker container.

docs/build.gradle
wzur-r3: Personally I would take these two `id` calls out of this particular line to a separate calls, but that's just me.

Still `bash -c make-docsite.sh` doesn't make any sense to me.

docs/build.gradle
wawrzek: What if I would build the whole Corda in Docker. With your setup it means that I would have a gradle running inside of docker creating another docker. Is that correct?

tlil: `$ ./make-docsite.sh` :-)

PR #4664, opened 2019-01-28
ENT-3000 Start pooling classpath scanning and related fixes

rick-r3:
Constrain threads used in ClassGraph calls.

Quickly patch synchronisation of attachment class loader cache.  Needs a revisit but more complicated due to DJVM.


rick-r3:
Turned out not making the threads daemons causes some tests to hang.

PR #4674, opened 2019-01-29
CORDA-2520: Add FetchParametersFlow

kasiastreich:
Add FetchParametersFlow. It requires minimumPlatformVersion to be >=4 because nodes on lower versions won't understand this flow (also, in this case transactions are not tagged). It doesn't solve the CORDA-2520 issue entirely for mixed bootstrapped networks when node 3 passes transaction that has v4 transaction in a chain, that node (also on v4) doesn't know about. But, network bootstrapper sets minimumPlatform version to 4 so it is unlikely.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
@@ -77,6 +77,7 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
     @Suspendable
     @Throws(FetchDataFlow.HashNotFound::class, FetchDataFlow.IllegalTransactionRequest::class)
     override fun call() {
+        val minimumPlatformVersion = serviceHub.networkParameters.minimumPlatformVersion
         val newTxns = ArrayList<SignedTransaction>(txHashes.size)
         // Start fetching data.
         for (pageNumber in 0..(txHashes.size - 1) / RESOLUTION_PAGE_SIZE) {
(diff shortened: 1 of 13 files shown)


core/src/main/kotlin/net/corda/core/internal/ResolveTransactionsFlow.kt
```diff
@@ -85,6 +86,11 @@ class ResolveTransactionsFlow(txHashesArg: Set<SecureHash>,
             newTxns += downloadDependencies(page)
             val txsWithMissingAttachments = if (pageNumber == 0) signedTransaction?.let { newTxns + it } ?: newTxns else newTxns
             fetchMissingAttachments(txsWithMissingAttachments)
+            // Fetch missing parameters flow was added in version 4. This check is needed so we don't end up with node V4 sending parameters
```
mikehearn: Hmmm. I wonder if we should start fetching and storing _before_ the flag day, and just have validation be the thing that depends on the min platform version? Seems like the fewer things change on flag day, the safer we are? In that case this would be gated on the platform version of the peer rather than the global version. This code isn't wrong, but, if we flip the switch and everything explodes....

node/src/main/kotlin/net/corda/node/internal/DBNetworkParametersStorage.kt
```diff
@@ -104,7 +90,7 @@ class DBNetworkParametersStorage(
         }
     }
 
-    // TODO For the future we could get them also as signed (by network operator) attachments on transactions.
+    // TODO Revisit this approach.
```
mikehearn: Hmmm. I don't fully understand why this TODO isn't simply now "done"?

kasiastreich: True

shamsasari:
At a quick glance this looks OK, but I'll leave it to Mike or Jose to approve.

PR #4676, opened 2019-01-29
ENT-2991 Deadlock between AppendOnlyPersistentMap and underlying database

blsemo:
During performance testing, a deadlock was discovered that could happen if two threads try to write the same key to the database, and also use the same instance of an AppendOnlyPersistentMap to read a key from a different table.
This changes removes the lock around database read access in AppendOnlyPersistentMap (added by @rick-r3 ) and adds a test that checks for the error condition that would have led to a deadlock before applying this change (added by @blsemo )

### DIFF
--- a/node/src/test/kotlin/net/corda/node/services/persistence/DbMapDeadlockTest.kt
+++ b/node/src/test/kotlin/net/corda/node/services/persistence/DbMapDeadlockTest.kt
@@ -16,11 +16,13 @@ import java.lang.Thread.sleep
 import java.util.*
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.CyclicBarrier
+import java.util.concurrent.atomic.AtomicReference
 import javax.persistence.Column
 import javax.persistence.Entity
 import javax.persistence.Id
 import kotlin.concurrent.thread
 import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
 
 class TestKey(val value: Int) {
     override fun equals(other: Any?): Boolean {
@@ -44,7 +46,6 @@ class MyPersistenceClass(
 
         @Column(name = "lValue", nullable = false)
         val value: Int)
-
 
 @Entity
 @javax.persistence.Table(name = "otherlockobjects")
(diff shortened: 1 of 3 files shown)


node/src/test/kotlin/net/corda/node/services/persistence/DbMapDeadlockTest.kt
```diff
@@ -0,0 +1,230 @@
...
+            // thread, it will wait for the main thread's db transaction to commit or rollback before proceeding with the read.
+            val otherThread = thread(name = "testThread2") {
+                try {
+                    log.info("Thread2 waiting")
+                    latch1.await()
+                    latch2.await()
+                    log.info("Thread2 starting transaction")
+                    persistence.transaction {
+                        log.info("Thread2 getting key 2")
+                        testMap.get(TestKey(2))
```
szymonsztuka: this is an operator, could be `testMap[TestKey(2)]`, but it's really a personal style, with get is more widely understood

node/src/test/kotlin/net/corda/node/services/persistence/DbMapDeadlockTest.kt
```diff
@@ -0,0 +1,230 @@
...
+                        log.info("Thread2 set other value 100")
+                        otherMap.set(TestKey(100), 100)
```
szymonsztuka: as above, could be `otherMap[TestKey(100)] = 100`

node/src/test/resources/migration/locktestschema.xml
```diff
@@ -0,0 +1,31 @@
...
+        <createTable tableName="locktestobjects">
+            <column name="lKey" type="INT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="lValue" type="INT">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+
+    </changeSet>
```
szymonsztuka: All `statements` could in one changeSet.

node/src/test/kotlin/net/corda/node/services/persistence/DbMapDeadlockTest.kt
```diff
@@ -0,0 +1,230 @@
...
+                        latch3.await()
+                        log.info("Thread2 getting value 10")
+                        val v = testMap.get(TestKey(10))
+                        assertEquals(10, v)
+                    }
+                    log.info("Thread2 done")
```
szymonsztuka: When exception is thrown this  line doesn't invoke, would it be any benefit if we expect/check or 1) reaching this line or 2) the error being thrown.  By e.g. setting some global `var` and requiring  it to be set or the error be present at the end, we would ensure we reach ``assertEquals(10, v)`` in the first case).

node/src/test/kotlin/net/corda/node/services/persistence/DbMapDeadlockTest.kt
```diff
@@ -0,0 +1,230 @@
...
+            otherThread.join()
+            checkException(otherThreadException)
+            log.info("MainThread done")
+        }
+    }
+
+    // We have to catch any exception thrown and check what they are - primary key constraint violations are fine, we are trying
+    // to insert the same key twice after all. Any deadlock time outs or similar are completely not fine and should be a test failure.
+    private fun checkException(exception: Exception?) {
+        if (exception == null){
```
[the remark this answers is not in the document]
rick-r3: I've hit reformat on the file.

node/src/test/kotlin/net/corda/node/services/persistence/DbMapDeadlockTest.kt
rick-r3: I've tidied up a bit to assert that the exception is set at the end of the MainThread.  I've also added logging when it throws.  I also decided that the visibility of changes to `otherThreadException` were not necessarily visible as not `@Volatile`, so switched to `AtomicReference`.

PR #4680, opened 2019-01-29
CORDA-2515 Fix bug in verifier for has to signature constraints (auto-migration).

josecoll:
Verification logic needs to use the correct Attachment in a hash to signature constraint auto-switch:
- signed for Signature constrained state (OUTPUT)
- unsigned for Hash constrained state (INPUT)

Note: the previous logic was working correctly only under certain conditions.
This fix has been validated against the QA TST: https://r3-cev.atlassian.net/browse/R3T-1560


josecoll:
Its actually stricter now as it *only* processes states which are Hash or Signature (and are auto-migratable as per the hashToSignatureConstrainedContracts list). Any other state is treated as per original processing rules.

PR #4698, opened 2019-01-31
CORDA-2543 object graph traversal

r3domfox:
Per [CORDA-2543](https://r3-cev.atlassian.net/browse/CORDA-2543), introduce an `ObjectGraphTraverser` which uses the type analysis from the serialization framework to traverse an object graph in depth-first order, emitting a `Sequence` of objects in the graph:

```kotlin
val peopleInObjectGraph = ObjectGraphTraverser.traverse(object) { it is Person }.toList()
```

To support code which must run in `core`, the `LocalTypeModel` code and related classes have been separated from the `serialization` context and now stand alone, in `net.corda.core.internal.reflection`.

The remaining classes in `serialization.amqp` have been re-organised with some additional packages being introduced - the aim was to clarify the dependencies between groups of classes and eliminate cyclic dependencies between packages, but some cycles remain which it would be troublesome to remove.

By re-implementing the very small part of `TypeToken` that is actually used in `serialization`, we remove the dependency on `TypeToken` from both `core` and `serialization`.


core/src/main/kotlin/net/corda/core/internal/reflection/PropertyReader.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import net.corda.core.KeepForDJVM
+import java.lang.reflect.Field
+import java.lang.reflect.Method
+
+/**
+ * Obtains the value of a property from an instance of the type to which that property belongs, either by calling a getter method
+ * or by reading the value of a private backing field.
+ */
+@KeepForDJVM
```
shamsasari: I thought reflection is disabled in the DJVM. @chrisr3 ?

shamsasari (changes requested):
Can you wire this up to `StatePointerSearch`. There's also `CordaSerializableBeanSerializerModifier` that can benefit from this I think.

Further, this is a big PR and I see that some classes have been moved into their own files. Can you keep such changes to a minimum please to reduce the amount of changes and to keep the git history. You can always do the refactoring in a separate PR.

core/src/main/kotlin/net/corda/core/internal/reflection/PropertyReader.kt
chrisr3: Corda's `djvm` module will rewrite the bytecode to replace all `java.lang.reflect.*` and `java.lang.invoke.*` methods with stubs that throw `RuleViolationError` inside Avian. So this code won't work inside the sandbox.

The `jdk8u-deterministic/jdk/jre/lib/rt.jr` is still including them at the moment, though - for reasons I cannot remember right now.

r3domfox: Object graph traversal, used in `StatePointerSearch`, will rely on reflection to traverse the object graph, whether it uses this code or not.

r3domfox:
Wiring up into `StatePointerSearch` is now done, with a couple of additional tests indicating the behaviour around calculated properties. @roger3cev, could you maybe kick the tires on `StatePointerSearch` and its associated tests and confirm whether we have the right behaviour for all the cases we care about?

roger3cev:
@r3domfox i'll have a look today - cheers!!

core/src/test/kotlin/net/corda/core/internal/reflection/PropertyEnumeratorTests.kt
```diff
@@ -0,0 +1,44 @@
...
+                else -> 0
+            }
+
+            override fun toString(): String = when(value) {
+                is String -> "Bottom($value)"
+                else -> "cycle"
+            }
+        }
+
+        data class Middle(val value: String, val bottoms: List<Bottom>)
```
shamsasari: Can you add separate simple test cases for List and Map

shamsasari (commented):
You have a merge conflict. Otherwise this is looking ready to merge.

core/src/test/kotlin/net/corda/core/internal/reflection/PropertyEnumeratorTests.kt
r3domfox: Done.

PR #4700, opened 2019-01-31
CORDA-2467: Moved non-contract things out of finance-contracts and into finance-workflows

shamsasari:
This includes all of the cash selection logic, JSON support using jackson and a bunch of utilities which are not relevant to contract verification. The exception to this are Interpolator which has been moved to the IRS demo, and PhysicalLocationStructures.kt which is now in explorer.


vkolomeyko:
@shamsasari 
What is the size of `finance-contracts.jar` after the change?

Unit tests failing with:
```
CordaRPCOpsImplTest.attachment uploaded with metadata has specified filename
java.lang.IllegalArgumentException: There is more than one CorDapp containing the package net.corda.finance.contracts on the classpath [/data/BuildAgent/work/996a671061afbd68/finance/contracts/build/libs/corda-finance-contracts-5.0-SNAPSHOT.jar, /data/BuildAgent/work/996a671061afbd68/node/build/libs/corda-node-5.0-SNAPSHOT.jar]. Specify a package name which is unique to the CorDapp.
```

shamsasari:
Tests should now be fixed.

The jar is now 175KB.

vkolomeyko:
Thanks @shamsasari 

How are we looking on compatibility side with this change?
I.e. if Cash state was saved before the change Corda 4, can it be loaded and transacted upon with current version in `master`?

Is Demobench still working after this change?

shamsasari:
The contract hasn't changed so I see no reason why compatibility should be affected. However, we have to be careful about placing unnecessary guarantees here. The finance app is now demo and I don't think we are providing any guarantees on compatibility. Certainly the flows from V3 don't work with V4.

Demobench works.

PR #4704, opened 2019-02-01
Remove old reference to publishing a single corda finance jar.

josecoll:
Finance is no longer published as a single JAR.


szymonsztuka:
Yes indeed, this is already correct in the current code.

PR #4707, opened 2019-02-01
CORDA-2423 document draining requirements in flow state machines

r3domfox:
Document the draining requirements for node upgrade in the flow state machines docs, explaining that while evolution of transaction data is supported, there is no equivalent mechanism for evolving persisted flow checkpoints.

### DIFF
--- a/docs/source/flow-state-machines.rst
+++ b/docs/source/flow-state-machines.rst
@@ -385,6 +385,15 @@ suspended into a continuation and saved to persistent storage. If the node crash
 effectively continue as if nothing had happened. Your code may remain blocked inside such a call for seconds,
 minutes, hours or even days in the case of a flow that needs human interaction!
 
+Note that if a node has flows still in a suspended state, with flow continuations written to disk, it will not be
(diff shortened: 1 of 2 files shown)


docs/source/flow-state-machines.rst
```diff
@@ -385,6 +385,15 @@ suspended into a continuation and saved to persistent storage. If the node crash
 effectively continue as if nothing had happened. Your code may remain blocked inside such a call for seconds,
 minutes, hours or even days in the case of a flow that needs human interaction!
 
+Note that if a node has flows still in a suspended state, with flow continuations written to disk, it will not be
+possible to upgrade that node to a new version of Corda, because flows must be completely "drained" before an upgrade
```
mikehearn: to a new version of Corda *or your app*, because flows must be ....

Maybe put this in a `.. warning::` block and put it after the note below, which introduces the concept of checkpointing.

r3domfox: Done.

PR #4709, opened 2019-02-02
[CORDA-2545] Verifier changes: plumb network parameters into the attachments classloader

mikehearn:
This is the first of a series of changes intended to fix package namespace ownership and simplify constraint migration logic, which has proven to be problematically complex during the C4 testing and a re-review. We'll back off and go for a simpler approach that involves making hash/zone constraints breakable via sig migrations, possibly reintroducing "unbreakable" hash constraints in a future evolution of the data model. It's unclear to what extent people really want such a feature anyway.

💣 This PR stream is all on top of the branch `mike-simplify-migration` - the goal is to merge these changes quickly onto the branch, such that each PR can itself be incomplete or partial, and then merge the branch to master at the end.

This change does several things:

1. Provides a `NetworkParameters` to the `AttachmentsClassLoader` which previously did not have it.
2. Changes the backwards compatibility fallback to fetch the current params out of the FlowLogic if a LedgerTransaction was created in flow context using old/deprecated constructors. The current code is trying to keep the verifier paths able to work with no parameters access, but that wasn't really being followed - the package namespace ownership check code has been null asserting it away for ages already and nothing broke, so it makes sense to just try harder to get params but then give up early if we can't get params from anywhere and accept the (likely very minor or non-existant) compatibility issue.
3. Misc renames/doc comments/additions.
4. Deletes a parameter from a new public API we seem to have added since v3, but the param is useless for users and existed only due to an implementation choice. This is triggering the API checker but is deliberate, as I don't think we shipped this entry point before.

Unfortunately Tudor is now on holiday so this PR sequence will have to be done without him.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
@@ -29,6 +29,16 @@ class TransactionResolutionException(val hash: SecureHash) : FlowException("Tran
 class AttachmentResolutionException(val hash: SecureHash) : FlowException("Attachment resolution failure for $hash")
 
 /**
+ * Thrown to indicate that a contract attachment is not signed by the network-wide package owner.
+ */
+@CordaSerializable
+class PackageOwnershipException(val attachmentHash: AttachmentId, val contractClass: String, val packageName: String) : FlowException(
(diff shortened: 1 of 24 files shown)


mikehearn:
Thanks Chris!

mikehearn:
This PR has been rolled into #4713 so I'll close it and continue there.

PR #4710, opened 2019-02-04
Adding new page on UAT environment

carolynequinn:
We need to have an external page which talks about UAT. As agreed with R3's marketing team and the Network team, this should be on docs.corda.net rather than elsewhere. 

(The idea is that this cannot be corda.network since UAT is part of R3, not the Foundation.)

This page + relevant sub-pages will continue to be updated, with more detail.

### DIFF
--- a/docs/source/corda-network/UAT.md
+++ b/docs/source/corda-network/UAT.md
@@ -1,7 +1,7 @@
 Corda Network: UAT Environment
 =============================
 
-For owners of tested CorDapps with a firm plan to take them into production, a bespoke UAT environment can be provided by R3 (generally, a commercal agreement will need to be in place). Here, such CorDapps can be further tested in the network configuration they will experience in production, utilising relevant Corda Network Services (including the Identity Operator, and trusted notaries). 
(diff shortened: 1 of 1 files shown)


docs/source/corda-network/UAT.md
```diff
@@ -0,0 +1,10 @@
+Corda Network: UAT Environment
+=============================
+
+For owners of tested CorDapps with a firm plan to take them into production, a bespoke UAT environment can be provided by R3 (generally, a commercal agreement will need to be in place). Here, such CorDapps can be further tested in the network configuration they will experience in production, utilising relevant Corda Network Services (including the Identity Operator, and trusted notaries).
```
sartin-r3: I read this that "a bespoke UAT environment can be provided" for each customer. It's slightly different to the Introduction to Corda Network page in Confluence which says that "a bespoke UAT environment isprovided" . Might be worth clarifying, but as Jay didn't notice it, it could just be me.

carolynequinn: yes nice catch. I'll adjust it.

mikehearn:
I've merged it but doesn't this need to be included in a toctree somewhere? It's not linked to from any other part of the docsite, currently.

PR #4713, opened 2019-02-04
[CORDA-2545] Rework package namespace ownership check

mikehearn:
This PR patches up the package namespace ownership check to work as originally specified, and moves it into the attachments classloader. It also fixes a few cases where the unit test simulation code wasn't accurate enough.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
@@ -1,5 +1,6 @@
 package net.corda.core.contracts
 
+import net.corda.core.CordaException
 import net.corda.core.DeleteForDJVM
 import net.corda.core.KeepForDJVM
 import net.corda.core.crypto.SecureHash
@@ -260,13 +261,13 @@ abstract class TransactionVerificationException(val txId: SecureHash, message: S
            Please check the source of this attachment and if it is malicious contact your zone operator to report this incident.
            For details see: https://docs.corda.net/network-map.html#network-parameters""".trimIndent(), null)
 
+    // TODO: Make this descend from TransactionVerificationException so that untrusted attachments cause flows to be hospitalized.
     /** Thrown during classloading upon encountering an untrusted attachment (eg. not in the [TRUSTED_UPLOADERS] list) */
     @KeepForDJVM
     @CordaSerializable
     class UntrustedAttachmentsException(txId: SecureHash, val ids: List<SecureHash>) :
-            TransactionVerificationException(txId, "Attempting to load untrusted transaction attachments: $ids. " +
+            CordaException("Attempting to load untrusted transaction attachments: $ids. " +
                     "At this time these are not loadable because the DJVM sandbox has not yet been integrated. " +
                     "You will need to install that app version yourself, to whitelist it for use. " +
-                    "Please follow the operational steps outlined in https://docs.corda.net/cordapp-build-systems.html#cordapp-contract-attachments to learn more and continue.",
-                    null)
+                    "Please follow the operational steps outlined in https://docs.corda.net/cordapp-build-systems.html#cordapp-contract-attachments to learn more and continue.")
 }
--- a/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
+++ b/core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
@@ -5,6 +5,7 @@ import net.corda.core.KeepForDJVM
 import net.corda.core.crypto.SecureHash
 import net.corda.core.flows.FlowException
 import net.corda.core.identity.Party
+import net.corda.core.node.NetworkParameters
 import net.corda.core.node.services.AttachmentId
 import net.corda.core.serialization.CordaSerializable
 import net.corda.core.utilities.NonEmptySet
@@ -17,7 +18,15 @@ import java.security.PublicKey
  * @property hash Merkle root of the transaction being resolved, see [net.corda.core.transactions.WireTransaction.id]
  */
 @KeepForDJVM
-class TransactionResolutionException(val hash: SecureHash) : FlowException("Transaction resolution failure for $hash")
+open class TransactionResolutionException @JvmOverloads constructor(val hash: SecureHash, message: String = "Transaction resolution failure for $hash") : FlowException(message) {
+    /**
+     * Thrown if a transaction specifies a set of parameters that aren't stored locally yet verification is requested.
+     * This should never normally happen because before verification comes resolution, and if a peer can't provide a
+     * new set of parameters, [TransactionResolutionException] will have already been thrown beforehand.
+     */
+    class UnknownParametersException(txId: SecureHash, paramsHash: SecureHash) : TransactionResolutionException(txId,
+            "Transaction specified network parameters $paramsHash but these parameters are not known.")
+}
 
 /**
  * The node asked a remote peer for the attachment identified by [hash] because it is a dependency of a transaction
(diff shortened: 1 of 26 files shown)


core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -35,20 +36,32 @@ private constructor(
...
         override val id: SecureHash,
+        /** The notary that the tx uses, this must be the same as the notary of all the inputs, or null if there are no inputs. */
         override val notary: Party?,
+        /** The time window within which the tx is valid, will be checked against notary pool member clocks. */
         val timeWindow: TimeWindow?,
+        /** Random data used to make the transaction hash unpredictable even if the contents can be predicted; needed to avoid some obscure attacks. */
         val privacySalt: PrivacySalt,
-        /** Network parameters that were in force when the transaction was notarised. */
+        /**
+         * Network parameters that were in force when the transaction was notarised. This is nullable only for backwards
```
rick-r3: Come to think of it, this must be the network parameters in force when the transaction was constructed.  Not all transactions are notarised so the current comment cannot be correct.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
```diff
@@ -113,21 +126,48 @@ private constructor(
...
+                throw UnsupportedOperationException("Cannot verify a LedgerTransaction created using deprecated constructors outside of flow context.")
+        }
+        return params
+    }
+
+    @StubOutForDJVM
+    private fun getParamsFromFlowLogic(): NetworkParameters? {
+        // TODO: Why can't I make this build, even with the stub/delete annotations?
+        // return FlowLogic.currentTopLevel?.serviceHub?.networkParameters
+        TODO()
```
rick-r3: Are we intending to fix in this PR?

core/src/main/kotlin/net/corda/core/transactions/WireTransaction.kt
```diff
@@ -188,7 +188,7 @@ class WireTransaction(componentGroups: List<ComponentGroup>, val privacySalt: Pr
 
         val resolvedAttachments = attachments.lazyMapped { att, _ -> resolveAttachment(att) ?: throw AttachmentResolutionException(att) }
 
-        val resolvedNetworkParameters = resolveParameters(networkParametersHash) ?: throw TransactionResolutionException(id)
+        val resolvedNetworkParameters = resolveParameters(networkParametersHash) ?: throw TransactionVerificationException.UnknownParameters(id, networkParametersHash!!)
```
rick-r3: I think this should be a resolution exception of some variety.  The caller (e.g. `SignedTransaction.toLedgerTransaction`) says it only throws  `@Throws(SignatureException::class, AttachmentResolutionException::class, TransactionResolutionException::class)` which is probably why it was the way it was.

core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
```diff
@@ -28,6 +28,16 @@ class TransactionResolutionException(val hash: SecureHash) : FlowException("Tran
 @KeepForDJVM
 class AttachmentResolutionException(val hash: SecureHash) : FlowException("Attachment resolution failure for $hash")
 
+/**
+ * Thrown to indicate that a contract attachment is not signed by the network-wide package owner.
+ */
+@CordaSerializable
+class PackageOwnershipException(val attachmentHash: AttachmentId, val contractClass: String, val packageName: String) : FlowException(
```
rick-r3: Need to be careful.  `LedgerTransaction.verify()` only throws `TransactionVerificationException` so we can't throw something that isn't.   Probably why this was inheriting from it before.

core/src/test/kotlin/net/corda/core/contracts/ConstraintsPropagationTests.kt
```diff
@@ -117,6 +117,7 @@ class ConstraintsPropagationTests {
     }
 
     @Test
+    @Ignore    // TODO(mike): rework
```
rick-r3: This to be left `@Ignore` in this PR?

core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
```diff
@@ -224,23 +234,28 @@ abstract class TransactionVerificationException(val txId: SecureHash, message: S
...
-            """The Contract attachment JAR: $attachmentHash containing the contract: $contractClass is not signed by the owner specified in the network parameters.
-           Please check the source of this attachment and if it is malicious contact your zone operator to report this incident.
-           For details see: https://docs.corda.net/network-map.html#network-parameters""".trimIndent(), null)
-
     /**
      * Thrown when multiple attachments provide the same file when building the AttachmentsClassloader for a transaction.
      */
     @CordaSerializable
     @KeepForDJVM
     class OverlappingAttachmentsException(path: String) : Exception("Multiple attachments define a file at path `$path`.")
```
rick-r3: I think this should have inherited `TransactionVerificationException`, since it will be thrown inside `verify` and is a nested inside it?

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -280,7 +333,8 @@ object AttachmentURLStreamHandlerFactory : URLStreamHandlerFactory {
 @KeepForDJVM
 @CordaSerializable
 class UntrustedAttachmentsException(val ids: List<SecureHash>) :
-        CordaException("Attempting to load untrusted Contract Attachments: $ids" +
-                "These may have been received over the p2p network from a remote node." +
-                "Please follow the operational steps outlined in https://docs.corda.net/cordapp-build-systems.html#cordapp-contract-attachments to continue."
+        CordaException("Attempting to load untrusted transaction attachments: $ids. " +
```
rick-r3: I think this also needs to be a `TransactionVerificationException` as it's thrown inside verify.

core/src/main/kotlin/net/corda/core/transactions/LedgerTransaction.kt
tlil: Chris fixed it in this commit (different PR): https://github.com/corda/corda/pull/4709/commits/fa09a462d4c1453a7465721be691685b7bbce52a

core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
```diff
@@ -224,23 +234,28 @@ abstract class TransactionVerificationException(val txId: SecureHash, message: S
...
 
+    /**
+     * Thrown when a transaction appears to be trying to downgrade a state to an earlier version of the app that defines it.
+     * This could be an attempt to exploit a bug in the app, so we prevent it.
+     */
     @KeepForDJVM
     class TransactionVerificationVersionException(txId: SecureHash, contractClassName: ContractClassName, inputVersion: String, outputVersion: String)
         : TransactionVerificationException(txId, " No-Downgrade Rule has been breached for contract class $contractClassName. " +
             "The output state contract version '$outputVersion' is lower that the version of the input state '$inputVersion'.", null)
```
tlil: is lower than

core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
```diff
@@ -224,23 +234,28 @@ abstract class TransactionVerificationException(val txId: SecureHash, message: S
...
+
+    /**
+     * Thrown if a transaction specifies a set of parameters that aren't stored locally yet verification is requested.
+     * This should never normally happen because before verification comes resolution, and if a peer can't provide a
+     * new set of parameters, [TransactionResolutionException] will have already been thrown beforehand.
+     */
+    @KeepForDJVM
+    class UnknownParameters(txId: SecureHash, paramsHash: SecureHash) : TransactionVerificationException(txId,
```
tlil: `UnknownParametersException`

mikehearn:
Thanks. I realised I managed to bugger up the PRs, so this one isn't based on the other branch. Sorry! I'll fix the review comments then consolidate into a single PR.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -188,6 +101,137 @@ class AttachmentsClassLoader(attachments: List<Attachment>, parent: ClassLoader
...
+    init {
+        val untrusted = attachments.mapNotNull { it as? ContractAttachment }.filterNot { isUploaderTrusted(it.uploader) }
+                .map(ContractAttachment::id)
+        if (untrusted.isNotEmpty())
+            throw UntrustedAttachmentsException(untrusted)
+        checkAttachments(attachments)
+    }
+
+    // This function attempts to strike a balance between security and usability when it comes to the no-overlap rule.
+    // TODO - investigate potential exploits.
```
tlil: Is this JIRA'fied? Or just left for later?

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -188,6 +101,137 @@ class AttachmentsClassLoader(attachments: List<Attachment>, parent: ClassLoader
...
+                    // filesystem tries to be case insensitive. This may break developers who attempt to use ProGuard.
+                    //
+                    // Also convert to Unix path separators as all resource/class lookups will expect this.
+                    val path = entry.name.toLowerCase(Locale.US).replace('\\', '/')
+
+                    // Namespace ownership. We only check class files: resources are loaded relative to a JAR anyway.
+                    if (path.endsWith(".class")) {
+                        // Get the package name from the file name. Inner classes separate their names with $ not /
+                        // in file names so they are not a problem.
+                        val pkgName= path.dropLast(6).replace('/', '.').split('.').dropLast(1).joinToString(".")
```
tlil: ".class".length for readability?

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -188,6 +101,137 @@ class AttachmentsClassLoader(attachments: List<Attachment>, parent: ClassLoader
...
+                // checks are only applied during a file read. Merely opening a signed JAR does not imply
+                // the files within it are correctly signed, but, we wish to verify package ownership
+                // at this point during construction because otherwise we may conclude a JAR is properly
+                // signed by the owners of the packages, even if it's not. We'd eventually discover that fact
+                // when trying to read the class file to use it, but if we'd made any decisions based on
+                // perceived correctness of the signatures or package ownership already, that would be too late.
+                attachment.openAsJAR().use { JarSignatureCollector.collectSigners(it) }
+            }
+            // Now open it again to compute the overlap and package ownership data.
+            attachment.openAsJAR().use { jar ->
```
tlil: Could we get some explicit test coverage for this? Would perhaps be easier if some of the functionality was pulled out into a helper class, and you could then test some of the cases that the overlap-checker is looking for without too much faff? :)

core/src/test/kotlin/net/corda/core/contracts/ConstraintsPropagationTests.kt
mikehearn: Yes, this PR will merge to a side branch, not master.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
mikehearn: I fixed this one and left the other for now, as I didn't touch that code in this PR.

core/src/main/kotlin/net/corda/core/contracts/TransactionVerificationException.kt
rick-r3: This still needs fixing otherwise we break the checked exceptions API for Java developers, since `verify` is throwing an exception type not in the `@Throws`

vkolomeyko:
The test seems to have hung:
```
[16:52:54]	[net.corda.node.services.AttachmentLoadingTests] net.corda.node.services.AttachmentLoadingTests.contracts downloaded from the network are not executed without the DJVM (2h:06m:10s)
```

mikehearn:
@vkolomeyko Yep. The issue is that on master, unknown/untrusted attachments will cause a propagating flow death, which isn't quite right. I've made a small change to avoid triggering that test and repushed, but the next PR will have to be to fix it - the right behaviour is surely to wait for the app to be upgraded, just like if it was missing.

PR #4714, opened 2019-02-04
[CORDA-2473] - Remove AMQP system property

dimosr:
With the recent changes in C4, the AMQP system property is effectively not used anymore, since Cordapps are loaded as attachments & the RPC client scans the classpath for any custom serializers/whitelists. This change removes this property and the scanning performed in `AbstractAMQPSerializationScheme` which is dead code currently, in an effort to simplify this part of the code base a bit.

Refer to the JIRA for more details.


mikehearn:
Let's hold this until after C4.0

PR #4715, opened 2019-02-04
[CORDA-2499] Set default UnknownConfigKeysPolicy to IGNORE

anthonykeenan:
Since Corda v4, the Corda node fails to start up by default if unknown configuration is found in the node.conf file.

It's been added to stop confusion for new developers, but it's causing backwards compatbility issues.

There was also an additional option `WARN` alongside `FAIL` and `IGNORE`, but that seems to have disappeared. That would have been a better default. It sort of makes the whole option redundant now if it's going back to the old way. Not sure whether it would be better to delete it altogether.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
+++ b/node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
@@ -40,7 +40,7 @@ open class SharedNodeCmdLineOptions {
             names = ["--on-unknown-config-keys"],
             description = ["How to behave on unknown node configuration. \${COMPLETION-CANDIDATES}"]
     )
-    var unknownConfigKeysPolicy: UnknownConfigKeysPolicy = UnknownConfigKeysPolicy.FAIL
+    var unknownConfigKeysPolicy: UnknownConfigKeysPolicy = UnknownConfigKeysPolicy.IGNORE
 
     @Option(
             names = ["-d", "--dev-mode"],


szymonsztuka:
An unknown property key prefixed with `custom.` will pass the unknown key check. At least there is one routine to allow that, not sure if it's now invoked all the time, see: https://github.com/corda/corda/blob/a5fb1a82f153ab3084045b0e368d727538682796/node-api/src/main/kotlin/net/corda/nodeapi/internal/config/ConfigUtilities.kt#L77

anthonykeenan:
Ok, so there is a way forward - but presumably braid requires the config in the root so we’re still breaking backwards compatibility with configuration that worked before. Is it worth reaching out to the braid guys to see whether they can change this? Wdyt @tlil?

tlil:
The parsing is stricter on v4 master at the moment. As in, it fails on config options that it previously ignored during validation that an app could subsequently read.

mikehearn:
Oh, right, we should be ignoring config options intended for apps, for sure.

szymonsztuka:
but we won't be sure if a config is intended for an app , e.g.  `dataSourcePorpertiesZ` is an app or misspelled `dataSourcePorperties`, the best would be to have specific entry where such properties are placed/ or some app name to property key discovery, but this is probably post 4.0

tlil:
There's the grouping under `custom` that could be a structured way of differentiating. But in the ticket, it seems like `braid` was attempted used at the top-level, and thus failing. :-)

anthonykeenan:
Ok, I'm going to close this PR - the actual solution here is to move the config into the _cordapp_ configuration file, and make sure it's properly documented, I'll raise a seperate PR for that.

tlil:
Thanks, @anthonykeenan

PR #4716, opened 2019-02-05
State pointers are always resolved, so the kdocs should reflect this.

roger3cev:
Just a kdocs change. We should get this on the release branch if we can! Someone on slack pointed it though.


roger3cev:
Thanks @exFalso !

PR #4726, opened 2019-02-08
Disable the forward compatibility check in state deserialisation.

mikehearn:
See https://groups.io/g/corda-dev/topic/more_versioning_notes/29676422?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,29676422 for details.


mikehearn:
@r3domfox Why didn't any unit tests fail?

r3domfox:
The test `preservesDataWhenFlagSet` in `EvolutionSerializerFactoryTests` tests that the evolution serializer enforces the forward compatibility test when configured to do so. There is no test that `AttachmentsClassLoader` creates a factory that is configured in this way - it's difficult to validate this outside of an acceptance test-like scenario.

It would be useful to define an acceptance test for this PR, validating that the proposed code change genuinely disables the forward compatibility test.

mikehearn:
Agreed.

PR #4727, opened 2019-02-08
CORDA-2563: Assign external IDs when creating new keys

roger3cev:
Added two overloads in key management service to create a key pair and assign it to an ID. Ideally we need this in the Corda 4 release.

JIRA: https://r3-cev.atlassian.net/browse/CORDA-2563

Thoughts? Objections?

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -7749,7 +7749,6 @@ public class net.corda.testing.node.MockServices extends java.lang.Object implem
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, int, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair...)
-  public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair[], int, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.node.cordapp.CordappLoader, net.corda.core.node.services.IdentityService, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity, java.security.KeyPair[], net.corda.core.node.services.KeyManagementService, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.testing.core.TestIdentity, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity...)
   public <init>(net.corda.testing.core.TestIdentity, net.corda.testing.core.TestIdentity...)
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -7750,6 +7750,7 @@ public class net.corda.testing.node.MockServices extends java.lang.Object implem
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, int, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair...)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair[], int, kotlin.jvm.internal.DefaultConstructorMarker)
+  public <init>(net.corda.node.cordapp.CordappLoader, net.corda.core.node.services.IdentityService, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity, java.security.KeyPair[], net.corda.core.node.services.KeyManagementService, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.testing.core.TestIdentity, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity...)
   public <init>(net.corda.testing.core.TestIdentity, net.corda.testing.core.TestIdentity...)
   public final void addMockCordapp(String)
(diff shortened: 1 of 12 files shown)


roger3cev:
@roastario You are right, we do.. Will pick it up on Monday. Cheers!

mikehearn:
Isn't in memory KMS needed for people to write unit tests for apps that use this feature?

roger3cev:
@mikehearn I'm not sure how to easily add support for accounts using the in memory KMS. Accounts requires database persistence to work properly :(

Maybe we could add a new interface that supports these methods only?

mikehearn:
Why does it require persistence to work properly? Isn't it conceptually just a HashMap?

roger3cev:
Right but the vault service needs access to this mapping for it to be useful.

mikehearn:
OK, but the KeyManagementService is an abstraction. I take it the vault code is bypassing the abstraction to access the underlying storage directly, instead of calling through the interface?

roger3cev:
@mike I'm not sure I understand?! The vault uses the table created by the persistent key management service variants to map public keys to "account ids". There's no way it would do this via the key management interface, right ?

roger3cev:
I think this is OK, as is now. @roastario WDYT ?

roger3cev:
I'm at a loss of what to do with this... I neglected the fact that `MockServices` also needs support for `externalId`s. However, as the mapping from state participant to external ID is done via Liquibase and Hibernate, there's no way to map states to external IDs using the in memory / mock variants of Corda services.

roger3cev:
Oh actually, ignore me... The vault isn't even supported via `MockServices`: 

`override val vaultService: VaultService get() = throw UnsupportedOperationException()`

What we need to do, is up `makeTestDatabaseAndMockServices` to create a `PersistentKeyManagementService` instead of using the in memory variant.

roger3cev:
@mikehearn 

Tried the above and cannot get it to work...

also tried just replacing all of the mock services stuff with mock nodes but the mock nodes use mock E2E key management service which is not persistent, so that doesn't work either. 

I think we just have to note that the only way to test "ACCOUNTS" is when using the node driver.

mikehearn:
Alright, can you file a JIRA ticket to make this feature work when using MockServices? If we need to switch it to using the persistent KMS then so be it - the backing db will still be an in memory h2 database.

node/src/main/kotlin/net/corda/node/services/keys/E2ETestKeyManagementService.kt
```diff
@@ -61,10 +61,18 @@ class E2ETestKeyManagementService(val identityService: IdentityService, private
         return keyPair.public
     }
 
+    override fun freshKey(externalId: UUID): PublicKey {
+        throw UnsupportedOperationException("")
```
mikehearn: Could you add a more useful message than the empty string, here?

node/src/main/kotlin/net/corda/node/services/keys/KeyManagementServiceInternal.kt
```diff
@@ -1,8 +1,32 @@
...
+import java.security.PublicKey
+import java.util.*
+import javax.persistence.*
 
 interface KeyManagementServiceInternal : KeyManagementService {
     fun start(initialKeyPairs: Set<KeyPair>)
 }
+
+@Entity
+@Table(name = "pk_hash_to_ext_id_map", indexes = [Index(name = "pk_hash_to_xid_idx", columnList = "public_key_hash")])
```
mikehearn: Does adding this table require a Liquibase migration script?

roger3cev:
@mikehearn made a bit more progress with it... Was throwing my toys out of the pram earlier :) The mapping now works for mock services. Now we just need to add the capability for mock nodes to use a key management service other than the `E2ETestKeyManagementService`, which shouldn't take that much effort.

node/src/main/kotlin/net/corda/node/services/keys/KeyManagementServiceInternal.kt
roger3cev: Yeah, there already is a script. There is also a bug associated with it that James Higgs is working on.

node/src/main/kotlin/net/corda/node/services/keys/E2ETestKeyManagementService.kt
roger3cev: Will do. Cheers

roger3cev:
P.S. I'll fix the broken tests tomorrow!

roger3cev:
I think this is good to go now....

roger3cev:
@mikehearn this is good to go now. Manged to fix the issues. Cheers

tlil:
@mnesbit as agreed, dismissed the review. Good for you or @rick-r3 to approve if you're happy with the updates.

.ci/api-current.txt
```diff
@@ -7749,7 +7749,7 @@ public class net.corda.testing.node.MockServices extends java.lang.Object implem
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, int, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair...)
-  public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair[], int, kotlin.jvm.internal.DefaultConstructorMarker)
+  public <init>(net.corda.node.cordapp.CordappLoader, net.corda.core.node.services.IdentityService, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity, java.security.KeyPair[], net.corda.core.node.services.KeyManagementService, kotlin.jvm.internal.DefaultConstructorMarker)
```
rick-r3: This is how kotlin privately reveals a private constructor to a privileged subclass.  It makes it public, but requires a `kotlin.jvm.internal.DefaultConstructorMarker` param which cannot be constructed.  So no one can be using it, and thus API safe.

In this case a prior `object : MockServices(args)` is replaced with another.

roger3cev:
@tlil Hey Tommy, can we merge? Cheers!

tlil:
Merged per Rick's approval

PR #4729, opened 2019-02-11
[CORDA-2574] Prevent non null values from being discarded

r3domfox:
We can prevent non-null values from being discarded during evolution serialisation by enforcing a rule inside the `ObjectBuilder` that constructs a deserialised object instance from a series of property values.

An `ObjectBuilder` holds a set of "slots" which map an ordered list of property values either to an array of constructor parameters, or to a collection of setter methods to be invoked after the target object has been constructed. The ordering of the values supplied to the `ObjectBuilder` corresponds to the order in which they appear in the serialised stream.

When a POJO-like class is evolved by re-ordering, adding and/or removing constructor parameters, an `EvolutionObjectBuilder` is used, which re-routes values supplied to its slots to the corresponding slots in a wrapped `ObjectBuilder`. For example, suppose we have an original class

```kotlin
class A(val b: Int, val a: Int)
```

with slot mappings

* slot 0 -> `a`, in constructor parameter index 1
* slot 1 -> `b`, in constructor parameter index 0

(since property values are typically written into the stream in alphabetical order by property name).

The evolved class is

```kotlin
class A(val a: Int, val c: Int?)
```

The value of `b` is discarded - it has nowhere to go in the new class - and the value of `a` is now written into the first constructor parameter. In this case, we have an `ObjectBuilder` for the new class with mappings 

* slot 0 -> `a`, in constructor parameter index 0
* slot 1 -> `c`, in constructor parameter index 1

To reconcile this with data received for the old class, we use an `EvolutionObjectBuilder` which wraps this `ObjectBuilder`, re-assigning values to slots as follows:

* slot 0 -> `ObjectBuilder` slot 0 -> `a`, in constructor parameter index 0
* slot 1 -> discarded, as there is no slot for `b`

Note that there is no slot mapped to `ObjectBuilder` slot 1, as the old-style class will have no serialised property value for `c`.

We can enforce that non-null values are not discarded during evolution deserialisation, by setting a flag on `EvolutionObjectBuilder` that indicates whether to allow non-null values to be written to a "discarded" slot. If this flag is set, and a non-null value is received into such a slot, an error is thrown with the message "Non-null value [value] provided for property [property name], which is not supported in this version".


mikehearn:
Thanks. So to ensure I understand this properly, if I evolve a class like this:

`data class Foo(val a: String)`

into

`data class Foo(val a: String, val b: Int)`

then this will throw when deserialising a v2 Foo onto a v1 class (with the mode activated) because there's no slot for the `b`. But if it's:

`data class Foo(val a: String, val b: Int?)`

then it'll work, because the slot will become from the Java perspective a `@Nullable Integer b` instead of `int b` and the default will be null.

I think this change looks correct from the code, but we'll need to discuss the rules in the docsite if we don't already.

r3domfox:
I've updated the design doc, which seems to be the only place where this is discussed.

The situation is now that if a state originates with version A, version B (with a new, nullable field) will accept it, filling in the new field with null, and will be able to send the same data back to A, which will accept the null value for the missing field. If a state originates with version B, and has a non-null value for the missing field, then version A must reject it.

PR #4731, opened 2019-02-11
Add a section to the coding guidelines about updating the docsite.

mikehearn:


mikehearn:
Actually I'd rather delete that checklist. I've never seen it actually help (we still have people adding public APIs with no documentation), and often people don't even bother to delete it from their PR descriptions!

PR #4740, opened 2019-02-12
Add a gitattributes file

blsemo:
Add a gitattributes file to make sure docker docs builds work on windows - otherwise the windows line endings make bash in docker very sad.


blsemo:
Actually not required, there is a .gitattributes file in the docs folder already.

PR #4743, opened 2019-02-12
[CORDA-2851] - Fix the way serialization whitelist is calculated for CordappImpl

dimosr:
### Changes
This change includes the actual whitelisted classes in the set of all classes of a cordapp.

### Testing
* Added a unit test that was previously failing, because the name of the `CordappImpl` was added, instead of the actual set of whitelisted classes.


dimosr:
rebased and kept only the changes for the whitelist.
The duplicate warnings has already been fixed by Dom in a separate PR: 
https://github.com/corda/corda/pull/4848

PR #4748, opened 2019-02-12
Updates docs to match new cordapp-example structure.

joeldudleyr3:
Updates docs to match new cordapp-example structure.


joeldudleyr3:
Related to https://github.com/corda/samples/pull/5.

PR #4754, opened 2019-02-14
[CORDA-2561] Use the attachments classloader to deserialise contract states in migrations

JamesHR3:
[CORDA-2487](https://r3-cev.atlassian.net/browse/CORDA-2487) introduces a custom liquibase migration to populate the state_party table and correctly mark states as relevant or not when upgrading from V3. In order to do this, it must deserialise transactions in the database to obtain the output contract states, and from these read the participants in the state. If the jar defining the contract state is on the classpath (as is likely if the migration is run at node startup), then this will succeed. However, if this isn't the case, then whether the participants can be read depends on how the state was serialised. The carpenter will only provide a readable participants list if this field was not calculated, or was marked as a `SerializableCalculatedProperty`. This could result in states skipped.

This change instead attempts to deserialise the states using the attachments classloader. As the transaction should have attachments containing the jars defining the contract states, this should mean that all states can be deserialised. I have attempted this manually and this appears to be the case. However, I have left the old code in as a fallback in the event that some states do not have the relevant attachments.

This also means that the unit tests can still be used to test the migrations manipulate the states correctly. However, as there is no mechanism for getting the attachments for the states in the unit tests without using the full infrastructure of the node, there is no way of testing that the attachment deserialisation part works in the unit tests. As such, a round of manual testing has been carried out to ensure this works.

Changes:
 - Use an attachments classloader context to deserialise states from the transaction output component group
 - This requires the network parameters, so add some code to work out the latest from the database. If there aren't any, fallback to some defaults (I don't think it matters what these are)
 - Improve logging and comments

### DIFF
--- a/node/src/main/kotlin/net/corda/node/migration/CordaMigration.kt
+++ b/node/src/main/kotlin/net/corda/node/migration/CordaMigration.kt
@@ -7,18 +7,20 @@ import liquibase.database.jvm.JdbcConnection
 import liquibase.exception.ValidationErrors
 import liquibase.resource.ResourceAccessor
 import net.corda.core.identity.CordaX500Name
+import net.corda.core.node.NetworkParameters
 import net.corda.core.schemas.MappedSchema
+import net.corda.node.internal.DBNetworkParametersStorage
 import net.corda.node.services.api.WritableTransactionStorage
 import net.corda.node.services.identity.PersistentIdentityService
-import net.corda.node.services.persistence.AbstractPartyToX500NameAsStringConverter
-import net.corda.node.services.persistence.DBTransactionStorage
-import net.corda.node.services.persistence.PublicKeyToTextConverter
+import net.corda.node.services.persistence.*
 import net.corda.nodeapi.internal.persistence.CordaPersistence
 import net.corda.nodeapi.internal.persistence.DatabaseConfig
 import net.corda.nodeapi.internal.persistence.SchemaMigration.Companion.NODE_X500_NAME
 import java.io.PrintWriter
 import java.sql.Connection
 import java.sql.SQLFeatureNotSupportedException
+import java.time.Clock
+import java.time.Duration
 import java.util.logging.Logger
 import javax.sql.DataSource
 
@@ -45,6 +47,16 @@ abstract class CordaMigration : CustomTaskChange {
 
     private lateinit var _dbTransactions: WritableTransactionStorage
 
+    val attachmentsService: AttachmentStorageInternal
+        get() = _attachmentsService
+
+    private lateinit var _attachmentsService: AttachmentStorageInternal
+
+    val latestNetworkParams: NetworkParameters
+        get() = _latestNetworkParams
+
+    private lateinit var _latestNetworkParams: NetworkParameters
+
     /**
      * Initialise a subset of node services so that data from these can be used to perform migrations.
      *
@@ -66,6 +78,10 @@ abstract class CordaMigration : CustomTaskChange {
         cordaDB.transaction {
             identityService.ourNames = setOf(ourName)
             _dbTransactions = DBTransactionStorage(cordaDB, cacheFactory)
+            _attachmentsService = NodeAttachmentService(metricRegistry, cacheFactory, cordaDB)
+            _latestNetworkParams = DBNetworkParametersStorage.createParametersMap(cacheFactory).allPersisted()
+                    .map { it.second.verified() }
+                    .maxBy { it.epoch } ?: defaultNetworkParameters()
         }
     }
 
@@ -83,6 +99,21 @@ abstract class CordaMigration : CustomTaskChange {
         // Liquibase handles closing the database connection when migrations are finished. If the connection is closed here, then further
         // migrations may fail.
         return CordaPersistence(configDefaults, schema, jdbcUrl, cacheFactory, attributeConverters, closeConnection = false)
+    }
+
+    private fun defaultNetworkParameters(): NetworkParameters {
+        val clock = Clock.systemUTC()
+        return NetworkParameters(
+                1,
+                listOf(),
+                1,
+                1,
+                 clock.instant(),
+                1,
+                mapOf(),
+                Duration.ZERO,
+                mapOf()
+        )
     }
 
     override fun validate(database: Database?): ValidationErrors? {
--- a/node/src/main/kotlin/net/corda/node/migration/CordaMigration.kt
+++ b/node/src/main/kotlin/net/corda/node/migration/CordaMigration.kt
@@ -7,10 +7,7 @@ import liquibase.database.jvm.JdbcConnection
 import liquibase.exception.ValidationErrors
 import liquibase.resource.ResourceAccessor
 import net.corda.core.identity.CordaX500Name
-import net.corda.core.node.NetworkParameters
 import net.corda.core.schemas.MappedSchema
-import net.corda.node.internal.DBNetworkParametersStorage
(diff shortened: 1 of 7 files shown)


JamesHR3:
The tests I've carried out:
 - Start up a V3 node and generate 1500 cash states, then upgrade to a V4. Check state party is populated.
 - Using the same database, upgrade to V4, and remove the cash contract cordapp jar (and everything else for good measure). Check that the state party table is populated, and the debug log indicating that attachment deserialisation failed is not present.
This database also had a consumed state that was (correctly) ignored.

node/src/main/kotlin/net/corda/node/migration/VaultStateMigration.kt
```diff
@@ -47,18 +49,36 @@ class VaultStateMigration : CordaMigration() {
...
+            // serialised such that the participants list cannot be accessed (participants is calculated and not marked as a
+            // SerializableCalculatedProperty.
             throw VaultStateMigrationException("Cannot add state parties as state class is not on the classpath " +
                     "and participants cannot be synthesised")
         }
     }
 
+    private fun extractStateFromTx(tx: SignedTransaction, stateIndex: Int): TransactionState<ContractState> {
+        return try {
+            val attachments = tx.tx.attachments.mapNotNull { attachmentsService.openAttachment(it)}
```
tudor-malene: This could fail for Notary change transactions or explicit upgrades because `SignedTransaction.tx` would throw.
I don't know if that's a big problem.
If it is you could add more logic to handle that.

JamesHR3: Potentially! I'll look into it

rick-r3: Well spotted @tudor-malene !  Yes, we need a `when` inside `extractStateFromTx` and branch to different functions for each type.

node/src/main/kotlin/net/corda/node/migration/CordaMigration.kt
```diff
@@ -85,6 +101,21 @@ abstract class CordaMigration : CustomTaskChange {
         return CordaPersistence(configDefaults, schema, jdbcUrl, cacheFactory, attributeConverters, closeConnection = false)
     }
 
+    private fun defaultNetworkParameters(): NetworkParameters {
+        val clock = Clock.systemUTC()
+        return NetworkParameters(
+                1,
```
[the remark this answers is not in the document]
JamesHR3: It isn't, no. One case where there will not be network parameters in the vault is if the database has been created in dev mode, so having some dummy network parameters there is helpful. I'll add some logging around this to indicate that these are being used though.

JamesHR3:
It turns out supporting notary change and contract upgrade transactions was slightly harder than I hoped - the outputs for these are not calculated until the transaction is resolved to a ledger transaction, and it cannot be guaranteed that these transactions have only one input. As a result, I've made the following changes:
 - Implement a new `ServicesForResolution` (note that all the pieces required were already available, this just formalises the loadState(s) methods, borrowing from the logic in `ServicesForResolutionImpl`).
 - Move the network parameters calculation stuff into this class.
 - Add a test case covering a notary change transaction. Unfortunately trying to do the same for contract upgrades falls foul of the same issue that affects the attachments classloading, so I have not added a test for this.

node/src/main/kotlin/net/corda/node/migration/CordaMigration.kt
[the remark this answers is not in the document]
JamesHR3: Actually, thinking about it, I don't think that's true. I tried testing without these dummy parameters and it failed though, so I need to go back and work out why that would be.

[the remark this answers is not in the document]
rick-r3: When you first register the node, there's no network parameters in the database, but the database has to be created, and this migration will run.  So I think you need the dummy value.

BUT, we won't actually find anything to convert, so they're not actively used in anger.

node/src/main/kotlin/net/corda/node/migration/MigrationServicesForResolution.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    companion object {
+        val logger = contextLogger()
+    }
+    override val cordappProvider: CordappProvider
+        get() = throw NotImplementedError()
+
+    private fun defaultNetworkParameters(): NetworkParameters {
+        logger.warn("Using a dummy set of network parameters for migration. This is expected if the database was created by a node started"
+            + " in dev mode.")
```
rick-r3: So do we not upload the network params to the database in dev mode ?

I wouldn't expect this to get called unless there was something in the vault to convert?

JamesHR3: I got that wrong, I think the message needs changing. I think this gets called when `MigrationServicesForResolution` is constructed if there are no network parameters in the database (the `networkParameters` property), so it would get called if there's nothing in the database.

rick-r3: Does it solve it if we do

`override val networkParameters: NetworkParameters get() = `

rather than

`override val networkParameters: NetworkParameters = `

rick-r3: Wait.  That will be slow as it's called a lot. Need it memoize, so:

`override val networkParameters: NetworkParameters by lazy { }`

rick-r3: I was thinking we could make this lazy and never see the message, but actually there's a common path here.  Any CE 3.x node has no network parameter storage, so actually this will always get called.  There seems to be no adverse affects of the dummy value though.  I'm assuming it's not actually used on the code path?  It can't be otherwise the crazy dummy values would be causing a problem.

On that basis, I think we're good here.

JamesHR3:
In running a test for C4 I spotted a good way of testing contract upgrades, so I'd like to do that before merging this.

JamesHR3:
It's a good thing I tested this - I've run into a few issues.

Firstly, I think the parameters may well be looked at for contract upgrades. Fortunately this isn't too hard to fix, as the infrastructure has already been put in for the `AttachmentVersionNumberMigration`. I've done this, and I'll put up a new version with these changes shortly.

Secondly, there appears to be a problem when loading the upgraded contract class while constructing a `ContractUpgradeLedgerTransaction`. The error I'm seeing is:

```
java.lang.ClassNotFoundException: com.upgrade.NewContract
        at java.net.URLClassLoader.findClass(Unknown Source) ~[?:1.8.0_192]
        at java.lang.ClassLoader.loadClass(Unknown Source) ~[?:1.8.0_192]
        at sun.misc.Launcher$AppClassLoader.loadClass(Unknown Source) ~[?:1.8.0_192]
        at java.lang.ClassLoader.loadClass(Unknown Source) ~[?:1.8.0_192]
        at net.corda.core.transactions.ContractUpgradeLedgerTransaction.loadUpgradedContract(ContractUpgradeTransactions.kt:304) ~[corda-core-5.0-SNAPSHOT.jar:?]
```
which seems to be saying that the upgraded contract class cannot be loaded by the classloader. I'll need to investigate this further.

tudor-malene:
@JamesHR3 , that bug is fixed by this PR: https://github.com/corda/corda/pull/4766

JamesHR3:
Ah, thanks @tudor-malene ! I'll retest this once that's in.

PR #4764, opened 2019-02-15
Fixed typo in bootstrapper doc; specify text for anchor

bpaunescu:
While testing the bootstrapper, came across 2 issues:
- 2 typos in the network params override command
- doc reference used title as text which didn't look ok (text as it appeared: more complicated consider using a The network map server.)

This PR fixes the 2 command typos and adds explicit link text to the anchor.


docs/source/network-bootstrapper.rst
```diff
@@ -105,7 +105,7 @@ Modifying a bootstrapped network
 
 The Network Bootstrapper is provided as a development tool for setting up Corda networks for development and testing.
 There is some limited functionality which can be used to make changes to a network, but for anything more complicated consider
-using a :doc:`network-map` server.
+using a :doc:`network-map</network-map>` server.
```
anthonykeenan: The human readable description should be "Network Map" should it not?

bpaunescu: Agreed.

PR #4765, opened 2019-02-15
[CORDA-2591] Prevent spurious logs in AttachmentVersionNumberMigration

JamesHR3:
The `AttachmentVersionNumberMigration` relies on the network parameters being available in order to find what contract implementations are whitelisted. These may not be available if, for example, the node is being bootstrapped or is first connecting to the doorman. However, the migrations will still run in these cases, resulting in an error message being printed to the console. This can be misleading, as it does not prevent operations such as connecting to the doorman from succeeding.

This PR updates the migration to handle this case in such a way that an error message is not printed to the console. The node logs will still contain an indication of what went wrong.

Changes:
 - Detect if the network parameters file is not present while trying to read them from the supplied path
 - Log at the info level if there are no network parameters or attachments
 - Log at the warn level if there are no network parameters but there are attachments
 - Make the X500 name passed to `SchemaMigration` not null

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
@@ -30,7 +30,7 @@ class SchemaMigration(
         private val databaseConfig: DatabaseConfig,
         private val classLoader: ClassLoader = Thread.currentThread().contextClassLoader,
         private val currentDirectory: Path?,
-        private val ourName: CordaX500Name? = null) {
+        private val ourName: CordaX500Name) {
 
     companion object {
         private val logger = contextLogger()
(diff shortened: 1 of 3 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
```diff
@@ -30,7 +30,7 @@ class SchemaMigration(
         private val databaseConfig: DatabaseConfig,
         private val classLoader: ClassLoader = Thread.currentThread().contextClassLoader,
         private val currentDirectory: Path?,
-        private val ourName: CordaX500Name? = null) {
```
JamesHR3: I made this not null because it allows for this class to be constructed in such a way that it results in a runtime error when the migrations are actually run. However, it occurs to me that this may break backwards compatibility in some way - do you know if this is a concern here?

szymonsztuka: As discussed SchemaMigration is not a public API, so this is fine.

PR #4770, opened 2019-02-17
CORDA-2619: Backwards compatibility: put c-i back on the node classpath by default.

mikehearn:
Also fixes some deprecation warnings whilst I'm here.

This will leave the handler open, but we need to revisit how we close the old handlers in future releases, as our current approach is too simplistic and can break under rolling upgrade.


mikehearn:
@shamsasari Reverted the changes and tested with DemoBench + my "upgrade one node to c4" test. Seems to work. Let's see what CI makes of it.

mikehearn:
There are two sample apps in the samples repo that would need adjustment as well: obligation-cordapp and whistleblower

shamsasari:
In V3 the SwapIdentitiesHandler was registered manually (https://github.com/corda/corda/blob/df343379d2acd6bd8ab9856435c2c9c0783ea07c/node/src/main/kotlin/net/corda/node/internal/AbstractNode.kt#L604) because it didn't have the `InitiatedBy` annotation. It does now in V4, but I'm not sure if the node scans locally for initiating-intiatedBy pairs like it does for apps.

mikehearn:
What's the best way to test this? I've been using the CashPaymentFlow and sending cash back and forth with "anonymous: true". I've put back the handler registration.

shamsasari:
Sounds about right, as long as the C-I jar is not in `cordapps` dir.

mikehearn:
OK, so it turns out the node is scanning itself, so the second registration was failing. Trying again ...

confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
```diff
@@ -128,7 +128,7 @@ open class SwapIdentitiesException @JvmOverloads constructor(message: String, ca
 
 // This only exists for backwards compatibility
 @InitiatedBy(SwapIdentitiesFlow::class)
-private class SwapIdentitiesHandler(private val otherSide: FlowSession) : FlowLogic<Unit>() {
+class SwapIdentitiesHandler(private val otherSide: FlowSession) : FlowLogic<Unit>() {
```
shamsasari: Why is this public now? Perhaps move this class into the node?

PR #4773, opened 2019-02-17
ENT-3053 Database connection pools leaking memory on every checkpoint

rick-r3:
Gradle changes are just to align OS build.gradles with ENT

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -156,7 +156,10 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
                 }
                 val continuation = processEvent(transitionExecutor, nextEvent)
                 when (continuation) {
-                    is FlowContinuation.Resume -> return continuation.result
+                    is FlowContinuation.Resume -> {
+                        openThreadLocalWormhole()
+                        return continuation.result
+                    }
                     is FlowContinuation.Throw -> {
                         continuation.throwable.fillInStackTrace()
                         throw continuation.throwable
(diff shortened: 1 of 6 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -208,13 +211,21 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
         MDC.put("thread-id", Thread.currentThread().id.toString())
     }
 
+    private fun openThreadLocalWormhole() {
+        val threadLocal = getTransientField(TransientValues::database).hikariPoolThreadLocal
+        if (threadLocal != null) {
+            val valueFromThread = swappedOutThreadLocalValue(threadLocal)
+            if (valueFromThread != null) threadLocal!!.set(valueFromThread)
```
exFalso: The !! shouldn't be needed

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -208,13 +211,21 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
         MDC.put("thread-id", Thread.currentThread().id.toString())
     }
 
+    private fun openThreadLocalWormhole() {
+        val threadLocal = getTransientField(TransientValues::database).hikariPoolThreadLocal
```
exFalso: I'm wondering, can we do filtering other than comparing specific threadlocals? Perhaps use the class of the value in the threadlocal somehow? That way we don;t need to retrieve a reference 

But in any case, the current comparison works fine for this bug.

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
rick-r3: Correct.  It was warning me.

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
rick-r3: In this case we would be able to find it because it subclasses ThreadLocal, but again, reflection of private classes etc required to get a reference to class we're looking for.

Luckily we have just the one pool, so this is sufficient as is.

I think there's no silver bullet.  If the ThreadLocal was not a subclass, it would be impossible to find it that way.

PR #4776, opened 2019-02-17
CORDA-2624 Tidy docs slightly

gendal:
Changed docs to point to Corda Network and Testnet, and tidied release notes slightly


gendal:
added Kat as reviewer since it touches release notes too

PR #4784, opened 2019-02-18
CORDA-2634 - Fix derivation of currentTargetVersion for abstract classes

tlil:
Abstract flow classes are not associated with CorDapps, and consequently, the node is unable to resolve the correct target platform version.

JIRA: [CORDA-2634](https://r3-cev.atlassian.net/browse/CORDA-2634)

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/cordapp/CordappImpl.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/cordapp/CordappImpl.kt
@@ -34,14 +34,15 @@ data class CordappImpl(
         override val targetPlatformVersion: Int,
         val notaryService: Class<out NotaryService>? = null,
         /** Indicates whether the CorDapp is loaded from external sources, or generated on node startup (virtual). */
-        val isLoaded: Boolean = true
+        val isLoaded: Boolean = true,
+        private val explicitCordappClasses: List<String> = emptyList()
 ) : Cordapp {
     override val name: String = jarName(jarPath)
 
     // TODO: Also add [SchedulableFlow] as a Cordapp class
     override val cordappClasses: List<String> = run {
         val classList = rpcFlows + initiatedFlows + services + serializationWhitelists.map { javaClass } + notaryService
-        classList.mapNotNull { it?.name } + contractClassNames
+        classList.mapNotNull { it?.name } + contractClassNames + explicitCordappClasses
     }
 
     companion object {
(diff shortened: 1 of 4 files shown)


vkolomeyko:
Looks like a unit test `TransactionBuilderTest.multiple commands with same data are joined without duplicates in terms of signers` is not too happy with those changes.

tlil:
> Looks like a unit test `TransactionBuilderTest.multiple commands with same data are joined without duplicates in terms of signers` is not too happy with those changes.

Yes, I'll look at that after I've stepped through a couple of things in the app. There's no obvious reason why that test should fail by this from what I can tell, but it does for any value < 4. Anyway, leave this PR for now. I'm investigating a related issue.

node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
```diff
@@ -145,7 +145,7 @@ class JarScanningCordappLoader private constructor(private val cordappJarPaths:
                 findPlugins(url),
                 findSerializers(this),
                 findCustomSchemas(this),
-                findAllFlows(this),
+                findAllFlows(this, true),
```
shamsasari: Let's use `Cordapp.cordappClasses` instead as it's meant to be an "exhaustive" list of all classes.

Use `RestrictedScanResult.scanResult.getAllStandardClasses()` (and perhaps `.getAllInterfaces()` as well). There's no need to load the classes as `cordappClasses` is just a list of strings, and thus we can avoid the performance penalty there.

tlil: Right, so you want to change the getter defined in `CordappImpl` altogether instead of just bringing in an extra collection? Fine, but keen to keep this change as small as possible

tlil: Ah, yes, fair enough. This is only covering flow classes atm. Update on the way.

tlil:
Updated and green, over to you guys

core/src/main/kotlin/net/corda/core/internal/cordapp/CordappImpl.kt
```diff
@@ -34,14 +34,15 @@ data class CordappImpl(
...
+        val isLoaded: Boolean = true,
+        private val explicitCordappClasses: List<String> = emptyList()
 ) : Cordapp {
     override val name: String = jarName(jarPath)
 
     // TODO: Also add [SchedulableFlow] as a Cordapp class
     override val cordappClasses: List<String> = run {
         val classList = rpcFlows + initiatedFlows + services + serializationWhitelists.map { javaClass } + notaryService
-        classList.mapNotNull { it?.name } + contractClassNames
+        classList.mapNotNull { it?.name } + contractClassNames + explicitCordappClasses
```
shamsasari: You don't need the other stuff as `explicitCordappClasses` has everything. But that's a minor thing that can be cleaned up later.

tlil: Yep, I assumed so, but wanted to play it 100% safe since it's coming in so last minute :-)

PR #4786, opened 2019-02-18
[CORDA-2636] Ensure states created with contract upgrades can be migrated

JamesHR3:
States created via a contract upgrade transaction currently do not migrate correctly, as in order to work out the participants the creating transaction must be deserialised and resolved, and the process of resolving requires a classloader, which is currently not supplied by the custom migration. This PR adds this by providing the cordapp loader used by the `SchemaMigration`, which is used to load classes in these cases. The classloader from this is also used as a parent when deserialising states from normal wire transactions.

Changes:
 - Add an extra field to the `SchemaMigration` class to pass in a `CordappLoader`, and pass this to the custom migration via a thread local
 - Move `CordappLoader` to the node-api module to break a circular dependency
 - Provide a `CordappProvider` implementation that does nothing other than return a `CordappLoader`. This is required to correctly load output states from the contract upgrade transaction.

### DIFF
--- a/.ci/api-current.txt
+++ b/.ci/api-current.txt
@@ -7744,7 +7744,6 @@ public class net.corda.testing.node.MockServices extends java.lang.Object implem
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, int, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair...)
-  public <init>(net.corda.node.cordapp.CordappLoader, net.corda.core.node.services.IdentityService, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity, java.security.KeyPair[], net.corda.core.node.services.KeyManagementService, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.testing.core.TestIdentity, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity...)
   public <init>(net.corda.testing.core.TestIdentity, net.corda.testing.core.TestIdentity...)
   public final void addMockCordapp(String)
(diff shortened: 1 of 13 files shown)


JamesHR3:
I'm putting this up to get it started through the PR checklist. This still doesn't work if the output state classes are not supplied by any CorDapps in the cordapps directory - possibly this could be fixed by providing an attachment classloader using the new attachment as a fallback to the current one, but it's possible we don't want/need to do this. Will discuss tomorrow.

rick-r3:
The API checker failed, but it's the same pattern as before.  A private constructor involving CordappLoader has changed because of its refactor to another package.  It's not publicly callable due to the extra constructor arg that folks can't pass in.  It's okay to remove that line from api-current.txt.  I'll approve.

rick-r3:
I think using just `AttachmentClassLoader`s would have been problematic because the upgrade `AttachmentClassLoader` would need the input state `AttachmentClassLoader` as the parent, otherwise they similarly would have had two different instances of the input class.    Maybe it's fine to just say for now that all explicit upgrades must be on the classpath.  We can revisit explicit upgrade class loading after 4.0 perhaps?

rick-r3:
Otherwise code is looking fine.

.ci/api-current.txt
```diff
@@ -7744,7 +7744,6 @@ public class net.corda.testing.node.MockServices extends java.lang.Object implem
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, int, kotlin.jvm.internal.DefaultConstructorMarker)
   public <init>(net.corda.core.identity.CordaX500Name, net.corda.core.node.services.IdentityService, java.security.KeyPair, java.security.KeyPair...)
-  public <init>(net.corda.node.cordapp.CordappLoader, net.corda.core.node.services.IdentityService, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity, java.security.KeyPair[], net.corda.core.node.services.KeyManagementService, kotlin.jvm.internal.DefaultConstructorMarker)
```
shamsasari: Another one @dimosr. I've created a Jira ticket to track this: https://r3-cev.atlassian.net/browse/CORDA-2638

PR #4790, opened 2019-02-20
Modify 'Docker with Testnet' docs

przemolb:
This change makes the docker example commands easier to script and modify (especially when you want to spin up several Corda nodes as I did during testing).
There should be more changes for the other parts of the docker docs so if you accept this direction I will update the other part of the docs as well.
If you disagree - will give up.

### DIFF
--- a/docs/source/docker-image.rst
+++ b/docs/source/docker-image.rst
@@ -73,22 +73,34 @@ Joining TestNet
 
 .. code-block:: shell
 
-    docker run -ti \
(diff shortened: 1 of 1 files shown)


shamsasari:
This PR is bit-rotting. Can we please bring it to a conclusion.

vkolomeyko:
This PR is now in conflicted state, which needs to be resolved.

TomAnthony (approved):

rnicoll:
@przemolb Can you raise this against the new docs repo please

PR #4793, opened 2019-02-20
CORDA-2655 - Document CorDapp upgradeability guarantees in Corda 4.

josecoll:


docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Corda 4.0
+---------
+
+Corda 4 introduces a number of advanced features (such as signature constraints), and data security model improvements (such as attachments
+trust checking and classloader isolation of contract attachments for transaction building and verification).
+
+The following guarantees are made for CorDapps running on Corda 4.0
+
+- CorDapps compiled with previous versions of Corda (from 1.0) will execute without change on Corda 4.0
```
gendal: soften to 3.0?  (put it this way: if somebody digs out a v1 app and it doesn't work, will we spend time trying to make it work?).  Also qualify: "Compliant CorDapps compiled..."  And add: "Note: by "compliant", we mean CorDapps that do not utilise Corda internal, non-stable or other non-committed public Corda APIs"

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+
+  Recommendation: security hardening changes in flow processing, specifically the ``FinalityFlow``, recommend upgrading existing CorDapp
+  receiver flows to use the new APIs. See :ref:`cordapp_upgrade_finality_flow_ref` for more information.
```
gendal: does this mean opting in to platform version 4? or can you do it without that?

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+
+- CorDapp Contract states generated on ledger using Hash constraints (introduced in Corda 1.0) and CZ Whitelisted constraints (introduced in Corda 3.0)
+  are consumable by CorDapps built on Corda 4 (by default these will use Signature constraints) but may not be combined as inputs into the
+  same transaction. A transaction may consume multiple states of the same constraint type but not combined.
+
+- CorDapp Contract states generated on ledger using Hash constraints are not migratable to Signature constraints in this release.
```
gendal: can you migrate them to zone constraints and then use the manual approach below?

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+The following guarantees are made for CorDapps running on Corda 4.0
+
+- CorDapps compiled with previous versions of Corda (from 1.0) will execute without change on Corda 4.0
+
+  Recommendation: security hardening changes in flow processing, specifically the ``FinalityFlow``, recommend upgrading existing CorDapp
+  receiver flows to use the new APIs. See :ref:`cordapp_upgrade_finality_flow_ref` for more information.
+
+- CorDapp Contract states generated on ledger using Hash constraints (introduced in Corda 1.0) and CZ Whitelisted constraints (introduced in Corda 3.0)
+  are consumable by CorDapps built on Corda 4 (by default these will use Signature constraints) but may not be combined as inputs into the
+  same transaction. A transaction may consume multiple states of the same constraint type but not combined.
```
gendal: This paragraph contradicts the one below, no?  This implies you can consume a hash-constrained state and spit out a sig constrained one... but the next section says you can't?

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+
+- CorDapp contract attachments are no longer trusted from remote peers over the p2p network for the purpose of transaction verification.
+  The implication of this data security hardening step is that a Node operator must locally install *all* versions of a Contract attachment
+  to be able to resolve a chain of contract states from its original version.
+  The RPC ``UploadAttachment`` mechanism can be used to achieve this (as well as conventional loading of a CorDapp by installing it in the nodes /cordapp directory).
+  See :ref:`cordapp_install_ref` and :ref:`cordapp_contract_attachments_ref` for more information.
+
+- CorDapp contract attachment classloader isolation has some important side-effects and edge cases to consider:
+
+  1. Contract attachments should include all 3rd party library dependencies in the same packaged JAR.
```
gendal: should we put "aka 'Fat Jar'" here or something? just so when people see/hear it in other contexts they know we're talking about the same thing?

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+  2. Contract attachments that depend on other Contract attachments are currently supported in so far as the Attachments Classloader
+     will attempt to resolve any external dependencies from the Nodes application classloader. It is thus paramount that dependent Contract
```
gendal: node's

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+     Attachments are loaded upon node startup from the respective /cordapps directory.
+
+- Rolling upgrades are partially supported.
+  A Node operator may choose to manually upload a later version of a Contract Attachment than the version their node is currently using
+  for the purposes of transaction verification (from remote peers). They will only be able to build new transactions with the version
+  that is currently loaded in their node (as installed from the nodes /cordapps directory)
```
gendal: you're using load/upload ambiguously here.  When you say "upload", you're talking about RPC upload attachment?  But when you use "load", you're talking about "obtained from cordapps directory"?

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+
+- Finance CorDapp
+  Whilst experimental, we guarantee that states generated with the Finance CorDapp are interchangeable across Open Source and Enterprise
```
gendal: Specify that we're talking about finance v4 only?  Soften guarantee to "Whilst experimental, our test coverage has included confirming that states..."

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+  distributions. This has been made possible by releasing a single version of the Finance Contracts CorDapp.
+  Please note the Finance application will be superseded shortly by the new Tokens SDK (https://github.com/corda/token-sdk)
+
+Corda 4.1
```
gendal: will it be 4.1?  Or, rather, will it require a minplatformversion bump on any associated network?

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+---------
+
+The following guarantees will be delivered in a close follow-up release to Corda 4.0:
```
gendal: Should we say: "The following additional capabilities are under consideration for delivery in a follow-up release to Corda 4.0"?  ie avoid binding the hands of the product management team if it turns out some are too complex or not being demanded by customers?

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+- Rolling upgrades will be fully supported.
+  A Node operator will be able to pre-register (by hash or code signing public key) versions of CorDapps they are not yet ready to install locally,
+  but wish to use for the purposes of transaction verification with peers running later versions of a CorDapp.
+
+.. note:: Trusted downloading of contract attachments from remote peers will not be available until secure JVM sand-boxing is available.
+
+Corda Enterprise concerns
+-------------------------
+
+- CorDapps compiled with the OS version of Corda 4.0 will execute without change on Enterprise Corda 4.0
```
gendal: add fullstop at end.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,86 @@
...
+  The reverse is not guaranteed. Whilst the Public API's are currently identical, R3 may introduce Enterprise-specific Public API's for
```
gendal: APIs, not API's

docs/source/cordapp-upgradeability.rst
josecoll: Not sure what you mean by opting in to platform version 4.
As the new API's are shipped with the version 4 release I don't think this is optional (eg. you must set a minimum platform version of 4). @mikehearn ?

docs/source/cordapp-upgradeability.rst
josecoll: In theory yes, but I don't believe we have tried this.

docs/source/cordapp-upgradeability.rst
josecoll: Damn I was trying to clarify the original paragraph.
Let me try again ;-)

docs/source/cordapp-upgradeability.rst
josecoll: Yes - I've added this explicitly in this section and the next (to make it clear).

docs/source/cordapp-upgradeability.rst
josecoll: Yes - let me make this clearer.

docs/source/cordapp-upgradeability.rst
josecoll: I don't think it will require a platform version bump. @mikehearn ?
PS: I'm assuming we release 4.1 after 4.0 (despite what our JIRA numbering says)

josecoll:
@fenryka I've added you for visibility of this docs change (to be included in Corda 4.0).

docs/source/cordapp-upgradeability.rst
```diff
@@ -33,25 +35,26 @@ The following guarantees are made for CorDapps running on Corda 4.0
...
+  A Node operator may choose to manually upload (via the RPC attachments uploader mechanism) a later version of a Contract Attachment than
+  the version their node is currently using for the purposes of transaction verification (received from remote peers). However, they will only
+  be able to build new transactions with the version that is currently loaded (installed from the nodes /cordapps directory) in their node.
 
-- Finance CorDapp
-  Whilst experimental, we guarantee that states generated with the Finance CorDapp are interchangeable across Open Source and Enterprise
-  distributions. This has been made possible by releasing a single version of the Finance Contracts CorDapp.
+- Finance CorDapp (v4)
+  Whilst experimental, our test coverage has confirmed that states generated with the Finance CorDapp are interchangeable across Open Source
+  and Enterprise distributions. This has been made possible by releasing a single 4.0 version of the Finance Contracts CorDapp.
```
gendal: Maybe add: "We have also confirmed that the finance v3 app works on Corda v4. However, the finance v3 and v4 apps are not themselves compatible/interoperable"

josecoll: I can add that line but it feels rather negative.
Eg. a Finance-V3 running on Node-V3 should be interoperable with Finance-V4 running on Node-V4. I even thought @mikehearn confirmed this scenario worked ???

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,90 @@
...
+
+- Contract attachments will be able to explicitly declare their dependencies on other Contract attachments such that these are automatically
+  loaded by the Attachments Classloader (rendering the 4.0 fallback to application classloader mechanism redundant).
+  This improved modularity removes the need to "Fat JAR" all dependencies together in a single jar.
+
+- Rolling upgrades will be fully supported.
+  A Node operator will be able to pre-register (by hash or code signing public key) versions of CorDapps they are not yet ready to install locally,
+  but wish to use for the purposes of transaction verification with peers running later versions of a CorDapp.
+
+.. note:: Trusted downloading of contract attachments from remote peers will not be available until secure JVM sand-boxing is available.
```
mikehearn: available -> integrated

docs/source/cordapp-upgradeability.rst
gendal: why would we expect it to?  OK... maybe if the finance app v4 had been written to enable the upgrade/downgrade stuff to work without data loss, etc... but did we design the v4 app to work that way?

josecoll: Finance app is a tricky one because of the C-I angle.
In general we guarantee that a CorDapp written to the same set of API's from one version (eg. V3) should be compatible / interoperable with an evolution of that CorDapp compiled against the same API's in the next release (eg. V4) - eg. min platform Version 3, target platform version 4 in this case. 
Of course, this no longer holds if V4 of the CorDapp uses platform version 4 API's (in which case the min platform version would be 4 and hence cannot by definition interoperate with a mpv 3 built app.)

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,90 @@
...
+- Compliant CorDapps compiled with previous versions of Corda (from 3.0) will execute without change on Corda 4.0
+
+  .. note:: by "compliant", we mean CorDapps that do not utilise Corda internal, non-stable or other non-committed public Corda APIs.
+
+  Recommendation: security hardening changes in flow processing, specifically the ``FinalityFlow``, recommend upgrading existing CorDapp
+  receiver flows to use the new APIs. See :ref:`cordapp_upgrade_finality_flow_ref` for more information.
+
+- CorDapp Contract states generated on ledger using Hash constraints (introduced in Corda 1.0) and CZ Whitelisted constraints (introduced in Corda 3.0)
+  are consumable by CorDapps built on Corda 4 (by default these will use Signature constraints) but may not be combined as inputs into the
+  same transaction. However a single transaction may combine multiple states of the same constraint type.
```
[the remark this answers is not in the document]
josecoll: Added further clarification.
Also added a note about the hash-to-CZ-whitelisting scenario.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,90 @@
...
+
+- CorDapp Contract states generated on ledger using Hash constraints are not migratable to Signature constraints in this release.
+
+- CorDapp Contract states generated on ledger using CZ Whitelisted constraints are migratable to Signature constraints using a manual process
+  that requires programmatic code changes. See :ref:`constraints_whitelist_to_signature_ref` for more information.
+
+- Explicit Contract Upgrades are only supported for Hash and CZ Whitelisted constraint types. See :ref:`explicit_contract_upgrades_ref` for more information.
+
+- CorDapp contract attachments are no longer trusted from remote peers over the p2p network for the purpose of transaction verification.
```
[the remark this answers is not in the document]
josecoll: Will remove this line (I honestly thought this was introduced in V4).

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Corda 4.1
+---------
+
+The following additional capabilities are under consideration for delivery in a follow-up release to Corda 4.0:
+
+- CorDapp contract states issued with different constraint types will be consumable within the same transaction.
+  eg. no longer need to consume hash, CZ whitelist and signature constraints in isolation.
+
+- CorDapp Contract states generated on ledger using Hash constraints will be automatically migrated to Signature constraints when building new transactions.
```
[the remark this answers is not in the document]
josecoll: We need to work through the details of how we deliver this.
The "disable hash-constraint checking" flag is only part of the story as I found out when re-testing.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,90 @@
...
+  Whilst experimental, our test coverage has confirmed that states generated with the Finance CorDapp are interchangeable across Open Source
+  and Enterprise distributions. This has been made possible by releasing a single 4.0 version of the Finance Contracts CorDapp.
+  Please note the Finance application will be superseded shortly by the new Tokens SDK (https://github.com/corda/token-sdk)
+
+Corda 4.1
+---------
+
+The following additional capabilities are under consideration for delivery in a follow-up release to Corda 4.0:
+
+- CorDapp contract states issued with different constraint types will be consumable within the same transaction.
```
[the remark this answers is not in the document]
josecoll: This is not currently possible as tested and explained above.
1. You can combine existing states constrained by the same type
2. You cannot mix existing states issued/generated with different constraint types.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,90 @@
...
+  eg. no longer need to consume hash, CZ whitelist and signature constraints in isolation.
+
+- CorDapp Contract states generated on ledger using Hash constraints will be automatically migrated to Signature constraints when building new transactions.
+
+- CorDapp Contract states generated on ledger using CZ Whitelisted constraints will be automatically migrated to Signature constraints when building new transactions.
+
+- Explicit Contract Upgrades will be supported for all constraint types: Hash, CZ Whitelisted and Signature.
```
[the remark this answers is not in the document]
josecoll: It doesn't presently work because you cannot CZ whitelist a signed jar.
And the current code defaults to checking the CZ whitelist when explicitly upgrading.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,90 @@
...
+  In practice, it should only be necessary to upgrade from hash or CZ whitelisted to new signature constrained contract types.
+  Signature constrained Contracts are upgradeable seamlessly (through built in serialization and code signing controls) without requiring explicit upgrades.
+
+- Contract attachments will be able to explicitly declare their dependencies on other Contract attachments such that these are automatically
```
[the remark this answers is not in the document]
josecoll: Agree it needs some design thinking which may push out the solution to 5.x
But given we have a workaround "hack" right now, why would this need a platform version change? (the Public API's will not change, the Jar metadata will be used to drive dependency classloading)

Tudor mentioned he has a WIP Branch that tackles this already: 
https://github.com/corda/corda/tree/feature/CORDA-2495/fix_contract_package_scanning

docs/source/cordapp-upgradeability.rst
mikehearn: * You don't have to set a min of 4, I don't think that's checked, though it could have been.
* Target version does need to be 4, I think.
* And yeah it means opting in to the new APIs.

docs/source/cordapp-upgradeability.rst
```diff
@@ -16,21 +16,23 @@ The following guarantees are made for CorDapps running on Corda 4.0
...
+- CorDapp Contract states generated on ledger using hash constraints (introduced in Corda 1.0) and CZ whitelisted constraints (introduced in Corda 3.0)
+  are consumable by CorDapps built on Corda 4 (by default these will use signature constraints). However, you may not combine inputs constrained by different
+  types (eg. cannot mix hash and CZ whitelist and signature constrained states) in the same transaction. Note you can combine multiple inputs
+  constrained by the same type (many of hash or CZ whitelist or signature) in the same transaction.
+  An exception to this rule is whre you may have subsequently CZ whitelisted a Jar previously used to issue hash constrained states (eg. the original hash-constrained states
+  and new CZ-whitelisted constrainted states may be combined in a new transaction in this scenario because they resolve to the same original hash).
 
-- CorDapp Contract states generated on ledger using Hash constraints are not migratable to Signature constraints in this release.
+- CorDapp Contract states generated on ledger using hash constraints are not migratable to signature constraints in this release.
```
gendal: I'm still confused... this directly contradicts lines 19-20, no?

docs/source/cordapp-upgradeability.rst
```diff
@@ -59,13 +61,13 @@ The following additional capabilities are under consideration for delivery in a
 - CorDapp contract states issued with different constraint types will be consumable within the same transaction.
   eg. no longer need to consume hash, CZ whitelist and signature constraints in isolation.
 
-- CorDapp Contract states generated on ledger using Hash constraints will be automatically migrated to Signature constraints when building new transactions.
+- CorDapp Contract states generated on ledger using hash constraints will be automatically migrated to signature constraints when building new transactions.
```
gendal: only if the Jar is signed?

docs/source/cordapp-upgradeability.rst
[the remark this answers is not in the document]
gendal: will a v4 node be able to understand and correctly process a transaction from a node that has this feature from the future?

docs/source/cordapp-upgradeability.rst
josecoll: Perhaps this is specific to  "Cash"  (and other fungible assets) where you may consume a hash-constrained (or zone whitelisted constraint) and obtain change (new outputs) using the same constraints as the inputs.
With the original Hash->Signature constraint logic, any subsequent outputs would automatically migrate to using signature-constraints (where a signed jar or the original unsigned jar was present in the node).

josecoll: Should I qualify the sentence somehow:
"CorDapp Contract states generated on ledger using hash constraints are not automatically migrated to signature constraints where a transaction generates further outputs from its inputs in this release. Any new outputs preserve the constraint type of their original inputs".

gendal: The bit that is confusing me is "(by default these will use signature constraints)" on line 20.  If this wasn't there then everything is consistent.  Line 19: you can consume hash-constrained inputs (implication: they'll remain hash constrained). Line 26: you can't migrate hash constraints to sig constraints.  But the bit in brackets on line 20 implies to the reader that if you consume a hash constrained input then by default it will become sig constrained. Is that what you mean to say? Am I misunderstanding?   So yes... the qualification would help.. but maybe the answer is to remove the bit in brackets on line 20 unless it's saying something I don't understand?

josecoll: Got it!. 
Yes - let me remove the misleading "(by default these will use signature constraints)" on line 20.

josecoll:
back to you @gendal

josecoll:
Thanks @gendal . 
@mikehearn final review from you please.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Corda Enterprise concerns
+-------------------------
+
+- CorDapps compiled with the OS version of Corda 4.0 will execute without change on Enterprise Corda 4.0.
+  The reverse is not guaranteed. Whilst the Public APIs are currently identical, R3 may introduce Enterprise-specific Public APIs for
+  advanced CorDapp functionality, therefore invalidating the ability to execute on Open Source nodes.
+  Wire-compatibility and ABI stability is maintained.
+
+- The Finance Contract CorDapp is only available in the Open Source distribution to ensure uniqueness and singularity of JAR "hash".
```
mikehearn: I guess the readers of this document have never seen the ENT codebase and so this doesn't really mean anything to them, as far as they're concerned, there was only ever one version of the app. So I think we can take this out too.

docs/source/cordapp-upgradeability.rst
josecoll: > * You don't have to set a min of 4, I don't think that's checked, though it could have been.

How come? 
"Applications can advertise a minimum platform version they require. If your app uses new APIs that were added in (for example) Corda 5, you should specify a minimum version of 5. This will ensure the app won’t be loaded by older nodes."
According to our definition you would need to specify this to use the new API's.

> * Target version does need to be 4, I think.
"Apps should advertise the highest version of the platform they have been tested against."

> * And yeah it means opting in to the new APIs.

Include in text to make this clear.

docs/source/cordapp-upgradeability.rst
josecoll: Ok - this paragraph is talking about Corda Version 4 only (not 3.x interop).

docs/source/cordapp-upgradeability.rst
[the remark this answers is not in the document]
josecoll: Removed paragraph to avoid confusion.

docs/source/cordapp-upgradeability.rst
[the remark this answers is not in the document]
josecoll: Right - so we're talking about a proper solution using metadata for this (where today there is none). This implies we will need a platform version number bump.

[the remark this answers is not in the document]
josecoll: Agreed - we need a Design Document on how we do this properly in a Java 8 and JDK 9+ world (with jigsaw modules).

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,94 @@
...
+The following guarantees are made for CorDapps running on Corda 4.0
+
+- Compliant CorDapps compiled with previous versions of Corda (from 3.0) will execute without change on Corda 4.0
+
+  .. note:: by "compliant", we mean CorDapps that do not utilise Corda internal, non-stable or other non-committed public Corda APIs.
+
+  Recommendation: security hardening changes in flow processing, specifically the ``FinalityFlow``, recommend upgrading existing CorDapp
+  receiver flows to use the new APIs. See :ref:`cordapp_upgrade_finality_flow_ref` for more information.
+
+- CorDapp Contract states generated on ledger using hash constraints (introduced in Corda 1.0) and CZ whitelisted constraints (introduced in Corda 3.0)
```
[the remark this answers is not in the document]
josecoll: We cannot say 
"States with hash constraints do not allow the associated app to be upgraded. If you issued such states by mistake"
because end users had no choice but to use Hash Constraints in V1.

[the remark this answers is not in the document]
josecoll: To avoid going round in more circles I will just delete the paragraph.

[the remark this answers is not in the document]
josecoll: Removed paragraph to avoid confusion.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,94 @@
...
+- Rolling upgrades will be fully supported.
+  A Node operator will be able to pre-register (by hash or code signing public key) versions of CorDapps they are not yet ready to install locally,
+  but wish to use for the purposes of transaction verification with peers running later versions of a CorDapp.
+
+.. note:: Trusted downloading of contract attachments from remote peers will not be integrated until secure JVM sand-boxing is available.
+
+Corda Enterprise concerns
+-------------------------
+
+- CorDapps compiled with the OS version of Corda 4.0 will execute without change on Enterprise Corda 4.0.
```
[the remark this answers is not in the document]
josecoll: In what way is it not true today?

docs/source/cordapp-upgradeability.rst
josecoll: I will need to add this back for the Enterprise docs.

josecoll:
Back to you @mikehearn

docs/source/cordapp-upgradeability.rst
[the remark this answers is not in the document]
mikehearn: But we can't read data from v1 anyway, v3 is the earliest that made data v4 can read because that's when we introduced wire stability. And in v3 you had zone constraints, so using hash constraints was probably a mistake.

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,84 @@
...
+
+  .. note:: by "compliant", we mean CorDapps that do not utilise Corda internal, non-stable or other non-committed public Corda APIs.
+
+  Recommendation: security hardening changes in flow processing, specifically the ``FinalityFlow``, recommend upgrading existing CorDapp
+  receiver flows to use the new APIs and thus opting in to platform version 4. See :ref:`cordapp_upgrade_finality_flow_ref` for more information.
+
+- All constraint types (hash, CZ whitelisted, signature) are consumable within the same transaction if there is an associated contract attachment that satisfies all of them.
+
+- CorDapp Contract states generated on ledger using hash constraints are not directly migratable to signature constraints in this release.
+  You may have CZ whitelisted a Jar previously used to issue hash constrained states, and then follow the manual process described in the paragraph
```
mikehearn: You may have CZ whitelisted a Jar -> "Your compatibility zone operator may whitelist a JAR previously used to issue hash constrained states, and then you can follow the manual process described ...."

docs/source/cordapp-upgradeability.rst
```diff
@@ -0,0 +1,84 @@
...
+- CorDapp contract attachments are not trusted from remote peers over the p2p network for the purpose of transaction verification.
+  A node operator must locally install *all* versions of a Contract attachment to be able to resolve a chain of contract states from its original version.
+  The RPC ``uploadAttachment`` mechanism can be used to achieve this (as well as conventional loading of a CorDapp by installing it in the nodes /cordapp directory).
+  See :ref:`cordapp_install_ref` and :ref:`cordapp_contract_attachments_ref` for more information.
+
+- CorDapp contract attachment classloader isolation has some important side-effects and edge cases to consider:
+
+  1. Contract attachments should include all 3rd party library dependencies in the same packaged JAR - we call this a "Fat JAR",
+     meaning that all dependencies are resolvable by the classloader by only loading a single Jar.
+  2. Contract attachments that depend on other Contract attachments are currently supported in so far as the Attachments Classloader
```
mikehearn: This bit is slightly confusing. Attachments that depend on other attachments are resolved from the attachments, no? Or that's how it's meant to work!

docs/source/cordapp-upgradeability.rst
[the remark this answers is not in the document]
mikehearn: I was meaning the second part - as you observe, the APIs are currently identical, so whilst we *might* introduce ENT specific APIs in future, we don't currently do so, and therefore, you should be able to compile and run in both directions.

docs/source/index.rst
```diff
@@ -34,6 +34,7 @@ We look forward to seeing what you can do with Corda!
    release-notes
    app-upgrade-notes
    node-upgrade-notes
+   cordapp-upgradeability
```
mikehearn: Maybe we should put this as a sub-page of `app-upgrade-notes`?

docs/source/cordapp-upgradeability.rst
josecoll: How about:
"Contract attachments that depend on other Contract attachments from a different packaged JAR are currently supported ..." ?

docs/source/index.rst
josecoll: > Maybe we should put this as a sub-page of `app-upgrade-notes`?

Good idea

PR #4794, opened 2019-02-20
ENT-3017: TestCordapp – use Gradle tooling API for builds

adagys:
The TestCordappImpl runs gradle to build cordapp jars required for tests. It's causing multiple issues including not actually building a jar or causing the JVM to die with SIGBUS.

Using the Gradle Tooling API to run the builds seems to fix the issues.

### DIFF
--- a/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/TestCordappImpl.kt
+++ b/testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/TestCordappImpl.kt
@@ -4,7 +4,8 @@ import io.github.classgraph.ClassGraph
 import net.corda.core.internal.*
 import net.corda.core.utilities.contextLogger
 import net.corda.testing.node.TestCordapp
-import org.apache.commons.lang.SystemUtils
+import org.gradle.tooling.GradleConnector
+import org.gradle.tooling.ProgressEvent
 import java.nio.file.Path
 import java.util.*
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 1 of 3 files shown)


testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/TestCordappImpl.kt
```diff
@@ -79,7 +79,8 @@ data class TestCordappImpl(val scanPackage: String, override val config: Map<Str
             return projectRootToBuiltJar.computeIfAbsent(projectRoot) {
                 val gradlew = findGradlewDir(projectRoot) / (if (SystemUtils.IS_OS_WINDOWS) "gradlew.bat" else "gradlew")
                 log.info("Generating CorDapp jar from local project in $projectRoot ...")
-                val exitCode = ProcessBuilder(gradlew.toString(), "jar").directory(projectRoot.toFile()).inheritIO().start().waitFor()
+                // The inner Gradle process has to be run with daemon disabled as it might reuse the same Gradle daemon as the main process running the tests, which leads to JVM crashing.
```
rick-r3: I don't think that is what is happening, but I could be wrong.  Locally it looks like it just spins up a new daemon, and the crash is because it rebuilds a JAR file that is in use.

Intrigued that this now causes some integration test failures?  Hmmmm....

adagys: Hmm yeah, and it's the same TraderDemo test that's failing! Although in a different way now

mikehearn:
I think all this code needs is the version metadata, right?

So grep the build.gradle file for the relevant lines and parse them? Hacky but should work and be fast.

adagys:
@mikehearn not exactly sure what you mean. The current problem I'm getting in OS is that the trader demo workflows jar doesn't get built when running gradle programmatically via the TestCordappImpl

adagys:
Ok this seems to work on both OS and ENT

testing/node-driver/src/main/kotlin/net/corda/testing/node/internal/TestCordappImpl.kt
```diff
@@ -77,24 +78,31 @@ data class TestCordappImpl(val scanPackage: String, override val config: Map<Str
...
+            val build = projectConnection.newBuild().apply {
+                forTasks("jar")
+                addProgressListener { event: ProgressEvent ->
+                    log.info(event.description)
                 }
-                current = current.parent
             }
+            // Blocks until the build is complete
+            build.run()
+            projectConnection.close()
```
shamsasari: Use `use`

shamsasari:
Huh? Was there an urgency to merge this? I would have liked by comment to have been addressed.

adagys:
Haha, just pushed an update.

fenryka:
Sorry, Andrius asked me to merge it so I assumed it was good to go and the block was because no one had reviewed it rather than blocking

fenryka:
We can alwways push those requested changes in another PR

shamsasari:
Yep, no problem. Thanks.

PR #4795, opened 2019-02-20
[CORDA-2626] Docsite: add a node upgrading page and repair misc damaged pages

mikehearn:
Two big changes:

1. Add a page that describes how to drain and upgrade a node (adapted from the enterprise and other doc pages)
2. Many small markup repairs and improvements to fix bugs introduced by people who aren't testing their docs changes!


docs/source/changelog.rst
```diff
@@ -43,10 +17,10 @@ Version 4.0
   retries have been disabled for single node notaries since in this case they offer no potential benefits, unlike for a notary cluster with
   several members who might have different availability.
 
-* New configuration property `database.initialiseAppSchema` with values `UPDATE`, `VALIDATE` and `NONE`.
-  The property controls the behavior of the Hibernate DDL generation. `UPDATE` performs an update of CorDapp schemas, while
-  `VALID` only verifies their integrity.  The property does not affect the node-specific DDL handling and
-   complements `database.initialiseSchema` to disable DDL handling altogether.
+* New configuration property ``database.initialiseAppSchema`` with values ``UPDATE``, ``VALIDATE`` and ``NONE``.
+  The property controls the behavior of the Hibernate DDL generation. ``UPDATE`` performs an update of CorDapp schemas, while
+  ``VALID`` only verifies their integrity.  The property does not affect the node-specific DDL handling and
```
gendal: VALIDATE?

docs/source/changelog.rst
```diff
@@ -345,1512 +319,25 @@ Version 4.0
...
-* Dependencies changes:
-    * Upgraded Kotlin to v1.1.1.
-    * Upgraded Gradle to v3.4.1.
-    * Upgraded requery to v1.2.1.
-    * Upgraded H2 to v1.4.194.
-    * Replaced kotlinx-support-jdk8 with kotlin-stdlib-jre8.
-
-* Improvements:
-    * Added ``--version`` command line flag to print the version of the node.
-    * Flows writt
```
gendal: Do we need to add something about BOTH Jars needing to be in the cordapps folder, even if you think you only need the contracts, owing to the database migration magic being in the workflows jar?  (I seem to recall a discussion about this yesterday). cc @szymonsztuka I think?

gendal: Now as opposed to when? How about: "Changes to Corda's handling of state relevancy means that additional information needs to be stored in the vault for each unspent state. This information was not captured by previous versions so if a Corda 4 node is started with historic data in its vault it will perform a one-time analysis of existing unspent states to calculate this information. This means Corda can be....

docs/source/getting-set-up.rst
```diff
@@ -5,7 +5,7 @@ Software requirements
 ---------------------
 Corda uses industry-standard tools:
 
-* **Oracle JDK 8 JVM** - minimum supported version **8u171**
+* **Oracle JDK 8 JVM** - minimum supported version **|java_version|** (please note that
```
gendal: we need to talk about the other supported jvms I guess?    Gavin sent this to corda-dev: https://groups.io/g/corda-dev/message/624?p=,,,20,0,0,0::Created,,zulu,20,2,0,28742916  Do we add something like: "We have also tested with Azul Zulu OpenJDK 8, Amazon Corretto 8 and Red Hat OpenJDK 8 at the same or higher patch level equivalents". Something like that?

docs/source/key-concepts-node.rst
```diff
@@ -81,16 +81,20 @@ The node also has several CorDapps installed by default to handle common tasks s
...
 
 * Commands requiring to start new flows through RPC will be rejected.
 * Scheduled flows due will be ignored.
 * Initial P2P session messages will not be processed, meaning peers will not be able to initiate new flows involving the node.
 * All other activities will proceed as usual, ensuring that the number of in-flight flows will strictly diminish.
 
 As their number - which can be monitored through RPC - reaches zero, it is safe to shut the node down.
-This property is durable, meaning that restarting the node will not reset it to its default value and that a RPC command is required.
\ No newline at end of file
+This property is durable, meaning that restarting the node will not reset it to its default value and that a RPC command is required.
```
gendal: is this tested I wonder?

docs/source/node-upgrade-notes.rst
```diff
@@ -0,0 +1,63 @@
...
+Draining flows is a key task for node administrators to perform. It exists to simplify applications by ensuring apps don't have to be
+able to migrate workflows from any arbitrary point to other arbitrary points, a task that would rapidly become infeasible as workflow
+and protocol complexity increases.
+
+To drain the node, run the ``gracefulShutdown`` command. This will wait for the node to drain and then shut down the node when the drain
+is complete.
+
+.. warning:: The length of time a node takes to drain depends on both how your applications are designed, and whether any apps are currently
+   talking to network peers that are offline or slow to respond. It is thus hard to give guidance on how long a drain should take, but in
+   an environment with well written apps and in which your counterparties are online, drains may need only a few seconds.
```
gendal: Any guidance on what to do if they can't drain for any reason or it takes forever?  What if they have old flows in their database that are never going to end?

docs/source/key-concepts-node.rst
mikehearn: IIRC it was, when implemented. I plan to experiment with this procedure some more as part of my next round of testing. For now I'm assuming that what's documented does work.

docs/source/node-upgrade-notes.rst
mikehearn: That would require introducing killFlow but it's a separate topic and needs some usability work in general, I think. But yes, these docs can be enhanced over time with various scrapes and pickles you might find yourself in.

mikehearn:
@gendal ptal

PR #4796, opened 2019-02-21
CORDA-2654: Repair the DJVM CLI tool.

chrisr3:
Realign the DJVM CLI with the more recent DJVM work so that it can build and run classes based on `java.util.function.Function` rather than its own `SandboxedRunnable`.


docs/source/key-concepts.rst
```diff
@@ -24,7 +24,6 @@ This section should be read in order:
    key-concepts-node
    key-concepts-tearoffs
    key-concepts-tradeoffs
-   key-concepts-djvm
```
tlil: Why is this pulled out?

chrisr3: Because I thought we were pulling the DJVM from the docs?

tlil: That was discussing the _release notes_, not this section. And I believe we've gone back on that now, with this PR.

chrisr3:
Actually, I think the line about "logging an informational message" no longer applies. This was previously implemented via a `Rule` but I don't seem able to trigger it any more.

PR #4806, opened 2019-02-23
CORDA-2669 - Reintroduce pendingFlowsCount

tlil:
Reintroduce `pendingFlowsCount` to public API (as deprecated). Advise to use the `gracefulShutdown` command in the shell instead.

The draining API should be revisited and tidied up for future releases.

See [CORDA-2669](https://r3-cev.atlassian.net/browse/CORDA-2669) for further details.


mikehearn:
Oh, we probably want to add it to the API checker tool by hand, so we can't break it again.

tlil:
Yes, good point. I'll generate and add that.

PR #4811, opened 2019-02-25
CORDA-2671: Remove broken ExplorerSimulator from Node Explorer.

chrisr3:
Node Explorer's `ExplorerSimulator` uses `:node-driver` to launch a cluster of Corda nodes. However, `:node-driver` now assumes that it is running from within a Gradle project so that it can deploy CorDapps. This breaks Node Explorer, which is a stand-alone application these days.


tools/explorer/README.md
```diff
@@ -13,37 +13,6 @@ The user can execute cash transaction commands to issue and move cash to other p
 
     ./gradlew tools:explorer:run
     
-
-## Running Demo Nodes
```
josecoll: Perhaps we should suggest some Corda Samples (eg. Bank of Corda) to run the Explorer against instead.

josecoll:
You also need to update the documentation that talks about using `explorer:runDemoNodes``
See node-explorer.rst.

chrisr3:
> Sad to see the old Explorer Simulator go ...

Boo-hoo. And I'm over it already :wink:.

PR #4814, opened 2019-02-25
ENT-3128: Tidy up DemoBench's "built-in CorDapp" management.

chrisr3:
Allow for more "built-in" CorDapps to be added to DemoBench at some point.


chrisr3:
> OK - the renaming away from `finance` makes it a little cleaner I suppose.
I am pro-actively reducing the difference between OS and ENT... :wink:

PR #4815, opened 2019-02-25
CORDA-2672: Tidy up CorDapp deployments in samples.

chrisr3:
- Extract CorDapp clients from their CorDapps.
- Fix warnings from `cordapp` plugin.
- Stop applying unnecessary Gradle plugins.


samples/cordapp-configuration/build.gradle
```diff
@@ -3,7 +3,7 @@ apply plugin: 'idea'
 apply plugin: 'net.corda.plugins.cordformation'
 
 dependencies {
-    compile project(':node-api')
+    runtimeOnly project(':node-api')
```
josecoll: I think we should update our docs to include this recommendation:
https://docs.corda.net/head/cordapp-build-systems.html#corda-dependencies

samples/irs-demo/cordapp/build.gradle
```diff
@@ -17,14 +17,22 @@ sourceSets {
...
+    }
+}
+
 dependencies {
     cordapp project(':finance:contracts')
     cordapp project(':finance:workflows')
 
     // Corda integration dependencies
     cordaRuntime project(path: ":node:capsule", configuration: 'runtimeArtifacts')
-    cordaCompile project(':core')
```
josecoll: Again, we need to update our CorDapp Dependencies Guidelines on when/why/how to add these two dependencies that you have just removed:
`cordaCompile project(':core')`
`cordaCompile project(':node-api')`

samples/notary-demo/build.gradle
```diff
@@ -1,17 +1,25 @@
...
+        name "Corda Notary Demo"
+        vendor "R3"
+        targetPlatformVersion corda_platform_version.toInteger()
+        minimumPlatformVersion 1
+    }
+}
+
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+    cordaCompile project(':client:rpc')
```
josecoll: In general, shouldn't these samples be using non-project internal dependency references:
eg. `cordaCompile net.corda:corda-rpc:$corda_release_version`

samples/notary-demo/workflows/build.gradle
```diff
@@ -6,7 +6,11 @@ description 'Corda Notary Demo - Workflows'
 dependencies {
     cordaCompile project(':core')
     cordaCompile project(':client:rpc')
-    cordaCompile project(':node')
+
+    // We need to compile against the Node, but also DO NOT
+    // want the Node bundled inside the CorDapp or added to
+    // Gradle's runtime classpath.
+    compileOnly project(':node')
```
josecoll: I thought that was the reason we used `cordaCompile` ???
"cordaCompile and cordaRuntime indicate dependencies that should not be included in the CorDapp JAR."

Ok ... I see, this change means the `Node` is not also bundled on the runtime classpath.

Is this a specialist sample case or should we also add this as a guideline in the docs?

samples/simm-valuation-demo/build.gradle
```diff
@@ -62,6 +62,11 @@ dependencies {
     testCompile "org.assertj:assertj-core:$assertj_version"
 }
 
+jar {
+    // The Node configures the logging, not the CorDapp!
```
josecoll: so why define a CorDapp `log4j2.xml` then?

samples/simm-valuation-demo/src/main/resources/log4j2.xml
```diff
@@ -2,8 +2,8 @@
 <Configuration status="info">
```
josecoll: you seem to be excluding this file in the associated `build.gradle`???

samples/trader-demo/build.gradle
```diff
@@ -1,29 +1,58 @@
...
+}
+
+configurations {
+    integrationTestCompile.extendsFrom testCompile
+    integrationTestRuntime.extendsFrom testRuntime
+}
+
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+    cordaCompile project(':samples:bank-of-corda-demo')
```
josecoll: This sample should depend on Finance, *not* bank-of-corda-demo !!!

samples/notary-demo/build.gradle
chrisr3: For as long as `:samples` is a module within Corda itself we _definitely_ want to use `project(':client:rpc')` here. We don't want two sources of the exact same classes and are testing the samples against the Corda components that we have just built.

samples/notary-demo/workflows/build.gradle
chrisr3: I'm trying to keep another SLF4J back-end off the runtime classpath so that the Network Bootstrapper can't find it.

samples/simm-valuation-demo/build.gradle
chrisr3: CorDapps shouldn't have `log4j2.xml` files because logging should be configured by the Node itself.

Any `log4j2.xml` files you see are actually for (e.g.) `:node-driver` to use. Or other testing apparatus.

josecoll: OK - thanks. 
All rather confusing ... but correct in light of lack of isolation between CorDapps and the Node they are executing within.

samples/trader-demo/build.gradle
```diff
@@ -30,7 +30,13 @@ configurations {
 
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
-    cordaCompile project(':samples:bank-of-corda-demo')
+    compile "net.sf.jopt-simple:jopt-simple:$jopt_simple_version"
+    cordaCompile project(':client:rpc')
+    runtimeOnly "org.slf4j:slf4j-simple:$slf4j_version"
+
+    // We only need this for its DUMMY_BANK constants, and
```
josecoll: What!?

chrisr3: Yes, I've broken the "Bank of Corda" dependency down, and the `DUMMY_BANK` constants can only be obtained via `:test-utils`. This is already on the `testRuntime` classpath as a transitive dependency of `:node-driver` but I don't want `node-driver` on the compile classpath.

samples/trader-demo/build.gradle
chrisr3: :+1:

samples/irs-demo/cordapp/build.gradle
chrisr3: This module isn't actually building a CorDapp, although it needs to apply the `cordapp` plugin for the sake of CordFormation :nauseated_face:.

samples/simm-valuation-demo/src/main/resources/log4j2.xml
chrisr3: CorDapps do not tell Nodes how to configure Log4J. These `log4j2.xml` files are intended for `:node-driver` and possibly Network Bootrapper (i.e. Cordformation).

docs/source/cordapp-build-systems.rst
```diff
@@ -88,15 +89,17 @@ Here are some guidelines for Corda dependencies:
...
 
 * When building an RPC client that communicates with a node (e.g. a webserver), you should include
-  ``net.corda:corda-rpc:$corda_release_version`` as a ``cordaCompile`` dependency
+  ``net.corda:corda-rpc:$corda_release_version`` as a ``cordaCompile`` dependency.
 
 * When you need to use the network bootstrapper to bootstrap a local network (e.g. when using ``Cordformation``), you
-  should include ``net.corda:corda-node-api:$corda_release_version`` as a ``cordaCompile`` dependency
+  should include ``net.corda:corda-node-api:$corda_release_version`` as either a ``cordaRuntime`` or a ``runtimeOnly``
+  dependency. You may also wish to include an implementation of SLF4J as a ``runtimeOnly`` dependency for the network
+  bootstrapper to use.
```
josecoll: qualify with "if using CordFormation to deploy a local network" 
(eg. its perfectly feasible to build CorDapps without all the extra baggage)

docs/source/cordapp-build-systems.rst
```diff
@@ -73,10 +73,11 @@ In certain cases, you may also wish to build against the unstable Master branch.
 
 Corda dependencies
 ^^^^^^^^^^^^^^^^^^
-The ``cordformation`` plugin adds two new gradle configurations:
+The ``cordapp`` plugin adds three new gradle configurations:
 
 * ``cordaCompile``, which extends ``compile``
 * ``cordaRuntime``, which extends ``runtime``
+* ``cordapp``, which extends ``compile``
```
josecoll: needs a one liner explanation of when this should be used (eg. how does it differ from `cordaCompile`?)

samples/attachment-demo/build.gradle
```diff
@@ -1,17 +1,60 @@
...
+    }
+}
+
+configurations {
+    integrationTestCompile.extendsFrom testCompile
+    integrationTestRuntime.extendsFrom testRuntime
+}
+
 dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
```
josecoll: where did all these extra dependencies come from ? (eg. how did this work before?)

samples/attachment-demo/src/main/resources/simplelogger.properties
```diff
@@ -0,0 +1,3 @@
+org.slf4j.simpleLogger.defaultLogLevel=info
```
josecoll: I think adding another type of logger (and configuration syntax) is just plain confusing.

chrisr3: Perhaps, but the classpath has been polluted with jars containing other `log4j2.xml` and `log4j2-test.xml` files.  _This means that I cannot guarantee that `:node-driver` / Network Bootstrapper will use **my** configuration file!_ The simplest solution is to use a completely different SLF4J backend instead.

I could always replace `slf4j-simple` with `slf4j-nop`, of course. That doesn't need any configuration.

samples/attachment-demo/build.gradle
chrisr3: I smited this line from `workflows/build.gradle`:
```gradle
cordaCompile project(':webserver')
```
The point being that this is a Web _client_, and so it doesn't need all of the Web server's sneaky transitive dependencies being added to its runtime classpath. So "Goodbye, Jetty! You have been replaced with just the RPC client, a command line parser and some Web annotations."

docs/source/cordapp-build-systems.rst
chrisr3: Eh? That paragraph _starts_ by saying:
> When you need to use the network bootstrapper to bootstrap a local network ...

and then I say that the SLF4J implementation would be "for the network bootstrapper to use". So there's no need for this paragraph to repeat itself.

josecoll: Apologies - I didn't start reading from the beginning of the paragraph.

samples/attachment-demo/src/main/resources/simplelogger.properties
josecoll: But until we properly deprecate the :webserver from Corda (eg. remove it), why would we want to make the samples dependencies look more complex?

chrisr3: This isn't about WebServer. We have `log4j2.xml` files baked into _both_ `corda.jar` _and_ `corda-node.jar`, and a `log4j2-test.xml` file baked into `corda-test-common.jar` too (for bad measure).

Log4J2's strategy for finding its configuration is described [here](https://logging.apache.org/log4j/2.x/manual/configuration.html), which would seem to suggest that I could probably override all of these XML files by providing a `log4j2-test.properties` file, except that would _definitely_ be more complicated that my three line `simplelogger.properties` file.

docs/source/cordapp-build-systems.rst
chrisr3: :+1:

chrisr3:
@josecoll I've rebased this PR against the `master` branch and confirmed that these samples all still work. Can we merge this one now, please?

josecoll:
Sure - as soon as the latest Integration TC gate passes ;-)

PR #4816, opened 2019-02-26
ENT-3059: RPC disconnect log entry changed to info

lockathan:
When an RPC client disconnects from the RPC server, the log entry created now has its log level set to info rather than warning. It is perfectly normal for an RPC client to disconnect - only the RPC client knows if it was intentional.


mikehearn:
Maybe it could even have been trace!

lockathan:
Should I change it from info to trace?

PR #4818, opened 2019-02-26
ENT-3165 Add unit test for reference state consumption.

rick-r3:
Adds a test that creates state on node 0 and does not share.  Then creates a transaction referring to it that is shared with node 1 (relevant only record).  Then evolve the reference state on node 0 and do not share it.   Then creates a transaction referring to it that is shared with node 1 (relevant only record).

That last transaction should then update the states recorded for the reference states transaction to include the second version of the reference state (since it wasn't originally relevant), thus consuming the original version.

PR coming in enterprise (which should be merged first to avoid breaking merge PR).


rick-r3:
Waiting for ENT PR to merge before merging to avoid a broken OS -> ENT merge.

PR #3789, opened 2018-08-15
* Added "isRelevant" functionality to the vault.

roger3cev:
* Added new Vault Query API: `StateModificationStatus` which ca be `MODIFIABLE`, `NOT_MODIFIABLE` or `ALL`
* Updated "StateMetaData"
* All the query criteria now take `StateModificationStatus` as a new parameter
* Default behaviour is to query for `ALL` states as that is what Corda has previously done
* Added new column to `VaultStates` table called `is_modifiable` which stores the value of the enum which can only be set to RELEVANT or NOT_RELEVANT, never ALL
* Updated `getCashBalances` to only query for `MODIFIABLE` states as we only want to count cash states which we own!
* Added tests

Probably needs some optimisation as we now perform multiple "isModifiable" checks (key lookups).

TODO: Add better docs.

PR #4819, opened 2019-02-26
CORDA-2665: Updated OwnableState relevancy check put back to V3 version

shamsasari:
https://github.com/corda/corda/pull/3789 changed the relevancy check of OwnableState to include the participants list in addition to the owner. This however breaks existing apps which assume (in their vault query) an OwnableState is recorded to the vault if-and-only-if the owner matches.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
+++ b/node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
@@ -6,6 +6,7 @@ import net.corda.core.contracts.*
 import net.corda.core.crypto.SecureHash
 import net.corda.core.crypto.containsAny
 import net.corda.core.internal.*
+import net.corda.core.internal.cordapp.CordappResolver
 import net.corda.core.messaging.DataFeed
 import net.corda.core.node.ServicesForResolution
 import net.corda.core.node.StatesToRecord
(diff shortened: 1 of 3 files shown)


node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -69,11 +70,19 @@ class NodeVaultService(
          * A state is relevant if any of the participants (or the owner for ownable states) has an owning key matching one of this node's
          * public keys.
          */
+        // TODO This could benefit from a small performance tweak by not using temporary collections
```
fenryka: ToDo's should have Jira's or they tend to just die on the vine

node/src/main/kotlin/net/corda/node/services/vault/NodeVaultService.kt
```diff
@@ -69,11 +70,19 @@ class NodeVaultService(
...
         fun isRelevant(state: ContractState, myKeys: Set<PublicKey>): Boolean {
             val keysToCheck = when (state) {
                 // Sometimes developers forget to add the owning key to participants for OwnableStates.
                 // TODO: This logic should probably be moved to OwnableState so we can just do a simple intersection here.
-                is OwnableState -> (state.participants.map { it.owningKey } + state.owner.owningKey).toSet()
+                is OwnableState -> {
```
rick-r3: I'm generally negative on the idea of switching behaviour under users.  It's a tall order to re-verify the behaviour of an existing app that just happens to be re-compiled on 4 given how much testing app writers seem to be doing.  I think I'd like to understand why we want to make this change to the relevancy a bit more.

PR #4820, opened 2019-02-26
CORDA-2674 observe the true type of an atomic type

r3domfox:
During evolution, the evolution serializer factory compares the types of properties on the remote (serialized, described by AMQP schema) and local (inspected by reflecting over the type on the receiver's classpath) type, and throws an exception if there is a mismatch.

For purposes of comparison, it uses the `observedType` of the local property, which should reflect the actual type of the field/getter/setter/constructor parameter. However, if the property's type is a boxed primitive such as `java.lang.Integer`, the `observedType` is mistakenly recorded as the unboxed primitive type. This causes an exception when trying to deserialize a type that requires evolution, but has properties of boxed primitive types.

We resolve this problem by modifying the `LocalTypeInformationBuilder` so that it faithfully records the _actual_ type of the observed property, rather than forcing it to the unboxed primitive type.


r3domfox:
Done!

serialization/src/test/java/net/corda/serialization/internal/amqp/JavaEvolutionTests.java
```diff
@@ -97,4 +97,53 @@ public void testN2AddsPrimitive() throws IOException {
...
+
+        public void setId(Integer id) {
+            this.id = id;
+        }
+    }
+    */
+
+    public interface ForceEvolution { }
+
+    // Class as it exists now with the newly added element
```
mikehearn: What's the newly added element? This looks identical to the class above above?

r3domfox: The marker interface. It changes the fingerprint, so we hit the evolution serializer factory.

Why not just add a new property? Because the "property types must match" rule is only actually checked if the set of property names is the same - we're looking to see if this is a spurious evolution case and we can get away with just returning the paleo-serializer.

We probably _should_ also check that property types match for properties that the old and new type have in common, but it turns out that we don't at the moment (the first version of this test added a new property, and mysteriously passed even without the fix in pace). Task for another day I think.

mikehearn: Good explanation but could you put it in the code? I was pretty confused by what I was reading because I didn't spot that the new element was the interface.

r3domfox: Done.

mikehearn:
Approved, thanks!

PR #4826, opened 2019-02-27
[CORDA-2687] Ensure collections of unknown types are not carpented

JamesHR3:
During schema processing when attempting to deserialize, the serialization engine will detect when a class is not available on the classpath and attempt to use the carpenter to patch this up. In the event that the missing class was found as a type parameter of a collection, the type it attempts to create is the collection of the missing type, which results in an error (the type name is not a valid java name). This change attempts to detect this condition and not carpent up the class in this case.

Update: Added a unit test and verified that it fails without the fix with the error seen.


node-api/src/main/kotlin/net/corda/nodeapi/internal/serialization/amqp/SerializerFactory.kt
```diff
@@ -249,8 +249,12 @@ open class SerializerFactory(
                 name to processSchemaEntry(notation)
             } catch (e: ClassNotFoundException) {
                 // class missing from the classpath, so load its interfaces and add it for carpenting (see method docs).
```
dimosr: Could you keep this comment inside the if-condition ?
Because currently there are 2 different comments intermingled and it's not very clear where each one refers to.

Also, (minor) but I remember I added a unit test for the interfaces case, could you add one more for this case ? It was in the `StaticInitialisationOfSerializedObjectTest.kt`.

JamesHR3: Thanks @dimosr . I've left the comment together as I think the two statements are related ("we're carpenting some stuff, here is what we're carpenting and how we're working out what that is"), but I've tried to clarify a little. I've also added a unit test.

PR #4837, opened 2019-03-01
CORDA-2684 Added new guide on CorDapp Constraints Migration procedures.

josecoll:
Documentation explaining how to consume existing hash and CZ whitelisted states using Corda 4.

### DIFF
--- a/docs/source/api-contract-constraints.rst
+++ b/docs/source/api-contract-constraints.rst
@@ -146,7 +146,7 @@ Hash constrained states in private networks
 Where private networks started life using CorDapps with hash constrained states, we have introduced a mechanism to relax the checking of
 these hash constrained states when upgrading to signed CorDapps using signature constraints.
 
-The java system property ``-Dnet.corda.node.disableHashConstraints="true"`` may be set to relax the hash constraint checking behaviour.
+The Java system property ``-Dnet.corda.node.disableHashConstraints="true"`` may be set to relax the hash constraint checking behaviour.
 
 This mode should only be used upon "out of band" agreement by all participants in a network.
 
--- a/docs/source/api-contract-constraints.rst
+++ b/docs/source/api-contract-constraints.rst
@@ -46,6 +46,8 @@ to issue the states was signed by Alice and Bob, every transaction must use an a
 the constraint used by equivalent output states (i.e. output states that use the same contract class name) must match the
 input state, so it can't be changed and you can't combine states with incompatible constraints together in the same transaction.
 
+.. _implicit_vs_explicit_upgrades:
+
 **Implicit vs explicit.** Constraints are not the only way to manage upgrades to transactions. There are two ways of handling
 upgrades to a smart contract in Corda:
 
@@ -136,15 +138,15 @@ From there it's suspended waiting to be retried on node restart.
 This gives the node operator the opportunity to recover from those errors, which in the case of constraint violations means
 adding the right cordapp jar to the ``cordapps`` folder.
 
+.. _relax_hash_constraints_checking_ref:
+
 Hash constrained states in private networks
 -------------------------------------------
 
 Where private networks started life using CorDapps with hash constrained states, we have introduced a mechanism to relax the checking of
 these hash constrained states when upgrading to signed CorDapps using signature constraints.
 
-The following java system property may be set to relax the hash constraint checking behaviour:
-
-   -Dnet.corda.node.disableHashConstraints="true"
+The java system property ``-Dnet.corda.node.disableHashConstraints="true"`` may be set to relax the hash constraint checking behaviour.
 
 This mode should only be used upon "out of band" agreement by all participants in a network.
 
@@ -191,58 +193,11 @@ During transaction building the ``AutomaticPlaceholderConstraint`` for output st
 will be selected based on a variety of factors so that the above holds true. If it can't find attachments in storage or there are no
 possible constraints, the ``TransactionBuilder`` will throw an exception.
 
+CorDapp constraints migration
+------------------------------
 
-.. _constraints_whitelist_to_signature_ref:
-
-How to use the ``SignatureAttachmentConstraint`` if states were already created on the network with the ``WhitelistedByZoneAttachmentConstraint``
--------------------------------------------------------------------------------------------------------------------------------------------------
-
-1. As the original developer of the corDapp, the first step is to sign the latest version of the JAR that was released (see :doc:`cordapp-build-systems`).
-The key used for signing will be used to sign all subsequent releases, so it should be stored appropriately. The JAR can be signed by multiple keys owned
-by different parties and it will be expressed as a ``CompositeKey`` in the ``SignatureAttachmentConstraint`` (See :doc:`api-core-types`).
-Use `JAR signing and verification tool <https://docs.oracle.com/javase/tutorial/deployment/jar/verify.html>`_ to sign the existing JAR.
-The signing capability of :ref:`corda-gradle-plugins <cordapp_build_system_signing_cordapp_jar_ref>` cannot be used in this context as it signs the JAR while building it from source.
-
-2. Whitelist this newly signed JAR with the Zone operator. The Zone operator should check that the JAR is signed and not allow any
-more versions of it to be whitelisted in the future. From now on the developer(s) who signed the JAR are responsible for new versions.
-
-3. Any flows that build transactions using this Cordapp will have the responsibility of transitioning states to the ``SignatureAttachmentConstraint``.
-This is done explicitly in the code by setting the constraint of the output states to signers of the latest version of the whitelisted jar.
-In the near future we will make this transition automatic if we detect that the previous 2 steps were executed.
-
-4. As a node operator you need to add the new signed version of the contracts cordapp to the "cordapps" folder together with the latest version of the flows jar
-that will contain code like:
-
-.. container:: codeset
-
-    .. sourcecode:: kotlin
-
-        // This will read the signers for the deployed cordapp.
-        val attachment = this.serviceHub.cordappProvider.getContractAttachmentID(contractClass)
-        val signers = this.serviceHub.attachments.openAttachment(attachment!!)!!.signerKeys
-
-        // Create the key that will have to pass for all future versions.
-        val ownersKey = signers.first()
-
-        val txBuilder = TransactionBuilder(notary)
-                // Set the Signature constraint on the new state to migrate away from the WhitelistConstraint.
-                .addOutputState(outputState, constraint = SignatureAttachmentConstraint(ownersKey))
-                ...
-
-    .. sourcecode:: java
-
-        // This will read the signers for the deployed cordapp.
-        SecureHash attachment = this.getServiceHub().getCordappProvider().getContractAttachmentID(contractClass);
-        List<PublicKey> signers = this.getServiceHub().getAttachments().openAttachment(attachment).getSignerKeys();
-
-        // Create the key that will have to pass for all future versions.
-        PublicKey ownersKey = signers.get(0);
-
-        TransactionBuilder txBuilder = new TransactionBuilder(notary)
-                // Set the Signature constraint on the new state to migrate away from the WhitelistConstraint.
-                .addOutputState(outputState, myContract, new SignatureAttachmentConstraint(ownersKey))
-                ...
-
+Please read :doc:`cordapp-constraint-migration` to understand how to consume pre-existing hash or CZ whitelisted constrained states
+issued using a previous version of Corda using Corda 4 signed CorDapps (using signature constraints).
 
 Debugging
 ---------
@@ -278,7 +233,7 @@ The same example in Java:
         });
 
 
-Staring a node missing CorDapp(s)
+Starting a node missing CorDapp(s)
 *********************************
 
 When running the Corda node ensure all CordDapp JARs are placed in ``cordapps`` directory of each node.
--- a/docs/source/api-contract-constraints.rst
+++ b/docs/source/api-contract-constraints.rst
@@ -193,11 +193,11 @@ During transaction building the ``AutomaticPlaceholderConstraint`` for output st
 will be selected based on a variety of factors so that the above holds true. If it can't find attachments in storage or there are no
 possible constraints, the ``TransactionBuilder`` will throw an exception.
 
-CorDapp constraints migration
-------------------------------
+Constraints migration to Corda 4
+--------------------------------
 
-Please read :doc:`cordapp-constraint-migration` to understand how to consume pre-existing hash or CZ whitelisted constrained states
-issued using a previous version of Corda using Corda 4 signed CorDapps (using signature constraints).
+Please read :doc:`cordapp-constraint-migration` to understand how to consume and evolve pre-Corda 4 issued hash or CZ whitelisted constrained states
+using a Corda 4 signed CorDapp (using signature constraints).
 
 Debugging
 ---------
--- a/docs/source/api-contract-constraints.rst
+++ b/docs/source/api-contract-constraints.rst
@@ -24,7 +24,7 @@ to it.
 There are several types of constraint:
 
 1. Hash constraint: exactly one version of the app can be used with this state.
-2. Zone whitelist constraint: the compatibility zone operator lists the hashes of the versions that can be used with this contract class name.
+2. Compatibility zone whitelisted (or CZ whitelisted) constraint: the compatibility zone operator lists the hashes of the versions that can be used with this contract class name.
 3. Signature constraint: any version of the app signed by the given ``CompositeKey`` can be used.
 4. Always accept constraint: any app can be used at all. This is insecure but convenient for testing.
 
--- a/docs/source/cordapp-constraint-migration.rst
+++ b/docs/source/cordapp-constraint-migration.rst
@@ -20,8 +20,21 @@ Faced with the exercise of upgrading an existing Corda 3.x CorDapp to Corda 4, y
 
    If you have existing **CZ whitelisted** constrained states see :ref:`Migrating CZ whitelisted constraints<cz_whitelisted_constraint_migration>`.
 
-   If you have existing **always accept** constrained states these are automatically consumable as they offer no security and should only
+   If you have existing **always accept** constrained states these are not consumable nor evolvable as they offer no security and should only
    be used in test environments.
+
+* What type of contract states does my CorDapp use?
+
+   **Linear states** typically evolve over an extended period of time (defined by the lifecycle of the associated business use case), and
+   thus are prime candidates for constraints migration.
+
+   **Fungible states** are created by an issuer and transferred around a Corda network until explicitly exited (by the same issuer).
+   They do not evolve as linear states, but are transferred between participants on a network. Their consumption may produce additional new
+   output states to represent adjustments to the original state (e.g. change when spending cash). For the purposes of constraints migration,
+   it is desirable that any new output states are produced using the new Corda 4 signature constraint types.
+
+   Where you have long transaction chains of fungible states, it may be advisable to send them back to the issuer for re-issuance (this is
+   called "chain snipping" and has performance advantages as well as simplifying constraints type migration).
 
 * Should I use the **implicit** or **explicit** upgrade path?
 
@@ -29,6 +42,14 @@ Faced with the exercise of upgrading an existing Corda 3.x CorDapp to Corda 4, y
 
    **Implicit** upgrades allow pre-authorising multiple implementations of the contract ahead of time.
    They do not require additional coding and do not incur a complex choreographed operational upgrade process.
+
+.. warning:: the steps outlined in this page assume you are using the same CorDapp Contract (eg. same state definition, commands and verification code) and
+   wish to use that CorDapp to leverage the upgradeability benefits of Corda 4 signature constraints. If you are looking to upgrade an existing
+   Contract CorDapp please read :ref:`Contract and state versioning<contract_upgrading_ref>` and :doc:`cordapp-upgradeability` to understand your options.
+
+Please also remember that *states are always consumable if the version of the CorDapp that issued (created) them is installed*.
+In the simplest of scenarios it may be easier to re-issue existing hash or CZ whitelisted states (eg. exit them from the ledger using
+the original unsigned CorDapp and re-issuing them using the new signed CorDapp).
 
 .. _hash_constraint_migration:
 
@@ -40,9 +61,10 @@ Hash constraints migration
 Corda 4.0
 ~~~~~~~~~
 
-Corda 4.0 requires some additional steps to consume pre-existing on-ledger **hash** constrained states:
+Corda 4.0 requires some additional steps to consume and evolve pre-existing on-ledger **hash** constrained states:
 
-1. The Corda Node must be started using relaxed hash constraint checking mode as described in :ref:`relax_hash_constraints_checking_ref`.
+1. All Corda Nodes in the same CZ or business network that may encounter a transaction chain with a hash constrained state must be started using
+   relaxed hash constraint checking mode as described in :ref:`relax_hash_constraints_checking_ref`.
 
 2. CorDapp flows that build transactions using pre-existing *hash-constrained* states must explicitly set output states to use *signature constraints*
    and specify the related public key(s) used in signing the associated CorDapp Contract JAR:
@@ -84,7 +106,8 @@ Later releases
 ~~~~~~~~~~~~~~
 
 The next version of Corda will provide automatic migration of *hash constrained* states. This means that signed CorDapps running on a Corda 4.x node will
-automatically consume any pre-existing on-ledger *hash-constrained* states (and generate *signature-constrained* outputs).
+automatically consume any pre-existing on-ledger *hash-constrained* states (and generate *signature-constrained* outputs) when the system property
+to break constraints is set.
 
 .. _cz_whitelisted_constraint_migration:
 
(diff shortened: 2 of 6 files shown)


docs/source/api-contract-constraints.rst
```diff
@@ -136,15 +138,15 @@ From there it's suspended waiting to be retried on node restart.
...
 Hash constrained states in private networks
 -------------------------------------------
 
 Where private networks started life using CorDapps with hash constrained states, we have introduced a mechanism to relax the checking of
 these hash constrained states when upgrading to signed CorDapps using signature constraints.
 
-The following java system property may be set to relax the hash constraint checking behaviour:
-
-   -Dnet.corda.node.disableHashConstraints="true"
+The java system property ``-Dnet.corda.node.disableHashConstraints="true"`` may be set to relax the hash constraint checking behaviour.
```
gendal: Java - capital J

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+   <script type="text/javascript" src="_static/codesets.js"></script>
+
+CorDapp constraints migration
+=============================
+
+.. note:: Before reading this page, you should be familiar with the key concepts of :doc:`Contract Constraints <api-contract-constraints>`.
+
+Faced with the exercise of upgrading an existing Corda 3.x corDapp to Corda 4, you need to consider the following:
+
+* What existing unconsumed states have been issued on ledger by a previous version of this corDapp and using other constraint types?
```
gendal: CorDapp (capital C)

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+   **Implicit** upgrades allow pre-authorising multiple implementations of the contract ahead of time.
+   They do not require additional coding and do not incur a complex choreographed operational upgrade process.
+
+.. _hash_constraint_migration:
+
+Hash constraints migration
+--------------------------
+
+.. note:: These instructions only apply to corDapp Contract JARs (unless otherwise stated).
```
gendal: CorDapp

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Corda 4.0
+~~~~~~~~~
+
+Corda 4.0 requires some additional steps to consume pre-existing on-ledger **hash** constrained states:
+
+1. The Corda Node must be started using relaxed hash constraint checking mode as described in :ref:`relax_hash_constraints_checking_ref`.
+
+2. corDapp flows that build transactions using pre-existing *hash-constrained* states must explicitly set output states to use *signature constraints*
```
gendal: CorDapp (I won't comment on this any more - but all other instances should be changed too)

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+   and specify the related public key(s) used in signing the associated corDapp Contract JAR:
+
+.. container:: codeset
+
+    .. sourcecode:: kotlin
+
+        // This will read the signers for the deployed corDapp.
+        val attachment = this.serviceHub.cordappProvider.getContractAttachmentID(contractClass)
+        val signers = this.serviceHub.attachments.openAttachment(attachment!!)!!.signerKeys
```
gendal: why does the java snippet show using Transaction Builder but the Kotlin sample doesn't?

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+        List<PublicKey> signers = this.getServiceHub().getAttachments().openAttachment(attachment).getSignerKeys();
+
+        // Create the key that will have to pass for all future versions.
+        PublicKey ownersKey = signers.get(0);
+
+        TransactionBuilder txBuilder = new TransactionBuilder(notary)
+                // Set the Signature constraint on the new state to migrate away from the WhitelistConstraint.
+                .addOutputState(outputState, myContract, new SignatureAttachmentConstraint(ownersKey))
+
+3. Both the original pre-Corda 4 corDapp JAR (generating *hash-constrained* states) and the new Corda 4 signed corDapp JAR must be present in the
```
gendal: This is interesting... feels like there's some additional detail/nuance behind this that we should expand on, if only to help the reader understand what's going on and how they need to write their contracts? The thoughts going through my mind are:

* Ahh... that's interesting... it hadn't occurred to me that the contracts would be different... I just assumed you'd attach the same Jar as before (so the tx is effectively a no op from a business perspective but now you have a sig constrained output that can be upgraded in a future tx).  But it seems we're transitioning the constraint AND upgrading the contract at the same time?  Is that because the no-op can't work because the original Jar is unsigned?
* Assuming I've understood this correctly, do we have any recommendations here?  eg would a sensible pattern simply be to create a signed jar that has IDENTICAL functionality as the hash-constrained Jar?  so that the tx that changes the constraint is purely technical and has no business function?
* Either way, perhaps also remind people that it is the NEW Jar that will be attached to the tx and so (obviously - but perhaps not obviously!) it needs to be compatible with the old one since it has to be able to deserialise and reason about the input state?

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+Corda 4.0
+~~~~~~~~~
+
+Corda 4.0 requires some additional steps to consume pre-existing on-ledger **CZ whitelisted** constrained states:
+
+1. As the original developer of the corDapp, the first step is to sign the latest version of the JAR that was released (see :doc:`cordapp-build-systems`).
+   The key used for signing will be used to sign all subsequent releases, so it should be stored appropriately. The JAR can be signed by multiple keys owned
+   by different parties and it will be expressed as a ``CompositeKey`` in the ``SignatureAttachmentConstraint`` (See :doc:`api-core-types`).
+   Use `JAR signing and verification tool <https://docs.oracle.com/javase/tutorial/deployment/jar/verify.html>`_ to sign the existing JAR.
+   The signing capability of :ref:`corda-gradle-plugins <cordapp_build_system_signing_cordapp_jar_ref>` cannot be used in this context as it signs the JAR while building it from source.
```
gendal: ouch... this feels like an exciting potential source of bugs.... ie how things are signed differs depending on whether you start with source or binary?  Or is signing jars utterly standardised and unremarkable?

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+2. Both the original pre-Corda 4 corDapp JAR (generating *CZ whitelisted* constrained states) and the new Corda 4 signed corDapp JAR must be
+   registered with the CZ Network operator (as whitelisted in the network parameters which are distributed to all nodes in that CZ).
+   The CZ network operator should check that the JAR is signed and not allow any more versions of it to be whitelisted in the future.
+   From now on the development organisation that signed the JAR is responsible for signing new versions.
+
+   The process of CZ Network corDapp whitelisting depends on how the Corda Network is configured:
```
gendal: lowercase N for network so it's clear we're talking about Corda networks in general, not just Corda Network in particular.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+    - if using a hosted CZ Network (such as `The Corda Network <https://docs.corda.net/head/corda-network/index.html>`_ or
+      `UAT Environment <https://docs.corda.net/head/corda-network/UAT.html>`_ ) running an Identity Operator (formerly known as Doorman) and
+      Network Map Service, you should manually send the hashes of the two JARs to the CZ network operator and request these be added using
```
gendal: why the two? Wouldn't the first one _already_ be whitelisted?  Since the previous transaction (that created the zone-constrained state wouldn't have been valid if it wasn't?)

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+3. Any flows that build transactions using this corDapp will have the responsibility of transitioning states to the ``SignatureAttachmentConstraint``.
+   This is done explicitly in the code by setting the constraint of the output states to signers of the latest version of the whitelisted jar:
+
+.. container:: codeset
+
+    .. sourcecode:: kotlin
+
+        // This will read the signers for the deployed corDapp.
+        val attachment = this.serviceHub.cordappProvider.getContractAttachmentID(contractClass)
```
gendal: how does this know which contract to use here?  Both the old one and the new one will be hanging around, no?

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+        List<PublicKey> signers = this.getServiceHub().getAttachments().openAttachment(attachment).getSignerKeys();
+
+        // Create the key that will have to pass for all future versions.
+        PublicKey ownersKey = signers.get(0);
+
+        TransactionBuilder txBuilder = new TransactionBuilder(notary)
+                // Set the Signature constraint on the new state to migrate away from the WhitelistConstraint.
+                .addOutputState(outputState, myContract, new SignatureAttachmentConstraint(ownersKey))
+
+4. As a node operator you need to add the new signed version of the contracts corDapp to the "cordapps" folder together with the latest version of the flows jar.
```
gendal: ah... ok - does that answer my question above?  If so, does it mean you have to ensure the OLD app is NOT in the folder?

gendal: and if so, what happens if this is not the case?

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+Hash constraints migration
+--------------------------
+
+.. note:: These instructions only apply to corDapp Contract JARs (unless otherwise stated).
+
+Corda 4.0
+~~~~~~~~~
+
+Corda 4.0 requires some additional steps to consume pre-existing on-ledger **hash** constrained states:
```
mikehearn: Not to consume them, right? They are consumable out of the box. This section only applies if you want to keep them around but "break" the constraint by changing what code is usable with them.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+        // This will read the signers for the deployed corDapp.
+        SecureHash attachment = this.getServiceHub().getCordappProvider().getContractAttachmentID(contractClass);
+        List<PublicKey> signers = this.getServiceHub().getAttachments().openAttachment(attachment).getSignerKeys();
+
+        // Create the key that will have to pass for all future versions.
+        PublicKey ownersKey = signers.get(0);
+
+        TransactionBuilder txBuilder = new TransactionBuilder(notary)
+                // Set the Signature constraint on the new state to migrate away from the WhitelistConstraint.
```
mikehearn: The comment here is talking about WhitelistConstraint but it's in a section about hash constraints.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+                .addOutputState(outputState, myContract, new SignatureAttachmentConstraint(ownersKey))
+
+3. Both the original pre-Corda 4 corDapp JAR (generating *hash-constrained* states) and the new Corda 4 signed corDapp JAR must be present in the
+   Corda nodes /cordapps directory or imported manually via RPC using the ``uploadAttachment`` command (see
+   :ref:`CorDapp Contract Attachments <cordapp_contract_attachments_ref>` for further information).
+
+
+Later releases
+~~~~~~~~~
```
mikehearn: Can you make the underline match the header length? Sphinx makes warnings during docs compile otherwise :(

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+The next version of Corda will provide automatic migration of *hash constrained* states. This means that signed corDapps running on a Corda 4.x node will
+automatically consume any pre-existing on-ledger *hash-constrained* states (and generate *signature-constrained* outputs).
```
mikehearn: ... *when the system property to break constraints is set*.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+--------------------------
+
+.. note:: These instructions only apply to corDapp Contract JARs (unless otherwise stated).
+
+Corda 4.0
+~~~~~~~~~
+
+Corda 4.0 requires some additional steps to consume pre-existing on-ledger **hash** constrained states:
+
+1. The Corda Node must be started using relaxed hash constraint checking mode as described in :ref:`relax_hash_constraints_checking_ref`.
```
mikehearn: Can we expand this to observe that it's not sufficient for your node to have it set. *Every node* that might encounter this transaction chain, must have it set! Maybe talk here a bit about business networks and such.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+
+CZ whitelisted constraints migration
+-------------------------------------
+
+.. note:: These instructions only apply to corDapp Contract JARs (unless otherwise stated).
+
+Corda 4.0
+~~~~~~~~~
+
+Corda 4.0 requires some additional steps to consume pre-existing on-ledger **CZ whitelisted** constrained states:
```
mikehearn: Ditto. Not to consume - they are consumable out of the box. You're talking here about *migrating* pre-existing states.

docs/source/cordapp-constraint-migration.rst
mikehearn: I'm actually not sure I understand this sentence. Why can't it be done in the usual way via the Gradle plugin? It's not hard to do it manually but - I don't see why? The CZ lists hashes, and the signing process will create a new hash, so, it's effectively an "upgrade" one way or another.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+   The signing capability of :ref:`corda-gradle-plugins <cordapp_build_system_signing_cordapp_jar_ref>` cannot be used in this context as it signs the JAR while building it from source.
+
+2. Both the original pre-Corda 4 corDapp JAR (generating *CZ whitelisted* constrained states) and the new Corda 4 signed corDapp JAR must be
+   registered with the CZ Network operator (as whitelisted in the network parameters which are distributed to all nodes in that CZ).
+   The CZ network operator should check that the JAR is signed and not allow any more versions of it to be whitelisted in the future.
+   From now on the development organisation that signed the JAR is responsible for signing new versions.
+
+   The process of CZ Network corDapp whitelisting depends on how the Corda Network is configured:
+
+    - if using a hosted CZ Network (such as `The Corda Network <https://docs.corda.net/head/corda-network/index.html>`_ or
```
mikehearn: CZ Network -> zone

docs/source/cordapp-constraint-migration.rst
mikehearn: Yep first should already be whitelisted.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,152 @@
...
+        PublicKey ownersKey = signers.get(0);
+
+        TransactionBuilder txBuilder = new TransactionBuilder(notary)
+                // Set the Signature constraint on the new state to migrate away from the WhitelistConstraint.
+                .addOutputState(outputState, myContract, new SignatureAttachmentConstraint(ownersKey))
+
+4. As a node operator you need to add the new signed version of the contracts corDapp to the "cordapps" folder together with the latest version of the flows jar.
+
+Later releases
+~~~~~~~~~
```
mikehearn: Wavy underline length

docs/source/cordapp-constraint-migration.rst
josecoll: search/replaced all references.

docs/source/cordapp-constraint-migration.rst
josecoll: Added

docs/source/cordapp-constraint-migration.rst
josecoll: Adjusteed

docs/source/cordapp-constraint-migration.rst
josecoll: Iv'ed consistently used "CZ network " which I think is clearer, consistent with other docs and adheres to our Corda network concepts education.

docs/source/cordapp-constraint-migration.rst
josecoll: Right ... so maybe I need to clarify a couple more things at the beginning of this doc:
1. States are always `consumable` if the version of the CorDapp that `issued` (created) them is installed.
2.  **Linear states** typically `evolve` over an extended period of time (defined by the lifecycle of the associated business use case), and thus are prime candidates for constraints migration.
**Fungible states** are created by an `issuer` and transferred around a Corda network until explicitly `exited` (by the same `issuer). They do not `evolve` as linear states, but are transferred between participants on a network. Note their consumption may produce additional  new output states to represent adjustments to the original state (e.g. change when spending cash). For the purposes of constraints migration, it is desirable that any new output states are produced using the new signature constraints.

docs/source/cordapp-constraint-migration.rst
josecoll: Good point!

josecoll: Documentation or references to Business Networks are mostly non-existent in Enterprise Corda docs.
There is a new section on something called a "Dynamic CZ" (never heard of this before) - `setting-up-a-dynamic-compatibility-zone.rst`

docs/source/cordapp-constraint-migration.rst
josecoll: > This is interesting... feels like there's some additional detail/nuance behind this that we should expand on, if only to help the reader understand what's going on and how they need to write their contracts? The thoughts going through my mind are:
> 
> * Ahh... that's interesting... it hadn't occurred to me that the contracts would be different... I just assumed you'd attach the same Jar as before (so the tx is effectively a no op from a business perspective but now you have a sig constrained output that can be upgraded in a future tx).  But it seems we're transitioning the constraint AND upgrading the contract at the same time?  Is that because the no-op can't work because the original Jar is unsigned?

I need to make clear that the steps outlined in this page assume we are dealing with the SAME Contract (ie. your "no op from a business perspective"). 

> * Assuming I've understood this correctly, do we have any recommendations here?  eg would a sensible pattern simply be to create a signed jar that has IDENTICAL functionality as the hash-constrained Jar?  so that the tx that changes the constraint is purely technical and has no business function?
Yes - will clarify with recommendation.

> * Either way, perhaps also remind people that it is the NEW Jar that will be attached to the tx and so (obviously - but perhaps not obviously!) it needs to be compatible with the old one since it has to be able to deserialise and reason about the input state?
Yes - mandatory prerequisite.

docs/source/cordapp-constraint-migration.rst
josecoll: Indeed, I copied this step "in-situ" from the other page.
My initial thought is that we shouldn't have this limitation in our gradle plugin (eg. it should be able to sign a pre-existing binary JAR). I suspect this requires a small JIRA to adjust the current behaviour.

docs/source/cordapp-constraint-migration.rst
josecoll: Corrected.

docs/source/cordapp-constraint-migration.rst
josecoll: The assumption is we are dealing with the same version.
I think we need to validate the code above, as I would have used the ``AttachmentQueryCriteria`` to  obtain the correct AttachmentId (eg. the signed version of the Contract CorDapp) using:
``queryAttachments(criteria: AttachmentQueryCriteria, sorting: AttachmentSort? = null): List<AttachmentId>``

docs/source/cordapp-constraint-migration.rst
josecoll: No - both CorDapps may be present in the /cordapps folder (even though the original one should already be in the nodes attachment store). 
I think we need to tighten the query in the code snippet.

docs/source/api-contract-constraints.rst
```diff
@@ -191,58 +193,11 @@ During transaction building the ``AutomaticPlaceholderConstraint`` for output st
...
-        // Create the key that will have to pass for all future versions.
-        PublicKey ownersKey = signers.get(0);
-
-        TransactionBuilder txBuilder = new TransactionBuilder(notary)
-                // Set the Signature constraint on the new state to migrate away from the WhitelistConstraint.
-                .addOutputState(outputState, myContract, new SignatureAttachmentConstraint(ownersKey))
-                ...
-
+Please read :doc:`cordapp-constraint-migration` to understand how to consume pre-existing hash or CZ whitelisted constrained states
+issued using a previous version of Corda using Corda 4 signed CorDapps (using signature constraints).
```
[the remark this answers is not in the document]
josecoll: Clarified title and paragraph.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,182 @@
...
+
+
+   <script type="text/javascript" src="_static/jquery.js"></script>
+   <script type="text/javascript" src="_static/codesets.js"></script>
+
+CorDapp constraints migration
+=============================
+
+.. note:: Before reading this page, you should be familiar with the key concepts of :doc:`Contract Constraints <api-contract-constraints>`.
+
```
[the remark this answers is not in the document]
josecoll: Updated.

josecoll:
I have created and validated 2 additional QA TST scenarios for these migration instructions:
https://r3-cev.atlassian.net/browse/R3T-2566
https://r3-cev.atlassian.net/browse/R3T-2566

josecoll:
@mikehearn Please re-review. Thanks.

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,187 @@
...
+Corda 4 introduces and recommends building signed CorDapps that issue states with signature constraints.
+Existing on ledger states issued before Corda 4 are not automatically propagated to new signature constraints when building transactions in Corda 4.
```
mikehearn: propagated? or is "upgraded" a better word?

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,187 @@
...
+This document explains how to modify existing CorDapp flows to explicitly consume and evolve pre Corda 4 states, and outlines a future mechanism
+where such states will propagate automatically (without explicit migration code).
```
mikehearn: propagate -> upgrade/transition

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,187 @@
...
+
+Faced with the exercise of upgrading an existing Corda 3.x CorDapp to Corda 4, you need to consider the following:
+
+* What existing unconsumed states have been issued on ledger by a previous version of this CorDapp and using other constraint types?
+
+   If you have existing **hash** constrained states see :ref:`Migrating hash constraints<hash_constraint_migration>`.
+
+   If you have existing **CZ whitelisted** constrained states see :ref:`Migrating CZ whitelisted constraints<cz_whitelisted_constraint_migration>`.
```
mikehearn: CZ whitelisted -> compatibility zone whitelisted (at least for the first use, the other uses can be left abbreviated)

docs/source/cordapp-constraint-migration.rst
```diff
@@ -0,0 +1,187 @@
...
+                .addOutputState(outputState, myContract, new SignatureAttachmentConstraint(ownersKey))
+
+3. As a node operator you need to add the new signed version of the contracts CorDapp to the ``/cordapps`` folder together with the latest version of the flows jar.
+   Please also ensure that the original unsigned contracts CorDapp is removed from the ``/cordapps`` folder (this will already be present in the
+   nodes attachments store) to ensure the lookup code in step 2 retrieves the correct signed contract CorDapp JAR.
+
+Later releases
+~~~~~~~~~~~~~~
+
+The next version of Corda will provide automatic propagation of *hash constrained* states. This means that signed CorDapps running on a Corda 4.x node will
```
mikehearn: propagation -> migration/upgrade/transition

docs/source/cordapp-constraint-migration.rst
mikehearn: I'm still confused. It doesn't have to be the pre-existing binary when using CZ constraints, right? The signing is just a part of the next upgrade, and the app can be upgraded in all the normal ways. You don't have to do a "non-upgrade upgrade" with this constraint type.

As far as I can tell this specified requirement doesn't really exist in the code.

docs/source/cordapp-constraint-migration.rst
josecoll: I originally used the word "upgraded" but a review comment suggested I use the word "propagated" to align with this description in the documentation:
https://docs.corda.net/releases/release-V4.0/api-contract-constraints.html#constraints-propagation

docs/source/cordapp-constraint-migration.rst
josecoll: "transition" sounds better as propagation is used in a slightly different context:
https://docs.corda.net/releases/release-V4.0/api-contract-constraints.html#constraints-propagation

docs/source/cordapp-constraint-migration.rst
josecoll: CZ whitelisted is used consistently throughout the text.

docs/source/cordapp-constraint-migration.rst
josecoll: will switch to "transition" (I already renamed this term from "upgrade" to "propagate")

docs/source/cordapp-constraint-migration.rst
josecoll: Will adopt the word "transition" to avoid clashing with propagation.

docs/source/cordapp-constraint-migration.rst
josecoll: The reader should be familiar with this term from reading the prerequisite background reading where this is introduced. I will modify the first reference to make this clear as the original text calls these "Zone whitelist constraint"

PR #4842, opened 2019-03-03
CORDA-2694: Prevent Node Explorer from crashing should it receive unknown transaction objects.

chrisr3:
Forcibly deserialise the `WireTransaction`'s `TransactionState<ContractState>` output objects up-front, replacing any that fail to deserialise with a dummy `UNKNOWN_TRANSACTION_STATE` object instead.

Also simplify some lambda-spaghetti and remove an unused import.

### DIFF
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
@@ -14,7 +14,7 @@ import net.corda.core.internal.LazyMappedList
 import net.corda.core.transactions.LedgerTransaction
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.transactions.WireTransaction
-import java.math.BigInteger
+import java.math.BigInteger.ZERO
 
 private class Unknown : Contract {
     override fun verify(tx: LedgerTransaction) = throw UnsupportedOperationException()
@@ -54,7 +54,7 @@ data class PartiallyResolvedTransaction(
     }
 
     companion object {
-        private val UNKNOWN_PARTY = Party(CordaX500Name("Unknown Party", "Nowhere", "ZZ"), entropyToKeyPair(BigInteger.ZERO).public)
+        private val UNKNOWN_PARTY = Party(CordaX500Name("Unknown Party", "Nowhere", "ZZ"), entropyToKeyPair(ZERO).public)
         private val UNKNOWN_TRANSACTION_STATE: TransactionState<ContractState> = TransactionState(Unknown.State, Unknown::class.java.name, UNKNOWN_PARTY)
 
         fun fromSignedTransaction(
--- a/client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
+++ b/client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
@@ -54,8 +54,7 @@ data class PartiallyResolvedTransaction(
     }
 
     companion object {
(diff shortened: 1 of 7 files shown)


JamesHR3:
Given the fix is being committed to OS, should the ticket be moved to CORDA from ENT?

chrisr3:
@JamesHR3 Possibly. Who is the correct "JIRA Bunny" to do that? :rabbit:

JamesHR3:
I think it's possible to do it yourself - if you go to the ticket in JIRA [here](https://r3-cev.atlassian.net/browse/ENT-3145?filter=-1), there's a "Move" option under the three dots on the top right of the page. If that doesn't work I'm not sure who the bunny is! @davidrapacchiale is a JIRA wizard though so perhaps he can produce one

client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
```diff
@@ -41,10 +54,19 @@ data class PartiallyResolvedTransaction(
...
         fun fromSignedTransaction(
                 transaction: SignedTransaction,
                 inputTransactions: Map<StateRef, SignedTransaction?>
         ): PartiallyResolvedTransaction {
+            /**
+             * Forcibly deserialize our transaction outputs up-front.
+             * Replace any [TransactionState] objects that fail to
+             * deserialize with [UNKNOWN_TRANSACTION_STATE].
+             */
+            transaction.coreTransaction.outputs.eagerDeserialise { _, _ -> UNKNOWN_TRANSACTION_STATE }
```
rick-r3: I think you could actually fill in the notary from the transaction?  I guess that means you need an unknown output per transaction rather than a global.  Some code (now or in the future) might care whether all outputs point at the same notary (because they are supposed to).

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
```diff
@@ -540,6 +550,16 @@ class LazyMappedList<T, U>(val originalList: List<T>, val transform: (T, Int) ->
  */
 fun <T, U> List<T>.lazyMapped(transform: (T, Int) -> U): List<U> = LazyMappedList(this, transform)
 
+/**
+ * Iterate over a [LazyMappedList], forcing it to transform all of its elements immediately.
+ * This transformation is assumed to be "deserialisation". Does nothing for any other kind of [List].
+ */
```
rick-r3: It's perhaps worth a warning here that this mutates the source `List` and thus be very careful since the mapping is permanent.

client/jfx/src/main/kotlin/net/corda/client/jfx/model/TransactionDataModel.kt
chrisr3: :+1:

core/src/main/kotlin/net/corda/core/internal/InternalUtils.kt
chrisr3: :+1:

rick-r3: I'm now thinking we should perhaps have made this work for any exception.

chrisr3:
That wouldn't have been necessary because the deserialiser code already wraps almost any exception into a `TransactionDeserialisationException`. Similarly, I doubt we'd want to allow developers to ignore `MissingAttachmentException`s either:
```kotlin
components.lazyMapped { component, internalIndex ->
    try {
        factory.deserialize(component, clazz.java, context)
    } catch (e: MissingAttachmentsException) {
        throw e
    } catch (e: Exception) {
        throw TransactionDeserialisationException(groupEnum, internalIndex, e)
    }
}
```

PR #4844, opened 2019-03-04
CORDA-2690: Reduce CordApp scanning logging

lockathan:
Only log a line about looking for notary implementations if at least notary implementation is found.

It might be unnecessary and confusing to log when no implementations are found.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
@@ -212,12 +212,14 @@ class JarScanningCordappLoader private constructor(private val cordappJarPaths: 
     }
 
     private fun findNotaryService(scanResult: RestrictedScanResult): Class<out NotaryService>? {
-        // Note: we search for implementations of both NotaryService and TrustedAuthorityNotaryService as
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
```diff
@@ -217,7 +217,9 @@ class JarScanningCordappLoader private constructor(private val cordappJarPaths:
         // present in the CorDapp.
```
adagys: just spotted the comment is outdated, could you replace `TrustedAuthorityNotaryService` with `SinglePartyNotaryService` while you're at it?

node/src/main/kotlin/net/corda/node/internal/cordapp/JarScanningCordappLoader.kt
```diff
@@ -217,7 +217,9 @@ class JarScanningCordappLoader private constructor(private val cordappJarPaths:
...
         val result = scanResult.getClassesWithSuperclass(NotaryService::class) +
                 scanResult.getClassesWithSuperclass(SinglePartyNotaryService::class)
-        logger.info("Found notary service CorDapp implementations: " + result.joinToString(", "))
+        if(!result.isEmpty()) {
```
adagys: kotlin also provides a handy `.isNotEmpty`

PR #4845, opened 2019-03-04
[CORDA-2698] Allow theoretical page size of `Integer.MAX_VALUE`

thschroeter:


rick-r3:
Expected outcome of existing test needs fixing up.

JamesHR3:
This change is the same in OS as well as ENT isn't it? If so, would you mind moving the ticket to CORDA from ENT? I think you can do this from the JIRA ticket ([here](https://r3-cev.atlassian.net/browse/ENT-3213)) - there should be a "Move" option under the three dots at the top right of the page.

PR #4847, opened 2019-03-04
CORDA-2706: Change logging level from warn to info

kasiastreich:
On existing session message when can't find a flow corresponding to
session ID. See JIRA: ENT-3058


JamesHR3:
Assuming this ticket has the same fix in OS and ENT, would you mind moving the JIRA ticket to CORDA from ENT? On the JIRA page [here](https://r3-cev.atlassian.net/browse/ENT-3058), there's a "Move" option under the three dots on the top right of the page. Thanks!

kasiastreich:
For 1 word change fix this is too much formalities. What is the reason for doing that apart from love for JIRA and rules?

PR #4848, opened 2019-03-04
CORDA-2696 eliminate unwanted duplicate class warnings

r3domfox:
To reduce the number of `More than one CorDapp registered...` warnings logged when registering cordapps, we log warnings only for classes that are listed in the cordapp's registry of `contractClasses`, and use `logger.warnOnce` to eliminate duplicate log messages.


josecoll:
As discussed, this small improvement can go onto master for post OS4.1/CE4.0.

PR #4853, opened 2019-03-05
DOCS - clarify the upgrade policy for finality flow

dimosr:
This is a small change in the upgrade docs related to the `FinalityFlow` in order to make a bit more explicit the difference between initiating & non-initiating flows and the implications on the possible ways the upgrade can be deployed.

### DIFF
--- a/docs/source/app-upgrade-notes.rst
+++ b/docs/source/app-upgrade-notes.rst
@@ -30,6 +30,8 @@ Although the RPC API is backwards compatible with Corda 3, the RPC wire protocol
 updated in lockstep with the node to use the new version of the RPC library. Corda 4 delivers RPC wire stability and therefore in future you
 will be able to update the node and apps without updating RPC clients.
 
+.. _cordapp_upgrade_version_numbers_ref:
+
 Step 2. Adjust the version numbers in your Gradle build files
 -------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/source/app-upgrade-notes.rst
```diff
@@ -163,11 +165,20 @@ The previous ``FinalityFlow`` API is insecure. It doesn't have a receive flow, s
...
 away to the new API, as otherwise things like business network membership checks won't be reliably enforced.
 
-This is a three step process:
+The flows that make use of ``FinalityFlow`` in a CorDapp can be classified in the following 2 basic categories:
+
+* **non-initiating flows**: these are flows that finalise a transaction without the involvement of a counterpart flow at all.
+* **initiating flows**: these are flows that initiate a counterpart (responder) flow.
+
+There is a main difference between these 2 different categories, which is relevant to how the CorDapp can be upgraded.
+The second category of flows can be upgraded to use the new ``FinalityFlow`` in a backwards compatible way, which means the upgraded CorDapp can be deployed at the various nodes using a rolling deployment.
```
josecoll: Can you add a note explaining what a "rolling deployment" means.

docs/source/app-upgrade-notes.rst
```diff
@@ -163,11 +165,20 @@ The previous ``FinalityFlow`` API is insecure. It doesn't have a receive flow, s
...
+On the other hand, the first category of flows cannot be upgraded to the new ``FinalityFlow`` in a backwards compatible way, so the changes to these flows need to be deployed simultaneously at all the nodes, using a lockstep deployment.
```
josecoll: Can you add a note explaining what a "lockstep deployment" means.

docs/source/app-upgrade-notes.rst
dimosr: Good point, yes I'll add a quick explanation for both terms.

PR #4854, opened 2019-03-05
[CORDA-2701] Ensure crlCheckSoftFail config option is respected

JamesHR3:
The `crlCheckSoftFail` config option is currently ignored by the node in some circumstances. In particular, it acts as if it were always set to `true`, which allows connections to be made between nodes and flows to be executed even if some CRLs in the certificate hierarchy are invalid or unavailable. This occurs because the default value in `AMQPConfiguration` is `true`, and this isn't overridden by the Bridge Manager. This fix plumbs the config value through to the Bridge Manager.

Changes:
 - Plumb the config value through to the Bridge Manager
 - Remove the default value from the `AMQPConfiguration` interface. This makes it easier to detect instances where the config hasn't been set correctly by failing the build if this is the case.
 - Add a test case to `AMQPBridgeTest` to show that with the option set to false, connections cannot be made to AMQP servers with an invalid certificate hierarchy
 - Fix the equivalent test case in `ProtonWrapperTests`. Previously, this test case would never connect even if the soft fail option were set to true, as it attempted to connect to an invalid server.

Testing:
Along with the unit tests above, also executed the test case in which this was found with a fixed version of the Corda jar. Logs from this are attached below.
[boc.log](https://github.com/corda/corda/files/2930790/boc.log)
[notary.log](https://github.com/corda/corda/files/2930791/notary.log)
[bankA.log](https://github.com/corda/corda/files/2930792/bankA.log)

### DIFF
--- a/docs/source/corda-configuration-file.rst
+++ b/docs/source/corda-configuration-file.rst
@@ -91,7 +91,8 @@ crlCheckSoftFail
   This is a boolean flag that when enabled (i.e. ``true`` value is set) causes certificate revocation list (CRL) checking to use soft fail mode.
   Soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
   If this parameter is set to ``false`` rigorous CRL checking takes place. This involves each certificate in the certificate path being checked for a CRL distribution point extension, and that this extension points to a URL serving a valid CRL.
-  If the CRL distribution point is inaccessible and this option is set to ``false``, then the revocation check will fail.
+  This means that if any CRL URL in the certificate path is inaccessible, the connection with the other party will fail and be marked as bad.
+  Additionally, if any certificate in the hierarchy, including the self-generated node SSL certificate, is missing a valid CRL URL, then the certificate path will be marked as invalid.
 
   *Default:* true
 
--- a/docs/source/corda-configuration-file.rst
+++ b/docs/source/corda-configuration-file.rst
@@ -88,9 +88,10 @@ cordappSignerKeyFingerprintBlacklist
   *Default:* not defined
 
 crlCheckSoftFail
-  This is a boolean flag that when enabled (i.e. ``true`` value is set) causes the certificate revocation list (CRL) checking will use the soft fail mode.
-  The soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
-  If this parameter is set to ``false`` rigorous CRL checking takes place, involving each certificate in the certificate path being checked needs to have the CRL distribution point extension set and pointing to a URL serving a valid CRL.
+  This is a boolean flag that when enabled (i.e. ``true`` value is set) causes certificate revocation list (CRL) checking to use soft fail mode.
+  Soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
+  If this parameter is set to ``false`` rigorous CRL checking takes place. This involves each certificate in the certificate path being checked for a CRL distribution point extension, and that this extension points to a URL serving a valid CRL.
+  If the CRL distribution point is inaccessible and this option is set to ``false``, then the revocation check will fail.
 
   *Default:* true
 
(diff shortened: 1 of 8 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPConfiguration.kt
```diff
@@ -37,7 +37,6 @@ interface AMQPConfiguration {
      */
     @JvmDefault
     val crlCheckSoftFail: Boolean
-        get() = true
```
mkit: I would leave that as this is a default value. Make sure that docs speak about what is the default value.

JamesHR3: I'm slightly nervous to revert that particular change, as it was the presence of this default that allowed the bug to happen in the first place! (Without that line, the project doesn't build because crlCheckSoftFail isn't overridden in the AMQP bridge, which immediately reveals the problem.)

I think the node configuration should make it explicit that the default is `true` though, so I've added this to the node config defaults. I've also slightly reworded the documentation around this config option that should hopefully make it a bit clearer what it does.

docs/source/corda-configuration-file.rst
```diff
@@ -88,9 +88,10 @@ cordappSignerKeyFingerprintBlacklist
...
 
 crlCheckSoftFail
-  This is a boolean flag that when enabled (i.e. ``true`` value is set) causes the certificate revocation list (CRL) checking will use the soft fail mode.
-  The soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
-  If this parameter is set to ``false`` rigorous CRL checking takes place, involving each certificate in the certificate path being checked needs to have the CRL distribution point extension set and pointing to a URL serving a valid CRL.
+  This is a boolean flag that when enabled (i.e. ``true`` value is set) causes certificate revocation list (CRL) checking to use soft fail mode.
+  Soft fail mode allows the revocation check to succeed if the revocation status cannot be determined because of a network error.
+  If this parameter is set to ``false`` rigorous CRL checking takes place. This involves each certificate in the certificate path being checked for a CRL distribution point extension, and that this extension points to a URL serving a valid CRL.
+  If the CRL distribution point is inaccessible and this option is set to ``false``, then the revocation check will fail.
```
mnesbit: The consequences of of check false need to be spelled out very clearly! They are horrible:
1. If you cannot reach EVERY CRL URL in the cert path the connection fails and will be marked as bad.
2. If every certificate INCLUDING the self-generate node SSL keys are not properly configured with a CRL URL then the cert path is INVALID.

node-api/src/main/kotlin/net/corda/nodeapi/internal/protonwrapper/netty/AMQPConfiguration.kt
```diff
@@ -37,7 +37,6 @@ interface AMQPConfiguration {
      */
     @JvmDefault
     val crlCheckSoftFail: Boolean
```
mnesbit: This default was intentional.
1. The only sensible value was true for reasons above
2. The interface in OS and ENT are different and I very intentionally don't want a Data clas anywhere to complicate things and cause yet more merge confilcts.

JamesHR3: Fair enough - will revert

node/src/main/kotlin/net/corda/node/services/config/NodeConfigurationImpl.kt
```diff
@@ -108,6 +108,7 @@ data class NodeConfigurationImpl(
         val jmxReporterType: JmxReporterType = NodeConfiguration.defaultJmxReporterType
         val cordappSignerKeyFingerprintBlacklist: List<String> = DEV_PUB_KEY_HASHES.map { it.toString() }
         val networkParameterAcceptanceSettings: NetworkParameterAcceptanceSettings = NetworkParameterAcceptanceSettings()
+        val crlCheckSoftFail: Boolean = true
```
mkit: Shouldn't that default be set around the interface AMQPConfiguration interface. It uses true there. Perhaps it would be good to have that constant defined somewhere there and then reuse it here.

JamesHR3: I think it probably makes sense to specify a default in one place only. If that's on the AMQPConfiguration interface, then it probably shouldn't be in the NodeConfiguration too.

PR #4856, opened 2019-03-06
CORDA-2707 defer exceptions to deserialisation-time

r3domfox:
We allow serializers to be generated in the presence of non-deserialisable types (types that aren't on the receiver's classpath, and which we don't want to carpent up), and defer throwing exceptions to when a non-deserialisable value is required in the construction of a new value. This enables the evolution serialiser to work in the case that the original type has a property of a type that is no longer available in the new type: if the property itself has been removed, then it doesn't matter if we haven't carpented up a type to deserialise its value into.

If carpentry is enabled, then all missing types will be carpented, whether they are used or not, as it's too early in the deserialisation pipeline to know exactly which types will be required.

A new type, `NonDeserialisable`, is introduced to represent a value in the serialisation stream that could not be deserialised. Most clients of `DeserialisationInput` use `DeserialisationInput.read` to obtain deserialised values, and this method fails with an exception rather than returning a `NonDeserialisable`. However, serialisers which obtain part-values from the stream and use them to build a result call into `DeserialisationInput.readObjectOrNull` instead, and this method will return a value of type `NonDeserialisable` if there is no serialiser available to read the type in the stream.

It is then up to the consumer of that value to signal whether the presence of non-deserialisable values means that deserialisation cannot be completed. This enables `ObjectBuilder`s to consume such values and throw them away when they are not needed to construct objects, typically because they are assigned to a property that the object under construction no longer possesses. Other serialisers which may have to handle a `NonDeserialisable` value are `CollectionSerializer`, `MapSerializer` and `ArraySerializer`, which have been modified to fail if any of the elements in the collection is non-deserialisable. All three are covered by unit tests.

There is a risk that other custom serialisers which use `DeserialisationInput.readObjectOrNull` to obtain part-values during deserialisation will fail with a `ClassCastException` when given `NonDeserialisable`, if they have not been explicitly modified to handle this case. Any custom deserialiser which accepts part-values of type `Any` will not fail at all, but may end up constructing an object containing `NonDeserialisable` values, which may cause trouble further down the line. One possible way to mitigate this would be to introduce a method `DeserialisationInput.unsafeReadObjectOrNull` to be used only by those serialisers which explicitly handle `NonDeserialisable` values, and have `readObjectOrNull` throw an exception rather than ever returning a `NonDeserialisable`. However, as `readObjectOrNull` is already marked as `internal`, it is unlikely that ad hoc use will occur.

Deserialisation should fail if data-loss protection is enabled and a non-null value of a non-deserialisable type is received for a property that no longer exists, but succeed if the value passed is null. The `NonDeserialisable` type represents the non-null value case, while `null` is still the null value. A test has been added to `EvolvabilityTests` which exercises this scenario.

---

This behaviour is controlled by a feature flag, specified by the configuration property `featureFlags.disableCorda2707`. Feature flags themselves are supported by the new `FeatureFlag` class. This class does a couple of _very_ disreputable things, for reasons I will try to defend here.

1) A feature flag is effectively a global constant, read from node configuration and set once and for all during node initialisation.

We _could_ try to plumb the node config through to the initialisation of serialisation environments, rather than using a global value in this way. However, the ability to place a simple boolean check around a block of code, without having to add the explicit path of dependencies between that block of code and node initialisation, better fits the pattern of how I imagine feature flags being used: in an ad hoc way, to control small pieces of logic here and there.

Ideally, wherever this happens the flag itself should be referenced, so that if we are going to remove a flag (or want to search for the places where it will be used) we can use the standard IDE mechanisms for searching for usages of some function or constant. (Of course we could use some commenting convention, and full text search, instead).

The major downside of this approach is that if we run tests that initialise multiple nodes in-process, we cannot pass feature flag configuration to them, since this configuration will be set for the _entire_ process and all of the nodes (along with any other tests) running within it.

2) The value of each feature flag is overridable during testing by setting a `ThreadLocal` value.

This is done to enable us to write tests that verify that a particular piece of code behaves in a particular way when some combination of flags is set. It is absolutely _not_ intended to be used outside of this scenario.


fenryka (changes requested):
In the spirit of "Changes gooing forward for the betterment of Corda" I wonder if we shouldn't put this (and my) change behind a flag. We're changing behaviour we *might* want to turn off in the future. Plus, it's a good testing ground  for adding such "feature" flags

Change Log?

Otherwise LGTM!

gendal:
good point @fenryka ! We keep saying we need to protect the codebase with more features we can turn off through a flag.  I've never 100% understood how it would work in practice... does that mean we need to build with it on and off from now on to check both codepaths work?  or is it simpler than that?  I'm all for proving it out here if it makes sense

serialization/src/main/kotlin/net/corda/serialization/internal/model/TypeLoader.kt
```diff
@@ -21,8 +22,16 @@ interface TypeLoader {
 /**
  * A [TypeLoader] that uses the [ClassCarpenter] to build a class matching the supplied [RemoteTypeInformation] if none
  * is visible from the current classloader.
+ *
+ * @param mustCarpentMissingTypes Controlled by feature flag DISABLE-CORDA-2707 - if this flag is set, then a [ClassNotFoundException]
```
mikehearn: Seems this param was taken out.

mikehearn (changes requested):

mikehearn (dismissed):

core/src/main/kotlin/net/corda/core/context/FeatureFlag.kt
```diff
@@ -0,0 +1,131 @@
...
+import kotlin.reflect.KProperty0
+import kotlin.reflect.jvm.isAccessible
+
+/**
+ * Encapsulates reading a feature flag from the node configuration.
+ *
+ * A feature flag can be temporarily set for testing using the withSet method.
+ */
+@KeepForDJVM
+object FeatureFlag  {
```
rick-r3: I think we should put this inside the `internal` package of `core` as an indicator it's for R3 not consumers of `core` (and eventually we will block access).

serialization/src/main/kotlin/net/corda/serialization/internal/amqp/CollectionSerializer.kt
```diff
@@ -117,15 +117,16 @@ class CollectionSerializer(private val declaredType: ParameterizedType, factory:
...
             obj: Any,
             schemas: SerializationSchemas,
             input: DeserializationInput,
             context: SerializationContext): Any = ifThrowsAppend({ declaredType.typeName }) {
         // TODO: Can we verify the entries in the list?
-        concreteBuilder((obj as List<*>).map {
-            input.readObjectOrNull(it, schemas, inboundType, context)
+        concreteBuilder((obj as List<*>).mapIndexed { index, serializedItem ->
+            val item = input.readObjectOrNull(serializedItem, schemas, inboundType, context)
+            if (item is NonDeserializable) throw NotSerializableException("Collection item $index is non-deserializable ${item.descriptor}")
```
rick-r3: Why do you check this return value after each call to `readObjectOrNull()`, rather than checking before return from `readObjectOrNull()` ?

r3domfox: Because `readObjectOrNull` is used in cases where we don't mind if the result is `NonDeserializable` (or don't yet know if we mind - it gets passed forward to an object builder which might or might not just throw it away):

> A new type, `NonDeserialisable`, is introduced to represent a value in the serialisation stream that could not be deserialised. Most clients of `DeserialisationInput` use `DeserialisationInput.read` to obtain deserialised values, and this method fails with an exception rather than returning a `NonDeserialisable`. However, serialisers which obtain part-values from the stream and use them to build a result call into `DeserialisationInput.readObjectOrNull` instead, and this method will return a value of type `NonDeserialisable` if there is no serialiser available to read the type in the stream.

core/src/main/kotlin/net/corda/core/context/FeatureFlag.kt
r3domfox: Done.

shamsasari:
This PR is now bit-rotting.

PR #4859, opened 2019-03-07
CORDA-2714 parallelize download of nodeInfos

roastario:


josecoll:
Nice .... give us some benchmark numbers !!!
eg. measure a full DayWatch NMS download with/without this patch.

shamsasari:
This is conceptually fine, but it has to be punted until 4.1 is out the door, since that will be released off master. If however the intention is for 4.1 then it needs to be approved by product/release management/powers-that-be.

roastario:
> Nice .... give us some benchmark numbers !!!
> eg. measure a full DayWatch NMS download with/without this patch.

Vanilla Corda: `Node for "BankOfCordaStefanoTest" started up and registered in 179.53 sec`
Super HFT Corda: `Node for "BankOfCordaStefanoTest" started up and registered in 42.84 sec`

josecoll:
> > Nice .... give us some benchmark numbers !!!
> > eg. measure a full DayWatch NMS download with/without this patch.
> 
> Vanilla Corda: `Node for "BankOfCordaStefanoTest" started up and registered in 179.53 sec`
> Super HFT Corda: `Node for "BankOfCordaStefanoTest" started up and registered in 42.84 sec`

Using DayWatch with >1000 nodeInfos right?

wawrzek:
1000 nodes is not a big number.  I don't think we can release CE4 without it.

mikehearn:
Of course we can, this is not essential right now. We are very far from zones with 1000 identities *AND* where people care about first-run startup times. Daywatch can easily change setup to delete unused identities. 

This PR introduces some risk into core code that'd need to go into Corda 5, not the 4.x branch.

@roastario Is there a reason why we couldn't use `.parallelStream().` and need a custom forkjoin pool? The general pool is meant to be usable for everything.

roastario:
> @roastario Is there a reason why we couldn't use `.parallelStream().` and need a custom forkjoin pool? The general pool is meant to be usable for everything.

Yes unfortunately the general pool is sized to the Runtime.availableProcessors(), Here I try to make it run more threads than CPU's. On my laptop, it runs about 2x faster with threads=CPUs*4 than with just threads=CPUs.

roastario:
> > > Nice .... give us some benchmark numbers !!!
> > > eg. measure a full DayWatch NMS download with/without this patch.
> > 
> > 
> > Vanilla Corda: `Node for "BankOfCordaStefanoTest" started up and registered in 179.53 sec`
> > Super HFT Corda: `Node for "BankOfCordaStefanoTest" started up and registered in 42.84 sec`
> 
> Using DayWatch with >1000 nodeInfos right?

Yep! It downloaded 1684 identities.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -159,30 +203,7 @@ open class PersistentNetworkMapCache(cacheFactory: NamedCacheFactory,
...
-                if (!verifyIdentities(node)) return
-                database.transaction {
-                    updateInfoDB(node, session)
-                    changePublisher.onNext(MapChange.Modified(node, previousNode))
-                }
-            } else {
-                logger.info("Previous node was identical to incoming one - doing nothing")
-            }
-        }
+
```
rick-r3: Shouldn't his call `addNodes(listOf(node))` or something?  How come no tests fail?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -170,20 +172,26 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
-            // Download new node info from network map
-            try {
-                networkMapClient.getNodeInfo(it)
-            } catch (e: Exception) {
-                // Failure to retrieve one node info shouldn't stop the whole update, log and return null instead.
-                logger.warn("Error encountered when downloading node info '$it', skipping...", e)
-                null
+        //at the moment we use a blocking HTTP library - but under the covers, the OS will interleave threads waiting for IO
+        //as HTTP GET is mostly IO bound, use more threads than CPU's
+        val poolToUse = ForkJoinPool(Runtime.getRuntime().availableProcessors() * 4)
```
rick-r3: Let's give the threads useful names (might need a `ThreadFactory` or something).  I also think that perhaps we should cap the threads at a maximum.  Think about nodes running on 128 core server, multiplied by number of nodes in the network.  Seems like at some point there's diminishing returns to the multiple threads. e.g. if we say it should be `max(16, Runtime.getRuntime().availableProcessors() * 4)` is that good enough?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -170,20 +172,26 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+        poolToUse.submit {
+            (allHashesFromNetworkMap - currentNodeHashes).chunked(100).stream().parallel().mapNotNull { nodeInfosToGet ->
+                // Download new node info from network map
+                try {
+                    logger.info("Using Thread: ${Thread.currentThread().name} to retrieve: ${nodeInfosToGet.size} nodeInfos")
```
rick-r3: Shouldn't need to log the thread name here, as it is part of the log message formatting.

rick-r3:
Can you create a JIRA and put in the title?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -170,20 +172,26 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+                    nodeInfosToGet.map { nodeInfo -> networkMapClient.getNodeInfo(nodeInfo) }
+                } catch (e: Exception) {
+                    // Failure to retrieve one node info shouldn't stop the whole update, log and return null instead.
+                    logger.warn("Error encountered when downloading node info '$nodeInfosToGet', skipping...", e)
+                    null
+                }
+            }.forEach { retrievedNodeInfos ->
+                // Add new node info to the network map cache, these could be new node info or modification of node info for existing nodes.
+                networkMapCache.addNodes(retrievedNodeInfos)
```
rick-r3: One further thing: Is this last `forEach` happening in parallel or single threaded?  There's no point in it being parallel since there's a mutex in `addNodes`.  We probably don't want it concurrent going to the database anyway or we need to change the database connection pool size calculations (currently it allows 1 connection for network map updater).

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
```diff
@@ -40,6 +40,50 @@ import javax.annotation.concurrent.ThreadSafe
...
+                                    updatedNodes.add(node to previousNode)
+                                }
+                            }
+                            else -> logger.info("Previous node was identical to incoming one - doing nothing")
+                        }
+                    }
+            updatedNodes.forEach { (node, previousNode) ->
+                //updated
+                database.transaction {
+                    updateInfoDB(node, session)
```
rick-r3: Shouldn't we have a single transaction around the bulk add/updates ?

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
roastario: It is executed in "parallel" but with the big `synchronized` block, in addNodes, it's not actually done in parallel. 

I figured the extra complexity of collecting the results and then flatmapping them into one wasn't worth it. I'll investigate.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
roastario: Turns out you cannot give a Thread a name when using a `ForkJoinWorkerThreadFactory` so I've migrated it to using a normal executor and `CompletableFuture`

I set it to 24 maximum. Which is what my AzureVM runs with.

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
roastario: many tests did :stuck_out_tongue:

node/src/main/kotlin/net/corda/node/services/network/PersistentNetworkMapCache.kt
roastario: we do now :+1:

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
roastario: removed this log line entirely - it was just for dev purposes

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
roastario: migrated to being multithreaded for download, single threaded for inserts.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -170,18 +176,49 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
-                networkMapClient.getNodeInfo(it)
-            } catch (e: Exception) {
-                // Failure to retrieve one node info shouldn't stop the whole update, log and return null instead.
-                logger.warn("Error encountered when downloading node info '$it', skipping...", e)
-                null
+        //at the moment we use a blocking HTTP library - but under the covers, the OS will interleave threads waiting for IO
+        //as HTTP GET is mostly IO bound, use more threads than CPU's
+        //maximum threads to use = 24.
+        val threadsToUseForNetworkMapDownload = min(Runtime.getRuntime().availableProcessors() * 4, 24)
+        val executorToUseForDownloadingNodeInfos = Executors.newFixedThreadPool(threadsToUseForNetworkMapDownload, object : ThreadFactory {
```
rick-r3: Use `NamedThreadFactory`

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
```diff
@@ -170,18 +176,49 @@ class NetworkMapUpdater(private val networkMapCache: NetworkMapCacheInternal,
...
+            val threadCounter = AtomicInteger(1)
+            override fun newThread(r: Runnable?): Thread {
+                return Thread(r, "NetworkMapDownloadThread${threadCounter.getAndIncrement()}")
+            }
+        })
+        //DB insert is single threaded - use a single threaded executor for it.
+        val executorToUseForInsertionIntoDB = Executors.newSingleThreadExecutor()
```
rick-r3: It's worth naming the thread.  There's a version that can take one of our `NamedThreadFactory`.

node/src/main/kotlin/net/corda/node/services/network/NetworkMapUpdater.kt
fenryka: Shouldn't it be cores+1 or something meaningful

josecoll:
@mikehearn this really useful PR seems to be bit-rotting (conflicting files).
Why didn't we merge it when it was ready (and thus would now be in CE 4.0 and C 4.1)?

roastario:
Closing to retarget c5

PR #4860, opened 2019-03-07
Docs: use version numbers read from the build system, not duplicated.

mikehearn:
Docs: use version numbers read from the build system, not duplicated.


anthonykeenan:
Have you checked that the substitutions work with the PDF as well? I have a niggling thought that someone tried to do this a while ago and it didn't work in some circumstances.

mikehearn:
@anthonykeenan Yep, it works for me.

mikehearn:
Unit test failure seems unrelated, -> build cop.

PR #4861, opened 2019-03-07
CORDA-2569: Add "flow kill <ID>" command to Corda's shell.

chrisr3:
Add "kill" to the list of explicit flow operations allowed from Corda's shell. Also fix Jackson support for `StateMachineRunId` so that we can serialise UUID values directly to flow IDs.

### DIFF
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
@@ -15,16 +15,20 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize
 import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier
 import com.fasterxml.jackson.databind.deser.ContextualDeserializer
 import com.fasterxml.jackson.databind.deser.std.DelegatingDeserializer
+import com.fasterxml.jackson.databind.deser.std.FromStringDeserializer
 import com.fasterxml.jackson.databind.module.SimpleModule
 import com.fasterxml.jackson.databind.node.IntNode
 import com.fasterxml.jackson.databind.node.ObjectNode
 import com.fasterxml.jackson.databind.ser.BeanPropertyWriter
 import com.fasterxml.jackson.databind.ser.BeanSerializerModifier
+import com.fasterxml.jackson.databind.ser.std.StdScalarSerializer
+import com.fasterxml.jackson.databind.ser.std.UUIDSerializer
 import com.google.common.primitives.Booleans
 import net.corda.client.jackson.JacksonSupport
 import net.corda.core.contracts.*
 import net.corda.core.crypto.*
 import net.corda.core.crypto.PartialMerkleTree.PartialTree
+import net.corda.core.flows.StateMachineRunId
 import net.corda.core.identity.*
 import net.corda.core.internal.DigitalSignatureWithCert
 import net.corda.core.internal.createComponentGroups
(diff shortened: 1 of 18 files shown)


tools/shell/src/main/java/net/corda/tools/shell/FlowShellCommand.java
```diff
@@ -63,11 +63,20 @@ static void startFlow(@Usage("The class name of the flow to run, or an unambiguo
...
+    @Usage("List flows that user can start")
     public void list(InvocationContext<String> context) throws Exception {
         logger.info("Executing command \"flow list\".");
         for (String name : ops().registeredFlows()) {
             context.provide(name + System.lineSeparator());
         }
     }
+
+    @Command
+    @Usage("Kill a flow that is running on this node.")
```
szymonsztuka: 2 usage in this class have a `.` at the end of a sentence and 2 others don't, maybe make it consistent with your PR?

chrisr3: :+1:

client/jackson/src/main/kotlin/net/corda/client/jackson/internal/CordaModule.kt
```diff
@@ -418,6 +422,28 @@ private interface SecureHashSHA256Mixin
...
+private interface StateMachineRunIdMixin
+
+private class StateMachineRunIdSerializer : StdScalarSerializer<StateMachineRunId>(StateMachineRunId::class.java) {
+    private val uuidSerializer = UUIDSerializer()
+
+    override fun isEmpty(provider: SerializerProvider?, value: StateMachineRunId): Boolean {
+        return uuidSerializer.isEmpty(provider, value.uuid)
+    }
+
+    override fun serialize(value: StateMachineRunId, gen: JsonGenerator?, provider: SerializerProvider?) {
```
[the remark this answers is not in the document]
chrisr3: It doesn't matter. The `UUIDSerializer` class is clearly expecting the `JsonGenerator` never to be `null` already. However, I'm never dereferencing this object from Kotlin and so there's no harm in just passing the reference from Jackson through to the `UUIDSerializer` instance.

chrisr3:
@szymonsztuka I've now updated the docs, simplified some Kotlin lambdas and extended the smoke tests to cover killing flows.

client/rpc/src/smoke-test/kotlin/net/corda/kotlin/rpc/StandaloneCordaRPClientTest.kt
```diff
@@ -232,6 +235,27 @@ class StandaloneCordaRPClientTest {
...
+            rpcProxy.killFlow(flowHandle.id)
+        }
+    }
+
+    @Test
+    fun `test kill flow with killFlow permission`() {
+        val flowHandle = rpcProxy.startFlow(::CashIssueFlow, 83.DOLLARS, OpaqueBytes.of(0), notaryNodeIdentity)
+        notary.connect(rpcUser).use { connection ->
+            val rpcProxy = connection.proxy
+            assertTrue(rpcProxy.killFlow(flowHandle.id))
```
szymonsztuka: I'm afraid this test can fail (and the previous one) as a flow needs to be running while being killed. But the chances are negligible.

chrisr3: You say that, but... :disappointed:

szymonsztuka: yeah, ideally we could have a flow which runs indefinitely, but it's for another PR

PR #4864, opened 2019-03-08
CORDA-2629: Provide a better error msg when notary type misconfig

lockathan:
If a notary service is misconfigured with the type (either validating or not validating) stored in the node's configuration not matching that advertised in the network map cache, the notary will throw an exception and fail on startup.
Previously, this misconfiguration would result in an exception being thrown when attempting to notarise a transaction. This change results in the exception being thrown at node startup and the node operator being aware of the misconfiguration earlier.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
@@ -75,7 +75,7 @@ class NotaryLoader(
     /** Validates that the notary is correctly configured by comparing the configured type against the type advertised in the network map cache */
     private fun validateNotaryType(myNotaryIdentity: PartyAndCertificate?, services: ServiceHubInternal) {
         var configuredAsValidatingNotary = services.configuration.notary?.validating
-        val notaryParty = myNotaryIdentity?.party ?: throw IllegalStateException("Notary doesn't have a party")
+        val notaryParty = myNotaryIdentity?.party ?: throw IllegalStateException("Could not establish notary identity of this node")
         var validatingNotaryInNetworkMapCache = services.networkMapCache.isValidatingNotary(notaryParty)
         
         if(configuredAsValidatingNotary != validatingNotaryInNetworkMapCache) {
--- a/node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
@@ -79,7 +79,8 @@ class NotaryLoader(
         var validatingNotaryInNetworkMapCache = services.networkMapCache.isValidatingNotary(notaryParty)
         
         if(configuredAsValidatingNotary != validatingNotaryInNetworkMapCache) {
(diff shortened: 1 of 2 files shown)


rick-r3:
I see that we have a test that violates the rules!

node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
```diff
@@ -70,6 +72,17 @@ class NotaryLoader(
         return constructor.newInstance(services, notaryKey)
     }
 
+    /** Validates that the notary is correctly configured by comparing the configured type against the type advertised in the network map cache */
+    private fun validateNotaryType(myNotaryIdentity: PartyAndCertificate?, services: ServiceHubInternal) {
+        var configuredAsValidatingNotary = services.configuration.notary?.validating
+        val notaryParty = myNotaryIdentity?.party ?: throw IllegalStateException("Notary doesn't have a party")
```
rick-r3: I think the exception message should be something more like "Could not establish notary identity of this node,"

node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
```diff
@@ -70,6 +72,17 @@ class NotaryLoader(
...
+        var validatingNotaryInNetworkMapCache = services.networkMapCache.isValidatingNotary(notaryParty)
```
rick-r3: I haven't looked at the start up sequencing, but will there always be an entry for this node by the time this is called?  I would think so but worth trying to reason if that is actually the case.

node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
```diff
@@ -70,6 +72,17 @@ class NotaryLoader(
...
     }
 
+    /** Validates that the notary is correctly configured by comparing the configured type against the type advertised in the network map cache */
+    private fun validateNotaryType(myNotaryIdentity: PartyAndCertificate?, services: ServiceHubInternal) {
+        var configuredAsValidatingNotary = services.configuration.notary?.validating
+        val notaryParty = myNotaryIdentity?.party ?: throw IllegalStateException("Could not establish notary identity of this node")
+        var validatingNotaryInNetworkMapCache = services.networkMapCache.isValidatingNotary(notaryParty)
+        
+        if(configuredAsValidatingNotary != validatingNotaryInNetworkMapCache) {
+            throw IllegalStateException("There is a discrepancy in the configured notary type and the one advertised in the network parameters - shutting down.")
```
adagys: might be worth specifying what the expected/configured type is

node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
lockathan: I investigated this, and it looks like we synchronously load notaries into the network cache by reading them from a file. So, I don't think that sequencing will be an issue.

node/src/main/kotlin/net/corda/node/utilities/NotaryLoader.kt
lockathan: Fixed

PR #4865, opened 2019-03-08
CORDA-2721: Update `djvm check` documentation, and fix CLI tool installation.

chrisr3:
* Fix DJVM CLI installation and runtime scripts.
* Update DJVM documentation to explain about `RuleViolationError`.


mikehearn:
Can you put a note in the constants.properties file that it's not only being read by the docsite conf.py but also this script? If I'd known something was parsing it I'd have fixed it up but nothing makes it clear that this can cause regressions in an untested tool.

PR #4876, opened 2019-03-12
CORDA-2733: Vault trackBy not pushing consumption updates (fixed)

shamsasari:


vkolomeyko:
Looks like unit test: `VaultQueryTests.trackCashStates_consumed` is failing with these changes.

rick-r3 (approved):
Makes sense to me.  Looks like there might be a unit test checking the old behaviour though.

shamsasari:
Actually, that failing test is making me realise this entire PR is invalid. We have a `VaultQueryCriteria.status` criteria option when querying the value. The default is `UNCONSUMED`, which would match the behaviour in the bug report.

I'm going to test against with a query value of `ALL` and see if that "fixes" the issue.

shamsasari:
I can confirm that `vaultTrackByCriteria(Cash.State::class.java, QueryCriteria.VaultQueryCriteria(status = Vault.StateStatus.ALL))` causes the updates to include both consumed and produced states.

Closing.

PR #4883, opened 2019-03-12
RELEASE - Add password reset, cope with all git tickets marked in Jira

fenryka:
Amusingly, if all the tickets mentioned as fixed in Jira are marked in the Git commits we fell over... so yay people getting things right.

Also allows for resetting your password in the keyring if, like me, you mess up typing it you're now no longer locked out

### DIFF
--- a/release-tools/jiraReleaseChecker.py
+++ b/release-tools/jiraReleaseChecker.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/local/bin/python
 
 #-------------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


release-tools/jiraReleaseChecker.py
```diff
@@ -232,6 +238,8 @@ def notInCommit(args_) :
     parser.add_argument("oldTag", help="The previous release tag")
     parser.add_argument("jiraTag", help="The current Jira release")
     parser.add_argument("jiraUser", help="Who to authenticate with Jira as")
+    parser.add_argument("--resetPassword", help="Set flag to allow reseting of password in the keyring",
```
tlil: `resetting` surely :)

fenryka: :D fixed

fenryka:
skipping CI as this is a python script that's not part of CI anyway

PR #4887, opened 2019-03-13
[CORDA-2647] - Prevent registration when previous state exists

dimosr:
## Changes
This changes slightly the registration process, so that it fails fast, in case there's existing state from previous runs (artemis files, DB tables etc.). 

## Testing
* Added some unit tests for the cases introduced
* Also, tested it manually, performing a registration twice (using H2 / postgreSQL) on a node

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
@@ -1,6 +1,7 @@
 package net.corda.node.internal.subcommands
 
 import net.corda.cliutils.CliWrapperBase
+import net.corda.core.internal.VisibleForTesting
 import net.corda.core.internal.createFile
 import net.corda.core.internal.div
 import net.corda.core.internal.exists
@@ -12,10 +13,14 @@ import net.corda.node.internal.NodeStartupLogging.Companion.logger
 import net.corda.node.services.config.NodeConfiguration
 import net.corda.node.utilities.registration.HTTPNetworkRegistrationService
 import net.corda.node.utilities.registration.NodeRegistrationHelper
+import net.corda.nodeapi.internal.persistence.NODE_DATABASE_PREFIX
 import picocli.CommandLine.Mixin
 import picocli.CommandLine.Option
 import java.io.File
+import java.lang.IllegalStateException
 import java.nio.file.Path
+import java.sql.DriverManager
+import java.sql.SQLException
 import java.util.function.Consumer
 
 class InitialRegistrationCli(val startup: NodeStartup): CliWrapperBase("initial-registration", "Start initial node registration with Corda network to obtain certificate from the permissioning server.") {
@@ -58,7 +63,12 @@ class InitialRegistration(val baseDirectory: Path, private val networkRootTrustS
 
     private val nodeRegistration = NodeRegistrationOption(networkRootTrustStorePath, networkRootTrustStorePassword)
 
-    private fun registerWithNetwork(conf: NodeConfiguration) {
+    private val EXISTING_STATE_GENERIC_WARNING = "Please ensure there is no state from previous runs, before initiating registration of a node."
+
+    @VisibleForTesting
+    fun registerWithNetwork(conf: NodeConfiguration) {
+        verifyNoStateFromPreviousRuns(conf)
+
         val versionInfo = startup.getVersionInfo()
 
         println("\n" +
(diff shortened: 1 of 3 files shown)


node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
```diff
@@ -58,7 +63,12 @@ class InitialRegistration(val baseDirectory: Path, private val networkRootTrustS
 
     private val nodeRegistration = NodeRegistrationOption(networkRootTrustStorePath, networkRootTrustStorePassword)
 
-    private fun registerWithNetwork(conf: NodeConfiguration) {
+    private val EXISTING_STATE_GENERIC_WARNING = "Please ensure there is no state from previous runs, before initiating registration of a node."
```
anthonykeenan: Might be better to say something like "Initial registration can only be run on a new, clean node" or something like that?

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
```diff
@@ -82,10 +92,35 @@ class InitialRegistration(val baseDirectory: Path, private val networkRootTrustS
...
     }
 
+    private fun verifyNoStateFromPreviousRuns(conf: NodeConfiguration) {
+        val artemisDirectory = baseDirectory / "artemis"
+        check(!artemisDirectory.exists()) { "The node folder contains an artemis directory. $EXISTING_STATE_GENERIC_WARNING" }
+        val brokersDirectory = baseDirectory / "brokers"
+        check(!brokersDirectory.exists()) { "There node folder contains a brokers directory. $EXISTING_STATE_GENERIC_WARNING" }
+
+        val datasourceProps = conf.dataSourceProperties
+        if (datasourceProps.isEmpty) throw IllegalStateException("There must be a database configured.")
```
anthonykeenan: Will this fail if the node is running H2 with default settings? The data settings will be empty then won't they? Or have they been pre-populated by this point?

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
```diff
@@ -82,10 +92,35 @@ class InitialRegistration(val baseDirectory: Path, private val networkRootTrustS
...
+        val datasourceUrl = datasourceProps.getProperty("dataSource.url")
+        val datasourceUser = datasourceProps.getProperty("dataSource.user")
```
anthonykeenan: Is the User/Pass always specified seperately? If the user/pass is in the Url will this pass?
Is there some data connection code we can re-use here?

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
dimosr: yes, I like it, clear and concise. I changed it to this one.

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
dimosr: No, the settings have been pre-populated with the default ones by this point. I did a manual run with an empty config just in case to verify it.

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
dimosr: Good point, looking at the docs of `DriverManager` it seems there is some ambiguity in some cases, where the behaviour is implementation-specific.

Anyway, after a bit of a search, I found out `DataSourceFactory`, which is used across the codebase to initiate a datasource connection, so I used this one instead. That way, this part is consistent with the rest of the system's behaviour.

node/src/test/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCliTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        private val nodeStartup = mock(NodeStartup::class.java)
+
+        private val baseDirectory = Files.createTempDirectory("base-dir")!!
+        private val networkTrustRootFile = Files.createTempFile("trust-root-store", "jks")
+
+        private lateinit var nodeConfiguration: NodeConfiguration
+        private lateinit var datasourceProperties: Properties
+        private lateinit var node: Node
+
+        private val h2jdbcUrl = "jdbc:h2:tcp://localhost:10009/~/node"
```
anthonykeenan: Can we change this to use a port allocated using the global port allocator?

node/src/test/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCliTest.kt
```diff
@@ -0,0 +1,98 @@
...
+            executeSqlStatement("CREATE TABLE NODE_ATTACHMENTS(USERNAME VARCHAR(20));")
+
+            initialRegistration = InitialRegistration(baseDirectory, networkTrustRootFile, networkTrustRootPassword, nodeStartup)
+        }
+
+        @AfterClass
+        @JvmStatic
+        fun cleanup() {
+            baseDirectory.deleteRecursively()
+            networkTrustRootFile.deleteRecursively()
```
anthonykeenan: Could you also run the test on a Windows server? I think you should be able to trigger a Windows CI build via this TC configuration: https://ci-master.corda.r3cev.com/viewType.html?buildTypeId=Corda_Build_BuildWindowsBranch

Deleting file/directory errors is the number 1 windows build breakage reason.

node/src/test/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCliTest.kt
dimosr: I've switched to the global port allocator. I also moved the test to the integration tests folder, since it's more of an integration test and this way it will also benefit from the global port allocation configured for all integ tests.

PR #4889, opened 2019-03-13
ENT-3057: Log hibernate warns and errors in different log

lockathan:
If a hibernate error occurs (deadlock, for example) that would cause a flow to be sent to the hospital, hibernate logs the warnings and errors before we do. This results in duplication in the logs, and pollutes the log. To solve this, we create a new log appender named diagnostic-{node-name}.log and log any org.hibernate messages of warn and above to that file. This way, messages are not lost, which means that the information can be retrieved if need be.

### DIFF
--- a/config/dev/log4j2.xml
+++ b/config/dev/log4j2.xml
@@ -4,6 +4,7 @@
     <Properties>
         <Property name="log-path">${sys:log-path:-logs}</Property>
         <Property name="log-name">node-${hostName}</Property>
+        <Property name="diagnostic-log-name">diagnostic-${hostName}</Property>
         <Property name="archive">${log-path}/archive</Property>
         <Property name="defaultLogLevel">${sys:defaultLogLevel:-info}</Property>
         <Property name="consoleLogLevel">${sys:consoleLogLevel:-error}</Property>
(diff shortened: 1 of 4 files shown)


config/dev/log4j2.xml
```diff
@@ -103,6 +104,46 @@
                 </Delete>
             </DefaultRolloverStrategy>
 
+        </RollingRandomAccessFile>
+		
+		<!-- Will generate up to 100 log files for a given day. During every rollover it will delete
```
rick-r3: The indentation here has gone a bit squiffy.  Can we get it indented as per the appender above?

rick-r3:
Indentation looks better

config/dev/log4j2.xml
lockathan: I updated the node-administration.rst, since it seems to relate to logging. Seems the same between Open Source and Enterprise. I added a sentence describing the change - let me know what you think.

chrisr3: I don't know about either of you, but I'm thinking
> Why is the Node's application-specific Log4J2 configuration file located in a directory called `config/dev`?

Answers on a postcard to...?

docs/source/node-administration.rst
```diff
@@ -10,7 +10,8 @@ Logging
 By default the node log files are stored to the ``logs`` subdirectory of the working directory and are rotated from time
 to time. You can have logging printed to the console as well by passing the ``--log-to-console`` command line flag.
 The default logging level is ``INFO`` which can be adjusted by the ``--logging-level`` command line argument. This configuration
-option will affect all modules.
+option will affect all modules. Hibernate specific log messages of level ``WARN`` and above will be logged to the diagnostic log
```
rick-r3: I think it is worth saying why we do this, so that there's no temptation to set up monitoring on the diagnostics file.  Perhaps something like:

`Hibernate (the JPA provider used by Corda) specific log messages of level ``WARN`` and above will be logged to the diagnostic log file, which is stored in the same location as other log files (``logs`` subdirectory by default).  This is because Hibernate may log messages at WARN and ERROR that are handled internally by Corda and do not need operator attention.  If they do, they will be logged by Corda itself in the main node log file.`

lockathan: Thanks I've implemented your suggested docs change.

config/dev/log4j2.xml
rick-r3: Yeah, it's a bit of a fail.  The idea was that would be used in development and then there would be a `config/prod` for real environments, but that idea got lost along the way...

PR #4892, opened 2019-03-14
CORDA-2745: Cache notary identity lookups

adagys:
Add a cache for notary identities in the PersistentIdentityService. This
solves a reported problem where notary identity lookup fails if its network
map entry is missing, which results in an exception when trying to insert
a state into the vault after recording a transaction.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
+++ b/node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
@@ -102,7 +102,7 @@ class PersistentIdentityService(cacheFactory: NamedCacheFactory) : SingletonSeri
     override val trustAnchor: TrustAnchor get() = _trustAnchor
 
     /** Stores notary identities obtained from the network parameters, for which we don't need to perform a database lookup. */
-    private lateinit var notaryIdentityCache: Set<Party>
+    private val notaryIdentityCache = HashSet<Party>()
 
     // CordaPersistence is not a c'tor parameter to work around the cyclic dependency
     lateinit var database: CordaPersistence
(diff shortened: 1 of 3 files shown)


node/src/main/kotlin/net/corda/node/services/identity/PersistentIdentityService.kt
```diff
@@ -170,7 +174,16 @@ class PersistentIdentityService(cacheFactory: NamedCacheFactory) : SingletonSeri
 
     override fun wellKnownPartyFromX500Name(name: CordaX500Name): Party? = certificateFromCordaX500Name(name)?.party
 
-    override fun wellKnownPartyFromAnonymous(party: AbstractParty): Party? = database.transaction { super.wellKnownPartyFromAnonymous(party) }
+    override fun wellKnownPartyFromAnonymous(party: AbstractParty): Party? {
+        // Skip database lookup if the party is a notary identity.
+        // This also prevents an issue where the notary identity can't be resolved if it's not in the network map cache. The node obtains
+        // a trusted list of notary identities from the network parameters automatically.
+        return if (party is Party && party in notaryIdentityCache) {
```
adagys: this extra lookup should be realyl cheap. Also we perform an identity lookup for a notary on every state we insert into the vault, so the net effect might even be a slight perf improvement.

rick-r3:
We restart on a network parameters change, so don't need to worry about that presumably?  Any complexities with clustered notary identities ?

Otherwise looking good.

adagys:
Yeah, as long as any change to the notaries list requiress restart it should be fine. A clustered notary identity is still a Party with a PublicKey so it shouldn't make a difference

PR #4893, opened 2019-03-14
CORDA-2747: Package DJVM CLI for delivery.

chrisr3:
* Modify DJVM CLI so that it runs as executable JAR.
 * Package DJVM CLI as a distributable ZIP.
 * Publish DJVM CLI.

### DIFF
--- a/djvm/cli/src/shell/install
+++ b/djvm/cli/src/shell/install
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$(dirname $(readlink -f ${BASH_SOURCE[0]}))
+
+# Generate auto-completion file for Bash and ZSH
+java -cp ${SCRIPT_DIR}/corda-djvm-cli.jar \
+    picocli.AutoComplete -n djvm net.corda.djvm.tools.cli.Commands -f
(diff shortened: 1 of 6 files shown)


djvm/cli/src/shell/install
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$(dirname $(readlink -f ${BASH_SOURCE[0]}))
+
+# Generate auto-completion file for Bash and ZSH
+java -cp ${SCRIPT_DIR}/corda-djvm-cli.jar \
```
tlil: Should we make this an option for the main executable? `djvm --install-shell-extensions` or something similar to trigger this.

chrisr3: The comment implies that it's specific to `bash` or `zsh`, so I'm only really providing the `install` script to align this ZIP artifact with the existing documentation.

For reference, I'm currently trying to write `djvm.bat` so that Anthony has more work to do :smiling_imp:.

tlil: Lol, yeah, can do that as a follow-up anyway.

chrisr3: OK - but I'd like to add the `djvm.bat` before merging please.

tlil: Yes, sure - I was referring to the `install-shell-extensions` bit. Provisionally approved. Ping me when your 🦇  is ready ;)

PR #4894, opened 2019-03-14
[CORDA-2738] Allow the ProgressTracker to cope with child trackers with the same steps

JamesHR3:
This PR is a set of robustness enhancements to the `ProgressTracker` and `AnsiProgressRenderer` that should mean these report the correct steps as being executed under more circumstances. The main fix is to allow multiple child trackers to have the same steps, which happens in, for example, the `CashIssueAndPaymentFlow`. There are also a few other changes to try and remove other sources of bugs, and a few commenting improvements and removal of unneeded code.

Changes:
 - Fixed the calculation of the `stepsTreeIndex` in `ProgressTracker` to account for the same step label existing under multiple top level steps.
 - Fixed the recalculation of indices seen in the `AnsiProgressRenderer` to account for the same step label possibly existing at multiple points in the step tree.
 - Fixed an issue where index and tree updates could race, such that if an index and tree update happened at about the same time the index update could end up applying to either the old or new tree in a non-deterministic way. The two observables are now joined into a single one on the renderer end, such that one update will be fully processed before another starts even if the update types are interleaved.
 - Removed some unneeded properties from the `ProgressTracker`.
 - Moved a comment back to the correct location.
 - Update `CashIssueAndPaymentFlow` to have top-level steps. This prevents an issue where the issuance progress tracker is overwritten by the payment tracker, but as they have the same steps the payment tracker has steps prematurely marked as complete. Unfortunately there is no way of solving this within the progress tracking and rendering code, as there is no way to tell between "tracker has been entirely removed and replaced" and "tracker has been amended by adding a new child".
 - Added unit tests covering fixed functionality.

Testing:
 - Unit tests
 - Multiple manual runs of cash issue, cash payment, and cash issue and payment flows to ensure the output looks sane and is repeatable.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
+++ b/core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
@@ -233,9 +233,23 @@ class ProgressTracker(vararg inputSteps: Step) {
         recalculateStepsTreeIndex()
     }
 
+    private fun getStepIndexAtLevel(level: Int): Int {
+        return if (level > 0) {
+            if (stepIndex - 2 >= 0) stepIndex - 2 else 0
+        } else {
+            stepIndex
+        }
+    }
+
+    private fun getCurrentStepTreeIndex(level: Int = 0): Int {
+        val newLevel = level + 1
+        val indexAtLevel = getStepIndexAtLevel(level)
+        val additionalIndex = getChildProgressTracker(currentStep)?.getCurrentStepTreeIndex(newLevel) ?: 0
+        return indexAtLevel + additionalIndex
+    }
+
     private fun recalculateStepsTreeIndex() {
-        val step = currentStepRecursiveWithoutUnstarted()
-        stepsTreeIndex = _allStepsCache.indexOfFirst { it.second == step }
+        stepsTreeIndex = getCurrentStepTreeIndex()
     }
 
     private fun _allSteps(level: Int = 0): List<Pair<Int, Step>> {
--- a/core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
+++ b/core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
@@ -50,7 +50,9 @@ class ProgressTracker(vararg inputSteps: Step) {
         }
     }
 
-    /** The superclass of all step objects. */
+    /**
+     * The superclass of all step objects.
+     */
     @CordaSerializable
     open class Step(open val label: String) {
         open val changes: Observable<Change> get() = Observable.empty()
@@ -84,7 +86,9 @@ class ProgressTracker(vararg inputSteps: Step) {
 
     private val childProgressTrackers = mutableMapOf<Step, Child>()
 
-    /** The steps in this tracker, same as the steps passed to the constructor but with UNSTARTED and DONE inserted. */
+    /**
+     * The steps in this tracker, same as the steps passed to the constructor but with UNSTARTED and DONE inserted.
+     */
     val steps = arrayOf(UNSTARTED, STARTING, *inputSteps, DONE)
 
     private var _allStepsCache: List<Pair<Int, Step>> = _allSteps()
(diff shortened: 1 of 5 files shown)


shamsasari:
I'd rather this be punted after CE4 is released. It can go into C4.2. It doesn't seem to be a high priority bug, unless you think otherwise.

JamesHR3:
My concern was that CORDA-2737 had made things worse with the `ProgressTracker` by exposing some weirdness with the `CashIssueAndPaymentFlow` tracking, and I wanted to tidy it up. I think it's probably ok to delay this though, I don't think it's urgently broken.

JamesHR3:
Actually, we might want this - it seems that some steps are being consistently skipped in the progress tracker on CE4.0-RC05, which I think this fixes. I think CORDA-2737 introduced this - all step tree updates are pushed from the beginning when the observable is subscribed to, which means the snapshot is unhelpful (the first update may send you backwards in time), but I think the initial state isn't pushed in the observable. This fix does do that.

shamsasari:
Fair enough, I'll take a look. 

The API Stability check is failing.

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -217,7 +209,10 @@ class ProgressTracker(vararg inputSteps: Step) {
     var parent: ProgressTracker? = null
         private set
 
-    /** Walks up the tree to find the top level tracker. If this is the top level tracker, returns 'this' */
+    /** Walks up the tree to find the top level tracker. If this is the top level tracker, returns 'this'.
+     *  Required for API compatibility.
+     */
+    @Deprecated("topLevelTracker is not expected to be used by external clients.")
```
shamsasari: Can you explain why

tools/shell/src/main/kotlin/net/corda/tools/shell/utlities/ANSIProgressRenderer.kt
```diff
@@ -75,27 +78,60 @@ abstract class ANSIProgressRenderer {
...
-                    tree = it
-                    draw(true)
-                }, { done(it) }, { done(null) })
+        if (treeUpdates == null || indexUpdates == null) {
+            renderInBold("Cannot print progress for this flow as the required data is missing", Ansi())
+        } else {
+            // By combining the two observables, a race condition where both emit items at roughly the same time is avoided. This could
+            // result in steps being incorrectly marked as skipped. Instead, whenever either observable emits an item, a pair of the
+            // last index and last tree is returned, which ensures that updates to either are processed in series.
+            updatesSubscription = combineLatest(treeUpdates, indexUpdates) {tree, index -> Pair(tree, index)}.subscribe({
```
shamsasari: To make this a bit easier to read, can you move the `{` to its own line and have the last two `done` lambas on separate lines.

tools/shell/src/main/kotlin/net/corda/tools/shell/utlities/ANSIProgressRenderer.kt
```diff
@@ -75,27 +78,60 @@ abstract class ANSIProgressRenderer {
...
+                    tree = newTree
+                }
+                draw(true)
+            }, { done(it) }, {done(null)})
+        }
+    }
+
+    // Create a new tree of steps that also holds a reference to the parent of each step. This is required to uniquely identify each step
+    // (assuming that each step label is unique at a given level).
+    private fun transformTree(inputTree: List<Pair<Int, String>>): StepsTree {
```
shamsasari: Can you introduce a new class to represent this list of pairs

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
```diff
@@ -159,27 +166,12 @@ class ProgressTracker(vararg inputSteps: Step) {
             }
         }
 
-    /**
-     * Reading returns the value of steps[stepIndex], writing moves the position of the current tracker. Once moved to
-     * the [DONE] state, this tracker is finished and the current step cannot be moved again.
-     */
-
     /** Returns the current step, descending into children to find the deepest step we are up to. */
+    @Deprecated("currentStepRecursive should not be used by external clients")
```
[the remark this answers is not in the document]
JamesHR3: I'm finding it a little tricky to describe this concisely. The problem with this property is that it's effectively internal (it was used to calculate what the current step index is), but ended up being public. If this information is required, it can be obtained by subscribing to the progress observable instead. I've changed the deprecation message to say this ("The current step should be obtained by subscribing to the progress observable").

core/src/main/kotlin/net/corda/core/utilities/ProgressTracker.kt
JamesHR3: Again, slightly tricky - this shouldn't really have ended up as part of the public API (I think). I've said "Obtaining tracker tree structure information should be done using the stepsTreeChanges observable".

shamsasari:
Those deprecated methods return `Step` objects which the alternatives don't. I think it be better to not deprecate, at least not in this PR.

PR #4895, opened 2019-03-14
CORDA-2506: Better handling of invalid log path

lockathan:
Test if we have access to the logging path (baseDirectory/logs) before attempting to write to them. This allows us to shut down gracefully with an easily understandable error message.
Without doing this, the log4j2 will attempt to access the logPath when it first uses the logger in the call() function.
This does involve changing the signature of initiLogging to return a Boolean. This allows call() to know whether initLogging was successful and hence either continue execution or shut down.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -72,7 +72,7 @@ open class NodeStartupCli : CordaCliWrapper("corda", "Runs a Corda Node") {
     private val initialRegistrationCli by lazy { InitialRegistrationCli(startup) }
     private val validateConfigurationCli by lazy { ValidateConfigurationCli() }
 
-    override fun initLogging():Boolean = this.initLogging(cmdLineOptions.baseDirectory)
+    override fun initLogging(): Boolean = this.initLogging(cmdLineOptions.baseDirectory)
 
     override fun additionalSubCommands() = setOf(networkCacheCli, justGenerateNodeInfoCli, justGenerateRpcSslCertsCli, initialRegistrationCli, validateConfigurationCli)
 
@@ -438,7 +438,7 @@ interface NodeStartupLogging {
     }
 }
 
-fun CliWrapperBase.initLogging(baseDirectory: Path) : Boolean {
+fun CliWrapperBase.initLogging(baseDirectory: Path): Boolean {
     System.setProperty("defaultLogLevel", specifiedLogLevel) // These properties are referenced from the XML config file.
     if (verbose) {
         System.setProperty("consoleLoggingEnabled", "true")
@@ -448,19 +448,17 @@ fun CliWrapperBase.initLogging(baseDirectory: Path) : Boolean {
 
     //Test for access to the logging path and shutdown if we are unable to reach it.
     val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
-    try{
+    try {
         logPath.createDirectories()
-    }
-    catch(e: IOException){        
-        println(ShellConstants.RED + "Unable to create logging directory ${logPath.toString()}. Node will now shutdown." + ShellConstants.RESET)
+    } catch (e: IOException) {
+        println("$ShellConstants.RED Unable to create logging directory ${logPath.toString()}. Node will now shutdown.$ShellConstants.RESET")
+        return false
+    } catch (e: SecurityException) {
+        println("$ShellConstants.RED Current user is unable to access logging directory ${logPath.toString()}. Node will now shutdown.$ShellConstants.RESET")
         return false
     }
-    catch(e: SecurityException){
-        println(ShellConstants.RED + "Current user is unable to access logging directory ${logPath.toString()}. Node will now shutdown." + ShellConstants.RESET)
-        return false
-    }
-    if(!logPath.exists()) {
-        println(ShellConstants.RED + "Unable to access logging directory ${logPath.toString()}. Node will now shutdown." + ShellConstants.RESET)
+    if (!logPath.isDirectory()) {
+        println("$ShellConstants.RED Unable to access logging directory ${logPath.toString()}. Node will now shutdown.$ShellConstants.RESET")
         return false
     }
 
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -37,6 +37,7 @@ import java.io.RandomAccessFile
 import java.lang.management.ManagementFactory
 import java.net.InetAddress
 import java.nio.channels.UnresolvedAddressException
+import java.io.File
 import java.nio.file.Path
 import java.time.DayOfWeek
 import java.time.ZonedDateTime
@@ -53,7 +54,7 @@ abstract class NodeCliCommand(alias: String, description: String, val startup: N
         const val LOGS_DIRECTORY_NAME = "logs"
     }
 
-    override fun initLogging() = this.initLogging(cmdLineOptions.baseDirectory)
+    override fun initLogging(): Boolean = this.initLogging(cmdLineOptions.baseDirectory)
 
     @Mixin
     val cmdLineOptions = SharedNodeCmdLineOptions()
@@ -71,7 +72,7 @@ open class NodeStartupCli : CordaCliWrapper("corda", "Runs a Corda Node") {
     private val initialRegistrationCli by lazy { InitialRegistrationCli(startup) }
     private val validateConfigurationCli by lazy { ValidateConfigurationCli() }
 
-    override fun initLogging() = this.initLogging(cmdLineOptions.baseDirectory)
+    override fun initLogging():Boolean = this.initLogging(cmdLineOptions.baseDirectory)
 
     override fun additionalSubCommands() = setOf(networkCacheCli, justGenerateNodeInfoCli, justGenerateRpcSslCertsCli, initialRegistrationCli, validateConfigurationCli)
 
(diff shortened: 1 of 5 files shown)


shamsasari:
Please make sure the correct Jira ticket reference is used in the PR title. Also include a short summary after it.

Feel free to change I've made to the summary if incorrect.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
```diff
@@ -123,12 +124,13 @@ abstract class CliWrapperBase(val alias: String, val description: String) : Call
...
     // This needs to be called before loggers (See: NodeStartup.kt:51 logger called by lazy, initLogging happens before).
     // Node's logging is more rich. In corda configurations two properties, defaultLoggingLevel and consoleLogLevel, are usually used.
-    open fun initLogging() {
+    open fun initLogging() : Boolean {
         System.setProperty("defaultLogLevel", specifiedLogLevel) // These properties are referenced from the XML config file.
         if (verbose) {
             System.setProperty("consoleLogLevel", specifiedLogLevel)
         }
         System.setProperty("log-path", Paths.get(".").toString())
+        return true
```
shamsasari: This can only return true so the boolean return is not needed.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -71,7 +72,7 @@ open class NodeStartupCli : CordaCliWrapper("corda", "Runs a Corda Node") {
     private val initialRegistrationCli by lazy { InitialRegistrationCli(startup) }
     private val validateConfigurationCli by lazy { ValidateConfigurationCli() }
 
-    override fun initLogging() = this.initLogging(cmdLineOptions.baseDirectory)
+    override fun initLogging():Boolean = this.initLogging(cmdLineOptions.baseDirectory)
```
shamsasari: Space after colon

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -437,14 +438,34 @@ interface NodeStartupLogging {
...
         System.setProperty("consoleLogLevel", specifiedLogLevel)
         Node.renderBasicInfoToConsole = false
     }
+
+    //Test for access to the logging path and shutdown if we are unable to reach it.
+    val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
+    try{
+        logPath.createDirectories()
+    }
+    catch(e: IOException){
```
shamsasari: Our coding guideline is to have the `else` and `catch` blocks on the same line as the preceding `}`
```
} catch (e: IOException) {
```

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -437,14 +438,34 @@ interface NodeStartupLogging {
...
     System.setProperty("defaultLogLevel", specifiedLogLevel) // These properties are referenced from the XML config file.
     if (verbose) {
         System.setProperty("consoleLoggingEnabled", "true")
         System.setProperty("consoleLogLevel", specifiedLogLevel)
         Node.renderBasicInfoToConsole = false
     }
+
+    //Test for access to the logging path and shutdown if we are unable to reach it.
+    val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
+    try{
```
shamsasari: Space before `{`

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -437,14 +438,34 @@ interface NodeStartupLogging {
...
+        logPath.createDirectories()
+    }
+    catch(e: IOException){        
+        println(ShellConstants.RED + "Unable to create logging directory ${logPath.toString()}. Node will now shutdown." + ShellConstants.RESET)
```
shamsasari: Use string templates

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -437,14 +438,34 @@ interface NodeStartupLogging {
     }
 }
 
-fun CliWrapperBase.initLogging(baseDirectory: Path) {
+fun CliWrapperBase.initLogging(baseDirectory: Path) : Boolean {
```
shamsasari: No space before the `:`

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
```diff
@@ -30,7 +30,7 @@ class InitialRegistrationCli(val startup: NodeStartup): CliWrapperBase("initial-
         return startup.initialiseAndRun(cmdLineOptions, InitialRegistration(cmdLineOptions.baseDirectory, networkRootTrustStorePath, networkRootTrustStorePassword, startup))
     }
 
-    override fun initLogging() = this.initLogging(cmdLineOptions.baseDirectory)
+    override fun initLogging() : Boolean = this.initLogging(cmdLineOptions.baseDirectory)
```
shamsasari: ditto

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -437,14 +438,34 @@ interface NodeStartupLogging {
...
+    }
+    catch(e: IOException){        
+        println(ShellConstants.RED + "Unable to create logging directory ${logPath.toString()}. Node will now shutdown." + ShellConstants.RESET)
+        return false
+    }
+    catch(e: SecurityException){
+        println(ShellConstants.RED + "Current user is unable to access logging directory ${logPath.toString()}. Node will now shutdown." + ShellConstants.RESET)
+        return false
+    }
+    if(!logPath.exists()) {
```
shamsasari: Use `isDirectory()` instead

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
kasiastreich: space

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
kasiastreich: Wrong formatting, both space and where catch is. The easiest is to do autofromat on those lines.

tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt
lockathan: I can't remove this, because the return type is used in the CordaCliWrapper.call() method. If I remove the Boolean return type here, I can't use it in call(). So, even though only the NodeStartup implementation of initLogging actually uses the return type, unfortunately I had to modify the signature of the base declaration and all implementations of it.

shamsasari: Understood, I didn't see that it was needed in the super class.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -437,14 +438,32 @@ interface NodeStartupLogging {
...
         System.setProperty("consoleLogLevel", specifiedLogLevel)
         Node.renderBasicInfoToConsole = false
     }
+
+    //Test for access to the logging path and shutdown if we are unable to reach it.
+    val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
+    try {
+        logPath.createDirectories()
+    } catch (e: IOException) {
+        println("$ShellConstants.RED Unable to create logging directory ${logPath.toString()}. Node will now shutdown.$ShellConstants.RESET")
```
shamsasari: You can avoid the leading space by doing `"${$ShellConstants.RED}Unable`

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
lockathan: Fixed and removed the leading space. Please have a look.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
lockathan: Fixed.

shamsasari:
The unit tests are failing due to a compile error

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
anthonykeenan: There is a convenience function for printing errors:
https://github.com/corda/corda/blob/master/tools/cliutils/src/main/kotlin/net/corda/cliutils/CordaCliWrapper.kt#L192

shamsasari: Ah, that's good to know.

lockathan: Switched to using the helper function

lockathan:
> The unit tests are failing due to a compile error

Argh! Sorry! Fixed it.

tools/shell-cli/src/main/kotlin/net/corda/tools/shell/StandaloneShell.kt
```diff
@@ -56,10 +56,11 @@ class StandaloneShell : CordaCliWrapper("corda-shell", "The Corda standalone she
 
     private fun getManifestEntry(key: String) = if (Manifests.exists(key)) Manifests.read(key) else "Unknown"
 
-    override fun initLogging() {
+    override fun initLogging() : Boolean {
         super.initLogging()
```
shamsasari: Shouldn't we return false here if the super call returns false?

shamsasari: No, it doesn't matter as the super call always returns true.

PR #4896, opened 2019-03-14
CORDA-2748 Always set the ThreadLocal in the Fiber from the Thread, even if not yet set in the Thread.

rick-r3:


rick-r3:
Going to take it for a workout to check the original memory leak remains fixed and no further issues.

rick-r3:
I'm happy that after reviewing memory leak related profiling reports the memory leak is still fixed.

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -170,6 +169,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
             }
         } finally {
             checkDbTransaction(isDbTransactionOpenOnExit)
+            openThreadLocalWormhole()
```
rick-r3: I decided it was a mistake not to also do this on the code path for the `Throw`.

rick-r3:
This is a logic fix, with no regression, on that basis.

PR #4897, opened 2019-03-15
CORDA-2750: Modularise the DJVM build.

chrisr3:
The DJVM code is independent of Corda and requires its own build / release cycle.


djvm/build.gradle
```diff
@@ -1,91 +1,132 @@
...
+    }
 
-repositories {
-    maven {
-        url "$artifactory_contextUrl/corda-dev"
+    repositories {
+        mavenCentral()
+        jcenter()
+        maven {
+            url "$artifactory_contextUrl/corda-releases"
```
Clintonio: What is the purpose for depending on this repository? If it is for the gradle plugins then you should only depend on the released versions in Jcenter/maven central.

chrisr3: Ah, if only...
```
A problem occurred configuring root project 'determinisic-jvm-sandbox'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not find net.corda.plugins:publish-utils:4.0.42.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/net/corda/plugins/publish-utils/4.0.42/publish-utils-4.0.42.pom
       - https://repo.maven.apache.org/maven2/net/corda/plugins/publish-utils/4.0.42/publish-utils-4.0.42.jar
       - https://jcenter.bintray.com/net/corda/plugins/publish-utils/4.0.42/publish-utils-4.0.42.pom
       - https://jcenter.bintray.com/net/corda/plugins/publish-utils/4.0.42/publish-utils-4.0.42.jar
       - https://plugins.gradle.org/m2/net/corda/plugins/publish-utils/4.0.42/publish-utils-4.0.42.pom
     Required by:
         project :
```

Clintonio:
Please fully break the dependency on Corda by not using constants.properties. Your build will experience breaks if Corda changes. This introduces a more complex build cycle than a full split or fully joined project and is the worst of both worlds. We did this for the gradle plugins, as you know, and it has worked fine in the last two years since.

Clintonio:
Although keeping the shared Kotlin version does make sense, for now.

djvm/build.gradle
chrisr3: Fixed :+1:

chrisr3:
*Sigh* The installer scripts still need to be updated... I will fix...

PR #4899, opened 2019-03-18
CORDA-2759: Use GlobalTestPortAllocation for Node's integration tests.

chrisr3:
Node's integration tests are repeatedly creating servers that listen on a small range of ports starting from 10,000. If a previous test's servers don't shut down quickly enough then the next test can accidentally connect to one of them, and then Chaos Happens.

Chaos is Bad. Avoid Chaos by incrementing port allocations globally across all of the Node's integration tests so that lingering servers will be ignored.


chrisr3:
That's OK - there was no documentation to change to begin with!

josecoll:
I think we need to document this somewhere - 
system variables used within code are dangerous without docs.

PR #4900, opened 2019-03-18
BUILD: Try to fix flaky FailNodeOnNotMigratedAttachmentContractsTableNameTests.

chrisr3:
* Ensure that this test stops nodes that it starts.
* Also ensure that random ephemeral ports can be reused.

It looks like the test doesn't fail per-se. Just that Gradle is left waiting indefinitely for it to complete properly, and then Team City kills the build when it eventually times out. This sounds like something isn't being closed down properly - try putting `nodeHandle.stop()` inside a `finally` block for starters. And setting `reuseAddress` on the ephemeral port that we want to grab can't hurt either.


anthonykeenan:
LGTM

PR #4901, opened 2019-03-18
 Create hibernate_sequence database sequence object only if it doesn't exist

alexeykoren:
Liquibase initial migration shouldn't fail if hibernate_sequence is already created in the target database. To avoid this we use 'preConditions' check.

My PR is in accordance with the [Developer's Certificate of Origin](https://github.com/corda/corda/blob/master/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/node/src/main/resources/migration/common.changelog-init.xml
+++ b/node/src/main/resources/migration/common.changelog-init.xml
@@ -5,11 +5,17 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
 
     <changeSet author="R3.Corda" id="1511451595465-1.1" dbms="h2">
+        <preConditions onFail="MARK_RAN">
+            <not><sequenceExists sequenceName="hibernate_sequence"/></not>
+        </preConditions>
         <createSequence sequenceName="hibernate_sequence"/>
     </changeSet>
 
    <changeSet author="R3.Corda" id="1511451595465-1.3" onValidationFail="MARK_RAN">
-       <preConditions onFail="MARK_RAN" onSqlOutput="TEST">  <not> <dbms type="h2"/> </not> </preConditions>
+        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
+            <not><sequenceExists sequenceName="hibernate_sequence"/></not>
+            <not><dbms type="h2"/></not>
+        </preConditions>
         <createSequence sequenceName="hibernate_sequence" minValue="1"/>
     </changeSet>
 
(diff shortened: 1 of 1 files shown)


node/src/main/resources/migration/common.changelog-init.xml
```diff
@@ -5,11 +5,17 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
 
     <changeSet author="R3.Corda" id="1511451595465-1.1" dbms="h2">
+        <preConditions onFail="MARK_RAN">
```
szymonsztuka: The best is to add `onSqlOutput="TEST"` to `preConditions`, so this is properly handles when generating DDL script as well.

node/src/main/resources/migration/common.changelog-init.xml
```diff
@@ -5,11 +5,17 @@
...
+            <not><sequenceExists sequenceName="hibernate_sequence"/></not>
+        </preConditions>
         <createSequence sequenceName="hibernate_sequence"/>
     </changeSet>
 
    <changeSet author="R3.Corda" id="1511451595465-1.3" onValidationFail="MARK_RAN">
-       <preConditions onFail="MARK_RAN" onSqlOutput="TEST">  <not> <dbms type="h2"/> </not> </preConditions>
+        <preConditions onFail="MARK_RAN" onSqlOutput="TEST">
+            <not><sequenceExists sequenceName="hibernate_sequence"/></not>
+            <not><dbms type="h2"/></not>
```
szymonsztuka: Just in case I would add then new contain after `<dbms>` check.

node/src/main/resources/migration/common.changelog-init.xml
alexeykoren: Done

node/src/main/resources/migration/common.changelog-init.xml
alexeykoren: Good point, done

alexeykoren:
@szymonsztuka I tried with Postgres 11.2 and Liquibase seems to ignore this precondition and it fails. From the other hand for Postgres you can use <sql>CREATE SEQUENCE IF NOT EXISTS hibernate_sequence;</sql> and hope the best for others. I tried and it works. Do you think this is a good approach?

szymonsztuka:
How does it fail, e.g it still tries to create a sequence and it fails? If yes, it could be some problem with schema namespaces, could you try on a database which has only a one schema for a Corda node?

alexeykoren:
> How does it fail, e.g it still tries to create a sequence and it fails? If yes, it could be some problem with schema namespaces, could you try on a database which has only a one schema for a Corda node?

I double-checked - that was my fancy docker setup, it picked up wrong version of corda.jar. Everything works fine. So I'm pushing your suggestions and I guess it's ready for review after this

mikehearn:
Thanks for the contribution!

PR #4905, opened 2019-03-20
 ENT-3327: Check for missing certificates directory

lockathan:
While the node is starting up, we now check for the presence of the certificates directory. This allows us to print out an easily understandable error message if the directory is not present. An exception is made for devMode, as devMode will result in the directory being created in any case.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -122,7 +122,6 @@ open class NodeStartup : NodeStartupLogging {
         private val logger by lazy { loggerFor<Node>() } // I guess this is lazy to allow for logging init, but why Node?
         const val LOGS_DIRECTORY_NAME = "logs"
         const val LOGS_CAN_BE_FOUND_IN_STRING = "Logs can be found in"
-        const val CERTIFICATES_DIRECTORY_NAME = "certificates"
     }
 
     lateinit var cmdLineOptions: SharedNodeCmdLineOptions
@@ -152,7 +151,7 @@ open class NodeStartup : NodeStartupLogging {
         val configuration = cmdLineOptions.parseConfiguration(rawConfig).doIfValid { logRawConfig(rawConfig) }.doOnErrors(::logConfigurationErrors).optional ?: return ExitCodes.FAILURE
        
         // Step 6. Check if we can access the certificates directory
-        if (!canReadCertificatesDirectory(cmdLineOptions.baseDirectory, configuration.devMode)) return ExitCodes.FAILURE
+        if (!canReadCertificatesDirectory(configuration.certificatesDirectory, configuration.devMode)) return ExitCodes.FAILURE
 
         // Step 7. Configuring special serialisation requirements, i.e., bft-smart relies on Java serialization.
         if (attempt { banJavaSerialisation(configuration) }.doOnFailure(Consumer { error -> error.logAsUnexpected("Exception while configuring serialisation") }) !is Try.Success) return ExitCodes.FAILURE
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -122,6 +122,7 @@ open class NodeStartup : NodeStartupLogging {
         private val logger by lazy { loggerFor<Node>() } // I guess this is lazy to allow for logging init, but why Node?
         const val LOGS_DIRECTORY_NAME = "logs"
         const val LOGS_CAN_BE_FOUND_IN_STRING = "Logs can be found in"
+        const val CERTIFICATES_DIRECTORY_NAME = "certificates"
```
kasiastreich: I think that we can have certificates directory in configuration and it can be a different path that the one you are using (I am not sure if it can be provided in config, it's worth checking how it works).

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -149,17 +150,20 @@ open class NodeStartup : NodeStartupLogging {
         // Step 5. Load and validate node configuration.
         val rawConfig = cmdLineOptions.rawConfiguration().doOnErrors(cmdLineOptions::logRawConfigurationErrors).optional ?: return ExitCodes.FAILURE
         val configuration = cmdLineOptions.parseConfiguration(rawConfig).doIfValid { logRawConfig(rawConfig) }.doOnErrors(::logConfigurationErrors).optional ?: return ExitCodes.FAILURE
+       
+        // Step 6. Check if we can access the certificates directory
+        if (!canReadCertificatesDirectory(cmdLineOptions.baseDirectory, configuration.devMode)) return ExitCodes.FAILURE
```
shamsasari: You can use `configuration.certificatesDirectory` instead

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -297,6 +301,19 @@ open class NodeStartup : NodeStartupLogging {
...
     }
 
+    private fun canReadCertificatesDirectory(baseDirectory: Path, devMode: Boolean): Boolean {
+        //Test for access to the certificates path and shutdown if we are unable to reach it.
+        //We don't do this if devMode==true because the certificates would be created anyway
+        if (devMode) return true
+
+        val certPath = baseDirectory / CERTIFICATES_DIRECTORY_NAME
+        if (!certPath.isDirectory()) {
+            printError("Unable to access certificates directory ${certPath.toString()}. Node will now shutdown.")
```
shamsasari: This isn't the issue that ENT-3327 is experiencing. The node will create the certificates directory in devMode (as you've mentioned) and during the initial doorman registration process (https://github.com/corda/corda/blob/5e16961fee3502390154b24ce4212196f50dfa43/node/src/main/kotlin/net/corda/node/utilities/registration/NetworkRegistrationHelper.kt#L84). So if it's missing it simply means that the node has been started without registering with the doorman. The error message should highlight that instead.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -149,17 +150,20 @@ open class NodeStartup : NodeStartupLogging {
...
         if (attempt { preNetworkRegistration(configuration) }.doOnFailure(Consumer(::handleRegistrationError)) !is Try.Success) return ExitCodes.FAILURE
 
-        // Step 8. Log startup info.
+        // Step 9. Log startup info.
         logStartupInfo(versionInfo, configuration)
 
-        // Step 9. Start node: create the node, check for other command-line options, add extra logging etc.
+        // Step 10. Start node: create the node, check for other command-line options, add extra logging etc.
         if (attempt {
                     cmdLineOptions.baseDirectory.createDirectories()
```
shamsasari: This no longer makes sense if we've already checked for the existence of the certificates directory.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
lockathan: Cool I'll updated it to use the certificates path from the configuration.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
lockathan: Fixed

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -297,6 +299,18 @@ open class NodeStartup : NodeStartupLogging {
         return true
     }
 
+    private fun canReadCertificatesDirectory(certDirectory: Path, devMode: Boolean): Boolean {
+        //Test for access to the certificates path and shutdown if we are unable to reach it.
+        //We don't do this if devMode==true because the certificates would be created anyway
+        if (devMode) return true
+
+        if (!certDirectory.isDirectory()) {
+            printError("Unable to access certificates directory ${certDirectory.toString()}. This could be because the node has not been registered with the Identity Operator. Node will now shutdown.")
```
shamsasari: You don't need the `.toString()`

shamsasari: Perhaps it would be helpful to print the docsite for more info: https://docs.corda.net/joining-a-compatibility-zone.html

PR #4906, opened 2019-03-20
CORDA-2763 - Log censored config at startup

tlil:
JIRA ticket: [CORDA-2763](https://r3-cev.atlassian.net/browse/CORDA-2763)

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/subcommands/ValidateConfigurationCli.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/subcommands/ValidateConfigurationCli.kt
@@ -26,7 +26,7 @@ internal class ValidateConfigurationCli : CliWrapperBase("validate-configuration
             return "for path: \"$pathAsString\": $message"
         }
 
-        internal fun logRawConfig(config: Config) = logger.debug("Actual configuration:\n${V1NodeConfigurationSpec.describe(config, Any?::toConfigValue).render(configRenderingOptions)}")
+        internal fun logRawConfig(config: Config) = logger.info("Actual configuration:\n${V1NodeConfigurationSpec.describe(config, Any?::toConfigValue).render(configRenderingOptions)}")
     }
 
     @Mixin


node/src/main/kotlin/net/corda/node/internal/subcommands/ValidateConfigurationCli.kt
```diff
@@ -26,7 +26,7 @@ internal class ValidateConfigurationCli : CliWrapperBase("validate-configuration
             return "for path: \"$pathAsString\": $message"
         }
 
-        internal fun logRawConfig(config: Config) = logger.debug("Actual configuration:\n${V1NodeConfigurationSpec.describe(config, Any?::toConfigValue).render(configRenderingOptions)}")
+        internal fun logRawConfig(config: Config) = logger.info("Actual configuration:\n${V1NodeConfigurationSpec.describe(config, Any?::toConfigValue).render(configRenderingOptions)}")
```
anthonykeenan: If it's just a case of changing the logging level, I'm inclined to ask Matthew whether he's alright with that and not make any changes. It does put a lot of junk in the log files. WDYT?

tlil: Well, if you switch on debug-level logging, you get a very verbose node log. :) Also, note that we're not dumping the fully annotated config here. Just the provided one with censored fields. Personally, I don't see a problem with 20-80 lines of config dumped in a 1000s-line long log. And to his point previously, it's in production that this is useful for validation during support requests, etc., and most people won't run a production node with debug-level logging, I don't think.

mnesbit: Customers run at info. We always get sent a garbled story and a full log encapsulating the true config is key to problem resolution. The config must hide any passwords (and probably the DB jdbc URL for that reason too)

anthonykeenan: Ok sounds sensible, in that case looks fine to me.

tlil: > Customers run at info. We always get sent a garbled story and a full log encapsulating the true config is key to problem resolution. The config must hide any passwords (and probably the DB jdbc URL for that reason too)

@mnesbit, yep, this is exactly what it does :-)

PR #4907, opened 2019-03-20
CORDA-2653 - during initialisation a Corda Service should have a thread context classloader

dazraf:
This is regarding JIRA: [CORDA-2653](https://r3-cev.atlassian.net/browse/CORDA-2653).

During the initalisation of a `CordaService` if `Thread.currentThread().getContextClassLoader()` is `null`, certain third party libraries may fail to load. 

This fix ensures that a classloader is available during initialisation.


dazraf:
This PR is in accordance with the [Developer's Certificate of Origin](https://github.com/corda/corda/blob/master/CONTRIBUTING.md#developer-certificate-of-origin).

PR #4909, opened 2019-03-20
CORDA-2767: Add missing cliutils dependency to explorer

cburlinchon:


anthonykeenan:
Can you change it to a public Jira please!

cburlinchon:
I removed the copying of log4j file from config/dev, we seem to do this in other places, https://r3-cev.atlassian.net/browse/CORDA-2768

cburlinchon:
> Does that cause the error to go away?

Yep as we no longer bundle the config/dev/log4j which has ErrorCodeRewritePolicy

anthonykeenan:
Does that not mean it just doesn't do any logging at all now?

cburlinchon:
> Does that not mean it just doesn't do any logging at all now?

no it logs, it picks up the log4j file from explorer, which makes more sense

PR #4912, opened 2019-03-21
CORDA-2769 - fixed infinite recursive call for `copy` implementation.

dazraf:
Addresses: [CORDA-2769](https://r3-cev.atlassian.net/browse/CORDA-2769).
Added JvmOverloads to provide backwards compatibility for Java consumption.


core/src/main/kotlin/net/corda/core/node/NetworkParameters.kt
```diff
@@ -98,36 +98,20 @@ data class NetworkParameters(
         require(noPackageOverlap(packageOwnership.keys)) { "Multiple packages added to the packageOwnership overlap." }
     }
 
-    fun copy(minimumPlatformVersion: Int,
```
shamsasari: THe API Stability check is failing because you removed this method from the public API.

shamsasari:
Unit tests are failing

PR #4913, opened 2019-03-21
CORDA-2775: Simplify SourceClassLoader logic for DJVM.

chrisr3:
Simplify the `SourceClassLoader` logic to make it easier for people to use DJVM as a library.

Removing `SandboxSourceClassLoader` means that we don't need `AbstractSourceClassLoader` any more either, leaving us with just `SourceClassLoader` again.

### DIFF
--- a/djvm/djvm/src/test/java/net/corda/djvm/execution/SandboxExecutorJavaTest.java
+++ b/djvm/djvm/src/test/java/net/corda/djvm/execution/SandboxExecutorJavaTest.java
@@ -16,6 +16,8 @@ public class SandboxExecutorJavaTest extends TestBase {
 
     @Test
     public void testTransaction() {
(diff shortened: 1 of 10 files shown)


djvm/djvm/src/test/java/net/corda/djvm/execution/SandboxExecutorJavaTest.java
```diff
@@ -16,6 +16,8 @@
 
     @Test
     public void testTransaction() {
+        //TODO: Transaction should not be a pinned class! It needs
```
tlil: This will be addressed in a separate PR I take it?

chrisr3: Yes, it's _far_ too large a body of work to address now.

PR #4918, opened 2019-03-22
CORDA-2779 - Fix Netmap REST endpoint docs

fenryka:
Trivial change to the docs

### DIFF
--- a/docs/source/network-map.rst
+++ b/docs/source/network-map.rst
@@ -48,6 +48,24 @@ The set of REST end-points for the network map service are as follows.
 | GET            | /network-map/my-hostname                | Retrieve the IP address of the caller (and **not** of the network map).                                                                      |
 +----------------+-----------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------+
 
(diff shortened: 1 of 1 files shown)


docs/source/network-map.rst
```diff
@@ -48,6 +48,24 @@ The set of REST end-points for the network map service are as follows.
 | GET            | /network-map/my-hostname                | Retrieve the IP address of the caller (and **not** of the network map).                                                                      |
 +----------------+-----------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------------+
 
+The above represent the core of the Network Map protocol. However, the Corda Network (and other compatibility zones deploying the Enterprise Network Manager suite), provide the following enspoints
```
gendal: s/represent/represents/
s/enspoints/endpoints/

How about: "INFO: Network maps hosted by R3 or other parties using R3's commercial network management tools typically also provide the following endpoints as a convenience to operators and other users; we include them here as they can aid debugging but, for the avoidance of doubt, they are not a formal part of the spec and the node will operate even in their absence."

oknowles: Slightly nit-picky, but (assuming we mean node operators here) would it read better to say "...convenience to _node_ operators and other users..."?

PR #4926, opened 2019-03-25
CORDA-2786: Certificates directory is only required for node startup

lockathan:
Certificates directory should only be required on node startup. For all other cases, it is optional. This was breaking initial registration in particular, which is usually responsible for creating the certificates directory.

initialiseAndRun is called from all of the different CLI commands, including but not limited to generate-node-info and initial-registration.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -6,7 +6,6 @@ import net.corda.cliutils.CliWrapperBase
 import net.corda.cliutils.CordaCliWrapper
 import net.corda.cliutils.CordaVersionProvider
 import net.corda.cliutils.ExitCodes
-import net.corda.cliutils.ShellConstants
 import net.corda.core.contracts.HashAttachmentConstraint
 import net.corda.core.crypto.Crypto
 import net.corda.core.internal.*
@@ -111,7 +110,7 @@ open class NodeStartupCli : CordaCliWrapper("corda", "Runs a Corda Node") {
             else -> startup.initialiseAndRun(cmdLineOptions, object : RunAfterNodeInitialisation {
                 val startupTime = System.currentTimeMillis()
                 override fun run(node: Node) = startup.startNode(node, startupTime)
-            })
+            }, requireCertificates = true)
         }
     }
 }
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -151,8 +151,9 @@ open class NodeStartup : NodeStartupLogging {
         val configuration = cmdLineOptions.parseConfiguration(rawConfig).doIfValid { logRawConfig(rawConfig) }.doOnErrors(::logConfigurationErrors).optional ?: return ExitCodes.FAILURE
        
         // Step 6. Check if we can access the certificates directory
-        if (!canReadCertificatesDirectory(configuration.certificatesDirectory, configuration.devMode)) return ExitCodes.FAILURE
-
+        if (requireCertificates) {
```
chrisr3: I don't like the nested `if`. What's wrong with `and`?

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -111,7 +111,7 @@ open class NodeStartupCli : CordaCliWrapper("corda", "Runs a Corda Node") {
             else -> startup.initialiseAndRun(cmdLineOptions, object : RunAfterNodeInitialisation {
                 val startupTime = System.currentTimeMillis()
                 override fun run(node: Node) = startup.startNode(node, startupTime)
-            })
```
shamsasari: If we're requesting changes, can you add the named parameter for the `true`.

lockathan: Hi Shams. Sorry, I don't understand this comment? I do pass the value "true" in here. Do you mean that I should name the parameter at the point where it is passed in?

shamsasari: Yes, that's what I meant. Apologies for not being clear. `requireCertificates = true`

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
lockathan: Fixed - please have a look.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
lockathan: Fixed - please have a look.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -151,8 +151,8 @@ open class NodeStartup : NodeStartupLogging {
         val configuration = cmdLineOptions.parseConfiguration(rawConfig).doIfValid { logRawConfig(rawConfig) }.doOnErrors(::logConfigurationErrors).optional ?: return ExitCodes.FAILURE
        
         // Step 6. Check if we can access the certificates directory
-        if (!canReadCertificatesDirectory(configuration.certificatesDirectory, configuration.devMode)) return ExitCodes.FAILURE
-
+        if (requireCertificates && !canReadCertificatesDirectory(configuration.certificatesDirectory, configuration.devMode)) return ExitCodes.FAILURE
+
```
chrisr3: _AAAGH! Evil Whitespace!_ (And the unused `import` too...)

PR #4928, opened 2019-03-25
CORDA-2783 - Improve error message for initial registration with devMode=true

tlil:
In response to [CORDA-2783](https://r3-cev.atlassian.net/browse/CORDA-2783):

If `devMode=true` is set in the config, and `devModeOptions.allowCompatibilityZone` isn't, the node would fail with an uninformative message `Failed requirement`. With this fix, the following output gets produced instead:

```
   ______               __
  / ____/     _________/ /___ _
 / /     __  / ___/ __  / __ `/         Computers are useless. They can only
/ /___  /_/ / /  / /_/ / /_/ /          give you answers.  -- Picasso
\____/     /_/   \__,_/\__,_/

--- Corda Open Source 5.0-SNAPSHOT (a1f7c4f) -------------------------------------------------------------


Logs can be found in                    : C:\Users\TommyLillehagen\IdeaProjects\test-node\logs
Cannot perform initial registration when 'devMode' is true, unless 'devModeOptions.allowCompatibilityZone' is also true.
```

### DIFF
--- a/node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
+++ b/node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
@@ -93,7 +93,9 @@ open class SharedNodeCmdLineOptions {
 class InitialRegistrationCmdLineOptions : SharedNodeCmdLineOptions() {
     override fun parseConfiguration(configuration: Config): Valid<NodeConfiguration> {
         return super.parseConfiguration(configuration).doIfValid { config ->
-            require(!config.devMode || config.devModeOptions?.allowCompatibilityZone == true)
+            require(!config.devMode || config.devModeOptions?.allowCompatibilityZone == true) {
+                "Cannot perform initial registration when 'devMode' is true, unless 'devModeOptions.allowCompatibilityZone' is also true."
+            }
             require(config.compatibilityZoneURL != null || config.networkServices != null) {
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/NodeCmdLineOptions.kt
```diff
@@ -93,7 +93,9 @@ open class SharedNodeCmdLineOptions {
 class InitialRegistrationCmdLineOptions : SharedNodeCmdLineOptions() {
     override fun parseConfiguration(configuration: Config): Valid<NodeConfiguration> {
         return super.parseConfiguration(configuration).doIfValid { config ->
-            require(!config.devMode || config.devModeOptions?.allowCompatibilityZone == true)
+            require(!config.devMode || config.devModeOptions?.allowCompatibilityZone == true) {
+                "Cannot perform initial registration when 'devMode' is true, unless 'devModeOptions.allowCompatibilityZone' is also true."
```
szymonsztuka: Just question about wording. Is 'initial registration' different than  'registration mode' (the next case)?

tlil: Nope, they're the same. And this is in the `InitialRegistrationCmdLineOptions` class, so went with that. :-)

PR #4931, opened 2019-03-25
CORDA-2730: Enforce finance cordapp currency restrictions

lockathan:
Changes CashIssueFlow to verify that the finance workflows cordapp that it is executing in
is configured to support the requested currency. The finance workflows cordapp exposes a
configuration property that stores issuable currencies. We check against this list of
issuable currencies before actually issuing currency.

Note that in the default cause, all currencies are supported. In other words, this only
takes effect if the operator deliberately overrides the list of supported currencies with
a custom list.

### DIFF
--- a/finance/workflows/src/main/kotlin/net/corda/finance/flows/AbstractCashFlow.kt
+++ b/finance/workflows/src/main/kotlin/net/corda/finance/flows/AbstractCashFlow.kt
@@ -46,6 +46,8 @@ abstract class AbstractCashFlow<out T>(override val progressTracker: ProgressTra
     abstract class AbstractRequest(val amount: Amount<Currency>)
 }
 
+class UnsupportedCurrencyException internal constructor(message: String) : Exception(message) {}
+
 // The internal constructor here allows the ThrowableSerializer to find a suitable constructor even if the `cause` field is removed using reflection by the RPC proxies.
 class CashException internal constructor(cause: Throwable?, message: String) : FlowException(message, cause) {
     constructor(message: String, cause: Throwable) : this(cause, message)
--- a/finance/workflows/src/main/kotlin/net/corda/finance/flows/CashIssueFlow.kt
+++ b/finance/workflows/src/main/kotlin/net/corda/finance/flows/CashIssueFlow.kt
@@ -9,6 +9,9 @@ import net.corda.core.transactions.TransactionBuilder
 import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.utilities.ProgressTracker
 import net.corda.finance.contracts.asset.Cash
+import net.corda.finance.internal.CashConfigDataFlow
+import net.corda.finance.flows.CashException
+import net.corda.finance.flows.UnsupportedCurrencyException
 import net.corda.finance.flows.AbstractCashFlow.Companion.FINALISING_TX
 import net.corda.finance.flows.AbstractCashFlow.Companion.GENERATING_TX
 import net.corda.finance.flows.AbstractCashFlow.Companion.SIGNING_TX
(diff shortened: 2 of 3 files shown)


josecoll:
And now begins the fun of fixing tests that made assumptions on issuable currencies ;-)

finance/workflows/src/main/kotlin/net/corda/finance/flows/AbstractCashFlow.kt
```diff
@@ -46,6 +46,8 @@ abstract class AbstractCashFlow<out T>(override val progressTracker: ProgressTra
     abstract class AbstractRequest(val amount: Amount<Currency>)
 }
 
+class UnsupportedCurrencyException internal constructor(message: String) : Exception(message) {}
```
shamsasari: The `internal` constructor isn't internal in Javaland.

shamsasari: Remove braces, they're not needed

finance/workflows/src/main/kotlin/net/corda/finance/flows/CashIssueFlow.kt
```diff
@@ -39,6 +42,9 @@ class CashIssueFlow(private val amount: Amount<Currency>,
 
     @Suspendable
     override fun call(): AbstractCashFlow.Result {
+        var cashConfig = subFlow(CashConfigDataFlow())
+        if (!cashConfig.issuableCurrencies.contains(amount.token))
+            throw CashException("Unsupported currency requested", UnsupportedCurrencyException("Unable to issue currency ${amount.token} because the CordApp is not configured to support it."))
```
shamsasari: Make UnsupportedCurrencyException extend CashException and throw it directly

finance/workflows/src/main/kotlin/net/corda/finance/flows/CashIssueFlow.kt
```diff
@@ -39,6 +42,9 @@ class CashIssueFlow(private val amount: Amount<Currency>,
 
     @Suspendable
     override fun call(): AbstractCashFlow.Result {
+        var cashConfig = subFlow(CashConfigDataFlow())
```
shamsasari: Use the `ConfigHolder` service directly, like how `CashConfigDataFlow` does.

finance/workflows/src/main/kotlin/net/corda/finance/flows/CashIssueFlow.kt
```diff
@@ -39,6 +42,9 @@ class CashIssueFlow(private val amount: Amount<Currency>,
...
+        if (!cashConfig.issuableCurrencies.contains(amount.token))
```
shamsasari: Use the kotlin `in` operator

finance/workflows/src/main/kotlin/net/corda/finance/flows/AbstractCashFlow.kt
lockathan: Would doing these sorts of checks at flow run time be against the intention of the cash issue flow then? It doesn't seem to have ever had such checking. At least, I went back to Corda 3 and could find no indication that it ever did. So, we can introduce it, but should we?
Should I just remove the internal keyword then?

shamsasari:
The issuable currencies list wasn't introduced to enforce any issuance checks (as far as I understand). It's simply there to help populate the issuable currencies dropdown list in demobench. We shouldn't make it out more than what it is - a demo-enabling feature.

lockathan:
If issuableCurrencies relates only to demos, I'd rather remove it entirely and handle the configuration of the demo in some other way. Perhaps extend the finance workflows cordapp with a finance workflows demo cordapp?

Reason being, we now have code in the finance workflows cordapp  that is not needed by the cordapp itself but needed for something external to it.

josecoll:
Right - but `CashConfigDataFlow` is part of the finance module as is `CashIssueFlow` so they are related together (it just so happens that we forgot to add the validation in the `CashIssueFlow` when `CashConfigDataFlow` was introduced, BUT did use it in a same GUI app).

shamsasari:
I agree. I thought by moving the `CashConfigDataFlow` to an internal package would be one step towards that.

shamsasari:
I think we're all agree this PR as is is not required.

josecoll:
I disagree - I think this needs to be cleaned up one way or the other.
It certainly caused a test failure in QA.

lockathan:
Hi. Okay, so in summary, we should remove CashConfigDataFlow from the finance workflows cordapp because it doesn't belong there. We could either move it to the demo app itself, if it is specific to that, or to a shared package if it is not specific to the demo.

Should we do that with this JIRA and PR, or is it better to create a new JIRA and PR for these changes? It sounds like this is more of a refactoring than a bugfix in any case, since the lack of enforcement in the corda finance cordapp is not actually a bug.

shamsasari:
It seems this PR should be closed.

PR #4932, opened 2019-03-25
CORDA-2802 use eventually to wait

r3domfox:
Tests which use `Thread.sleep` to wait for some asynchronous process to complete are liable to flicker due to inconsistent running times. It is usually preferable to test some explicit condition repeatedly, timing out if it is not met within a reasonable time period - the test will then be able to continue immediately as soon as the condition is met, and can have a more generous timeout allowance since we will not usually be waiting that long.

The `eventually` method provides a convenient common way for tests to do this, and should be used where appropriate:

```kotlin
inline fun <R> eventually(
        duration: Duration = Duration.ofSeconds(5),
        waitBetween: Duration = Duration.ofMillis(100),
        waitBefore: Duration = waitBetween,
        test: () -> R): R {
    val end = System.nanoTime() + duration.toNanos()
    var times = 0
    var lastFailure: AssertionError? = null

    if (!waitBefore.isZero) Thread.sleep(waitBefore.toMillis())

    while (System.nanoTime() < end) {
        try {
            return test()
        } catch (e: AssertionError) {
            if (!waitBetween.isZero) Thread.sleep(waitBetween.toMillis())
            lastFailure = e
        }
        times++
    }

    throw AssertionError("Test failed with \"${lastFailure?.message}\" after $duration; attempted $times times")
}
```

This function is added to `test-common`, as it is used in multiple modules.


testing/test-common/src/main/kotlin/net/corda/testing/common/internal/Eventually.kt
```diff
@@ -0,0 +1,50 @@
...
+}
+
+
+/**
+ * Use when the action you want to retry until it succeeds throws an exception, rather than failing a test.
+ */
+inline fun <R> succeeds(action: () -> R): R =
+    try {
+        action()
+    } catch (e: Throwable) {
```
chrisr3: I think we should be catching `Exception` rather than `Throwable`.

r3domfox: Done.

PR #4935, opened 2019-03-25
investigate Party, x500 / key decoupling

roastario:
NOTE, COLLECTSIGNATURES TODO

see: 
https://github.com/corda/accounts/pull/3
https://github.com/corda/accounts-demo/pull/7


core/src/main/kotlin/net/corda/core/identity/Party.kt
```diff
@@ -36,4 +36,9 @@ class Party(val name: CordaX500Name, owningKey: PublicKey) : AbstractParty(ownin
     fun anonymise(): AnonymousParty = AnonymousParty(owningKey)
     override fun ref(bytes: OpaqueBytes): PartyAndReference = PartyAndReference(this, bytes)
     override fun toString() = name.toString()
+
+    fun withNewKey(owningKey: PublicKey) : Party{
+        return Party(this.name, owningKey)
+    }
```
roger3cev: extension function?

roger-that-dev (commented):
Excellent! Looks like we can get this in a minor release if we make the new method on `Party` an extension function.

PR #4939, opened 2019-03-26
Minor tweaks and improvements in docs TOC to aid navigation and access.

josecoll:
Minor re-structuring to include:
- Corda API link moves to top-level (near top)
- Cheatsheet moves to top-level
- Design documents grouped under sub-header in Contributions
- Vault / Soft-locking moved to key concepts
- added toc entries for missing references as reported by `WARNING: document isn't included in any toctree`

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -34,6 +34,8 @@ We look forward to seeing what you can do with Corda!
    release-notes
    app-upgrade-notes
    node-upgrade-notes
+   corda-api
+   cheat-sheet
 
 .. toctree::
    :caption: Development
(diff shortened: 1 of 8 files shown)


josecoll:
@nimmaj Please have a quick scan to see if I've missed anything you haven't already covered.

docs/source/index.rst
```diff
@@ -88,24 +90,6 @@ We look forward to seeing what you can do with Corda!
 
    contributing-index.rst
    deterministic-modules.rst
+   design/design-docs-index.rst
```
mikehearn: Hmm. I think we've lost the htmlmode-only conditionality. maybe it can go into `design-docs-index.rst` or maybe it can stay in the root index.rst

PR #4940, opened 2019-03-26
CORDA-2782 allow un-whitelisted Comparable

r3domfox:
Because vault queries have properties of type `Comparable<T>`, and the default whitelist used in RPC serialisation does not include `Comparable`. we need to make special exception for `Comparable` when checking against the whitelist. (The alternative approach of adding `Comparable` to the default whitelist was tried, but amounted to a wide-ranging change in what gets sent on the wire during RPC serialisation, which caused breakages elsewhere).

This was not an issue in Corda OS v3 because Kryo serialisation was used for RPC.


josecoll:
This feels like a very tactical fix (which may be fine to get CE 4.0 over the line), but I think we will just be postponing further pain (until another interface type comes along).

PR #4941, opened 2019-03-26
allow abstractParty to open a flow session

roastario:
see: https://github.com/corda/accounts-demo/pull/8 and https://github.com/corda/accounts/pull/4

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     file("$projectDir/constants.properties").withInputStream { constants.load(it) }
 
     // Our version: bump this on release.
-    ext.corda_release_version = constants.getProperty("cordaVersion") + "-NEW_IDENTITY_SERVICE"
+    ext.corda_release_version = constants.getProperty("cordaVersion")
     ext.corda_platform_version = constants.getProperty("platformVersion")
     ext.gradle_plugins_version = constants.getProperty("gradlePluginsVersion")
 
@@ -468,10 +468,4 @@ wrapper {
 buildScan {
     termsOfServiceUrl = 'https://gradle.com/terms-of-service'
     termsOfServiceAgree = 'yes'
-}
-
-task printVersionBuilt(){
-    doLast{
-        println project.version
-    }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
     file("$projectDir/constants.properties").withInputStream { constants.load(it) }
 
     // Our version: bump this on release.
-    ext.corda_release_version = constants.getProperty("cordaVersion")
+    ext.corda_release_version = constants.getProperty("cordaVersion") + "-NEW_IDENTITY_SERVICE"
     ext.corda_platform_version = constants.getProperty("platformVersion")
     ext.gradle_plugins_version = constants.getProperty("gradlePluginsVersion")
 
@@ -468,4 +468,10 @@ wrapper {
 buildScan {
     termsOfServiceUrl = 'https://gradle.com/terms-of-service'
     termsOfServiceAgree = 'yes'
+}
+
+task printVersionBuilt(){
+    doLast{
+        println project.version
+    }
 }
--- a/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
@@ -110,7 +110,7 @@ class CollectSignaturesFlow @JvmOverloads constructor(val partiallySignedTx: Sig
         if (unsigned.isEmpty()) return partiallySignedTx
 
         val setOfAllSessionKeys = sessionsToCollectFrom.groupBy { it.sessionOwningKey }.map {
-            require(it.value.size == 1) {"There are multiple sessions initiated for party key ${it.key.toStringShort()}"}
+            require(it.value.size == 1) { "There are multiple sessions initiated for party key ${it.key.toStringShort()}" }
             it.key to it.value.first()
         }.toMap()
 
@@ -131,22 +131,23 @@ class CollectSignaturesFlow @JvmOverloads constructor(val partiallySignedTx: Sig
             }
         }
 
-        val keysMissingASession = unsigned.filterNot {
-            //every unsigned key must have a session it can ask
+        val keyToSessionMap = unsigned.map {
             if (it in setOfAllSessionKeys) {
-                // the unsigned key exists directly as a sessionKey
-                true
+                // the unsigned key exists directly as a sessionKey, so use that session
+                it to setOfAllSessionKeys[it]
             } else {
                 //it might be delegated to a wellKnownParty
                 val wellKnownParty: Party? = keyToSigningParty[it]
                 if (wellKnownParty != null) {
-                    //there is a wellKnownParty for this key, check if it has a session
-                    setOfAllSessionKeys[wellKnownParty.owningKey] != null
+                    //there is a wellKnownParty for this key, check if it has a session, and if so - use that session
+                    val sessionForWellKnownParty = setOfAllSessionKeys[wellKnownParty.owningKey]
+                            ?: throw IllegalStateException("No session available to request signature for key: ${it.toStringShort()}")
+                    it to sessionForWellKnownParty
                 } else {
-                    false
+                    throw IllegalStateException("Could not find a session or wellKnown party for key ${it.toStringShort()}")
                 }
             }
-        }
+        }.toMap()
 
 //
 //        require(keysNotProvidedBySessions.isEmpty() && sessionsNotRelatedToKeys.isEmpty()) {
--- a/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
@@ -3,12 +3,15 @@ package net.corda.core.flows
 import co.paralleluniverse.fibers.Suspendable
 import net.corda.core.crypto.TransactionSignature
 import net.corda.core.crypto.isFulfilledBy
+import net.corda.core.crypto.toStringShort
+import net.corda.core.identity.AbstractParty
 import net.corda.core.identity.Party
 import net.corda.core.identity.groupPublicKeysByWellKnownParty
 import net.corda.core.node.ServiceHub
 import net.corda.core.transactions.SignedTransaction
 import net.corda.core.transactions.WireTransaction
 import net.corda.core.utilities.ProgressTracker
+import net.corda.core.utilities.toBase58String
 import net.corda.core.utilities.unwrap
 import java.security.PublicKey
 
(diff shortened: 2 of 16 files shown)


gendal:
No comments on the implementation (except to wonder if the exception message string should begin with a capital letter) so just some higher level ones:

* In your planned usage for a first pass at 'accounts', the only code that would call initiateFlow with an AnonymousParty would be inside the putative accounts library... ie customers building against the accounts API would not directly call this API. Is that right?
* If so, it seems that the risk we take by doing this in a 4.1 release rather than 5.0 is that somebody else might discover it and use it?
* So the question is:  is that good or bad?  If good then let's document it properly. If bad, what is the risk we're taking?
* My assumption (perhaps wrong?) is that the risk is: somebody builds their app against C4.1 and uses this API... maybe what they have is actually a Party but they pass it in as an AbstractParty... so they're not even taking advantage of the new feature!!  But THEN...  they or somebody else takes their app and tries to run it on C4.0.  And then bad things happen?  Is that the risk?    If so, do we need to mitigate it and, if so, how?
* Which leads me to my knowledge gap...   we use platform version to talk about consensus/network-critical rules... how do we reason about/communicate API versions?  Is there a way to write an app that says "I only run on 4.1 or higher"?  And given that even if there was many people wouldn't populate the manifest correctly in any case so perhaps it actually doesn't matter that this "build on 4.1, run on 4.0, explosion" issue could happen...?
* To be clear: it seems to be that the risk of doing this in 4.1 is low (assuming the implementation itself makes sense!!). We just have to carry a bit of complexity around for a bit.... and since we will be strongly encouraging everybody to move to 4.1 asap we won't even have to live with it for long in any case.
* And yes.. if our 'rules' say this change means the API has changed and so it means it must be Corda 5, I really worry about that...  after three years of trying, people still don't understand our approach to versioning and the evidence is that there are LOTS of mental shortcuts that have been hard-wired into our customers' deployment processes...    bumping from 4 to 5 for this would be very hard to message. (we got away with it with c2 - I'm not sure we'll be as lucky this time)

roger3cev:
> In your planned usage for a first pass at 'accounts', the only code that would call initiateFlow with an AnonymousParty would be inside the putative accounts library... ie customers building against the accounts API would not directly call this API. Is that right?

Note quite :) I think the key thing to note is that all CorDapps which need to initiate a session using `AnonymousParty`s would use the new `initiateFlow` method. The reason being is that if they write a flow which uses accounts then they must deal with `AnonymousParty`s and call `FinalityFlow` / `CollectSignaturesFlow`, this API change prevents them from writing their own shims around these flows to deal with `AnonymousParty`s. Therefore, accounts will "just work".

The other change is to `PersistentIdentityService` but this is an internal API change only. The compromise here is that the accounts SDK must be compiled against the Corda node JAR until a public API has been added for this functionality in Corda 5. I think this is less of an issue, though.

> If so, it seems that the risk we take by doing this in a 4.1 release rather than 5.0 is that somebody else might discover it and use it?

Well, given that those using accounts would _have_ to use this API, the risk might be that they build their app against version 4.1 and expect to run it on a version 4 node.

> My assumption (perhaps wrong?) is that the risk is: somebody builds their app against C4.1 and uses this API... maybe what they have is actually a Party but they pass it in as an AbstractParty... so they're not even taking advantage of the new feature!! But THEN... they or somebody else takes their app and tries to run it on C4.0. And then bad things happen? Is that the risk? If so, do we need to mitigate it and, if so, how?

Maybe we can mitigate this by specifying the parameter as `AnonymousParty`?

> Which leads me to my knowledge gap... we use platform version to talk about consensus/network-critical rules... how do we reason about/communicate API versions? Is there a way to write an app that says "I only run on 4.1 or higher"? And given that even if there was many people wouldn't populate the manifest correctly in any case so perhaps it actually doesn't matter that this "build on 4.1, run on 4.0, explosion" issue could happen...?

As I understand, targetVersion must be an integer.

> To be clear: it seems to be that the risk of doing this in 4.1 is low (assuming the implementation itself makes sense!!). We just have to carry a bit of complexity around for a bit.... and since we will be strongly encouraging everybody to move to 4.1 asap we won't even have to live with it for long in any case.

Agree, I think it's low. If we don't document this then most devs won't know it's there. Also, as you note, given that we will release version 4.1 very soon, the time window for this to have an impact is quite narrow.

Cheers

build.gradle
```diff
@@ -4,7 +4,7 @@ buildscript {
     file("$projectDir/constants.properties").withInputStream { constants.load(it) }
 
     // Our version: bump this on release.
-    ext.corda_release_version = constants.getProperty("cordaVersion")
+    ext.corda_release_version = constants.getProperty("cordaVersion") + "-NEW_IDENTITY_SERVICE"
```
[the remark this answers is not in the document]
roastario: this will be fixed, I just wanted to not overwrite my local install.

core/src/main/kotlin/net/corda/core/flows/FlowSession.kt
```diff
@@ -49,6 +50,7 @@ abstract class FlowSession {
      *   [counterparty] is the same Party as the one passed to that function.
      */
     abstract val counterparty: Party
+    abstract val sessionOwningKey: PublicKey
```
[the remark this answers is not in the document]
mikehearn: The name of this field is a bit confusing. Also it seems like it's unnecessarily key specific, if we were to introduce other kinds of things you could initiate flows with that subclass AbstractParty in future, this API wouldn't work.

Can we instead introduce a new field and deprecate counterparty (fixing up the call sites):

`abstract val destination: AbstractParty`

with the existing property becoming just

`@Deprecated val counterparty: Party get() destination as? Party ?: throw IllegalStateException("This flow session is not directly with a node identity, it's actually a ${destination.javaClass.simpleName}")`

or words to that effect.

What I'm thinking of here is future possible upgrades to allow flows to communicate with things like enclaves, groups of peers instead of single peers, internal services (core team is starting to look at that) etc.

mikehearn:
Obviously needs tests, docsite changes, all the usual paraphernalia.

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
```diff
@@ -105,14 +108,55 @@ class CollectSignaturesFlow @JvmOverloads constructor(val partiallySignedTx: Sig
         // If the unsigned counterparties list is empty then we don't need to collect any more signatures here.
         if (unsigned.isEmpty()) return partiallySignedTx
 
-        val partyToKeysMap = groupPublicKeysByWellKnownParty(serviceHub, unsigned)
-        // Check that we have a session for all parties.  No more, no less.
-        require(sessionsToCollectFrom.map { it.counterparty }.toSet() == partyToKeysMap.keys) {
+        val setOfAllSessionKeys = sessionsToCollectFrom.groupBy { it.sessionOwningKey }.map {
```
shamsasari: Use `mapValues` instead of `map` and `toMap`

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
```diff
@@ -105,14 +108,55 @@ class CollectSignaturesFlow @JvmOverloads constructor(val partiallySignedTx: Sig
...
+
+        val unrelatedSessions = sessionsToCollectFrom.filterNot {
+            if (it.sessionOwningKey == it.counterparty.owningKey) {
+                //this session was initiated by a wellKnownParty
+                //the session must have a corresponding unsigned
+                it.counterparty in partyToKeysItSignsFor
+            } else {
+                //this session was not initiated by a wellKnownParty
+                //so must directly exist in the unsigned
+                unsigned.contains(it.sessionOwningKey)
```
shamsasari: `in`

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
```diff
@@ -105,14 +108,55 @@ class CollectSignaturesFlow @JvmOverloads constructor(val partiallySignedTx: Sig
...
+            } else {
+                //it might be delegated to a wellKnownParty
+                val wellKnownParty: Party? = keyToSigningParty[it]
+                if (wellKnownParty != null) {
+                    //there is a wellKnownParty for this key, check if it has a session, and if so - use that session
+                    val sessionForWellKnownParty = setOfAllSessionKeys[wellKnownParty.owningKey]
+                            ?: throw IllegalStateException("No session available to request signature for key: ${it.toStringShort()}")
+                    it to sessionForWellKnownParty
+                } else {
+                    throw IllegalStateException("Could not find a session or wellKnown party for key ${it.toStringShort()}")
```
shamsasari: You can remove the if-else by doing `val wellKnownParty = checkNotNull(keyToSigningParty[it]) { "..." }`

core/src/main/kotlin/net/corda/core/flows/FlowLogic.kt
```diff
@@ -115,6 +116,19 @@ abstract class FlowLogic<out T> {
...
 
+    /**
+     * Creates a communication session with [party]. Subsequently you may send/receive using this session object. Note
+     * that this function does not communicate in itself, the counter-flow will be kicked off by the first send/receive.
+     */
+    @Suspendable
+    fun initiateFlow(requested: AbstractParty): FlowSession {
+        val wellKnown = serviceHub.identityService.wellKnownPartyFromAnonymous(requested)
+        if (wellKnown == null) {
+            throw IllegalStateException("could not initiate flow with party $requested as they are not in the node identity service")
```
shamsasari: Use `checkNotNull`

core/src/main/kotlin/net/corda/core/node/services/IdentityService.kt
```diff
@@ -24,6 +26,10 @@ interface IdentityService {
     val trustAnchor: TrustAnchor
     val caCertStore: CertStore
 
+    companion object {
+        val log = contextLogger()
```
shamsasari: this needs to be private

gendal:
btw... I agree with and support shams here... my understanding of this ticket is that it is intended to solicit feedback on a path forward but it isn't mergeable (no docs, no justification/explanation, etc)...  assuming the outcome is "rough consensus" that this approach can work then I would expect to see a short write-up as part of the associated jira so we can trace back in the future to why we did this, etc.  cc @roger3cev

mikehearn:
There was some discussion off-PR about adding a `Destination` supertype (or interface?) to AbstractParty. Probably should be supertype.

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
roastario: you can, but this is much easier to read, and has the same performance characteristics.

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
roastario: I prefer to make what is going on clearer over making the code a few lines shorter.

roastario:
closing to re-target release/5

PR #4957, opened 2019-03-28
CORDA-2586 explorer exception handling

StefanIliev545:
Changed the dialog used for exception handling in the login view from the default ExceptionDialog which always shows a stack trace to a custom Dialog that won't show stack trace unless asked to. RPC exceptions won't show a stack trace as from my understanding running through the code that throws them, they are either network or protocol issues, thus there isn't much point. Other exceptions that reach the catch will still be shown. The Dialog should display a hyperlink to whatever is generated by errorCodeUrl() extension function. (Only tested that it works on mac) 

CordaVersionProvider now contains a CordaVersion, so that the error url generation logic could be moved to a different library, because CordaVersionProvider extends some  third party module interface and there is no point in linking that interface whenever you want to get the versioning info.

Note that I'm still navigating around gradle and the build.gradle in common-logging was copied from another common/ lib so there might be an obvious mistake there or something useless.

### DIFF
--- a/common/logging/src/main/kotlin/net/corda/common/logging/CordaVersion.kt
+++ b/common/logging/src/main/kotlin/net/corda/common/logging/CordaVersion.kt
@@ -2,10 +2,6 @@ package net.corda.common.logging
 
 import com.jcabi.manifests.Manifests
 
-/**
- * Simple version printing when command is called with --version or -V flag. Assuming that we reuse Corda-Release-Version and Corda-Revision
- * in the manifest file.
- */
 class CordaVersion  {
     companion object {
         private const val UNKNOWN = "Unknown"
--- a/common/logging/src/main/kotlin/net/corda/common/logging/CordaVersion.kt
+++ b/common/logging/src/main/kotlin/net/corda/common/logging/CordaVersion.kt
@@ -1,0 +1,32 @@
+package net.corda.common.logging
+
+import com.jcabi.manifests.Manifests
+
+/**
+ * Simple version printing when command is called with --version or -V flag. Assuming that we reuse Corda-Release-Version and Corda-Revision
+ * in the manifest file.
+ */
+class CordaVersion  {
+    companion object {
+        private const val UNKNOWN = "Unknown"
+        const val current_major_release = "4.0-SNAPSHOT"
+        const val platformEditionCode = "OS"
+
+        private fun manifestValue(name: String): String? = if (Manifests.exists(name)) Manifests.read(name) else null
+
+        val releaseVersion: String by lazy { manifestValue("Corda-Release-Version") ?: UNKNOWN }
+        val revision: String by lazy { manifestValue("Corda-Revision") ?: UNKNOWN }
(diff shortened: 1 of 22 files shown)


common/logging/src/main/kotlin/net/corda/common/logging/CordaVersion.kt
```diff
@@ -0,0 +1,32 @@
+package net.corda.common.logging
+
+import com.jcabi.manifests.Manifests
+
+/**
+ * Simple version printing when command is called with --version or -V flag. Assuming that we reuse Corda-Release-Version and Corda-Revision
+ * in the manifest file.
+ */
+class CordaVersion  {
```
josecoll: This information already exists elsewhere in the code base.
Please do not duplicate.

StefanIliev545: Thanks, I've removed it in the last commit.

common/logging/src/main/kotlin/net/corda/common/logging/ExceptionsErrorCodeFunctions.kt
```diff
@@ -1,21 +1,22 @@
-package net.corda.cliutils
+package net.corda.common.logging
 
 import org.apache.logging.log4j.Level
 import org.apache.logging.log4j.message.Message
 import org.apache.logging.log4j.message.SimpleMessage
 import java.util.*
 
-internal fun Message.withErrorCodeFor(error: Throwable?, level: Level): Message {
+
```
anthonykeenan: extra line

tools/cliutils/src/main/kotlin/net/corda/cliutils/ErrorCodeRewritePolicy.kt
```diff
@@ -6,6 +6,8 @@ import org.apache.logging.log4j.core.appender.rewrite.RewritePolicy
 import org.apache.logging.log4j.core.config.plugins.Plugin
 import org.apache.logging.log4j.core.config.plugins.PluginFactory
 import org.apache.logging.log4j.core.impl.Log4jLogEvent
+import net.corda.common.logging.withErrorCodeFor
+
```
anthonykeenan: extra line

tools/cliutils/src/main/kotlin/net/corda/cliutils/ErrorCodeRewritePolicy.kt
```diff
@@ -6,6 +6,8 @@ import org.apache.logging.log4j.core.appender.rewrite.RewritePolicy
...
 
 @Plugin(name = "ErrorCodeRewritePolicy", category = Core.CATEGORY_NAME, elementType = "rewritePolicy", printObject = false)
 class ErrorCodeRewritePolicy : RewritePolicy {
```
anthonykeenan: Shouldn't this go in the logging module?

PR #4958, opened 2019-03-29
First attempt to plumb CorDapp config through to whitelisting.  Faili…

rick-r3:
…ng local unit tests when run through gradle.


rick-r3:
I can't get this to pass local unit tests when run from the command line.  Individually tests pass, and don't think I've even touched a relevant bit of code, but maybe I did.

rick-r3:
Established there's a unit test instability

PR #4960, opened 2019-03-29
CORDA-2814: Check if result set is closed when getting tables in initial registration

thschroeter:
A SQLException is thrown when calling first() on a closed result set.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
@@ -1,10 +1,7 @@
 package net.corda.node.internal.subcommands
 
 import net.corda.cliutils.CliWrapperBase
-import net.corda.core.internal.createFile
-import net.corda.core.internal.div
-import net.corda.core.internal.exists
-import net.corda.core.utilities.Try
+import net.corda.core.internal.*
 import net.corda.node.InitialRegistrationCmdLineOptions
 import net.corda.node.NodeRegistrationOption
 import net.corda.node.internal.*
@@ -12,10 +9,12 @@ import net.corda.node.internal.NodeStartupLogging.Companion.logger
 import net.corda.node.services.config.NodeConfiguration
 import net.corda.node.utilities.registration.HTTPNetworkRegistrationService
 import net.corda.node.utilities.registration.NodeRegistrationHelper
+import net.corda.nodeapi.internal.persistence.NODE_DATABASE_PREFIX
 import picocli.CommandLine.Mixin
 import picocli.CommandLine.Option
 import java.io.File
 import java.nio.file.Path
+import java.sql.SQLException
 import java.util.function.Consumer
 
 class InitialRegistrationCli(val startup: NodeStartup): CliWrapperBase("initial-registration", "Start initial node registration with Corda network to obtain certificate from the permissioning server.") {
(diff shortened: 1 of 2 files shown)


rick-r3:
@thschroeter @dimosr I think this initial registration checking code needs some attention.   For example, it doesn't close the data source, connection etc.

dimosr:
> @thschroeter @dimosr I think this initial registration checking code needs some attention. For example, it doesn't close the data source, connection etc.

Yes, that's a good point, not sure if the datasource can/needs to be closed, but the connection definitely needs to.

@thschroeter another question I had is: how come the result set is closed ? Is it returned like that, because I can't see it being closed somewhere else.

thschroeter:
@dimosr will update the pr. I think we execute two statements on the connection before reading the results of the first statement.

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
```diff
@@ -98,17 +98,26 @@ class InitialRegistration(val baseDirectory: Path, private val networkRootTrustS
...
             val tablesUpperCaseResultSet = connectionMetadata.getTables(null, null, "${NODE_DATABASE_PREFIX.toUpperCase()}%", null)
-            check(!tablesLowerCaseResultSet.first() && !tablesUpperCaseResultSet.first()) {
-                "The database contains Corda-specific tables, while it should be empty. $EXISTING_STATE_GENERIC_WARNING"
+            tablesUpperCaseResultSet.use {
+                check(!tablesUpperCaseResultSet.first()) {
+                    "The database contains Corda-specific tables, while it should be empty. $EXISTING_STATE_GENERIC_WARNING"
+                }
             }
         } catch (exception: SQLException) {
             throw Exception("An error occurred whilst connecting to \"${conf.dataSourceProperties.getProperty("dataSource.url")}\". ", exception)
```
thschroeter: not sure this catch block is needed any more.

thschroeter: also if we catch a SQL exception here we're not having an error whiles connecting any more, since we connect above. At least the error message needs updating. But i think we can get rid of the catch block.

dimosr: 1. I think the catch is still needed. The reason is after registration the node will start up, so database will have to be configured properly anyway.
2. Yes, in that case, it might be worth slightly updating the message. But, I have a memory that this is also thrown during the datasource retrieval, it's just wrapped from Hikari, so it's not visible at compile time. It'd be easy to check that by running the code though, I'll also try to have a look at the source code to see if I can find this place.

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
```diff
@@ -98,17 +98,26 @@ class InitialRegistration(val baseDirectory: Path, private val networkRootTrustS
...
+        } finally {
+            datasource.connection.close()
```
dimosr: are we closing the same connection here or a new one ?

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
dimosr: An update regarding 2, I re-run the test with an incorrectly configured DB and the exception is actually thrown from the connection instantiation. Datasource is initialised without errors, so we're good there.

dimosr:
@thschroeter I added the changes we discussed offline. These are mainly about using assertJ for exceptions to avoid masking failures, closing connections in the tests and a small issue around datasource retrieval that was masked.

node/src/main/kotlin/net/corda/node/internal/subcommands/InitialRegistrationCli.kt
dimosr: I believe this invokes the getter in Java, using a new connection. I removed the `finally` and used a `use` instead to avoid some gymnastics that would be needed to retain this structure. 

Note that if an `SQLException` is thrown by the `getTables` method, it will still be handled by the existing catch in this way. Let me know what you think.

thschroeter:
@dimosr 👍 thanks. will ask Rick to take another look once i've deployed and tested our changes.

shamsasari:
The JIRA ticket is closed. What's the status of this PR?

PR #4961, opened 2019-03-29
make StatePointerSearch allow for null packages (like primitives)

roastario:
backport changes to allow StatePointerSearch to handle primitives.

Should we REALLY be doing this graph search of the entireTX without the user requesting it? There are some usecases where they might not want to attach all statepointer states ..

### DIFF
--- a/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
+++ b/core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
@@ -43,7 +43,7 @@ class StatePointerSearch(val state: ContractState) {
         val fieldsWithObjects = fields.mapNotNull { field ->
             // Ignore classes which have not been loaded.
             // Assumption: all required state classes are already loaded.
-            val packageName = field.type.`package`?.name
+            val packageName = field.type.packageNameOrNull
             if (packageName == null) {
                 null
             } else {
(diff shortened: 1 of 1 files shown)


roger-that-dev (approved):
Agree that we should make this optional - we can introduce a flag on transaction builder methods. Perhaps the default is to include reference states but developers can specify not to.

core/src/main/kotlin/net/corda/core/internal/StatePointerSearch.kt
```diff
@@ -85,6 +85,10 @@ class StatePointerSearch(val state: ContractState) {
         handleObject(newObj)
     }
 
+
+    //interesting - this seems to do a graph search of EVERY FIELD IN EVERY STATE in the TX
```
mikehearn: Can we put design discussion into a PR or mailing list thread rather than comments please.

Also note the comment style in the code style guide: https://docs.corda.net/codestyle.html#comments

mikehearn: Can we put design discussion into a PR or mailing list thread rather than comments please.

Also note the comment style in the code style guide: https://docs.corda.net/codestyle.html#comments

mikehearn (changes requested):

PR #4962, opened 2019-03-29
CORDA-2813 Handle loss of database

blsemo:
Add DbException handler to terminate process when the DB goes away or gives up.

Also add a helper to terminate the node and use it instead of calling Runtime.halt() in random places.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/utilities/errorAndTerminate.kt
+++ b/node/src/main/kotlin/net/corda/node/utilities/errorAndTerminate.kt
@@ -1,0 +1,17 @@
+package net.corda.node.utilities
+
+import net.corda.core.utilities.seconds
+import org.slf4j.LoggerFactory
+
+/**
+ * Log error message and terminate the process. This might not clean up resources and could leave
+ * the system in a messy state.
+ */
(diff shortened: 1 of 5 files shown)


node/src/main/kotlin/net/corda/node/utilities/errorAndTerminate.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import net.corda.core.utilities.seconds
+import org.slf4j.LoggerFactory
+
+/**
+ * Log error message and terminate the process. This might not clean up resources and could leave
+ * the system in a messy state.
+ */
+fun errorAndTerminate(message: String, e: Throwable?) {
+    val log = LoggerFactory.getLogger("errorAndTerminate")
```
rick-r3: Let's wrap the logging in a `thread { }`.  I've seen this attempt to halt get blocked by the logger blocking in other projects I've worked on.  You then head off down the path of "What if I can't create a thread?", but we have to stop somewhere and hopefully that will result in OOM which we can tell the JVM to exit on!

Oh, and make `errorAndTerminate` `@Synchronized` so we only log/start one shutdown sequence.

PR #4964, opened 2019-03-29
CORDA-2647 ROLLBACK

thschroeter:


shamsasari:
Please provide information on why this was rolled back for the benefit of others. There's also no explanation in https://r3-cev.atlassian.net/browse/CORDA-2647, which is still marked as done.

thschroeter:
The reason of the rollback is that initial registration doesn't work on some databases. See https://r3-cev.atlassian.net/browse/CORDA-2814.

PR #4967, opened 2019-04-01
CORDA-2632 Handle exceptions when file does not exist.

StefanIliev545:
Added a catch clause for the specific case of StringToMethodCallParser.UnparseableCallException.FailedParse which is only thrown during argument parsing.
CORDA-2632

### DIFF
--- a/client/jackson/src/main/kotlin/net/corda/client/jackson/StringToMethodCallParser.kt
+++ b/client/jackson/src/main/kotlin/net/corda/client/jackson/StringToMethodCallParser.kt
@@ -127,7 +127,7 @@ open class StringToMethodCallParser<in T : Any> @JvmOverloads constructor(
         return method.parameters.mapIndexed { index, param ->
             when {
                 param.isNamePresent -> param.name
-            // index + 1 because the first Kotlin reflection param is 'this', but that doesn't match Java reflection.
+                // index + 1 because the first Kotlin reflection param is 'this', but that doesn't match Java reflection.
                 kf != null -> kf.parameters[index + 1].name ?: throw UnparseableCallException.ReflectionDataMissing(method.name, index)
                 else -> throw UnparseableCallException.ReflectionDataMissing(method.name, index)
             }
@@ -153,7 +153,7 @@ open class StringToMethodCallParser<in T : Any> @JvmOverloads constructor(
         class MissingParameter(methodName: String, val paramName: String, command: String) : UnparseableCallException("Parameter $paramName missing from attempt to invoke $methodName in command: $command")
         class TooManyParameters(methodName: String, command: String) : UnparseableCallException("Too many parameters provided for $methodName: $command")
         class ReflectionDataMissing(methodName: String, argIndex: Int) : UnparseableCallException("Method $methodName missing parameter name at index $argIndex")
-        class NoSuchFile(filename: String) : UnparseableCallException("File ${filename} not found")
+        class NoSuchFile(filename: String) : UnparseableCallException("File $filename not found")
         class FailedParse(e: Exception) : UnparseableCallException(e.message ?: e.toString(), e)
     }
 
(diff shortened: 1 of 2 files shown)


tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -497,6 +497,12 @@ object InteractiveShell {
                     throw e.rootCause
                 }
             }
+        } catch(e: StringToMethodCallParser.UnparseableCallException.FailedParse) {
+            out.println("Failed while parsing arguments.", Color.red)
```
anthonykeenan: Does it not make more sense to print the original exception message instead?

Does it als not make sense to continue printing out `Please try 'man run' to learn what syntax is acceptable` like it does for the more generic exception below?

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -497,6 +497,12 @@ object InteractiveShell {
...
+            val cause = e.cause
+            if (cause is java.nio.file.NoSuchFileException) {
+                out.println("File not found.")
```
anthonykeenan: Is there any way we can say where we can print the failed file name out here? For example if there are two file inputs to the command? Or at least the parameter name that had the incorrect argument supplied?

StefanIliev545: That's a good idea, the exception should be easily reworkable to include the file on which the argument on which the parsing failed.

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -499,7 +499,9 @@ object InteractiveShell {
             }
         } catch (e: StringToMethodCallParser.UnparseableCallException) {
             out.println(e.message, Color.red)
-            out.println("Please try 'man run' to learn what syntax is acceptable")
+            if (e !is StringToMethodCallParser.UnparseableCallException.NoSuchFile) {
+                out.println("Please try 'man run' to learn what syntax is acceptable")
```
StefanIliev545: For that one I'm not sure if there is a case where something weird is required as a file for the command that might actually need man explaining.

client/jackson/src/main/kotlin/net/corda/client/jackson/StringToMethodCallParser.kt
```diff
@@ -153,6 +153,7 @@ open class StringToMethodCallParser<in T : Any> @JvmOverloads constructor(
         class MissingParameter(methodName: String, val paramName: String, command: String) : UnparseableCallException("Parameter $paramName missing from attempt to invoke $methodName in command: $command")
         class TooManyParameters(methodName: String, command: String) : UnparseableCallException("Too many parameters provided for $methodName: $command")
         class ReflectionDataMissing(methodName: String, argIndex: Int) : UnparseableCallException("Method $methodName missing parameter name at index $argIndex")
+        class NoSuchFile(filename: String) : UnparseableCallException("File ${filename} not found")
```
anthonykeenan: $filename

client/jackson/src/main/kotlin/net/corda/client/jackson/StringToMethodCallParser.kt
```diff
@@ -200,6 +201,8 @@ open class StringToMethodCallParser<in T : Any> @JvmOverloads constructor(
             val entryType = om.typeFactory.constructType(argType)
             try {
                 om.readValue<Any>(entry.traverse(om), entryType)
+            } catch( e : java.nio.file.NoSuchFileException) {
```
anthonykeenan: `} catch (e: java.nio.File.NoSuchFileException)`

Try and do a pass with the formatter before you commit!

PR #4969, opened 2019-04-01
CORDA-2832: Improved error reporting in interactive shell when an error occurs after a ctor is matched.

adelel1:
The interactive shell implementation will try to match the passed in flow start argument list with each flow constructor in turn. Apart from the matching constructor there will be either too few or too many arguments specified, or missing parameters. All these errors are accumulated into a list.

If the matching constructor also fails for a legitimate reason, e.g. the Party cannot be found then this error is also added to the cumulative list and the whole list is sent back to the user and displayed in the shell. This is not right, only the errors related to the matching constructor should be displayed. This is now done in this PR, if there is no matching constructor then all errors are returned as before.

Note that we do not support two matching constructors with the same parameter names but different only by type. However the previous implementation would allow this. The first matching constructor that succeeded would be used, previous matching constructors that failed would be ignored. In this implementation if the first matching constructor fails then the error is reported.

### DIFF
--- a/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
+++ b/tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
@@ -389,17 +389,31 @@ object InteractiveShell {
                               inputData: String,
                               clazz: Class<out FlowLogic<T>>,
                               om: ObjectMapper): FlowProgressHandle<T> {
-        // For each constructor, attempt to parse the input data as a method call. Use the first that succeeds,
-        // and keep track of the reasons we failed so we can print them out if no constructors are usable.
-        val parser = StringToMethodCallParser(clazz, om)
-        var allErrors = ArrayList<String>()
-        val errors = ArrayList<String>()
-        var constructorFound = false
 
+        val errors = ArrayList<String>()
+        val parser = StringToMethodCallParser(clazz, om)
+        val nameTypeList = getMatchingConstructorParamsAndTypes(parser, inputData, clazz, om)
+
+        try {
+            val args = parser.parseArguments(clazz.name, nameTypeList, inputData)
+            return invoke(clazz, args)
+        } catch (e: StringToMethodCallParser.UnparseableCallException.ReflectionDataMissing) {
+            val argTypes = nameTypeList.map { (name, type) -> type }
+            errors.add("$argTypes: <constructor missing parameter reflection data>")
+        } catch (e: StringToMethodCallParser.UnparseableCallException) {
+            val argTypes = nameTypeList.map { (name, type) -> type }
+            errors.add("$argTypes: ${e.message}")
+        }
+        throw NoApplicableConstructor(errors)
+    }
+
+    private fun <T> getMatchingConstructorParamsAndTypes(parser: StringToMethodCallParser<FlowLogic<T>>, inputData: String, clazz: Class<out FlowLogic<T>>,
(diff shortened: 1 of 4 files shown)


tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -392,10 +392,13 @@ object InteractiveShell {
         // For each constructor, attempt to parse the input data as a method call. Use the first that succeeds,
         // and keep track of the reasons we failed so we can print them out if no constructors are usable.
         val parser = StringToMethodCallParser(clazz, om)
+        var allErrors = ArrayList<String>()
         val errors = ArrayList<String>()
+        var constructorFound = false
```
chrisr3: This `Boolean` bothers me because it makes this function's control-flow a lot less obvious. This function has three parts:
- selecting the correct constructor
- parsing the constructor arguments
- invoking this constructor with the arguments

Unfortunately, this is being obscured by this "One Try-Catch Block To Rule Them All!", which is really just handling  `StringToMethodCallParser.UnparseableCallException` and its descendants. I'm sure that this can be simplified, e.g.
```kotlin
val nameTypeList = run {
    val allErrors = ArrayList<String>()
    for (ctor in clazz.constructors) {
        try {
            // etc
            return@run nameTypeList
        } catch (...) {
            // etc
        }
    }
    throw NoApplicableConstructor(allErrors)
}

try {
    val args = parser.parseArguments(clazz.name, nameTypeList, inputData)
    return invoke(clazz, args)
} catch (...) {
    throw NoApplicableConstructor(...)
}
```
The problem (of course) is these multi-layered lambdas for generating all of the error messages! Is there any way that you can untangle this function please?

adelel1: Hi,
Points noted. I have simplified code. Ready for your review.
Thanks, Adel

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
```diff
@@ -389,14 +389,31 @@ object InteractiveShell {
...
+            errors.add("$argTypes: <constructor missing parameter reflection data>")
+        } catch (e: StringToMethodCallParser.UnparseableCallException) {
+            val argTypes = nameTypeList.map { (name, type) -> type }
+            errors.add("$argTypes: ${e.message}")
+        }
+        throw NoApplicableConstructor(errors)
+    }
 
+    private fun <T> getMatchingConstructorParamsAndTypes(parser: StringToMethodCallParser<FlowLogic<T>>, inputData: String, clazz: Class<out FlowLogic<T>>,
+                                            om: ObjectMapper) : List<Pair<String, Type>> {
```
chrisr3: This function doesn't use its `om` parameter, so can you remove it please?
But the line defining this function's prototype is a bit _long_, so can you reformat it too? The `runFlowFromString` function is a good example.

chrisr3 (changes requested):
Nice; small things though.

tools/shell/src/main/kotlin/net/corda/tools/shell/InteractiveShell.kt
adelel1: All updates done, ready for your review.
Thanks,

adelel1:
Ready for review

chrisr3 (approved):

PR #4976, opened 2019-04-02
CORDA-2519 Changed crash version to our latest.

StefanIliev545:
Changed library version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,7 @@ buildscript {
     ext.dependency_checker_version = '4.0.0'
     ext.commons_collections_version = '4.1'
     ext.beanutils_version = '1.9.3'
-    ext.crash_version = '810d2b774b'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -53,7 +53,7 @@ buildscript {
     ext.dependency_checker_version = '4.0.0'
     ext.commons_collections_version = '4.1'
     ext.beanutils_version = '1.9.3'
-    ext.crash_version = 'cadb53544fbb3c0fb901445da614998a6a419488'
```
anthonykeenan: Why is the hash much smaller for the current release? What has changed?

StefanIliev545: The full commit hash wouldn't work for some reason. I tried the short one that was in jitpack and it worked. Guessed it's an issue on their side (or was, maybe it's alright now) https://github.com/jitpack/jitpack.io/issues/11

StefanIliev545: @anthonykeenan it builds now as i supposed with the long revision which can be found in the last commit.

PR #4981, opened 2019-04-03
Tidy up codesets in contract constraints documentation

lankydan:
Code for these docs had the java and kotlin samples but were not combined into a single codeset with tabs for each language.


josecoll:
Nice cleanup!

PR #4982, opened 2019-04-03
 CORDA-2821 Fix issue with Quasar errors redirecting to useless page

StefanIliev545:
Just changed the url

### DIFF
--- a/docs/source/tutorial-cordapp.rst
+++ b/docs/source/tutorial-cordapp.rst
@@ -486,41 +486,7 @@ You can run the CorDapp's integration tests by running the ``Run Integration Tes
 Running tests in IntelliJ
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-We recommend editing your IntelliJ preferences so that you use the Gradle runner - this means that the quasar utils
-plugin will make sure that some flags (like ``-javaagent`` - see :ref:`below <tutorial_cordapp_alternative_test_runners>`) are
-set for you.
-
-To switch to using the Gradle runner:
-
-* Navigate to ``Build, Execution, Deployment -> Build Tools -> Gradle -> Runner`` (or search for `runner`)
-
-  * Windows: this is in "Settings"
-  * MacOS: this is in "Preferences"
-
-* Set "Delegate IDE build/run actions to gradle" to true
-* Set "Run test using:" to "Gradle Test Runner"
-
-.. _tutorial_cordapp_alternative_test_runners:
-
-If you would prefer to use the built in IntelliJ JUnit test runner, you can add some code to your ``build.gradle`` file and
-it will copy your quasar JAR file to the lib directory. You will also need to specify ``-javaagent:lib/quasar.jar``
-and set the run directory to the project root directory for each test.
-
-Add the following to your ``build.gradle`` file - ideally to a ``build.gradle`` that already contains the quasar-utils plugin line:
-
-.. sourcecode:: groovy
-
-    apply plugin: 'net.corda.plugins.quasar-utils'
-
-    task installQuasar(type: Copy) {
-        destinationDir rootProject.file("lib")
-        from(configurations.quasar) {
-            rename 'quasar-core(.*).jar', 'quasar.jar'
-        }
-    }
-
-
-and then you can run ``gradlew installQuasar``.
+See :ref:`Running tests in IntelliJ<tutorial_cordapp_alternative_test_runners>`
 
 Debugging your CorDapp
 ----------------------
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
@@ -309,7 +309,7 @@ class SingleThreadedStateMachineManager(
     private fun checkQuasarJavaAgentPresence() {
         check(SuspendableHelper.isJavaAgentActive()) {
             """Missing the '-javaagent' JVM argument. Make sure you run the tests with the Quasar java agent attached to your JVM.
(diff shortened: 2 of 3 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/SingleThreadedStateMachineManager.kt
```diff
@@ -309,7 +309,7 @@ class SingleThreadedStateMachineManager(
     private fun checkQuasarJavaAgentPresence() {
         check(SuspendableHelper.isJavaAgentActive()) {
             """Missing the '-javaagent' JVM argument. Make sure you run the tests with the Quasar java agent attached to your JVM.
-               #See https://docs.corda.net/troubleshooting.html - 'Fiber classes not instrumented' for more details.""".trimMargin("#")
+               #See https://docs.corda.net/head/tutorial-cordapp.html?highlight=quasar#running-tests-in-intellij - 'Fiber classes not instrumented' for more details.""".trimMargin("#")
```
josecoll: I'm all for this change but I would like to see the information displayed once and consistently in the actual troubleshooting page.

The new link above (https://docs.corda.net/head/tutorial-cordapp.html?highlight=quasar#running-tests-in-intellij) has a sentence that states:
"You will also need to specify -javaagent:lib/quasar.jar and set the run directory to the project root directory for each test."
This is actually incorrect and should be replaced with the general Junit configuration advise given here:
https://docs.corda.net/head/tutorial-building-transactions.html?highlight=javaagent%20lib%20quasar%20jar#gathering-inputs (see NOTE)

anthonykeenan: The original troubleshooting page was removed by @joeldudleyr3, presumably as part of a move towards directing people to StackOverflow.

IMO we should either point directly to a comprehensive SO page that contains an answer for this, or we should re-introduce it back to the docsite. Although I'd be worried about hardocoding stackoverflow links into the product. What do you think @joeldudleyr3??

StefanIliev545: The stackoverflow https://stackoverflow.com/questions/47411363/error-when-running-corda-flow-tests-from-intellij but somehow the tutorial cordapp seemed easier to read to me.

josecoll: Well, even the SO page doesn't give you the really useful advice:

`Before creating the IntelliJ run configurations for these unit tests go to Run -> Edit Configurations -> Defaults -> JUnit, add -javaagent:lib/quasar.jar to the VM options, and set Working directory to $PROJECT_DIR$ so that the Quasar instrumentation is correctly configured.`

StefanIliev545: Should we edit it then? We can update the documentation but I'm not sure if developers reading it will expect to find info on the issue in /tutorial-cordapp link, apart for those brought there by the console message.

josecoll: @chrischabot, who runs Dev Relations, might like to comment on how we handle these scenarios going forward.
My personal opinion is:
- we should not embed SO URL's in out exception messages
- we should include obvious/constant trip-me-up issues to our Troubleshooting page (the one in this PR)
- everything else goes to SO
The developer experience should be really simple, and SO should not be a continuous part of that experience for trivialities.

anthonykeenan: FYI this is the PR where the troubleshooting page was removed: https://github.com/corda/corda/pull/4071 - I don't want to add it back if it's been deleted for a reason, but I agree we've made things more difficult for people here.

josecoll: Sure - I think we just don't have the balance right.
And furthermore, far reaching changes in documentation structure and approach should not be authorised (approved) by one person only - eg. the previous PR you referenced.

anthonykeenan: Ok, in the interests of making some progress on this, I think we should do the following:

Combine the information in the hello world tutorial and the building transactions tutorial into a section on the https://docs.corda.net/testing.html page, which seems like a logical place for it.

Point both the above pages _and_ the original error to the testing page so the information is all in one place.

StefanIliev545: Updated in latest commit. Notice that I duplicated the note, wasn't sure whether to link to the whole thing, seemed rather an overkill.

docs/source/tutorial-cordapp.rst
```diff
@@ -486,41 +486,7 @@ You can run the CorDapp's integration tests by running the ``Run Integration Tes
...
-    task installQuasar(type: Copy) {
-        destinationDir rootProject.file("lib")
-        from(configurations.quasar) {
-            rename 'quasar-core(.*).jar', 'quasar.jar'
-        }
-    }
-
-
-and then you can run ``gradlew installQuasar``.
+See :ref:`Running tests in IntelliJ<tutorial_cordapp_alternative_test_runners>`
```
anthonykeenan: Have you built the docs locally and checked whether this link works? (I think it doesn't)

StefanIliev545: I'll double check, I think it worked.

StefanIliev545: Yep, redirects to testing.html#tutorial-cordapp-alternative-test-runners

PR #4983, opened 2019-04-03
CORDA-2645 Do not remove exception information in dev mode

lankydan:
`ExceptionSerialisingRpcOpsProxy` was removing information about the original exception in dev mode. This code has been removed. Although there is no check in `ExceptionSerialisingRpcOpsProxy` for dev mode (and also due to it being about serialising), `ExceptionMaskingRpcOpsProxy` will handle the removal of exception data in non dev mode (production mode)

### DIFF
--- a/node/src/integration-test/kotlin/net/corda/node/services/rpc/RpcExceptionHandlingTest.kt
+++ b/node/src/integration-test/kotlin/net/corda/node/services/rpc/RpcExceptionHandlingTest.kt
@@ -21,13 +21,32 @@ import org.assertj.core.api.AssertionsForInterfaceTypes.assertThat
 import org.hibernate.exception.GenericJDBCException
 import org.junit.Test
 import java.sql.SQLException
+import kotlin.test.assertEquals
 
 class RpcExceptionHandlingTest {
     private val user = User("mark", "dadada", setOf(Permissions.all()))
     private val users = listOf(user)
 
     @Test
-    fun `rpc client receive client-relevant exceptions regardless of devMode`() {
+    fun `rpc client receives wrapped exceptions in devMode`() {
+        val params = NodeParameters(rpcUsers = users)
+        val clientRelevantMessage = "This is for the players!"
+
+        fun NodeHandle.throwExceptionFromFlow() {
+            rpc.startFlow(::ClientRelevantErrorFlow, clientRelevantMessage).returnValue.getOrThrow()
+        }
+
+        driver(DriverParameters(startNodesInProcess = true, notarySpecs = emptyList())) {
+            val devModeNode = startNode(params, BOB_NAME).getOrThrow()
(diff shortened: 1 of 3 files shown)


node/src/integration-test/kotlin/net/corda/node/services/rpc/RpcExceptionHandlingTest.kt
```diff
@@ -21,13 +21,32 @@ import org.assertj.core.api.AssertionsForInterfaceTypes.assertThat
 import org.hibernate.exception.GenericJDBCException
 import org.junit.Test
 import java.sql.SQLException
+import kotlin.test.assertEquals
 
 class RpcExceptionHandlingTest {
```
shamsasari: Can you try and migrate this to a unit test class using mock network. It seems overkill to be using the driver to test this.

lankydan:
I believe so yes, https://r3-cev.atlassian.net/browse/CORDA-2740 might be taking it a bit further so should have another look at it. There is the possibility that 2740 can be closed with no further changes if the changes in this PR cover everything.

node/src/integration-test/kotlin/net/corda/node/services/rpc/RpcExceptionHandlingTest.kt
```diff
@@ -21,13 +21,32 @@ import org.assertj.core.api.AssertionsForInterfaceTypes.assertThat
...
+        val params = NodeParameters(rpcUsers = users)
+        val clientRelevantMessage = "This is for the players!"
+
+        fun NodeHandle.throwExceptionFromFlow() {
+            rpc.startFlow(::ClientRelevantErrorFlow, clientRelevantMessage).returnValue.getOrThrow()
+        }
+
+        driver(DriverParameters(startNodesInProcess = true, notarySpecs = emptyList())) {
+            val devModeNode = startNode(params, BOB_NAME).getOrThrow()
+            assertThatThrownBy { devModeNode.throwExceptionFromFlow() }.isInstanceOfSatisfying(ClientRelevantException::class.java) { exception ->
```
shamsasari: To confirm, the stacktrace even in devMode doesn't include the node code?

lankydan: Yes, the stacktrace is removed somewhere else in the serialization code as well. That code does not depend on the `devMode` flag meaning it is always removed.

shamsasari: Is it empty, in which case can you assert that?

lankydan: It is covered by one of the other tests, that being said, I have added an explicit test for it since it is pretty important.

shamsasari:
Can't figure out why TC is not building this. Easiest solution is probably just to re-submit under a different branch.

PR #4984, opened 2019-04-04
CORDA-2827 extend timeout on test

r3domfox:
The default 5-second timeout for `eventually` isn't always sufficient for `RetryFlowMockTest.Restart does not set senderUUID` to complete its work, so we extend it to 30 seconds, with a 100ms retry interval.


node/src/test/kotlin/net/corda/node/services/statemachine/RetryFlowMockTest.kt
```diff
@@ -101,7 +101,7 @@ class RetryFlowMockTest {
         })
         val count = 10000 // Lots of iterations so the flow keeps going long enough
         nodeA.startFlow(KeepSendingFlow(count, partyB))
-        eventually(waitBetween = Duration.ofMillis(10)) {
+        eventually(duration = Duration.ofSeconds(30), waitBetween = Duration.ofMillis(100)) {
```
blsemo: Not sure this first thing needs to wait for 30s, and should probably stay on 10ms repeats?

PR #4985, opened 2019-04-04
Willh docupdate

willhr3:
Adding guidance for mac users when using the `runnodes` task as it fails to start nodes when they're opened in separate windows.


vkolomeyko:
@joeldudleyr3 - since @roger3cev (team TL) is on holidays would you be able to merge this *docs* PR in his absence?

PR #4988, opened 2019-04-05
CORDA-2824: Do not start the P2P consumer until we have at least one registered handler

mnesbit:
This modification just pauses the P2P consumer start until the state machine has registered its message handler, which in practice requires network map checks to have finished.

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
+++ b/node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
@@ -133,6 +133,7 @@ class P2PMessagingClient(val config: NodeConfiguration,
     private val delayStartQueues = Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())
 
     private val handlers = ConcurrentHashMap<String, MessageHandler>()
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/services/messaging/P2PMessagingClient.kt
```diff
@@ -528,12 +534,18 @@ class P2PMessagingClient(val config: NodeConfiguration,
...
+        synchronized(handlersChangedSignal) {
+            handlersChangedSignal.notifyAll()
+        }
         return HandlerRegistration(topic, callback)
     }
 
     override fun removeMessageHandler(registration: MessageHandlerRegistration) {
         registration as HandlerRegistration
         handlers.remove(registration.topic)
+        synchronized(handlersChangedSignal) {
```
exFalso: Do we need this one?

mnesbit: probably not, but I wanted to include some symmetry, I don't think we ever care right now but might if we change to do more intelligent failover.

PR #4990, opened 2019-04-05
[CORDA-2837] Ensure certain flows cannot start a session with the local node

JamesHR3:
When calling the `initiateFlow` API, it is possible to specify the local node as the counterparty. There are some situations where this is desirable, but there are other flows for which this does not make sense. In the platform, `CollectSignaturesFlow` and `SwapIdentitiesFlow` fall into this second category. This PR adds some logic to prevent these flows from using the local node as a counterparty.

Testing:
 - Unit tests
 - Ran the Obligation sample and transferred an obligation to the borrower (so the same node is the borrower and lender). This causes the new logic in `CollectSignaturesFlow` to be hit.
 - Tested the `CashPaymentFlow` when paying the local node in the `SwapIdentitiesFlow` case. There's also a fix for the `CashPaymentFlow` included in this PR.

I have a few points I'd like to make sure get double checked:
 - The change to `SwapIdentitiesFlow` here represents an API behaviour change (I've had to change the finance CorDapp to accommodate it). Is that a safe change to make? Would it benefit from being behind a `targetPlatformVersion` gate? (Note that the same is not true of the `CollectSignaturesFlow`, which doesn't work at the moment if a session to the local node is passed in. This PR just makes the failure happen earlier and is more explicit about what went wrong.)
 - Is this compatible with the plans for accounts?

### DIFF
--- a/confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
+++ b/confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
@@ -95,6 +95,9 @@ private constructor(private val otherSideSession: FlowSession?,
                     "SwapIdentitiesFlow with FlowSessions. (${CordappResolver.currentCordapp?.info})")
             initiateFlow(otherParty!!)
         }
+        require(!serviceHub.myInfo.isLegalIdentity(session.counterparty)) {
+            "Do not attempt to sync identities with the local node"
+        }
         progressTracker.currentStep = GENERATING_IDENTITY
         val ourAnonymousIdentity = serviceHub.keyManagementService.freshKeyAndCert(ourIdentityAndCert, false)
         val data = buildDataToSign(ourAnonymousIdentity)
--- a/confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
+++ b/confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
@@ -95,8 +95,8 @@ private constructor(private val otherSideSession: FlowSession?,
                     "SwapIdentitiesFlow with FlowSessions. (${CordappResolver.currentCordapp?.info})")
             initiateFlow(otherParty!!)
         }
-        require(!serviceHub.myInfo.isLegalIdentity(session.counterparty)) {
-            "Do not attempt to sync identities with the local node"
+        if (serviceHub.myInfo.isLegalIdentity(session.counterparty)) {
+            throw SwapIdentitiesException("Attempted to swap identities with the local node")
         }
         progressTracker.currentStep = GENERATING_IDENTITY
         val ourAnonymousIdentity = serviceHub.keyManagementService.freshKeyAndCert(ourIdentityAndCert, false)
--- a/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
@@ -91,6 +91,10 @@ class CollectSignaturesFlow @JvmOverloads constructor(val partiallySignedTx: Sig
             "The Initiator of CollectSignaturesFlow must have signed the transaction."
         }
 
+        require(sessionsToCollectFrom.none { serviceHub.myInfo.isLegalIdentity(it.counterparty) }) {
+            "Do not call CollectSignaturesFlow with sessions for the local node - the initiator should have already signed the transaction"
+        }
+
         // The signatures must be valid and the transaction must be valid.
         partiallySignedTx.verifySignaturesExcept(notSigned)
         partiallySignedTx.tx.toLedgerTransaction(serviceHub).verify()
--- a/finance/workflows/src/main/kotlin/net/corda/finance/flows/CashPaymentFlow.kt
+++ b/finance/workflows/src/main/kotlin/net/corda/finance/flows/CashPaymentFlow.kt
@@ -53,7 +53,11 @@ open class CashPaymentFlow(
         val recipientSession = initiateFlow(recipient)
         recipientSession.send(anonymous)
         val anonymousRecipient = if (anonymous) {
-            subFlow(SwapIdentitiesFlow(recipientSession))[recipient]!!
+            if (!serviceHub.myInfo.isLegalIdentity(recipient)) {
+                subFlow(SwapIdentitiesFlow(recipientSession))[recipient]!!
+            } else {
+                serviceHub.keyManagementService.freshKeyAndCert(ourIdentityAndCert, false).party.anonymise()
+            }
         } else {
             recipient
         }
--- a/confidential-identities/src/test/kotlin/net/corda/confidential/SwapIdentitiesFlowTests.kt
+++ b/confidential-identities/src/test/kotlin/net/corda/confidential/SwapIdentitiesFlowTests.kt
@@ -72,7 +72,7 @@ class SwapIdentitiesFlowTests {
     fun `cannot swap identities with local node`() {
         assert.that(
                 aliceNode.services.startFlow(SwapIdentitiesInitiator(alice)),
-                willThrow<IllegalArgumentException>()
+                willThrow<SwapIdentitiesException>()
         )
     }
 
--- a/confidential-identities/src/test/kotlin/net/corda/confidential/SwapIdentitiesFlowTests.kt
+++ b/confidential-identities/src/test/kotlin/net/corda/confidential/SwapIdentitiesFlowTests.kt
@@ -17,11 +17,13 @@ import net.corda.core.identity.PartyAndCertificate
 import net.corda.testing.core.*
 import net.corda.testing.internal.matchers.allOf
 import net.corda.testing.internal.matchers.flow.willReturn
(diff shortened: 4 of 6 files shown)


josecoll:
So there are 2 updates here that sounds perfectly sensible:

1. `CollectSignaturesFlow` which was broken as surfaced by the Obligations CorDapp.
2. `SwapIdentitiesFlow` which semantically falls into the same problem, but *was* working previously (eg. you could transfer Cash to yourself with no problem).

The change to `CashPaymentFlow` would seem to produce the same outcome (albeit with local generation of the anonymous key pair), so the fix looks good.
You rightly point out this is a semantic break from the point of view of behaviour previously working (that shouldn't) is now being trapped/thrown. 
@shamsasari what would you recommend in this scenario? (gating does feel like the right thing to do in my mind, but would now mean gating on mpv = 5)

finance/workflows/src/main/kotlin/net/corda/finance/flows/CashPaymentFlow.kt
```diff
@@ -53,7 +53,11 @@ open class CashPaymentFlow(
         val recipientSession = initiateFlow(recipient)
         recipientSession.send(anonymous)
         val anonymousRecipient = if (anonymous) {
-            subFlow(SwapIdentitiesFlow(recipientSession))[recipient]!!
+            if (!serviceHub.myInfo.isLegalIdentity(recipient)) {
```
josecoll: Feel like you may need to apply the same logic on `TwoPartyDealFlow` L58-60

confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
```diff
@@ -95,6 +95,9 @@ private constructor(private val otherSideSession: FlowSession?,
                     "SwapIdentitiesFlow with FlowSessions. (${CordappResolver.currentCordapp?.info})")
             initiateFlow(otherParty!!)
         }
+        require(!serviceHub.myInfo.isLegalIdentity(session.counterparty)) {
```
josecoll: Should we not be throwing a `SwapIdentitiesException` in keeping with other Exceptions thrown by this flow?

core/src/main/kotlin/net/corda/core/flows/CollectSignaturesFlow.kt
```diff
@@ -91,6 +91,10 @@ class CollectSignaturesFlow @JvmOverloads constructor(val partiallySignedTx: Sig
             "The Initiator of CollectSignaturesFlow must have signed the transaction."
         }
 
+        require(sessionsToCollectFrom.none { serviceHub.myInfo.isLegalIdentity(it.counterparty) }) {
```
josecoll: Hmmm .... seems like in this Flow code we are using the general `IllegalArgumentException` (triggered by require), so not very consistent Exception handling in our code ;-)

josecoll:
@shamsasari (re: Exception handling in flows), can we raise a story to ensure that Exceptions thrown by flow are documented in the API (and where possible, a consistent Exception definition/usage guideline is adopted).
eg.  in this PR I noticed:
`SwapIdentitiesFlow` throws `SwapIdentitiesException` exceptions (not documented in API)
`CollectSignaturesFlow` uses `require` that throws `IllegalArgumentException` exceptions (not documented in API)

JamesHR3:
As an addition to @josecoll 's comment, I think the flow framework handles anything based on `FlowException` differently from other exceptions - see https://docs.corda.net/api-flows.html#flowexception. It's probably worth reviewing that we're doing the right thing with different exception types.

I can raise the story as Shams is away today.

gendal:
i've not reviewed this but i know @roger3cev  and @roastario are building some work right now that makes various assumptions about how you start flow sessions... can you quickly check with one of them that nothing in this conflicts with what they need please? thanks!

finance/workflows/src/main/kotlin/net/corda/finance/flows/CashPaymentFlow.kt
JamesHR3: It turns out convincing `TwoPartyDealFlow` to work when the counterparty is yourself is harder than just this change. I'm most of the way there but I think this might end up being quite a sizeable change to the structure of the flow, and as a result there's a better chance something could go wrong. I'm wondering if we might be better off throwing an exception if the counterparty is the same as us in this case too? Are there cases where it makes sense to propose a deal to yourself?

(I think the behaviour of this flow without the other changes proposed in this PR would be to fail partway through collecting signatures with the same error as in the original Jira ticket.)

confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
JamesHR3: I've opted to try and keep the exception types the same within each flow for this PR. Sounds like we might need to do a pass to make sure we're being consistent across flows.

confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt
```diff
@@ -95,6 +95,9 @@ private constructor(private val otherSideSession: FlowSession?,
                     "SwapIdentitiesFlow with FlowSessions. (${CordappResolver.currentCordapp?.info})")
             initiateFlow(otherParty!!)
         }
+        if (serviceHub.myInfo.isLegalIdentity(session.counterparty)) {
```
shamsasari: The V3 behaviour in this scenario is to generate just one identity and not have any flow I/O: https://github.com/corda/corda/blob/release-V3/confidential-identities/src/main/kotlin/net/corda/confidential/SwapIdentitiesFlow.kt#L81. V3 only supported `Party` objects so in our case this special handling has to be done when `otherSideSession` is null (i.e. `otherParty` is non-null). It seems I forgot about this case when I was updating SIF for V4.

For the scenario when a flow session is passed then I'm not actually sure we need to be changing anything.  It's not clear to me how the current behaviour is broken.  Or even how the original JIRA ticket is related to this. It talks about missing responder flows.

JamesHR3: The original ticket has quite a lot going on. One of the issues highlighted was that `CollectSignaturesFlow` doesn't work if one of the sessions is to the local node (you'll get an error saying you've passed the wrong sessions to the flow, because as the initiator you've already signed the transaction). That could be fixed by filtering out any sessions to the local node, but it doesn't make much sense to collect a signature from yourself, so instead the flow has been changed to raise an error detecting this case.

The argument is that the same is true of `SwapIdentitiesFlow` - it doesn't make much sense to swap an identity with yourself, so the fact that this is allowed represents (possibly) a bug. I think this PR implements the equivalent of the V3 logic in `CashPaymentFlow` (i.e. pushing that logic to the calling flow).

There's possibly an argument to be made that these flows should be accepting sessions to themselves. The case I'm concerned about is accounts, which could potentially result in multiple parties on the same node, so a session might be opened to one of the local node's identities. This is the case that I think Richard is getting at, and that I'd like Roger or Stefano to look over. Unfortunately they're both away today.

shamsasari: In V3 you could "swap" an anonymous identity with yourself. Sure, the term "swapping" doesn't make sense in this scenario, but the end result was that you got an anonymous identity if you wanted an anonymous tx involving yourself. The change here breaks that behaviour, and so this check needs to be reverted. And since it worked in V3 it doesn't make sense to gate it behind target version.

The other point that I was trying to get at was that in V3, the self scenario just generated one anonymous identity and returned that directly (without doing any actual  "swapping"). That behaviour got wiped (my me accidently) and show be returned. It's simply a matter of before initiating the session (`initiateFlow(otherParty!!)`), check if `otherParty` is self and just return the generated identity. This change can be done in a separate PR though.

JamesHR3:
I had a quick chat with @roger3cev. It sounds like with accounts it would become an expected condition that the session is set up with the local node. If that's the case, then this PR is probably going the wrong way and instead we'll need to ensure the flows we provide can handle receiving sessions to the local node.

I'm aware of the following flows that don't work if a local session is provided:
 - `FinalityFlow`
 - `CollectSignaturesFlow`

`SwapIdentitiesFlow`, as @shamsasari points out, generates two identities instead of one if a session to the local node is provided (this PR effectively reinstates the V3 behaviour for `CashPaymentFlow` specifically).

There's a few other flows that I'm less sure about, but I think should work:
 - `DataVendingFlow`
 - `ReceiveTransactionFlow`
 - `ResolveTransactionFlow`

(Making the changes described should also allow `TwoPartyDealFlow` to work with a local counterparty.)

roastario:
Hmmm, even ignoring Accounts - at the moment it's totally possible that a state owned by a CI identity on the local node is spent to a CI identity on the same node. 

I think we should make corda agnostic to whether the target node is local or not, rather than restricting sessions to remote nodes only.

JamesHR3:
I think that's the conclusion I've reached as well (although there's some interesting versioning concerns around that change). I'll close this PR, as I think it implements the wrong fix.

josecoll:
Great summary @JamesHR3, and given the discussions above, makes sense to re-target the original bug with a different approach.

PR #4992, opened 2019-04-05
CORDA-2834 Clarify error message when base directory doesn't exist

Bartman250:
https://r3-cev.atlassian.net/browse/CORDA-2834

# PR Checklist:

- [x] Have you run the unit, integration and smoke tests as described [here](https://docs.corda.net/head/testing.html)?
- [x] If you added public APIs, did you write the JavaDocs?
- [x] If the changes are of interest to application developers, have you added them to the [changelog](https://github.com/corda/corda/blob/master/docs/source/changelog.rst), and potentially the [release notes](https://github.com/corda/corda/blob/master/docs/source/release-notes.rst)?
- [x] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/corda/corda/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/corda/corda/blob/master/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -76,6 +76,24 @@ open class NodeStartupCli : CordaCliWrapper("corda", "Runs a Corda Node") {
 
     override fun additionalSubCommands() = setOf(networkCacheCli, justGenerateNodeInfoCli, justGenerateRpcSslCertsCli, initialRegistrationCli, validateConfigurationCli)
 
+    override fun call(): Int {
+        if (!validateBaseDirectory()) {
+            return ExitCodes.FAILURE
+        }
+        return super.call()
+    }
+
+    private fun validateBaseDirectory(): Boolean {
+        //Ensure that the base directory actually exists before initialising and the rest of the node.
+        val baseDirectory = cmdLineOptions.baseDirectory
+
+        if (!baseDirectory.exists()) {
+            printError("Base directory $baseDirectory does not exist. Node will now shutdown")
+            return false
+        }
+        return true
+    }
+
     override fun runProgram(): Int {
         return when {
             InitialRegistration.checkRegistrationMode(cmdLineOptions.baseDirectory) -> {
@@ -146,8 +164,10 @@ open class NodeStartup : NodeStartupLogging {
         Node.printBasicNodeInfo(LOGS_CAN_BE_FOUND_IN_STRING, System.getProperty("log-path"))
 
         // Step 5. Load and validate node configuration.
-        val rawConfig = cmdLineOptions.rawConfiguration().doOnErrors(cmdLineOptions::logRawConfigurationErrors).optional ?: return ExitCodes.FAILURE
-        val configuration = cmdLineOptions.parseConfiguration(rawConfig).doIfValid { logRawConfig(rawConfig) }.doOnErrors(::logConfigurationErrors).optional ?: return ExitCodes.FAILURE
+        val rawConfig = cmdLineOptions.rawConfiguration().doOnErrors(cmdLineOptions::logRawConfigurationErrors).optional
+                ?: return ExitCodes.FAILURE
+        val configuration = cmdLineOptions.parseConfiguration(rawConfig).doIfValid { logRawConfig(rawConfig) }.doOnErrors(::logConfigurationErrors).optional
+                ?: return ExitCodes.FAILURE
 
         // Step 6. Check if we can access the certificates directory
         if (requireCertificates && !canReadCertificatesDirectory(configuration.certificatesDirectory, configuration.devMode)) return ExitCodes.FAILURE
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -390,7 +390,8 @@ open class NodeStartup : NodeStartupLogging {
                     "When I discovered my toaster wasn't\nwaterproof, I was shocked.",
                     "Where do cryptographers go for\nentertainment? The security theatre.",
                     "How did the Java programmer get rich?\nThey inherited a factory.",
-                    "Why did the developer quit his job?\nHe didn't get ar-rays."
+                    "Why did the developer quit his job?\nHe didn't get ar-rays.",
+                    "Quantum computer jokes are funny and not funny at the same time"
```
anthonykeenan: The jokes have need a newline in the middle I think or they don't print right.

Should it not be "Quantum computer jokes are _both_ funny and not funny at the same time"

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -465,6 +466,10 @@ fun CliWrapperBase.initLogging(baseDirectory: Path): Boolean {
     //Test for access to the logging path and shutdown if we are unable to reach it.
     val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
     try {
+        if(!logPath.exists()){
+            printError("base directory $baseDirectory does not exist. Node will now shutdown")
```
anthonykeenan: Capital B please!

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -465,6 +466,10 @@ fun CliWrapperBase.initLogging(baseDirectory: Path): Boolean {
     //Test for access to the logging path and shutdown if we are unable to reach it.
     val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
     try {
+        if(!logPath.exists()){
```
anthonykeenan: A space is needed between before the curly brace

mnesbit: Should this not in fact be if(!baseDirectory.exists()) or similar? There is no explanation of the rationale here and it seems to contradict the intent of the logPath.createDirectories() that follows. Please add some commentary to explain why this is useful.

Bartman250: thanks - corrections done - just rebuilding/testing

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
florian-f: Jokes should go into their own individual PR. Otherwise we risk losing/ delaying good jokes.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
Bartman250: When I tested this in QA I got the following message
'Unable to create logging directory /some/random/dir/that/does/not/exist/logs. Node will now shutdown.'
Which is misleading, as you may thing it is a permission issue when in fact you have specified a base directory that does not exist. Hope that make sense.
I'll add suitable comments in the code if you are ok with this.

florian-f: The JIRA gives some insight into the intention behind this: https://r3-cev.atlassian.net/browse/CORDA-2834

Basically the error message we print when whe can't create the log directories is confusing to the user.
What we should do is:

- check if the base directory exists and handle the case when it doesn't (I'm pretty sure we already do that at an earlier point, and if we don't we should)

- check if the If the log directory (is it one directory or multiple directories?) exists, else print "Log Directories don't exist, attempting to create them",

- and then if that fails "base directory $baseDirectory does not exist.

node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -465,6 +466,12 @@ fun CliWrapperBase.initLogging(baseDirectory: Path): Boolean {
     //Test for access to the logging path and shutdown if we are unable to reach it.
     val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
     try {
+        //Ensure that the base directory actually exists before creating the log file
+        //so that we can warn the developer/operator. The subsequent create directory call will yield any issues related to permissions.
+        if(!logPath.exists()) {
```
mnesbit: This will never call logPath.createdirectories()

Bartman250:
Thanks for the feedback, talking to Florian this morning we need to find an earlier point to check if the base directory exists. Also thanks Matthew - you are correct - I meant to put base directory in that check and not log directory. I will add unit test for this as well, so that it fails local tests.

Bartman250:
Added base directory check ahead of the log files being create. Also raised Jira to improve testablity of NodeStartUpCli https://r3-cev.atlassian.net/browse/CORDA-2844 <<--- This will enable us to automate those QA command line tests

PR #4993, opened 2019-04-05
CORDA-2801: Test to check compatibility between TLS 1.2 and TLS 1.3

vkolomeyko:
The test is currently disabled till we move to Java 11 (or beyond) when TLS 1.3 becomes available as part of JDK.
Local testing been performed with Open JDK 12 (12+33) and the test is passing.


vkolomeyko:
@mnesbit - gentle reminder to review and approve this PR.

PR #4995, opened 2019-04-08
CORDA-2535 Move classes from test src to main

lankydan:
For some reason a load of classes were in the test src rather than the
main src even though they were needed during runtime. These have been
moved back into main so they are compiled for use in the demo app.

To resolve https://r3-cev.atlassian.net/browse/CORDA-2535


mnesbit:
These lived in test, because only integration tests should be messing with the node clock and it would be actively dangerous to expose that in say the finance package.

samples/irs-demo/cordapp/workflows-irs/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
 
     // Corda integration dependencies
     cordaCompile project(':core')
+    cordaCompile project(':node')
```
josecoll: CorDapps should never directly depend on `node` (only public APIs).

lankydan: `DemoClock` which is used in this demo app lives in `node`

lankydan:
Changes to the deal ref have been made to correctly display the id. You can now open the trade/deal from the home screen

samples/irs-demo/cordapp/workflows-irs/build.gradle
josecoll: hmmm ... so how was this demo working prior to this addition?

anthonykeenan: It wasn't working, it hasn't been working for a while, I suspect since the test API refactoring we did for the v3 release. It compiles and the tests run but the actual demo doesn't work.

josecoll: Either way, we can't have sample CorDapps depending on an internal module (eg. `node`).
Some creative refactoring required ;-)

lankydan: Not sure how creative we want to get. 

`DemoClock` extends `CordaClock` which also lives in `node`. Therefore moving `DemoClock` somewhere else doesn't solve the problem as a compile dependency to `node` needs to exist somewhere. We could rely on `node-driver` but that is really only for tests and not for runtime code.

Furthermore, the code in `node` is making use of `DemoClock` and even if it wasnt we would still need access to classes that exist within node.

So the question is how far do we want to take this. Do we create a new module and move all the clocks out of `node` or is that overkill to fix this sample cordapp.

Open to other suggestions if I have missed something.

anthonykeenan: I think as long as it is clear (possibly a change to the README, or comments in the code) that this is a _demo_ rather than a _sample_ I'm fine with using the internals in this very specific case.

lankydan: sure, that seems like a good compromise.

lankydan: these changes have been made 👍

anthonykeenan:
Looks good now

PR #4998, opened 2019-04-09
CORDA-2572 Alter the message of a `FlowException` on receiving side

lankydan:
When an exception is thrown the receiving party / counter flow will receive the `FlowException` with an altered message indicating it has come from the sending party / counter flow.

Locally the thrown exception will remain the same.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/flows/FlowException.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowException.kt
@@ -2,6 +2,7 @@ package net.corda.core.flows
 
 import net.corda.core.CordaException
 import net.corda.core.CordaRuntimeException
+import net.corda.core.identity.Party
 
 // DOCSTART 1
 /**
@@ -28,6 +29,8 @@ open class FlowException(message: String?, cause: Throwable?, var originalErrorI
     constructor(cause: Throwable?) : this(cause?.toString(), cause)
     constructor() : this(null, null)
 
+    var peer: Party? = null
+
     override fun getErrorId(): Long? = originalErrorId
 }
 // DOCEND 1
@@ -41,6 +44,8 @@ class UnexpectedFlowEndException(message: String, cause: Throwable?, val origina
         CordaRuntimeException(message, cause), IdentifiableException {
     constructor(message: String, cause: Throwable?) : this(message, cause, null)
     constructor(message: String) : this(message, null)
+
+    var peer: Party? = null
 
     override fun getErrorId(): Long? = originalErrorId
 }
--- a/core/src/main/kotlin/net/corda/core/flows/FlowException.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowException.kt
@@ -29,7 +29,8 @@ open class FlowException(message: String?, cause: Throwable?, var originalErrorI
     constructor(cause: Throwable?) : this(cause?.toString(), cause)
     constructor() : this(null, null)
 
-    var peer: Party? = null
+    // private field with obscure name to ensure it is not overridden
+    private var _peer: Party? = null
 
     override fun getErrorId(): Long? = originalErrorId
 }
@@ -45,7 +46,8 @@ class UnexpectedFlowEndException(message: String, cause: Throwable?, val origina
     constructor(message: String, cause: Throwable?) : this(message, cause, null)
     constructor(message: String) : this(message, null)
 
-    var peer: Party? = null
+    // private field with obscure name to ensure it is not overridden
+    private var _peer: Party? = null
 
     override fun getErrorId(): Long? = originalErrorId
 }
--- a/core/src/main/kotlin/net/corda/core/flows/FlowException.kt
+++ b/core/src/main/kotlin/net/corda/core/flows/FlowException.kt
@@ -30,7 +30,7 @@ open class FlowException(message: String?, cause: Throwable?, var originalErrorI
     constructor() : this(null, null)
 
     // private field with obscure name to ensure it is not overridden
-    private var _peer: Party? = null
+    private var peer: Party? = null
 
     override fun getErrorId(): Long? = originalErrorId
 }
@@ -47,7 +47,7 @@ class UnexpectedFlowEndException(message: String, cause: Throwable?, val origina
     constructor(message: String) : this(message, null)
 
     // private field with obscure name to ensure it is not overridden
-    private var _peer: Party? = null
+    private var peer: Party? = null
 
     override fun getErrorId(): Long? = originalErrorId
 }
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -173,9 +173,10 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
     private fun Throwable.fillInLocalStackTrace(): Throwable {
         fillInStackTrace()
         // provide useful information that can be displayed to the user
+        // reflection use to access private field
         when (this) {
             is UnexpectedFlowEndException -> {
-                peer?.let {
+                this.declaredField<Party?>("_peer").value?.let {
                     stackTrace = arrayOf(
                         StackTraceElement(
                             "Received unexpected counter-flow exception from peer",
@@ -187,7 +188,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
                 }
             }
             is FlowException -> {
-                peer?.let {
+                this.declaredField<Party?>("_peer").value?.let {
                     stackTrace = arrayOf(
                         StackTraceElement(
                             "Received counter-flow exception from peer",
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -176,7 +176,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
         // reflection use to access private field
         when (this) {
             is UnexpectedFlowEndException -> {
-                this.declaredField<Party?>("_peer").value?.let {
+                DeclaredField<Party?>(UnexpectedFlowEndException::class.java, "peer", this).value?.let {
                     stackTrace = arrayOf(
                         StackTraceElement(
                             "Received unexpected counter-flow exception from peer",
@@ -188,7 +188,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
                 }
             }
             is FlowException -> {
-                this.declaredField<Party?>("_peer").value?.let {
+                DeclaredField<Party?>(FlowException::class.java, "peer", this).value?.let {
                     stackTrace = arrayOf(
                         StackTraceElement(
                             "Received counter-flow exception from peer",
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -171,33 +171,36 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
     }
 
     private fun fillInStackTrace(throwable: Throwable): Throwable {
-        throwable.apply {
+        return throwable.apply {
             fillInStackTrace()
             // provide useful information that can be displayed to the user
             when (this) {
                 is UnexpectedFlowEndException -> {
                     if (peer != null) {
-                        stackTrace += StackTraceElement(
-                            "Received unexpected counter-flow exception from peer",
-                            " ${peer!!.name}",
-                            null,
-                            -1
-                        )
+                        stackTrace = arrayOf(
+                            StackTraceElement(
+                                "Received unexpected counter-flow exception from peer",
+                                " ${peer!!.name}",
+                                null,
+                                -1
+                            )
+                        ) + stackTrace
                     }
                 }
                 is FlowException -> {
                     if (peer != null) {
-                        throwable.stackTrace += StackTraceElement(
-                            "Received counter-flow exception from peer",
-                            " ${peer!!.name}",
-                            null,
-                            -1
-                        )
+                        stackTrace = arrayOf(
+                            StackTraceElement(
+                                "Received counter-flow exception from peer",
+                                " ${peer!!.name}",
+                                null,
+                                -1
+                            )
+                        ) + stackTrace
                     }
                 }
             }
         }
-        return throwable
     }
 
     /**
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -159,7 +159,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
                     is FlowContinuation.Resume -> {
                         return continuation.result
                     }
-                    is FlowContinuation.Throw -> throw fillInStackTrace(continuation.throwable)
+                    is FlowContinuation.Throw -> throw continuation.throwable.fillInLocalStackTrace()
                     FlowContinuation.ProcessEvents -> continue@eventLoop
                     FlowContinuation.Abort -> abortFiber()
                 }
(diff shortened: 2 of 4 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/transitions/DeliverSessionMessageTransition.kt
```diff
@@ -114,8 +114,10 @@ class DeliverSessionMessageTransition(
         val exception: Throwable = if (payload.flowException == null) {
             UnexpectedFlowEndException("Counter-flow errored", cause = null, originalErrorId = payload.errorId)
         } else {
-            payload.flowException.originalErrorId = payload.errorId
-            payload.flowException
+            payload.flowException.apply {
+                originalErrorId = payload.errorId
+                setMessage("Counter-flow threw FlowException: $message")
```
shamsasari: I think this is the wrong approach. Apps may well be assuming that the message in the FlowException when thrown is the same received, and it makes sense for it to be the same. Rather let's make it clear in the log lines which party threw the exception.

lankydan: Hopefully they aren't checking the message 😟 . I can change it to a log line instead although i think it loses some of its clarity since the exception is big and clear to read and the log line might be skipped over.

If we did go for the log line would we want it at INFO or WARN? I think INFO is probably right since throwing `FlowExceptions` is a norm.

rick-r3: I think we _should_ modify the exception, because it gets re-thrown and at that point we've lost the context.  So as it goes through the flow hospital etc it will not be possible to identify it came from a peer if we just log something here.

I get the changed message is a risk (as would changing the type - we should have wrapped these exceptions from the beginning in the `ExecutionException` style - but too late now).  Can we set/change the cause in some way?  And the stack trace if possible.

Couple of reasons to change the cause/stacktrace.

1. It's misleading for the cause from the other side to be present (and who knows what it reveals about the remote node).
2. We don't want causes with e.g. database deadlocks from the other side to be present as it might make our flow hospital think our database deadlocked! Not actually sure this is what could be happening but I think maybe it's possible if the whole cause comes over from the other node. (Deadlocks is probably a bad example as they retry forever on the other side.  But some stuff we catch a limited number of times before giving up and then that would potentially reach the initiating node).

The cause of this and the `UnexpectedFlowEndException` in the case where `flowException` is null could be a new `CounterFlowException` which does have the peer name in the message (maybe other things about the session).

It's tempting to change this under a target version flag in the future so that we swap the cause and the `FlowException` round because currently it would be hard for a flow to be able to figure out a peer was the cause (if they were attempting to catch the exception).  I bet we kept it like this for the notary, which throws an exception on double spend.

JamesHR3: To address the last paragraph: I'm about to send an email about using the `targetPlatformVersion` check, as it turns out trying to reason about when you should use it is very difficult. I think in this case a `tPV` gate doesn't buy you anything anyway.

If I've understood correctly, you're proposing to change the exception type thrown if the counterparty errors during a flow. That would imply that the outputs of the flow API have become stricter (the conditions under which an exception is thrown remain the same, but the exception types are better differentiated). That in turn implies that if an app depends on this behaviour, its `minimumPlatformVersion` number must be bumped. Putting the behaviour behind a `tPV` gate doesn't help regarding compatibility here - either the app doesn't care, in which case it works if it gets either behaviour, or it does, in which case it ceases to work on a V4 node. (Whether it cares or not I think depends on whether it has different behaviour if the app sees the new exception compared to if it sees the old one - if it does the same thing in either case then it will continue to work on both versions.)

rick-r3: Well, I said it was tempting.  But then I said I think we should basically ban behavioural change under version checks unless the fate of Corda depends on it.  i.e. we shouldn't make the change I said was tempting (in the last paragraph).

lankydan: I'm looking at making the sort of changes Rick mentioned around changing the `cause` to new class `CounterFlowException`. Although it does help clarify to the user that the error occurred on the counterparty but it goes back to the same sort of issue in altering the message. If anyone is depending on the `cause` to a specific exception then existing code will break. It also raises the question of what do we do with the original `cause` if it had one. It could be added to the new _cause_ but now the path to the possibly important exception info has been moved further away from the top level exception.

To me changing the `message` seems like a smaller change than altering the `cause`.

Let me know if I have misunderstood anything.

JamesHR3: Strictly speaking, changing the `cause` is a semantic change that makes the outputs stricter (it ensures the cause is a specific exception type with particular info in the cases where the counterparty has errored). Following the reasoning from my email, that means an app that depends on it won't work on an older version of the node. That means the change shouldn't be introduced until an app can perform a `minimumPlatformVersion` bump - i.e. not until Corda 5.

That being said, I think the chances of anyone actually depending on it are pretty much zero, so it's probably ok and we shouldn't stress about it too much. (Hope I'm not wrong on that...)

Regarding the message change, you're totally changing the set of possible messages that could be provided. That has some weird implications for versioning, but again I don't think we should worry about it - I don't think we should be setting our API in stone at the level of exception messages. So from my perspective I think I'm ok with either, but I'll let others chime in if they disagree.

lankydan: I have updated the code to reflect what has been discussed here. Let me know what you think.

lankydan: After delving further down the route of altering the `FlowException`'s cause, me and James think just logging the exception is the best route. 

This is due to the process of overriding the `cause` of the `FlowException` is not reliable enough.

For example, if someone has extended `FlowException` and written:
```kotlin
class MyFlowException(override val message: String, override val cause: Throwable) : FlowException(message, cause)
```
Overwriting the `cause` value using `setCause` (from `CordaException`) replaces the `_cause` (an internal cause value). But, the `cause` will remain the orignal cause since it is a `val` and lives on the subclass.

This problem would also arise if the `message` of the exception was changed.

If developers did not write exceptions in this way then it would work, but it is likely that a lot of developers will do it this way since they might not know how to pass the values into the base class in kotlin or want to access the values within their subclass.

rick-r3: Hmm.... we need to fix up _something_ about the exception otherwise the node is going to break because that is what is examined by the flow hospital, as well as confusing users (the original motivation for this PR).  And we can't have the flow hospital reacting to low level causes on the remote node and I have personally fallen for a node reporting an exception that was actually from the remote node.

This is a pain.  Must do something but seems like everything is incompatible in some way!  Haven't thought of a solution yet, but wondering what the exception serializer does?

rick-r3: Okay, here's my idea.  We can change the `stacktrace` to contain an extra element at the top to indicate it's from a remote node.  The elements are just `StackTraceElement` and mostly strings.  No one (unless they are insane) is going to be overriding those or writing code against them.  `Throwable` has a `setStackTrace` so I think we're good.

It just needs to be tested that a fake class/method/filename does not cause issues with logging (that seems to try and identify which JAR file has the file name).  Perhaps:

`class = "Received error from peer"`
`method = " peer = ${peer}"`
`filename = null`
`lineNumber = -1`

should give

`Received error from peer.  <peer> (Unknown Source)`

It is even perhaps correct to replace the entire stack trace with just this one element, rather than just putting it at the top, as this is less leaky and less confusing.

shamsasari: Yes

shamsasari: Though it is very hacky...

shamsasari: I was thinking of a `FlowException` sub-type, but that would be introducing new API, though we could make it internal for now.

rick-r3: Oh, yes, very hacky! :-) But we messed up a bit originally and now have to hack our way out of the corner.

I think introducing a new type is problematic as we've discussed above how developers might be testing the type/message/cause (or overriding them).  We could put a flag on `FlowException` that we set when deserializing (only) so we know something is from a remote process, but that only helps the flow hospital and not the humans.

lankydan: Ok, so another version has been pushed 😄 .

This time the info has been added to the stack trace. Leaving you with a stacktrace looking something like the below (peer info at the bottom of the stack):
```log
net.corda.core.flows.FlowException: BOOM
	at net.corda.node.services.statemachine.FlowStateMachineImpl.fillInStackTrace(FlowStateMachineImpl.kt:175) ~[corda-node-5.0-SNAPSHOT.jar:?]
	at net.corda.node.services.statemachine.FlowStateMachineImpl.processEventsUntilFlowIsResumed(FlowStateMachineImpl.kt:162) ~[corda-node-5.0-SNAPSHOT.jar:?]
	at net.corda.node.services.statemachine.FlowStateMachineImpl.suspend(FlowStateMachineImpl.kt:443) ~[corda-node-5.0-SNAPSHOT.jar:?]
	at net.corda.node.services.statemachine.FlowSessionImpl.sendAndReceive(FlowSessionImpl.kt:54) ~[corda-node-5.0-SNAPSHOT.jar:?]
	at net.corda.node.services.statemachine.FlowSessionImpl.sendAndReceive(FlowSessionImpl.kt:61) ~[corda-node-5.0-SNAPSHOT.jar:?]
	at net.corda.confidential.SwapIdentitiesFlow.call(SwapIdentitiesFlow.kt:128) ~[corda-confidential-identities-5.0-SNAPSHOT.jar:?]
	at net.corda.confidential.SwapIdentitiesFlow.call(SwapIdentitiesFlow.kt:34) ~[corda-confidential-identities-5.0-SNAPSHOT.jar:?]
	at net.corda.node.services.statemachine.FlowStateMachineImpl.subFlow(FlowStateMachineImpl.kt:326) ~[corda-node-5.0-SNAPSHOT.jar:?]
	at net.corda.core.flows.FlowLogic.subFlow(FlowLogic.kt:314) ~[corda-core-5.0-SNAPSHOT.jar:?]
	at com.lankydanblog.tutorial.base.flows.MyCashPaymentFlow.call(MyCashPaymentFlow.kt:58) ~[?:?]
	at com.lankydanblog.tutorial.base.flows.MyCashPaymentFlow.call(MyCashPaymentFlow.kt:34) ~[?:?]
.........
........
	at Received counter-flow exception from peer. O=PartyA, L=London, C=GB(Unknown Source) ~[?:?]
```

JamesHR3: Can we put the new element at the top of the stack trace? Feels like it's probably the most important bit of information, so should be as visible as possible.

lankydan: sure, could make my mind up which way to put it

lankydan: changes made

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -173,6 +170,39 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
         }
     }
 
+    private fun fillInStackTrace(throwable: Throwable): Throwable {
```
JamesHR3: Is this better defined as an extension method on `Throwable`? (Probably should be private, and I'm guessing it'll want renaming as there's already a `fillInStackTrace`.)

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -173,6 +170,39 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
...
+        return throwable.apply {
+            fillInStackTrace()
+            // provide useful information that can be displayed to the user
+            when (this) {
+                is UnexpectedFlowEndException -> {
+                    if (peer != null) {
+                        stackTrace = arrayOf(
+                            StackTraceElement(
+                                "Received unexpected counter-flow exception from peer",
+                                " ${peer!!.name}",
```
JamesHR3: I thought Kotlin was aware that peer could not be null because you've checked it in the if block, so you shouldn't need the `!!` here. Is that the case?

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -173,6 +170,39 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
...
+                            )
+                        ) + stackTrace
+                    }
+                }
+                is FlowException -> {
+                    if (peer != null) {
+                        stackTrace = arrayOf(
+                            StackTraceElement(
+                                "Received counter-flow exception from peer",
+                                " ${peer!!.name}",
```
JamesHR3: Same again here.

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
lankydan: the only reason i didnt make it an extension function is that i couldnt come up with a suitable name that doesnt conflict with `fillInStackTrace` ☹️

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
lankydan: nope, intellij complained that it couldn't be changed... 
Screenshot 2019-04-12 at 17 19 07

lankydan: changed to use `?.let` to get around this

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
lankydan: changed to extension function named `fillInLocalStackTrace`.

core/src/main/kotlin/net/corda/core/flows/FlowException.kt
```diff
@@ -28,6 +29,8 @@ open class FlowException(message: String?, cause: Throwable?, var originalErrorI
     constructor(cause: Throwable?) : this(cause?.toString(), cause)
     constructor() : this(null, null)
 
+    var peer: Party? = null
```
rick-r3: I've had an annoying realisation.    What if someone's subclass of this has a `val` or `var` called `peer`?  Their stuff might not compile anymore.  Do we need to make the name more obscure?  I think we should.  Are we allowed to call is `__peer` (double underscore)?

core/src/main/kotlin/net/corda/core/flows/FlowException.kt
```diff
@@ -42,5 +45,7 @@ class UnexpectedFlowEndException(message: String, cause: Throwable?, val origina
     constructor(message: String, cause: Throwable?) : this(message, cause, null)
     constructor(message: String) : this(message, null)
 
+    var peer: Party? = null
```
rick-r3: Same here.   See comment above.

core/src/main/kotlin/net/corda/core/flows/FlowException.kt
rick-r3: Put a comment on the code as to why we gave it such a weird name.

lankydan: ugh it just gets worse and worse, sure I can do that. I dont think we have used double underscore anywhere in the code, we have used single underscores for "private" backing vars so I'll probably go for a `_peer` (1 underscore) unless you really think we should make it super safe and go for 2.

lankydan: this is ready for re-review, made `_peer` private and accessed via reflection.

core/src/main/kotlin/net/corda/core/flows/FlowException.kt
```diff
@@ -28,6 +29,9 @@ open class FlowException(message: String?, cause: Throwable?, var originalErrorI
     constructor(cause: Throwable?) : this(cause?.toString(), cause)
     constructor() : this(null, null)
 
+    // private field with obscure name to ensure it is not overridden
+    private var _peer: Party? = null
```
shamsasari: Remove the underscore. It's not needed now that it's private.

lankydan: I dont think this is right. If someone had extended `FlowException` and had their own `peer` field with a different type our code will no longer work. If their field happens to be the same type (`Party?`) then it works ok. The `peer` data is lost while passing it across anyway and our code just overwrites their public `peer` field instead of our private one. But if the type is different at all it will all start blowing up.

Therefore sticking with the more obscure field name is probably better.

shamsasari: A subclass can have its own `peer` field, and it won't conflict with this in anyway. However, because we're using `declaredField`, and the way that it's written (it uses the runtime Class of the object to look for the field) means we can't use it. What we want is to look for "peer" in FlowException (`FlowException::class.java.getDeclaredField("peer")`). 

Also the comment above it is not quite correct. The obscure name doesn't prevent it from being overridden, it's the fact that it's private.

If anything, I might even prefer to have `peer` public and document it accordingly as part of the API (assuming this fix doesn't need to be backported). I'm not sure I share the previous concern of adding public members to open classes. Surely the JDK has done that numerous times? But anyway, this PR has dragged on long enough.

node/src/test/kotlin/net/corda/node/services/statemachine/FlowFrameworkTests.kt
```diff
@@ -209,6 +212,29 @@ class FlowFrameworkTests {
         assertThat((lastMessage.payload as ErrorSessionMessage).flowException!!.stackTrace).isEmpty()
     }
 
+    @Test
+    fun `FlowException can have a peer field without causing serialisation problems`() {
```
shamsasari: sub-class of FlowException ...

PR #4999, opened 2019-04-09
CORDA-2651 Check if resources are in classpath

rui-r3:
[CORDA-2651](https://r3-cev.atlassian.net/browse/CORDA-2651) 

Check if resources are in classpath before passing them to Liquibase. Improve exception logging.

### DIFF
--- a/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
+++ b/node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
@@ -13,7 +13,6 @@ import net.corda.nodeapi.internal.MigrationHelpers.getMigrationResource
 import net.corda.core.schemas.MappedSchema
 import net.corda.core.utilities.contextLogger
 import net.corda.nodeapi.internal.cordapp.CordappLoader
-import sun.security.x509.X500Name
 import java.io.ByteArrayInputStream
 import java.io.InputStream
 import java.nio.file.Path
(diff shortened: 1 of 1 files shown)


node-api/src/main/kotlin/net/corda/nodeapi/internal/persistence/SchemaMigration.kt
```diff
@@ -217,6 +218,14 @@ class SchemaMigration(
         }
         return isExistingDBWithoutLiquibase || isFinanceAppWithLiquibaseNotMigrated
     }
+
+    private fun checkResourcesInClassPath(resources: List<String?>) {
+        for (resource in resources) {
+            if (classLoader.getResource(resource) == null) {
+                throw CouldNotCreateDataSourceException("Could not find Liquibase resource $resource. Please ensure the jar file containing it is deployed in the cordapps directory.")
```
szymonsztuka: ``throw CouldNotCreateDataSourceException`` change to ``throw DatabaseMigrationException``
The change of the error class does't make much difference now, but there's other PR which will benefit of it. 

``Could not find Liquibase resource $resource.`` -> ``Could not find Liquibase database migration script $resource.``
The better naming consistent with our docs, not sure if ``Liquibase`` is needed but the more the better.

rui-r3: Fixed in 7b6edb749f3bc67633638d6c7b5cf368c7494055

shamsasari:
@rui-r3 be sure to mark the JIRA ticket as done.

PR #5000, opened 2019-04-10
[CORDA-2522] Improve error reporting around failed flows

JamesHR3:
Error reporting around flow failures can be difficult to parse, owing to the error being buried within a lot of other information. This PR makes a few tweaks that aim to improve the clarity of error messages, reduce the amount of duplicate information, and move some of the more verbose logging to lower logging levels.

Changes:
 - In `StaffedFlowHospital`, reword the log that prints each error for a flow to make it clear that (possibly) multiple errors are being printed. Additionally, remove extra information that is printed in the logging context, and promote the log to warn. This is where the exception traceback is printed.
 - In `FlowStateMachineImpl`, remove a duplicate printing of the exception traceback
 - In `AttachmentsClassloader`, add a log at the point where an unverified attachment exception is thrown.
 - In `DumpHistoryOnErrorInterceptor`, move the logs to debug. These logs record all state machine transitions when a flow fails, which is often unnecessary in practice.

Testing
 - Used the upgrade cordapp to send a transaction with the new contract version to a node without it, and looked through the logs to observe the changes.

### DIFF
--- a/core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
+++ b/core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
@@ -127,7 +127,7 @@ class AttachmentsClassLoader(attachments: List<Attachment>,
 
         if (untrusted.isNotEmpty()) {
             log.warn("Cannot verify transaction $sampleTxId as the following attachment IDs are untrusted: $untrusted." +
-                    "You will need to install that app version yourself, to whitelist it for use. " +
+                    "You will need to manually install the CorDapp to whitelist it for use. " +
                     "Please follow the operational steps outlined in https://docs.corda.net/cordapp-build-systems.html#cordapp-contract-attachments to learn more and continue.")
             throw TransactionVerificationException.UntrustedAttachmentsException(sampleTxId, untrusted)
         }
--- a/core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
+++ b/core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
@@ -125,8 +125,12 @@ class AttachmentsClassLoader(attachments: List<Attachment>,
                 }
                 .map(Attachment::id)
 
-        if (untrusted.isNotEmpty())
+        if (untrusted.isNotEmpty()) {
+            log.warn("Cannot verify transaction $sampleTxId as the following attachment IDs are untrusted: $untrusted." +
+                    "You will need to install that app version yourself, to whitelist it for use. " +
+                    "Please follow the operational steps outlined in https://docs.corda.net/cordapp-build-systems.html#cordapp-contract-attachments to learn more and continue.")
             throw TransactionVerificationException.UntrustedAttachmentsException(sampleTxId, untrusted)
+        }
 
         // Enforce the no-overlap and package ownership rules.
         checkAttachments(attachments)
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -243,7 +243,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
             if(t.isUnrecoverable()) {
                 errorAndTerminate("Caught unrecoverable error from flow. Forcibly terminating the JVM, this might leave resources open, and most likely will.", t)
             }
-            logger.info("Flow raised an error... sending it to flow hospital", t)
+            logger.info("Flow raised an error: ${t.message}. Sending it to flow hospital to record and diagnose.")
             Try.Failure<R>(t)
         }
         val softLocksId = if (hasSoftLockedStates) logic.runId.uuid else null
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
@@ -243,7 +243,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
             if(t.isUnrecoverable()) {
                 errorAndTerminate("Caught unrecoverable error from flow. Forcibly terminating the JVM, this might leave resources open, and most likely will.", t)
             }
-            logger.info("Flow raised an error: ${t.message}. Sending it to flow hospital to record and diagnose.")
+            logger.info("Flow raised an error: ${t.message}. Sending it to flow hospital to be triaged.")
             Try.Failure<R>(t)
         }
         val softLocksId = if (hasSoftLockedStates) logic.runId.uuid else null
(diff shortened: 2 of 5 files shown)


core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
```diff
@@ -125,8 +125,12 @@ class AttachmentsClassLoader(attachments: List<Attachment>,
                 }
                 .map(Attachment::id)
 
-        if (untrusted.isNotEmpty())
+        if (untrusted.isNotEmpty()) {
+            log.warn("Cannot verify transaction $sampleTxId as the following attachment IDs are untrusted: $untrusted." +
+                    "You will need to install that app version yourself, to whitelist it for use. " +
```
anthonykeenan: Is "You will need to manually install the CorDapp to whitelist it for use." slightly clearer?

node/src/main/kotlin/net/corda/node/services/statemachine/FlowStateMachineImpl.kt
```diff
@@ -243,7 +243,7 @@ class FlowStateMachineImpl<R>(override val id: StateMachineRunId,
             if(t.isUnrecoverable()) {
                 errorAndTerminate("Caught unrecoverable error from flow. Forcibly terminating the JVM, this might leave resources open, and most likely will.", t)
             }
-            logger.info("Flow raised an error... sending it to flow hospital", t)
+            logger.info("Flow raised an error: ${t.message}. Sending it to flow hospital to record and diagnose.")
```
anthonykeenan: Sending it to flow hospital for triage? I think that's the terminology used by the hospital.

anthonykeenan: to be triaged?

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
```diff
@@ -160,7 +160,8 @@ class StaffedFlowHospital(private val flowMessaging: FlowMessaging, private val
         return errors
                 .asSequence()
                 .mapIndexed { index, error ->
-                    log.info("Flow ${flowFiber.id} has error [$index]", error)
+                    // Rely on the logging context to print details of the flow ID.
+                    log.warn("Error ${index + 1} of ${errors.size}:", error)
```
anthonykeenan: We're removing the `flowFiber.id` here - are we removing any useful information?

JamesHR3: This one is deliberate - the logging context prints out `flowFiber.id` (along with a bunch of other stuff), so I removed it.

core/src/main/kotlin/net/corda/core/serialization/internal/AttachmentsClassLoader.kt
JamesHR3: That does sound better. I actually borrowed this from the `UntrustedAttachmentException`, so I'll change it there too.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
rick-r3: I'm not sure this is the right approach.  The flow hospital here should just be INFO I think, because if it is discharged, it has been handled automatically.  The actual error should be logged when it propagates, or when it is kept in for observation.  In the former, I think it will already log at an elevated level.  I think what we want to do here is when there is `Diagnosis.OVERNIGHT_OBSERVATION` generate a warning (In the scenario in the ticket, it's kept in for observation isn't it, and currently I think there would be nothing above INFO level when that happens).

The existing message in this scenario does so at INFO level and does not include the cause (line 114).

node/src/main/kotlin/net/corda/node/services/statemachine/interceptors/DumpHistoryOnErrorInterceptor.kt
```diff
@@ -39,11 +39,12 @@ class DumpHistoryOnErrorInterceptor(val delegate: TransitionExecutor) : Transiti
             (record ?: ArrayList()).apply { add(transitionRecord) }
         }
 
-        // Just if we decide to propagate, and not if just on the way to the hospital.
+        // Just if we decide to propagate, and not if just on the way to the hospital. Only log at debug level here - the flow transition
+        // information is often unhelpful in the logs, and the actual cause of the problem will be logged elsewhere.
         if (nextState.checkpoint.errorState is ErrorState.Errored && nextState.checkpoint.errorState.propagating) {
-            log.warn("Flow ${fiber.id} errored, dumping all transitions:\n${record!!.joinToString("\n")}")
+            log.debug("Flow ${fiber.id} errored, dumping all transitions:\n${record!!.joinToString("\n")}")
```
shamsasari: This should be `debug { ... }`

rick-r3: This was intentionally left at a higher log level because this is only installed for dev mode to help developers.  Is debug on by default in dev mode ?

anthonykeenan: It definitely isn't on by default in dev mode.

node/src/main/kotlin/net/corda/node/services/statemachine/StaffedFlowHospital.kt
JamesHR3: Good point - I think you're right that this probably isn't the right place to log the error, given the logs in `flowErrored`. How about the following:
 - Change the log for `Diagnosis.NOT_MY_SPECIALITY` and `Diagnosis.OVERNIGHT_OBSERVATION` to warn and print the error here. (Stop printing `flowFiber.id` because it's in the logging context)
 - Change the log for the other Diagnosis to print the error that was handled (perhaps just the message?)
 - Change the log in `consultStaff` back to INFO (or possibly even DEBUG - we're going to be printing the stack trace for the most important error we couldn't deal with anyway, and the others are essentially ignored. Leaving at INFO would result in an error being printed twice with the other changes suggested here)

node/src/main/kotlin/net/corda/node/services/statemachine/interceptors/DumpHistoryOnErrorInterceptor.kt
JamesHR3: No, it isn't on by default. My argument was that most of the time, you don't want to see this, even when developing. Problems I've hit have generally been things going wrong somewhere in the flow, and all I've really needed to see has been the exception stack trace. The flow framework history is usually not helpful in that case. It tends to be quite verbose too, so I end up having to track back through a lot of logs to find the thing I care about.

I can see the argument the other way though - presumably there will come a time when this is helpful and not having it on by default will be annoying, as you'll have to reproduce with debug on. My thought was that that condition was fairly rare and so the mainline case of not needing it should win, but I'm happy to change back if we think I'm underestimating its utility.

PR #5006, opened 2019-04-11
[CORDA-2522] Follow up changes to error reporting around failed flows

JamesHR3:
The previous PR (#5000) ended up getting reviewed at the same time that it got merged, resulting in a few pieces of tidying up that didn't find their way into master. This PR addresses those issues:
 - Change a warning back to info, as it has been pointed out that the error condition has been logged elsewhere.
 - `FinalityDoctor` now reports the error message stack trace, as this is the point with the most context about the problem
 - Use the Corda provided `debug` utility where appropriate
 - Remove some other `flowFiber.id`s that should be in the logging context

### DIFF
--- a/node/src/main/kotlin/net/corda/node/services/statemachine/interceptors/DumpHistoryOnErrorInterceptor.kt
+++ b/node/src/main/kotlin/net/corda/node/services/statemachine/interceptors/DumpHistoryOnErrorInterceptor.kt
@@ -43,9 +43,9 @@ class DumpHistoryOnErrorInterceptor(val delegate: TransitionExecutor) : Transiti
         // Just if we decide to propagate, and not if just on the way to the hospital. Only log at debug level here - the flow transition
         // information is often unhelpful in the logs, and the actual cause of the problem will be logged elsewhere.
         if (nextState.checkpoint.errorState is ErrorState.Errored && nextState.checkpoint.errorState.propagating) {
-            log.debug { "Flow ${fiber.id} errored, dumping all transitions:\n${record!!.joinToString("\n")}" }
+            log.warn("Flow ${fiber.id} errored, dumping all transitions:\n${record!!.joinToString("\n")}")
             for (error in nextState.checkpoint.errorState.errors) {
-                log.debug("Flow ${fiber.id} error", error.exception)
+                log.warn("Flow ${fiber.id} error", error.exception)
             }
         }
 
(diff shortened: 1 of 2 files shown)


node/src/main/kotlin/net/corda/node/services/statemachine/interceptors/DumpHistoryOnErrorInterceptor.kt
```diff
@@ -42,7 +43,7 @@ class DumpHistoryOnErrorInterceptor(val delegate: TransitionExecutor) : Transiti
         // Just if we decide to propagate, and not if just on the way to the hospital. Only log at debug level here - the flow transition
         // information is often unhelpful in the logs, and the actual cause of the problem will be logged elsewhere.
         if (nextState.checkpoint.errorState is ErrorState.Errored && nextState.checkpoint.errorState.propagating) {
-            log.debug("Flow ${fiber.id} errored, dumping all transitions:\n${record!!.joinToString("\n")}")
+            log.debug { "Flow ${fiber.id} errored, dumping all transitions:\n${record!!.joinToString("\n")}" }
```
rick-r3: Rest of PR looks okay.  WRT this, is debug on in dev mode by default?  The idea was this was for Cordapp developers, and they currently get all this diagnostic stuff when developing their apps in prior versions.

JamesHR3: It wouldn't be on. I was actually trying to get rid of it in the mainline case as I thought the diagnostic information was largely unhelpful. For a Cordapp developer, is there anything they can do with the flow state machine transition information that would help them anyway? I would have thought if something has gone wrong there that implies a platform bug, but I'm not familiar enough with the code to say for certain.

Ultimately I don't mind too much - if we think having this information regardless is valuable, then I'll reinstate this as an info/warn log. (Think it was warn before.)

JamesHR3: Err, to clarify, when I say largely unhelpful, I mean that most of the issues I've encountered were due to an exception being thrown somewhere in the flow that was then correctly handled by the framework. Typically I've found everything I've needed in the stack trace in these cases, and the framework transitions just add an extra couple of pages of log output to sift through to find it. But again, I'm not totally familiar with this, and it's more than possible there's a class of problems for which this is very helpful that I simply haven't encountered.

rick-r3: I'm not entirely certain, but the idea was that if a developer mis-aligns their sends and receives or has other bugs then perhaps this stuff is useful to understand exactly what a flow did leading up to the error.  I'm not arguing strongly either way because I'm not really a "user" of this.

JamesHR3: I'm going to revert this, as I don't think we have a clear picture of how useful it is, and possibly someone does find it helpful. We can lower it to debug if anyone complains!

PR #5019, opened 2019-04-12
CORDA-2602 Update getting setup guide java details

lankydan:
Raised to address https://r3-cev.atlassian.net/browse/CORDA-2602 . 

Remove some information that is way to specific for installing the
oracle jdk.

Specify that the docs will go through how to install the oracle jdk
only but provide links to amazon corretto and openjdk.

Also say that we support intellij 2019 which has been released now.


lankydan:
zulu added and fixed typo 👍

PR #5024, opened 2019-04-15
Fix tut-two-party-flow kotlin docs + make both versions easier to read

lankydan:
- `DOCSTART`s were not in sync
- Incorrect indentation
- Documentation not very clear


shamsasari:
We should probably back port this to the 4.1 branch.

lankydan:
ok, ill check 4.0 as well

PR #5028, opened 2019-04-16
CORDA-2860 relax property type checking

r3domfox:
This PR corrects the property type checking code in `PropertyDescriptor`, and relaxes the rule to avoid issues around type parameters and wildcards, for example that the getter and setter generated by Kotlin for `foo` in

```kotlin
class ContainsSet(var foo: Set<Foo>)
```

have the signatures

```java
public Set<Foo> getFoo()
public void setFoo(foo: Set<? extends Foo>)
```

which fails the requirement that the type received by the setter must be a subtype of the type returned by the getter.

This PR also stops the type inspector from doing property inspection on classes with custom serializers, since we should not be throwing exceptions if such classes happen to have invalid field/getter/setter type combinations (see `LocalDate.month`, for example).

### DIFF
--- a/serialization/src/test/kotlin/net/corda/serialization/internal/SetsSerializationTest.kt
+++ b/serialization/src/test/kotlin/net/corda/serialization/internal/SetsSerializationTest.kt
@@ -6,6 +6,7 @@ import net.corda.core.serialization.deserialize
 import net.corda.core.serialization.serialize
 import net.corda.node.serialization.kryo.kryoMagic
 import net.corda.node.services.statemachine.DataSessionMessage
+import net.corda.serialization.internal.amqp.propertyDescriptors
 import net.corda.testing.core.SerializationEnvironmentRule
 import net.corda.testing.internal.kryoSpecific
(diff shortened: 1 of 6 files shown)


serialization/src/test/kotlin/net/corda/serialization/internal/SetsSerializationTest.kt
```diff
@@ -64,4 +65,14 @@ class SetsSerializationTest {
         }
         assertArrayEquals(output.toByteArray(), serializedForm.bytes)
     }
+
+    open class P
+    class VarOfP(var p: Set<P>)
+
+    // See CORDA-2860
```
fenryka: can we have a brief comment why this tests the problem? I assume other devs are lazy and wont' go to the Corda Jira

r3domfox: Done.

r3domfox:
Will update changelog on 4.1 backport

PR #5034, opened 2019-04-17
CORDA-2862: Fix to allow softlinks of logs directory

manojkrgupta:
- If logs directory in corda base directory is a soft link to some other existing director, it does not work in Corda 4.
- This corda is a fix to allow such soft links to work.
- Have test the code, it works.
- Tried]$ ./gradlew   samples:trader-demo:deployNodes
- Got a working node -- Test passed.

# PR Checklist:

- [ ] Have you run the unit, integration and smoke tests as described [here](https://docs.corda.net/head/testing.html)?
- [ ] If you added public APIs, did you write the JavaDocs?
- [ ] If the changes are of interest to application developers, have you added them to the [changelog](https://github.com/corda/corda/blob/master/docs/source/changelog.rst), and potentially the [release notes](https://github.com/corda/corda/blob/master/docs/source/release-notes.rst)?
- [ ] If you are contributing for the first time, please read the agreement in [CONTRIBUTING.md](https://github.com/corda/corda/blob/master/CONTRIBUTING.md) now and add a comment to this pull request stating that your PR is in accordance with the [Developer's Certificate of Origin](https://github.com/corda/corda/blob/master/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
+++ b/node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
@@ -38,6 +38,7 @@ import java.io.RandomAccessFile
 import java.lang.management.ManagementFactory
 import java.net.InetAddress
 import java.nio.channels.UnresolvedAddressException
+import java.nio.file.Files
 import java.nio.file.Path
 import java.time.DayOfWeek
 import java.time.ZonedDateTime
@@ -486,7 +487,11 @@ fun CliWrapperBase.initLogging(baseDirectory: Path): Boolean {
     //Test for access to the logging path and shutdown if we are unable to reach it.
     val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
     try {
-        (logPath / "/").createDirectories()
+        if (Files.isSymbolicLink(logPath)){
+            Files.readSymbolicLink(logPath).createDirectories()
+        } else {
+            logPath.createDirectories()
+        }
     } catch (e: IOException) {
         printError("Unable to create logging directory ${logPath.toString()}. Node will now shutdown.")
         return false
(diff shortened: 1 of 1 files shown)


node/src/main/kotlin/net/corda/node/internal/NodeStartup.kt
```diff
@@ -486,7 +486,7 @@ fun CliWrapperBase.initLogging(baseDirectory: Path): Boolean {
     //Test for access to the logging path and shutdown if we are unable to reach it.
     val logPath = baseDirectory / NodeCliCommand.LOGS_DIRECTORY_NAME
     try {
-        logPath.createDirectories()
+        (logPath / "/").createDirectories()
```
shamsasari: It's not clear how this resolves the issue. Instead use `Files.readSymbolicLink` and `Files.readSymbolicLink`.

manojkrgupta: Most of the conditions are internally handled at os layer. 

When we have a situation of L --> pointedBySoftLinkdirectory
* Reference to L is L itself(softlink itself)
* Reference to L/ is the destination directory (pointedBySoftLinkdirectory) which L (softlink) is pointing to

So the code initially had a call of create directly L 
And if L is already existing as a soft link, operating system won't be able to create a directory named L since that name=L is already taken by a soft link.

When we change the create directory call to L/ , operating system now tries to create the directory pointedBySoftLinkdirectory/ which is  destination directory in case of softlink and in case of non soft link it will be able to create a directory L/ locally.

Hence both the situation/need is covered.

shamsasari: I tested with this piece of code:
```
        if (Files.isSymbolicLink(logPath)) {
            Files.readSymbolicLink(logPath).createDirectories()
        } else {
            logPath.createDirectories()
        }
```
It's clearer what's going on and relies on Java to deal with any differences at FS layer.

manojkrgupta: This is exactly same.
I have updated the code.
Have tested looks good.
Have pushed the code for your review.

PR #5047, opened 2019-04-20
CashUtils.generateSpend: add anonymous flag, default to true

cxyzhang0:
Mike,
  Can we add a flag to the generateSpend functions in CashUtils to indicate whether or not to use CI for the party who is going to receive the change? Default to true so this change won't impact existing usages.
  We have disabled the CA function of our node certs so the current generateSpend which uses CI exclusively would fail for our use case. Providing that flag will give the developer the tool to override that behavior.


cxyzhang0:
That way, we don't have to fork away.

cxyzhang0:
Close this one because it touches too many unnecessary files. Use  #5048 instead.

PR #5048, opened 2019-04-20
CashUtils.generateSpend: add anonymous flag, default to true

cxyzhang0:
This one replaces #5047 which touches too many unnecessary files. We only need to change one file CashUtils.kt. 
Mike,
Can we add a flag to the generateSpend functions in CashUtils to indicate whether or not to use CI for the party who is going to receive the change? Default to true so this change won't impact existing usages.
We have disabled the CA function of our node certs so the current generateSpend which uses CI exclusively would fail for our use case. Providing that flag will give the developer the tool to override that behavior.
That way, we don't have to fork away the Corda codebase.

### DIFF
--- a/finance/workflows/src/main/kotlin/net/corda/finance/workflows/asset/CashUtils.kt
+++ b/finance/workflows/src/main/kotlin/net/corda/finance/workflows/asset/CashUtils.kt
@@ -29,7 +29,6 @@ object CashUtils {
      * @param onlyFromParties if non-null, the asset states will be filtered to only include those issued by the set
      *                        of given parties. This can be useful if the party you're trying to pay has expectations
      *                        about which type of asset claims they are willing to accept.
-     * @param anonymous whether or not to use CI to send the change to
      * @return A [Pair] of the same transaction builder passed in as [tx], and the list of keys that need to sign
      *         the resulting transaction for it to be valid.
      * @throws InsufficientBalanceException when a cash spending transaction fails because
@@ -43,8 +42,7 @@ object CashUtils {
                       tx: TransactionBuilder,
                       amount: Amount<Currency>,
                       to: AbstractParty,
-                      onlyFromParties: Set<AbstractParty> = emptySet(),
-                      anonymous: Boolean = true): Pair<TransactionBuilder, List<PublicKey>> {
+                      onlyFromParties: Set<AbstractParty> = emptySet()): Pair<TransactionBuilder, List<PublicKey>> {
         return generateSpend(services, tx, listOf(PartyAndAmount(to, amount)), services.myInfo.legalIdentitiesAndCerts.single(), onlyFromParties, anonymous)
     }
 
@@ -58,7 +56,9 @@ object CashUtils {
      *           to move the cash will be added on top.
      * @param amount How much currency to send.
      * @param to the recipient party.
-     * @param ourIdentity well known identity to create a new confidential identity from if anonymous is true, for sending change to.
+     * @param ourIdentity ourIdentity is used to determine the where the change will be sent.
+     *                    If anonymous is true then an anonymous identity will be generated from this and the change
+     *                    will be spent to that, otherwise ourIdentity will be used as is.
      * @param onlyFromParties if non-null, the asset states will be filtered to only include those issued by the set
      *                        of given parties. This can be useful if the party you're trying to pay has expectations
      *                        about which type of asset claims they are willing to accept.
@@ -94,7 +94,6 @@ object CashUtils {
      * @param onlyFromParties if non-null, the asset states will be filtered to only include those issued by the set
      *                        of given parties. This can be useful if the party you're trying to pay has expectations
      *                        about which type of asset claims they are willing to accept.
-     * @param anonymous whether or not to use CI to send the change to
      * @return A [Pair] of the same transaction builder passed in as [tx], and the list of keys that need to sign
      *         the resulting transaction for it to be valid.
      * @throws InsufficientBalanceException when a cash spending transaction fails because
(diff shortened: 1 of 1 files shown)


finance/workflows/src/main/kotlin/net/corda/finance/workflows/asset/CashUtils.kt
```diff
@@ -42,8 +43,9 @@ object CashUtils {
                       tx: TransactionBuilder,
                       amount: Amount<Currency>,
                       to: AbstractParty,
-                      onlyFromParties: Set<AbstractParty> = emptySet()): Pair<TransactionBuilder, List<PublicKey>> {
-        return generateSpend(services, tx, listOf(PartyAndAmount(to, amount)), services.myInfo.legalIdentitiesAndCerts.single(), onlyFromParties)
+                      onlyFromParties: Set<AbstractParty> = emptySet(),
+                      anonymous: Boolean = true): Pair<TransactionBuilder, List<PublicKey>> {
```
shamsasari: Deprecated API should not be updated. Let's leave this method as is.

finance/workflows/src/main/kotlin/net/corda/finance/workflows/asset/CashUtils.kt
```diff
@@ -102,8 +107,9 @@ object CashUtils {
     fun generateSpend(services: ServiceHub,
                       tx: TransactionBuilder,
                       payments: List<PartyAndAmount<Currency>>,
-                      onlyFromParties: Set<AbstractParty> = emptySet()): Pair<TransactionBuilder, List<PublicKey>> {
-        return generateSpend(services, tx, payments, services.myInfo.legalIdentitiesAndCerts.single(), onlyFromParties)
+                      onlyFromParties: Set<AbstractParty> = emptySet(),
+                      anonymous: Boolean = true): Pair<TransactionBuilder, List<PublicKey>> {
```
shamsasari: likewise

finance/workflows/src/main/kotlin/net/corda/finance/workflows/asset/CashUtils.kt
```diff
@@ -56,10 +58,11 @@ object CashUtils {
      *           to move the cash will be added on top.
      * @param amount How much currency to send.
      * @param to the recipient party.
-     * @param ourIdentity well known identity to create a new confidential identity from, for sending change to.
+     * @param ourIdentity well known identity to create a new confidential identity from if anonymous is true, for sending change to.
```
shamsasari: The description is not correct. Perhaps something like "`ourIdentity` is used to determine the where the change will be sent. If `anonymous` is true then an anonymous identity will be generated from this and the change will be spent to that, otherwise `ourIdentity` will be used as is."

finance/workflows/src/main/kotlin/net/corda/finance/workflows/asset/CashUtils.kt
```diff
@@ -43,7 +43,7 @@ object CashUtils {
                       amount: Amount<Currency>,
                       to: AbstractParty,
                       onlyFromParties: Set<AbstractParty> = emptySet()): Pair<TransactionBuilder, List<PublicKey>> {
-        return generateSpend(services, tx, listOf(PartyAndAmount(to, amount)), services.myInfo.legalIdentitiesAndCerts.single(), onlyFromParties)
+        return generateSpend(services, tx, listOf(PartyAndAmount(to, amount)), services.myInfo.legalIdentitiesAndCerts.single(), onlyFromParties, anonymous)
```
shamsasari: It should be `true`, not `anonymous`

cxyzhang0: Good catch. I actually had a compiler error I overlooked.
I removed anonymous to make it identical to the pre-PR state because the API is deprecated.
The default in the called function will take care of it.

cxyzhang0:
@shamsasari Thank you for your help. Is it also going into 4.1 or 4.2?
Do I need a new PR for 4.1 or the current PR will take care of it?

shamsasari:
Unfortunately we don't include new API in point releases. So your changes would land in Corda 5.

gendal:
Hi @cxyzhang0 - just to explain the versioning thing as it confuses some people the first time they hear about it.  App developers and apps themselves need a way to know which APIs are provided by the node - an app that uses the API added by this PR wouldn't work if run on C4.0, for example. The way we do this is through the major version...   our rule is: "add/change an API, then bump the major version".  That way, app developers know that the v3 API is consistent for all 3.x releases and the v4 API is consistent for all 4.x releases and so on.  This means that, sometimes, a major release is actually very small(!) - eg Corda 2.0 just added a single API and nothing else.  It is looking likely (although not yet confirmed) that Corda 5.0 will fall into this latter category.. actually fairly minor in terms of functionality - and shipping in the coming months.  No promises but watch this space.

cxyzhang0:
@gendal Hi Richard, thank you for the info. I appreciate Corda's versioning policy and have no intention to violate it. Because of the way the PR uses default to preserve the API's original behavior, my tests have shown C4.0 apps compile and run without any changes against the PR and their behavior is the same with and without the PR. I think that is what default gives us in theory. But I could be wrong and will be interested in learning any test case that shows otherwise.
Mike, James and Guy are familiar with our production dependency. The short story is that we have removed the CA function from all the nodes' certs due to our PKI policy. So any attempt to generate CI would throw an exception. Hence the PR to provide us a way to use the finance cash module in the main repository without having to fork away. BTW, even a simple fork has not worked for us due to the package sealing restriction introduced in C4.0. So we are stuck. We could use the token sdk instead, but that is a major change for us - our CorDapp has already gone thru two POCs and tests and is slated for Pilot in the coming months. 5.0 may be too late for us.

anthonykeenan:
FYI - this caused a Java compatibility break - see https://r3-cev.atlassian.net/browse/CORDA-2899 for the issue and https://github.com/corda/corda/pull/5048 for the fix

shamsasari:
>Because of the way the PR uses default to preserve the API's original behavior, my tests have shown C4.0 apps compile and run without any changes against the PR and their behavior is the same with and without the PR. I think that is what default gives us in theory. But I could be wrong and will be interested in learning any test case that shows otherwise.

C4.0 apps will work if re-compiled, but we provide ABI guarantees whereby an existing app without recompilation should work. An app built with this change and explicitly using the new parameter will not work with finance 4.0.

>BTW, even a simple fork has not worked for us due to the package sealing restriction introduced in C4.0

You can always copy the entire module into your own package namespace to avoid this issue.

cxyzhang0:
@shamsasari @anthonykeenan Thank you for the info. I did not test ABI so am glad you caught it. So we will work on a fork until 5.0.

PR #5049, opened 2019-04-24
ENT-3422 switch to JUnit 5

r3domfox:
* Adds JUnit 5 version numbers and dependencies to gradle build
* Converts serialization unit tests to use JUnit 5 where possible (remaining with JUnit 4 where tests would require more significant changes to be compatible)
* Introduces `SerializationEnvironmentExtension` as a JUnit 5-compatible replacement `SerializationEnvironmentRule`
* Configures serialization test suite to run tests concurrently.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,6 +39,8 @@ buildscript {
     ext.typesafe_config_version = constants.getProperty("typesafeConfigVersion")
     ext.fileupload_version = '1.3.3'
     ext.junit_version = '4.12'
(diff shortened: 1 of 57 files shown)


rick-r3:
Were you going to raise a PR with just the JUnit upgrade first ?

r3domfox:
Yes, and first I have to stop it breaking IntelliJ on on other modules (it's fine on serialization, but blows up on core)

